PR #152, opened 2021-11-21
CIP-0007 to Proposed

shawnim:
Proposed.
I understand that a better solution might be to apply the a0 pledge benefit fully and only to the pledge and not to the total stake.
I am attempting to move this CIP to Porposed in order to have IOG respond with their plans in regards to the a0 pledge benefit factor.

### DIFF
--- a/CIP-0007/README.md
+++ b/CIP-0007/README.md
@@ -5,7 +5,7 @@ Authors: Shawn McMurdo <shawn_mcmurdo@yahoo.com>
 Discussions-To: 
 Comments-Summary: 
 Comments-URI: 
-Status: Draft
+Status: Proposed
 Type: Standards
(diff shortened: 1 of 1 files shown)


crptmppt (commented):
A change to 'Proposed' should have a defined "Path to 'Active'" - see https://github.com/cardano-foundation/CIPs/tree/master/CIP-0001 
While here the **Path to 'Active**' is understood as (being implemented by IOG) it might be worth specifically adding this in the PR. (meanwhile will see if we can adjust CIP-0001 to be more explicit)

KtorZ (approved):

crptmppt (approved):

crptmppt:
merging under the assumption that the Path to Active is it being implemented

PR #153, opened 2021-11-21
Fair min fees

shawnim:
Note that with the price of ADA being 2x what it was when I proposed this CIP, it is even more critical to adopt this CIP or to prepay min fees.

### DIFF
--- a/CIP-0023/README.md
+++ b/CIP-0023/README.md
@@ -3,7 +3,7 @@ CIP: 23
 Title: Fair Min Fees
 Authors: Shawn McMurdo <shawn_mcmurdo@yahoo.com>
 Comments-URI: https://forum.cardano.org/t/fair-min-fees-cip/47534
-Status: Draft
+Status: Proposed
 Type: Standards
 Created: 2021-02-04
 License: CC-BY-4.0


mark-stopka:
> Note that with the price of ADA being 2x what it was when I proposed this CIP, it is even more critical to adopt this CIP or to prepay min fees.

No it really isn't, unless we implement full fee market, considering we already have large chain utilization, and we're expecting significantly more transactions once dApps go live over next few months.

shawnim:
> > Note that with the price of ADA being 2x what it was when I proposed this CIP, it is even more critical to adopt this CIP or to prepay min fees.
> 
> No it really isn't, unless we implement full fee market, considering we already have large chain utilization, and we're expecting significantly more transactions once dApps go live over next few months.

I don't understand what you mean by "full fee market" in this context or how the number of transactions on the chain is relevant to the issue of the minimum fixed fee (340 ADA( paid to pools that make at least 1 block.

mark-stopka:
> > > Note that with the price of ADA being 2x what it was when I proposed this CIP, it is even more critical to adopt this CIP or to prepay min fees.
> > 
> > 
> > No it really isn't, unless we implement full fee market, considering we already have large chain utilization, and we're expecting significantly more transactions once dApps go live over next few months.
> 
> I don't understand what you mean by "full fee market" in this context or how the number of transactions on the chain is relevant to the issue of the minimum fixed fee (340 ADA( paid to pools that make at least 1 block.

Sorry, I confused this (stakepool fees) proposal with a different proposal debated on twitter few weeks back regarding minimal TX fees.

KtorZ (approved):

DanTup:
@kevinhammond 

Yesterday on the SPO call there were both small and large SPOs advocating for removal of this 340 min fee. It's around half the block reward now making small pools uncompetitive, and we see stake moving from single pool operators to multi-pool operators every epoch.

The response given was that this cannot be removed for "security reasons", but no more information was given (despite several people asking).

If that's true, then I don't understand why this CIP wasn't rejected a long time ago. Why is it being dragged out giving small SPOs the false impression that the playing field might become more level in the future? And where is this documented that we can read about? We wouldn't keep asking about it every SPO call if this information was clear and unambiguous.

If it's not true (and there was some misunderstanding), then how does this progress? IOG said they want _us_ to be in control of parameters, but any time we want a dialog about it, it gets brushed away with vague responses.

SebastienGllmt (approved):

mangelsjover:
@KtorZ according to my notes this PR needed github clean-up before moving it to Last Check. Are we going to wait or shall be added to Last Check for next week meeting?

KtorZ:
@shawnim as discussed in the editor meeting, this PR has a weird git commit history (probably due to some merge or rebase from a wrong revision) so I'll clean up the history to keep only the last commit that is actually the change proposed by the PR (moving it to proposed) and proceed with merging it :+1: 

Meanwhile, I'll also bring it back to the attention of the core team to see if we can have anyone commenting on its feasibility and inclusion in the ledger or, if we should mark this CIP as obsolete given some rebuttal.

shawnim:
> @shawnim as discussed in the editor meeting, this PR has a weird git commit history (probably due to some merge or rebase from a wrong revision) so I'll clean up the history to keep only the last commit that is actually the change proposed by the PR (moving it to proposed) and proceed with merging it +1
> 
> Meanwhile, I'll also bring it back to the attention of the core team to see if we can have anyone commenting on its feasibility and inclusion in the ledger or, if we should mark this CIP as obsolete given some rebuttal.

Thanks @KtorZ !

PR #158, opened 2021-11-24
CIP 34: Network Registry

SebastienGllmt:
Currently Cardano has two easily-usable networks: "mainnet" and "testnet". However, in the future, we expect more networks to exist and so we need some way to refer to these networks to be able to write better multi-network applications and systems.

This will help with standards like CIP30 for the "getNetworkId" endpoint and having a standardized registry for networks allows easy integration into the broader crypto ecosystem via standards like [CAIP-2](https://github.com/ChainAgnostic/CAIPs/blob/master/CAIPs/caip-2.md)


NetworkRegistry/NetworkRegistry.md
```diff
@@ -0,0 +1,65 @@
...
+
+- User-friendly name
+- Network ID
+- Network magic
+- Genesis hash
+
+When representing these networks in a human-readable string, the following format shall be used:
+
+```
+cipxxx:NetworkId-NetworkMagic
```
SebastienGllmt, line 54: I think this is the one point for debate. It's not great that we use "NetworkMagic" which is semi-deprecated and also not human friendly.
I was thinking we could take the first byte of the genesis block hash (8bits) so that we have 12bits total (4096 possibilities) but I'm not sure if this is enough. We could do first two bytes of the genesis block, but at that point I'm not sure if it's particularly better than just using the NetworkMagic

AndrewWestberg: NetworkMagic is not deprecated at all. It is used when doing the ouroboros-network handshake to other nodes and prevents nodes on different networks from talking to each other.

NetworkRegistry/NetworkRegistry.md
```diff
@@ -0,0 +1,65 @@
...
+All entries in this registry should have the following entries:
+
+- User-friendly name
+- Network ID
+- Network magic
+- Genesis hash
+
+When representing these networks in a human-readable string, the following format shall be used:
+
+```
```
SebastienGllmt, line 53: Another option for the format could be bech32. The result would look something like this:
```
cipxx:1pugjyv6y3rn0jc
```

This is less human readable, but not that bad and has the checksum in it which is also nice (but the checksum is a little aggressive for the size of the payload. Checksum is 6 characters and so around half of the resulting string is just the checksum)

NetworkRegistry/NetworkRegistry.md
SebastienGllmt: Good point. I added this to the terminology section 👍

NetworkRegistry/NetworkRegistry.md
```diff
@@ -0,0 +1,65 @@
...
+- The "network ID" included in every address and also optionally present in the transaction body. This only stores 16 possibilities (4 bits)
+- The "network magic" used for Byron addresses and in the handshake with other nodes in the network layer. This is a random 32-bit number
+- The genesis block hash (a 28-byte number)
+
+## Motivation
+
+Blockchains can have multiple deployments of the same codebase. For example:
+
+1. Test networks where the base asset has no value (so devs can test at no cost)
+1. Test networks where the protocol is simplified for ease of testing (ex: Cardano but with a Proof of Authority conensus for stable block production in testing)
```
KtorZ, line 30: Small typo: **conensus** -> consensus

NetworkRegistry/NetworkRegistry.md
```diff
@@ -0,0 +1,65 @@
...
+1. Test networks for new features
+    1. PAB testnet
```
KtorZ, line 32: PAB -> Plutus Application Backend (+ link to repository?)

NetworkRegistry/schema.json
```diff
@@ -0,0 +1,31 @@
...
+            "type": "string"
+          },
+          "NetworkId": {
+            "type": "integer"
+          },
+          "NetworkMagic": {
+            "type": "integer"
+          },
+          "GenesisHash": {
+            "type": "string"
```
KtorZ, line 17: Possibly, we could add `"encoding": "base16", "minLength":  64, "maxLength": 64`

NetworkRegistry/schema.json
```diff
@@ -0,0 +1,31 @@
...
+        },
+        "required": [
+          "Name",
+          "NetworkId",
+          "NetworkMagic",
+          "GenesisHash"
+        ]
+       }
+    },
+    "$schema": "http://json-schema.org/schema#",
```
KtorZ, line 28: https

NetworkRegistry/NetworkRegistry.md
```diff
@@ -0,0 +1,65 @@
...
+    2. Shelley incentivized testnet
+1. Forks that diverge in feature set (the many forks of Bitcoin and Ethereum)
+
+dApps may be deployed on specific testnets that match their criteria and wallets need to know about these networks to know how to behave.
+
+Additionally, having a standardized registry for networks allows easy integration into the broader crypto ecosystem via standards like [CAIP-2](https://github.com/ChainAgnostic/CAIPs/blob/master/CAIPs/caip-2.md)
+
+## Specification
+
+We create a machine-readable registry of networks
```
michaelpj, line 42: Where will this be stored? How will the data be governed? Will it be decentralized?

Or is this proposal just about the format of the data, leaving the actual form of the _registry_ unspecified?

SebastienGllmt: It will be stored here in this CIP. The governance is that anybody who submits a chain ID for any actual network they launched gets it. Similar to other blockchains that have similar IDs with similar governance. This system will not be uniquely associating chain IDs to projects. It's just a way for people to declare they are using an ID in an agreed-upon place to avoid people accidentally colliding.
This is similar to the way other chains deal with chain IDs and similar to other registries like the metadata label registry (CIP10) and SLIP44.

michaelpj: Ah, that makes sense. It would be good to have something about t hat in the text!

What about short-lived networks, e.g. networks spun up for testing and then spun down shortly afterwards? Is the idea that they should use the same format to identify themselves, but not bother to register here? Will that affect how e.g. wallets interact with such networks?

SebastienGllmt: Yeah I guess short-lived networks could just pick a IDs to avoid any chance of collision and they can double-check this CIP to make sure they're not conflicting. We could reserve one of the network IDs for short-term networks and then have them pick a random network magic within it if that sounds like something useful

KtorZ (approved):

crptmppt (approved):

PR #159, opened 2021-11-29
CIP-31: Reference inputs

michaelpj:
This CIP proposes adding "reference inputs" in the style of Ergo's "data inputs".

### DIFF
--- a/CIP-0031/README.md
+++ b/CIP-0031/README.md
@@ -1,0 +1,206 @@
+---
+CIP: 31
+Title: Reference inputs
+Authors: Michael Peyton Jones <michael.peyton-jones@iohk.io>
+Comments-Summary: No comments
+Comments-URI: 
+Status: Draft
+Type: Standards Track
+Created: 2021-11-29
+License: CC-BY-4.0
+---
+
+# Reference inputs
+
+## Abstract
+
+We introduce a new kind of input, a _reference_ input, which allows looking at an output without spending it.
+This will facilitate access to information stored on the blockchain without the churn associated with spending and recreating UTXOs.
+
+## Motivation
+
+Datums in transaction outputs provide a way to store and access information on the blockchain.
+However, they are quite constrained in a number of ways.
+Most notably, in order to access the information which is contained in them, you have to _spend_ the output that the datum is attached to.
+
+This has a number of undesirable features:
+1. Even if the output is recreated after being spent, it is still a _new_ output. Any other user who wishes to look at the data cannot spend the old output (which is gone), but must rather spend the new output (which they will not know about until the next block). In practice this throttles some applications to one "operation" per block.
+2. Looking at the _information_ in the datum requires _spending_ the output, which means that care must be taken over the distribution of any funds in the output (possibly requiring scripts), and spending conditions must be met. This is overly stringent, inconvenient, and expensive.
+
+We would like to have a mechanism for accessing the information in datums that avoids these issues.
+
+### Use cases
+
+Here are some cases where we expect this to be helpful.
+
+1. Inspecting the state (datum, or locked value) of an on-chain application without having to consume the output, e.g. checking the current state of a stablecoin state machine.
+2. On-chain data providers that store data in outputs, to be referenced by other scripts.
+3. (With inline datums) Creating a single UTXO with data to be used in multiple subsequent transactions, but only paying the cost for submitting it once.
+
+### Enabling further improvements
+
+This proposal was designed in tandem with two further proposals which make use of reference scripts, CIP-32 (Inline datums) and CIP-32 (Reference scripts).
+A full assessment of the worth of this proposal should take into account the fact that it is an enabler for these other proposals.
+
+## Specification
+
+### Reference inputs
+
+We introduce a new kind of transaction input, a reference input.
+Transaction authors specify inputs as either normal (spending) inputs or reference inputs.[^1]
+
+A reference input is a transaction input, which is linked to a particular transaction output as normal, except that it _references_ the output rather than _spending_ it. That is:
+
+- A referenced output must exist in the UTXO set.
+- Any value on a referenced output is _not_ considered when balancing the transaction.
+- The spending conditions on referenced outputs are _not_ checked, nor are the witnesses required to be present. 
+    - i.e. validators are not required to pass (nor are the scripts themselves or redeemers required to be present at all), and signatures are not required for pubkey outputs.
+- Referenced outputs are _not_ removed from the UTXO set if the transaction validates.
+- Reference inputs _are_ visible to scripts.
+
+Finally, a transaction must _spend_ at least one output.[^2] 
+
+[^1]: Note in particular that this proposal does not affect _outputs_ in any way.
+[^2]: This restriction already exists, but repeat it here for clarity.
+
+### Script context
+
+Scripts are passed information about transactions via the script context.
+The script context therefore needs to be augmented to contain information about reference inputs.
+
+Changing the script context will require a new Plutus language version in the ledger to support the new interface.
+The change in the new interface is: a _new_ field is added to the structure which contains the list of reference inputs.
+
+Old versions of the language will retain the old interface.
+We do not propose to include information about reference inputs in the old interface.
+
+### Optional datums
+
+Today, transactions can contain optional datums that are not required for validation.
+Currently the optional datums must all be pre-images of datum hashes that appear in outputs.
+We change this so that the optional datums can also be pre-images of datum hashes that appear in reference inputs.[^3]
+
+Note that the existing mechanism includes the hash of the optional datum structure in the transaction body, so that additional datums cannot be stripped by an attacker without voiding transaction signatures.
+
+[^3]: Pre-images of datum hashes that appear in _spent_ inputs are of course mandatory.
+
+### CDDL
+
+The CDDL for transaction bodies will change to the following to reflect the new field.
+```
+transaction_body =
+ { 0 : set<transaction_input>    ; inputs
+ ...      
+ , ? 16 : set<transaction_input> ; reference inputs
+ }
+```
+
+## Rationale
+
+The key idea of this proposal is to use UTXOs to carry information. 
+But UTXOs are currently a bad fit for distributing information. 
+Because of locality, we have to include outputs that we use in the transaction, and the only way we have of doing that is to _spend_ them - and a spent output cannot then be referenced by anything else. 
+To put it another way: outputs are resource-like, but information is not resource-like.
+
+The solution is to add a way to _inspect_ ("reference") outputs without spending them. 
+This allows outputs to play double duty as resource containers (for the value they carry) and information containers (for the data they carry). 
+
+### Requirements
+
+We have a number of requirements that we need to fulfil.
+- Determinism
+    - It must be possible to predict the execution of scripts precisely, given the transaction.
+- Locality 
+    - All data involved in transaction validation should be included in the transaction or the outputs which it spends.
+- Non-interference
+    - As far as possible, transactions should not interfere with others. The key exception is when transactions consume resources that other transactions want (usually by consuming UTXO entries).
+- Replay protection
+    - The system should not be attackable (e.g. allow unexpected data reads) by replaying old traffic.
+- Storage control and garbage-collection incentives 
+    - The amount of storage required by the system should have controls that prevent it from overloading nodes, and ideally should have incentives to shrink the amount of storage that is used over time.
+- Optimized storage 
+    - The system should be amenable to optimized storage solutions.
+- Data transfer into scripts
+    - Scripts must have a way to observe the data.
+- Tool and library support
+    - It should not be too difficult for tools and libraries to work with the new system.
+- Hydra
+    - The system should be usable without compromising the needs of Hydra, such as parallel processing of independent parts of the transaction graph.
+
+Reference inputs satisfy these requirements, mostly by inheriting them from the corresponding properties of UTXOs.
+
+- Referencing an output still requires the output to be presented as part of the transaction and be unspent, so determinism is preserved.
+- Referenced outputs appear as transaction inputs, so locality is preserved.
+- Outputs can be referenced multiple times, so we have non-interference even under heavy usage of referencing. Conflicts can occur only if the output is actually spent.
+- Replay protection is inherited because a transaction must always spend at least one UTXO, and UTXOs can only be spent once.
+- Storage control is inherited from control of the UTXO set, via the min UTXO value incentives. Data can be "retired" by spending the input (rather than referencing it), reclaiming the min UTXO value.
+- Optimized storage is inherited directly from work on improving UTXO storage.
+- Data transfer into scripts works automatically since scripts can see transaction inputs.
+- Tools and libraries will only have to deal with a new kind of transaction input, and unless they care about scripts they can likely ignore them entirely.
+- Hydra is able to work with reference inputs (see below).
+
+### Why UTXOs?
+
+There are various approaches to augmenting Cardano with the ability to store and retrieve data more easily.
+All of these require some kind of resource-like system for the _storage_ that is used for data.
+The main argument in favour of the reference input approach is that it reuses our existing resource-like system: UTXOs.
+
+As we've seen above, this allows us to satisfy most of our requirements for free.
+Any other approach would need new solutions to these problems.
+
+As a brief example, suppose we wanted to instead implement data storage as an on-chain, hash-indexed data store (a proposal we considered).
+
+Then we would need to answer a lot of questions:
+- How is the data accessed and retrieved? Are these stateful operations? Does this make transaction ordering more significant, threatening determinism?
+- How is the storage usage controlled? Do people pay for storing data? Do they pay for a fixed period, or perpetually? How is data retired?
+- How is the storage going to be implemented? How will this affect node memory usage?
+- How are scripts going to access the data in the store?
+- How will tools interact with and visualize the store and changes to it?
+
+### How should we present the information to scripts?
+
+Scripts definitely need to see reference inputs.
+But we have at least two options for how we represent this in the script context: put the reference inputs in their own field; or include them with the other inputs, but tag them appropriately.
+
+Keeping them separate seems wise given the potential for confusing reference inputs for normal inputs.
+That would be quite a dangerous programming error, as it might lead a script to believe that e.g. an output had been spent when in fact it had only been referenced.
+
+### Accessing the datums of reference inputs
+
+Currently this proposal has somewhat limited utility, because datums in outputs are just hashes, and the spending party is required to find and provide the preimage themselves.
+Therefore the CIP-32 proposal for inline datums is complementary, as it allow UTXOs to carry the data itself rather than a hash.
+
+In the mean time (or if CIP-32 is not implemented), we still want users to be _able_ to provide the datums corresponding to datum hashes in reference outputs, so that at least it is possible to reference the data, albeit clumsily.
+The easiest solution here is to reuse the existing mechanism for providing pre-images of datum hashes which appear in outputs.
+
+Providing datum hash pre-images remains optional, since there are reasons to use reference inputs even without looking at the datum, e.g. to look at the value in an output.
+
+### Accessing the value in outputs
+
+The motivation of this proposal mainly requires looking at the _datums_ of outputs.
+But reference inputs allow us to do more: they let us look at the _value_ in an output.
+
+This is a bonus feature, in that we don't think it's critical, but it is still useful.
+If a user wants to use reference inputs to actually look at the _state_ of an on-chain application, the amount of value locked in the script is likely to be an important part of the state, so it's valuable to allow access to that.
+
+### Hydra
+
+We want reference inputs to be usable inside Hydra heads.
+This raises a worry: since Hydra processes independent parts of the transaction graph in parallel, what happens if it accepts one transaction that references an output, and one that spends the output, but then the first transaction gets put into a checkpoint before the second?
+
+Fortunately, Hydra already has to deal with double-spend conflicts of this kind (although in the naive protocol this results in a decommit).
+This proposal simply introduces a new kind of conflict, a reference-spend conflict.
+Reference-spend conflicts should be handled by the same mechanism that is used to handle double-spend conflicts.
+
+### Controlling referencing
+
+One thing that a user might want to do is to control who can reference an output.
+For example, an oracle provider might want to only allow a transaction to reference a particular output if the transaction also pays them some money.
+
+Reference inputs alone do _not_ provide any way to do this.
+
+### Related work
+
+Reference inputs are very similar to Ergo's "data inputs".
+We chose to name them differently since "data" is already a widely used term with risk for confusion.
+We might also want to introduce other "verb" inputs in future, e.g. "check" inputs that check that an output _can_ be spent, but don't actually spend it.


CIP-0031/README.md
```diff
@@ -0,0 +1,206 @@
...
+A reference input is a transaction input, which is linked to a particular transaction output as normal, except that it _references_ the output rather than _spending_ it. That is:
+
+- A referenced output must exist in the UTXO set.
+- Any value on a referenced output is _not_ considered when balancing the transaction.
+- The spending conditions on referenced outputs are _not_ checked, nor are the witnesses required to be present. 
+    - i.e. validators are not required to pass (nor are the scripts themselves or redeemers required to be present at all), and signatures are not required for pubkey outputs.
+- Referenced outputs are _not_ removed from the UTXO set if the transaction validates.
+- Reference inputs _are_ visible to scripts.
+
+Finally, a transaction must _spend_ at least one output.[^2]
```
L-as, line 61: Even if you did somehow "allow" this, wouldn't this be impossible since you need to pay the fee somehow?

CIP-0031/README.md
```diff
@@ -0,0 +1,206 @@
...
+The easiest solution here is to reuse the existing mechanism for providing pre-images of datum hashes which appear in outputs.
+
+Providing datum hash pre-images remains optional, since there are reasons to use reference inputs even without looking at the datum, e.g. to look at the value in an output.
+
+### Accessing the value in outputs
+
+The motivation of this proposal mainly requires looking at the _datums_ of outputs.
+But reference inputs allow us to do more: they let us look at the _value_ in an output.
+
+This is a bonus feature, in that we don't think it's critical, but it is still useful.
```
L-as, line 183: It's definitely critical. Information without any proof of authenticity is not worth much.

jfischoff: This section does not concern proof of authenticity. The OP was suggesting the assets, Ada and non-native tokens, stored at UTxO could also be accessed. 

The motivating use case, access state in a datum, can be addressed without the ability to access the value on the UTxO. I agree with the OP. This is a bonus feature.

CIP-0031/README.md
```diff
@@ -0,0 +1,206 @@
...
+The solution is to add a way to _inspect_ ("reference") outputs without spending them. 
+This allows outputs to play double duty as resource containers (for the value they carry) and information containers (for the data they carry). 
+
+### Requirements
+
+We have a number of requirements that we need to fulfil.
+- Determinism
+    - It must be possible to predict the execution of scripts precisely, given the transaction.
+- Locality 
+    - All data involved in transaction validation should be included in the transaction or the outputs which it spends.
```
Quantumplation, line 114: Considering this proposal, should this be `or the previous outputs which it references.`

CIP-0031/README.md
L-as: I'm not sure I understand what you mean? This CIP in isolation, without considering the other ones, is completely useless if you can not see the value contained in the UTXO.

Quantumplation: @L-as I think this comes down to a confusing over-loading of the word "Value".

In one sense, "Value" could mean "The data which hashes to the datum hash", in which case the previous section addresses how this data is made available.

In another sense, "Value" could mean "The ADA and cardano-native-tokens stored in the UTXO".  So, beyond just seeing that the referenced datum has X hash and Y data, you could also see that it contains 133 ADA and 7 HOSKY tokens.

I believe the "bonus feature, in that we don't think it's critical" here refers to this second meaning.

CIP-0031/README.md
```diff
@@ -0,0 +1,206 @@
...
+
+Here are some cases where we expect this to be helpful.
+
+1. Inspecting the state (datum, or locked value) of an on-chain application without having to consume the output, e.g. checking the current state of a stablecoin state machine.
+2. On-chain data providers that store data in outputs, to be referenced by other scripts.
+3. (With inline datums) Creating a single UTXO with data to be used in multiple subsequent transactions, but only paying the cost for submitting it once.
+
+### Enabling further improvements
+
+This proposal was designed in tandem with two further proposals which make use of reference scripts, CIP-32 (Inline datums) and CIP-32 (Reference scripts).
```
flyerman, line 42: 🎨 Reference Script is CIP-33

CIP-0031/README.md
mkazlauskas: Implicit coin (reward withdrawals + deposit reclaims) can be used to pay the fee, so you could allow it

CIP-0031/README.md
```diff
@@ -0,0 +1,206 @@
...
+### Reference inputs
+
+We introduce a new kind of transaction input, a reference input.
+Transaction authors specify inputs as either normal (spending) inputs or reference inputs.[^1]
+
+A reference input is a transaction input, which is linked to a particular transaction output as normal, except that it _references_ the output rather than _spending_ it. That is:
+
+- A referenced output must exist in the UTXO set.
+- Any value on a referenced output is _not_ considered when balancing the transaction.
+- The spending conditions on referenced outputs are _not_ checked, nor are the witnesses required to be present.
```
mkazlauskas, line 56: > A referenced output must exist in the UTXO set.
> The spending conditions on referenced outputs are _not_ checked, nor are the witnesses required to be present.

This might be a breaking change. I think you can currently imply owner of utxo just by looking at output addresses. An alternative of adding a new field in transaction body would be breaking too, but in an obvious way - this is more stealthy.

CIP-0031/README.md
michaelpj: I'm not sure whether we're disagreeing, but actually I was wrong here, this *is* very important. The reason for that is that the main way we have of identifying *specific* UTXOs is via their (value) contents, in particular we usually use tokens for this. So if you care about looking at particular UTXOs this is very important. I'll amend.

CIP-0031/README.md
michaelpj: Yes!

CIP-0031/README.md
michaelpj: I've attempted to clarify in the text.

CIP-0031/README.md
michaelpj: You could potentially rely on a mechanism like this. Nonetheless, transactions today are already required to spend at least one UTXO, even if e.g. they could cover the fee with withdrawals. We simply don't change that restriction here.

delicious-lemon:
I think we need a feature like this but I have a suggestion for an alternate implementation. My concern is that we may regret losing the _can only be read once_ affine nature of UTXOs.

We could implement something that looks more like a Reader Monad than a reference input. To do that we would need to introduce the concept of a constrained input and defer some construction of the Tx to the node.

All we need to say is that we want any TxOut that matches exactly the data we want to read

```
data TxOut = TxOut {
    txOutAddress   :: Address,
    txOutValue     :: Value,
    txOutDatumHash :: Maybe DatumHash
    }
```

and so long as the transaction reproduces an identical TxOut we can form a Reader Monad like chain through which the state is threaded.

If the transaction does not reproduce exactly this output then further reading transactions will fail in the node transaction construction phase since they cannot find a matching input.

WhatisRT:
> I think we need a feature like this but I have a suggestion for an alternate implementation.

While it's nice to keep some of the UTXO's current properties, I see three problems with this proposal:

1. It introduces a non-deterministic aspect into the UTXO. It is not entirely specified here how looking up an output would work, but there are essentially two options: either the 'old' `TxIn` and the 'new' `TxIn` both work, or only the 'old' one works (there's also the option to only do these things for the current block, and all future blocks use the 'new' `TxIn`). If we disallow for the 'new' `TxIn` to be used, there can be an unexpected failure for users that didn't know about this feature and unexpectedly triggers it. And if we allow for both, then an output can obtain a large number of `TxIn`'s that refer to it.
2. It's a lot more difficult to implement than what this CIP proposes.
3. It requires active collaboration. This could be enforced by validators, but that would make it expensive.

michaelpj:
@delicious-lemon 

> To do that we would need to introduce the concept of a constrained (or content addressable) input and defer some construction of the Tx to the node.

This is an interesting idea, but it loses another property that we care about: that the transaction specifies exactly what it does, and is self-contained. For example, losing that threatens the determinism of script execution (I think we could *probably* keep it working, but we'd have to think carefully about it).

You could imagine implementing something like this as a layer 2 solution, perhaps, where "partial" transactions are elaborated into "fully-specified" transactions in order to be included in the base settlement layer. But I feel less positive about it as a feature of layer 1.

Plus, this "constrained script inputs" feature is much more powerful than just reference inputs. It feels odd to implement it just to do referencing. Perhaps there are other use cases that make it compelling, but otherwise it feels like we're adding a powerful feature somewhat blindly, which might lead to unexpected outcomes.

> and so long as the transaction reproduces an identical TxOut we can form a Reader Monad like chain through which the data is threaded

Notably, this means you need to implement more of the logic in the script that locks the output: it has to insist that you produce a matching script when you reference it.

> we get "controlled referencing" automatically since you must be able to spend the UTXO to read it

It's not even clear if "controlled referencing" is desirable. And if we did do it, we'd probably want to use a different script than the script that controls spending: taking the oracle example, you don't want the _same_ script to control who can use the data and who can reference it. (Well, maybe you could make it work with a clever script and redeemers.)

At any rate, I think the design there is less clear, and your proposal actually gets rid of *uncontrolled* referencing, which I think *is* definitely useful.

-----

Anyway, I think this is definitely an interesting idea and I'd encourage you to write it up if you're keen.

delicious-lemon:
>     1. It introduces a non-deterministic aspect into the UTXO. It is not entirely specified here how looking up an output would work, but there are essentially two options: either the 'old' `TxIn` and the 'new' `TxIn` both work, or only the 'old' one works (there's also the option to only do these things for the current block, and all future blocks use the 'new' `TxIn`). If we disallow for the 'new' `TxIn` to be used, there can be an unexpected failure for users that didn't know about this feature and unexpectedly triggers it. And if we allow for both, then an output can obtain a large number of `TxIn`'s that refer to it.

I am not sure what you mean. Perhaps there is some misunderstanding. I propose we add the ability to look up TxOutRefs on the ledger by the hash of the TxOut. This would be an extension to the current functionality so there is no danger that it will interfere with how anything currently works. This would be an opt-in feature since this requires storing additional data on the ledger. i.e. you make a Tx with an output and flag that it should be stored in such a way that its TxOutRef can be looked up by its TxOutHash. I propose that we can specify inputs and outputs using these TxOutHash references such that we can define cheap Reader monad like threads. There is no way this feature can be unexpectedly triggered by those that don't know about it.

On non-determinism - that's the whole point. We get non-determinism w.r.t. TxOutRef but this doesn't matter since we can still verify that the Tx will work if a TxOut can be found with the specified TxOutHash. The lookup for a TxOutHash will be a similar cost to the lookup for a TxOutRef. If this step is successful validation proceeds as normal otherwise we fail as normal.

>     2. It's a lot more difficult to implement than what this CIP proposes.
Why do you believe this? I state the opposite to be true.

>     3. It requires active collaboration. This could be enforced by validators, but that would make it expensive.
I am not sure what you mean by this. I am not proposing we change how validation works.

delicious-lemon:
@michaelpj 

Perhaps constrained inputs is the wrong name for this. That's a more powerful bag of magic that I agree we shouldn't look into for layer 1. I'm suggesting we could specify inputs and outputs to be TxOuts that already exist on the ledger with exactly matching Address, Value, Datum, identified by a hash of this triple. The node can complete this Tx closure by computing the TxOutRefs as required. This keeps validation deterministic and allows us just enough non-determinism to get reference-input like behaviour without having reference inputs.

I'm not sure whether this would be more work to implement than reference inputs. I feel like it may be considered less powerful though since we would lose uncontrolled referencing. My concern is that losing the affine read-once nature of UTXOs might make the computation model more complex and thus harder to audit and prove properties for - though I have no evidence for this, it's just a feeling.

> Anyway, I think this is definitely an interesting idea and I'd encourage you to write it up if you're keen.

I'll put some markdown together. :+1:

L-as:
I think that "content-addressed UTXOs" is a good idea, because the location/transaction of the UTXO doesn't matter, the only thing that matters is its content. I don't see why determinism would be lost if you have to specify the exact content of the UTXO, but if you support only specifying it partially, e.g. a UTXO with X token, without caring about the datum, that is a lot harder to make deterministic.

However, this doesn't replace the need for a reference input, and it will not work well at all with CIP-33. The two extensions to the ledger aren't mutually exclusive as they don't conflict, so I don't think it's a good reason not to add this feature to the ledger.

delicious-lemon:
@L-as I think CIP-33 could work with "content-addressed UTXOs". We would just need a way of saying "this is the hash of the script, it should be in this UTXO" and have a mechanism for pulling it out of the "content-addressed UTXO".

I am in agreement that these two ideas could coexist in the system. Read-only access to the ledger would be very useful. Additionally - perhaps content-addressed UTXOs are more powerful than I initially estimated - consider a UTXO that represents a DEX as a content-addressed UTXO supporting an order book with execution/matching decided by the node. Users could place an order relative to a content-addressed UTXO that they speculate will come into existence and hope that a node will be able to engineer a sequence that supports their transaction. The node could be performing some application specific arbitrage operation to enable the sequencing.

michaelpj:
> I'm not sure whether this would be more work to implement than reference inputs.

FWIW my belief is that yes, this would be much more work to implement. Reference inputs requires a tag on inputs, some small changes to the ledger rules, and some small changes to the transaction context. "Constrained inputs" requires changes to how nodes construct blocks, changes to the transaction format, new kinds of entity, possibly a content-addressed lookup store... lots of things.

WhatisRT:
@delicious-lemon 
1. Yes, I misunderstood your proposal there, but of course the problem of non-determinism is still present. Which output is picked if there are multiple outputs with the same content? Maybe some of them are spendable with the transactions witnesses and some are not. One could of course make up a consistent scheme (use the one with the smallest `TxIn` that the transaction can spend) but we care a lot about knowing exactly what a transaction will do before sending it. Even if we can make sure that Plutus scripts won't change their result, this is still a property that I'd argue we care more about than the read once nature of the UTxO.
2. I've already written the spec for this proposal, which is very small. Your proposal does touch a lot more places.
3. This is the same point as 'controlled referencing'. My argument is that to allow anyone to use a reference would require a Plutus validator to force the person using it to also return it to its place. Which means that then using it for CIP-33 could then pull in a bunch of extra scripts that are just used for validating the reference. It should still be possible to not have to supply any scripts in that scenario, but there's a lot of expensive overhead.

CIP-0031/README.md
```diff
@@ -0,0 +1,210 @@
...
+Transaction authors specify inputs as either normal (spending) inputs or reference inputs.[^1]
+
+A reference input is a transaction input, which is linked to a particular transaction output as normal, except that it _references_ the output rather than _spending_ it. That is:
+
+- A referenced output must exist in the UTXO set.
+- Any value on a referenced output is _not_ considered when balancing the transaction.
+- The spending conditions on referenced outputs are _not_ checked, nor are the witnesses required to be present. 
+    - i.e. validators are not required to pass (nor are the scripts themselves or redeemers required to be present at all), and signatures are not required for pubkey outputs.
+- Referenced outputs are _not_ removed from the UTXO set if the transaction validates.
+- Reference inputs _are_ visible to scripts.
```
WhatisRT, line 59: Together with CIP-33, reference inputs would then do two things: Be part of the context passed to scripts, and if they contain a reference script, that would get added to the witness set. This combination is a bit arbitrary. I worry about a future where we suddenly need to have different types of reference inputs because we need to be able to reference some of the things but not others.

amin1419: michaelpj:mpj/reference-inputs

CIP-0031/README.md
GeorgeFlerovsky: I agree that this is critical. NFTs are a major way to pinpoint UTXOs of interest.

Without accessing a particular NFT inside a reference input's "Value", I think it would be hard to prove to a script that you're providing it with authentic data, because the adversary can always create UTXOs with arbitrary data under the script address.

CIP-0031/README.md
GeorgeFlerovsky: > Together with CIP-33, reference inputs would then do two things: Be part of the context passed to scripts, and if they contain a reference script, that would get added to the witness set. This combination is a bit arbitrary. I worry about a future where we suddenly need to have different types of reference inputs because we need to be able to reference some of the things but not others.

If a transaction's witnesses map contains additional (script hash, script source) pairs, does it really make a difference? All of the transaction's inputs were intentionally locked by specific script hashes, so they would never unintentionally refer to these additional scripts from the reference inputs...

michaelpj: > Be part of the context passed to scripts, and if they contain a reference script, that would get added to the witness set.

My intention was that with CIP-33 _any_ input that corresponds to an output with a reference script would see it added to the witness set. So I claim that reference inputs still only do one (conceptual) thing: they let you _look at_ all the information in an output. It seems reasonable to me that looking at the information in an output that contains a reference script should let you use the reference script as a witness.

Rephrasing your worry, though, what you're suggesting is that we might want to e.g. restrict the information that a reference input lets us look at. I can't see a reason for that, but maybe there is one.

CIP-0031/README.md
```diff
@@ -0,0 +1,210 @@
...
+### Reference inputs
+
+We introduce a new kind of transaction input, a reference input.
+Transaction authors specify inputs as either normal (spending) inputs or reference inputs.[^1]
+
+A reference input is a transaction input, which is linked to a particular transaction output as normal, except that it _references_ the output rather than _spending_ it. That is:
+
+- A referenced output must exist in the UTXO set.
+- Any value on a referenced output is _not_ considered when balancing the transaction.
+- The spending conditions on referenced outputs are _not_ checked, nor are the witnesses required to be present.
```
asutherlandus, line 56: You may want to clarify that the spending conditions are checked if the referenced output is spent? I assume this is the case, since not doing so constitutes a DOS attack.

CIP-0031/README.md
```diff
@@ -0,0 +1,210 @@
...
+Fortunately, Hydra already has to deal with double-spend conflicts of this kind (although in the naive protocol this results in a decommit).
+This proposal simply introduces a new kind of conflict, a reference-spend conflict.
+Reference-spend conflicts should be handled by the same mechanism that is used to handle double-spend conflicts.
+
+### Controlling referencing
+
+One thing that a user might want to do is to control who can reference an output.
+For example, an oracle provider might want to only allow a transaction to reference a particular output if the transaction also pays them some money.
+
+Reference inputs alone do _not_ provide any way to do this.
```
asutherlandus, line 204: It seems like having some sort of access control over the datum in a reference input would be a very common use case. It might be good to mention how to do this or at least the rationale for this decision since to the uneducated (me) it seems like a simple matter of checking the witness on the reference input.

CIP-0031/README.md
L-as: s/referenced output/reference input/ would be better IMO

asutherlandus: indeed

michaelpj: I can make it explicit. Since there are no changes to outputs, there is no change to the behaviour of spending an output.

CIP-0031/README.md
michaelpj: I'll incorporate some of the discussion from this PR into the text.

michaelpj:
I've updated the text with some clarifications and a small discussion about controlling referencing.

CIP-0031/README.md
```diff
@@ -0,0 +1,236 @@
...
+- The same script would be used for both referencing and spending, overloading the meaning of the validator script. This is still _usable_, however, since the redeemer could be used to indicate which action is being taken.
+- We would need a flag on outputs to say "this output cannot be referenced, but only checked". Exactly what this should look like is an open question, perhaps it should be generic enough to control all the possible ways in which an output might be used (of which there would be three).
+
+#### Referencing conditions
+
+"Referencing conditions" would mean adding a new field to outputs to indicate under what conditions the output may be referenced.
+This could potentially be an entire additional address, since the conditions might be any of the normal spending conditions (public key or script witnessing).
+
+However, this would make outputs substantially bigger and more complicated.
+
```
L-as, line 231: I think it's cleaner to add an optional "reference validator script" field rather than having "check inputs". How much overhead will an empty field add to the serialisation of a UTXO?

michaelpj:
Dear community, I need some input. It is unclear to me whether controlled referencing (as defined in the CIP text currently) is a key feature for people or not.

Please react to this comment with your feelings: 
- :+1: for "reference inputs are useful to me regardless of controlled referencing"
- :confused: for "reference inputs are useful to me without controlled referencing, but I *also* want controlled referencing"
- :-1: for "reference inputs are useless to me without controlled referencing"

CIP-0031/README.md
asutherlandus: I also prefer the optional "reference validator script" field, it does seem cleaner. To be really explicit about the behavior: If the reference validation script was absent it would be the same as validation passing. I like the model of having separate validators for the two semantic actions of referencing vs spending.

GeorgeFlerovsky:
Reference inputs _without_ controlled referencing are useful in applications where the referenced data is _not_ monetized (i.e. compensation not expected to the data originator), because the dApp is referencing its own intrinsic data. Examples:

- Users accessing a dApp's shared global state, so that they can actually use the dApp. This state does not need to be monetized directly itself, because usage of the dApp can be monetized separately via the usual spending validators.
- Parallelized computation where one thread needs to know how far the other threads have progressed, as of the current active UTXO set.
- Updateable dApps (using this CIP and https://github.com/cardano-foundation/CIPs/pull/161), where general dApp scripts reference a script from a UTXO that holds a particular dApp NFT. The reference script can be updated via dApp governance, whereby community voting can allow the dApp NFT UTXO to be consumed and its NFT to be moved to a new UTXO with an updated script.

Referencing conditions (i.e. separate conditions for referencing than spending) are useful in applications where the referenced data itself is being monetized, because the data comes from outside of the dApps that use it. In other words, the data provider does not derive sufficient incentive from the benefits that users gain from using the data, and must be compensated separately in order to provide the data. Examples:

- Oracles that inject real world data into the blockchain (e.g. financial data, world events, births/deaths), as a service.
- One dApp wants to allow other dApps to "officially" subscribe to its state, with designated UTXOs under controlled referencing.

The main benefit that referencing conditions provide is the ability for a user to prove on-chain that she met the data provider's terms for the data that she has used. For example, upon referencing a UTXO with referencing conditions, the user can mint an NFT that will witness to subsequent transactions that the referencing conditions have been met in this transaction. Such NFT witnesses would allow:

- The data provider to accept liability claims for its data only from users who hold such NFT witnesses of compliant use.
- One dApp community to mandate that its scripts must use another dApp's designated UTXOs with referencing conditions. In other words, disallow the unsanctioned use of the other dApp's data.

On the other hand, the usefulness of "check inputs" controlled referencing (i.e. referencing allowed if spending conditions are met) varies by context:

- In the multi-user context, it's useless because it means that anyone who can reference the input can also spend/destroy it. In my opinion, granting spending rights to users defeats the purpose of reference inputs in the multi-user context, because now any user can unilaterally prevent other users from referencing the UTXO in subsequent blocks.
- In the single-user context, it can be seen as a degenerate case of referencing conditions, whereby the UTXO owner can legitimately reference the UTXO without meeting its referencing conditions (if any), and can mint an NFT witness of having done so that can be shown to any scripts that require such witnesses.

michaelpj:
@GeorgeFlerovsky thank you for that useful summary. I agree with most of what you wrote. A couple of things.

A key use case for reference inputs is to support CIP-33 (reference scripts). I think those would be pretty useful with just reference inputs, although there are perhaps interesting opportunities with controlled referencing too.

On check inputs:
> In the multi-user context, it's useless because it means that anyone who can reference the input can also spend/destroy it.

That's not necessarily true, as I hinted in the text. You can use the redeemer to control it:
- Redeemer is `data Action = Check | Spend ...`
- Validator checks that
    - If the redeemer is `Check`, then the input appears in the list of check inputs, and the referencing conditions are met
    - If the redeemer is `Spend`, then the input appears in the list of normal inputs, and the spending conditions are met

So you can encode different referencing and spending conditions into a single validator.

(Aside: this makes me realise that it is *not* true that check inputs give a proof that you *could* spend the output... the example above gives exactly a case where you could "check" an output but not spend it!)

asutherlandus:
I think that is an interesting suggestion, but am wondering if these two are not equivalent in terms of side effects. I believe one reason for the separate fields:

> We introduce a new kind of transaction input, a reference input.
> Transaction authors specify inputs as either normal (spending) inputs or reference inputs.

Is for the transaction author to specify the intended semantics to the interpreter to determine if the input participates in the transaction balance check and if the UTXO gets marked spent.

Which makes me wonder if this idiom:
> Redeemer is data Action = Check | Spend ...
would get used very much in practice, unless for some reason you wanted to handle both cases with only 1 script.

I believe you could combine them if you added a flag to the Context and it was visible to the interpreter. The current proposal seems very economical, but perhaps not as explicit.

Please correct me if I got any of the semantics wrong. I'm not super confident in my understanding of how this all works.

anton-k:
I'm for separate reference input field in the context and not to mix them with ordinary inputs. And can they be not a packed to a set on a node level? So that it's possible to just index them as it's possible for outputs. That would be great to have. Fixing that for inputs is also great to have. But it's a different story.

michaelpj:
Thanks for the input, everyone. Given the timelines on which we'd like to do this work, the lack of design consensus, and the lack of anyone saying that the lack of controlled referencing makes this CIP worthless for them, I'm going to leave it **out of scope** for this CIP. We can revisit it in future.

With that said, I think this is ready to be merged as Draft. I will revisit it and update it to Active once the implementation has progressed and e.g. the CDDL is pinned down.

--------------

> I'm for separate reference input field in the context and not to mix them with ordinary inputs.

Yep, that's the current proposal.

> And can they be not a packed to a set on a node level? So that it's possible to just index them as it's possible for outputs. 

That's definitely out of scope. For this proposal I think the interface should be the same as for normal inputs, and if we change them we should change them both. Perhaps you should write a CIP :)

peterVG:
This CIP will be ground-breaking for Cardano oracles which is what I'm working on. There's no point continuing with the current architecture restrictions (having to spend transaction outputs to read data + only one script can read per block) if CIP-31 is nearby. I realize there are many extenuating factors but it sounds like IOG is fast-tracking this CIP, is that right @michaelpj? I'm trying to determine whether I can expect this CIP feature in the next fork/chain update. AFAIK those are scheduled for February and then June right?

InderKumarRathore (approved):

KtorZ:
> it sounds like IOG is fast-tracking this CIP,

No one is fast-tracking any CIP :blush: ! Plus, there's a clear separation between CIPs (which are proposals of possible solutions) and actual implementations. While IOG is seemingly working on implementing CIP-0031, CIP-0032 and CIP-0033; they are still following the same process as others CIPs, going through multiple rounds of reviews and validations by editors and the community :+1:

peterVG:
Well, to be clear. I would love for IOG to fast track this particular CIP. It can't come on-chain soon enough IMHO. I say "fast track" because John Woods, Director of Cardano Architecture at IOG has publicly mentioned this CIP and it's two related CIPs twice now in Cardano 360 updates. That's how i found out about them. I understand and respect that all CIPs have to go through the same review and editorial process but I also realize that the individuals involved can choose to prioritize that work for whatever reason (i.e. "fast track").

michaelpj:
This isn't the place to discuss timelines.

peterVG:
Forgive my ignorance @michaelpj, I am brand new to the CIP process. I am an interested party in this CIP, my oracle project will benefit greatly from it. I  am trying to plan my own Cardano development activities accordingly. I currently have no sense of whether to expect to see this CIP live in 1 month or 1 year. Please direct me to the correct forum to ask about timelines. Thank you.

rphair:
@peterVG there would be others willing to discuss the timeline(s) & other advocacy in this forum category if you create a thread there: https://forum.cardano.org/c/developers/cips/122

If & when these forum discussion threads generate insight into a CIP itself, sometimes the authors will also include them in the `Comments-URI` section in the CIP header.

CIP-0031/README.md
```diff
@@ -0,0 +1,236 @@
...
+Providing datum hash pre-images remains optional, since there are reasons to use reference inputs even without looking at the datum, e.g. to look at the value in an output.
+
+### Accessing the value locked in outputs
+
+The motivation of this proposal mainly requires looking at the _datums_ of outputs.
+But reference inputs allow us to do more: they let us look at the _value_ locked in an output (i.e. how much Ada or other tokens it contains).
+
+This is actually a very important feature.
+Since anyone can lock an output with any address, addresses are not that useful for identifying _particular_ outputs on chain, and instead we usually rely on looking for particular tokens in the value locked by the output.
+Hence, if a script is interested in referring to the data attached to a _particular_ output, it will likely want to look at the value is locked in the output.
```
koslambrou, line 188: Fix: look at the value **that** is locked in the output

michaelpj:
I updated the proposal to follow the principle that we should never silently omit information. That means that instead of silently omitting reference inputs when creating the context for old scripts, instead we will fail a transaction which spends from an old script and includes reference inputs in phase 1.

michaelpj:
See also https://github.com/cardano-foundation/CIPs/pull/161#issuecomment-1030181440.

crptmppt:
This was discussed at the [Editor meeting 38](https://www.crowdcast.io/e/cip-editors-meeting-38) - it is assumed on hold until further development (do flag if ready to review again)

KtorZ (approved):
Reviewed addition about forbidding usage of reference inputs in conjunction with old Plutus script (i.e V1). This makes sense; limiting foot-guns when building a DeFi ecosystem definitely makes sense.

SebastienGllmt (approved):

mangelsjover:
Hi @michaelpj 

Could you please add a justification of why this should be added as a Phase 1?

Thank you!

M.

M.

KtorZ:
@michaelpj in the last editor, if you recall, someone in the chat brought up the question of why should the ledger fail during phase-1 validation when presented with reference inputs and a PlutusV1 script; So the idea was to provide a rationale for that, but, I see that the rationale is already there actually:

```
- Omitting information may lead scripts to make assumptions about the transaction that are untrue; for this reason we prefer not to silently omit information as a general principle.
- That leaves us only one option: reject transactions where we would have to present information about reference inputs to old scripts.
```

Thus, happy to proceed with that one as discussed :+1:

L-as:
Question @michaelpj: Can a transaction reference a UTXO if it's consumed in the same block by another transaction?

WhatisRT:
Yes, blocks are essentially irrelevant to what's going on in the UTxO set. As long as the transaction that references an output comes before the one that spends it everything works, even if they are in the same block. It also works in reverse, an output can be created and referenced in the same block as well.

But if you send transactions that depend on each other in such short succession that they might end up in the same block, there's some chance that they might not be in the order you've sent them, which would mean that only one of them would actually end up on the chain (and the other one would need to be submitted again).

PR #160, opened 2021-11-29
CIP-32: Inline datums

michaelpj:
This CIP proposes adding "inline datums" which allow attaching a datum itself rather than a datum hash to an output.

### DIFF
--- a/CIP-0032/README.md
+++ b/CIP-0032/README.md
@@ -24,7 +24,8 @@ However, in practice datums are implemented by attaching _hashes_ of datums to o
 
 This is quite inconvenient for users. 
 Datums tend to represent the result of computation done by the party who creates the output, and as such there is almost no chance that the spending party will know the datum without communicating with the creating party.
-That means that either the datum must be communicated between parties off-chain, or communicated on-chain by attaching it to the transaction that creates the output (which is also inconvenient in that the spending party must watch the whole chain to spot it).
+That means that either the datum must be communicated between parties off-chain, or communicated on-chain by including it in the witness map of the transaction that creates the output ("extra datums").
+This is also inconvenient for the spending party, who must watch the whole chain to spot it.
 
 It would be much more convenient to just put the _datum itself_ in an output, which is what we propose.
 
@@ -94,25 +95,25 @@ In practice, what is implemented here may depend on whether the UTXO-on-disk wor
 
 ### Other modes of specifying datums
 
-We could deprecate the other methods of specifying datums (datum hashes or datum hashes+optional datums).
+We could deprecate the other methods of specifying datums (datum hashes or datum hashes+extra datums).
 However, the other approaches also have some advantages.
 
 - Transmission costs: creator pays versus consumer pays
     - Inline datums: creator pays
     - Datum hashes: consumer pays
-    - Datum hashes+optional datums: both pay
+    - Datum hashes+extra datums: both pay
 - Min UTXO value costs
     - Inline datums: depends on data size
     - Datum hashes: fixed cost
-    - Datum hashes+optional datums: fixed cost
+    - Datum hashes+extra datums: fixed cost
 - Privacy
     - Inline datums: datum is immediately public
     - Datum hashes: datum is not public until consumed
-    - Datum hashes+optional datums: datum is immediately public, but only to chain-followers
+    - Datum hashes+extra datums: datum is immediately public, but only to chain-followers
 - Communication of datums
     - Inline datums: easy on-chain communication
     - Datum hashes: off-chain communication necessary
-    - Datum hashes+optional datums: complicated on-chain communication
+    - Datum hashes+extra datums: complicated on-chain communication
 
 Any one of these factors could be important to particular use cases, so it is good to retain the other options.
 
--- a/CIP-0032/README.md
+++ b/CIP-0032/README.md
@@ -1,0 +1,129 @@
+---
+CIP: 32
+Title: Inline datums
+Authors: Michael Peyton Jones <michael.peyton-jones@iohk.io>
+Comments-Summary: No comments
+Comments-URI: 
+Status: Draft
+Type: Standards Track
+Created: 2021-11-29
+License: CC-BY-4.0
+---
+
+# Inline datums
+
+## Abstract
+
+We propose to allow datums themselves to be attached to outputs instead of datum hashes.
+This will allow much simpler communication of datum values between users.
+
+## Motivation
+
+Conceptually, datums are pieces of data that are attached to outputs. 
+However, in practice datums are implemented by attaching _hashes_ of datums to outputs, and requiring that the spending transaction provides the actual datum.
+
+This is quite inconvenient for users. 
+Datums tend to represent the result of computation done by the party who creates the output, and as such there is almost no chance that the spending party will know the datum without communicating with the creating party.
+That means that either the datum must be communicated between parties off-chain, or communicated on-chain by attaching it to the transaction that creates the output (which is also inconvenient in that the spending party must watch the whole chain to spot it).
+
+It would be much more convenient to just put the _datum itself_ in an output, which is what we propose.
+
+### Use cases
+
+We expect that, provided we are able to bring the cost low enough, a large proportion of dapp developers will make use of this feature, as it will simplify their systems substantially.
+
+## Specification
+
+Transaction outputs are changed so that the datum field can contain either a hash or a datum (an "inline datum").
+
+The min UTXO value for an output with an inline datum depends on the size of the datum, following the `coinsPerUTxOWord` protocol parameter.
+
+When an output with an inline datum is spent, the spending transaction does not need to provide the datum itself. 
+
+### Script context
+
+Scripts are passed information about transactions via the script context.
+The script context therefore needs to be augmented to contain information about inline datums.
+
+Changing the script context will require a new Plutus language version in the ledger to support the new interface.
+
+There are two changes in the new version of the interface:
+- The datum field on transaction outputs can either be a hash or the actual datum.
+- The datum field on transaction inputs can either be a hash or the actual datum.
+
+Old versions of the language will retain the old interface, but the construction of the script context in the old interface is also changed as follows: if either an input or an output has an inline datum, we replace it by its hash, and put the datum in the existing datum-hash-to-datum mapping.
+
+### CDDL
+
+The CDDL for transaction outputs will change as follows to reflect the new alternative.
+```
+transaction_output =
+  [ address
+  , amount : value
+  , ? datum : $hash32 / plutus_data
+  ]
+```
+TODO: should there be a dedicated production for datum-hash-or-datum? Does it need to be tagged?
+
+## Rationale
+
+The key idea of this proposal is simply to restore the conceptually straightforward situation where datums are attached to outputs.
+Historically, this was the way that the EUTXO model was designed, and switching to datum hashes on outputs was done to avoid bloating UTXO entries, which at that time (pre-multiasset) were constant-size (see [1] page 7).
+
+Now that we have variable-sized UTXO entries and the accounting to support them, we can restore inline datums.
+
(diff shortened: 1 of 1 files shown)


L-as:
Does it make sense to also allow inline scripts at the same time?

michaelpj:
> Does it make sense to also allow inline scripts at the same time?

Inline scripts are logical, but less pressing, since scripts change less often. Plus they have the added complication that they make the "address" field much more complicated (and, generally, a lot bigger), and many places make assumptions that addresses are roughly hash-sized. So given that there isn't really compelling benefit, I'm not proposing those right now.

CIP-0032/README.md
```diff
@@ -0,0 +1,129 @@
...
+
+The CDDL for transaction outputs will change as follows to reflect the new alternative.
+```
+transaction_output =
+  [ address
+  , amount : value
+  , ? datum : $hash32 / plutus_data
+  ]
+```
+TODO: should there be a dedicated production for datum-hash-or-datum? Does it need to be tagged?
```
SebastienGllmt, line 66: The main use for creating this is that the Rust codebase and a few others are automatically generated from the CDDL definition so if you create a `datum-hash-or-datum` definition, it means this will be a common struct that represents this for all implementations

CIP-0032/README.md
```diff
@@ -0,0 +1,129 @@
...
+
+Any one of these factors could be important to particular use cases, so it is good to retain the other options.
+
+### Changing the script context
+
+We don't strictly need to change the script context.
+We could use the fall-back approach that we use for the old interface, namely continuing to pretend that datums are always hashes, and providing the mapping from hashes to datums.
+This would be functionally equivalent.
+
+The only advantage of changing it is that it's more honest to the real representation of transactions, and therefore allows scripts to insist on inline datums if they want.
```
SebastienGllmt, line 124: For data-availability uses cases being able to enforce the usage of inline datums is crucial so I would definitely make some way to enforce it (be it through its presence in the script context or some other way)

CIP-0032/README.md
```diff
@@ -0,0 +1,129 @@
...
+
+### UTXO set size
+
+This proposal gives users a way to put much larger amounts of data into the UTXO set. 
+Won’t this lead to much worse UTXO set bloat?
+
+The answer is that we already have a mechanism to discourage this, namely the minimum UTXO value. 
+If inline datums turns out to drive significantly increased space usage, then we may need to increase `coinsPerUTxOWord` in order to keep the UTXO size down.
+That will be costly and inconvenient for users, but will still allow them to use inline datums where they are most useful and the cost is bearable.
+Furthermore, we hope that we will in fact be able to _reduce_ `coinsPerUTxOWord` when the upcoming work on moving the UTXO mostly to on-disk storage is complete.
```
SebastienGllmt, line 85: Agreed. To add to this, probably we should have a follow-up CIP to this that changes the tx fee function so that inline datums contribute less towards tx fee than other fields (since unlike other fields, it's just data that don't require extra computation to simply inline into the tx)

WhatisRT: They do contribute to block size though, which is also a very limited resource. Maybe the correct way of dealing with this is to split up `txfee` into a memory and a time component, and have two constants that can be used to adjust them independently.

michaelpj: There are two things here:
- UTXO size. This is mostly a memory constraint, and as such there is no reason to treat inline datums differently, we just need a single parameter based on size.
- Transaction size fee. This is really about processing time, and there there arguably is a reason to pay more or less for different parts of the transaction. But that would be a fairly major change.

CIP-0032/README.md
WhatisRT: This only makes a difference off-chain. IICR validators can already enforce datums for new outputs to be present in a transaction, so the only real difference between inline or not is whether to look it up in the transaction history or the UTXO. Making the inlining status visible to validators would simplify this lookup, at the cost of adding complexity to the on-chain code for everyone.

michaelpj: > is whether to look it up in the transaction history or the UTXO

Which is a big deal! So I think Sebastian's point is relevant: people will want to be able to enforce this.

theganant: @WhatisRT Can you explain how to find datums trustlessly offchain? I can't seem to find any answers online, this issue comes closest

WhatisRT: Since the hash is published on-chain, there's no requirement for trustlessness here

michaelpj:
I think this CIP is ready to be merged as Draft. I will revisit it and update to Active once the implementation has progressed and the CDDL is nailed down.

WhatisRT:
During a meeting today, we've discussed how inline datums should be provided to the transaction context. There are multiple options (including the one currently mentioned in this CIP) with different kinds of advantages and drawbacks. First, here are some potential requirements:
- Low ledger overhead
- Low overhead for scripts
- The ability to distinguish whether a datum is inlined
- Backwards compatibility

Having a representation that is as close to the actual ledger representation as possible reduces the probability of getting bitten by something unexpected down the line. Since scripts _should_ access datums via an API that would allow the same kind of access for datums whether they are inlined or not, the overhead should always be pretty low. Of course there would be some runtime costs that would be reduced if the ledger provided uniform access, but that's probably a small issue. In the case we did do a uniform access approach in the ledger, we could provide a boolean that tells us whether a datum is inlined or not.

However, the big issue seems to be backwards compatibility, i.e. should inline datums be provided to scripts of previous (i.e. current) versions. If not, a user could send funds to a script address with an inline datum and that datum would still be passed to the script that's spending it, but not to other scripts, which breaks the assumption that scripts can always see the datum of other scripts that spend outputs, so we can't do that. This forces us to either provide inline datums to old scripts, or to make them incompatible by force, for example by disallowing inline datums with previous Plutus versions.

Disallowing inline datums and scripts of previous Plutus versions in the same transaction has an other issue though, which is that there'd be a subtle way of making an output unspendable, which is to send it to such a script but with an inline datum. If inline datums become popular, having them might become the default which sounds dangerous. We have no way of distinguishing what language version a script address belongs to (in general, i.e. without knowing the script), so we can't prevent users from making this mistake in the ledger. Wallets could potentially do that, if they know the script.

Providing inline datums to previous scripts is also at odds with having low overhead for the ledger, since they would have to be provided in the same way as datums currently are.

All of this suggests the following options:
1. Staying as close to the actual representation in the ledger as possible, i.e. providing inline datums as part of the outputs (currently in this CIP), and disable inline datums with older scripts. Potentially dangerous for users.
2. Same as above, but with inserting inline datums into old script contexts. This has the worst kind of ledger overhead, since the ledger needs to do different things with inline datums depending on the language version. But it could solve the problem now, while possibly introducing less (or more!) future problems.
3. A uniform approach for all datums, inserting inline datums in the same map where datums live right now, with an additional boolean for newer versions of Plutus to distinguish the type of datum. Not great for the ledger, but should have less runtime overhead, and has backwards compatibility.

Maybe there's something that we've missed when we discussed it, or maybe having more types of unspendable outputs is fine since almost anyone will send money to scripts only though applications that generate the transaction for them, which should really know what kind of input to use. Or maybe it's really really bad for some reason I can't think of right now.

michaelpj:
Thank you for the excellent summary. I'll incorporate this discussion into the CIP text.

For the moment, my inclination is to pick option 1. The reason is:
1. I do think we want scripts to be able to enforce the use of inline datums. If an application developer wants to use inline datums to make things convenient for their users, then it would be quite annoying if someone who interacts with the application can just choose _not_ to do that.
2. Unless there is a pressing reason, I think it's very good to keep the representation of transactions in the script context as close as possible to the representation of transactions in the ledger. This avoids surprises and conceptual overhead for users, which can really hurt us if it builds up.
3. Adding a new way to make outputs unspendable is bad. I think we should avoid these wherever possible. However, I think that in this instance breakage is likely to be rare because: a) it should easily be noticed in testing, b) the client will typically know what kind of script they are using and can avoid this, c) hopefully most users will migrate to Plutus V2 quickly since it will support a number of desirable features.

michaelpj:
I've updated the text.

CIP-0032/README.md
```diff
@@ -115,14 +115,51 @@ However, the other approaches also have some advantages.
...
+While the conceptual distance here may be small, if we let it grow over time then it may become quite confusing.
+
+We then have the choice of what to do about backwards compatibility.
+Option 2 would work, but is more complicated for the ledger and is inconsistent in representation with our choice for the new script context (inline datums are sometimes represented faithfully, and sometimes put in the witness map).
+Option 1 is simple, but doesn't allow old scripts to work with inline datums.
+This would not be so bad if it just meant that old scripts could not be spent in transactions that include inline datums, but it also introduces a new way to make an unspendable output.
+If a user creates an output with an old script and an inline datum, then any transaction spending that output will include an inline datum, which we would not allow.
+
+Unfortunately, we cannot prevent users from creating such outputs in general, since script addresses do not include the script language, so the ledger cannot tell whether the inline datum is permissible or not.
```
L-as, line 157: Even if it did include this information, it would still not be enough. The script might be V2, but to consume it you might need another input that uses a V1 validator, or you might need to mint/burn with a V1 minting policy, which wouldn't work because those scripts can't be run.

CIP-0032/README.md
```diff
@@ -115,14 +115,51 @@ However, the other approaches also have some advantages.
...
+There are a couple of options for how to change the representation of the script context to include the new information, and whether to make a backwards compatibility effort for old language versions.
+
+For the new script context:
+
+1. Match the ledger representation as much as possible: change the fields on inputs and outputs to be either a datum hash or a datum.
+2. Try to only have one way to look up datums: put inline datums in the datum witness map and insert their hashes into the corresponding inputs and outputs; optionally add a boolean to inputs and outputs to indicate whether the datum was originally inline.
+
+For backwards compatibility:
+
+1. Don't try to represent inline datums for scripts using old language versions: old scripts simply can't be run in transactions that use inline datums (since we can't represent the information).
```
L-as, line 144: You probably want to make it explicit that the transaction *fails*, not just that they can't be run.

L-as (approved):
This approach seems fine to me.

CIP-0032/README.md
L-as: This could be because e.g. the datum of the UTXO in question contains a reference to a minting policy hash, and the script asserts that a token of that minting policy is minted. In general, this problem is only solvable by the client knowing the contents of the scripts in question, so I don't think there's any way around this.

michaelpj: Right, that would be a more elaborate way of getting yourself stuck in this position indeed.

CIP-0032/README.md
michaelpj: This is more explicit earlier, but I'll make it even more explicit.

CIP-0032/README.md
```diff
@@ -0,0 +1,167 @@
...
+Another guard rail would be to enforce upper limits on the size of inline datums.
+At the extreme, we could bound them to the size of a hash, which would guarantee no more space usage than today.
+However, this is much worse for users, since it introduces a sharp discontinuity where an inline datum is entirely acceptable, until it crosses the size threshold at which point it is unacceptable, and there is no way to avoid this.
+Generally we prefer to avoid such discontinuities in favour of gradually increasing costs.
+
+In practice, what is implemented here may depend on whether the UTXO-on-disk work is completed by the time that this proposal is implemented.
+
+### Other modes of specifying datums
+
+We could deprecate the other methods of specifying datums (datum hashes or datum hashes+optional datums).
```
koslambrou, line 97: I'm having a hard time understanding what we mean here by `optional datums`. Currently, only datum hash can be inside a tx output. The way this sentence is phrased makes it seem like it's possible to specify the actual datum in a tx output. Unless what `optional datum` means is the datum in the datum witness set?

michaelpj:
It's ready for review.

KtorZ (approved):
Fine to go as draft while the implementation is being worked on :+1:

KtorZ:
(cc @rphair @SebastienGllmt)

rphair (approved):
> Fine to go as draft while the implementation is being worked on

agreed :+1:

PR #163, opened 2021-12-03
CIP-0037? | Dynamic Saturation based on Pledge

CyberCyclone:
This CIP is a proposal to correlate a SPOs total pledge to the pools saturation level. This takes the importance of the pledge and pool saturation away from the pool entity and places it on the SPO entity. This is to patch the loophole that single entity SPOs have exploited in running pool farms with significant amounts of stake. 

The intention of the pool saturation was to limit how dominant a single entity could become, however they have bypassed this intention by creating secondary pools.

This proposal will mean that a SPO will have a saturation point across all their pools based on how much pledge they have across them.

### DIFF
--- a/CIP-0037/README.md
+++ b/CIP-0037/README.md
@@ -1,5 +1,5 @@
 ---
-CIP: CIP-0037 
+CIP: 37 
 Title: Dynamic Saturation 
 Authors: Casey Gibson <caseygibson@protonmail.ch>
 Status: Proposed
--- a/CIP-0037/README.md
+++ b/CIP-0037/README.md
@@ -1,10 +1,11 @@
 ---
-CIP: TODO 
+CIP: CIP-0037 
 Title: Dynamic Saturation 
 Authors: Casey Gibson <caseygibson@protonmail.ch>
-Status: Draft
+Status: In Review
 Type: Standards Track
 Created: 2021-12-03
+Updated: 2022-07-19
 License: CC-BY-4.0  
 ---
 
--- a/CIP-0037/README.md
+++ b/CIP-0037/README.md
@@ -2,7 +2,7 @@
 CIP: CIP-0037 
 Title: Dynamic Saturation 
 Authors: Casey Gibson <caseygibson@protonmail.ch>
-Status: In Review
+Status: Proposed
 Type: Standards Track
 Created: 2021-12-03
 Updated: 2022-07-19
--- a/CIP-0037/README.md
+++ b/CIP-0037/README.md
@@ -83,6 +83,70 @@ One of the questions raised by the community was, will the lower limit stop the 
 1. The value is a percentage of k, such as 10%. This percentage could increase as needed, such as to 15% of k.
 2. The value could be calculated based on the average of active stake compared to active pools. E.g, active stake = 23837 M / 3000 = saturation point of 7.94 M ADA
 
+## Proposals Based On Feedback
+
+After some discussions among the community and some help from https://github.com/cffls, the below code example has been proposed for how the dynamic saturation could work.
+
+```
+let lovelace = 1000000;
+
+function calc_sat(pledge){
+    k = 500;
+    e = 0.2;
+    l = 125;
+    total_supply = 33719282563 * lovelace;
+    orig_sat = total_supply / k;
+    new_sat = orig_sat * Math.max(e, min(1 / k, pledge / orig_sat * l));
+    final_sat = max(new_sat, orig_sat);
+    console.log(`pledge: ${pledge / lovelace}, sat: ${final_sat / lovelace}`);
+}
+
+function max(val1, val2){
+    if(val1 < val2){
+        return val1;
+    }
+    return val2;
+}
+
+function min(val1, val2){
+    if(val1 > val2){
+        return val1;
+    }
+    return val2;
+}
+
+calc_sat(50000 * lovelace);
+calc_sat(100000 * lovelace);
+calc_sat(150000 * lovelace);
+calc_sat(250000 * lovelace);
+calc_sat(500000 * lovelace);
+calc_sat(750000 * lovelace);
+calc_sat(1000000 * lovelace);
+calc_sat(2000000 * lovelace);
+```
+
+Results:
+
+```
+[Log] pledge: 50000, sat: 13487713.0252
+[Log] pledge: 100000, sat: 13487713.0252
+[Log] pledge: 150000, sat: 18750000
+[Log] pledge: 250000, sat: 31250000
+[Log] pledge: 500000, sat: 62500000
+[Log] pledge: 750000, sat: 67438565.126
+[Log] pledge: 1000000, sat: 67438565.126
+[Log] pledge: 2000000, sat: 67438565.126
+```
+
+## Path To Active
+
+CIP-0037 has had active discussions for over 6 months and is now waiting for a review from IOG to provide feedback on the feasibility and soundness of the approach.
+
+1. This CIP should be considered a medium priority as it directly impacts the health and growth of the Cardano ecosystem. Large Cardano pools have had several years to take advantage of the lack of oversight and have control of a large portion of mining operations. This continued lack of restrictions will further damage the trust and reliability of the framework.
+
+2. Timelines around the implementation of the CIP will depend on urgency, however its implementation should be trivial as there are no new parameters required or risks involved. Further this CIP from a high-level aspect only requires an update to the existing algorithm. From an external context, the CIP will require trivial updates as it should be self-contained in cardano-node.
+
+
 ## Copyright
 
 This CIP is licensed under [CC-BY-4.0](https://creativecommons.org/licenses/by/4.0/legalcode)
--- a/cip-caseygibson-dynamicsaturation.md
+++ b/cip-caseygibson-dynamicsaturation.md
@@ -1,0 +1,88 @@
+---
+CIP: TODO 
+Title: Dynamic Saturation 
+Authors: Casey Gibson <caseygibson@protonmail.ch>
+Status: Draft
+Type: Standards Track
+Created: 2021-12-03
+License: CC-BY-4.0  
+---
+
+## Simple Summary
+
+Currently Cardano has been plagued with an ever increasing amount of single entity Stake Pool Operators (SPO) creating multiple pools. The pools that are known to be operated by single entity SPOs account for just 18.72% of the total stake and 50% of the total stake can be attributed to at least 23 single entities (as of 3rd Dec 2021).
+
+The vision of Cardano is that it everyone should be able to have the opportunity to be able to run a stake pool regardless of their financial capabilities and this is even more important for developing countries.
+
+## Abstract
+
+The issue we're currently facing is that many SPOs have been able to exploit a loophole that has allowed them to use their influence to create many sub-pools without any restrictions. As the size of their pools grow, instead of honouring the saturation limits already in place, they are bypassing them by creating more sub-pools. From a technical point, it is theoretically possible for a single entity to run enough stake pools to control more then 50% of the total stake. 
+
+While unlikely, the system does not take into account external influences (eg. political) that could harm the system. Shelly has been active for less then 2 years and we are already seeing the risks on a small scale. As real world adoption continues, there could be situations in the next decade that could jeopardise the decentralisation of Cardano. One example could be that a political party could run a ISPO (Initial Stake Pool Offering) on a mass scale. Seeing as Cardano has the potential to be used for voting (even at a political level), the integrity of Cardano could be questioned if political parties controlled stake pools with large shares of total stake. 
+
+## Motivation
+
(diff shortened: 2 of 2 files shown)


cip-caseygibson-dynamicsaturation.md
```diff
@@ -0,0 +1,88 @@
...
+
+To make the pledge a meaningful metric that is fair to all SPOs and aligns with the core values of Cardano, the pledge should be used to calculate the saturation point of a pool. This will mean that SPOs, no matter how many pools they operate, will have a maximum saturation based on their total pledge. For example, if a pool operates a single pool and wanted to open up another pool with the same amount of stake, they will have to assign the equal amount of pledge against that new pool. If a pool wishes to up their saturation point, they will need to assign a higher amount of pledge.
+
+## Specification
+
+For this to be able to work, there firstly needs to be an upper limit and a lower limit. The K parameter can still be used as the upper saturation limit of a single pool. As in, if a SPO has enough pledge assigned to a single pool, that pool will be able to run at the maximum saturation point of K. The lower limit is in place to safe guard small SPOs and allow them to grow.
+
+An example of how Dynamic Saturation would be calculated:
+
+500,000 ADA Pledged = Saturation point at 100% of K
```
cffls, line 37: We will need a mechanism to determine these pledge thresholds. For example, why does 500,000 ADA pledge, instead of 600,000 ADA, deserve 100% saturation point.

mark-stopka: It's not smart in general to limit saturation, dynamic impact of pledge, relative to total delegated stake would be much more beneficial, and provide a level playing field...

CyberCyclone: > We will need a mechanism to determine these pledge thresholds. For example, why does 500,000 ADA pledge, instead of 600,000 ADA, deserve 100% saturation point.

The example I gave was based on the fact that the values would be adjustable in the same way all the other params were chosen and could be changed based on network consensus.

CyberCyclone: > It's not smart in general to limit saturation, dynamic impact of pledge, relative to total delegated stake would be much more beneficial, and provide a level playing field...

For the purpose of the discussion, can you please explain why "it's not good to limit saturation"? Technically the k param already limits saturation, so I'm not sure what your point is.

The idea is to shift the saturation to the pledge instead of on the pool.

cffls: Here is my idea of formulating the thresholds as a function of `k` and two additional parameters, `e` (entry threshold) and `l` (leverage):

```
orig_sat = total_supply / k
new_sat = orig_sat * max(e, min(1 / k, pledge / orig_sat * l))
```

We can use function `new_sat` to calculate the saturation point for a pool given its pledge, so we don't need to define these thresholds manually in the protocol.The parameter `e` is to, like Casey mentioned, safe guard small SPOs so they can at least have a reasonable chance to generate a block in a certain period of time. The parameter `l` (leverage) could be a value the community define, e.g. 20, which means the saturation point will be 20x the amount of pledge. 

@mark-stopka Not limiting the saturation point means we will discard `k` completely, which will be a much bigger change than the scope of this proposal I think.

CyberCyclone: Thanks for the formula @cffls. Based on my initial analysis it looks pretty promising.

I wrote up your formula as Javascript just to make sure we're clear on your thinking and to see what it would look like.

I played around with `e` and `l` as I thought they were a bit aggressive. In the CIP proposal I picked 500,000 pledge as it seems like a reasonable amount based on what I've seen from pool data. So I've tried to compress it to the lower end so that max saturation is reasonably achievable.

It would be good to see arguments for shifting the range either way.

```
lovelace = 1000000;

function calc_sat(pledge){
    k = 500;
    e = 0.2;
    l = 125;
    total_supply = 33719282563 * lovelace;
    orig_sat = total_supply / k;
    new_sat = orig_sat * Math.max(e, min(1 / k, pledge / orig_sat * l));
    final_sat = max(new_sat, orig_sat);
    console.log(`pledge: ${pledge / lovelace}, sat: ${final_sat / lovelace}`);
}

function max(val1, val2){
    if(val1 < val2){
        return val1;
    }
    return val2;
}

function min(val1, val2){
    if(val1 > val2){
        return val1;
    }
    return val2;
}

calc_sat(50000 * lovelace);
calc_sat(100000 * lovelace);
calc_sat(150000 * lovelace);
calc_sat(250000 * lovelace);
calc_sat(500000 * lovelace);
calc_sat(750000 * lovelace);
calc_sat(1000000 * lovelace);
calc_sat(2000000 * lovelace);

```

results:
```
[Log] pledge: 50000, sat: 13487713.0252
[Log] pledge: 100000, sat: 13487713.0252
[Log] pledge: 150000, sat: 18750000
[Log] pledge: 250000, sat: 31250000
[Log] pledge: 500000, sat: 62500000
[Log] pledge: 750000, sat: 67438565.126
[Log] pledge: 1000000, sat: 67438565.126
[Log] pledge: 2000000, sat: 67438565.126
```

Overall it does do basically what I had in mind.

cffls: Thanks Casey @CyberCyclone  for testing my idea and sharing the result here. I didn't think much about the values 
 for `e` and `l`. Yours (`e=0.2` and `l=125`) seems more reasonable than mine. Another good thing about formulating the saturation point is that we can tune parameters (`e` and `l`) based on the price of ADA and `k`, and eventually have community to vote and decide their values.

StakeWithPride:
The example uses these numbers: "500,000 ADA Pledged = Saturation point at 100% of K". 

This is way too low. I'm pretty confident most of the OG pool group operators have enough ADA to cover this and not collapse pools.

CyberCyclone:
> The example uses these numbers: "500,000 ADA Pledged = Saturation point at 100% of K".
> 
> This is way too low. I'm pretty confident most of the OG pool group operators have enough ADA to cover this and not collapse pools.

I'm not sure who you're referring to @StakeWithPride or what OG means, but the idea is to make single entities the majority operators of the network and force large stake pool operators to put down a decent pledge. If a small number of entities have enough ADA to meet the pledge and want to operate multiple pools then that's not an exploitation.

SmaugPool:
Just a comment to avoid the idea being lost: Tobias Fancee mentioned a possible variant by calculating the pledge benefit from the stake/pledge ratio:

"Pledge/pool stake ratio pledge benefit pretty much does the same thing and is way easier to implement. Also, you wouldn't need saturation at all because the more leverage a pool has the smaller their pledge benefit which gives smaller pools a chance to compete."

https://twitter.com/TobiasFancee/status/1466437270651686922

leo42:
> This CIP is a proposal to correlate a SPOs total pledge to the pools saturation level. This takes the importance of the pledge and pool saturation away from the pool entity and places it on the SPO entity. This is to patch the loophole that single entity SPOs have exploited in running pool farms with significant amounts of stake.
> 
> The intention of the pool saturation was to limit how dominant a single entity could become, however they have bypassed this intention by creating secondary pools.
> 
> This proposal will mean that a SPO will have a saturation point across all their pools based on how much pledge they have across them.

Why define it as a concreate number and not a Leverage cap ? 

For example at a leverage cap of 10 you can have 10X the stake given your pledge.

this way it is also easy to translate into K changes etc.

leo42:
A good way to calculate a starting value for the max leverage(dynamic saturation is to find the total leverage of the network at present and use that) this way theoretically all stake should go to the pools to saturate base on pledge.

Operators will naturally increase pledge as much as they can making the system more fluid and competitive again.

Based on data from (https://adapools.org/press) this suggested starting values would be: 
12.58 


This would mean that for a full saturation of 64M an operator would need ~ 5M Pledge.

This is the extreme end of values that we can support but we can support it and I would make things like new pledge changes and K increases less relevant and necessary

cip-caseygibson-dynamicsaturation.md
santonode: as a small pool, reaching saturation at 13m stake vs 64m seems more reasonable to me and targeting 50k pledge also seems like a realistic target this would encourage me to increase my pledge to avoid declining rewards for my delegates - do I understand this correctly?

cffls: These numbers are just used for discussion purpose and nothing is finalized. In this context, increasing the pledge means increasing the saturation threshold for your pool. If your pool's pledge is at 50k, then the pool will be saturated at 13.49M ADA, meaning that any more delegation to the pool will result in decreasing reward. In this case, you are correct in a sense that you would need to increase the pledge above 50k in order to receive more delegations beyond 13.49M ADA without hurting the ROI of your pool.

cffls: @CyberCyclone Casey, if this formula looks good to you, would you mind including it in your proposal? I can also create a pull request to your clone, and then you can merge it from your clone and resubmit this pull request. I am fine with either way.

cffls:
> 



> > This CIP is a proposal to correlate a SPOs total pledge to the pools saturation level. This takes the importance of the pledge and pool saturation away from the pool entity and places it on the SPO entity. This is to patch the loophole that single entity SPOs have exploited in running pool farms with significant amounts of stake.
> > The intention of the pool saturation was to limit how dominant a single entity could become, however they have bypassed this intention by creating secondary pools.
> > This proposal will mean that a SPO will have a saturation point across all their pools based on how much pledge they have across them.
> 
> Why define it as a concreate number and not a Leverage cap ?
> 
> For example at a leverage cap of 10 you can have 10X the stake given your pledge.
> 
> this way it is also easy to translate into K changes etc.

I proposed a formula which introduced in a new parameter `l` (leverage) for exactly the same reason. Please take a look at the review above.

cffls:
> 



> A good way to calculate a starting value for the max leverage(dynamic saturation is to find the total leverage of the network at present and use that) this way theoretically all stake should go to the pools to saturate base on pledge.
> 
> Operators will naturally increase pledge as much as they can making the system more fluid and competitive again.
> 
> Based on data from (https://adapools.org/press) this suggested starting values would be: 12.58
> 
> This would mean that for a full saturation of 64M an operator would need ~ 5M Pledge.
> 
> This is the extreme end of values that we can support but we can support it and I would make things like new pledge changes and K increases less relevant and necessary

Top 19 pools are IOG and anonymous pools with full pledge (64M). We should exclude these pools since they are outliers. 

If those 19 pools are excluded, we will end up with leverage of 31.09, which means full saturation of 64M will require ~2M pledge, which sounds a bit better.

Calculation:

```
total_delegation = 23930M
total_pledge = total_delegation / 12.294 = 1946.48M
total_pledge_curated = total_pledge - 64M*19 = 730.48M
leverage_curated = (total_delegation - 64M*19) / total_pledge_curated = (23930 - 64*19) / 730.48 = 31.09
```

cip-caseygibson-dynamicsaturation.md
CyberCyclone: Thanks for the reminder @cffls Jerry. I'll try and look into adding it in the next few days. I'm working on another CIP so once that's ready I'll come back to this one.

KtorZ:
Moving forward, we would like to:

- Mark and merge this CIP as proposed
- Add a "Path to active" section which clearly indicates that the next step is at the moment upon IOG to provide feedback on the feasibility and soundness of the approach; as well as (if applicable) possible timelines or priorities when it comes to implement the proposition.

CyberCyclone:
> Moving forward, we would like to:
> 
> * Mark and merge this CIP as proposed
> * Add a "Path to active" section which clearly indicates that the next step is at the moment upon IOG to provide feedback on the feasibility and soundness of the approach; as well as (if applicable) possible timelines or priorities when it comes to implement the proposition.

@KtorZ The CIP has been updated for IOG review. Thank you!

cip-caseygibson-dynamicsaturation.md
CyberCyclone: @cffls I've finally gotten around to updating the CIP. I've attached the code to the bottom of the CIP.

CIP-0037/README.md
```diff
@@ -0,0 +1,153 @@
+---
+CIP: CIP-0037 
+Title: Dynamic Saturation 
+Authors: Casey Gibson <caseygibson@protonmail.ch>
+Status: In Review
```
rphair, line 5:
```suggestion
Status: Proposed
```
... `Status:` must be one of these keywords 🤓 https://github.com/cardano-foundation/CIPs/tree/master/CIP-0001#cip-statuses
... and as per https://github.com/cardano-foundation/CIPs/pull/163#issuecomment-1136204287 we will merge this as `Status: Proposed` while awaiting the review suggested in your _Path To Active_.

rphair (commented):
@CyberCyclone thanks for adding _Path To Active_ although the `Status:` in the header does not have to reflect the fact that you/we are waiting for IOG to review...

CIP-0037/README.md
```diff
@@ -0,0 +1,153 @@
+---
+CIP: CIP-0037
```
rphair, line 2:
```suggestion
CIP: 37 
```

rphair (commented):
thanks for making that last change but my apologies (for duplication of effort) we also missed this one 😖

CyberCyclone:
No problem @rphair. Let me know if there's anything else needed.

rphair (approved):

CyberCyclone:
@rphair just pinging this in case it's been forgotten about 👍

rphair:
@CyberCyclone I've requested reviews from the other editors already, plus I've also just now requested that we add this to the upcoming CIP meeting agenda... so it's sure to progress somehow from there 😎 here's the Discord where that meeting will take place: https://discord.com/channels/971785110770831360

KtorZ (approved):

PR #177, opened 2021-12-17
CIP-0005 | Add prefixes for output datum hash and script data hash

gabrielKerekes:
We're suggesting the additions mainly for HW wallets.

For "Output datum hash" we also considered using `output_data`.
For "Script data hash" we also considered using `script_data` or `redeemers` (the plural rather than singular).

The `datum` and `redeemer` prefixes were actually proposed to us by @KtorZ, with the reasoning that this naming was used from the start in regard to the eUTXO model.

### DIFF
--- a/CIP-0005/README.md
+++ b/CIP-0005/README.md
@@ -70,19 +70,20 @@ We define the following set of common prefixes with their corresponding semantic
 
 #### Hashes
 
-| Prefix             | Meaning                                            | Contents                                               |
-| ---                | ---                                                | ---                                                    |
-| `asset`            | Fingerprint of a native asset for human comparison | See [CIP-0014]                                         |
-| `pool`             | Pool operator verification key hash (pool ID)      | blake2b\_224 digest of an operator verification key    |
-| `script`           | Script hash                                        | blake2b\_224 digest of a serialized transaction script |
-| `addr_vkh`         | Address verification key hash                      | blake2b\_224 digest of a payment verification key      |
-| `addr_shared_vkh`  | Shared address verification key hash               | blake2b\_224 digest of a payment verification key      |
-| `policy_vkh`       | Policy verification key hash                       | blake2b\_224 digest of a policy verification key       |
-| `stake_vkh`        | Stake address verification key hash                | blake2b\_224 digest of a delegation verification key   |
-| `stake_shared_vkh` | Shared stake address verification key hash         | blake2b\_224 digest of a delegation verification key   |
-| `vrf_vkh`          | VRF verification key hash                          | blake2b\_256 digest of a VRF verification key          |
-| `datum`            | Output datum hash                                  | blake2b\_256 digest of output datum                    |
-| `script_data`      | Script data hash                                   | blake2b\_256 digest of script data                     |
+| Prefix             | Meaning                                            | Contents                                                 |
+| ---                | ---                                                | ---                                                      |
+| `asset`            | Fingerprint of a native asset for human comparison | See [CIP-0014]                                           |
+| `pool`             | Pool operator verification key hash (pool ID)      | blake2b\_224 digest of an operator verification key       |
+| `script`           | Script hash                                        | blake2b\_224 digest of a serialized transaction script    |
+| `addr_vkh`         | Address verification key hash                      | blake2b\_224 digest of a payment verification key         |
+| `addr_shared_vkh`  | Shared address verification key hash               | blake2b\_224 digest of a payment verification key         |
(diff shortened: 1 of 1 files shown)


CIP-0005/README.md
```diff
@@ -78,6 +78,8 @@ We define the following set of common prefixes with their corresponding semantic
 | `stake_vkh`        | Stake address verification key hash                | blake2b\_224 digest of a delegation verification key   |
 | `stake_shared_vkh` | Shared stake address verification key hash         | blake2b\_224 digest of a delegation verification key   |
 | `vrf_vkh`          | VRF verification key hash                          | blake2b\_256 digest of a VRF verification key          |
+| `datum`            | Output datum hash                                  | blake2b\_256 digest of output datum                    |
+| `redeemer`         | Script data hash                                   | blake2b\_256 digest of script data                     |
```
michaelpj, line 82: I don't know what this means. Redeemers aren't hashed, I don't think. Do you mean the hash of an actual script? Then that is definitely not a redeemer.

KtorZ: I believe this refers to the script integrity hash, doesn't it? In which case (a) `redeemer` is definitely not a good name, and (b) I am not sure we need / want a bech32 prefix for that one. It's very "internal" and not quite relevant to end users.

gabrielKerekes: It refers to the `script_data_hash` field in the `tx_body`(see [CDDL](https://github.com/input-output-hk/cardano-ledger/blob/master/eras/alonzo/test-suite/cddl-files/alonzo.cddl#L61)). As it's included in the tx body, it gets processed by HW wallets and should be shown to the user.

Perhaps `script_data` would be a better prefix?

gabrielKerekes: I've renamed `redeemer` to `script_data` -> https://github.com/cardano-foundation/CIPs/pull/177/commits/769b7c6595fdad5bd5561bef30cde231af6587f0

CIP-0005/README.md
```diff
@@ -68,18 +68,19 @@ We define the following set of common prefixes with their corresponding semantic
...
+| `asset`            | Fingerprint of a native asset for human comparison | See [CIP-0014]                                            |
+| `pool`             | Pool operator verification key hash (pool ID)      | blake2b\_224 digest of an operator verification key       |
+| `script`           | Script hash                                        | blake2b\_224 digest of a serialized transaction script    |
+| `addr_vkh`         | Address verification key hash                      | blake2b\_224 digest of a payment verification key         |
+| `addr_shared_vkh`  | Shared address verification key hash               | blake2b\_224 digest of a payment verification key         |
+| `stake_vkh`        | Stake address verification key hash                | blake2b\_224 digest of a delegation verification key      |
+| `stake_shared_vkh` | Shared stake address verification key hash         | blake2b\_224 digest of a delegation verification key      |
+| `signer_vkh`       | Required signer verification key hash              | blake2b\_224 digest of a required signer verification key |
```
gabrielKerekes, line 80: After discussions on IOHK Slack we'd also like to add the `signer_vkh` prefix which is to be used to display the [`required_signers`](https://github.com/input-output-hk/cardano-ledger/blob/master/eras/alonzo/test-suite/cddl-files/alonzo.cddl) field of `tx_body`.

The existing prefixes are not suitable, because a HW wallet cannot determine whether the key is a payment (1852`) key, shared key (1854`) or a stake key so it wouldn't know which prefix to use. Another alternative would be to display the key hashes as hex, but that's also not ideal.

(the diff is showing also changes in the other prefixes but only the alignment of the `Contents` column has changed in those lines - GitHub UI doesn't show it too well)

michaelpj: Do you not want to include "required" or "req" somewhere in that prefix? We could have other cases where we need a vhk from a signer, this is a bit ambiguous.

gabrielKerekes: Makes sense, thank you. Changed in https://github.com/cardano-foundation/CIPs/pull/177/commits/f991d4dee03f322225e950592fd9cbbf45bb15cf.

rvl (approved):
No objections from me.

KtorZ (approved):

crptmppt (approved):

KtorZ:
@gabrielKerekes We'd like to merge this PR but it has conflicts and need to be rebased on top of current master. I would do it myself if I could but I can't push on your branch it seems.

If you can either grant rights or, do it yourself :pray:

gabrielKerekes:
@KtorZ I've rebased the branch on top of current master 👍

KtorZ:
Thanks!

Issue #178, opened 2021-12-17
CIP-0030 GetBalance and Account/Network change


josemmrocha:
Hello,

After integrating ccvault wallet (making use of CIP-0030), I have two suggestions.

1) getBalance(). It returns now the full balance of the account. Could it be possible to add another endpoint that returns the actual available balance? (total balance - ada attached to tokens). I am having the issue where I retrieve the balance to see if a user can pay for a CNFT, and total balance is higher than CNFT price, but their actual available balance is not. So when submitting the tx, it fails and user dont know why (when it is only that they dont have enough 'free' balance). It will always happen this. And problem will be bigger the more assets an account has.

2) Account/Network change: Is it possible to have events when account/network is changed? Just like Nami wallet has now. That way is easier to subscribe to this event, and act accordingly (calling .enable() again to resfresh the API for example). Without this events it is more difficult to manage when user changes to testnet, or changes account, having the website with dApp connector 'opened' and not knowing this changes took place.

Thanks

rooooooooob:
@josemmrocha 

>Could it be possible to add another endpoint that returns the actual available balance? (total balance - ada attached to tokens).

We talked about this in [the original CIP-30 PR](https://github.com/cardano-foundation/CIPs/pull/88#discussion_r651558195) as well as #171  about how we might want to have more granularity for the balance/utxos endpoints but no consensus has been reached as of yet.

>(total balance - ada attached to tokens)

What do you mean attached? Perhaps I'm misunderstanding you but can't you use ADA that is together with a non-ADA multiasset? You'll just want to return said asset(s) in a change output to the user to only use the ADA. Although if it's juts a single one then you'll still have the min ADA per utxo calculation coming into play so if that UTXO only had the minimum attached to it potentially wouldn't help if you weren't going to have a change output for some reason. Or if you're using multiple and bunching them in a single change output that ADA is still usable. Of course the user may or may not want that to happen but I'm not sure what the ideal general API for that would be about controlling what the user wants done with his or her UTXOs/bundling and that seems like it may be more on the wallet's side than the CIP-30 spec side but we haven't had this come up in discussion so far so maybe there is a better solution.

>Account/Network change: Is it possible to have events when account/network is changed? 

See #151 these were removed from CIP-30's first PR in order to get it merged as we had not reached a consensus yet by that time in the original PR and #151 was opened up to continue work there.

We should move this discussion over to #151 and #171 respectively instead of this issue.

josemmrocha:
[...]

PR #181, opened 2021-12-23
CIP30: getUtxos null instead of undefined

SebastienGllmt:
Currently CIP30 defines the `getUtxos` endpoint should return `undefined` if the requested amount can't be satisfied

The problem is that browser message passing uses JSON and `undefined` is not valid JSON. This can easily introduce subtle bugs (some libraries silently convert `undefined` to `null`) and it's needless complexity when we could just define the spec to return null instead.

Strictly speaking this is a breaking change, but I'm not sure if it will affect app at the moment.

Another option would be to return an error instead of null which I think also makes sense

### DIFF
--- a/CIP-0030/README.md
+++ b/CIP-0030/README.md
@@ -191,11 +191,11 @@ Errors: `APIError`
 
 Returns the network id of the currently connected account. 0 is testnet and 1 is mainnet but other networks can possibly be returned by wallets. Those other network ID values are not governed by this document. This result will stay the same unless the connected account has changed.
 
-### api.getUtxos(amount: cbor\<value> = undefined, paginate: Paginate = undefined): Promise\<TransactionUnspentOutput[] | undefined>
+### api.getUtxos(amount: cbor\<value> = undefined, paginate: Paginate = undefined): Promise\<TransactionUnspentOutput[] | null>
 
 Errors: `APIError`, `PaginateError`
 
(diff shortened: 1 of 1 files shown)


alessandrokonrad:
I'm not even sure if `undefined` or `null` make even sense here, because the return type is `TransactionUnspentOutput[]`. If no utxos were found by filtering by certain assets, it should return an empty array imo.

SebastienGllmt:
It's not filtering by an asset -- it's filtering by an amount

Honestly though I think taking in an argument list for everything in CIP30 was a mistake. We should have made function only have a single parameter -- a JS object. It would have make backward/forward compatibility much easier -- especially for functions like this where people may want to add different filters, etc. in the future but can't anymore because of the optional parameter

alessandrokonrad:
> It's not filtering by an asset -- it's filtering by an amount

Yeah, but as you say it's filtering, so normally when I apply a filter on an array I always get back an array again regardless of the result.

SebastienGllmt:
Returning an empty array would only make sense if this was element-wise filtering though. Otherwise you wouldn't be able to differentiate between a wallet with no UTXOs and a wallet with UTXOs but just not enough to sum to the value you requested

mkazlauskas:
> Returning an empty array would only make sense if this was element-wise filtering though. Otherwise you wouldn't be able to differentiate between a wallet with no UTXOs and a wallet with UTXOs but just not enough to sum to the value you requested

Agree. I'm wondering what is the intended use case for this filtering. If it's for input selection then this interface is not sufficient and should be redesigned in my opinion (see [this comment](https://github.com/cardano-foundation/CIPs/issues/178#issuecomment-998641446)).

vsubhuman (approved):

KtorZ (approved):

rphair (approved):

PR #183, opened 2021-12-28
CIP30: experimental api section

SebastienGllmt:
# Problem

1. Many wallets have started defining their own functions that aren't part of CIP30. It makes it hard for dApp devs to know which functions are standardized (supported by all wallets) and which are wallet-specific
1. We currently have no clear process for updating the CIP30 api version

Adding an "Experimental API" section to CIP30 aims to solve these issues.

## Experimental API

Functions in this section must be under the `experimental` namespace (ex: `api.experimental.myFunctionality`).

The benefits of this are:
1. Wallets can non-standardized features while still following the CIP30 structure
1. dApp developers can use this functions explicitly knowing they are experimental (not stable or standardized)
1. New features can be added to CIP30 as experimental features and only moved to non-experimental once multiple wallets implement it
1. Provides a clear path to updating the CIP version number (when functions move from experimental -> stable)

### DIFF
--- a/CIP-0030/README.md
+++ b/CIP-0030/README.md
@@ -247,6 +247,16 @@ Errors: `APIError`, `TxSendError`
 
 As wallets should already have this ability, we allow dApps to request that a transaction be sent through it. If the wallet accepts the transaction and tries to send it, it shall return the transaction id for the dApp to track. The wallet is free to return the `TxSendError` with code `Refused` if they do not wish to send it, or `Failure` if there was an error in sending it (e.g. preliminary checks failed on signatures).
 
+## Experimental API
+
+Functions in this section must be under the `experimental` namespace (ex: `api.experimental.myFunctionality`).
+
+The benefits of this are:
+1. Wallets can non-standardized features while still following the CIP30 structure
+1. dApp developers can use this functions explicitly knowing they are experimental (not stable or standardized)
+1. New features can be added to CIP30 as experimental features and only moved to non-experimental once multiple wallets implement it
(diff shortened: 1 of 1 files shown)


alessandrokonrad:
That is a great idea. Then I can finally move Nami to CIP-0030, while still supporting some of the custom endpoints and event functions.

MarcelKlammer:
Just to clarify, that would be?:
```
const fullApi = await window.cardano.walletName.enable() // walletName: eg. nami or ccvault

await fullApi.experimental.functionName()
```

SebastienGllmt:
@MarcelKlammer yes, that's correct

MarcelKlammer:
Today I had a feature request of a nft marketplace for an appVersion property, which can be read ahead of calling enable().

So I put it here:

```
window.cardano.ccvault.experimental.appVersion = { major: 1, minor 3, patch: 9 }
```

So we could also add such an object in the wallet namespace.

SebastienGllmt:
Just an FYI, Nami, ccvault and Flint now all use this `experimental` field so I think probably this should be merged

In general probably we need a modification to CIP30 to define when we can merge these though

KtorZ (approved):

rooooooooob (approved):
LGTM

rphair (approved):
Although I don't understand every detail of this spec, the additions look understandable, useful & well precedented to me.

vsubhuman (approved):

vsubhuman:
Note, Yoroi is updating to utilise this as well

PR #185, opened 2022-01-03
Smart Contract Software Licenses

Quantumplation:
Allows specifying, in transaction metadata, the license for some or all contracts in the transaction

### DIFF
--- a/CIP-0034/README.md
+++ b/CIP-0034/README.md
@@ -1,0 +1,82 @@
+---
+CIP: 34
+Title: Smart Contract Software Licensing
+Authors: Pi Lanningham <pi@sundaeswap.finance>
+Comments-URI:
+Status: Draft
+Type: Informational
+Created: 2022-01-03
+Post-History: 
+License: CC-BY-4.0
+---
+
+## Abstract
+
+This proposal defines a metadata standard for specifying the license for a smart contract.
+
+## Motivation
+
+Smart contracts, specified in source code or compiled to raw bytecode as presented on the blockchain, are creative and often highly valuable works and are often subject to legal licensing concerns.  It is important that companies building on Cardano have the ability to link their smart contracts unambiguously to the license.
+
+This standard allows such a dApp builder to specify in the metadata of a transaction a reference to the license for each script.
+
+Ensuring that there is a consistent way to do this will have a number of benefits:
+ - Explorer projects, such as CardanoScan, can be updated to show the script license before showing the bytecode
+ - Projects have legal recourse against outright, low-effort forks of their projects (such as the Sushiswap / Uniswap fiasco)
+ - In theory, parts of the ecosystem could (at their discretion) choose to honor these licenses, and refuse to interoperate with projects that violate the license; A DEX, for example, could justify excluding a token on the basis of violation of an appropriate software license.
+
+Ultimately, the above encourages and fosters a more Open and collaborative Cardano.  With some confidence that their license-restricted code won't be copied outright, projects can feel more comfortable open-sourcing more of their intellectual property for the sake of knowledge sharing and innovation, with the risk of a whole-sale copy mitigated.
+
+## Specification
+
+This is the proposed `transaction_metadatum_label` value
+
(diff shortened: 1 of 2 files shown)


Quantumplation:
Note: Just noticed there's already a 0034 in PR review.  Since the CIP number is a moving target, let me know which CIP number I should assign it to; otherwise I'll update it to something unused right before merging.

michaelpj (changes requested):
This CIP is missing a Rationale section, which is important since it doesn't explain how the proposed change will satisfy the motivation, and indeed I'm not sure that it does.

In particular, there is nothing to:
- Enforce the presence of the metadata
- Enforce the correctness of the metadata

If I submit a transaction that uses a script I can claim whatever license I want with this system, and nobody can tell whether it's correct or not... unless there is some other way of confirming the license, in which case this is redundant.

I think a solution like https://github.com/cardano-foundation/CIPs/tree/master/CIP-0026 is more appropriate, since it would provide at least a hope of having trustworthy metadata.

CIP-0034/README.md
```diff
@@ -0,0 +1,82 @@
...
+
+This value is chosen as it is unused, and because it is the year that the [Open Source Initiative was founded](https://opensource.org/history).
+
+### Structure
+
+The structure allows for multiple smart contracts, also with different licenses, in a single transaction.  It also allows specifying licenses for only a subset of the contracts.
+
+```
+{
+  "1998": {
```
SebastienGllmt, line 46: Should register the label in CIP10 at the same time

CIP-0034/README.md
```diff
@@ -0,0 +1,82 @@
...
+    "<contract_address>": {
```
SebastienGllmt, line 47: Could we get a CDDL specification instead of a JSON specification? Transaction metadata is a subset of valid JSON so we should avoid JSON definitions

KtorZ:
As discussed today:

- This CIP lacks indeed a rationale section, at least explaining what are the legal implication of such a proposal and what happens in case of conflicting metadata submission. 

- A short CDDL specification of the on-chain format would make it unambiguous 

- Make sure to rename it as CIP-0039 and change the status as `Proposed`. 

- Also, do you know of any other entity beside SundaeSwap who's been adopting this (therefore, possibly moving it as _active_ directly)?

CIP-0034/README.md
```diff
@@ -0,0 +1,82 @@
...
+License: CC-BY-4.0
+---
+
+## Abstract
+
+This proposal defines a metadata standard for specifying the license for a smart contract.
+
+## Motivation
+
+Smart contracts, specified in source code or compiled to raw bytecode as presented on the blockchain, are creative and often highly valuable works and are often subject to legal licensing concerns.  It is important that companies building on Cardano have the ability to link their smart contracts unambiguously to the license.
```
L-as, line 19: > are often subject to legal licensing concerns

Is this in fact the case? You put your smart contract on the blockchain, and now everyone has it. What happens if I take a random script from my local copy, and put it on my website? You could sue me, but I think in many jurisdictions, you'd have a hard time winning the suit, especially when this is what blockchain explorers etc. are already doing.

L-as: Of course, this doesn't apply to the *source form*, but that's not relevant to this CIP, because that's not put on the chain, and if it is, the same concerns as above apply.

micahkendall: I would assume that attaching the license only matters when you are providing/using the source code; for example, when you are submitting transactions and providing the source or using reference inputs. Would it make more sense then to attach the metadata to the Plutus script instead? There could be a standard for wrapping a Plutus script, which would encode a license without adding greatly to execution costs.

rphair:
@Quantumplation without any response to the last editorial enquiries for some time now, we've marked this for closure.  If you're still interested in pursuing this, can you address the suggestions above & update your code baseline?

If continuing, note the CIP number 34 has already been officially adopted: https://github.com/cardano-foundation/CIPs/tree/master/CIP-0034

PR #193, opened 2022-01-14
CIP-0025 | Specified asset name encoding

alessandrokonrad:
Since the encoding was never really specified for asset names and most projects use `UFT-8`, `UTF-8` is now the standard  for version `1.0`. All future versions will use the `hex` encoding for asset names. Then we can also keep backwards compatibility.

### DIFF
--- a/CIP-0025/README.md
+++ b/CIP-0025/README.md
@@ -69,6 +69,8 @@ The structure allows for multiple token mints, also with different policies, in 
 }
 ```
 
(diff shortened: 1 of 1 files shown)


KtorZ (approved):
Approving with mixed feelings since asset names are not necessarily representable as valid UTF-8 byte sequences! This is mildly okay in this context because it only excludes those non representable asset-name from complying with this CIP. 

Version 2 would need to address this issue.

SebastienGllmt (approved):

crptmppt (approved):

PR #200, opened 2022-01-25
CIP-36? | Add support for Catalyst multi-delegation

mark-stopka:
Create repacement to CIP-15 to support the latest requirements for Catalyst and Voltaire, namely the ability to delegate one's rights to vote to (possibly multiple) representatives and/or expert voters.

### DIFF
--- a/CIP-0036/README.md
+++ b/CIP-0036/README.md
@@ -1,0 +1,172 @@
+---
+CIP: 36
+Title: Catalyst/Voltaire Registration Transaction Metadata Format (Updated)
+Authors: Sebastien Guillemot <sebastien@emurgo.io>, Rinor Hoxha <rinor.hoxha@iohk.io>, Mikhail Zabaluev <mikhail.zabaluev@iohk.io>, Giacomo Pasini <giacomo.pasini@iohk.io>, Kevin Hammond <kevin.hammond@iohk.io>
+Comments-URI: https://forum.cardano.org/t/cip-catalyst-registration-metadata-format/44038
+Status: Draft
+Type: Standards
+Created: 2021-12-06
+License: CC-BY-4.0
+---
+
+## Abstract
+
+Cardano uses a sidechain for its treasury system. One needs to "register" to participate on this sidechain by submitting a registration transaction on the mainnet chain. This CIP details the registration transaction format.
+This is a revised version of the original CIP-15.
+
+## Motivation
+
+Cardano uses a sidechain for its treasury system ("Catalyst") and for other voting purposes. One of the desirable properties of this sidechain is that even if its safety is compromised, it doesn't cause loss of funds on the main Cardano chain. To achieve this, instead of using your wallet's recovery phrase on the sidechain, we need to use a brand new "voting key".
+
+However, since 1 ADA = 1 vote, a user needs to associate their mainnet ADA to their new voting key. This can be achieved through a registration transaction.
+
+In addition, to encourage participation by a broader range of ADA holders, it should be possible to delegate one's rights to vote to (possibly multiple) representatives and/or expert voters. Such delegations will still be able to receive Catalyst rewards. 
+
+We therefore need a registration transaction that serves three purposes:
+
+1. Registers a "voting key" to be included in the sidechain and/or delegates to existing "voting key"s
+2. Associates mainnet ADA to this voting key(s)
+3. Declares an address to receive Catalyst rewards
+
+
+Note: This schema does not attempt to differentiate delegations from direct registrations, as the two options have exactly the same format.  It also does not distinguish between delegations that are made as "private" arrangements (proxy votes)
+from those that are made by delegating to representatives who promote themselves publicly.
+Distinguishing these possibilities is left to upper layers or future revisions of this standard, if required.
+In this document, we will use the term 'delegations' to refer to all these possibilities.
+
+## Specification
+
+### Registration metadata format
+
+A Catalyst registration transaction is a regular Cardano transaction with a specific transaction metadata associated with it.
+
+Notably, there should be five entries inside the metadata map:
+ - A non-empty array of delegations, as described below;
+ - A stake address for the network that this transaction is submitted to (to point to the Ada that is being delegated);
+ - A Shelley address discriminated for the same network  this transaction is submitted to to receive rewards.
+ - A nonce that identifies that most recent delegation
+ - A number that indicates the purpose of the vote
+
+### Delegation format
+
+A delegation assigns (a portion of) the ADA controlled by one or more UTxOs on mainnet to the voting key in the sidechain as voting power.  The UTxOs can be identified via the stake address at some designated point in time.
+
+Each delegation therefore contains:
+  - a voting key: simply an ED25519 public key. This is the spending credential in the sidechain that will receive voting power from this delegation. For direct voting it's necessary to have the corresponding private key to cast votes in the sidechain. How this key is created is up to the wallet.
+  - the weight that is associated with this key: this is an unsigned integer (CBOR major type 0) that represents the relative weight of this delegation over the total weight of all delegations in the same registration transaction.
+  The weight may range from 0 to 2^32-1.  Any greater value is capped to 2^32-1.
+
+
+### Associating voting power with a voting key
+
+This method has been used since Fund 2.
+For future fund iterations, a new method making use of time-lock scripts may
+be introduced as described [below][future-development].
+
+Recall: Cardano uses the UTXO model so to completely associate a wallet's balance with a voting key (i.e. including enterprise addresses), we would need to associate every payment key to a voting key individually. Although there are attempts at this (see [CIP-0008]), the resulting data structure is a little excessive for on-chain metadata (which we want to keep small)
+
+Given the above, we choose to only associate staking keys with voting keys. Since most Cardano wallets only use base addresses for Shelley wallet types, in most cases this should perfectly match the user's wallet.
+
+The voting power that is associated with each delegated voting key is derived from the user's total voting power
+as follows.
+
+1. The total weight is calculated as a sum of all the weights;
+2. The user's total voting power is calculated as a whole number of ADA (rounded down);
+3. The voting power associated with each voting key in the delegation array is calculated as the weighted fraction of the
+   total voting power (rounded down);
+4. Any remaining voting power is assigned to the last voting key in the delegation array.
+
+This ensures that the voter's total voting power is never accidentally reduced through poor choices of weights,
+and that all voting powers are exact ADA.
+
+### Example 
+
+Voting registration example:
+```
+61284: {
+  // delegations - CBOR byte array 
+  1: [("0xa6a3c0447aeb9cc54cf6422ba32b294e5e1c3ef6d782f2acff4a70694c4d1663", 1), ("0x00588e8e1d18cba576a4d35758069fe94e53f638b6faf7c07b8abd2bc5c5cdee", 3)],
+  // stake_pub - CBOR byte array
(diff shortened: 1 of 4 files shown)


mark-stopka:
@SebastienGllmt, @KtorZ what changes should be made, so this can me merged on the next CIP review meeting?

KtorZ:
@mark-stopka it won't be merged in the next CIP meeting; but it may go into last-check. The process is lengthy on purpose to allow folks to comment and react on proposals. As for the content itself, I haven't looked at it in depth yet but the overall looks good and well-structured. So I don't foresee any particular issue given that we already discussed the approach in the last meeting.

CIP-0036/README.md
```diff
@@ -0,0 +1,172 @@
...
+## Specification
+
+### Registration metadata format
+
+A Catalyst registration transaction is a regular Cardano transaction with a specific transaction metadata associated with it.
+
+Notably, there should be five entries inside the metadata map:
+ - A non-empty array of delegations, as described below;
+ - A stake address for the network that this transaction is submitted to (to point to the Ada that is being delegated);
+ - A Shelley address discriminated for the same network  this transaction is submitted to to receive rewards.
```
disassembler, line 46: this needs to be a shelley stake address. It can't be a regular address because of minutxo requirements.

disassembler (commented):
Is this really what we want for multi-delegation? This allows you to split your voting rights across multiple keys with different weights, which might be useful, but I always assumed multi-delegation for voting was going to be the ability to designate different experts for different challenges with the full voting power... e.g. I might delegate rights to the SPO I delegate to for an SPO challenge.

CIP-0036/README.md
SebastienGllmt: The problem is that the challenges are not static -- they dynamically change every round. It means you can't "fire and forget" for a specific category and you'd have to explicitly re-delegate your vote every funding round. You could argue whether or not this is a desirable property or not

mark-stopka: I would argue it is a desirable property for now; later after DCF emerges and voting on parameters will be enabled we will have to revisit this CIP anyway; so IMO for now it fits the current crucial need to enable Project Catalyst to grow and get us where we need to be this year.

MichalPetro:
We discussed this today and there are two options how to implement this proposal in hw wallets:
1) Drop the old format and support only the proposed one. We would need to sync the breaking change with sw wallets.
2) Support both formats, even on fw level. This would increase fw code complexity and size (significantly).
We prefer the first option since it is easier to implement and doesn't increase hw wallet app size that much.

CIP-0036/README.md
kevinhammond: Note that the voting purpose field allows different delegation assignments for specific reasons (and the CIP deliberately doesn't restrict what those reasons might be).  So we simply need to define the mappings from voting purpose to challenges etc. and reflect that in the voting weights.

kevinhammond: We can explicitly make the rewards address a Shelley address.  I think that was just carried over from CIP-15

kevinhammond: Mark, we are definitely looking ahead to non-Catalyst uses for this

kevinhammond:
We need to add the following text to this CIP.  I can't do that.

### Voting key derivation path

To avoid linking voting keys directly with Cardano spending keys,  the voting key derivation path must start with a specific segment:

`m / 1694' / 1815' / account' / chain / address_index`

CIP-0036/README.md
```diff
@@ -0,0 +1,172 @@
...
+4. Any remaining voting power is assigned to the last voting key in the delegation array.
+
+This ensures that the voter's total voting power is never accidentally reduced through poor choices of weights,
+and that all voting powers are exact ADA.
+
+### Example 
+
+Voting registration example:
+```
+61284: {
```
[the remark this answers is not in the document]
janmazak: `CIP-0036/schema.cddl` should also be referenced before the example.

CIP-0036/README.md
```diff
@@ -0,0 +1,172 @@
...
+the staking key on the Cardano network, the address to receive rewards,
+a nonce, and a voting purpose, respectively. A registration with these metadata will be
+considered valid if the following conditions hold:
+
+- The nonce is an unsigned integer (of CBOR major type 0) that should be 
+  monotonically rising across all transactions with the same staking key.
+  The advised way to construct a nonce is to use the current slot number.
+  This is a simple way to keep the nonce increasing without having to access
+  the previous transaction data.
+- The reward address is a Shelley address discriminated for the same network
```
janmazak, line 109: As mentioned by @disassembler , it should be a stake address.

CIP-0036/README.md
```diff
@@ -0,0 +1,172 @@
...
+Delegation to the voting key `0xa6a3c0447aeb9cc54cf6422ba32b294e5e1c3ef6d782f2acff4a70694c4d1663` will have relative weight 1 and delegation to the voting key `0x00588e8e1d18cba576a4d35758069fe94e53f638b6faf7c07b8abd2bc5c5cdee` relative weight 3 (for a total weight of 4).
+Such a registration will assign 1/4 and 3/4 of the value in ADA to those keys respectively, with any remainder assigned to the second key.
+
+To produce the signature field, the CBOR representation of a map containing
+a single entry with key 61284 and the registration metadata map in the
+format above is formed, designated here as `sign_data`.
+This data is signed with the staking key as follows: first, the
+blake2b-256 hash of `sign_data` is obtained. This hash is then signed
+using the Ed25519 signature algorithm. The signature metadata entry is
+added to the transaction under key 61285 as a CBOR map with a single entry
```
janmazak, line 124: "added to the transaction" is imprecise, we are still talking about constructing tx metadata, right?

CIP-0036/README.md
```diff
@@ -0,0 +1,172 @@
...
+### Registration metadata format
+
+A Catalyst registration transaction is a regular Cardano transaction with a specific transaction metadata associated with it.
+
+Notably, there should be five entries inside the metadata map:
+ - A non-empty array of delegations, as described below;
+ - A stake address for the network that this transaction is submitted to (to point to the Ada that is being delegated);
+ - A Shelley address discriminated for the same network  this transaction is submitted to to receive rewards.
+ - A nonce that identifies that most recent delegation
+ - A number that indicates the purpose of the vote
```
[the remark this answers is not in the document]
janmazak: According to the CDDL below, the purpose number is optional. Shouldn't it be mentioned here? Or some discussion of allowed/typical/default values of that?

CIP-0036/test-vector.md
```diff
@@ -0,0 +1,112 @@
...
+
+Legacy:
+```javascript
+{
+  '61284': {
+    '1': '0x0036ef3e1f0d3f5989e2d155ea54bdb2a72c4c456ccb959af4c94868f473f5a0',
+    '2': '0x1c5d88aa573da97e5a4667e0f7c4a9c6a3d848934c3b0a5b9296b401540f2aef',
+    '3': '0xe0ae3a0a7aeda4aea522e74e4fe36759fca80789a613a58a4364f6ecef',
+    '4': 1234,
+    '5': 0
```
janmazak, line 90: I'm not sure one can call this "legacy" --- it does not comply with CIP-15 which does not contain key `5`.

Zeegomo: correct, this (and the related occurrence above) should be changed

CIP-0036/README.md
Zeegomo: yes, would "The signature metadata is added to the transaction metadata under key 61285" be better?

janmazak: Yes, definitely.

CIP-0036/README.md
```diff
@@ -0,0 +1,172 @@
...
+and that all voting powers are exact ADA.
+
+### Example 
+
+Voting registration example:
+```
+61284: {
+  // delegations - CBOR byte array 
+  1: [("0xa6a3c0447aeb9cc54cf6422ba32b294e5e1c3ef6d782f2acff4a70694c4d1663", 1), ("0x00588e8e1d18cba576a4d35758069fe94e53f638b6faf7c07b8abd2bc5c5cdee", 3)],
+  // stake_pub - CBOR byte array
```
SebastienGllmt, line 89: Is there a reason only allow public keys and not script addresses? Only supporting public keys would block multisigs from delegating their vote. It wouldn't add any complexity to the voting app because at the end of the day you're delegating to a (non-script) voting key

Zeegomo: No particular reasons besides increased complexity on the backend to validate such scripts. Plutus for sure seems out of scope for now, but I'm worried even native scripts might be a bit of a stretch atm, considering required changes to the format as well. 
Do we know how common / requested this feature is? We can add such functionality in the future if needed

SebastienGllmt: @Zeegomo I think right now there's 100M+ locked in native script multisigs including ADA owned by the Milkomeda DAO

KtorZ:
@kevinhammond @mark-stopka @Zeegomo 

In the absence of Mark, I have updated the [original PR from Giacomo](https://github.com/cardano-foundation/CIPs/pull/188) to include commits from this PR that made the CIP a standalone one (I also took the liberty to amend the authors accordingly, removing Sebastien, Rhinor and Mikhail from the list, but adding Mark to it).  

As discussed during today's CIP meeting, we agreed that in order to proceed with this proposal, the CDDL format for the voting key must be changed to be forward-compatible with the future introduction of different types of credentials (e.g. native multi-sig scripts) and not only staking keys. This could be achieved by representing it as a classic sum-type using tags in CBOR. 

The changes will be done on #188.

gitmachtl:
What would be the outcome if you use a vote public key multiple times in the registration/delegation like:
```js
{
61284: {
   1: [
      [h'51F117D26E29AEA7DB3D1F2F874AB5F585F619A95AED6D71D31A7404CB6557B5', 1],
      [h'51F117D26E29AEA7DB3D1F2F874AB5F585F619A95AED6D71D31A7404CB6557B5', 1], 
      [h'51F117D26E29AEA7DB3D1F2F874AB5F585F619A95AED6D71D31A7404CB6557B5', 1], 
      [h'51F117D26E29AEA7DB3D1F2F874AB5F585F619A95AED6D71D31A7404CB6557B5', 1], 
      [h'51F117D26E29AEA7DB3D1F2F874AB5F585F619A95AED6D71D31A7404CB6557B5', 1], 
      [h'51F117D26E29AEA7DB3D1F2F874AB5F585F619A95AED6D71D31A7404CB6557B5', 1], 
      [h'51F117D26E29AEA7DB3D1F2F874AB5F585F619A95AED6D71D31A7404CB6557B5', 1], 
      [h'51F117D26E29AEA7DB3D1F2F874AB5F585F619A95AED6D71D31A7404CB6557B5', 1]
      ],
   2: h'51F117D26E29AEA7DB3D1F2F874AB5F585F619A95AED6D71D31A7404CB6557B5', 
   3: h'E0C13582AEC9A44FCC6D984BE003C5058C660E1D2FF1370FD8B49BA73F', 
   4: 123456789, 
   5: 0
   },

61285: {
   1: 
h'8C8C35BC2E3D08F0D0A418A1098E8BF85E42D0A58D7198DF0EAE739866BA3F8D4DB2C84D24F83923259BEB40691456E2B5DE366156B38DDAF1A95AF8D8749E05'
   }
}
```
Will it only count a single time, or will the voting power accumulate? Or is this an invalid delegation/registration at all?

gitmachtl:
I would propose, to add a line in the CIP that is saying: "no duplicated public-voting-key entries are allowed, otherwise such a delegation metadata content would be invalid".

Frontend tools can also take care to make it clear for the user.

gitmachtl:
Created a PR: https://github.com/cardano-foundation/CIPs/pull/335

PR #206, opened 2022-02-01
CIP-0030 | deprecate signTx & add chain signing

SebastienGllmt:
The signTx endpoint for the dApp connector needs to change. The way we used to have a list of arguments with an optional argument makes any breaking change in the future extremely tedious. It's much better to just provide a JSON object so we can add new features to it

One of the reasons I want to do this is so that we can add the ability for dApps to request signing a chain of transactions. We can achieve this by having the dApp itself provide hints about the contents of the input for the next transaction even if they don't appear on-chain yet.

## FAQ

Q: Can't wallets just keep track of pending txs so they know how to sign other transactions in the chain
A: This doesn't work in the generic sense. Here are some cases where this approach would fail:
- Some of the inputs may not belong to the user (ex: the dApp may have its own transactions it needs to broadcast as part of its flow)
- There is no guarantee the dApp submits transactions through the wallet's submit endpoint

### DIFF
--- a/CIP-0030/README.md
+++ b/CIP-0030/README.md
@@ -227,11 +227,26 @@ Errors: `APIError`
 
 Returns the reward addresses owned by the wallet. This can return multiple addresses e.g. CIP-0018.
 
-### api.signTx(tx: cbor\<transaction>, partialSign: bool = false): Promise\<cbor\<transaction_witness_set>>
+### 👎 api.signTx(tx: cbor\<transaction>, partialSign: bool = false): Promise\<cbor\<transaction_witness_set>>
 
 Errors: `APIError`, `TxSignError`
 
 Requests that a user sign the unsigned portions of the supplied transaction. The wallet should ask the user for permission, and if given, try to sign the supplied body and return a signed transaction. If `partialSign` is true, the wallet only tries to sign what it can. If `partialSign` is false and the wallet could not sign the entire transaction, `TxSignError` shall be returned with the `ProofGeneration` code. Likewise if the user declined in either case it shall return the `UserDeclined` code. Only the portions of the witness set that were signed as a result of this call are returned to encourage dApps to verify the contents returned by this endpoint while building the final transaction.
+
+
+### ⚗️ api.signTx(request: {
+	tx: cbor\<transaction>,
+	partialSign?: bool = false,
+	inputHints?: cbor\<transaction>[] = []
+}): Promise\<cbor\<transaction_witness_set>>
+
+Errors: `APIError`, `TxSignError`
+
+Requests that a user sign the unsigned portions of the supplied transaction. The wallet should ask the user for permission, and if given, try to sign the supplied body and return a signed transaction.
+
+- `partialSign`: if true, the wallet only tries to sign what it can. If false and the wallet could not sign the entire transaction, `TxSignError` shall be returned with the `ProofGeneration` code. Likewise if the user declined in either case it shall return the `UserDeclined` code. Only the portions of the witness set that were signed as a result of this call are returned to encourage dApps to verify the contents returned by this endpoint while building the final transaction.
+
+- `inputHints`: to allow dApps to chain transactions together, they may need to provide to the wallet hints about some of the inputs of the transaction that are not on-chain yet. In order to do this in a trustless way, we have to provide the full transaction body (so that the wallet can hash the tx body and compare it to the hash in the inputs). Providing the full transaction instead of just the transaction body also allows access to any scripts or data in the witness that may be useful for the wallet. Note these hints could be use by the wallet to resolve regular transaction inputs, collateral or any other feature of Cardano in the future
 
 ### api.signData(addr: cbor\<address>, sigStructure: cbor\<Sig_structure>): Promise\<Bytes>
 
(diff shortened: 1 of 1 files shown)


CIP-0030/README.md
```diff
@@ -227,12 +227,23 @@ Errors: `APIError`
...
+
+- `partialSign`: if true, the wallet only tries to sign what it can. If false and the wallet could not sign the entire transaction, `TxSignError` shall be returned with the `ProofGeneration` code. Likewise if the user declined in either case it shall return the `UserDeclined` code. Only the portions of the witness set that were signed as a result of this call are returned to encourage dApps to verify the contents returned by this endpoint while building the final transaction.
+
+- `inputHints`: to allow dApps to chain transactions together, they may need to provide to the wallet hints about some of the inputs of the transaction that are not on-chain yet. In order to do this in a trustless way, we have to provide the full transaction body (so that the wallet can hash the tx body and compare it to the hash in the inputs). Providing the full transaction instead of just the transaction body also allows access to any scripts or data in the witness that may be useful for the wallet. Note these hints could be use by the wallet to resolve regular transaction inputs, collateral or any other feature of Cardano in the future
```
SebastienGllmt, line 245: The main part I'm not entirely sure about is whether or not these input hints should be the full transaction or just the tx body. For chaining transactions, you just need the tx body. For things like data hints of script hints, we could make these separate fields instead of putting this all as part of the same field.  Open to thoughts on this.

longngn: I have a question, why can't we just sign continuing txs by calling `signTx` multiple times? Does the wallet implementation check for existing input and throw error if inputs don't exist whenever we call `signTx`? If this is the case, then isn't the solution is to change the wallet implementation to not check for existing inputs?
On a side note, to support chaining tx, I think we would also need support on the infrastructure side (i.e. Blockfrost) to submit multiple txs in a single API call to the same node

SebastienGllmt: @longngn if the wallet doesn't have access to the input, it can't know which input should be signed with which key

vsubhuman (approved):

CIP-0030/README.md
SebastienGllmt: I asked around and it seems nobody needs the full tx so just the tx body should be enough for everyone

MarcelKlammer: To present the user will all information about a transaction, including metadata, I would like to see transactions instead of transaction bodies as hints.

MarcelKlammer: Apart from Allowing to review all non on-chain transactions at once that let to this input, no.

MarcelKlammer (approved):

KtorZ:
@SebastienGllmt would you mind providing a quick update on that one? Is still relevant / has it been added to CIP-0030 implementations already?

SebastienGllmt:
This is still an important CIP that should be implemented. It just got caught up in the CIP30 governance discussion and so it stalled waiting for a decision on that front

rphair:
That's good to know @SebastienGllmt but then this needs a rebase on CIP-0030 in `master` particularly after the changes in https://github.com/cardano-foundation/CIPs/pull/446.  I tried to resolve the merge conflict but it would have involved taking material from both sides & so would have had to ask you to proofread the result anyway.

Ryun1:
Hey @SebastienGllmt ,

This proposal was discussed today in CIP editors call 68. Are you still interested in pursing this amendment or perhaps there has been a work around discovered? because as it stands I do not think it is advisable to continue to alter CIP-30. Rather I would strongly recommend any more alterations to CIP-30 utilize the extensibility mechanism introduced in [#446](https://github.com/cardano-foundation/CIPs/pull/446). This is to avoid the past issues of versioning of CIP-30 for both dApps and wallets, see [#446](https://github.com/cardano-foundation/CIPs/pull/446) for discussions.

I would suggest packaging this in with other CIP-30 amendments (such as [#443](https://github.com/cardano-foundation/CIPs/pull/443) and/or maybe [#206](https://github.com/cardano-foundation/CIPs/pull/206)) to create a multi-functional CIP-30 extension. I mentioned this idea in this [issue comment](https://github.com/cardano-foundation/CIPs/issues/489#issuecomment-1483409155). Would you be willing to pursue something like this?

MartinSchere:
We would like to re-visit this PR. It's important now that more dApps are using transaction chaining.

MartinSchere (approved):

joacohoyos (approved):

TURTLdao:
> We would like to re-visit this PR. It's important now that more dApps are using transaction chaining.

Agreed, this PR would benefit the whole of Cardano if it was merged

Ryun1:
Although I agree that transaction chaining is a great feature to support, altering the CIP-30 base API would be a mistake, due to lack of versioning.
Instead, this would make a great CIP-30 extension, if someone wishes to take this forward.
Although, with the landscape of #587 this might not be needed.

Due to the age of this PR and it not fitting with the current landscape of CIP-30 extensions, I think we can move to close it cc @rphair @Crypto2099

rphair:
In sympathy with https://github.com/cardano-foundation/CIPs/pull/209#issuecomment-1988228035 and the last comments from @Ryun1 I'm closing this by quorum but please @Crypto2099, @SebastienGllmt please re-open (after a proper rebase) if there's a need & someone can carry this forward.

PR #207, opened 2022-02-01
Update schema/readme with new ticker max length

Quantumplation:
In the wild and on other chains, tickers get as long as 9 characters.

Of relevance to me, SUNDAE is 6 characters, but CoinMarketCap recognizes
a MILKSHAKE token with 9 characters.

See https://ethereum.stackexchange.com/questions/25619/is-there-length-limits-on-token-symbols
for real world data.

Also relevant: https://github.com/input-output-hk/offchain-metadata-tools/pull/45

### DIFF
--- a/CIP-0026/README.md
+++ b/CIP-0026/README.md
@@ -178,7 +178,7 @@ The following properties are considered well-known, and the JSON in their values
 {
   "type": "string",
   "description": "A short identifier for the metadata subject, suitable to show in listings or tiles.",
-  "maxLength": 5,
+  "maxLength": 9,
   "minLength": 2
 }
 ```
--- a/CIP-0026/schema.json
+++ b/CIP-0026/schema.json
@@ -50,7 +50,7 @@
   , "ticker":
     { "type": "string"
     , "description": "A short identifier for the metadata subject, suitable to show in listings or tiles."
-    , "maxLength": 5
+    , "maxLength": 9
     , "minLength": 2
     }
   , "decimals":


rphair (approved):
Since there's already been consensus as of https://github.com/input-output-hk/offchain-metadata-tools/pull/45#issuecomment-1016953616 I think this is _fait accompli_.

Quantumplation:
@rphair Sorry about that; I figured those in a position to object would be the ones reviewing that PR.  If you have major objections happy to revisit them now.

The main argument here was that CMC recognizes tickers up to 9 characters, so the cat might be out of the bag on that.

Originally I was going to bump it up to 6 (to cover SUNDAE, since that's what people are using in the wild), but after doing some digging on what other systems support, came to the 9 character PR.

rphair:
> @rphair Sorry about that;

no worries @Quantumplation ... I didn't follow the discussion but read it retroactively & I agree with the proposition and the apparent consensus about the 9-character limit.  As a CIP editor I'm only interested in the consistency between that discussion & this proposed change to the CIP, which I believe we have. 😎

Personally I think it's also a good idea... though the industry as a whole has some 11-character token names, I feel these are no longer "tickers" because they become descriptive rather than abbreviations.  But that is just my opinion & doesn't factor into my approval of this CIP change as originally posted. 🤓

rvl (approved):
Thanks @Quantumplation

SebastienGllmt (approved):

PR #208, opened 2022-02-04
[CIP-0030] Adding `getCollateral` function to the connector API

vsubhuman:
Added function `getCollateral` which is made to be symmetrical with `getUtxos`. The main idea there is to allow the dapps to not care as much as possible (within the existing API in the spec) and for wallets to encapsulate the exact logic of how the collaterals are being handled in reality (whether they have opted-in to https://github.com/cardano-foundation/CIPs/pull/104 or implemented some different approach).


CIP-0030/README.md
```diff
@@ -197,6 +197,16 @@ Errors: `APIError`, `PaginateError`
 
 If `amount` is `undefined`, this shall return a list of all UTXOs (unspent transaction outputs) controlled by the wallet. If `amount` is not `undefined`, this request shall be limited to just the UTXOs that are required to reach the combined ADA/multiasset value target specified in `amount`, and if this cannot be attained, `undefined` shall be returned. The results can be further paginated by `paginate` if it is not `undefined`.
 
+### api.getCollateralUtxos(amount: cbor\<Coin>): Promise\<TransactionUnspentOutput[] | null>
```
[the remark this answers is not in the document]
vsubhuman: Agreed about the positional parameters, can change this to be an object like `{ amount }`.

The `amount` must be a required parameter to incentivize dapps to not request more resources than they actually require. Depending on the specific wallet implementation a difference between asking for an abstract maximum collateral amount and the amount actually required at the moment might be the difference between the wallet responding immediately with already existing utxo and having to bother the user with a reorganisation when it's not critical.

Additionally, not specifying the required amount means the wallet have to just return whatever they have and then the dapp has to either use it all and rist additional unnecessary funds or have to implement some additional logic on how to select minimum amount form the returned utxos.

The name can be changed to `getCollateral` although it poses a question then if it's better for the existing daps to have a function that matches the spec by name, but not by parameters, or better to keep `getCollateral` as the experimental pre-spec name and then wallets can upgrade to support the function as described in the spec with the spec name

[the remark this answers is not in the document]
vsubhuman: I have updated the spec, renaming the function and changing the argument to an object. It now also comments that it is possible for wallets to not require the `amount` parameter, but it is not by-spec and is not recommended. This means the dapps should be working to be compatible with the spec as much as possible, and pass the required parameter, but the wallets are not forced to update, until they are ready.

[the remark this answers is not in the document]
vsubhuman: Think it's the best compromise solution that ensures the existing functions are kept backward compatible, but also allows to define the **right thing** in the spec for other wallets to have various options for the implementation and for dapps to follow.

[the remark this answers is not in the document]
SebastienGllmt: Yeah I would allow specifying the amount optionally. Right now wallets tend to return collateral in a wallet-dependent way, but usually everybody has some sort of arbitrarily-defined minimum (ex: 5 ADA minimum) based on the assumption that probably no dApp will require more than 5. It's not very scientific so it would be nice if the dApp could just specify how much it needs.

CIP-0030/README.md
```diff
@@ -197,6 +197,23 @@ Errors: `APIError`, `PaginateError`
...
+1. Dapps must be motivated to understand what they are doing with the collateral, in case they decide to handle it manually.
+2. Depending on the specific wallet implementation, requesting more collateral than necessarily might worsen the user experience with that dapp, requiring the wallet to make explicit wallet reorganisation when it is not necessary and can be avoided.
+3. If dapps don't understand how much collateral they actually need to make their transactions work - they are placing more user funds than necessary in risk.
+
+So requiring the `amount` parameter would be a by-spec behavior for a wallet. Not requiring it is possible, but not specified, so dapps should not rely on that and the behavior is not recommended.
+
+This shall return a list of all UTXOs (unspent transaction outputs) controlled by the wallet that are required to reach the combined ADA value target specified in `amount` **AND** the best suitable to be used as collateral inputs for transactions with plutus script inputs. If this cannot be attained, `null` shall be returned.
```
SebastienGllmt, line 211: I would prefer if we returned some kind of "Requirement not met" error so that we can specify which requirement wasn't met (which scales better if we need to add more requirements in the future)

CIP-0030/README.md
```diff
@@ -197,6 +197,23 @@ Errors: `APIError`, `PaginateError`
...
+
+The `amount` parameter is required, specified as a `string` (BigNumber) or a `number`, and the maximum allowed value must be agreed to be something like 5 ADA. Not limiting the maximum possible value might force the wallet to attempt to purify an unreasonable amount of ADA just because the dapp is doing something weird. Since by protocol the required collateral amount is always a percentage of the transaction fee, it seems that the 5 ADA limit should be enough for the foreseable future.
```
SebastienGllmt, line 215: Fyi both ccvault and Flint use 5~20 ADA and I think Nami does <50 ADA

CIP-0030/README.md
vsubhuman: Updated the description to `If this cannot be attained, an error message with an explanation of the blocking problem shall be returned.`

CIP-0030/README.md
SebastienGllmt: The bottom of the range (5ADA) is what is meant to make sure you have enough collateral to cover any requirement

The upper part of the range (20ADA / 50 ADA) is just because wallets wants to protect users from accidentally losing a lot of ADA due to a poorly programmed dApp and so they don't want to allow using millions of ADA as collateral. 20/50 was picked as an arbitrary cutoff point so that it's not the end of the world if you lose it.

vsubhuman: Sounds like it then can be just the range from 5 ADA min to 5 ADA max. E.g. if we assume 5 ADA should be enough to cover any requirements then why give dapps any idea to ever have more than that at all? 🤔

SebastienGllmt: Because the user may have a 10 ADA utxo entry by chance, and it would be a bit excessive to say to the user "sorry, you can't use your 10 ADA as collateral because it has to be exactly 5 ADA".

vsubhuman: > Because the user may have a 10 ADA utxo entry by chance, and it would be a bit excessive to say to the user "sorry, you can't use your 10 ADA as collateral because it has to be exactly 5 ADA".

I am confused here. What will happen with the dapp call to `getCollateral` in case they specify they want to get 20 ADA, but there's no UTXO below 50 ADA, for example?

The whole point of what I am describing in the spec is that the `amount` **parameter** limit should be something like 5 ADA, so that **DAPPS** can NOT **request** amounts above 5 ADA. But then the wallet is free to return **anything** as long as it covers the amount the dapp has requested.

Example:
1. Dapp requests 3 ADA
2. Wallet sees there's a 10 ADA utxo
3. Wallet returns the 10 ADA utxo if the wallet wants
4. Dapp doesn't care and just includes it into the tx without having to think

Example 2:
1. Dapp requests 3 ADA
2. Wallet see there are few utxos of 1.4 ADA
3. Wallet returns 3 utxos like `[1.4 , 1.4 , 1.4]` total sum of 4.2
4. Dapp doesn't care and just includes it into the tx without having to think

Example 3:
1. Dapp requests 7 ADA
2. Wallets returns an error saying that's too much and shouldn't be reasonable for a collateral

SebastienGllmt (approved):

KtorZ (approved):
As discussed:

- Let's merge this one

then,

- Move CIP-0030 as active and "set in stone" the current API.
- Define via another CIP a structured process for bringing extensions to this API.

PR #209, opened 2022-02-04
CIP-0030 | Adding optional `networkId` parameter to `.enable`

vsubhuman:
To allow dapps optionally to force user to select only compatible network wallets.

See comment here https://github.com/cardano-foundation/CIPs/pull/151#discussion_r799248001


KtorZ:
I second @SebastienGllmt on this :+1: , would be better to avoid "integer blindness".

mangelsjover:
This is on hold until the governance path can be discussed on CIP 30.

Ryun1:
Due to little activity on this proposal and the current state of CIP-30 extensions, I think we can move to close this one @rphair @Crypto2099

rphair:
Like wise with https://github.com/cardano-foundation/CIPs/pull/323#issuecomment-1988222025 (apparent abandonment, the current framework of network identification being "good enough" in practice) and dire need of a rebase), we're closing by quorum unless a current need to reopen is presented.

PR #211, opened 2022-02-07
Add reference implementation for CIP15

SebastienGllmt:
I noticed there was no reference implementation for CIP15 so I created one


KtorZ (approved):

rphair (approved):
I can't vouch for the implementation itself but I can see its presence supports this CIP 😎

PR #137, opened 2021-10-12
CIP-???? |  On-Chain Token Metadata Standard

savaki:
The following CIP proposes an on-chain metadata standard to work along aside existing off-chain metadata standards.

PR #215, opened 2022-02-09
CIP-35: Plutus Core Evolution

michaelpj:
This CIP proposes a process for making CIP proposals to change Plutus
Core or its interface to the ledger.

### DIFF
--- a/CIP-0035/README.md
+++ b/CIP-0035/README.md
@@ -1,0 +1,302 @@
+---
+CIP: 35
+Title: Plutus Core Evolution
+Authors: Michael Peyton Jones <michael.peyton-jones@iohk.io>
+Comments-Summary: No comments
+Comments-URI: 
+Status: Draft
+Type: Process
+Created: 2022-02-09
+License: CC-BY-4.0
+---
+
+# Evolution of Plutus Core in the Cardano ledger
+
+## Abstract
+
+This CIP proposes a process for proposing changes to Plutus Core itself, and the parts of the Cardano ledger which relate to it.
+
+## Motivation
+
+The Plutus Core language, its builtins, and its interface to the ledger are all likely to evolve significantly over time. There are many reasons for this:
+- We may be able to increase performance, improve safety, or reduce script sizes by changing the language.
+- We may be able to improve performance by providing builtin versions of expensive functions.
+- We may need to provide builtin versions of sensitive functions (e.g. cryptography) in order to ensure access to high-quality implementations.
+- We may find bugs in the implementation that need to be fixed.
+- We may need to change the interface to the ledger in order to represent changes in transaction formats.
+- We may wish to remove elements which have been deprecated due to the addition of improved versions.
+- … and more
+
+At the moment there is no process for making such changes other than the discretion of the core developers. 
+This CIP gives a taxonomy of changes, explains how such changes might be implemented in Cardano, and prescribes processes for proposing such changes.
+
+## Background
+
+This CIP assumes general familiarity with Plutus Core and the Cardano ledger, but we give some brief background here.
+
+### Plutus Core
+
+_Plutus Core_ is a script language used in the Cardano ledger. For the purposes of this document, Plutus Core consists of various _language constructs_, and also _builtin types and functions_.
+
+Plutus Core has a number of builtin types, such as integers, and builtin functions, such as integer addition. 
+Builtin functions provide access to functionality that would be difficult or expensive to implement in Plutus Core code. 
+Builtin functions can operate only over builtin types. Builtin types come with a _size metric_ which is used by costing functions. 
+For example, the size metric for integers returns the bit-size of the integer.
+
+The performance of Plutus Core scripts has two components: how expensive the script actually is to run (_real performance_) and how expensive we say it is to run in the ledger (_model performance_). 
+
+Model performance is calculated by costing _evaluation_ in abstract resource units ("exunits") of CPU and memory. 
+Individual steps of evaluation are costed, and builtin functions must also come with a _costing function_ that provides costing information for them. 
+The costing function for a builtin function is a mathematical function which takes the sizes of the arguments (as computed by their size metrics), and returns an estimate of the budget that will be used to perform that function. 
+For example, the costing function for addition says that the CPU and memory costs are both proportional to the maximum of the sizes of the input integers (with some appropriate constants).
+
+Determining costing functions is done empirically, by running the function in question against a large number of inputs and choosing a costing function that fits the data well.
+
+### Scripts in the Cardano ledger
+
+The Cardano ledger recognizes various kinds of _scripts_ identified by _language_. 
+This language tag is the only way that the ledger has to distinguish between different types of script. 
+Hence if we require different behaviour, we need a new language.
+
+We omit this distinction when talking about Plutus Core: we refer instead to Plutus Core _language versions_, which are actually entirely distinct _languages_ from the perspective of the ledger.
+All Plutus Core language versions which are used in the ledger must be supported forever, in order to be able to validate the history of the chain.
+
+Part of the specification of a language in the ledger is how scripts of that language are run, what arguments they are passed, how those arguments are structured, etc. 
+We refer to this as the _ledger-script interface_.
+
+Languages also have an associated subset of _protocol parameters_ which provide the ability to control some aspects of evaluation without a software update. 
+The most notable example is a set of parameters which parameterize the costing of program execution.
+
+### Types of change
+
+This document considers the following types of change:
+
+1. The Plutus Core language
+    1. Adding a construct to the language
+    2. Removing a construct from the language
+    3. Changing the behaviour of a construct in the language
+    4. Changing the binary format of the language in a backwards-compatible way
+    5. Changing the binary format of the language in a backwards-incompatible way
+2. The Plutus Core builtin functions and types
+    1. Adding a new builtin function or type
+    2. Removing a builtin function or type
+    3. Changing the behaviour of a builtin function or type
+3. The ledger-script interface
+    1. Changing the interface between the ledger and the interpreter
+4. Protocol parameters
+    1. Improving model performance (i.e. changing the costing parameters so that scripts use less budget)
+    2. Regressing model performance (i.e. changing the costing parameters so that scripts use more budget)
+    3. Adding costing parameters
+    4. Removing costing  parameters
+5. Performance of the Plutus Core interpreter
+    1. Improving real performance
+    2. Regressing real performance
+
+### Types of release
+
+Changes to Plutus Core can be released onto Cardano in four ways, with ascending levels of difficulty:
+
+1. A _protocol parameter_ change (PP), taking effect as soon as the new parameters are accepted (in a new epoch).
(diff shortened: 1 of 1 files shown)


CIP-0035/README.md
```diff
@@ -0,0 +1,302 @@
...
+Type: Process
+Created: 2022-02-09
+License: CC-BY-4.0
+---
+
+# Evolution of Plutus Core in the Cardano ledger
+
+## Abstract
+
+This CIP proposes a process for proposing changes to Plutus Core itself, and the parts of the Cardano ledger which relate to it.
```
kozross, line 17: What specifically are these? Considering Plutus' nontrivial dependency footprint, it would be good to have a list of these, even if it's not 100% exhaustive.

CIP-0035/README.md
```diff
@@ -0,0 +1,302 @@
...
+- In the Specification:
+    - Names and types/kinds for the new functions or types.
+    - A source for the implementation (e.g. a library which can be linked against); or a generic description of the functionality which is implementable in any programming language.
+    - In the Rationale:
+    - An argument for the utility of the new builtins.
+
+It SHOULD also include:
+- In the Rationale
+    - Examples of real-world use cases where the new additions would be useful.
+    - A comparison with an implementation using the existing features, and an argument why the builtin is preferable (e.g. better performance).
```
kozross, line 159: Who decides that the requirements to move to Draft have been met?

CIP-0035/README.md
```diff
@@ -0,0 +1,302 @@
...
+
+In order to move to Proposed status, it MUST include:
+- In the Specification:
+    - For new types: a precise description of the measure used for the size of a value of that type.
+    - For new builtin functions: a costing function for the builtin function.
+- In the Rationale:
+    - If an external implementation is provided: an argument that it is trustworthy.
+    - Discussion of how any measures and costing functions were determined.
+- In the Path to Active section:
+    - A named implementor (person or organization) who will implement the support for the new builtins in the Plutus Core interpreter, or a pull request to the `plutus` repository with an implementation.
```
kozross, line 169: Several issues here:

* Again, who decides when the requirements to move to Proposed have been met?
* Based on [here](https://github.com/input-output-hk/plutus/blob/master/plutus-core/cost-model/CostModelGeneration.md) and [here](https://github.com/input-output-hk/plutus/pull/4368#issuecomment-1032025802), the requirement of a costing function being provided is completely unworkable. The only group of people able to cost anything are IOG themselves, and while there is clearly [a goal to extending this](https://github.com/input-output-hk/plutus/issues/4375#issuecomment-1029864939) and [some discussion towards this](https://github.com/input-output-hk/plutus/issues/4384), very non-trivial limitations remain.

CIP-0035/README.md
```diff
@@ -0,0 +1,302 @@
...
+
+In order to move to Active status, the following must be true:
+- The external implementations must be available.
+- The `plutus` repository MUST be updated with an implementation including costing.
+- The Plutus Core specification MUST be updated to include the new builtins.
+- The ledger MUST be updated to include new protocol parameters to control costing of the new builtins.
+- The completion of these steps MUST be tracked in the Path to Active section.
```
kozross, line 176: All the issues raised in the 'move to Proposed' section apply here as well. In addition:

* No mention of testing. Given issues such as [this](https://github.com/input-output-hk/plutus/pull/4284#issuecomment-996185889), [this](https://github.com/input-output-hk/plutus/pull/4322#discussion_r782817268), and (while not from Plutus, _critically_ relevant) [this](https://github.com/input-output-hk/cardano-base/pull/253), the importance and quality of tests cannot be overstated. We also need to formalize practices such as use of Hedgehog, 1-in-a-thousand as the acceptable failure standard for Hedgehog's bad coverage implementation, and so forth: all the above are described in the links.

kozross (changes requested):
In general, I think it's a good idea to have a more specified and clear structure around modifying Plutus and the libraries it depends on. However, I am concerned at the degree of bureaucracy this adds. Consider, for example, [this issue here](https://github.com/input-output-hk/plutus/issues/4233). Based on the current write-up, I see the process looking like this:

1. CIP for this gets sent. We await its approval, however long that takes, deal with whatever discussions and concerns arise, etc. This is complicated by the fact that we have to modify `cardano-base` as well, which by my reading, requires a _separate_ CIP.
1. Once accepted, the process for getting to Draft must be completed. Some as-yet-unclear party decides when this goal is met, which may require more discussion, dealing with more concerns, etc.
1. Once gotten to Draft, the process for Proposed must be completed. Assuming costing can be done at all (see my note), it's yet another round of the same.
1. Once gotten to Proposed, the process for getting to Active must be completed. You get the idea by now.

This would take something that is already a lengthy and complex process, make it _lengthier_ and _more_ complex, while arguably as-written still failing to solve key issues (for example, see my statement regarding testing). Additionally, a document such as this should document proper practices and give specifics about what needs to be done (again, see my comment regarding testing): at the moment, it's not even clear who gets to decide when we can move, for example, to Draft. Additionally, if my reading is correct, for something like the linked task, this process might have to be completed _twice_.

In short, much more clarity is needed, and the amount of steps specified here need to be considered for usefulness: personally, they seem excessive, and fairly unclear, at this stage.

CIP-0035/README.md
michaelpj: Does the list in "Types of change" not answer this question?

CIP-0035/README.md
michaelpj: Implementation may be difficult :) But in general the idea isn't meant to be that the implementor has to do everything and the Plutus team just sit back and chill. As with any contribution, if we think we should accept it we're going to help. In the short term that means that the Plutus team will probably do the costing functions.

(See my main comment for why I didn't include more "implementation process" stuff like this in the text.)

CIP-0035/README.md
michaelpj: As discussed in my main comment, this is an implementation issue so is out of scope.

michaelpj:
> However, I am concerned at the degree of bureaucracy this adds, and what we actually gain by it.

Partially, the bureaucracy is the point. I believe that the bar for changing the blockchain _should_ be higher. We should be sure that many people in the community want the change, and that the change is well-specified. 

Moreover, the idea is for the development of Cardano to be increasingly driven by people other than IOG, and that includes the design work. It's an easier process for the community to say "we want X" and get IOG to do the design work, but we *want* the community to do the design work!

> Consider, for example, https://github.com/input-output-hk/plutus/issues/4233.

That's a great example. Let me sketch out what I think the process for that would have been:

1. You open a Draft CIP, which is more-or-less like the issue that you opened, but with more details.
2. There is discussion on the CIP. Hopefully, this will surface any difficulties due to the increased number of eyes, for example, it might have identified the fact that you wanted Schnorr signatures.
3. In parallel with that, you can start on the implementation work. The CIP doesn't say anything about when the implementation has to happen. If you can persuade the `cardano-base` maintainers to merge your PR with the implementation etc. then you can go ahead and do that.
4. The remaining details are hammered out in discussion on the CIP and potentially the prototype implementation. Somewhere around here the design is completed enough that the CIP becomes Proposed.
5. The implementation gets reviewed and accepted (or not! making a CIP doesn't guarantee that something will get accepted! it's just an open design doc), and the CIP gets changed to Active.

I think this would have been not that much more onerous than what actually happened. The main differences would be:
- More design work up front for the proposer. This is, as I said, deliberate, but is annoying for you.
- More eyes on the proposal by the community. This is important: part of the point is that we _shouldn't_ just be changing the blockchain because one community member makes a `plutus` issue and I go "seems fine".

> This is complicated by the fact that we have to modify cardano-base as well, which by my reading, requires a separate CIP.

Why do you think that? That wasn't the intention. 

> Some as-yet-unclear party decides when this goal is met

As I commented inline, it's the CIP editors, who are in general responsible for assessing the progression of CIPs.

> Additionally, a document such as this should document proper practices and give specifics about what needs to be done

Unfortunately, this document can't be as precise as I would like, since _implementation_ is out of scope for CIPs. A lot of requirements are buried in "it is the responsibilty of the implementor to get their change accepted", such as testing. I agree that that makes it less helpful than it could be. When I realised this I considered not even making a CIP and just adding this to the `plutus` `CONTRIBUTING` doc or something. But I think it's still valuable to use the CIP process for the _design_ part of the process, even if it has to leave something out.

At least, I'll try and clarify the text to make the implementation requirements more obvious.

CIP-0035/README.md
michaelpj: Or perhaps a better way of saying it is: there might be named _implementors_ which could be "MLabs and the Plutus team (for cost modelling)"

CIP-0035/README.md
kozross: Thanks for clarifying this: an inline link would help make that clearer in the text.

L-as (approved):
I think this is fine honestly.

CIP-0035/README.md
michaelpj: I guess it's this sentence that's mostly confusing. I thought "This document considers the following types of change: <explicit list of changes>" was pretty explicit but okay.

Benjmhart: i think the direct question here is "My change requires modifications to `cardano-base` as well as the plutus interpreter, does this require two separate CIP procedures?"

CIP-0035/README.md
Benjmhart: I'd like to see a process for generating cost reports when an implementation is ready to go,  the point should be to _enable_ the community to do this in the correct way, rather than gating it off to specific implementors.   Teach them to fish.  Ideally the request process would not need to go through a CIP editors meeting, but that may become a standard practice.

CIP-0035/README.md
Benjmhart: this seems to be answered 'No' in the below longform comment

michaelpj: What I was trying to clarify is what in the text makes it seem that you *would* need a CIP to modify `cardano-base`, given that "changes to dependencies" is not listed as a type of change covered by the CIP.

michaelpj:
Updated:
- I've attempted to clarify the scope of the CIP a bit more, and that it applies only to the listed types of change.
- I've made it clearer that implementation is out of scope and may introduce more work. I've also removed the requirement that the CIP include named implementors: now it just says that in order to reach Active, it must have been implemented. How that happens (who does it, how they manage it), doesn't need to be specified.

rhyslbw:
CIP Identifier clash with https://github.com/cardano-foundation/CIPs/pull/137

michaelpj:
> CIP Identifier clash with https://github.com/cardano-foundation/CIPs/pull/137

My understanding is that the Editors are responsible for allocating a CIP number anyway: I've only picked one here because otherwise you can't put it in a logical subdirectory :shrug: 

We can change the number of whichever one doesn't get merged first ;)

kozross (approved):

KtorZ (approved):

SebastienGllmt (approved):

michaelpj:
Given the lack of further comments and the approvals from the editors, I've changed the status to Active to represent that fact that we will be using this process going forwards.

KtorZ:
@michaelpj whatever you did, you _might_ want to push it :blush: ?

PR #217, opened 2022-02-10
CIP-???? | Collateral reward

SebastienGllmt:
This PR describes adding a scheme to transfer change in collateral to the reward account during phase-2 validation failure which removes the need for wallets to explicitly set collateral (which is bad UX) and improves safety of collateral at the same time.

This approach, however, has multiple problems (outlined in the CIP) and would probably be best combined with a more general discussion about reviving the idea of chimeric accounts (as it would make implementing this CIP a much more natural solution)

This PR is part of a series of different ways we can improve collateral handling in Cardano. You can find the three options below

[Collateral Key](https://github.com/cardano-foundation/CIPs/pull/104)
[Collateral Output](https://github.com/cardano-foundation/CIPs/pull/216)
[Collateral Reward](https://github.com/cardano-foundation/CIPs/pull/217)

### DIFF
--- a/CollateralReward/README.md
+++ b/CollateralReward/README.md
@@ -1,0 +1,66 @@
+---
+CIP: ?
+Title: Collateral Reward
+Authors: Sebastien Guillemot <seba@dcspark.io>
+Status: Draft
+Type: Standards
+Created: 2022-02-10
+License: CC-BY-4.0
+---
+
+# Abstract
+
+This document describes providing change on phase-2 validation through the reward account
+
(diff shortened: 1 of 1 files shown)


KtorZ:
@SebastienGllmt did you intend to deprecate / make obsolete that one?

KtorZ:
Closing this now, doing some cleanup. It's been "likely deprecated" for a while, without any new activity. Feel free to re-open with new elements if needs be.

PR #220, opened 2022-02-11
CIP-0381? | Plutus support for pairings over curve Bls12-381

iquerejeta:
This CIP proposes to include plutus support for pairings over curve Bls12-381

### DIFF
--- a/CIP-0037/CIP-0037.md
+++ b/CIP-0037/CIP-0037.md
@@ -1,0 +1,136 @@
+###### tags: `CIP`
+# CIP-37
+---
+CIP: 37
+Title: Plutus support for Pairings over BLS12_381
+Authors: Iñigo Querejeta-Azurmendi <inigo.querejeta@iohk.io>
+Discussions-To: https://github.com/cardano-foundation/CIPs/pull/219
+Comments-URI: https://github.com/cardano-foundation/CIPs/pull/219
+Status: Draft
+Type: Standards Track
+Created: 2022-02-11
+License: Apache-2.0
+---
+
+
+## Abstract
+This CIP proposes an extension of the current plutus functions to provide support for basic operations over BLS12_381 curve to the plutus language. We expose some candidate implementations, and describe clearly the benefits that this would bring. In a nutshell, pairing friendly curves will enable a large number of cryptographic primitives that will be essential for the scalability of Cardano. 
+
+## Motivation
+Pairing Friendly Curves are a type of curves that provide the functionality of computing pairings. A pairing is a binary function that maps two points from two groups to a third element in a third target group. For a more in-depth introduction to pairings, we recommend reading [Pairings for Beginners](https://www.craigcostello.com.au/tutorials) or [Pairings for Cryptographers](https://eprint.iacr.org/2006/165). In a nutshell, pairings enable a range of interesting cryptographic primitives which could (and will) be used for scaling Cardano and many other use cases. 
+
+* **Sidechains** are a crucial component for the scalability of Cardano, and its interoperability with other chains/tokens/smart contracts. However, sidechains need to periodically commit their state to the Cardano mainnet to provide the same security guarantees as the latter. This periodical commitment is performed through a threshold signature by the dynamic committee of the Sidechain. The most interesting construction for medium sized committees is presented in the paper [Proof of Stake Sidechains](https://cointhinktank.com/upload/Proof-of-Stake%20Sidechains.pdf), in section 5.2 and requires pairings. We have yet not found an acceptable solution that does not require pairings.
+* **Zero Knowledge Proofs** are an incredibly powerful tool. There exist different types of zero knowledge proofs, but the most succinct (and cheaper to verify) rely on pairings for verification. Zero knowledge proofs can be used to make Mithril certificates, or sidechain checkpoints even more succinct, or to create layer 2 solutions to provide scalability (by the means of [ZK-Rollups](https://ethereum.org/en/developers/docs/scaling/layer-2-rollups/), which are used to scale Ethereum, e.g. [Loopring](https://loopring.org/#/), [zkSync](https://zksync.io/) or [Aztec](https://aztec.network/) among others)
+* **Hydra** is another crucial component for scalability of Cardano (there is a series of blog posts available, with a good summary of the solution [here](https://iohk.io/en/blog/posts/2021/09/17/hydra-cardano-s-solution-for-ultimate-scalability/)). Hydra relies on a multisignature scheme, where all participants of the side channel need to agree on the new state. This can be achieved with non-pairing friendly curves (as it is currently designed), but pairing based signature schemes provide much more elegant constructions that reduce interaction among signers. 
+* **Mithril** currently does not require plutus support. However, Mithril, as a technology, allows for signature generation representing all stakeholders of Cardano (or any proof of stake system). These types of certificates might eventually be used for certifying sidechains, and plutus support will be crucial. Again, Mithril relies on pairing based signatures. 
+* **ATALA** is a decentralized identification mechanism. One of the properties they want to provide is anonymity: users can selectively disclose attributes of their certificate or prove statements regarding them without disclosing their identity. Up to date, the most recent, efficient and interesting solutions to provide these are pairing based ([Hyperledger Fabric/Idemix standardisation effort](https://hyperledger-fabric.readthedocs.io/en/release-2.2/idemix.html#underlying-cryptographic-protocols), [coconut credentials used by Nym](https://blog.nymtech.net/nyms-coconut-credentials-an-overview-4aa4e922cd51), among others)
+
+
+### Reason for choosing BLS12-381
+The reason for choosing BLS12-381 over BN256 curve is that the former is claimed to provide 128 bits of security, while the latter was reduced to 100 bits of security after the extended number field sieve (a new algorithm to compute the discrete logarithm) was [shown to reduce the security](https://eprint.iacr.org/2016/1102.pdf) of these curves. 
+
+An [EIP](https://eips.ethereum.org/EIPS/eip-2537) for precompiles of curve BLS12-381 already exists, but has been stagnant for a while. Nonetheless, Zcash, MatterLabs and Consensys support BLS12-381 curve, so it is certainly widely used in the space. 
+
+Further reading regarding curve BLS12-381 can be found [here](https://hackmd.io/@benjaminion/bls12-381) and the references thereof cited. For level of detail required in this document, it is sufficient to understand that one can define a pairing over BLS12-381, where a pairing is a map: $e:G_{1}\times G_{2}\rightarrow G_{T}$, which satisfies the following properties:
+
+* Bilinearity: $\forall a,b\in F_{q}^{*},\ \forall P\in G_{1},Q\in G_{2}:\ e\left(aP,bQ\right)=e\left(P,Q\right)^{ab}$
+* Non-degeneracy: $e\neq 1$
+* Computability: There exists an efficient algorithm to compute $e$
+
+where $G_1, G_2$ and $G_T$ are three distinct groups of order a prime $q$. Given that all three groups have the same order, we can refer to the scalars of each group using the same type/structure. 
+
+## Specification
+We now provide the technical specification. 
+
+### Names and types/kinds for the new functions or types
+The added types will be the following, all of which can be represented as a byte array. Even if these types are equivalent to byte arrays of a given size, I think it makes sense to include these types for clarity of the code.
+* `Bls12381G1Element`
+* `Bls12381G2Element`
+* `Bls12381GTElement`
+* `Bls12381Scalar`
+
+We need to support the binary operation of $G_1$ and $G_2$ (which are additive groups) as well as some operations over the scalars. In particular, we need the following functions: 
+
+* Group operations: 
+    * `BLS12_381_G1_add :: Bls12381G1Element -> Bls12381G1Element -> Bls12381G1Element`
+    * `BLS12_381_G1_mult :: Bls12381Scalar -> Bls12381G1Element -> Bls12381G1Element`
+    * `BLS12_381_G2_add :: Bls12381G2Element -> Bls12381G2Element -> Bls12381G2Element`
+    * `BLS12_381_G2_mult :: Bls12381Scalar -> Bls12381G2Element -> Bls12381G2Element`
+* Scalar operations: 
+    * `BLS12_381_Scalar_mult :: Bls12381Scalar -> Bls12381Scalar -> Bls12381Scalar`
+    * `BLS12_381_Scalar_add :: Bls12381Scalar -> Bls12381Scalar -> Bls12381Scalar`
+    * `BLS12_381_Scalar_inv :: Bls12381Scalar -> Bls12381Scalar`
+    * `BLS12_381_Scalar_neg :: Bls12381Scalar -> Bls12381Scalar`
+* Pairing operations:
+    * `BLS12_381_pairing :: Bls12381G1Element -> Bls12381G2Element -> Bls12381GTElement` 
+
+
+### Source implementations
+* [BLST](https://github.com/supranational/blst) library
+* [ZCash](https://github.com/zkcrypto/pairing)'s library
+* [Consensys](https://github.com/ConsenSys/gnark)' library
+
+Other libraries of interest
+* [Ethereum support for BLS12-381](https://eips.ethereum.org/EIPS/eip-2537). Not directly relevant as this is a Ethereum Improvement Proposal for a precompiled solidity contracts.
+
+### Comparison with existing function
+We present what would be the alternatives of using pairings in the different usecases presented above. 
+
+* Sidechain bridges using the current technology would rely on either of the two possibilities: 
+    * Require the bridge committee to interact during signature, or to rely on a precomputation phase. Current solutions only support non-robust signature schemes, meaning that if one signer misbehaves, the whole signature procedure needs to be restarted. This could seriously hinder sidechains.
+    * Non-aggregation of signatures. This would result in a linear "checkpoint certificate" with respect to the number of signers (both in communication and computation complexity). Basically, all committee members need to submit their signature, and the smart contract needs to verify all ed25519 signatures. 
+* Zero Knowledge Proofs cannot be verified with current functions available in Plutus. There exists proofs that can be instantiated over non-pairing friendly curves, but these result in logarithmic sized proofs with respect to the computation to prove, while solutions that rely on pairings can be represented more concisely. 
+
+### Reason for exposing curve operations API
+One might be concerned of why we are exposing such low-level primitives, instead of exposing higher level protocol functions, such as $\texttt{VerifyBlsSignature}$ or $\texttt{VerifyZKP}$. The motivation behind that is because pairings can enable a big number of use cases, and covering all of those can considerably extend the list of required functions.. 
+
+### Curve specifications
+BLS12 curve is fully defined by the following set of parameters (coefficient A=0 for all BLS12 curves). Taken from [EIP 2537](https://eips.ethereum.org/EIPS/eip-2537):
+```
+Base field modulus = 0x1a0111ea397fe69a4b1ba7b6434bacd764774b84f38512bf6730d2a0f6b0f6241eabfffeb153ffffb9feffffffffaaab
+B coefficient = 0x000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000004
+Main subgroup order = 0x73eda753299d7d483339d80809a1d80553bda402fffe5bfeffffffff00000001
+Extension tower
+Fp2 construction:
+Fp quadratic non-residue = 0x1a0111ea397fe69a4b1ba7b6434bacd764774b84f38512bf6730d2a0f6b0f6241eabfffeb153ffffb9feffffffffaaaa
+Fp6/Fp12 construction:
+Fp2 cubic non-residue c0 = 0x000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000001
+Fp2 cubic non-residue c1 = 0x000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000001
+Twist parameters:
+Twist type: M
+B coefficient for twist c0 = 0x000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000004
+B coefficient for twist c1 = 0x000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000004
+Generators:
+G1:
+X = 0x17f1d3a73197d7942695638c4fa9ac0fc3688c4f9774b905a14e3a3f171bac586c55e83ff97a1aeffb3af00adb22c6bb
+Y = 0x08b3f481e3aaa0f1a09e30ed741d8ae4fcf5e095d5d00af600db18cb2c04b3edd03cc744a2888ae40caa232946c5e7e1
+G2:
+X c0 = 0x024aa2b2f08f0a91260805272dc51051c6e47ad4fa403b02b4510b647ae3d1770bac0326a805bbefd48056c8c121bdb8
+X c1 = 0x13e02b6052719f607dacd3a088274f65596bd0d09920b61ab5da61bbdc7f5049334cf11213945d57e5ac7d055d042b7e
+Y c0 = 0x0ce5d527727d6e118cc9cdc6da2e351aadfd9baa8cbdd3a76d429a695160d12c923ac9cc3baca289e193548608b82801
+Y c1 = 0x0606c4a02ea734cc32acd2b02bc28b99cb3e287e85a763af267492ab572e99ab3f370d275cec1da1aaa9075ff05f79be
+Pairing parameters:
+|x| (miller loop scalar) = 0xd201000000010000
+x is negative = true
+```
+One should note that base field modulus is equal to 3 mod 4 that allows an efficient square root extraction.
+
+## Moving to Proposed
+### Trustworthiness of implementations
+* [BLST](https://github.com/supranational/blst) library—[audited by NCC Group](https://research.nccgroup.com/wp-content/uploads/2021/01/NCC_Group_EthereumFoundation_ETHF002_Report_2021-01-20_v1.0.pdf) and [being formally verified](https://github.com/GaloisInc/BLST-Verification) by Galois
+* [ZCash](https://github.com/zkcrypto/pairing)'s library---used by zcash to validate ZKSNARKs at scale. 
+* [Consensys](https://github.com/ConsenSys/gnark)' library---
+### Costing of function
+We now describe the relation of evaluating the functions described above with respect to ed25519 verification. This estimates will be more precise once we decide which library to go forward with. We represent each function cost as a ratio wrt to ed25519 verification (op / ed25519_verif):
+* `BLS12_381_G1_add`: 0,016
+* `BLS12_381_G1_mult`: 4,018
+* `BLS12_381_G2_add`: 0,054
+* `BLS12_381_G2_mult`: 13,8
+* `BLS12_381_Scalar_mult`: 0,0000003
+* `BLS12_381_Scalar_add`: 0,0000003
+* `BLS12_381_Scalar_inv`: 0,036
+* `BLS12_381_Scalar_neg`: 0,0000003
+* `BLS12_381_pairing`: 30,58
+### Plutus implementor
+IOHK internal
+
--- a/CIP-0037/CIP-0037.md
+++ b/CIP-0037/CIP-0037.md
@@ -31,13 +31,13 @@ The reason for choosing BLS12-381 over BN256 curve is that the former is claimed
 
 An [EIP](https://eips.ethereum.org/EIPS/eip-2537) for precompiles of curve BLS12-381 already exists, but has been stagnant for a while. Nonetheless, Zcash, MatterLabs and Consensys support BLS12-381 curve, so it is certainly widely used in the space. 
 
-Further reading regarding curve BLS12-381 can be found [here](https://hackmd.io/@benjaminion/bls12-381) and the references thereof cited. For level of detail required in this document, it is sufficient to understand that one can define a pairing over BLS12-381, where a pairing is a map: $e:G_{1}\times G_{2}\rightarrow G_{T}$, which satisfies the following properties:
+Further reading regarding curve BLS12-381 can be found [here](https://hackmd.io/@benjaminion/bls12-381) and the references thereof cited. For level of detail required in this document, it is sufficient to understand that one can define a pairing over BLS12-381, where a pairing is a map: e:G_1 X G_2 -> G_T, which satisfies the following properties:
 
-* Bilinearity: $\forall a,b\in F_{q}^{*},\ \forall P\in G_{1},Q\in G_{2}:\ e\left(aP,bQ\right)=e\left(P,Q\right)^{ab}$
-* Non-degeneracy: $e\neq 1$
-* Computability: There exists an efficient algorithm to compute $e$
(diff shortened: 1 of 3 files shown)


L-as:
I don't think this approach is viable. Even if you add support for pairings, executing the algorithms in question in the CEK machine would be horrifically slow. It is a much better idea to add support for a single universal proof system, and I highly recommend Halo 2. The main blocker is the API for recursive proofs not being done yet, but that should change soon AFAICT.

iquerejeta:
I've added some estimates of the number of operations for each of the different application listed in the CIP. Even for Plonk verification I get estimates that could fit within the plutus execution budget. In particular the significant operations in a verification are 
* 6 hash computations
* 1 vanishing polynomial (1 scalar exponantiation)
* Quotient polynomial evaluation (1 scalar inverse)
* First part of batch poly commitment (7 scalar mults and 6 additions in G1)
* Fully batched poly commitment (7 scalar mults and 8 additions over G1)
* Group encoded batch verification (1 scalar mult over G1)
* Batch-validate all evaluations (3 scalar mults and 4 additions over G1, plus two pairing evaluations + equality check)

@L-as, could you elaborate on why the CIP is not a viable approach in general for the other primitives please? And also, could you elaborate on why you think that Halo 2 (with IPA) is the best choice for a universal proof system (over Plonk, for instance)? I believe that verification of relatively complex circuits with Halo2 will most probably exceed plutus execution budget. 

Thanks!

CIP-0037/CIP-0037.md
```diff
@@ -134,3 +134,50 @@ We now describe the relation of evaluating the functions described above with re
...
+* Check that $n\leq t - k$ 
+* $n\cdot \log_2(t)$ hash computations
+* Equality check of two hashes
+* $n$ $\mathbb{G}_1$ additions and $1$ $\mathbb{G}_1$ negation [WE NEED TO ADD THIS]
+* 2 pairing evaluations + equality check
+
+## Mithril (using trivial proving system)
+Let $k$ be the number of submitted signatures (as above, $k$ is a security parameter determining the number of required signatures). However, in mithril $k << N$ where $N$ is the total number of eligible signers. Mithril verification goes as foollows: 
+* Check the $k$ is large enough
+* $t$ $\mathbb{G}_1$ additions
```
michaelpj, line 158: k?

michaelpj: and below

iquerejeta:
Ok, yes. I will add that as well. It's true that the numbers with respect to ed25519 might not be very helpful.

CIP-0037/CIP-0037.md
```diff
@@ -0,0 +1,183 @@
...
+
+
+### Reason for choosing BLS12-381
+The reason for choosing BLS12-381 over BN256 curve is that the former is claimed to provide 128 bits of security, while the latter was reduced to 100 bits of security after the extended number field sieve (a new algorithm to compute the discrete logarithm) was [shown to reduce the security](https://eprint.iacr.org/2016/1102.pdf) of these curves. 
+
+An [EIP](https://eips.ethereum.org/EIPS/eip-2537) for precompiles of curve BLS12-381 already exists, but has been stagnant for a while. Nonetheless, Zcash, MatterLabs and Consensys support BLS12-381 curve, so it is certainly widely used in the space. 
+
+Further reading regarding curve BLS12-381 can be found [here](https://hackmd.io/@benjaminion/bls12-381) and the references thereof cited. For level of detail required in this document, it is sufficient to understand that one can define a pairing over BLS12-381, where a pairing is a map: $e:G_{1}\times G_{2}\rightarrow G_{T}$, which satisfies the following properties:
```
michaelpj, line 34: Github markdown doesn't support latex maths, so these look pretty bad throughout. Maybe we're okay just asking people to render the latex in their heads, but it might be nice to try and use ascii.

L-as: You can use inline SVG IIRC.

L-as:
> And also, could you elaborate on why you think that Halo 2 (with IPA) is the best choice for a universal proof system (over Plonk, for instance)? I believe that verification of relatively complex circuits with Halo2 will most probably exceed plutus execution budget.

If we are to support Halo 2 directly in the interpreter, it would have a budget defined by the interpreter. It wouldn't be extremely fast, but we are free to choose whatever cost model we want when putting it into the interpreter. In practice, it would probably be half the CPU and memory budget + 1 or something similar.

As for why Halo 2, there isn't really any competitor. The four important properties are recursive proofs, trustlessness, open source license, and quality code base. There isn't any alternative to Halo 2 that accomplishes all of this.

iquerejeta:
> As for why Halo 2, there isn't really any competitor.

Plonk is being used in several applications, so I would consider it as a competitor. 

> The four important properties are recursive proofs, trustlessness, open source license, and quality code base.

Agree, however, plonk trusted setup is universal (meaning that it has to be performed only once) and is updatable which, IMO, it is enough for a lot of scenarios (maybe not all). What I'm concerned of Halo2 is the time it takes to verify a proof. It is my understanding that the plutus execution budget is 5-10ms. I don't have recent benchmarks at hand, but at least what was published in the Halo paper indicates that verifying a circuit of 2^11 constraints (pretty quite limiting) takes 0,1 seconds, and growth is linear. And verifying 2^17 constraints takes close to 1 second. I know these numbers are definitely better using Halo2, but I couldn't find any benchmarks. Maybe you have some numbers you can share to support the statement that it will be around half the CPU budget? For reference, using [plonk](https://github.com/ZK-Garage/plonk), times are 4,13ms and 5,15ms respectively.

michaelpj:
> It wouldn't be extremely fast, but we are free to choose whatever cost model we want when putting it into the interpreter. In practice, it would probably be half the CPU and memory budget + 1 or something similar.

We certainly can't just pick whatever cost model we like. It's important that it's at least somewhat accurate, because ultimately those limits are what stop us from blowing out our block diffusion times. If Halo 2 takes 2ms to verify something, it's going to have to be costed like that.

CIP-0037/CIP-0037.md
```diff
@@ -0,0 +1,217 @@
+###### tags: `CIP`
```
markulf, line 1: Suggestions: 

Also expose non-pairing curves using same API (except for pairing).

L-as:
@iquerejeta I think I agree now that we should just add support for the curve to UPLC now instead of adding support for the verification directly. However, do the Pasta curves also work for your use case?

iquerejeta:
The support that we will be including in this CIP is only for BLS12-381 curve, not the pasta curves. However, other curves could be considered if we see that they provide attractive properties. I'm guessing that the main attractive point of pasta curves is its recursive nature, right? If that is the case, considering plonky2 might be worth it, that combines ZKPs using FRI commitments (for speed in producing the proof), and batches them using a Plonk proof with KZG10 commitments (for succinctness in the proof size). I'm no expert in the field, but I believe this would provide attractive performance when batching several proofs, and would be compatible with the introduction of BLS12-381 curves. I would be more than happy in having a more in-depth discussion on this and the general topic of zkps and roll ups.

L-as:
FYI even the smallest Plonky2 proof is too big to fit in a single transaction AFAICT. In any case, we should definitely talk about this. Feel free to set up something.

CIP-0037/CIP-0037.md
```diff
@@ -0,0 +1,338 @@
...
+attributes does not require further pairing evaluations. 
+* Proof verification to a relation involving elements in G_1 and G_2. This verification does not require the 
+* pairing evaluation, but does require G_1 and G_2 additions and multiplications. 
+* 2 pairing evaluations + equality check
+
+The complexity of the statement one can prove with anonymous credentials will be bounded by plutus computational 
+limits. However, it is hard to make individual functions for each possible statement accepted. 
+
+## Plonk verification
+Plonk verification is quite complex, and might differ depending on the number of custom gates, or specific wire
```
davidnevadoc, line 319: The specific wire connections do not affect the verifier algorithm complexity. The connections are enforced as copy constraits with a permutation argument whose complexity is the same no matter the permutation we are using.

CIP-0037/CIP-0037.md
```diff
@@ -0,0 +1,338 @@
...
+* pairing evaluation, but does require G_1 and G_2 additions and multiplications. 
+* 2 pairing evaluations + equality check
+
+The complexity of the statement one can prove with anonymous credentials will be bounded by plutus computational 
+limits. However, it is hard to make individual functions for each possible statement accepted. 
+
+## Plonk verification
+Plonk verification is quite complex, and might differ depending on the number of custom gates, or specific wire 
+connections. In this section we expose the verifier complexity as described 
+[here](https://hackmd.io/Vr0DUYPCR0iWADpO8kpkAA?view#Verifier-Algorithm). We count every challenge computation
```
davidnevadoc, line 321: Better use a reference to the [paper](https://eprint.iacr.org/2019/953.pdf) as it is more updated. 
In particular the current version has a new way of computing the linearisation polynomial that saves some operations for the verifier.

CIP-0037/CIP-0037.md
```diff
@@ -0,0 +1,338 @@
...
+Plonk verification is quite complex, and might differ depending on the number of custom gates, or specific wire 
+connections. In this section we expose the verifier complexity as described 
+[here](https://hackmd.io/Vr0DUYPCR0iWADpO8kpkAA?view#Verifier-Algorithm). We count every challenge computation 
+as a single hash evaluation. We omit the `scalar X scalar` multiplications and additions. As we can see above, 
+these operations are insignificant with respect to group operations. We assume that point validation is performed 
+by the external C library, and therefore do not count that in this estimate. The computation is the following: 
+
+* 6 hash computations
+* 1 vanishing polynomial (1 scalar exponantiation)
+* Quotient polynomial evaluation (1 scalar inverse)
```
davidnevadoc, line 328: No need to compute the quotient evaluation in the newest version.

CIP-0037/CIP-0037.md
```diff
@@ -0,0 +1,338 @@
...
+* First part of batch poly commitment (7 scalar mults and 6 additions in G_1)
```
davidnevadoc, line 329: Now: 10 scalar mults and 9 additions in G_1

CIP-0037/CIP-0037.md
```diff
@@ -0,0 +1,338 @@
...
+* Fully batched poly commitment (7 scalar mults and 8 additions over G_1)
```
davidnevadoc, line 330: I've counted 9 scalar mults in the previous in the previous version.
Now: 6 scalar mults and 5 additions in G_1

CIP-0037/CIP-0037.md
```diff
@@ -0,0 +1,338 @@
...
+* Group encoded batch verification (1 scalar mult over G_1)
+* Batch-validate all evaluations (3 scalar mults and 4 additions over G_1, plus two pairing evaluations + equality check)
+
+
+Implementations may differ, and adding custom gates or additional wires will affect these estimates. Pairing
```
davidnevadoc, line 335: This is a good estimate as a reference, but we have to expect higher costs because this is the most basic version of Plonk and real life implementations will likely be more complex. 
For example adding lookups gates introduces tables (with its corresponding commitments), an extra permutation argument (increases the number of scalar multiplications and additions) and the need for new challenges (more hashes).

davidnevadoc (changes requested):
I have added some comments in the Plonk section with suggestions to make it a bit more precise.
In general, I think this is a very good and useful proposal! :)

CIP-0037/CIP-0037.md
iquerejeta: I count 9 and 9 in the first part of the poly com, and 5 and 5 in the second part. Thanks for the suggestion of referencing the paper, will do that 👍

CIP-0037/CIP-0037.md
CPerezz: Blinder costs should also be considered as well as the custom gates that any implementation uses.

From my perspective, plonk is "completely agnostic to this PR". In the sense that any project will be free to implement it's own Plonk Verifier contract/lib with their own custom gate design picks for their use-cases.
What I think would be really valuable is to first make sure that this is indeed doable for `vanilla-plonk` or `turbo-plonk` (excluding custom gates) as we still don't have any idea of how much this execution will cost or even if it is possible to execute all that stuff inside a contract.

iquerejeta: Thanks. Indeed, as you say this CIP is agnostic to what the pairings are used for. This list served as an estimate of some of the use cases that could leverage pairings. On your second point, I complete agree as well. Once we get the pairings running on a test version of plutus (and we have the functions costed), we'll see how it performs for verifying a vanilla plonk, and how much of the execution budget we spend.

michaelpj (commented):
I think this is starting to look viable and we should maybe move to a prototype implementation.

CIP-0037/CIP-0037.md
```diff
@@ -0,0 +1,338 @@
...
+```
+* generator G2:
+```
+[147, 224, 43, 96, 82, 113, 159, 96, 125, 172, 211, 160, 136, 39, 79, 101, 89, 107, 208, 208, 153, 32, 182, 
+26, 181, 218, 97, 187, 220, 127, 80, 73, 51, 76, 241, 18, 19, 148, 93, 87, 229, 172, 125, 5, 93, 4, 43, 126, 
+2, 74, 162, 178, 240, 143, 10, 145, 38, 8, 5, 39, 45, 197, 16, 81, 198, 228, 122, 212, 250, 64, 59, 2, 180, 
+81, 11, 100, 122, 227, 209, 119, 11, 172, 3, 38, 168, 5, 187, 239, 212, 128, 86, 200, 193, 33, 189, 184]
+```
+
+#### An important note on GT elements
```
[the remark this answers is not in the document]
iquerejeta: What we had before (of only the equality check) allowed us to compare two pairing evaluations, e.g. `e(A,B) =? e(C,D)` where `e(,)` is the pairing function. However, in some cases, we want to have more complex linear relations of pairing evaluations, e.g. `e(A, B) * e(C, D) =? e(E, F) * e(G, H)`. The benefit of using this type of representation comes here. If we did not use that, we would take around 700-800 us for each pairing, resulting in ~3ms for the whole. Instead, by using this representation we expect around 1,6ms. 

I'll clarify this in the section.

CIP-0037/CIP-0037.md
```diff
@@ -0,0 +1,338 @@
...
+    * `BLS12_381_H2G2 :: ByteArray -> Bls12381G2Element`
+    * `BLS12_381_GT_mul :: Blst12381GTElement -> Blst12381GTElement -> Blst12381GTElement`
+* Pairing operations:
+    * `BLS12_381_pairing :: Bls12381G1Element -> Bls12381G2Element -> Bls12381GTElement`
+
+On top of the elliptic curve operations, we also need to include deserialization functions, and equality definitions
+among the different types. 
+
+* Deserialisation: 
+  * `deserializeG1 :: ByteString -> Either BLSTError (Bls12381G1Element)`
```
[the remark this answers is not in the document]
michaelpj: Ignore me, you can't lambda-encode it, that's a higher-order function. This is actually a bit tricky.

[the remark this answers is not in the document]
iquerejeta: But then, if we want to handle the error, what would be your recommendation?

L-as:
One very important thing to note here, is that we're likely going to add more curves in the future, notably the Pasta (Pallas and Vesta) Curves, secp256k1 (the curve).

The core question I have then is, should we for each curve add the same primitives again separately?
Should basic primitives like addition work for many curves (and even integers)?

CIP-0037/CIP-0037.md
```diff
@@ -0,0 +1,352 @@
...
+* Pairing operations:
+    * `BLS12_381_pairing :: BLS12_381G1Element -> BLS12_381G2Element -> BLS12_381GTElement`
+    * `BLS12_381_final_verify :: BLS12_381GTElement -> BLS12_381GTElement -> bool` This performs the final 
+  exponantiation (see section `An important note on GT elements` below).
+
+On top of the elliptic curve operations, we also need to include deserialization functions, and equality definitions
+among the G1 and G2 types. 
+
+* Deserialisation: 
+  * `deserializeG1 :: ByteString -> (BLSTError -> a) -> (Bls12_381G1Element -> a) -> a`
```
L-as, line 118: Why not just `deserialiseG1 :: ByteString -> BuiltinList Bls12_381G1Element`, where the output is empty if it errs and otherwise a single element? We don't really care about error handling too much on-chain.
You could also just make it partial, `deserialiseG1 :: ByteString -> Bls12_381G1Element` looks reasonable to me.

CIP-0037/CIP-0037.md
```diff
@@ -0,0 +1,352 @@
...
+* `BLS12_381G1Element`
+* `BLS12_381G2Element`
+* `BLS12_381GTElement`
+
+We need to support the binary operation of G1 and G2 (which are additive groups), as well as the binary operation 
+over GT (which is represented as a multiplicative group). We also want to enable hashing to G1 and G2.
+In particular, we expose the hash to curve (which we denote with `H2G1` and `H2G2` for G1 and G2 
+respectively) algorithm as described in 
+[hash to curve draft](https://datatracker.ietf.org/doc/html/draft-irtf-cfrg-hash-to-curve#section-8), 
+using `BLS12381G1_XMD:SHA-256_SSWU_RO_` and `BLS12381G2_XMD:SHA-256_SSWU_RO_` for G1 and G2 respectively.
```
L-as, line 95: Shouldn't we use BLAKE2 as is common in Cardano AFAICT? There are a lot of reasons to avoid SHA2, see https://github.com/zcash/zcash/issues/706.

CIP-0037/CIP-0037.md
```diff
@@ -0,0 +1,338 @@
...
+then `A` can simply move to the other side of the equality sign. These limitations allow us for a performance 
+trick, which is also used for the verification of BLS signatures. In a nutshell, a pairing is divided into two 
+operations: (i) Miller loop, and (ii) final exponentiation. Both are expensive, but what we can do is first 
+compute the miller loop, which already maps two points from G_1 and G_2 to G_T. Then we can use this result 
+to perform the operations over these points (mults). Finally, when we want to check for equality, we invert 
+one of the two points, and multiply the result by the other point. Only now we compute the final exponentiation,
+and we verify that the result is the identity element. In other words, the equality check function, `eq`, first
+computes an inverse, then a multiplication, and finally a final exponentiation.
+                                         
+For this reason it is important to limit what can be done with GTElements, as the pairing really is not the full
```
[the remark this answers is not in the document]
L-as: So essentially, you'd add `BLS12_381_full_pairing`? Shouldn't `BLS12_381_pairing` just have a different name then?

CIP-0037/CIP-0037.md
```diff
@@ -0,0 +1,352 @@
...
+over GT (which is represented as a multiplicative group). We also want to enable hashing to G1 and G2.
+In particular, we expose the hash to curve (which we denote with `H2G1` and `H2G2` for G1 and G2 
+respectively) algorithm as described in 
+[hash to curve draft](https://datatracker.ietf.org/doc/html/draft-irtf-cfrg-hash-to-curve#section-8), 
+using `BLS12381G1_XMD:SHA-256_SSWU_RO_` and `BLS12381G2_XMD:SHA-256_SSWU_RO_` for G1 and G2 respectively.
+We do not include the type of the Scalar, nor operations related to it. These type of operations can already be 
+performed with the Integral type. In particular, we need the following functions: 
+
+* Group operations: 
+    * `BLS12_381_G1_add :: BLS12_381G1Element -> BLS12_381G1Element -> BLS12_381G1Element`
```
L-as, line 100: Idea that @morganthomas told me, we could have each of these generic group operations be indexed with the type of the group essentially, so `add :: (g :: G) -> g -> g -> g` in a dependently typed language, where `g` is internally represented as an integer. In the implementation, the interpreter would then err for invalid choices of `g`.

L-as: This might be better for supporting other curves in the future, what do you think @michaelpj ?

CIP-0037/CIP-0037.md
L-as: I can't think of a single case where handling the error is necessary.

CIP-0037/CIP-0037.md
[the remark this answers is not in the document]
iquerejeta: Yes, right. If there is the possibility that in the future we would add the full pairing, we should name it differently here. Maybe something like `BLS12_381_ppairing_ml` (standing for partial pairing miller loop), and then have clear documentation of the function.

CIP-0037/CIP-0037.md
iquerejeta: I think having a generic strategy is worth considering, though we need the opinion of @michaelpj here. I don't think we will add support for a lot of pairing friendly curves though, as this moves somewhat slowly. But in the mid/long term, this might be the case. 

My understanding is that the motivation for this is to include non-pairing friendly curves (such as pasta curves). My concern is that this generic strategy will become quite complex if we want to support all kind of EC families, and I'm not sure I see a huge benefit over simply adding support for different curves. For instance, do we want the interpreter to be aware of the cycles of the pasta curves (or any cycles of pairing friendly curves)? how would the pairing fit in such a generic framework? Would we have something like `BLS12_381_pairing :: (g1 :: G, g2::G) -> g1 -> g2 -> something`? (sorry if my notation is incorrect, I'm not a haskell developer). How would we represent the target group? 

I think that having a generic strategy that supports pairing friendly and non-pairing friendly curves would be beneficial, as we could also use it to expose curve operations over secp256k1 or curve25519.

CIP-0037/CIP-0037.md
iquerejeta: We are using the recommendation of the standard for BLS12 381, which is also what is implemented in the blst library. BLS signatures depend on this choice, and the goal is to follow the standard to be compatible with as many implementations of BLS signatures over BLS12 381. 

While Blake is the common and default choice in  Cardano, there are some exceptions. These exceptions are made to follow the standard specifications and make Cardano verification procedures compatible with other implementations. Notable exceptions are ed25519 and VRF (where both use SHA512 to compute the challenge).

CIP-0037/CIP-0037.md
michaelpj: Why can't we just fail the whole thing if deserialization fails? Surely this indicates malformed input of some kind?

CIP-0037/CIP-0037.md
michaelpj: I don't think we can do this easily. Firstly, we don't have the dependent typing. Secondly, if we added new curves to the support here, that would mean that some builtin calls which had previously failed would start succeeding, which is a visible semantics change.

L-as: @michaelpj I agree in that it *is* a change to the semantics, but isn't that true for *every* change? A transaction that attempts to use Plutus V2 will fail now, but won't in the future, no?

wrt. dependent typing, that was simply to make it clear what the intent was. You could have the signature be `add :: Int -> G -> G -> G` where `G` is a sum type that represents all the possible choices.

L-as: @michaelpj I agree in that it *is* a change to the semantics, but isn't that true for *every* change? A transaction that attempts to use Plutus V2 will fail now, but won't in the future, no?

wrt. dependent typing, that was simply to make it clear what the intent was. You could have the signature be `add :: Int -> G -> G -> G` where `G` is a sum type that represents all the possible choices.

CIP-0037/CIP-0037.md
L-as: I am very much for just failing the whole thing.

CIP-0037/CIP-0037.md
michaelpj: It's a change to the semantics of *existing scripts*. That we can't do easily. And sadly even failures are visible, because you can post a failed script to the chain if it's a phase 2 failure.

L-as: @michaelpj Well the Plutus version could determine what indices are valid. That would solve the issue AFAICT.

michaelpj: If you were willing to do a new Plutus version every time. But we're expecting those to be very infrequent, we can normally add builtins without needing one.

L-as: Well that sounds fine to me honestly. We don't want that many anyway.
However, can you think of a better solution rather than this? Is it more sensible to duplicate the operations for each curve? Are we duplicating type checking functionality? What is the most efficient? Is having that many builtins (O(n) for n curves) practical? Could the implementation be simpler if they were the same builtin?

iquerejeta: They would certainly not have identical performance characteristics. And as I said above, different curves (or more precisely, groups) will even have different functions. For instance, if we were to include operations over the group defined over secp256k1, we would only need `add`, `mult` and `neg` functions once (together with serialization and equality checks), but there would be no concept of pairing, `G1` or `G2`, and target group. If, on the other hand, we were to include operations over the groups defined over the pasta curves, we would need those operations together with some definition of `cycles` (something not covered in this CIP). I'm still not sure that having a single generic API covering all types of groups over ECs (including non-prime order groups, pairing friendly curves and cycles over different curves) will be simpler, provided that we don't plan on supporting a very big number of curves.

L-as: I agree with this, albeit it's unfortunate.

iquerejeta:
I believe we are in a position to start prototyping. Not sure who should approve the CIP, maybe @KtorZ ? I guess we need to adapt the number of the CIP, but I guess we can leave to the last moment.

CIP-0037/CIP-0037.md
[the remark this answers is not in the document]
L-as: @michaelpj Thoughts on this? Should we try to make a complete API or should we add it in another Plutus language version in the future if someone needs it?

L-as:
You can start working on an implementation before acceptance/rejection FWIW.

CIP-0037/CIP-0037.md
```diff
@@ -0,0 +1,357 @@
...
+    * `BLS12_381_G1_mult :: Integer -> BLS12_381G1Element -> BLS12_381G1Element`
+    * `BLS12_381_G1_neg :: BLS12_381G1Element -> BLS12_381G1Element`
+    * `BLS12_381_H2G1 :: ByteArray -> BLS12_381G1Element`
+    * `BLS12_381_G2_add :: BLS12_381G2Element -> BLS12_381G2Element -> BLS12_381G2Element`
+    * `BLS12_381_G2_mult :: Integer -> BLS12_381G2Element -> BLS12_381G2Element`
+    * `BLS12_381_G2_neg :: BLS12_381G2Element -> BLS12_381G2Element`
+    * `BLS12_381_H2G2 :: ByteArray -> BLS12_381G2Element`
```
L-as, line 107: Shouldn't this be `ByteString` instead of `ByteArray`?

CIP-0037/CIP-0037.md
```diff
@@ -0,0 +1,357 @@
...
+    * `BLS12_381_GT_mul :: Blst12381GTElement -> Blst12381GTElement -> Blst12381GTElement`
+* Pairing operations:
+    * `BLS12_381_ppairing_ml :: BLS12_381G1Element -> BLS12_381G2Element -> BLS12_381GTElement`
+    * `BLS12_381_final_verify :: BLS12_381GTElement -> BLS12_381GTElement -> bool` This performs the final 
+  exponantiation (see section `An important note on GT elements` below).
```
L-as, line 112: typo

L-as:
what happened to PR name?

CIP-0037/CIP-0037.md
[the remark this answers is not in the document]
L-as: bump

KtorZ:
@L-as regarding PR's name: CIP numbers are assigned on a best-effort basis. It sometimes happen that two CIPs are submitted along the same time and authors allocate the same numbers. That's why encourage authors not to assign numbers themselves, we do it as part of the housekeeping. Plus, a number is only truly settled once the CIP is accepted and merged. Hence the `?` on candidates CIPs. This is a summarized on the main README.

CIP-0037/CIP-0037.md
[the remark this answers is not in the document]
michaelpj: It's hard to say in advance. I think the efficiency arguments are compelling, but it would indeed be good to identify clearly what the functions that we are adding do, e.g. naming and documenting them appropriately. That will also reduce confusion if we do add more functions later.

iquerejeta:
@KtorZ , do you think the PR is ready to be merged? We are going to start doing some prototyping, so IMO we could merge this as a draft, WDYT?

KtorZ:
@iquerejeta we can add it to the agenda for next session (cc @mangelsjover) for review. 

I am yet to go through the comments and discussions that happened as part of this one, but I can already make a few points that would need to be fixed before we merge it:

- Change the name & folder to CIP-0381
- Rename _Reason for choosing BLS12_381_  as _Rationale_ and move it after the _Specification_ section. 
- Rename _Moving to Proposed_ as _Path to Proposed_ and consider perhaps having a _Path to Active_ as well (though for this one, I don't see much beyond seeing this actually implemented in Plutus and released in an upcoming update). 
- I am not quite sure about the purpose of the _Some operation count_ section? It seems to sits in-between the realm of recommendations for some of considered use-cases and a justification of the use of pairing in those situations; so perhaps this last section ought to be nested within the `Rationale` section?

iquerejeta:
I've included your suggestions. What I'm not entirely convinced is about the `Some operation count`. I've removed that, and added the operation count in the motivation (beginning of the document). Do you think it fits well there? Happy to make further changes, and thanks for looking into it 👍

KtorZ (approved):

rphair (approved):
author has completed changes requested in https://github.com/cardano-foundation/CIPs/pull/220#issuecomment-1174808014.

dcoutts:
In addition to security audits on the proposed implementation, it is important that we review the inclusion of the library for practical issues of build systems, and long term maintainability.

In particular in this case the library will be used in the chain format and in chain verification. This means we have special constraints that the format and verification logic must _never change_. Or more specifically: it must be possible _forever_ to be able to verify the existing uses on the chain. So even if there are upgrades or format changes in future, it must still be possible to decode and verify the old uses. This is not a constraint that most software has, and so many libraries are not designed to work within that constraint.

The proposed implementation is https://github.com/supranational/blst

* The implementation is in C and assembly for x86 and ARM v8. This is good for the existing Haskell implementation of the Cardano node because integrating with C libraries is relatively straightforward, it's well supported by the build and CI system and does not pull in too many extra dependencies. (Contrast for example if it were a Rust library where we would have serious practical problems on this front.)
* The implementation appears to have been designed with blockchain implementations in mind. This is a good sign for the long term maintainability because it probably means that the library authors will continue to support the existing format and semantics even if there are changes or improvements.
* The implementation is claimed to be compliant with draft IETF specifications. There is a risk that the specs may change before being declared final, and the library may be updated to follow. There is a risk that the Cardano community will have to support the old version forever. Though given the point above, it's probably the case that library updates would still provide compatibility with the current IETF drafts and serialisation formats.

So overall this seems like something the core development team and Cardano community could support long term without too high a cost. Though we should be aware of the risk that we may have to support an old version of the library, if the library gets changed in incompatible ways.

To ensure no compatibility surprises, it is worth considering forking the repository at a specific commit/version and building the node using that version. This is to guarantee the version remains available. Then for any future updates, the fork repo could be updated to a new version explicitly, with appropriate compatibility checks to ensure the existing on-chain uses are still compatible.

iquerejeta:
> To ensure no compatibility surprises, it is worth considering forking the repository at a specific commit/version and building the node using that version.

I'm not sure I see what is the benefit of forking here. This would only be useful if it happened that the repo maintainers overwrote history, right? As long as that does not happen, we are good sticking to a particular version/commit of the original repo without the need to fork it.

PR #222, opened 2022-02-15
CIP-0043 | New Plutus Core built-in `dataHash`

L-as:
Replaces #218 

[Rendered](https://github.com/mlabs-haskell/CIPs/blob/dataHash/datahashlasmlabscity.md)

### DIFF
--- a/datahashlasmlabscity.md
+++ b/datahashlasmlabscity.md
@@ -1,0 +1,101 @@
+---
+CIP: ?
+Title: New Plutus built-in `dataHash`
+Authors: Las Safin <las@mlabs.city>
+Discussions-To:
+Comments-URI:
+Status: Draft
+Type: Standards Track
+Created: 2022-02-15
+* License: Apache-2.0
+* Requires: 35
+* Replaces: 36
+---
+
+# New Plutus built-in `dataHash`
+
+## Abstract
+
+We propose adding a new built-in `dataHash` for calculating the hash of `Data` on-chain,
+in such a way that it is coherent with `txInfoData`.
+
+## Motivation
+
+This is useful for essentially "putting" `Data` into token names, the contents of which a script
+can access by having the concrete `Data` be passed in the redeemer (whether it's a minting policy
+or validator).
+
+It is also possible to verify signatures of (hashes of) `Data` with this built-in.
+
+This likely has many other uses, as it's a powerful feature.
+
+Currently it is essentially infeasible to do this without the built-in, as it's
+very expensive to first serialise the `Data` into CBOR then hashing it using `blake2b_256`.
+
+## Specification
+
+### Function definition
+
+The new built-in is:
+
+```haskell
+dataHash :: data -> bytestring
+```
+
+The reason for the name is that an off-chain `dataHash` already exists.
+
+The semantics should be such that for any key-value pair `(k, v)` in `txInfoData`,
+it should be true that `k == dataHash v`.
+
+### Cost model
+
+The CPU cost should be linear in the size metric of the input.
+
+The memory cost of `blake2b_256` is constant, but the memory cost
+of serialisation might not be constant even if it's streamed.
+
+It's possible that the memory cost will be linear with the depth of
+the data, in which case it might make sense to make the memory
+cost logarithmic in the size metric.
+
+## Rationale
+
+A possible alternative is CIP-36, which proposes a function `serialiseData`, such that
+`blake2b_256 . serialiseData` is equal to `dataHash`.
+
+The reason `dataHash` is considered to be a better approach by the author,
+is that in practice, all sensible uses of `serialiseData` are also possible
+using `dataHash`.
+
+Hashing using a different algorithm isn't something you need in practice,
+because no existing protocol hashes Plutus's `Data` with something other
+than `blake2b_256`. If in a future version of Plutus the hashing algorithm
+were to be changed, then `dataHash` would also be changed accordingly.
+
+With this approach, you can not inspect the raw serialised `Data`,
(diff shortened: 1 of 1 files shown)


datahashlasmlabscity.md
```diff
@@ -0,0 +1,101 @@
...
+We propose adding a new built-in `dataHash` for calculating the hash of `Data` on-chain,
+in such a way that it is coherent with `txInfoData`.
+
+## Motivation
+
+This is useful for essentially "putting" `Data` into token names, the contents of which a script
+can access by having the concrete `Data` be passed in the redeemer (whether it's a minting policy
+or validator).
+
+It is also possible to verify signatures of (hashes of) `Data` with this built-in.
```
ch1bo, line 28: This would require having the same implementation of `dataHash` off-chain to produce the signature or a sufficiently detailed specification on how the `Data` is serialized (and hashed) such that the system creating the signature can follow the spec.

datahashlasmlabscity.md
```diff
@@ -0,0 +1,101 @@
...
+`blake2b_256 . serialiseData` is equal to `dataHash`.
+
+The reason `dataHash` is considered to be a better approach by the author,
+is that in practice, all sensible uses of `serialiseData` are also possible
+using `dataHash`.
+
+Hashing using a different algorithm isn't something you need in practice,
+because no existing protocol hashes Plutus's `Data` with something other
+than `blake2b_256`. If in a future version of Plutus the hashing algorithm
+were to be changed, then `dataHash` would also be changed accordingly.
```
ch1bo, line 73: While this may be true for the `Data` provided by the ledger, it is not necessarily true for "user-supplied" `Data`. Furthermore, using a different hashing algorithm than `blake2b_256` is not unrealistic. For example, in Hydra we are using `sha2_256` right now! Although we could pick the used hash algorithm ourselves in our use case, other situations where some specific hashing scheme might be dictated by other, off-chain systems are perfectly thinkable and would not be able to use this builtin.

datahashlasmlabscity.md
```diff
@@ -0,0 +1,101 @@
...
+
+With this approach, you can not inspect the raw serialised `Data`,
+however, this is not something that is useful in practice, as the
+serialised `Data` doesn't give you any useful information other
+than its hash.
+
+Another important point to note is that `dataHash` is potentially more
+efficient on large `Data`, since the hash can be computed without storing
+the entire serialisation of the `Data` in memory.
```
ch1bo, line 82: This is something which should be detailed in the "Path to active" section as the most straight-forward `blake2b_256 . serialiseData` implementation (and thus easiest to develop and maintain) would not have that property. Only a hand-crafted serialize + hash routine might have these resource benefits.

datahashlasmlabscity.md
```diff
@@ -0,0 +1,101 @@
...
+
+Implementation-wise, it should also be simpler than `serialiseData`, since
+there are no considerations wrt. what the format of the serialisation should be,
+because it's never exposed to users.
```
ch1bo, line 86: This is unfortunately not the case and tying the semantics of how the data is serialized and then hashed within this primitive to "whatever is available in off-chain `dataHash`" is not sufficient for non-Haskell use cases and potential source of regressions. As commented above, a specification of the actual binary representation for `Data` is required.

ch1bo (commented):
As one of the authors of the proposed CIP-36 (#218) which this would want to replace, I left some comments on where we (Hydra) would additional specification analogous to our PR.

We were considering this approach as well (or even a much more specific to Hydra variant), but we felt that `sha3_256 . serialiseBuiltinData` (or a different hash algorithm of choice) is straight-forward to write, does not add any additional cost, is more flexible and does not increase the scope of CIP-36 unnecessarily.

datahashlasmlabscity.md
```diff
@@ -0,0 +1,101 @@
...
+* Requires: 35
+* Replaces: 36
+---
+
+# New Plutus built-in `dataHash`
+
+## Abstract
+
+We propose adding a new built-in `dataHash` for calculating the hash of `Data` on-chain,
+in such a way that it is coherent with `txInfoData`.
```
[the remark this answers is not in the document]
L-as: Can you explain this? How is this not sufficient?

datahashlasmlabscity.md
L-as: I am not sure what you mean. You are saying that users could want to choose a different hashing algorithm, but without specifying a reason as to why. Can you give a realistic use case that would be prevented by this?

datahashlasmlabscity.md
L-as: I fundamentally disagree with this. This is a core property of `dataHash`. According to CIP-35, this is where it ought to be. You can make bad implementations of anything.

datahashlasmlabscity.md
L-as: ...non-Haskell use cases? If you want to replicate this hashing algorithm in e.g. Rust, you just use whatever hashing algorithm Plutus already uses. How do you think https://github.com/input-output-hk/jormungandr would be implemented?

L-as:
Do note https://github.com/cardano-foundation/CIPs/pull/218#discussion_r803682900

KtorZ:
The Plutus team decided to adopt #218 as a new built-in, making this proposal now redundant.
 
For the sake of keeping a track record, we may still want to merge this proposal as `rejected`.

michaelpj:
The primary reason for adopting https://github.com/cardano-foundation/CIPs/pull/218 over this was the argument from increased composability.

L-as:
What should I change concretely? status to rejected?

PR #51, opened 2020-12-10
fixing deadlink for BIP32-ed25519 paper

crptmppt:

PR #225, opened 2022-02-19
CIP-0030 | Added `walletExperimentalApiVersion`

vsubhuman:
There must be two separate version parameters:
1. The version of the supported spec API
2. The version of the wallet API implementation itself

Example usage:

- A dapp knows a specific wallet supports a new experimental feature that they want to utilize, they check `.specificWallet.walletExperimentalApiVersion` and compare it to the version in the wallet documentation, since when the feature has appeared, because even if the user has the `.specificWallet` installed - they might have an older version, before the feature were added.

### DIFF
--- a/CIP-0030/README.md
+++ b/CIP-0030/README.md
@@ -26,6 +26,8 @@ In order to facilitate future dApp development, we will need a way for dApps to 
 
 The API specified in this document will count as version 0.1.0 for version-checking purposes below.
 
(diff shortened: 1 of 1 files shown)


SebastienGllmt:
The way that ccvault tackled this is by adding their app's version number into the walletname namespace

KtorZ:
@vsubhuman is this anything you're willing to pursue? 

Or is the solution described by @SebastienGllmt something you deem acceptable and thus, precludes the CIP?

rphair:
Agreed to close at CIP meeting #51 just ending now.

PR #227, opened 2022-02-24
Add more details on main's README

KtorZ:
(And thus consequently, on the generated website front-page.)

There seems to be a lot of confusion in the community about what CIPs are and what stages they go through. So I hope that these few lines will explain a bit better the situation and gives a clearer view of the bigger picture.


rphair (approved):
I think this will help people get involved with the process more easily, whether coming with initial questions or with submissions, and also will help CIPs be presented & discussed with better quality in outside places like the Cardano Forum.

There are a couple of minor grammar mistakes but they don't affect the readability or understandability of the document: so for working documentation I think the language is OK.

SebastienGllmt (approved):

PR #130, opened 2021-09-16
CIP-0013 update

tomislavhoracek:
This PR introduces a **_CIP-0013_** update where "Handling Daedalus specific links" section is added as well as "Transaction balancing and signing URI queries" draft section.

PR #234, opened 2022-03-18
CIP-0047? | Proposal for open Daedalus or desktop wallet via URL

danielmain:
This PR introduces CIP-0035 where the standard is defined of how websites or desktop applications can open and interact with the installed desktop wallet like Daedalus using a URL.

This is a replacement of the previously [draft PR](https://github.com/cardano-foundation/CIPs/pull/130) created on 16 Sep 2021 by @tomislavhoracek

### DIFF
--- a/CIP-0035/CIP-0035.md
+++ b/CIP-0035/CIP-0035.md
@@ -1,0 +1,1 @@
+Moved to [CIP-0035/README.md](./README.md).
--- a/CIP-0035/README.md
+++ b/CIP-0035/README.md
@@ -48,5 +48,5 @@ Wrong examples:
 Since Daedalus is shiped for a single network separately (mainnet and testnet) and can be installed alongside a fligt version (mainnet), the url needs to specify what version is being refered.
 
 # ⚠️ Open questions
-1. Most of the DEX relies on timing due to price volatility. If the user is only using a secure full-node wallet and when clicking on a URL handler that opens a wallet, which is not yet synchronized is the wallet, how is a timeout handled?
-2. Similar to 1. what if the starting of the wallet takes 2-10 mins? Is that still acceptable for a DEX? Should be defined timeout parameter to the URL so that the user DOES NOT make a transaction if the dApp runs in a timeout?
+1. Most of the DEX relies on timing due to price volatility. If the user is only using a secure full-node wallet, which is not yet fully synchronized, how is a timeout handled?
+2. Similar to 1. what if the starting of the wallet takes 2-10 mins? Is that still acceptable for a DEX? Should a timeout parameter be defined in the URL so that the user DOES NOT make the transaction if the dApp runs in a timeout?
--- a/CIP-0035/README.md
+++ b/CIP-0035/README.md
@@ -34,5 +34,5 @@ Correct examples:
 
 ## Considerations
 
-- Since Daedalus is shiped for a single network separately (mainnet and testnet) and can be installed alongside a fligt version (mainnet), the url needs to specify what version is being refered.
-- Adding the unix-timestamp-timeout parameter, in other words: the time in (Unix format) in the future where the transaction is no longer valid. This will allow using a desktop wallet that might take minutes to start or not be yet synchronized will avoid the user to send unnecessarily a transaction.
+- Since Daedalus is shipped for a single network separately (mainnet and testnet), and can be installed alongside a flight version (mainnet), the URL needs to specify which version is being referred.
+- Adding the unix-timestamp-timeout parameter, in other words: the time in (Unix format) in the future where the transaction is no longer valid. This will allow using a desktop wallet that might take minutes to start, or not yet be synchronized, will prevent the user from unnecessarily sending a transaction.
--- a/CIP-0035/README.md
+++ b/CIP-0035/README.md
@@ -23,30 +23,16 @@ The usage of a different URL will solve the problem of users, which have multipl
 
 # 📖 Specification
 
-The core implementation should follow the [BIP-21](https://github.com/bitcoin/bips/blob/master/bip-0021.mediawiki) standard (with `bitcoin:` replaced with `desktop+cardano:`)
+The core implementation should follow the [BIP-21](https://github.com/bitcoin/bips/blob/master/bip-0021.mediawiki) standard (with `bitcoin:` replaced with `web+cardano:`)
 
 Correct examples:
 ```
-<a href="desktop+cardano-mainnet://Ae2tdPwUPEZ76BjmWDTS7poTekAvNqBjgfthF92pSLSDVpRVnLP7meaFhVd">Buy/Send/Donate</a>
-<a href="desktop+cardano-testnet://Ae2tdPwUPEZ76BjmWDTS7poTekAvNqBjgfthF92pSLSDVpRVnLP7meaFhVd">Buy/Send/Donate</a>
-<a href="desktop+cardano-mainnet-flight://Ae2tdPwUPEZ76BjmWDTS7poTekAvNqBjgfthF92pSLSDVpRVnLP7meaFhVd">Buy/Send/Donate</a>
-
-<a href="desktop+cardano-mainnet://stake?c94e6fe1123bf111b77b57994bcd836af8ba2b3aa72cfcefbec2d3d4">Stake with us</a>
-<a href="desktop+cardano-mainnet://stake?POOL1=3.14159&POOL2=2.71828">Split between our 2 related pools</a>
-<a href="desktop+cardano-mainnet://stake?COSD">Choose our least saturated pool</a>
-```
-
-Wrong examples:
-```
-<a href="desktop+cardano://Ae2tdPwUPEZ76BjmWDTS7poTekAvNqBjgfthF92pSLSDVpRVnLP7meaFhVd?network=testnet">Buy/Send/Donate</a>
-<a href="desktop+cardano-testnet://Ae2tdPwUPEZ76BjmWDTS7poTekAvNqBjgfthF92pSLSDVpRVnLP7meaFhVd?network=mainnet">Buy/Send/Donate</a>
-<a href="desktop+cardano-mainnet://stake?POOL1=3.14159&POOL2=2.71828&network=testnet">Split between our 2 related pools</a>
+<a href="web+cardano://Ae2tdPwUPEZ76BjmWDTS7poTekAvNqBjgfthF92pSLSDVpRVnLP7meaFhVd?unix-timestamp-timeout=1647896861">Buy/Send/Donate</a>
+<a href="web+cardano-testnet://Ae2tdPwUPEZ76BjmWDTS7poTekAvNqBjgfthF92pSLSDVpRVnLP7meaFhVd?unix-timestamp-timeout=1647896861">Buy/Send/Donate (Testnet)</a>
+<a href="web+cardano-flight://Ae2tdPwUPEZ76BjmWDTS7poTekAvNqBjgfthF92pSLSDVpRVnLP7meaFhVd?unix-timestamp-timeout=1647896861">Buy/Send/Donate (Mainnet and using Daedalus Flight)</a>
 ```
 
 ## Considerations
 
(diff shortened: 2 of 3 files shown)


rphair:
thanks for following the suggestion to move this into a new CIP 😎 ... also let me remind other editors & reviewers there is useful discussion about this proposal in the draft PR linked above: https://github.com/cardano-foundation/CIPs/pull/130

@danielmain - is this already being used in work you're doing with the Daedalus group?  If so the status could be changed to `Active` (if not, we can change it the future).

CIP-0035/README.md
```diff
@@ -0,0 +1,52 @@
...
+An application could ask for a payment and offer options like PayPal, Visa, cardano-web. Although a user might want to help cardano to get more decentralized using a full-node-wallet that is why the application could also offer payments with a cardano-desktop.
+The usage of a different URL will solve the problem of users, which have multiple wallets: browser-wallet for daily usage and a secure desktop wallet for the savings.
+
+# 📖 Specification
+
+The core implementation should follow the [BIP-21](https://github.com/bitcoin/bips/blob/master/bip-0021.mediawiki) standard (with `bitcoin:` replaced with `desktop+cardano:`)
+
+Correct examples:
+```
+<a href="desktop+cardano-mainnet://Ae2tdPwUPEZ76BjmWDTS7poTekAvNqBjgfthF92pSLSDVpRVnLP7meaFhVd">Buy/Send/Donate</a>
```
rphair, line 30: Why is the URI scheme prefix changed from `web+cardano` in https://github.com/cardano-foundation/CIPs/pull/130 to `desktop+cardano` in this PR?  CIP-0013 itself specifies `web+cardano` which as far as I know dates back to https://github.com/Emurgo/yoroi-frontend/pull/511.

`The web+` prefix in turn was chosen because simpler scheme literals proposed earlier like `cardano` and `ada` (https://github.com/input-output-hk/daedalus/issues/883) would have [required safelisting with IANA](https://en.wikipedia.org/wiki/Uniform_Resource_Identifier#Syntax) while anything beginning with `web+` could be freely used on the fly (i.e. by apps registering `web+cardano` as a handler).  It doesn't look like `desktop+` has this same property: https://html.spec.whatwg.org/multipage/system-state.html#custom-handlers

rphair (commented):
Github requires me to leave a comment here 🤓

CIP-0035/README.md
danielmain: Thank you @rphair for the review!

We need a distinction because there will be cases of users, which will have multiple wallets installed, secure ones like Daedalus (full-node) for savings and an unsecure one (any light-wallet) for fast transactions.

What would happen by clicking on a `web+cardano`? Which wallet will be opened desktop or web is the dilema 🤷‍♂? 

> It doesn't look like desktop+ has this same property: [html.spec.whatwg.org/multipage/system-state.html#custom-handlers](https://html.spec.whatwg.org/multipage/system-state.html#custom-handlers)

I see the point that it needs to start with `web+` but how to distinguish them between desktop and light-wallets?

`web+desktop+cardano` ?

Also why other apps like visual studio code can be opened with `vscode://` 🤔
https://code.visualstudio.com/docs/editor/command-line#_opening-vs-code-with-urls

danielmain:
> thanks for following the suggestion to move this into a new CIP 😎 ... also let me remind other editors & reviewers there is useful discussion about this proposal in the draft PR linked above: #130
> 
> @danielmain - is this already being used in work you're doing with the Daedalus group? If so the status could be changed to `Active` (if not, we can change it the future).

Yes we are working on it 
https://github.com/input-output-hk/daedalus/pull/2606

CIP-0035/README.md
SebastienGllmt: @danielmain the reason they can use `vscode://` is because the restriction for `web+` does not apply to desktop apps.

> how to distinguish them between desktop and light-wallets

I don't see how this requirement matters. It's entirely possible somebody publishes a desktop application that works as a light wallet and it's entirely possible that somebody has multiple fullnode wallets installed in the future if somebody ever creates another fullnode wallet for Cardano. Prompting the user about which app they want to use for a URI scheme is something handled at the OS level for this reason. For example, if you have multiple mail clients installed, clicking on foo@foo.com will prompt you asking you which mail client you want to use for the URL.

danielmain: > @danielmain the reason they can use vscode:// is because the restriction for web+ does not apply to desktop apps.

Well, Daedalus is, in fact, a desktop app 😅 that is why I did not understand the reason to use `web+`

Anyway, I could live with the idea to use only one protocol for any wallet (not so friendly for users using two wallets and switching between light/desktop in the OS settings). Just we would need to point to mainnet/flight/testnet and add an extra parameter with the unix-timestamp plus a timeout.

I will update this PR.

thank you for your feedback!

CIP-0035/README.md
```diff
@@ -0,0 +1,38 @@
...
+License: CC-BY-4.0
+---
+
+# Abstract
+
+This proposal is an extension of CIP-0013, which describes a basic URI scheme to handle ada transfers or partial transactions with Daedalus or any desktop Wallet installed in the operating system.
+
+# 💡 Motivation
+
+DEX, Web-dApps and Desktop-dApps (e.g. games or desktop applications) should have the possibility to interact directly with desktop wallets. By defining a URI scheme and registering it in the operating system, it will allow opening Daedalus or any desktop wallet same as when you call or click a mailto:email@sample.com and your email client is opened.
```
rphair, line 19: changing this to `mailto:email@sample.com` (i.e, adding backticks around it) should prevent the email address from being formatted differently (as a clickable & invalid email address) in the formatted Markdown (as it appears now).

CIP-0035/README.md
```diff
@@ -0,0 +1,38 @@
+---
+CIP: 35
+Title: Daedalus URI Scheme
+Authors: Daniel Main <daniel.main@iohk.io>
+Comments-URI:
+- https://github.com/cardano-foundation/CIPs/pull/130
```
rphair, line 6: great that you have the last PR discussion in there, but can you also add this one?
`- https://github.com/cardano-foundation/CIPs/pull/234`

rphair (commented):
Thanks for your continued work on this 😎

CIP-0035/README.md
```diff
@@ -0,0 +1,38 @@
...
+
+# 📖 Specification
+
+The core implementation should follow the [BIP-21](https://github.com/bitcoin/bips/blob/master/bip-0021.mediawiki) standard (with `bitcoin:` replaced with `web+cardano:`)
+
+Correct examples:
+```
+<a href="web+cardano://Ae2tdPwUPEZ76BjmWDTS7poTekAvNqBjgfthF92pSLSDVpRVnLP7meaFhVd?unix-timestamp-timeout=1647896861">Buy/Send/Donate</a>
+<a href="web+cardano-testnet://Ae2tdPwUPEZ76BjmWDTS7poTekAvNqBjgfthF92pSLSDVpRVnLP7meaFhVd?unix-timestamp-timeout=1647896861">Buy/Send/Donate (Testnet)</a>
+<a href="web+cardano-flight://Ae2tdPwUPEZ76BjmWDTS7poTekAvNqBjgfthF92pSLSDVpRVnLP7meaFhVd?unix-timestamp-timeout=1647896861">Buy/Send/Donate (Mainnet and using Daedalus Flight)</a>
```
SebastienGllmt, line 32: what's the point of `web-cardano-flight`? This seems very Daedalus-specific. Why not just use `web+cardano` and leave it up to the OS prompt to choose which app to launch (Daedalus or Daedalus Flight)?

CIP-0035/README.md
```diff
@@ -0,0 +1,38 @@
...
+The usage of a different URL will solve the problem of users, which have multiple wallets: browser-wallet for daily usage and a secure desktop wallet for the savings.
+
+# 📖 Specification
+
+The core implementation should follow the [BIP-21](https://github.com/bitcoin/bips/blob/master/bip-0021.mediawiki) standard (with `bitcoin:` replaced with `web+cardano:`)
+
+Correct examples:
+```
+<a href="web+cardano://Ae2tdPwUPEZ76BjmWDTS7poTekAvNqBjgfthF92pSLSDVpRVnLP7meaFhVd?unix-timestamp-timeout=1647896861">Buy/Send/Donate</a>
+<a href="web+cardano-testnet://Ae2tdPwUPEZ76BjmWDTS7poTekAvNqBjgfthF92pSLSDVpRVnLP7meaFhVd?unix-timestamp-timeout=1647896861">Buy/Send/Donate (Testnet)</a>
```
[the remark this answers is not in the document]
danielmain: > I would strongly advise against doing `web+cardano-testnet`. Cardano, just like Ethereum, it sure to have many different networks in the future and we don't want to be stuck with a bunch of different URI schemes for every network. Instead, I suggest you include something similar to [CIP34](https://github.com/cardano-foundation/CIPs/tree/master/CIP-0034?rgh-link-date=2022-03-24T15%3A06%3A06Z) inside the payload

@SebastienGllmt : the network cannot be "selected" in Daedalus. For each Network exist a different Daedalus installation. This is how it is implemented, and I am afraid this will not be changed.

Coming back to the vscode example. 

`vscode://` => opens Visual Studio
 `vscode-insiders://` => opens Visual Studio Insiders

grafik

So, Flight is for us, like the Insider version.
cc: @nikolaglumac

michalrus:
> Cardano, just like Ethereum, it sure to have many different networks in the future and we don't want to be stuck with a bunch of different URI schemes for every network. Instead, I suggest you include something similar to [CIP34](https://github.com/cardano-foundation/CIPs/tree/master/CIP-0034) inside the payload

At least in Daedalus’ case, each network is handled by a different build.

If we are to define only a single `protocol:`, by which the OS differentiates which app to choose, we would have to either:
* let the user choose, after clicking *a* link, in a system prompt, which Daedalus build to use to handle it, and
    * how easy it would be to mis-click is not for me to decide,
    * I think some OS-es don’t show the full list of possible apps, just a few recent ones?
* or create a completely separate small app that opens an appropriate Daedalus build (if it is installed) based on `protocol:THIS_PART` which the OS is not concerned with – in effect re-implementing a great deal of what already is in all 3 OS-es, and having to make it cross-platform, having to detect *where* each Daedalus is installed etc. etc. – a mess.

CIP-0035/README.md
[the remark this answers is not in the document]
michalrus: > Cardano, just like Ethereum, it sure to have many different networks in the future and we don't want to be stuck with a bunch of different URI schemes for every network. Instead, I suggest you include something similar to [CIP34](https://github.com/cardano-foundation/CIPs/tree/master/CIP-0034) inside the payload

At least in Daedalus’ case, each network is handled by a different build.

If we are to define only a single `protocol:`, by which the OS differentiates which app to choose, we would have to either:
* let the user choose, after clicking *a* link, in a system prompt, which Daedalus build to use to handle it, and
    * how easy it would be to mis-click is not for me to decide,
    * I think some OS-es don’t show the full list of possible apps, just a few recent ones?
* or create a completely separate small app that opens an appropriate Daedalus build (if it is installed) based on `protocol:THIS_PART` which the OS is not concerned with – in effect re-implementing a great deal of what already is in all 3 OS-es, and having to make it cross-platform, having to detect *where* each Daedalus is installed etc. etc. – a mess.

CIP-0035/CIP-0035.md
```diff
@@ -0,0 +1 @@
+Moved to [CIP-0035/README.md](./README.md).
```
KtorZ, line 1: This is not needed for new CIPs.

KtorZ (commented):
As currently written and framed, the proposal is too Daedalus-specific to pretend to be included as a CIP. 

If the goal is to define a URI scheme for desktop wallets in general, then it ought to be framed as such for more clarity :+1: , only taking Daedalus as one of the possible example of implementation.

KtorZ:
Closing for now, as it's been inactive for a while. Feel free to re-open if there's any update on this topic.

PR #41, opened 2020-11-11
Revert "reworked file and dir structure for future numbering, added notes for…"

crptmppt:
Reverts cardano-foundation/CIPs#39 so we can merge CIP-10 in repo beforehand, 
and to fix folder conversion mismatch 
 CIP10/CIP10.md -> CIP-0010/CIP-0010.md

PR #239, opened 2022-03-21
 Adding meeting notes Editors Meeting #41

mangelsjover:


rphair (approved):
A lot of material here, but this is pretty much as I remember it 😎 ... there are occasional sound-alikes in the transcription process (one was "pool request" appearing in place of "pull request") but readers are likely to know these minutes are automated transcriptions and homonyms are to be expected.

@KtorZ in an earlier meeting you'd mentioned another PR was edited against an old baseline (or something like that) and ended up with lots of previous commits outside the scope of recent editing.  I think that may have happened again here... but I'm not good enough with Github to understand how you said you fixed it or to suggest what to do in this case.

rphair (approved):
This 2nd review request is redundant (I'm using mouse in the opposite hand & clicked the wrong thing 🤪 ... no way to undo)

KtorZ (approved):

PR #42, opened 2020-11-11
added 11-03-2020 meeting notes, changed dir structure and naming

crptmppt:
PR attempts to adjusts dir structure, naming, and adds meeting notes for 11-03-2020

PR #246, opened 2022-04-11
 Adding meeting notes Editors Meeting #42

mangelsjover:


rphair (approved):
Minutes look OK formatted to me (I was not present); I'll leave it to those present @SebastienGllmt @KtorZ to confirm content.

I still think the computer these minutes are being prepared on needs to sync its git repo so it doesn't keep racking up the cumulative commits. As I remember this can be fixed by doing a `git pull`, then `git merge` with any changes in a working branch, in the local repo; maybe Matthias or Sebastien can confirm the specifics and/or we can talk about this at the planned "retrospective" meeting on 19 April. 😎

KtorZ:
This PR re-add (:thinking:  ?) the notes from March 1st and March 15th, which is weird since they're already on the repository. It could be that it's made from an old state / branch and the github UI gets confused by it. 

Perhaps we could spend some time in a short call @mangelsjover to see what's going on with your git client?

PR #249, opened 2022-04-21
CIP-0048? | NFT metadata references and payloads

Jack-0:
Introduces methods to store and reference 721 metadata. For details and questions please contact me here or on the Cardano Forums

---

[see rendered Markdown](https://github.com/ada9000/CIPs/blob/master/CIP-0048/README.md)


Jack-0:
Discussion and example use cases can be found on the Cardano forum.
https://forum.cardano.org/t/cip-extended-721

Jack-0:
I would mark this as an informational CIP as it recommends a practice within the current 721 metadata which users are free to ignore or implement as willing.

Jack-0:
Major points mentioned on the forum post:

**A blockchain like Cardano is not for storing jpegs or to archive data files:**
This is an opinion. I agree congesting the chain is a bad idea. However, the intention is not for jpegs. Most ‘sane’ developers will understand IPFS or similar decentralized filestores are better in most scenarios. 

**How do you distinguish between two or more very similar 721s metadata:**
Last payload minted is treated as the most relevant. (You can query txhash data for slot etc)

**Is it more flexible or more generic than existing standards for NFT meta data:**
Yes. Because we give the user multiple ways to reference other transactions or onchain data (it doesn’t have to be in the same policy either).

**Could type be implicitly inferred from value:**
Yes it could, but I think using the shorthand ‘p’:<policy_id> is cleaner and is a mere 4-5 bytes extra (‘p’:,). As for mime type such as test/plain that could be done by the client but providing a standard helps avoid confusion.

**Is it backwards compatible:**
Backwards compatibility due to the version 2.0 tag this just expands upon EIP721 (which is explain in cip25)

**More compact than CBOR:**
This could be argued. The choice to use json is as such because it’s popular for web dev and extend the current 721 standard. Using some other format could save a few bytes but at the cost of usability (json parsing is easy and commonplace on the web).

**Future compatibility:**
The standard could also easily be updated in the future by simply altering the version tag.

**Personal note:**
Lot of the criticism seems to be narrow. I believe most people see the word NFT and assume this is meant for profile picture type NFT’s. Although this can create some interesting use cases for those projects most will stick to using 721 (version 1) with IPFS. The intended use is for innovative ideas for nfts (mentioned in forum post). I believe references in combination with IPFS could create some powerful tooling for developers. 

**How this CIP can also be used to reduce chain data:**
https://forum.cardano.org/t/cip-extended-721/99646/10?u=jack7e

**Opensource query example**
https://github.com/pixel-pool/extended721

Jack-0:
I also was just glancing over the 'Smart NFTs' proposal https://github.com/cardano-foundation/CIPs/pull/263.
I believe my Extended metadata CIP solves the issues raised here without coupling itself to a specific language like JS.

CIP-0048/CIP-0048.md
```diff
@@ -0,0 +1,141 @@
...
+
+*inside the refences tag*
+|tags that are reseved inside references |description | example
+|---|---| --- |
+|data| contains the data type MIME | for example if base 64, data:"text/plain;base64"|
+|src|contains a list of poitners to payloads | ```"src":[0,1]``` or ``` "src":["payload0","payload1"]```|
+|policy, p | references another policy. If none, assume current policy | ```'r':{"data:"text/plain","src":{[0,1]},"p":"<POLICY_ID>}"```|
+|txhash, tx | references a specific tx hash (optional) | ```'r':{"data:"text/plain","src":{[0,1]},"tx":"<TX_HASH>}"```|
+|type | the format/type of the data inside the payload |```'r':{"data:"image/jpeg","src":{[0,1]},"type":"ipfs"```|
+|b, back, backup | fallback, if a link to the data is used in the 'src' tag and is broken revert to a fallback link. | ```'r':{"src":{[0,1]},"b":{[2,3]}} ``` or for multiple fallbacks use ```'r':{"src":{[0,1]},"b":"{'0':{[2,3]}, '1'{[4,5]} }"``` |
```
KtorZ, line 48: How is a "broken" data defined?

CIP-0048/CIP-0048.md
```diff
@@ -0,0 +1,141 @@
...
+|version| this is described in 721 | not sure if it should be version 1.1? instead of 2|
+
+*inside the refences tag*
+|tags that are reseved inside references |description | example
+|---|---| --- |
+|data| contains the data type MIME | for example if base 64, data:"text/plain;base64"|
+|src|contains a list of poitners to payloads | ```"src":[0,1]``` or ``` "src":["payload0","payload1"]```|
+|policy, p | references another policy. If none, assume current policy | ```'r':{"data:"text/plain","src":{[0,1]},"p":"<POLICY_ID>}"```|
+|txhash, tx | references a specific tx hash (optional) | ```'r':{"data:"text/plain","src":{[0,1]},"tx":"<TX_HASH>}"```|
+|type | the format/type of the data inside the payload |```'r':{"data:"image/jpeg","src":{[0,1]},"type":"ipfs"```|
```
KtorZ, line 47: What difference is made between `data` and `type`?

CIP-0048/CIP-0048.md
```diff
@@ -0,0 +1,141 @@
...
+*defintion of the different reference types*
+| type | desciption |
+|---| ---|
+|ipfs|data is stored at ipfs|
+|url|data is found a url|
+|api|data is found at api get request|
+|utf-8| data is directly in payload as utf-8 |
+|b64 | data is in base 64 encoded in payload |
+|reference? | advance feature, points to a reference |
+|none| assume data is raw and using the defined media type |
```
KtorZ, line 59: It seems like each of those types has a precise semantic, what is it?

CIP-0048/CIP-0048.md
```diff
@@ -0,0 +1,141 @@
...
+      }
+   }
+}
+```
+# Retrieve payload information using references
+1. Find all transactions with the 721 label
+2. Check the version tag
+3. If a payload is found append that to an map or some data structure 
+4. Given a nft find there references
+5. Concatenate the data for all given references
```
KtorZ, line 93: How is "concatenate" defined here? What if the data isn't text or bytes?

CIP-0048/CIP-0048.md
```diff
@@ -0,0 +1,141 @@
...
+            "0":["Hello"],"1":["World!"]
+         },
+         "version":"2"
+      }
+   }
+}
+```
+# Retrieve payload information using references
+1. Find all transactions with the 721 label
+2. Check the version tag
```
KtorZ, line 90: What should the version tag be?

CIP-0048/CIP-0048.md
```diff
@@ -0,0 +1,141 @@
...
+         "payload":{
+            "0":["Hello"],"1":["World!"]
+         },
+         "version":"2"
+      }
+   }
+}
+```
+# Retrieve payload information using references
+1. Find all transactions with the 721 label
```
KtorZ, line 89: What is the scope of research here? A block? The whole chain? Should only payload under the same policy be considered?

KtorZ (commented):
I don't think this CIP currently meets the required quality we would demand of a CIP, and I would strongly suggest to spend more time polishing it on various aspects:

- I found quite many typos or confusing sentences; I fixed a few but there are more left. 
- There are many _"holes"_ in the specification at the moment which would lead to potentially many different incompatible implementations. Such a standard should be very precise on how the data is specified and interpreted. Few recommendations: 
    - Avoid double-notations (e.g. `references` or `r`; chose one); 
    - Avoid references to other similar systems in the specification itself (e.g. mention of `ext2`) unless you clearly point to another specification from which you re-use pieces. It's fine to mention as part of the rationale or abstract, but in the specification it can be quite confusing especially if the other system is only vaguely referred to. 
    - Define terms and operations that are ambiguous (e.g. 'concatenate')
    - Be consistent in the wording being used, don't use two different words to refer to the same concept (e.g. `pointers` vs `links`). 
    - Avoid _doubts_ / open-questions (e.g. "not sure if..."), or clearly mark them as "TODO" to make them points to address.

rphair:
@Jack-0 marking "Waiting for Author" since editor consensus at CIP meeting #51 today was that this proposal is technically sound but that the presentation & explanation needs to be cleaned up according to @KtorZ's recent review. 🙏

Jack-0:
I'm going to review this again. It's been a couple months since I wrote this and I can see some ways to improve this whilst also simplifying the specification. Thanks for the feedback I'll be incorporating it, I really like the extensions tag suggestion. @rphair @KtorZ

Jack-0:
I have taken the feedback into consideration and greatly reduced complexity.

**I now propose two CIPS**
CIP-48 - Adds a ext tag to the metadata to allow for more modular extension of NFTs on Cardano
CIP-49 - Adds a payload and reference tag to the 721 metadata (relies on CIP-48)

Please review @rphair @ktorZ when you get the time :)

rphair (dismissed):
@Jack-0 I can't see how it helps for your proposal to be "forked" in this way.  To me there are two possibilities:

1) the "Metadatum extension tag" can _only_ be used as "Extended 721 metadata"... i.e. _not_ by any other mechanism.  If this is the case, these should be combined into a single CIP.

2) the "Metadatum extension tag" can be used by other mechanisms _besides_ your scheme of "Extended 721 metadata"... in this case, the two CIPs would be considered independently and must therefore be split into two separate pull requests.

If taking the latter option # 2 I would recommend creating two **_new_** PR's and digesting / linking the points back to the comments, reviews & discussions in this one.  I'm basing this suggestion on my work with CIP-0013 which changed substantially in scope twice before it was finally merged.

Both times it was a lot of work to summarize and link to the previous PR's (still linked in the CIP-0013 header) but it helped people focus on the matters currently being discussed.  I would encourage you to do the same & include this PR URL in the `Comments-URI:` tag in _all_ your CIP documents remaining in this PR _or_ created in any new PR's.

> [CIP49] This CIP depends on CIP48

That does not logically imply that CIP48 should depend on CIP49, so please first decide if there can be use cases for CIP48 other than the one(s) you've imagined for CIP49.  That should help you determine whether these two threads should be split or combined.

> Please review @rphair @KtorZ when you get the time :)

Thanks but I'm not qualified to judge the validity of the metadatum strategy itself, being more like the editors' group librarian.  Since I have no expertise in this area, my reviews would be mostly based on whether the material here is presented logically in a way that favours critical debate & consensus... e.g. there is no requirement of "one PR per CIP" but if you clarify the question above I believe you will see it breaks down exactly that way.

Jack-0:
Option 2.

CIP 48 is an independent CIP that would add modular expansion to all Cardano NFTS under the 721 metadatum tag via the use of the keyword "ext"

CIP 49 is the original suggested proposal to allow for references and payloads in the 721 metadatum tag.

I segregated this into two CIPs as they are both different specifications. I would be happy to merge into separate CIPs if required for now I will summarize below.

### CIP-48 Introduce a EXT tag to 721 Metadatum metadata
CIP 48 is based of this [comment](https://github.com/cardano-foundation/CIPs/pull/249#discussion_r946446766)
Is a simple implementation that introduces the ```"ext"``` keyword. This will allow other optional 721 metadatum NFT CIPs to be easily appended to the current metadata schema.

### CIP-49 Introduce payloads and references
Introduces payloads and references as mentioned in the current discussion. The key takeaway is that an NFT would use the ```"ext"``` tag defined in CIP-48 to enable the optional feature of payloads and references. Then the metadata would contain the keyword ```"r"``` (similar to the current ```"files"``` keyword usage) to define some on chain references. While the ```"p"``` (payload) keyword is used to store data on-chain. With both ```"p"``` and ```"r"``` tags on chain references are enabled allowing for increased or reduced on-chain data storage.  (Examples in the CIP file)


Thanks for you response. I hope this helps with the discussion. Please tag the appropriate as I am not aware who is best to include. I would be happy to join a CIP or dev meeting to discuss / answer queries in detail.

rphair:
> I segregated this into two CIPs as they are both different specifications. I would be happy to **merge into separate CIPs**

If you mean to merge into two separate **PR's** then I think you have yourself a deal, as they say.  In addition to the reasons I already gave above, having two **new** PR's will help attract feedback from the well populated NFT & metadata community who have commented on similar proposals.

Jack-0:
> > @Jack-0: I segregated this into two CIPs as they are both different specifications. I would be happy to **merge into separate CIPs**
> 
> 
> 
> If you mean to merge into two separate **PR's** then I think you have yourself a deal, as they say.  In addition to the reasons I already gave above, having two **new** PR's will help attract feedback from the NFT & metadata community who have commented on similar proposals.

I'll be turning this into two CIP's today! I will update and reference once done.

rphair (commented):
thanks for splitting @Jack-0 ... can you leave this document in the `CIP-0048` subdirectory but rename the file itself to `README.md`?

Jack-0:
# For reviewers:
Since the last review I have rethought about the problem and generated 2 CIP's as defined below.

### CIP-0049 NFT extension tag
- Aims to make extending nfts easier for developers 
- Makes it easier to define optional metadata standards
- Deprecated the version tag in 721 metadata and replaces it with the `ext` tag that contains an array of CIP's to define what to expect in the metadata.
- PR [CIP-49](https://github.com/cardano-foundation/CIPs/pull/343/files?short_path=3831354#diff-3831354b434087782b3be3990f81c4b719a469899b3dac31f6a6c75f74e385f9)

### CIP-0048 Payloads and Reference
- I've updated CIP-48 and reduced unnecessary complexity. 
- CIP-48 is an optional metadata standard (CIP-49 was created to make it easier to define optional metadata standards)
- Defines the `p` (payloads) and `refs` tags inside the 721 metadata
- Allows an NFT asset within the 721 metadata to use reference data outside it's own metadata / mint transaction scope

rphair:
@Jack-0 a naming collision for CIP-0049  with https://github.com/cardano-foundation/CIPs/pull/250 came up at this week's CIP meeting so we've removed the pending CIP number from https://github.com/cardano-foundation/CIPs/pull/343.  Consequently in your last comment, so it provides a proper index for readers without confusing that reference, can you please also remove CIP-0049 from the heading? (e.g. simply "NFT metadata extension tag")

Jack-0:
I have done that, assuming nothing needs changing with this CIP.

If for some reason we decide that https://github.com/cardano-foundation/CIPs/pull/343 is not appropriate then this CIP would not use such a feature and it would just be yet another CIP that defines optional properties. I was hoping 343 would bring some clarity to NFT optional metadata properties.

rphair:
that's all fine... I was just asking (to avoid confusion in the repository) if you would please remove the terms CIP-0049 from this comment, where it currently says:
> # For reviewers:
> ...
> ### CIP-0049 NFT extension tag
and
> PR [CIP-49](https://github.com/cardano-foundation/CIPs/pull/343)

Jack-0:
Would also like to mention CIP 68 as it would be worth considering Reference and Payload usage in the Datum metadata standard

rphair:
@Jack-0 in the last hour's CIP meeting we discussed how we might finally get this one out of the queue.  Though it would otherwise be in a state where it could be merged as a document for NFT developers to "take it or leave it" (since there is no documented support from existing projects or NFT marketplaces), it seems this is dependent on https://github.com/cardano-foundation/CIPs/pull/343 which in turn requires modifications to [CIP-0025](https://github.com/cardano-foundation/CIPs/tree/master/CIP-0025).

So for now we can only proceed once @alessandrokonrad @SmaugPool can comment here and/or in https://github.com/cardano-foundation/CIPs/pull/343 about how they feel about adding support for the **NFT metadata extension tag**.

And @Jack-0 I think I've asked this question before (when you were separating this PR from #343) but still don't know if it's possible that this mechanism could work _without_ that extension tag.... if so please indicate that clearly here & if so we might be free to go ahead and finally merge this.

rphair (commented):
@Jack-0 likewise with https://github.com/cardano-foundation/CIPs/pull/343#pullrequestreview-1343098726 I'm marking this `Waiting for Author` with the intent to close as superseded by the pending CIP-0088 (also inviting that author to contribute to the discussion here).

We can keep this alive if someone can point out there is any feature provided, or use case handled, which is not addressed in CIP-0088.  Otherwise with no other supportive action taken we will probably close this soon.

Jack-0:
In reply to https://github.com/cardano-foundation/CIPs/pull/249#issuecomment-1339810423

This proposal that adds support for linking multiple nfts via references and payloads will work without the `ext` property, and this can be omitted from the standard it was simply introduced to help the 'developer' make sense of the meta before attempting to parse (like meta meta) but cip [88](https://github.com/cardano-foundation/CIPs/pull/467/files#diff-98412944349d972a6fac6d0fca4bc29265f7b9c275df337b97c0f2fb2310cd3f) might also be on that now. Which would deprecate [my proposal for meta data extension tags](https://github.com/cardano-foundation/CIPs/pull/343)

CIP-0048/README.md
```diff
@@ -0,0 +1,249 @@
...
+      },
+      // payload
+      "p": {
+        "<payload_id>": ["<data>", "<data>"], // payload0
+        "<payload_id>": ["<data>"] // payload1
+      }
+    }
+  }
+}
+```
```
Crypto2099, line 126: I think the best way to embed references in existing metadata structures would be via a structured URI format as defined in the spec. Rather than just using a name and forcing all parsers to scan for references a defined URI schema would make it easier to detect.

**URI Format**
```cbor
uriSchema = {
  scheme : string .size (4..64) ; Examples: ar://, ipfs://, etc
  path * : string .size (1..64) ; Examples: [ar/ipfs hash]
}
```

**Hypothetical CIP-48 URI format**:

```cbor
uriSchema = ["cip48://","referenceA"]
```

In this way a downstream parser may quickly and easily identify a CIP-48-specific reference only on fields that match the format.

CIP-0048/README.md
```diff
@@ -0,0 +1,249 @@
...
+          "name": "...",
+          "mediaType": "...",
+          "type": {
+            "txhash": ["<mint_txhash>", "<mint_txhash>"]
+          }
+        }
+      ]
+    ```
+  - `txhash` is faster than parsing all transactions in a given policy.
+  - `txhash` can dramatically reduce onchain data usage if duplicated data is referenced here _(however, the bytes required to define the reference would have to be less than the reference data itself)_
```
Crypto2099, line 82: If we presume that the majority of references could be declared at a top-level utilizing CIP-88, this is just some rough thoughts but perhaps we could expand the references "type" to optionally point to another on-chain pointer or to a direct payload.

```cbor
{
  48: {
    0: '1.0.0', ; Version?
    1: [ ; Array of references
      {
        "name": "referenceA", ; Name of the reference
        "type": "pointer | payload", ; Maybe an enum of accepted options?
        "payload": { ; Example of a direct payload reference
          "mediaType": "text/plain",
          "content": "My Cool NFT Project"
        },
        "payload": { ; Example of a tx pointer reference?
          "type": "tx",
          "content": "[txhash]"
        },
        "payload": { ; Example of a token point reference?
          "type": "token",
          "content": [(policy_id),(asset_id)]
        }
      }
    ]
  }
}
```

Crypto2099 (commented):
Just some thoughts about how the specification of CIP-0048 could be amended to work well with the proposed CIP-0088. Happy to discuss, review, and iterate.

CIP-0048/README.md
rphair: @Crypto2099 do I understand correctly, with this new approach, that there would be no need for the corresponding proposal that has long accompanied this one (which details the external metadata references)?

- https://github.com/cardano-foundation/CIPs/pull/343

Crypto2099: Yes, I believe the OP and myself agree that the ext flag would be made redundant if CIP-88 were implemented as both have the same function of signaling intent to support a further extension to CIP-25.

rphair: great then, I'll remove the `Likely Deprecated` tag from this CIP and leave it on #343.

CIP-0048/README.md
```diff
@@ -0,0 +1,249 @@
+---
+CIP: 48?
```
rphair, line 2:
```suggestion
CIP: 48
```

CIP-0048/README.md
```diff
@@ -0,0 +1,249 @@
...
+Title: 721 References and Payloads
```
rphair, line 3:
```suggestion
Title: NFT metadata references and payloads
```
For consistency with PR title.

CIP-0048/README.md
```diff
@@ -0,0 +1,249 @@
...
+Authors: Jack (ada9000, https://forum.cardano.org/u/jack7e/, @ada9000_ twitter)
```
rphair, line 4:
```suggestion
Authors:
    - Jack <some@real.email.example.org>
```
There's no means to embed URLs in header text, which makes social links impossible.  Even if getting an anonymous email to use here, we need something that looks like an email address if only so our web site builders work properly.  This is always a list even if having only one member.

CIP-0048/README.md
```diff
@@ -0,0 +1,249 @@
...
+Comments-URI: https://forum.cardano.org/t/cip-extended-721/
```
rphair, line 5:
```suggestion
Discussions:
    - https://forum.cardano.org/t/cip-extended-721
    - https://github.com/cardano-foundation/CIPs/pull/249
```

CIP-0048/README.md
```diff
@@ -0,0 +1,249 @@
...
+Status: Draft
```
rphair, line 6:
```suggestion
Status: Proposed
```

CIP-0048/README.md
```diff
@@ -0,0 +1,249 @@
...
+Type: Informational
```
rphair, line 7:
```suggestion
Category: Metadata
```
see https://github.com/cardano-foundation/CIPs/blob/master/CIP-0001/README.md#categories

CIP-0048/README.md
```diff
@@ -0,0 +1,249 @@
...
+Created: 2022-04-22 (updated 2022-09-22)
```
rphair, line 8:
```suggestion
Created: 2022-04-22
```
Version history strictly follows GitHub, so "updated" dates don't need to be in the header & our collation scripts can't process these because there's no standard form for them.  Current consensus is that particular dated versions can be identified using the GitHub commit hash (forming its own version policy).

CIP-0048/README.md
```diff
@@ -0,0 +1,249 @@
...
+---
+
+## Important note
+
+This CIP mentions the `ext` json key defined by [CIP49](https://github.com/cardano-foundation/CIPs/pull/343/files) which is an outcome of work on CIP48.
```
rphair, line 13:
```suggestion
```
My understanding is that bridging this to CIP88 means that the proposal formerly called CIP49 is now deprecated & I've marked it such on https://github.com/cardano-foundation/CIPs/pull/343 - so if you think otherwise be sure to post in that CIP.  In any case any "important note" should be removed from here and placed on one of the top level document sections.

Duplicate of https://github.com/cardano-foundation/CIPs/pull/249/files#r1149624522.

CIP-0048/README.md
```diff
@@ -0,0 +1,249 @@
...
+
+# Abstract
```
rphair, line 15:
```suggestion
## Abstract
```
H2 section titles make room for an H1 title which will be added programmatically from the title in the preamble. https://github.com/cardano-foundation/CIPs/blob/master/CIP-0001/README.md#structure

CIP-0048/README.md
```diff
@@ -0,0 +1,249 @@
...
+
+- This standard proposes a method to allow NFT assets to reference other onchain data.
+- This would allow users of 721 metadata to increase or reduce the amount of data an NFT asset uses.
+
+# Motivation
```
rphair, line 20:
```suggestion
## Motivation
```

CIP-0048/README.md
```diff
@@ -0,0 +1,249 @@
...
+
+- Large token mints that duplicated data could be dramatically reduced in size by pointing to one transaction payload that contains a ‘boilerplate’ structure.
+- 16kB is the upper limit of data in each transaction but If a user wanted more there is no alternative than to store that data off-chain using an external service such as ipfs.
+- There is no current mechanism to reduce duplicated metadata
+- NFT assets are restricted by there own metadata scope, optional references would prevent this restriction
+- There is currently no way to reference on chain data stored in a different policy
+
+# Specification
```
rphair, line 28:
```suggestion
## Specification
```

CIP-0048/README.md
```diff
@@ -0,0 +1,249 @@
...
+  ```json
+    "p":
+      {
+        "1": ["1 is a valid payload"],
+        "payloadA": ["a...64","b...64"], //a string has a max length of 64 characters
+        "payloadB": ["I'm found by referencing payloadB"],
+      }
+  ```
+
+## General structure
```
rphair, line 98:
```suggestion
### General structure
```

CIP-0048/README.md
```diff
@@ -0,0 +1,249 @@
...
+
```
rphair, line 99: It keeps coming up at CIP meetings recently that, since the chain itself stores CBOR (never JSON), that specifications should be written in CDDL.  If this is one of those cases where that's not important then @KtorZ will correct me... and also he has offered to help with translating these specs into CDDL.

We also recommend that the schemas be put in separate files: https://github.com/cardano-foundation/CIPs/blob/master/CIP-0001/README.md#repository-organization

CIP-0048/README.md
```diff
@@ -0,0 +1,249 @@
...
+      "p": {
+        "<payload_id>": ["<data>", "<data>"], // payload0
+        "<payload_id>": ["<data>"] // payload1
+      }
+    }
+  }
+}
+```
+
+## Notes
```
rphair, line 128:
```suggestion
### Notes
```

CIP-0048/README.md
```diff
@@ -0,0 +1,249 @@
...
+          "mediaType":"image/svg+xml;utf8"
+          "src": ["refToPayloadA", "refToPayloadB"]
+        }
+      ]
+    }
+    ```
+  - The `image` tag can still be used as a thumbnail if required.
+- If there are multiple `refs` defined the references that are _NOT_ defined in the `image` tag have the same behavior as the `files` tag currently defined in CIP25
+
+# Example
```
rphair, line 148:
```suggestion
### Example
```

CIP-0048/README.md
```diff
@@ -0,0 +1,249 @@
...
+     ```
+
+5. use the mediaType (mimetype) to determine the data.
+   - mediaType = "test/plain"
+   - NFT0 contains text "HelloWorld"
+
+# Backwards compatibility
+
+Handled via the use of the `"ext"` tag defined in CIP-48.
+
```
rphair, line 240:
```suggestion
```
I think we've established that the tag will not be used, right?  However compatibility would be handled is a good item to included in the Rationale section which still needs to be defined below.

CIP-0048/README.md
```diff
@@ -0,0 +1,249 @@
...
+# Further considerations
+
+- Later we may want to use different reference types, maybe even plutus contract references. Due to this I have defined the `type` tag inside the `refs` (references) tag.
+- `p` is used instead of **payload** to reduce bytes. Specifically `p` is not used within the `<asset_name>` tag therefore readability is less of a priority than data size. Unlike `refs` which is defined in the `<asset_name>` tag
+- JSON is kept over CBOR or alternatives as this CIP is encapsulated in the 721 metadatum tag and JSON is easier for developers to adopt
+
+### Duplicate data
+
+There could be issues with duplicate payloads. To solve the payload defined in the most recently minted tx takes priority.
```
rphair, line 249:
```suggestion
## Rationale: how does this CIP achieve its goals?

### Further considerations

- Later we may want to use different reference types, maybe even plutus contract references. Due to this I have defined the `type` tag inside the `refs` (references) tag.
- `p` is used instead of **payload** to reduce bytes. Specifically `p` is not used within the `<asset_name>` tag therefore readability is less of a priority than data size. Unlike `refs` which is defined in the `<asset_name>` tag
- JSON is kept over CBOR or alternatives as this CIP is encapsulated in the 721 metadatum tag and JSON is easier for developers to adopt

### Duplicate data

There could be issues with duplicate payloads. To solve the payload defined in the most recently minted tx takes priority.
```
Just sketching out a _Rationale_ from the points already included here... maybe also some of what you already wrote above regarding your design decisions.

Especially I think in the _Rationale_ section you should include:
- why you originally solved this problem with the extra tag, and
- why you abandoned that idea when CIP88 came along.

rphair, line 249: Finally (last section) include a copyright notice of your choice as defined here: https://github.com/cardano-foundation/CIPs/blob/master/CIP-0001/README.md#licensing

rphair (commented):
@Jack-0 we talked about moving this forward at the CIP meeting today and I summarised that you & @Crypto2099 as of 3 weeks ago were working on making this compatible with CIP-88 / https://github.com/cardano-foundation/CIPs/pull/467.

This review is to clean up the formatting as per CIP-0001 which was updated since last year, so we can merge it once you & @Crypto2099 can confirm here the two CIPs are compatible... so we also need you to address his https://github.com/cardano-foundation/CIPs/pull/249#pullrequestreview-1359625936.  (cc @KtorZ)

CIP-0048/README.md
rphair, line 249:
```suggestion
There could be issues with duplicate payloads. To solve the payload defined in the most recently minted tx takes priority.

## Path to Active
```
As per https://github.com/cardano-foundation/CIPs/blob/master/CIP-0001/README.md#path-to-active.  I don't understand the industry well enough to advise on this, but if @Jack-0 you're not sure of acceptance or commercial implementation targets then maybe @Crypto2099 could suggest some.

rphair (commented):
(missed a section in above review: something we check for sooner in the submission process these days)

rphair:
@Jack-0 after review at the CIP meeting today, following last month of inactivity with no response to last review trying to get the document itself in line with existing CIP standards, and without @Crypto2099 providing a bridge to CIP88 which is itself moving forward, this is again marked for deprecation and we intend to close it soon.

rphair:
Final call for any reservations at CIP meeting today regarding deprecating this & the related PR (https://github.com/cardano-foundation/CIPs/pull/343#issuecomment-1549337523).  @Jack-0 please address the last review if you decide to pursue this further.

PR #254, opened 2022-05-08
CIP-0053? | Standard for light wallet backend connection

miguelaeh:
Signed-off-by: Miguel A. Cabrera Minagorri <devgorri@gmail.com>

The intention of this PR is to create a CIP to standardize how light wallets connect to backend services in order to support the connection of wallets to different Cardano networks.

The specification is partially created, just as a reference to what we are looking for with this CIP. Since the intention is to discuss the proposal before finishing the specification, please add any thoughts so we can reach a consensus on the specification. The example is based on how Nami wallet connects with Blockfrost, but there are several wallets in the ecosystem and I would like to hear the opinions of some of them.
I also considered basing the specification on the REST API of `cardano-db-sync` but it was deprecated in favor of the GraphQL one.

Thanks!

---

[see rendered Markdown](https://github.com/miguelaeh/CIPs/blob/standardWalletBackend/standard-wallet-backend/README.md)

### DIFF
--- a/standard-wallet-backend/README.md
+++ b/standard-wallet-backend/README.md
@@ -1,0 +1,201 @@
+---
+CIP: ?
+Title: Standard for light wallet backend connection
+Authors: Miguel Angel Cabrera Minagorri <devgorri@gmail.com>
+Comments-URI: 
+Status: Draft
+Type: Standards Track 
+Created: 2022-05-08
+License: CC-BY-4.0
+---
+
+# **Summary**
+
+Wallets and backend services should consume/expose standard endpoints in order to make wallets work across Cardano networks.
+
+# **Abstract**
+
+Currently each light wallet uses a different backend and most of them do not offer a way to configure the backend host. Each backend exposing a different API makes it really difficult to be able to configure the backend host into the wallet settings.
+Being able to configure the backend host is really important to connect a wallet to a different Cardano network, so anyone can deploy a backend for the new network and connect the wallet to it.
+
+# **Motivation**
+
+As the Cardano ecosystem grows, more Cardano networks start to appear. Leading a project called Shareslake, I deployed a Cardano mainnet, finding that allowing people to connect existing wallets to this new network is really complex.
+
+To illustrate the problem let’s use Nami wallet as example. It allows to configure the transaction submit endpoint, nevertheless, it uses Blockfrost to read on-chain data such us the funds on the wallet address. So, even though you can change the submit endpoint, the wallet won’t show the address funds or transactions to the user.
+
+This CIP is created to suggest a standard on how wallets obtain the on-chain information from the backend, allowing people to easily edit the light wallet connection settings to connect it to a different Cardano network.
+What we are looking to achieve with this CIP is already possible into the Ethereum ecosystem. Let's think on a wallet called Metamask that you can easily connect to different Ethereum networks.
+
+# **Rationale**
+
+Focusing this CIP on the communication between backends indexing the ledger and the light wallets we should be able to suggest a bunch of standard endpoints that allows light wallet's users to just change the backend host and the creators of a new network to deploy a backend knowing that it will work with most of the light wallets.
+
+Cardano mainnet addresses do not include the network magic allowing to use the same address and keys between Cardano networks, so just changing of the backend host on the wallet settings is enough. Testnet addresses are different, they include the network magic, requiring to create a new address per network.
+
+There are two different points to standardise here:
+
+1. Endpoints used to read the on-chain data from the backend.
+2. Endpoints used to submit transactions to the network.
+
+The proposal must be agnostic to the wallet and backend implementations or programming languages so we will be defining REST endpoints that will be exposed by the backends and consumed by the wallets.
+
+To define those endpoints using [OpenAPI](https://swagger.io/specification/) seems ideal as it is widely used to specify RESTful APIs and there are several tools for automatically generating APIs code.
+
+About the first point, we could follow `cardano-db-sync` REST endpoints as standard for consuming on-chain data, but they were deprecated.
+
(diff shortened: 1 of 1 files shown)


standard-wallet-backend/README.md
```diff
@@ -0,0 +1,201 @@
...
+There are two different points to standardise here:
+
+1. Endpoints used to read the on-chain data from the backend.
+2. Endpoints used to submit transactions to the network.
+
+The proposal must be agnostic to the wallet and backend implementations or programming languages so we will be defining REST endpoints that will be exposed by the backends and consumed by the wallets.
+
+To define those endpoints using [OpenAPI](https://swagger.io/specification/) seems ideal as it is widely used to specify RESTful APIs and there are several tools for automatically generating APIs code.
+
+About the first point, we could follow `cardano-db-sync` REST endpoints as standard for consuming on-chain data, but they were deprecated.
```
SebastienGllmt, line 45: You may be interested in Carp. It's a replacement for db-sync that also contains an openapi definition for rest endpoints for webclients. 

However, Carp doesn't handle ledger state (ex: reward balance) or off-chain information (pool metadata), so it will need to be combined in some kind of Cardano OneBox with other tools like Ogmios to provide a complete API that can easily be used for other deployments of Cardano. This is something we're interested in at dcSpark and I think TxPipe is also interested in this.

standard-wallet-backend/README.md
```diff
@@ -0,0 +1,201 @@
...
+
+The endpoints proposed into the specification are based on the required by Nami wallet, which is an open source widely used Cardano light wallet that at the date of writting this CIP is using Blockfrost as backend.
```
SebastienGllmt, line 47: I would discourage using Blockfrost's api. It's not open source and it doesn't handle pagination across rollbacks.

miguelaeh: I agree, just added a couple of endpoints as examples to illustrate the intention of the CIP. Looks like Blockfrost plans to open-source the code regarding their website but it isn't at this moment.

standard-wallet-backend/README.md
miguelaeh: I didn't know about those tools. I will take a look, if they are open-source maybe we could package all of them creating an easy way for deploying the backend.

standard-wallet-backend/README.md
```diff
@@ -0,0 +1,201 @@
...
+
+This CIP is created to suggest a standard on how wallets obtain the on-chain information from the backend, allowing people to easily edit the light wallet connection settings to connect it to a different Cardano network.
+What we are looking to achieve with this CIP is already possible into the Ethereum ecosystem. Let's think on a wallet called Metamask that you can easily connect to different Ethereum networks.
+
+# **Rationale**
+
+Focusing this CIP on the communication between backends indexing the ledger and the light wallets we should be able to suggest a bunch of standard endpoints that allows light wallet's users to just change the backend host and the creators of a new network to deploy a backend knowing that it will work with most of the light wallets.
+
+Cardano mainnet addresses do not include the network magic allowing to use the same address and keys between Cardano networks, so just changing of the backend host on the wallet settings is enough. Testnet addresses are different, they include the network magic, requiring to create a new address per network.
```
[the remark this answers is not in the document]
miguelaeh: About Shelly-Era addresses, there are currently two options for the network id, either `testnet` or `mainnet`, and `testnet` requires the network magic. From [code](https://input-output-hk.github.io/cardano-node/cardano-api/lib/src/Cardano.Api.NetworkId.html):
```
data NetworkId = Mainnet
               | Testnet !NetworkMagic
```

In practical terms, the network is identified by both the network id (`Mainnet` or `Testnet`) and the network magic, as you detailed [here](https://cips.cardano.org/cips/cip34/). It looks like you can actually use the same network ID, for example `1`, but not the same network magic.

Coming back to addresses, in our case, the network id is the same as mainnet, while the network magic isn't. Shelley addresses contain the network id, the same as Cardano mainnet, so they are compatible. Byron addresses won't be compatible as you commented, I assumed for new networks those are just used on the bootstrapping process.

Maybe I should add this clarification to the CIP text.

About losing the funds, I don't see that point, networks are physically separated, so Shelley keys will work on both just depending on which network the user is connected to.

standard-wallet-backend/README.md
miguelaeh: I was told about this project that could be a good alternative: https://api.koios.rest/#overview

standard-wallet-backend/README.md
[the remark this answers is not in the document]
KtorZ: Note that the code pointed above is misleading, and an artifact of the Byron era.

In Byron, testnet addresses required a network magic to be included in the malleable payload. However, in Shelley, this was dropped in favor of a network id, encoded over 4 bits -- which leaves indeed 16 possibilities. By convention, `0` is reserved for testnets (plural), whereas the 15 others are reserved for mainnets, with `1` being the current mainnet and the rest future side-chains and the like. Thus, in principle, a more correct type to encapsulate the network id would be:

```hs
data NetworkId 
    = Mainnet NetworkIndex 
    | Testnet 
    | LegacyTestnet NetworkMagic
```

or, a little fancier with a GADTs and an extra type-parameter (`NetworkId era`)

rphair:
@miguelaeh we discussed this proposal at our CIP Editors' Meeting 2 weeks ago and the resolution was to reject this proposal because the motives & needs for it haven't been established (with a lot of time passing since your last update).

I'd been waiting for the comments to be officially reviewed & published but that might not happen; you can see the draft comments from the meeting here: https://github.com/AntMeyer1/CIPs/blob/Create-Meeting-Notes-54/BiweeklyMeetings/2022-09-30.md#pr254

PR #258, opened 2022-05-15
CIP-0057? | Plutus Smart-Contract Blueprints

KtorZ:
This proposal specifies a language for documenting Plutus contracts in a machine-readable manner. This is akin to what [OpenAPI](https://swagger.io/specification) or [AsyncAPI](https://www.asyncapi.com/docs/specifications/v2.4.0) are for, documenting HTTP services and asynchronous services respectively. In a similar fashion, A Plutus contract has a binary interface which is mostly defined by its datum and redeemer. Since all interactions with contracts are done through transactions, we can also model _endpoints_ of a contract as transitions from a known state to another purely declaratively. 

This document is therefore a meta-specification defining the vocabulary and validation rules with which one can specify a Plutus contract interface, a.k.a **Plutus contract blueprint**.

---

[see rendered Markdown](https://github.com/cardano-foundation/CIPs/tree/KtorZ/CIP-0054/plutus-contract-blueprint/CIP-0057)

### DIFF
--- a/CIP-0057/README.md
+++ b/CIP-0057/README.md
@@ -1,0 +1,422 @@
+---
+CIP: 57
+Title: Plutus Contract Blueprint
+Status: Active
+Category: Tools
+Authors:
+  - KtorZ <matthias.benkort@cardanofoundation.org>
+  - scarmuega <santiago@carmuega.me>
+Implementors:
+  - Aiken <https://aiken-lang.org>
+  - Plu-Ts <https://github.com/HarmonicPool/plu-ts>
+  - ImperatorLang <https://github.com/ImperatorLang>
+  - Lucid <https://lucid.spacebudz.io/>
+  - Mesh.js <https://martify.io/>
+Discussions:
+  - https://github.com/cardano-foundation/CIPs/pull/258
+  - https://discord.gg/yUkkhqBnyV
+Created: 2022-05-15
+License: CC-BY-4.0
+---
+
+## Abstract
+
+This document specifies a language for documenting Plutus contracts in a machine-readable manner. This is akin to what [OpenAPI](https://swagger.io/specification) or [AsyncAPI](https://www.asyncapi.com/docs/specifications/v2.4.0) are for, documenting HTTP services and asynchronous services respectively. In a similar fashion, A Plutus contract has a binary interface which is mostly defined by its datum and redeemer.
+
+This document is therefore a meta-specification defining the vocabulary and validation rules with which one can specify a Plutus contract interface, a.k.a **Plutus contract blueprint**.
+
+## Motivation: why is this CIP necessary?
+
+While publicly accessible, on-chain contracts are currently inscrutable. Ideally, one would want to get an understanding of transactions revolving around script executions. This is both useful to visualize and to control the evolution of a contract life-cycle; but also, as a user interacting with a contract, to ensure that one is authorizing a transaction to do what it's intended to. Having a machine-readable specification in the form of a JSON-schema makes it easier (or even, possible) to enable a wide variety of use cases from a single concise document, such as:
+
+- Code generators for serialization/deserialization of Contract's elements
+- Contract API Reference / Documentation, also automatically generated
+- Extra automated transaction validation layers
+- Better wallet UI / integration with DApps
+- Automated Plutus-code scaffolding
+
+Moreover, by making the effort to write a clear specification of their contracts, DApps developers make their contracts easier to audit (as they're able to specify the expected behavior).
+
+## Specification
+
+### Overview
+
+This specification introduces the notion of a _Plutus contract blueprint_, as a JSON document which it itself a _JSON-schema_ as per the definition of given in [JSON Schema: A Media Type for Describing JSON Documents: Draft 2020-12](https://json-schema.org/draft/2020-12/json-schema-core.html).
+
+Said differently, Plutus blueprints are first and foremost, valid JSON schemas (according to the specification linked above). This specification defines a core vocabulary and additional keywords which are tailored to the specification of Plutus contracts. Tools supporting this specification must implement the semantic and validation rules specified in this document.
+
+Meta-schemas for Plutus blueprints (i.e. schemas used for validating Plutus blueprints themselves) are given [in annexe](./schemas/README.md).
+
+A _Plutus contract blueprint_ is made of a single document describing one or more on-chain validators. By convention, the document is named `plutus.json` and should be located at the root of a project's repository to facilitate its discoverability.
+
+### Document Structure
+
+The document itself is a JSON object with the following fields:
+
+| Fields                       | Description                                               |
+| ---                          | ---                                                       |
+| [preamble](#preamble)        | An object with meta-information about the contract        |
+| [validators](#validators)    | An object of named validators                             |
+| ?[definitions](#definitions) | A registry of definition re-used across the specification |
+
+Note that examples of specifications are given later in the document to keep the specification succinct enough and not bloated with examples.
+
+#### preamble
+
+The `preamble` fields stores meta-information about the contract such as version numbers or a short description. This field is mainly meant for humans as a mean to contextualize a specification.
+
+| Fields         | Description                                                              |
+| ---            | ---                                                                      |
+| title          | A short and descriptive title of the application                         |
+| ?description   | A more elaborate description                                             |
+| ?version       | A version number for the project.                                        |
+| ?plutusVersion | The Plutus version assumed for all validators                            |
+| ?license       | A license under which the specification and contract code is distributed |
+
+#### validators
+
+Validators are the essence of the blueprint. This section describes each validator involved in the contract (simple applications will likely have only a single validator). A validator is mainly defined by three things: a title, arguments (i.e parameters, redeemer and/or datum) and some compiled code. Parameters refer to compile-time arguments that can be applied to a validator template. They must be instantiated to produce a final compiled code as they are embedded in the code of the validator itself. This is often the case for validators that must hold on unique external nonce to produce unique hashes.
+
+| Fields        | Description                                                                                                                     |
+| ---           | ---                                                                                                                             |
+| title         | A short and descriptive name for the validator                                                                                  |
+| ?description  | An informative description of the validator                                                                                     |
+| redeemer      | A description of the redeemer format expected by this validator                                                                 |
+| ?datum        | A description of the datum format expected by this validator                                                                    |
+| ?parameters   | A list of parameters required by the script in addition of the datum and redeemer                                               |
+| ?compiledCode | The full compiled and cbor-encoded serialized flat script                                                                       |
+| ?hash         | A blake2b-224 hash digest of the validator script, as found in addresses. Optional, but mandatory if `compiledCode` is provided |
+
+##### redeemer, datum and parameters
+
+`redeemer`, `datum` and `parameters` items all share the same schema structure. They must define a `schema` that describes how to construct valid on-chain values for each of these fields, and they also specify a purpose (`spend`, `mint`, `withdraw` or `publish`) that indicates in which context it can figure. The purpose is either a string, or an applicator `oneOf` that specifies multiple (distinct) purposes. Similarly, an argument is either an object as described below, or an applicator `oneOf` of such objects. In case where it is defined as an applicator, `purpose` values between objects must be strictly non-overlapping as they are used as discriminant in chosing a schema. This allows, for example, to define different redeemer schemas for different purposes.
+
+| Fields       | Description                                                                                                      |
+| ---          | ---                                                                                                              |
+| ?title       | A short and descriptive name for the redeemer, datum or parameter                                                |
+| ?description | An informative description of the redeemer, datum or parameter                                                   |
+| ?purpose     | One of `"spend"`, `"mint"`, `"withdraw"` or `"publish"`, or a `oneOf` applicator of those                        |
+| schema       | A _Plutus Data Schema_ using the core vocabulary defined below, or a `oneOf` applicator of _Plutus Data Schemas_ |
+
+#### definitions
+
+A set of extra schemas to be re-used as references across the specification.
+
+### Core vocabulary
+
+Plutus blueprints ultimately describes on-chain data value that can be found at the validator's interface boundaries. This means that while we would generally operate at the level of _Plutus Data_, the vocabulary covers in practice any of the possible Untyped Plutus Core (abbrev. UPLC) primitives that can appear at a validator's boundary (e.g. compile-time parameters). Any UPLC primitive is therefore represented as a schema with a `dataType` keyword. The possible values for `dataType` are detailed just below. In addition, and depending on the value of `dataType`, we may find additional keywords in the vocabulary.
+
+| dataType      | UPLC Type  | Description                                                               |
+| ---           | ---        | ---                                                                       |
+| `integer`     | Data       | A signed integer at an arbitrary precision, wrapped as `iData`.           |
+| `bytes`       | Data       | A bytes string of an arbitrary length, wrapped as `bData`.                |
+| `list`        | Data       | An ordered list of Plutus data, wrapped as `listData`                     |
+| `map`         | Data       | An associative list of Plutus data keys and values, wrapped as `mapData`. |
+| `constructor` | Data       | A constructor with zero, one or many fields, wrapped as `constrData`.     |
+| `#unit`       | Unit       | A builtin unit value (the unary constructor).                             |
+| `#boolean`    | Boolean    | A builtin boolean value.                                                  |
+| `#integer`    | Integer    | A builtin signed integer at an arbitrary precision.                       |
+| `#bytes`      | ByteString | A builtin bytes string of an arbitrary length.                            |
+| `#string`     | String     | A builtin UTF-8 text string.                                              |
+| `#pair`       | ProtoPair  | A builtin pair of `Data` elements.                                        |
+| `#list`       | ProtoList  | A builtin list of `Data` elements.                                        |
+
+> **Warning**
+>
+> While they exist for completeness, frameworks are strongly discouraged to use any of the constructs starting with a `#` as they refer to Plutus Core builtins types used by the Plutus virtual machines but aren't meant to figure in outward-facing interfaces. Validators should, as much as possible, stick to `integer`, `bytes`, `list`, `map` and `constructor` (and any composition of those) for their binary interface.
+
+Using these primitives, it becomes possible to represent the entire domain (i.e. possible values) which can be manipulated by Plutus contracts.
+
+### Additional keywords
+
+Similarly to JSON schemas, we provide extra validation keywords and keywords for applying subschemas with logic to further refine the definition of core primitives. Keywords allow to combine core data-types into bigger types and we'll later give some pre-defined definitions which we assume to be part of the core vocabulary and therefore, recognized by any tool supporting this standard.
+
+When presented with a validation keyword with a malformed value (e.g. `"maxLength": "foo"`), programs are expected to return an appropriate error.
+
+Beside, we define a _Plutus Data Schema_ as a JSON object with a set of fields depending on its corresponding data-type. When we refer to a _Plutus Data Schema_, we refer to the entire schema definition, with its validations and with the semantic of each keywords applied.
+
+Unless otherwise specified, keywords are all considered optional.
+
+Here below are detailed all the accepted keywords for each data-type.
+
+#### For any data-type
+
+> **Note** Keywords in this section applies to any instance data-type described above.
+
+##### `dataType`
+
+The value of this keyword must be a string, with one of the following value listed in the first column of the table above. This keyword is **optional**. When missing, the instance is implicitly typed as an opaque Plutus Data. When set, it defines the realm of other applicable keywords for that instance.
+
+##### `title`
+
+This keyword's value must be a string. This keyword can be used to decorate a user interface and qualify an instance with some short title.
+
+##### `description`
+
+This keyword's value must be a string. This keyword can be used to decorate a user interface and provide explanation about the purpose of the instance described by this schema.
+
+##### `$comment`
+
+This keyword's value must be a string. It is meant mainly for programmers and humans reading the specification. This keyword should be ignored by programs.
+
+##### `allOf`
+
+This keyword's value must be a non-empty array.  Each item of the array MUST be a valid _Plutus Data Schema_. An instance validates successfully against this keyword if it validates successfully against all schemas defined by this keyword's value.
+
+##### `anyOf`
+
+This keyword's value must be a non-empty array. Each item of the array must be a valid _Plutus Data Schema_. An instance validates successfully against this keyword if it validates successfully against at least one schema defined by this keyword's value.
+
+##### `oneOf`
+
+This keyword's value must be a non-empty array. Each item of the array must be a valid _Plutus Data Schema_. An instance validates successfully against this keyword if it validates successfully against exactly one schema defined by this keyword's value.
+
+##### `not`
+
+This keyword's value must be a valid _Plutus Data Schema_. An instance is valid against this keyword if it fails to validate successfully against the schema defined by this keyword.
+
+#### For `{ "dataType": "bytes" }`
+
+> **Note** Keywords in this section only applies to `bytes`. Using them in conjunction with an invalid data-type should result in an error.
+
+##### `enum`
+
+The value of this keyword must be an array of hex-encoded string literals. An instance validates successfully against this keyword if once hex-encoded, its value matches one of the elements of the keyword's values.
+
+##### `maxLength`
+
+The value of this keyword must be a non-negative integer. A bytes instance is valid against this keyword if its length is less than, or equal to, the value of this keyword.
+
+##### `minLength`
+
+The value of this keyword must be a non-negative integer. A bytes instance is valid against this keyword if its length is greater than, or equal to, the value of this keyword.
+
+#### For `{ "dataType": "integer" }`
+
+> **Note** Keywords in this section only applies to `integer`. Using them in conjunction with an invalid type should result in an error.
+
+##### `multipleOf`
+
+The value of "multipleOf" must be a integer, strictly greater than 0. The instance is valid if division by this keyword's value results in an integer.
+
+##### `maximum`
+
+The value of "maximum" must be a integer, representing an inclusive upper limit. This keyword validates only if the instance is less than or exactly equal to "maximum".
+
+##### `exclusiveMaximum`
+
+The value of "exclusiveMaximum" must be an integer, representing an exclusive upper limit. The instance is valid only if it has a value strictly less than (not equal to) "exclusiveMaximum".
+
+##### `minimum`
+
+The value of "minimum" must be an integer, representing an inclusive lower limit. This keyword validates only if the instance is greater than or exactly equal to "minimum".
+
+##### `exclusiveMinimum`
+
+The value of "exclusiveMinimum" must be a integer, representing an exclusive lower limit. The instance is valid only if it has a value strictly greater than (not equal to) "exclusiveMinimum".
+
+#### For `{ "dataType": "list" }`
+
+> **Note** Keywords in this section only applies to `list`. Using them in conjunction with an invalid data-type should result in an error.
+
+##### `items`
+
+The value of this keyword must be either another _Plutus Data Schema_ or a list of _Plutus Data Schema_. When this keyword is a single schema, this keywords applies its subschema to all child instances of the list. When it is a list, then the list is expected to have exactly the same number of elements as specified by the keyword and each element must match against the schema corresponding to its position. The list variation is useful to represent product types such as tuples.
+
+##### `maxItems`
+
+The value of this keyword must be a non-negative integer. An array instance is valid against "maxItems" if its size is less than, or equal to, the value of this keyword.
+
+##### `minItems`
+
+The value of this keyword must be a non-negative integer. A list instance is valid against "minItems" if its size is greater than, or equal to, the value of this keyword. Omitting this keyword has the same behavior as a value of 0.
+
+##### `uniqueItems`
+
+The value of this keyword must be a boolean. If this keyword has boolean value false, the instance validates successfully. If it has boolean value true, the instance validates successfully if all of its elements are unique.
+
+#### For `{ "dataType": "map" }`
+
+> **Note** Keywords in this section only applies to `map`. Using them in conjunction with an invalid data-type should result in an error.
+
+##### `keys`
+
+The value of this keyword must be either another _Plutus Data Schema_. This keywords applies its subschema to all keys of the map.
+
+##### `values`
+
+The value of this keyword must be either another _Plutus Data Schema_. This keywords applies its subschema to all values of the map.
+
+##### `maxItems`
+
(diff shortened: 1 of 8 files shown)


scarmuega:
It seems very comprehensive, all the features I had in mind are contemplated (and more). I like the idea of including the state machine transitions as part of the document.

I do have a few question:

- Should we think of how to make it forward compatible with future versions of the Plutus language? We could provide some field to specify to which version of Plutus the document refers to (which, IMO, shouldn't be tied to the version of the schema of this specification)
- Should we provide a way to offload the compiled code templates into an external file / url? Not only because of the potential size of the value, but also because in a CI pipeline they would probably be generated as an independent step.

I can start drafting the json schema document if / when it makes sense.

satran004:
The example looks complete to me and has everything required for the caller.

We may also want to add a "description" field in each datum / redeemer level. Also an additional field "example" if it makes sense. 

>> Should we think of how to make it forward compatible with future versions of the Plutus language? We could provide some field to specify to which version of Plutus the document refers to

Probably  plutus version at global level and then at each validator level (optional).

>> Should we provide a way to offload the compiled code templates into an external file / url? 

If we limit the scope of this document to the contract API / interface for the caller, do we even need a  "compile-code" field in this document ? The "compiled-code" field sounds like an implementation in this case.

KtorZ:
Thanks for the feedback @satran004 & @scarmuega :pray:  

> @scarmuega: **Should we think of how to make it forward compatible with future versions of the Plutus language.**

Yes, we at least need a `plutus-version` field for each validator, or as @satran004 is suggesting, a top-level "default" one which can be overridden per-validator.

> @scarmuega: **Should we provide a way to offload the compiled code templates into an external file / url?**
> @satran004: **If we limit the scope of this document to the contract API / interface for the caller, do we even need a  "compiled-code" field in this document ?**

I like the idea of an external file-template; probably quite easier to generate if needs be; Keep in mind also that I originally imagined that this `compiled-code` field could be totally optional; though strongly encouraged to really enable consumers to identify contracts on-chain in a reliable way. The spec can then really become a kind of ID card of the on-chain part. 

> @satran004: **We may also want to add a "description" field in each datum / redeemer level. Also an additional field "example" if it makes sense.**

Yup! This is totally planned, as well as few other things that are handy for documentation afterwards (e.g. `title`). Overall, I am thinking of drawing a lot of "inspiration" from OpenAPI and AsyncAPI which have been battle-tested for years. 

> @scarmuega: **I can start drafting the json schema document if / when it makes sense.**

If you want to give it a spin, go-ahead. I may give a go at all the high-level types and defining them in terms of the smaller primitives in the next few days. So perhaps we only need to align on the definition of those primitives. 

This is what I have in mind when it comes to the Plutus data "vocabulary" (without the validations): 

```yaml
title: Plutus Data Core Vocabulary

$schema: https://json-schema.org/draft/2020-12/schema
$id: https://raw.githubusercontent.com/cardano-foundation/CIPs/master/CIP-0054/v1/schemas/data

$vocabulary:
  https://json-schema.org/draft/2020-12/vocab/core: true
  https://json-schema.org/draft/2020-12/vocab/validation: true
  https://json-schema.org/draft/2020-12/vocab/applicator: true
  https://json-schema.org/draft/2020-12/vocab/meta-data: false

$dynamicAnchor: plutus-data

oneOf:
  - $ref: "#/$defs/data[sum]"
  - $ref: "#/$defs/data[list]"
  - $ref: "#/$defs/data[map]"
  - $ref: "#/$defs/data[integer]"
  - $ref: "#/$defs/data[bytes]"

$defs:

  data[sum]:
    type: object
    required:
      - type
      - cases
    properties:
      type:
        type: string
        const: sum
      cases:
        type: object
        additionalProperties:
          $dynamicRef: "#plutus-data/$defs/refOrSchema"
        propertyNames:
          pattern: /^[A-Z]/ # In principle, constructor can be any 'uppercase' unicode character.

  data[list]:
    type: object
    required:
      - type
      - items
    properties:
      type:
        type: string
        const: list
      items:
        $dynamicRef: "#plutus-data/$defs/refOrSchema"

  data[map]:
    type: object
    required:
      - type
      - elements
    properties:
      type:
        type: string
        const: map
      elements:
        type: array
        items:
          - $dynamicRef: "#plutus-data/$defs/refOrSchema"
          - $dynamicRef: "#plutus-data/$defs/refOrSchema"

  data[bytes]:
    type: object
    required:
      - type
    properties:
      type:
        type: string
        const: bytes

  data[integer]:
    type: object
    required:
      - type
    properties:
      type:
        type: string
        const: integer

  refOrSchema:
    oneOf:
      - $dynamicRef: "#plutus-data"
      - type: object
        required:
          - $ref
        additionalProperties: false
        properties:
          $ref:
            $ref: "https://json-schema.org/draft/2020-12/meta/core/$defs/uriReferenceString"
```

Note: we could/should also "formally" define it as a new vocabulary and specify the semantic of each operator on the side. Perhaps "overkill" given that the semantic is likely clear-enough as-is and can be expressed using JSON-schema already.

CIP-0054/README.md
```diff
@@ -0,0 +1,267 @@
...
+
+## Motivation
+
+While publicly accessible, on-chain contracts are currently inscrutable. Ideally, one would want to get an understanding of transactions surrounding certain script executions. This is both useful to visualize and control the evolution of a contract life-cycle; but also, as a user interacting with a contract, to ensure that I am authorizing a transaction to do what I intend to. Having a machine-readable specification in the form of a JSON-schema makes it easier (possible) to create a wide variety of use-cases around a single concise document, such as:
```
michaelpj, line 18: This seems tricky, given that many applications have wildly differing structures and interaction models. I think you may need to make some fairly big assumptions about how the interaction models are structured in order to get a usable language for describing the different ways you can interact with things.

KtorZ: So far, I think we're mainly focusing on two things, the first being relatively ubiquitous:

1. Get a machine-readable description of the datum and redeemers, which is basically just a JSON-schema for the Plutus `Data` (see above).

2. For state-machine based contracts, a way to represent transitions between states of the state-machine. Surely, this only applies to a subset of the applications but perhaps long term, may also encourage more and more DApps to organize around state-machines if it gives them access to a variety of tools built on top :)

KtorZ: Note that the first point in itself, while not being anything revolutionary, has already value when it comes to automation and building capable interfaces.

pgwadapool:
I am working on smart contract (rather onchain validator) proof using proof assistant like Coq and extracting from there.
So this will be quite nice feature to have as I can also generate this schema while extracting the contract
I wish to closely follow this CIP

KtorZ:
@satran004 @scarmuega I'll probably make room to continue on this CIP later this week. Any progress on your sides or may I just continue where I left things :sweat_smile: ?

satran004:
@KtorZ No progress from my side :(  You can continue.

scarmuega:
@KtorZ I have made progress on the json schema side, but I do have some blockers. In particular:

- If I understand correctly, the `transitions` concept assumes that the redeemer and datum will be a plutus data type `sum`. Should we enforce the top level of all redeemers / datums to a `sum`? (since any data can be interpreted as a sum of a single case)
- How do we constrain compiled-code arguments? should it be an unbound object? something matching plutus data core vocabulary? or something new?
- Should we treat sub-schema references using the same URI based / JSON pointer resolution mechanism as JSON-schema? or do we want to have a more restrictive / less complex approach?

Where should I share the WIP schema? Is it appropriate CIP etiquette to include support files as part of the PR?

KtorZ:
@scarmuega great news! 

> **@scarmuega: Should we enforce the top level of all redeemers / datums to a sum?**

I'd say: no; because even though any type can be represented as a sum of products, there's still the case of the core primitives such as bytes or integers and I think that a fair number of datums or redeemers may actually be as simple as bytes or integers. 

Having said that, regarding transitions and state-machines per se, I think it makes sense indeed to enforce that redeemers and datums are top-level sums. 

> **@scamuega**: How do we constrain compiled-code arguments? should it be an unbound object? something matching plutus data core vocabulary? or something new?

Compiled code argument _has to_ be Plutus data, hasn't it :thinking:? At least, from a Plutus-Tx standpoint. Not sure about alternatives. So I'd say that yes, it should be a plutus-data, according to the core vocabulary above :point_up: 

> **@scamuega**: Should we treat sub-schema references using the same URI based / JSON pointer resolution mechanism as JSON-schema? or do we want to have a more restrictive / less complex approach?

I'd be more inclined to say the former. Do you have any concrete idea of a more restrictive reference mechanism? The JSON meta-schema definition is bit involved but, mostly for those who define the meta-schema (i.e us), not so much for those using it as a reference. So I'd rather bite the bullet to do things _by the rules_, so that we can hopefully get all sort of existing tooling to work out-of-the-box if we don't mess up the meta-schemas :nerd_face: (which is far from guaranteed in the first iteration).

> **@scamuega**: Where should I share the WIP schema? Is it appropriate CIP etiquette to include support files as part of the PR?

Perhaps we can work from Github gists initially? And move a somewhat final draft to the PR once we have it. As for later, yes it makes total sense to include all the schemas, annexes and side-definitions alongside the CIP with the PR. We've done it for a few CIPs in the past. Keeps things tidy.

scarmuega:
This is the gist that I'm using to work on the "Script" meta-schema:

https://gist.github.com/scarmuega/e7a52753ddea144ee6f58e4fca168e39

Some comments:
- this is WIP, I'll be doing another pass later this week
- I'm using json-schema draft-4 since it's easy to test with my current tooling. We can switch to a newer draft if required.
- I'm using camelCase to maintain consistency with json-schema itself. We can switch if required.
- I'm renamed the `definitions` property to `$defs` to provide conventions similar to what newer json-schema drafts use.
- We should discuss a policy regarding how much "flexibility" should be allowed at the schema-level (eg: should we allow additional properties by default?)

MitchTurner:
Couple of questions:
1. Does this take into account MintingPolicies at all? I didn't see them mentioned--maybe I missed it.
2. It seems like a common pattern for SCs to require multiple Validators/Policies run in parallel, not just as a DAG of state changes. How could you capture something like that in a Blueprint?

KtorZ:
@MitchTurner sorry for not coming back earlier with this; I have been quite busy and didn't really had much time to put back on this proposal. 

The proposal is indeed meant to cover minting policies as well. I have just made a big update, which, I hope cover also your second question. Although, as I mention in the _Path to Active_ section, there are still open questions w.r.t how to represent transitions that go beyond a simple state-machine as you highlight.

CIP-0057/README.md
```diff
@@ -0,0 +1,414 @@
...
+      "to": { "$refs": "#/validators/main/datum" },
+      "via": { "$refs": "#/validators/main/redeemer" },
+    },
+    "vote": {
+      "description": "A vote that is increasing the voting count and the name counter by minting more tokens.", 
+      "from": { "$refs": "#/validators/main/datum" },
+      "to": { "$refs": "#/validators/main/datum" },
+      "via": { "$refs": "#/validators/main/redeemer" },
+    }
+  },
```
alessandrokonrad, line 366: Are these endpoints meant to be implemented like that in the actual off-chain code? To me it seems like these information are not sufficient. A validator could also depend on the ScriptContext e.g. valid time range. Or it is required to spend a certain input to prove ownership of token X, etc.   
But maybe I understand this wrong here, happy to be corrected.

KtorZ: What do you mean by "implemented like that" ? 

This is mainly meant to be a description of the various contract endpoints, and ideally yes, it should contain enough information to allow to fully construct a transaction from the specification (and user inputs). It is insufficient at the moment indeed, which I touch upon in the rationale. There are additional constraints of the transaction that needs to be expressed somehow. I think that the tx-constraints from plutus-tx offers a quite nice abstraction for this and it seems possible to me to translate that in plain JSON. 

Ideally, I would like this effort to be driven by one or more real-life use case, trying to fully specify an existing contract and address see where the shortcomings are in the current specification.

alessandrokonrad: > What do you mean by "implemented like that" ?

Do these endpoints represent the unit state machine transition or endpoints which are practically used. The reason why I'm saying this is because an endpoint in the actual off-chain code could have multiple transitions and also be conditional.

KtorZ: They represent how an application should interact with the contract; so they describe much more the off-chain interactions and conditions, as well as additional constraints that apply. 

Conceptually, a long-term could be to have UI automatically generated from a blueprint, all users would have to do is connect a wallet.

CIP-0057/README.md
```diff
@@ -0,0 +1,414 @@
...
+    "main": {
+      "description": "The lobster contract validator, verifying vote counts and driving token evolution.",
+      "datum": {
+        { "$ref": "plutus-types.json#/$defs/Unit" }
+      },
+      "redeemer": {
+        { "$ref": "plutus-types.json#/$defs/Unit" }
+      },
+      "compiledCode": {
+        "template": "59094c01000{{ lobsterParams }}006002003",
```
L-as, line 328: What does this mean?

CIP-0057/README.md
```diff
@@ -0,0 +1,414 @@
...
+
+## Motivation
+
+While publicly accessible, on-chain contracts are currently inscrutable. Ideally, one would want to get an understanding of transactions revolving around script executions. This is both useful to visualize and to control the evolution of a contract life-cycle; but also, as a user interacting with a contract, to ensure that one is authorizing a transaction to do what it's intended to. Having a machine-readable specification in the form of a JSON-schema makes it easier (or even, possible) to enable a wide variety of use cases from a single concise document, such as:
+
+- Code generators for serialization/deserialization of Contract's elements
+- Contract API Reference / Documentation, also automatically generated
+- Extra automated transaction validation layers 
+- Better wallet UI / integration with DApps
+- Automated Plutus-code scaffolding
```
L-as, line 26: - I'm not sure what this is supposed to mean, do you mean you want to describe the datums, redeemers, token names in question?
- Human readable? Why would this be better than writing something manually?
- Not sure what this means.
- Improving wallet UX makes sense, noto sure what you mean about integration with dapps.
- I'm not sure this is useful for that. You usually go the other way around.

CIP-0057/README.md
```diff
@@ -0,0 +1,414 @@
...
+Type: Process
+Created: 2022-05-15
+License: CC-BY-4.0
+---
+
+## Abstract
+
+This document specifies a language for documenting Plutus contracts in a machine-readable manner. This is akin to what [OpenAPI](https://swagger.io/specification) or [AsyncAPI](https://www.asyncapi.com/docs/specifications/v2.4.0) are for, documenting HTTP services and asynchronous services respectively. In a similar fashion, A Plutus contract has a binary interface which is mostly defined by its datum and redeemer. Since all interactions with contracts are done through transactions, we can also model _endpoints_ of a contract as transitions from a known state to another purely declaratively. 
+
+This document is therefore a meta-specification defining the vocabulary and validation rules with which one can specify a Plutus contract interface, a.k.a **Plutus contract blueprint**.
```
L-as, line 16: To make sure I understand the wording here, you want to define a language in which you can define the behaviour of your protocol?

CIP-0057/README.md
```diff
@@ -0,0 +1,414 @@
...
+- Better wallet UI / integration with DApps
+- Automated Plutus-code scaffolding 
+
+Moreover, by making the effort to write a clear specification of their contracts, DApps developers make their contracts easier to audit (as they're able to specify the expected behavior). 
+
+## Specification
+
+### Overview 
+
+This specification introduces the notion of a _Plutus contract blueprint_, as a JSON document which it itself a _JSON-schema_ as per the definition of given in [JSON Schema: A Media Type for Describing JSON Documents: Draft 2020-12](https://json-schema.org/draft/2020-12/json-schema-core.html).
```
L-as, line 34: I don't think the encoding of the blueprint is important. You probably want to make this section much shorter and perhaps move it around.

L-as (commented):
My over all comment is I'm not sure what this CIP is about.
I'm having a hard time understanding the language you've designed here,
which AFAICT is meant to model protocol behaviour.

What's the point in making this machine-readable by the way?

CIP-0057/README.md
KtorZ: The encoding of the blueprint is the whole purpose of this CIP :)

CIP-0057/README.md
KtorZ: Not the behavior (though to some extend, this is a stretch goal / bonus), but the application binary interface. That is, what other programs need to know in order to interact with a particular Plutus contract.

CIP-0057/README.md
KtorZ: It seems like you aren't familiar with tools such as OpenAPI or AsyncAPI, which I suggest to familiarize yourself with. Although, this would likely only speak to you if you've ever interacted with an HTTP or async service -- or designed one. These kind of use cases are very common for client applications integrating with an API service. In this scenario, client applications are wallets, explorers or any other tool really that is interacting with the chain to a certain extend, and the service is the on-chain DApp itself. At present, most DApps are also building their own frontend and somewhat obfuscating their contract interface. This proposal aims to make it possible for anyone to implement a somewhat robust frontend for a specified DApp (though this isn't the only use case).

CIP-0057/README.md
KtorZ: Read the specification: https://mustache.github.io/mustache.5.html

Although, this part -- the templating / compiled code -- is arguably clunky. There's just no good way today to obtain a parameterized contract code than to recompile it; I'll probably end up removing this one from the proposal in the end.

KtorZ:
> @L-as My over all comment is I'm not sure what this CIP is about. I'm having a hard time understanding the language you've designed here, which AFAICT is meant to model protocol behaviour.

Because it's not a "language"; but rather a specification format which can abide by to define the interface of a Plutus smart-contract. The goal isn't to specify a protocol behavior, but to document its interface. 

> @L-as  Perhaps if you speak less about JSON and more about the essence of the language, it will be more clear.

JSON is the essence of the proposal. It's about representing smart contract interfaces as JSON schemas.  

> @L-as What's the point in making this machine-readable by the way?

This is stated quite clearly in the motivation I believe, but I'll give it another shot. Being able to specify a contract interface in a machine readable format enables tooling to reduce the amount of implementation logic needed to interact with a particular DApp. There's a lot of commonality between DApps, and capturing these in the form of the proposed format will help to automate implementation steps and to reduce the boilerplate needed for any new DApp integration. On can support the format specified by this proposal and be seamlessly integrated at the same time with all DApp which have defined an abiding specification.

CIP-0057/README.md
KtorZ: I realize the only mention to mustache is in the meta-schema; which is a bit like -- hidden far away. If kept, this should be mentioned more clearly in the main specification document itself.

L-as:
@KtorZ OK, if you want to use this to allow easy interfacing with contracts, then I don't think this is suitable (yet). The dApps covered by this are very limited unless you extend the functionality here.

FWIW, this _is_ a language, that's what a language is, but I don't think taking too much inspiration from the HTTP world is a good idea.

JSON isn't a core part of this proposal, it is a way to encode data. You should formulate the possible interfaces then describe the JSON encoding rather than conflating the two (i.e. essentially denotational design).

KtorZ:
The proposal is indeed only able to represent limited use case at this stage (it's still a draft after all :)) , and this is why we want to actually work with a few DApps developers and write a specification for 'real world DApps'. This in order to identify the gaps.

Given that HTTP is, by far, the most used communication protocol in the world, I think there's some sense in taking inspiration from it. Don't you think? I am not interested in defining an abstract language for this. The proposal is very down-to-earth and should enable implementation of tools from an unambiguous format. This goes with defining what the medium of encoding is, what the acceptable keywords are and what is their expected semantic. Wording may not be perfect but we'll find out as we implement tools and write specifications with it. 

It's a good opportunity to try to take what works well from the existing industry and apply it to ours. However, It may as well be a dead end. We'll see. 

If you have concrete suggestions, please say so.

KtorZ:
> You should formulate the possible interfaces then describe the JSON encoding rather than conflating the two

To add a little precision to this, it is not the goal of this proposal to formulate the possible interfaces because they already exists: Plutus does define them. It is not something up for debate (or at least, it's far from the goal of this CIP to argue about these interfaces -- it embraces them as such).
This proposal is precisely about describing how to encode these interfaces as JSON, and as much details at possible to automate front-end interacting with them.
Another option could be to start from Haskell types and generate code from them. Though, this only works is you know Haskell and makes it much harder to build tool that aren't written in Haskell. Json schema offers a language agnostic interface that is widely used across the industry.

MitchTurner:
> this is why we want to actually work with a few DApps developers and write a specification for 'real world DApps'

I'm still early in development with my Naumachia library, but I'd be happy to help workshop the blueprint specs on actual contracts. My `SCLogic` trait kinda fits the same role, as an interface:
https://github.com/MitchTurner/naumachia/blob/aa42eb95593137ad069b9a460fd23edb2835469c/always-succeeds-contract/src/logic.rs#L44

Once I get the contracts compiling to Wasm, where they become portable to different languages, a universal blueprint might be necessary (if only for a subset of your intended use cases).

bladyjoker (commented):
Hey @KtorZ nice work, make me happy to see someone else thinking about this as well. (Also please check our Catalyst proposal at https://cardano.ideascale.com/c/idea/421376)

I have a couple of concerns about the overall proposal...

**No separation of Abstract vs Concrete syntax**

In essence this proposal documents how to JSON encode Plutus Data encoded types...which imo misses some opportunities.

Datum/redeemer types are not only used to interact with the onchain Plutus script via Datum/Redeemers, they are manipulated at the frontend, exchanged with the backend and maintained in relational databases. The Plutus Data structure of a type is for the most part only relevant when interacting with the onchain part.

Codegen seems to be limited because users remain at Plutus Data level and manage all of their Domain types via the `plutus-data/types.json`. This will most likely result in shallow translation layer in various codebases that take true native Domain types to their 'blueprint' counterpart.

What is needed is a proper separation of Concrete vs. Abstract syntax, for instance, let's imagine a Haskell ADT like language for describing shareable Domain types (ie. Abstract syntax)

```haskell
data User = User {
  userId :: Integer,
  userProfileUtxo: TxOutRef
}

data UserProfile = UserProfile {
  userType :: UserType
  userDob :: Date,
  userFirstName, userLastName :: String
}

data UserType = Regular | Admin
```

Let's imagine these domain types are embedded in Datum/Redeemers as well. From such a specification we can 'derive' a Plutus Data encoding, as is already the case in Haskell/PlutusTx. However, we can do a lot more:

  - Codegen types and APIs for constructing such types as close to what's considered native in a target language as possible (SQL, typescript/javascript/purescript, haskell/plutarch/plutustx)
  - Formally specify various encodings and achieve separation of concerns. For instance JSON and Plutus Data, that are then implemented in various language ecosystems as part of some runtime libraries and is completely opaque to users (**no user should care about encodings, whether it's JSON or Plutus Data**). 
  - Why stop at encodings? Say you want to provide Eq, Show, Arbitrary instances, documentation? It seems to me that we don't want to talk in terms of one specific encoding such Plutus Data to carry around shareable knowledge about our Domain types.

**Extending JSON schemas means losing supporting tooling**

AFAIU even JS Schema vocabs and meta-schemas, that serve the purpose of embedding some domain specific schemas, are loosely specified, and you essentially lose most relevant tooling support as you'd have to extend the associated runtime with understanding for your 'extended' schema.

This is a minor concern but perhaps warranted. Aside from JSON schemas being bloated, overly verbose for what is essentially a JSON eDSL in JSON and tightly coupled to Javascript and JSON based web environments, they seems to lack facilities for type polymorphism and proper sum types. 

**Separate Blueprint from data type (schema) definitions**

AFAICT the blueprint file eventually contains a type definition for a Plutus script, such as a Validator or a Minting Policy, but also any CompiledCode. That's fine and definitely needed, but it's also a very small addition to the core of the issue which is _How to specify and share domain types_

The same can be said about Endpoint, good idea, but it's unnecessary in this proposal.

Perhaps consider moving the blueprint part into a separate cip?

KtorZ:
@bladyjoker thanks for the feedback and my apologies for not coming back earlier with an answer. It's been pretty busy on my end and the proposal's been on hold for a while. 

Regarding your suggestions; note that I am not interested in providing yet-another-abstraction layer -- especially in Haskell. As stated previously, this proposal is precisely about finding a way to document the binary interface (i.e. the PlutusData on-chain encoding) in a machine-readable way. The blueprint can then be annotated with details that may be helpful for frontend applications to automatically render UI based on it.  

> @bladyjoker:
> what we want at a high level is:
>
> ```
> type Validator a b = a -> b -> ...
> ```

I don't think that this is what we want, no. Validators are fundamentally `Data -> Data -> ...` and this ain't going to change anytime soon looking at Plutus evolutions. We need a way to properly document how to construct a `Data`. And this needs to be language-agnostic so that it can serve multiple languages. 

---

Regarding progresses on the proposal itself, I am currently working on a Rust implementation of this encoding and we'll make a attempt at _producing_ blueprints as part of the [Aiken](aiken-lang.org/) toolkit.

KtorZ:
Specification updated following from recent implementation in [aiken](https://github.com/txpipe/aiken) + internal discussions on TxPipe's discord.

rphair (approved):
@KtorZ I will leave it to you to merge this if & when you're satisfied that @michaelpj & other interested parties have had a chance to comment on the latest round of changes.  cc @SebastienGllmt

KtorZ:
@rphair thanks. I believe Mesh & Lucid have now integrated the blueprint format. I don't know if @christianschmitz, @michele-nuzzi, @nielstron or @nau have any intention to also support Plutus Blueprints as a compilation artifacts for their respective framework, but I would welcome their feedback here. 

That's also an official invitation to support Plutus Blueprints and take a first step towards interoperability of smart-contract solutions in our ecosystem. I'd love to see every on-chain project with a top-level `plutus.json` that describes the contract's binary interface, regardless of the framework used to write on-chain code.

christianschmitz:
I'll have a look

michele-nuzzi:
added to the to-do list

nielstron:
Yeah will definitely add to todo list for @imperatorlang too

nielstron:
Just a side note: this only specifies an interface for spending purpose smart contracts (because the datum is required) and hence leaves out minting, staking and cerification contracts. Is this intentional?

KtorZ:
@nielstron the datum is optional. Only mandatory for spending validators. 
So, no, the schema is meant to cover all kind of validators.

nielstron:
@KtorZ Right I didn't see that. It says that the datum should be present iff purpose=spend - but there appears to be no (machine readable) way to specify the intended script purpose or am I missing that too?

KtorZ:
There was one, and I removed it. So that's an artifact I need to fix. An earlier version of the blueprint used to have a field "purpose". However, we realized that having such a field doesn't make much sense because a same validator is often reused between different purpose (that's especially the case between withdrawals and certificates validators). 

The purpose is an _intrinsic_ properties of validators, so it doesn't quite make sense in the outward blueprint. In the end, we truly have two kinds of validators:  those who have 2 parameters, and those who have 3.

nielstron:
Would it maybe make sense to also allow the anyOf syntax for purposes? I think it would indeed be useful to know what purposes a validator is intended to be used for while I agree there are validators that don't serve a single purpose

The 2/3 parameter distinction is also not entirely true. The [wrapped token contract](https://github.com/ImperatorLang/eopsin/blob/master/examples/smart_contracts/wrapped_token.py) written in eopsin serves both as a minting and as a spending script.

KtorZ:
Yes indeed. We've been thinking about doing something similar for Aiken-generated validators. Which raises an interesting question w.r.t blueprints. We may need to be able to represent a datum _as optional_. Right now, it's optional in the meta-schema, but that's different. 

I reckon the blueprint should be able to express just what you describe: "this validator can have 2 or 3 arguments", which is another way of saying "this validator can have a datum, or not" and provide the expected format for the datum when it's provided.

nielstron:
Maybe a better way to describe this behaviour is to generate distinct validator descriptions but with the same validator hash?

In a sense, even though it _is_ the same validator, it validates different things (in the token wrapping case, the spending contract checks that the correct amount of tokens is unlocked, while the minting contract validates the correct amount of wrapped tokens is minted)

This makes the purpose necessary to be specified to describe what this _invocation_ of the validator checks.

KtorZ:
I think it's even wider than the purpose then. You could imagine capturing all possible invocation of a specific validator as separate blueprint object. They'd all have the same validator hash, but the datum / redeemers may be different, as well as the description of the validator.

nielstron:
More points in favor of explicitly specifying the purpose

1) I think the fact that the same validator may be used for different purposes should be more or less transparent for the user. The user just wants to perform an action that requires invocation of one or more contracts (for one or more different purposes) and whether or not these are the same doesn't have to bother the user.

2) The tooling should be able to inform the user that they are using a contract against its intended specification. This is only possible if the tooling can identify the intended usage by combination of contract + purpose.

nielstron:
> I think it's even wider than the purpose then. You could imagine capturing all possible invocation of a specific validator as separate blueprint object. They'd all have the same validator hash, but the datum / redeemers may be different, as well as the description of the validator.

True. Of course at some point one has to cut a line and abstract.

I just think in practice different datums at the spending contract will be very unlikely to describe entirely different behaviour, because this implies that the same contract is used for completely unrelated things. But I would argue that a UTxO locked at a contract with a different datum describes a potentially completely different invocation.

The Redeemer describes what the user wants to do with the contract. It makes sense IMO to capture the different redeemers in the same description? But yeah one can imagine that the tooling may output different information about contract invocation based on the type of Redeemer passed into it.

nielstron:
Speaking in REST terms, in a sense you can imagine contract + purpose + optional datum as a path and the redeemer as method

nielstron:
Maybe it makes more sense to considerthe contract as a path, the purpose as the method and the Redeemer as auxiliary data that is always passes. 


The datum is a query parameter, it modifies how a path reacts to the spend method but should not  ause any drastic changes in behaviour.


This does the same as the REST specification: It takes a very powerful protocol and specifies how specific invocations _should_ behave. Of course the webapp or dApp may behave j  completely different ways but these are the desired ways.

KtorZ:
`purpose` added as an optional discriminant for validators arguments. As a consequence, datums, redeemers and parameters schemas are now nested under a `schema` field.

Ryun1 (approved):
Well written proposal with a good amount of discussion. 

Since this is not my area of expertise I don't have any comments to add, so I will approve.

michele-nuzzi:
@KtorZ I'm starting to implement a very raw version of the CIP in plu-ts.

Question: should we add support for function types and delayed types?

I know these can't be expressed as Data **but** they _could_ be used as script parameters.

KtorZ:
@michele-nuzzi I've also pondered about those indeed... Technically speaking, you're right, we can apply any Plutus term to a parameterized validator (including `error`). Though I did chose to omit them from the blueprint representation and only have constants and data in there. 

The reason being that I couldn't really find a use-case (in terms of interoperability) that would justify having them. Do you have any 🤔 ?

KtorZ:
@michele-nuzzi also small note: we noticed some minor discrepancy in the JSON schemas currently specified in annex. I'll try to update them A.S.A.P. but do not hesitate to cross-check things with some Aiken examples and the Aiken implementation at the moment.

michele-nuzzi:
For the reasons specified in the edit above, I usually prefer to compile dynamically scripts that might require these parameters, so I never end up having to deserialize just to apply the parameter. So no, I don't have an example, but I believe other languages do not have the option of compiling dynamically.

PR #85, opened 2021-04-09
CIP-NFTMetdataStandard

alessandrokonrad:

PR #263, opened 2022-05-18
CIP-0054? | Cardano Smart NFTs

kieransimkin:
I've received a lot of feedback on this proposal both on [the forum](https://forum.cardano.org/t/cip-draft-cardano-smart-nfts/100470) and in direct communication with various NFT creators. I've redrafted this proposal based on that feedback - I'm now moving on to create an initial implementation at Artifct - I will of course continue to take feedback and update the document as the implementation progresses.


CIP-0054/README.md
```diff
@@ -0,0 +1,225 @@
+---
+CIP: 54
+Title: Cardano Smart NFTs
+Authors: Kieran Simkin
+Comments-URI: https://forum.cardano.org/t/cip-draft-cardano-smart-nfts/100470
+Status: Draft
+Type: Standards
```
KtorZ, line 7: This one would be a "Process" track as it does not entail any changes in the _core_ Cardano components (no ledger or consensus changes; it is purely built upon the current system capabilities).

rphair:
FYI good discussion of this proposal in last CIP Editors' meeting here: https://github.com/cardano-foundation/CIPs/blob/cip-meeting-notes-46/BiweeklyMeetings/2022-06-07.md#pr263

michaelpj:
Just continuing my prophet-of-doom role to point out that CIP-25 still has [security problems](https://github.com/cardano-foundation/CIPs/pull/85#issuecomment-1054123645).

CIP-0054/README.md
Mirris: Support of the mutable NFT part is necessary for recording system eternal changes and will allow to chase use cases like gaming (build your character, add items to stash, work your stats) as well as multi-verse cases (transfer your character and play in another game; interactions with multi-verse market places, etc.). You can also chase cases as Cardano Weather Station, Cardano Seismic Activity, Cardano Financial Markets, New Google Maps layers from Cardano NFTs (Helium internet coverage, Favourite fruit prices in your surroundings,...). And many many many more!

CIP-0054/README.md
```diff
@@ -0,0 +1,227 @@
...
+Created: 2022-05-18
+License: CC-BY-4.0
+---
+# Cardano “Smart NFTs”
+
+## Abstract
+
+This CIP specifies a standard for an API which should be provided to Javascript NFTs, it also defines some additions to the 721 metadata standard defined in [CIP-0025](https://github.com/cardano-foundation/CIPs/tree/master/CIP-0025) which allow an NFT to specify it would like to receive certain current information from the blockchain.
```
KtorZ, line 15: Anything below this paragraph seems to belong more the "Motivation" than the abstract.

CIP-0054/README.md
```diff
@@ -0,0 +1,227 @@
...
+CIP: 54
+Title: Cardano Smart NFTs
+Authors: Kieran Simkin
+Comments-URI: https://forum.cardano.org/t/cip-draft-cardano-smart-nfts/100470
+Status: Draft
+Type: Standards
+Created: 2022-05-18
+License: CC-BY-4.0
+---
+# Cardano “Smart NFTs”
```
KtorZ, line 11: Please no title in the markdown :pray:, the title already is part of the front-matter data.

CIP-0054/README.md
```diff
@@ -0,0 +1,227 @@
...
+
+It would also be nice to be able to retrieve data that has been previously committed to the blockchain, separately to the NFT which wishes to access it. This would be useful for retrieving oracle data such as current Ada price quotes – as well as for allowing an NFT to import another NFT’s data.
+
+Further to this - for on-chain programatically generated NFTs, it makes sense to mint the code to render the NFT as one token, and then have the individual NFTs contain only the input variables for that code. This CIP specifies an additional metadata option which specifies that an NFT should be rendered by another token - this will massively reduce code duplication in on-chain NFTs.
+
+This combination of functionality enables many exciting new possibilities with on-chain NFTs.
+
+## Description:
```
KtorZ, line 25: Please rename as `Specification` (no column)

CIP-0054/README.md
```diff
@@ -0,0 +1,227 @@
...
+
+In tandem with the additional metadata, we also define a standard for the Javascript API which is provided to the NFT within the sandbox.
+
+## The Metadata
+
+Minting metadata for Smart NFTs – based on the existing [CIP-0025](https://github.com/cardano-foundation/CIPs/tree/master/CIP-0025) standard:
+
+```
+{
+	"721": {
```
KtorZ, line 39: This proposal _should_ use new key or, a new `version` number for 721. But either way, there needs to be a way to programmatically tell the expected structure of the metadata.

CIP-0054/README.md
```diff
@@ -0,0 +1,227 @@
...
+					"id": <string>
+					"name": <string>,
+					"mediaType": <mime_type>,
+					"src": <uri | array>,
+					<other_properties>
+				}],
+
+				"uses": {
+					"transactions": <string | array>,
+					"tokens": <string | array>,
```
KtorZ, line 59: I'd suggest to avoid double types when designing API as much as possible. If there's only one item, then an array of one item is just fine and makes it much easier for code dealing with that structure later on.

KtorZ: Also, consider providing a CDDL specification for that on-chain format, rather than JSON. JSON is only how on-chain metadata are typically rendered, but they are actual stored as binary objects following in CBOR.

CIP-0054/README.md
```diff
@@ -0,0 +1,227 @@
...
+
+### The transactions key
+
+To enable evolving NFTs where the NFT monitors a transaction history and changes in response to new transaction metadata, we define a sub-key called “transactions”, which can contain either a string or an array, specifying the tokens or addresses the NFT wishes to receive the transaction history for. These transaction histories will be provided to the NFT via the Javascript API detailed below.
+
+We also define a special keyword “own” which can be used to monitor the NFT’s own transaction history. So if we wish to create an “evolvable” NFT that can respond to additions to its own transaction history, the “uses” key within the metadata would look like:
+
+```
+	"uses": {
+		"transactions": "own"
```
KtorZ, line 80: What would be the semantic of "own"? How would a transaction identify as "own"? Would that refer to any transaction that has moved this NFT around (even if it didn't change wallets but was just carried as change)? Would that include the minting transaction? etc..

CIP-0054/README.md
```diff
@@ -0,0 +1,227 @@
...
+When specifying an external token to monitor, you should do so via the token’s fingerprint as in this example:
+
+```
+ 	"uses": {
+		"transactions": [
+			"own",			
+			"asset1frc5wn889lcmj5y943mcn7tl8psk98mc480v3j"
+		]
+	}
+```
```
KtorZ, line 105: Both examples with the asset id and the address are heavily user-facing / JSON-specific. In principle, this metadata would be stored on-chain and I would advise against storing these things are text strings. Both addresses and asset ids have well-defined on-chain binary representation, use them! 

Also, regarding asset fingerprints, these are not meant to be global unique identifiers. They are fine in the context of a single wallet but they should not be used as id in a global setup. Rather, use the asset id (i.e. policy id + asset name). Incidentally, it would be _okay_ to however display the asset fingerprint to the user, so long as it is stored as an asset id.

CIP-0054/README.md
```diff
@@ -0,0 +1,227 @@
...
+
+### The tokens key
+
+To enable modifier tokens - (that is, a token which you can hold alongside a Smart NFT which changes the Smart NFT's appearance or behaviour in some way); we provide a way for the Smart NFT to monitor the tokens held by a specific address. Similarly to the “transactions” key, the “tokens” key will also accept either a string or array.
+
+In this case we define the “own” keyword to mean the tokens held by the same stake key that currently holds the Smart NFT itself.
```
KtorZ, line 111: What about stake part not directly associated to a single stake key? For instance, a stake part captured by a native monetary script as a combination of three keys? Or simply, an address without stake part? Would those be simply excluded from the 'own' keyword?

CIP-0054/README.md
```diff
@@ -0,0 +1,227 @@
...
+
+It would also be nice to be able to retrieve data that has been previously committed to the blockchain, separately to the NFT which wishes to access it. This would be useful for retrieving oracle data such as current Ada price quotes – as well as for allowing an NFT to import another NFT’s data.
+
+Further to this - for on-chain programatically generated NFTs, it makes sense to mint the code to render the NFT as one token, and then have the individual NFTs contain only the input variables for that code. This CIP specifies an additional metadata option which specifies that an NFT should be rendered by another token - this will massively reduce code duplication in on-chain NFTs.
```
KtorZ, line 21: side-note: I'd love to see some example of existing JavaScript NFT in introduction to build a mental picture of those :)

CIP-0054/README.md
```diff
@@ -0,0 +1,227 @@
...
+Via the Javascript API, the rendering token will always receive the properties of the child token which specified it as its renderer. This means if you wish to use a renderer with a token which also evolves based on its own transaction history, you will need to specify both "renderer" and "transactions" keys within the child token, and within the renderer token you do not need to specify these keys. 
+
+For example, to create a Smart NFT which is rendered by another token, and is also evolvable based on its own transaction history, the "uses" key would look like this:
+
+```
+	"uses": {
+		"transactions": "own",
+		"renderer": "asset1frc5wn889lcmj5y943mcn7tl8psk98mc480v3j"
+	}
+```
```
KtorZ, line 148: I am not sure to understand the rationale for storing the rendering code _on-chain_ :thinking: ? It seems to me like a misuse of a blockchain as a database (which it is not!). If the goal is to enforce that only a certain code is executed for rendering an certain NFT, I'd suggest to instead store a hash of that code and have client applications select the right rendering code based on that. 

If the issue is to keep the client as dumb as possible then, I'd suggest to store the code on a system meant to store files such as IPFS, and only record a link / hash to that file on the Cardano blockchain.

CIP-0054/README.md
rphair: i.e. no "colon" (`:`) 😁

KtorZ: Whoops, not sure where my mind was.

rphair:
today's IO Dev Digest has linked back to this thread, calling for "community feedback" - https://mailchi.mp/iohk/dev-april-digest-675582

crptmppt (approved):

rphair (approved):
confirmed OK based on today's meeting

PR #264, opened 2022-05-18
CIP-0026 | subject properties return multiple claims

ehanoc:
This small change proposal makes so that there can be competing claims about a subject's property. Prevents hijacking of a property's by simply allowing the first value. Also for the metadata server client to decide (based on annotated signature), which claim it trust

### DIFF
--- a/CIP-0026/README.md
+++ b/CIP-0026/README.md
@@ -315,7 +315,7 @@ The metadata server SHOULD implement the following HTTP methods:
 GET /metadata/{subject}/property/{property name}
 ```
 
-This SHOULD return the property value for the given property name (if any) associated with the subject. This is returned as a single-entry JSON object whose key is the property name and return the complete JSON entry for that subject+name (including `value`, `sequenceNumber` and `signatures`).
+This SHOULD return the property values for the given property name (if any) associated with the subject. This is returned as a array of JSON objects whose key is the property name and return the complete JSON entries for that subject+name (including `value`, `sequenceNumber` and `signatures`).
 
 The metadata server SHOULD set the Content-Length header to allow clients to decide if they wish to download sizeable metadata.
 
@@ -329,7 +329,7 @@ This SHOULD return all the property names which are available for that subject (
 GET /metadata/{subject}
 ```
 
-This SHOULD return the full metadata object associated with this subject, including the subject and all properties associated with it. 
+This SHOULD return a array of the full metadata objects associated with this subject, including the subject and all properties associated with it. 
 
 ```
 POST /metadata/query


KtorZ:
I am not sure to follow your train of thoughts regarding "Highjacking a property"? What is the stated problem here exactly?

ehanoc:
> I am not sure to follow your train of thoughts regarding "Highjacking a property"? What is the stated problem here exactly?

You are correct; i'll re-phrase. The spirit of the change is that different parties might disagree on a particular claim and we want to return all the claims. It's then up to the client to choose where to anchor trust.

KtorZ:
I still don't get it :no_mouth: , or more exactly, how we could end up in such a situation. From the server's perspective, there's no ambiguity as for what is the most recent version of a metadata entry and as per the **Recommendations for Metadata Servers** section:

> 1. The server MUST only accept updates for metadata entries that have a higher sequence number than the previous entry.

Therefore, the server only record a strict sequence of metadata entries. Previous entries can't be modified. Are you maybe suggesting that the server, instead of returning only the latest entry, returns the whole history of entries for all known sequence numbers and let clients resolve possibly conflicting histories from different servers?

ehanoc:
I guess the recommendation should be expanded to make it clearer. 

The sequence numbers should remain working as suggested. The only difference to what i'm suggesting is that the servers instead of keeping track of a single claim / subject by some "entity" (author, developer, anyone... ). It can keep track of multiple claims and it's up to the client to use whatever mechanism of anchoring trust it's chooses too. A single entry doesn't offer the option to handle competing claims. 

Also in general terms, shouldn't metadata be created from any party, not just the "owner" / first-time-claimer? 

// example fetching multiple claims for a subject
```json
[ 
{ //party A
  "subject": "4c7bceca6b5739fe4c3fa02d0bc4007f",
  "isWorldRound": {
    "value": "true",
    "sequenceNumber": 0,
    "signatures": [
      {
        "signature": "939d14709a2f8d0eb5fd3f6499aa9c46",
        "publicKey": "90d120610b7e029257f2d5d26167cbda"
      }
    ]
  }
},
{ // party B
  "subject": "4c7bceca6b5739fe4c3fa02d0bc4007f",
  "isWorldRound": {
    "value": "false",
    "sequenceNumber": 0,
    "signatures": [
      {
        "signature": "050679dff79ff17354b3412a25ae0251",
        "publicKey": "10b32a9aea7ff0aff39c7cca07c1f940"
      }
    ]
  }
},
{ //party C
  "subject": "4c7bceca6b5739fe4c3fa02d0bc4007f",
  "isWorldRound": {
    "value": "true",
    "sequenceNumber": 1,
    "signatures": [
      {
        "signature": "c9361234013c8950d83ffd8931901952",
        "publicKey": "ae941363dd167289b4731bf0cf1e30dd"
      }
    ]
  }
},
]
```

Thoughts?

KtorZ:
I see @ehanoc, thanks for clarifying. So this is indeed to come as a mitigation for possibly "first-come first-serve" behavior where a malicious client may provide wrong (or competing) metadata on a particular entry. 

I imagine that for a given set of credentials, it would be sufficient to only return the highest sequence number; but return however all the potential version of a metadata given different credentials (and let the client make choice).

KtorZ (approved):
@rphair @SebastienGllmt  I was just going through the list of PR and saw that one which again got a bit lost in translation. I think that's a small update worth merging sooner than later.

rphair (approved):
thanks, I agree 😎

PR #266, opened 2022-05-19
Fix value for a0

SebastienGllmt:
You can see from the [current configuration](https://hydra.iohk.io/build/13695229/download/1/mainnet-shelley-genesis.json) that a0 is current 0.3

The value is correct in the JSON, but the markdown section had the wrong value


KtorZ (approved):
Indeed: 

https://github.com/input-output-hk/cardano-configurations/blob/master/network/mainnet/genesis/shelley.json#L25

rphair (approved):
confirmed easily enough 🤩

PR #267, opened 2022-05-24
CIP-0025: Added version 2 and cddl

alessandrokonrad:


CIP-0025/README.md
```diff
@@ -64,27 +64,90 @@ The structure allows for multiple token mints, also with different policies, in
...
+files_details = 
+  {
+    name : text,
+    mediaType : text,
+    src : text / [* text]
+  }
+
+metadata_details = 
+  {
+    name : text,
```
SebastienGllmt, line 118: We have to be a little careful with this definition since text here isn't the standard "text" in cbor. Instead, Cardano metadata enforces that all text or byte array has a maximum length. We should probably either define these as a special type that matches the same max length or explicitly allow people to use an array to combine chunks together

alessandrokonrad: Yeah, I just fixed it. `name` should definitely stay below 64 characters. The ones that can exceed 64 characters have the type `string / [* string]`.

KtorZ (approved):
:+1: Approving. 
@alessandrokonrad perhaps consider moving the CDDL as separate files so they can be linked / downloaded by tools / CI for validation & automation. 

Will leave the PR open for a bit, and we'll merge in the next editor bi-weekly at latest.

rphair (approved):
really easy to read with CDDL in separate files 🤓

SebastienGllmt (approved):
LGTM

PR #269, opened 2022-05-30
CIP-0056? | Treasury donations

SebastienGllmt:
Cardano's treasury is currently denominated purely in ADA and has no source of revenue other than the existing reward splitting mechanism implemented at the protocol level. This CIP intends to diversity the treasury holdings and increase protocol revenue to the treasury by allowing Catalyst-funded private ventures to both diversity and increase Cardano's treasury balance.

Not handled in this CIP is the legal implications of accepting donations:
- How to handle donations from questionable sources? This is a hard question for accounting-style chains, but I think for UTXO-based chains it's not as big of an issue because you can just ignore the UTXO entry
- Some countries have regulation regarding taking donations. For example, it's possible somebody donates money and then the treasury uses the money in a way that they disagree with. This is usually solved by having donors sign to terms agreeing they can't complain about how the money is spent, but this is of course harder to enforce on-chain
- If a US entity donates to the treasury and then the treasury is used to fund a Catalyst proposal from a sanctioned country, the US entity may, depending on the situation, be considered in violation of sanctions. My current understanding (I'm not a lawyer) is that as long as (1) Catalyst is not an entity actively with a history of funding sanctioned countries, (2) the US entity has no reason to believe their funds would be used this way, (3) the US entity voiced their objection to the use of funds in this way publicly and by voting against such a proposal, then they should have a decent argument

---

([rendered proposal from branch](https://github.com/cardano-foundation/CIPs/blob/treasury-donations/treasury-donation/README.md))

### DIFF
--- a/treasury-donation/README.md
+++ b/treasury-donation/README.md
@@ -11,7 +11,7 @@ License: CC-BY-4.0
 
 ## Abstract
 
-Cardano's treasury is currently denominated purely in ADA and has no source of revenue other than the existing reward splitting mechanism implemented at the protocol level. This CIP intends to diversity the treasury holdings and increase protocol revenue to the treasury.
+Cardano's treasury is currently denominated purely in ADA and has no source of revenue other than the existing reward splitting mechanism implemented at the protocol level. This CIP intends to diversify the treasury holdings and increase protocol revenue to the treasury.
 
 ## Motivation
 
--- a/treasury-donation/README.md
+++ b/treasury-donation/README.md
@@ -1,0 +1,54 @@
+---
+CIP: ??
+Title: Treasury donation
+Authors: Sebastien Guillemot <seba@dcspark.io>
+Comments-URI: TBD
+Status: Proposed
+Type: Standards
+Created: 2022-05-30
+License: CC-BY-4.0
+---
+
+## Abstract
+
+Cardano's treasury is currently denominated purely in ADA and has no source of revenue other than the existing reward splitting mechanism implemented at the protocol level. This CIP intends to diversity the treasury holdings and increase protocol revenue to the treasury.
+
+## Motivation
+
+Cardano currently spends 12.8M USD every 3 months (~25M ADA at the current price) on Catalyst. Additionally, there is discussion of increase the amount drawn from the treasury through initiatives like the DCF.
+
+However, there are two key points with the current system relevant to this CIP:
+
+1. **Catalyst funds are entirely denominated in ADA**. This is not ideal because Catalyst rewards are currently denominated in USD. It means that the amount of funds available can vary wildly between bear & bull markets and, in the worst case, could lead to Catalyst going bankrupt if the price of ADA falls below its obligations. You can read more about this problem [here](https://uncommoncore.co/a-new-mental-model-for-defi-treasuries/)
+2. **Treasury funds for private ventures does not necessarily lead to increased Catalyst revenue.**. To expand on this, a lot of the spending on Catalyst right now is for direct benefit of the protocol (marketing, open source tools, etc). However,  for fund 9, there is 7.8M USD allocated to products & integrations for Catalyst fund 9 which shows there is interest in funding efforts that only may only indirectly lead to protocol revenue (funding private ventures). Despite this, companies are [hesitant to take Catalyst rewards](https://twitter.com/blakelockbrown/status/1530968246332952576) because of this missing link in incentives.
+
+Notably, we can leverage point (2) to tackle point (1) by giving private ventures a way to pay it forward by contributing to treasury funds in a variety of (hopefully uncorrelated) assets.
+
+## Specification
+
+Reminder of the current system: at the time of writing, the Cardano treasury balance is not accumulated through on-chain transactions, but rather kept track of as part of the ledger state (similar to reward balances). To access treasury rewards, a special certificate called the MIR certificate is used to withdraw from the treasury balance. This means that treasury balances, just like reward addresses, only contain ADA (no tokens) and there is no way to send to the treasury address.
+
+This means that to allow companies to donate to the treasury, we need to define an on-chain address for the treasury. This can be done following the same format as the [upgradeable treasury smart contract](https://github.com/dcSpark/plutus-smart-contract-upgrade-dApp) from dcSpark.
+
+### Updating the treasury
+
+Cardano does not have an on-chain governance method for deciding protocol parameters. This missing functionality also makes it hard to decide on how to upgrade the treasury contract. As a temporary workaround similar to the current behavior, we could add MIR certificates to the Plutus context so that the contract can look at the context to decide if the contract can be upgraded.
+
+### Convincing projects to pitch in
+
+Once private ventures receive Catalyst funding, there is nothing to force them to actually donate back to the treasury after they've achieved market success. This should instead be encouraged at the social level and Catalyst proposals for these categories should be encouraged to define in their proposal whether they have any plans to donate funds back to the treasury.
+
+Additionally, the community should encourage projects to pitch in with assets that are as uncorrelated with ADA as possible (different baskets)
+
+### Compatibility with future treasury plans
+
+The upgradeability of the contract should make it easy to keep this system compatible with existing plans for the Cardano treasury. For clarity, I will detail some of the treasury plans and how this proposal is compatible with them:
+
+1. **Allowing Catalyst rounds in non-ADA tokens** (ex: COTI rewards for COTI proposals which has happened in the past). This proposal is compatible with this idea because projects could send their tokens to the treasury address. Additionally, it makes itt easier for this to happen because we could allow projects to put in 10K ADA (arbitrary number) as a fee to mint a project ID to the treasury address which can be redeemable once to add the project to Catalyst using the funds for that token present in the address. The fee of this mint would be used to both fund the main ADA treasury balance and to cover the cost of distributing rewards for the Catalyst challenge of the project.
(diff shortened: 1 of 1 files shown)


mach-diamond:
I love this idea, but I think it there are some preceding dependencies that need to be defined first. 

We have a 721 standard for digital assets, but I think businesses operating as DAOs deserve their own metadata standard along with some standardized smart contracts which would be important here. I've done a lot of preliminary research on that, but I still wouldn't be ready to pull the trigger on a draft CIP. (Or should I just post the idea with some supporting arguments at this point?)

Also playing into that, what trustless mechanism would be in place for a stake-holder in a revenue generating organization 
to receive their share of profits? 
I think in the perfect world, we have a smart contract for the full suite of treasury management where revenue is sent to this script address. Then a distribution is defined and voted upon to ensure expenses are covered, funds are set aside for internal investment, and finally a proportional dividend for remaining profits. 

And then on a different note, what mechanisms would be in place for quality control so the treasury doesn't get polluted?

anselmorei:
Probably out of scope but food for fought:

In this initial stage, where tooling and scaling solutions are critical, It could make sense to allocate a % of the treasury funds too a more centralized funding mechanism but a more efficient one - we currently have ~ 500M in the treasury with teams willing to build if properly funded.

This while the other % of the treasury is allocated to the current process of Project Catalyst so we can really build the infrastructure to have a truly decentralized governance system.

This would create a hybrid funding mechanism where teams with critical value prepositions could be funded fast and the project catalyst could evolve to create the foundations to the Voltaire era. 

The question is, which team could be responsible for this more centralized but more efficient process of funding? Probably the CF, IOG? I know @KtorZ is moving to CF with Dirk Hohndel, so probably he can enlighten us if there is any plans to use the Cardano treasury for future developments?

treasury-donation/README.md
```diff
@@ -0,0 +1,54 @@
...
+Comments-URI: TBD
+Status: Proposed
+Type: Standards
+Created: 2022-05-30
+License: CC-BY-4.0
+---
+
+## Abstract
+
+Cardano's treasury is currently denominated purely in ADA and has no source of revenue other than the existing reward splitting mechanism implemented at the protocol level. This CIP intends to diversity the treasury holdings and increase protocol revenue to the treasury.
```
rphair, line 14:
```suggestion
Cardano's treasury is currently denominated purely in ADA and has no source of revenue other than the existing reward splitting mechanism implemented at the protocol level. This CIP intends to diversify the treasury holdings and increase protocol revenue to the treasury.
```

treasury-donation/README.md
```diff
@@ -0,0 +1,54 @@
...
+
+1. **Allowing Catalyst rounds in non-ADA tokens** (ex: COTI rewards for COTI proposals which has happened in the past). This proposal is compatible with this idea because projects could send their tokens to the treasury address. Additionally, it makes itt easier for this to happen because we could allow projects to put in 10K ADA (arbitrary number) as a fee to mint a project ID to the treasury address which can be redeemable once to add the project to Catalyst using the funds for that token present in the address. The fee of this mint would be used to both fund the main ADA treasury balance and to cover the cost of distributing rewards for the Catalyst challenge of the project.
+2. **DCF**. DCF is a more complicated version of Catalyst meant for higher amounts with higher auditability on spending. DCF plans to have its own Plutus scripts, so there may be an overlap in the approach, but these DCF scripts could receive their initial funding from the treasury address so there should not be a compatibility issue.
```
rphair, line 48: The best source I could find on **DCF** was the video from last year's Cardano Summit.  If there's anything written on the subject it would help to link it here, so people can see how the ideas are compatible without having to go through the video. 😎

rphair (commented):
I think this is a brilliant plan 🤠

SebastienGllmt:
@anselmorei 
> It could make sense to allocate a % of the treasury funds too a more centralized funding mechanism but a more efficient one - we currently have ~ 500M in the treasury with teams willing to build if properly funded.

This is what the DCF is to a certain extent

@mach-diamond 
> I think in the perfect world, we have a smart contract for the full suite of treasury management where revenue is sent to this script address. Then a distribution is defined and voted upon to ensure expenses are covered, funds are set aside for internal investment, and finally a proportional dividend for remaining profits

The goal of this CIP is to start with a very simple script that is upgradeable. If we add too many features, we'll never end up with a v1 that everybody agrees on

@mach-diamond 
> And then on a different note, what mechanisms would be in place for quality control so the treasury doesn't get polluted?
This isn't really required because of the way the UTXO model works. If the treasury thinks a certain UTXO entry is spam, it never needs to spend it. This applies both to spam and to funds received from questionable source.

alexanderwatanabe:
> Some countries have regulation regarding taking donations. For example, it's possible somebody donates money and then the treasury uses the money in a way that they disagree with. This is usually solved by having donors sign to terms agreeing they can't complain about how the money is spent, but this is of course harder to enforce on-chain.

Perhaps obfuscating the Catalyst treasury's wallet addresses, and taking donations through a script address with appropriate T&C/disclaimers, which then forwards the funds on to appropriate addresses.

SebastienGllmt:
> > Some countries have regulation regarding taking donations. For example, it's possible somebody donates money and then the treasury uses the money in a way that they disagree with. This is usually solved by having donors sign to terms agreeing they can't complain about how the money is spent, but this is of course harder to enforce on-chain.
> 
> Perhaps this could be solved by obfuscating the Catalyst treasury's wallet addresses, and taking donations through a script address with appropriate T&C/disclaimers, which then forwards the funds on to appropriate addresses.

Yes, this is the standard solution in accounting-style chains. I don't see why this would be required for a UTXO chain since you could just decide not to spend a specific UTXO until some condition is met like they signed T&C

alexanderwatanabe:
> > > Some countries have regulation regarding taking donations. For example, it's possible somebody donates money and then the treasury uses the money in a way that they disagree with. This is usually solved by having donors sign to terms agreeing they can't complain about how the money is spent, but this is of course harder to enforce on-chain.
> > 
> > 
> > Perhaps this could be solved by obfuscating the Catalyst treasury's wallet addresses, and taking donations through a script address with appropriate T&C/disclaimers, which then forwards the funds on to appropriate addresses.
> 
> Yes, this is the standard solution in accounting-style chains. I don't see why this would be required for a UTXO chain since you could just decide not to spend a specific UTXO until some condition is met like they signed T&C

More considerate to the people making the donation, rather than leaving their funds in limbo they are explicitly being allocated. It could also be beneficial in that they could actually earmark it for specific catalyst initiatives/catagories.

SebastienGllmt:
> It could also be beneficial in that they could actually earmark it for specific catalyst initiatives/categories.

Probably this is best handled via the project mint mechanism specified in the "Allowing Catalyst rounds in non-ADA tokens" because the changes to support this feature would be the same we need for this functionality.

KtorZ:
Putting this back to draft as, per CIP-0084, changes proposed to the ledger must come with accompanying ledger rules modifications and formal specifications. This CIP is still well-articulated but lacks the necessary level of details that would make it a valid candidate.

perturbing:
Donating to the treasury is already possible, to do it

1. Register a stake key (requires a deposit of `>=2 Ada`)
2. Register a stake pool (requires a deposit of `>=500 Ada` which will be the donation).
4. Deregister the stake pool after `current epoch + 2`.
5. Deregister the stake key before the above epoch (get deposit of `>= 2 Ada` back).

The result will be that the deposit of `>= 500` Ada will not have a destination, and will be sent to the treasury. So all donations have to be at least 500 Ada (native assets can always be swapped for Ada at a dex)

Source: Ledger specs and https://iohk.zendesk.com/hc/en-us/articles/900001951926-Retiring-a-stake-pool

perturbing:
@rphair

1) The minpool cost of 500 Ada is minimum. I think you can donate as much in one go as you want, you can only claim 500 Ada back atm. This has to be checked though, as far as I understand, no ledger of who deposited how much is recorded, for every registration you just know that they at least deposited the minimum amount as per the parameters.

2) I shared this knowledge out just for technical inspiration :) So that anyone can donate if they'd like to.

rphair:
thanks @perturbing I appreciate your thoughts... I'd never thought of "treasury donation" in that way before, so this is a learning experience for me.  It would be good to know that so much bureaucratic work & ledger load... not to mention processes of pool analytics services carefully stabilised on a long term set size of 3000+ pools... might not be necessary for a large Treasury donation. 🤓

perturbing:
@rphair, I was wrong in my statement that 500 Ada is the minimum, the registration has to be exactly 500 Ada. I tested the behavior in the key registration case, you will get a value not conserved error if you try to deposit something apart from 2 Ada.

That said, one can easily register multiple pools in one transaction to donate more than 500 Ada. Though this is bound by the maximum transaction size. One has to check how many of those minimal registration one can fit in a transaction, not an ideal solution, but it works :)

Ryun1:
Hey @perturbing @rphair,

In the Conway Ledger era it appears such Ada donations are going to be introduced, see [draft Conway specification](https://github.com/input-output-hk/cardano-ledger/blob/5b90d156c2071a55095cc535483cc2da4eea01e6/eras/conway/test-suite/cddl-files/conway.cddl#L72-#L73).

Crypto2099:
@rphair @Ryun1 is this CIP still necessary given that there has been no movement or changes on it in several years and this has been implemented (at least for pure-ADA) in the Conway era?

Thoughts @SebastienGllmt @KtorZ ?

Ryun1:
Happy to close this one out for now,
can always reopen in the future

rphair:
I'm also OK with closing this, especially if the original proponents don't respond after a couple of business days (it's Sunday now).  I can also see an argument that swapping native tokens for ADA before donation to the Treasury (supported in Conway) would protect against volatility and possible conflicts of interest (i.e. Treasury governance being obligated to favour a certain token).

PR #271, opened 2022-06-03
CIP-27 Update Proposal: Multiple Royalty Addresses

NicholasMaselli:
As NFT Projects become more mainstream, projects may need to handle multiple royalty addresses for team members. This PR is created to propose the idea and potential implementation to the Cardano Community

### DIFF
--- a/CIP-0027/README.md
+++ b/CIP-0027/README.md
@@ -27,7 +27,7 @@ A new tag of 777 is proposed for this implementation.  The community guidelines 
 1) A brand new unused policy for implementation is required.
 2) The royalties tags are to be written to an unnamed token, using the policy to be used for the intended Cardano Assets.
 3) Only the first minted set of instructions will be honored.  Any future updates or rewrites will be ignored.  This prevents a Cardano Asset maker from changing the royalties at a future date.
-4) Within this created asset will be the metadata for royalties distributions.  It will use a tag of 777, and then have two tags to identify the percentage of future sales requested as a royalty, and the payment address to forward those royalties to.  Those tags will be "rate" and "addr" respectively.
+4) Within this created asset will be the metadata for royalties distributions.  It will use a tag of 777, have either an object or array depending on the number of royalty addresses, then have two tags to identify the percentage of future sales requested as a royalty, and the payment address to forward those royalties to.  Those tags will be "rate" and "addr" respectively.
 5) The "rate" key tag can be any floating point value from 0.0 to 1.0, to represent between 0 and 100 percent.  For example, a 12.5 percent royalty would be represented with "rate": "0.125".  Previous version 1.0 of this proposal used pct instead of rate.  Marketplaces to continue to honor legacy pct tag.
 6) The "addr" key tag can be a string value, or an array.  It is to include a single payment address.  By allowing for an array, the payment address can exceed the per line 64 character limitation.  This payment address could be part of a smart contract, which should allow for greater flexibility of royalties distributions, controlled by the asset creator.
 7) The royalty token must be minted prior to creating any assets off the policy.  All markets will be instructed to look only for the first minted asset on a policy, which would need to be the unnamed 777 token.
@@ -53,6 +53,23 @@ A new tag of 777 is proposed for this implementation.  The community guidelines 
 	}
 }
 
+## Example JSON with multiple addresses
+
+{
+	"777": [
+	{
+		"rate": "0.2",
+		"addr": "addr1v9nevxg9wunfck0gt7hpxuy0elnqygglme3u6l3nn5q5gnq5dc9un"
+	},
+	{
+		"rate": "0.12",
+		"addr": [
+			"addr1q8g3dv6ptkgsafh7k5muggrvfde2szzmc2mqkcxpxn7c63l9znc9e3xa82h",
+			"pf39scc37tcu9ggy0l89gy2f9r2lf7husfvu8wh"
+		]
+	}]
+}
+
 ## Process Flow
 1) Create policy for planned assets.
 2) Mint no name token with community standard royalties metadata.
@@ -61,7 +78,7 @@ A new tag of 777 is proposed for this implementation.  The community guidelines 
 
 ## Rationale
 
-By creating a new tag for the distinct purpose of royalties distributions, Cardano Asset makers, and Marketplaces can uniformly apply royalties to assets with predictable results.  By creating the instructions on a single, no name token, all marketplaces will know the correct location of the royalties asset, without having to further locate it.  By enforcing the requirement of honoring only the first mint, cardano asset buyers and owners can predict the future resale value of the assets in their possession.  The solution is scalable to any desired royalty percentage.  It is easy to work with this new standard, and does not require an in depth understanding of smart contracts.
+By creating a new tag for the distinct purpose of royalties distributions, Cardano Asset makers, and Marketplaces can uniformly apply royalties to assets with predictable results.  By creating the instructions on a single, no name token, all marketplaces will know the correct location of the royalties asset, without having to further locate it.  By enforcing the requirement of honoring only the first mint, cardano asset buyers and owners can predict the future resale value of the assets in their possession.  The solution is scalable to any desired royalty percentage and number of royalty addresses.  It is easy to work with this new standard, and does not require an in depth understanding of smart contracts.
 
 ## Change Log
 
@@ -71,8 +88,8 @@ By creating a new tag for the distinct purpose of royalties distributions, Carda
 ## Minor Revision - September 27th, 2021
 ## Minor Revision - November 8th, 2021
 ## Minor Revision - December 30th, 2021
-- Change CIP Type from Standards to Informational
-- Change CIP Status from Draft to Active
+## Major Revision - June 3rd, 2022
+- Adding multiple royalty address support proposal
 
 Copyright
 


CIP-0027/README.md
```diff
@@ -53,6 +53,23 @@ A new tag of 777 is proposed for this implementation.  The community guidelines
 	}
 }
 
+## Example JSON with multiple addresses
+
+{
+	"777": [
```
NicholasMaselli, line 59: It's possible we may want the value in the "777" object to always be an array. However, the current implementation is an object so marketplaces would need to be backwards compatible anyway

Crypto2099:
This was considered during initial discussions surrounding royalties on-chain and there is a multi-fold problem that arises when attempting to declare multiple royalty receive addresses:
1. Which address receives which amount of the royalties?
2. MinUTXO being 1 Ada would mean that no royalties can be paid unless each royalty recipient address receives at least 1 Ada meaning that the minimum sale price to trigger a royalty payout scales to the number of royalty recipient addresses.
3. Introducing multiple royalty payout addresses introduces additional complexity/validation in the smart contract for marketplaces, most likely leading to less adoption (and subsequently royalty payouts).

NicholasMaselli:
> This was considered during initial discussions surrounding royalties on-chain and there is a multi-fold problem that arises when attempting to declare multiple royalty receive addresses:
> 
> 1. Which address receives which amount of the royalties?
> 2. MinUTXO being 1 Ada would mean that no royalties can be paid unless each royalty recipient address receives at least 1 Ada meaning that the minimum sale price to trigger a royalty payout scales to the number of royalty recipient addresses.
> 3. Introducing multiple royalty payout addresses introduces additional complexity/validation in the smart contract for marketplaces, most likely leading to less adoption (and subsequently royalty payouts).

Yes there are definitely some challenges.

1: I think this PR does a decent job at distinguishing royalty percentages through the array method. Backwards compatibility could be issue however.

2: The MinUTXO being 1 Ada is the biggest roadblock to this currently. Perhaps in the future if / when this parameter decreases the idea would become more feasible.

3: This would increase SC complexity but with the Vasil Hardfork around the corner and reference scripts / inputs being included, I think this becomes more easier to accomplish. I would be interested to see what the Cardano Marketplaces have to say about SC complexity in this area

qunosteve:
I'm still brand new to smart contracts, so excuse my ignorance here.

What if marketplaces sent the ADA to the royalty address with the assetid (that was sold) in the metadata of the tx? Ideally the recipient address is a smart contract owned by the NFT project and they could then decide what to do with it, knowing which asset was sold.

One way a project could accomplish this is when they mint each nft, they can have a custom tag in the metadata of each token that provides info to their (the project's) royalty smart contract on the parameters on who and how much to pay out.

This would potentially appease both camps: It would cap the size of the tx on the marketplace end, but also give the project ultimate flexibility on distributing royalties.

NicholasMaselli:
> I'm still brand new to smart contracts, so excuse my ignorance on the terminology here.
> 
> What if marketplaces sent the ADA to the royalty address with the assetid (that was sold) in the metadata of the tx? Ideally the recipient address is a smart contract owned by the NFT project and they could then decide what to do with it, knowing which asset was sold.
> 
> One way a project could accomplish this is when they mint each nft, they can have a custom tag in the metadata of each unique token that provides info to their (the project's) royalty smart contract on the parameters on who and how much to pay out.
> 
> This would appease both camps: It would cap the size of the tx on the marketplace end, but also give the project ultimate flexibility on distributing royalties to their camp.

The issue with having royalty information on each NFT is that the transaction data would increase substantially since 10k +  NFTs will need to have the address data. Cardano currently uses a Royalty token for CIP-27 which has royalty information for the entire project. The issue is that this royalty token only supports 1 address.

It is possible that this address could be a smart contract, and as such this PR would be updated to have an empty datum sent along with the ada so that ada is not locked away forever in the contract. (A really nice solution would be if SC's were updated such that sending no datum has the same behavior as sending an empty datum but that is an extremely deep technical update that I don't think we are ready to consider yet)

Another issue with smart contracts for this is that they require another manual transaction for project owners to receive the ada. This could be fine but it is a limitation that the community would need to discuss.

This is why I ultimately came up with the proposal in this PR, using an array for royalty addresses rather than an object (with limitations such as MinUTXO).

rphair:
> 2: The MinUTXO being 1 Ada is the biggest roadblock to this currently. Perhaps in the future if / when this parameter decreases the idea would become more feasible.

@NicholasMaselli coincidentally a new CIP draft came onto the Cardano forum today, with a building rationale for lowering `minUTxOValue` from ≈1 ada.  The discussion so far has favoured *not* changing the long time value, so could you write a bit about the multiple-royalty use case there? https://forum.cardano.org/t/reduce-min-utxo/102527

KtorZ:
@NicholasMaselli reading through the proposal, I have mainly two questions / remarks:

1) It sounds like this update may quite substantially change the initial proposal; which suggests it'd be better off as a separate CIP that'd be extending the existing CIP-0027. Or, make sure to propose a change that remains backward compatible. 

2) If I understand correctly, the motivation for this change is to allow scheme where royalties are split amongst different owners. This seems like a perfect use-case for a script address / smart-contract! In its simplest form, you could imagine a multisig address that is collectively owned by all the owners so that, when withdrawing royalties into their own wallet, they all get to endorse the distribution. Or, a slightly more elaborate setup would involve a small Plutus contract that entirely defines how the royalties can be distributed. So, it still suffices to use a single address and by leveraging smart-contract, royalties can be managed in a quite clever way.

NicholasMaselli:
> > 2: The MinUTXO being 1 Ada is the biggest roadblock to this currently. Perhaps in the future if / when this parameter decreases the idea would become more feasible.
> 
> @NicholasMaselli coincidentally a new CIP draft came onto the Cardano forum today, with a building rationale for lowering `minUTxOValue` from ≈1 ada. The discussion so far has favoured _not_ changing the long time value, so could you write a bit about the multiple-royalty use case there? https://forum.cardano.org/t/reduce-min-utxo/102527

Yes I will post here as well, thank you for bringing this to my attention! I believe the initial rational for the 1 minUTXO was to make bot attacks impossible?

rphair:
> I believe the initial rational for the 1 minUTXO was to make bot attacks impossible?

maybe indirectly (so bots don't fill the ledger with frivolous entries), but the official reason is to constrain the growth of the ledger size (https://docs.cardano.org/native-tokens/minimum-ada-value-requirement):

> *the constraint protects the Cardano ledger from growing beyond a certain size.*

CIP-0027/README.md
```diff
@@ -53,6 +53,23 @@ A new tag of 777 is proposed for this implementation.  The community guidelines
...
+		"addr": "addr1v9nevxg9wunfck0gt7hpxuy0elnqygglme3u6l3nn5q5gnq5dc9un"
+	},
+	{
+		"rate": "0.12",
+		"addr": [
+			"addr1q8g3dv6ptkgsafh7k5muggrvfde2szzmc2mqkcxpxn7c63l9znc9e3xa82h",
+			"pf39scc37tcu9ggy0l89gy2f9r2lf7husfvu8wh"
+		]
+	}]
+}
```
KtorZ, line 71: Make sure to put that example in a proper blockquote (using \`\`\`) so that it renders properly. Example:

```json
{
	"777": [
	{
		"rate": "0.2",
		"addr": "addr1v9nevxg9wunfck0gt7hpxuy0elnqygglme3u6l3nn5q5gnq5dc9un"
	},
	{
		"rate": "0.12",
		"addr": [
			"addr1q8g3dv6ptkgsafh7k5muggrvfde2szzmc2mqkcxpxn7c63l9znc9e3xa82h",
			"pf39scc37tcu9ggy0l89gy2f9r2lf7husfvu8wh"
		]
	}]
}
```

KtorZ:
> @NicholasMaselli:  I would be perfectly happy making this a separate CIP if you think that is the better overall.

I do :smile: ! As I said, this feels more like a side-extension than an update of the current CIP. Especially because the current CIP already is in activity and use by a few marketplaces. So it is better to come up with a new proposal / specifications. If it was still a draft, that wouldn't matter as much.

> @NicholasMaselli:  it makes the processes much smoother from a trustless perspective

I don't understand how **trustless** comes in the picture here. Because, not using a smart-contract and relying on marketplaces to willfully abide by the royalty policy actually makes things more trustful than trustless :thinking: 

> @NicholasMaselli: This is an extra step which is negative from a UX perspective. [...] The smart contract requires a solid technical understanding to implement.

Fair point. I'd love to see that argument mentioned in the rationale once you move it in a new CIP :+1:

NicholasMaselli:
> > @NicholasMaselli:  I would be perfectly happy making this a separate CIP if you think that is the better overall.
> 
> I do 😄 ! As I said, this feels more like a side-extension than an update of the current CIP. Especially because the current CIP already is in activity and use by a few marketplaces. So it is better to come up with a new proposal / specifications. If it was still a draft, that wouldn't matter as much.
> 
> > @NicholasMaselli:  it makes the processes much smoother from a trustless perspective
> 
> I don't understand how **trustless** comes in the picture here. Because, not using a smart-contract and relying on marketplaces to willfully abide by the royalty policy actually makes things more trustful than trustless 🤔
> 
> > @NicholasMaselli: This is an extra step which is negative from a UX perspective. [...] The smart contract requires a solid technical understanding to implement.
> 
> Fair point. I'd love to see that argument mentioned in the rationale once you move it in a new CIP 👍

I think it makes sense for me a write a new CIP on this, now that I am thinking about it. It actually makes sense for both an array  implementation AND smart contract implementation to be added for maximum flexibility. I think perhaps we want to do this slowly though. And maybe start with an array where the max numbers of addresses can be 2 for now. (And we can gradually extend it later.) I will write a new CIP this weekend (maybe even an A and B CIP) for the implementation.

I do think UX is critically important to Cardano. When the next bull market hits in several years, users will navigate to the most intuitive chains. I think Native Asset NFTs are far more intuitive than other industry competitors, so I think our solutions in other areas should work intuitively as well.

PR #273, opened 2022-06-07
Update README's top-level tables CIPs & candidate CIPS.

KtorZ:


rphair (approved):
> Reviewed Proposals (as of 2022-04-07)
> Proposals Under Review (as of 2022-03-17)

Lists look great 😎 but don't the dates also need to be updated each time, to the date the PR was was submitted? (unless no changes in that section)

Also I would recommend moving the date out of the heading into a lesser heading below it. Github generates an anchored link to these sections & it would prevent that anchor from changing with every revision, e.g.:

https://github.com/cardano-foundation/CIPs/blob/README-update/README.md#proposals-under-review-as-of-2022-03-17

If separated like this:
```
### Proposals Under Review
#### as of 2022-03-17
```
... then the link to the greater heading would be the same across document revisions.

KtorZ:
Doh! Good catch @rphair and, I concur with your suggestion. Will do.

rphair (approved):

PR #274, opened 2022-06-09
CIP-0059? | Terminology Surrounding Core Features

JaredCorduan:
This informational CIP seeks to clarify the language around groups of features. At the very least, it provides some history.

### DIFF
--- a/CIP-namesSurroundingFeatures/feature-table.md
+++ b/CIP-namesSurroundingFeatures/feature-table.md
@@ -1,0 +1,12 @@
+# Cardano Features
+
+| Date    | Phase   | Era     | Slot Number | Epoch Number | Protocol Version | Ledger Protocol | Consensus Mechanism | Notes              |
+| ------- | ------- | -------:| -----------:| ------------:| ----------------:| ---------------:| -------------------:| ------------------:|
+| 2017/09 | Byron   | Byron   | 0           | 0            | 0,0              | -               | Ouroboros Classic   |                    |
+| 2020/02 | Byron   | Byron   | 3801600     | 176          | 1,0              | -               | Ouroboros BFT       |                    |
+| 2020/07 | Shelley | Shelley | 4492800     | 208          | 2,0              | TPraos          | Ouroboros Praos     |                    |
+| 2020/12 | Goguen  | Allegra | 16588800    | 236          | 3,0              | TPraos          | Ouroboros Praos     |                    |
+| 2021/03 | Goguen  | Mary    | 23068800    | 251          | 4,0              | TPraos          | Ouroboros Praos     |                    |
+| 2021/09 | Goguen  | Alonzo  | 39916975    | 290          | 5,0              | TPraos          | Ouroboros Praos     |                    |
+| 2021/10 | Goguen  | Alonzo  | 43372972    | 298          | 6,0              | TPraos          | Ouroboros Praos     | intra-era hardfork |
+|         | Goguen  | Babbage |             |              | 7,0              |  Praos          | Ouroboros Praos     | Vasil HF           |
--- a/CIP-0059/README.md
+++ b/CIP-0059/README.md
@@ -1,8 +1,8 @@
 ---
-CIP: ?
-Title: Names surrounding features
+CIP: 59
+Title: Terminology Surrounding Core Features
 Authors: Jared Corduan <jared.corduan@iohk.io>
-Status: Draft
+Status: Active
 Type: Informational
 Created: 2022-06-09
 License: CC-BY-4.0
--- a/CIP-namesSurroundingFeatures/README.md
+++ b/CIP-namesSurroundingFeatures/README.md
@@ -53,8 +53,6 @@ an intra-era hard fork.
 This is an implementation detail which involves bumping the major protocol version
 but not creating a new ledger era.
 The Alonzo era experienced an intra-era hard fork when going from major protocol version 5 to 6.
-See this [table](https://github.com/input-output-hk/cardano-ledger/wiki/First-Block-of-Each-Era)
-in the cardano-ledger repository's wiki.
 
 Yet another complication stems from the named releases.
 We chose to honor the late Cardano community member and Bulgarian mathematician Vasil Dabov
@@ -73,6 +71,8 @@ how block bodies are validated (era).
 The Shelley era used the "transitional Praos" protocol (or TPraos for short).
 It consisted of Praos together with a transition system to move away from Ouroboros-BFT.
 The Babbage era replaced TPraos with Praos.
+
+A table of all the features, as of the time this CIP was submitted, can be found [here](feature-table.md).
 
 ## Specification
 
--- a/CIP-namesSurroundingFeatures/README.md
+++ b/CIP-namesSurroundingFeatures/README.md
@@ -1,0 +1,103 @@
+---
+CIP: ?
+Title: Names surrounding features
+Authors: Jared Corduan <jared.corduan@iohk.io>
+Status: Draft
+Type: Informational
+Created: 2022-06-09
+License: CC-BY-4.0
+---
+
+## Simple Summary / Abstract
+
+This CIP seeks to clarify the language around groups of features.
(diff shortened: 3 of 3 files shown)


rphair:
To me this seems logically complete, with nothing extraneous or left out.  I think this will help a lot of people like me who've been bewildered by the inconsistent nomenclature.

Because of these haphazard beginnings maybe it would help to include a "timeline" table (like your [ledger era table](https://github.com/input-output-hk/cardano-ledger/wiki/First-Block-of-Each-Era) but more complete) of all phases, ledger eras, intra-era hardforks, and changes to the consensus mechanism & ledger protocol so far: since if this proposal is adopted it will make a lot of the writing around them incorrect retroactively.

I originally dismissed this idea because your proposed nomenclature is used throughout your own document, especially in the first big paragraph so it already includes most of the historical examples with the updated terms, in mostly chronological order.

But new SPOs, marketing & technical writers, and bloggers / vloggers would need something to refer to more easily.  Generally these won't have the technical experience & presence of mind to line up all the proposed terms with things they've experienced.  So I do think a timeline of all events characterised by those terms— from genesis up through the planned hard fork release dates before the end of 2022— would help the Cardano community and those beyond it.

Note I don't think such a timeline would need to be updated going forward in order to remain useful, especially since this document is about the nomenclature itself and not the history.  For that reason it might also be better, if including a timeline, to have it on a separate page.

JaredCorduan:
Thank you @rphair , that is great feedback! I think your idea for an expanded table is fantastic. I'm not sure exactly what you mean by using a separate page, though. A separate markdown file, to live in this repository? A wiki page on the `cardano-node` repository?

I think you have something like this in mind (this is totally off the cuff, I forget the details around the Byron phase, etc).

| Phase     | Era     | Slot Number | Epoch Number | Protocol Version | Ledger Protocol | Consensus Mechanism |
| --------| -----------:| ------------:| ----------------:| ----------------:| ----------------:| -------------:|
| Byron | Byron  |     0 |          0 | 1?              | - | Ouroboros Classic |
| Byron | Byron  |     ? |          ? | ?              | - | Ouroboros BFT |
| Shelley | Shelley |     4492800 |          208 | 2,0              | TPraos | Ouroboros Praos |
| Goguen | Allegra |    16588800 |          236 | 3,0              | TPraos | Ouroboros Praos |
| Goguen | Mary    |    23068800 |          251 | 4,0              | TPraos | Ouroboros Praos |
| Goguen | Alonzo  |    39916975 |          290 | 5,0              | TPraos | Ouroboros Praos |
| Goguen | Alonzo |    43372972 |          298 | 6,0              | TPraos | Ouroboros Praos |
| Goguen | Babbage  |   ?  |       ?    | 7,0              |  Praos | Ouroboros Praos |

rphair:
> A separate markdown file, to live in this repository?

yes, that's what I meant... I think a relative link to a file in the same directory as this CIP README will also work on `cips.cardano.org`.  As expected this has information going beyond the scope of the "ledger" so it makes more sense to include it in this CIP than replacing your earlier ledger table, unless you think otherwise.

The table itself is exactly what I was hoping for 🤩 although from other people's point of view it would be useful to have a Date column so they can line it up with other communications from these periods. (So people focus on the most relevant detail, which is the epoch / slot number, maybe it would be best to have an imprecise date: just month & year.)

JaredCorduan:
> yes, that's what I meant... I think a relative link to a file in the same directory as this CIP README will also work on `cips.cardano.org`. As expected this has information going beyond the scope of the "ledger" so it makes more sense to include it in this CIP than replacing your earlier ledger table, unless you think otherwise.

That sounds great to me! I'll need to do some digging to get all these details, but I agree that it sounds very helpful.

> Date column ... just month & year.

That sounds great too!

JaredCorduan:
(I updated the table to see what it would look like with dates and a notes column)

rphair (approved):
It all looks great, including the table 😎

I do think this will be an important reference document for writers & so also would like it to be easily found by nontechnical people looking for this subject.  I don't know if the title _Names surrounding features_ will stand out from a list of CIP titles in that regard (partly because of the nonliteral use of the word "surrounding"), whereas _Names for Cardano episodic features_ would be more literal and specific.

But I also don't have a common synonym for "episodic" and any common synonym for "episodic feature" would leave us right back with the word "era" whose definition isn't clarified until inside the document 🤪 ... so I can't pin this down perfectly & hope the group as a whole can come up with a real winner of a title.

KtorZ:
Shouldn't `Babbage` be part of `Basho` though :thinking: ?

JaredCorduan:
> It all looks great, including the table sunglasses

:tada:

> ... _Names for Cardano episodic features_ 

I like that better! It somehow doesn't bother me that the "name for the names" isn't defined :laughing: . I'm happy to got with whatever title folks think is the most helpful.

---

> Shouldn't Babbage be part of Basho though thinking ?

I'd never though of it that way. :thinking:  I was thinking about it as optimizations and ergonomics improvements for Plutus. I think of things like sidechains and Hydra as squarely within Basho.

KtorZ:
Hydra is layer 2, so arguably, it'll never be part of any changes carried on the core protocol. 😊

I'd say that any update that relates to performance / scalability on the layer 1 is part of Basho. This includes Babbage's Plutus improvements, pipelining, and Leios later on.

JaredCorduan:
> I'd say that any update that relates to performance / scalability on the layer 1 is part of Basho. This includes Babbage's Plutus improvements, pipelining, and Leios later on.

@dcoutts do you have an opinion on this? Are we considering the Babbage era a part of Goguen or Basho?

CIP-namesSurroundingFeatures/feature-table.md
```diff
@@ -0,0 +1,12 @@
+# Cardano Features
+
+| Date    | Phase   | Era     | Slot Number | Epoch Number | Protocol Version | Ledger Protocol | Consensus Mechanism | Notes              |
```
rdlrt, line 3: Might be worth mentioning:
- Protocol version is not the same as [Network Protocol Version](https://github.com/input-output-hk/ouroboros-network/blob/master/ouroboros-consensus-shelley/src/Ouroboros/Consensus/Shelley/Ledger/NetworkProtocolVersion.hs), which is at higher index than Protocol version referred by Ledger)? Not sure if it's something that can be mapped in this table, probably not because it's from networking layer between nodes?
- Perhaps a condition with link at the bottom to clarify about Protocol Versions in block header signify not the Protocol Version itself, but the maximum supported ledger-protocol-version for node?

JaredCorduan: * indeed, the "protocol version" is separate from the node-to-node and node-to-client versions. I'm happy to make a note of this, but I'd need help lining the node-to-x versions with this table (and indeed, it may not even line up at all).
* will do!

JaredCorduan: @rdlrt I added a commit to address these two points.

SebastienGllmt (approved):
Makes sense to me

KtorZ:
@JaredCorduan We'll move this PR to review/last-check in the next CIP meeting (cc @mangelsjover). In the meantime, may you:

- Number the proposal as `59` 
- Rename the main file as `README.md` 
- Move the files into a folder named `CIP-0059` 
- Change the status to `Active` (we'll discuss that in the meeting, but this being an informational CIP pushed by the core to clarify existing terminology, I believe it can go as `Active` right away).

And, more of a personal suggestion regarding the title: what do you think of: _Terminology Surrounding Core Features_
(stressing the 'Core' part, which adds a bit of context for when the CIP will be listed in other places).

JaredCorduan:
@KtorZ the main file is already named `README.md`. did you want the table moved, or was this just a checklist that I happened to have already completed one item? :sweat_smile:

KtorZ:
It was more of a general checklist, but I genuinely thought it wasn't named README hence why I mentioned it. My bad :no_mouth:

KtorZ (approved):

PR #282, opened 2022-06-20
CIP Editors Meetings Notes #46.md

mangelsjover:


BiweeklyMeetings/2022-06-07.md
```diff
@@ -0,0 +1,370 @@
...
+
+**Matthias** -: I mean, the situation we are in today was kind of already included. So there is probably better to do but whether CIP 50 is going in that direction, it's also good question and it's not clear right now. We don't have any quite clear proof that it is actually going in the right direction. So far it seems to be more of in sentiment than an actual well calculations or simulation of that new theory. Maybe that's something which would be worth trying on the site network. But importantly, we don't really have that.
+
+**Matthias** -: And the stake on the test net would be for instance, much more different because it's all free money and people don't have the same behavior, therefore. So, yeah, I don't know. We can try to bring IOG back again, see what's the answer and keep to have short update on that every week or every two weeks.
+
+**M. Àngels** -:  Okay. So we are going to keep it through for next meeting agenda?
+
+**Matthias** -: I would say for now, yes. And let's see, we probably won't get any update from put outputs before the Hard Fork. Because I guess all minds are focused on that, but we might a bit later.
```
rphair, line 267:
```suggestion
**Matthias** -: I would say for now, yes. And let's see, we probably won't get any update from Input Output before the Hard Fork. Because I guess all minds are focused on that, but we might a bit later.
```

BiweeklyMeetings/2022-06-07.md
```diff
@@ -0,0 +1,370 @@
...
+
+**Matthias** -: Okay. Sorry for that.
+
+**M. Àngels** -:  No problem. Do you want to make a brief summary of CIP 267?
+
+**Matthias** -: CIP-0025: Added version 2 and cddl #267
+
+**Matthias** -: No, sorry. Yeah. Pull Request 267. Yeah. So this is a iteration on CIP 25, right? Which was the NFT metadata standard we ha not exposed, but mentioned a few points that could be improved last time. And the idea was to make a version two that would address those points or this is basically it. Okay. What I see here. Yeah. I think one thing we discussed and that's also Sebastian maybe mentioning it, or. So it would be good to have the CGTL as a separate files as always, but at least now we've got a clear onchain CDDL specifications of the metadata for the CIP.
+
+**Matthias** -: I think it is mainly compatible version one in the sense that it embeds the same informations, but there is a slight difference in the way the asset names gets included. For instance, right now it used to be just playing UTF-8 text string, whereas this new one uses actually raw binary bites through CBO coding. So since this is entry data any way, that makes much more sense.
```
rphair, line 293:
```suggestion
**Matthias** -: I think it is mainly compatible version one in the sense that it embeds the same informations, but there is a slight difference in the way the asset names gets included. For instance, right now it used to be just playing UTF-8 text string, whereas this new one uses actually raw binary bytes through CBOR encoding. So since this is entry data any way, that makes much more sense.
```

rphair (approved):
some other grammatical changes but most can be recognisable from the context.

One idea I had is that when Meeting Notes are officially merged it will help for some contended PR's to back-link those sections of the minutes to the PR as a GitHub comment in the PR.  I was tempted to do this in a couple of cases even without these minutes being fully approved.

SebastienGllmt (approved):

PR #285, opened 2022-07-02
fix CIP-0040 table omissions

rphair:
forgot to make some changes to the line for CIP 40 after it was merged (https://github.com/cardano-foundation/CIPs/pull/216#issuecomment-1172955353)

PR #286, opened 2022-07-02
fix CIP-0040 table omissions (try 2)

rphair:
Trying again to fix something accidentally left out of https://github.com/cardano-foundation/CIPs/pull/216#issuecomment-1172955353 without the 18 month commit history as seen in https://github.com/cardano-foundation/CIPs/pull/285.


rphair:
Well, here we go again.  I'm very sorry but I have tried my best to create this PR without the enormous commit history.  Nothing I do will squash it.  I have even fixed this problem before but the solution as I understood it isn't working in this case.

Sorry about the mess.  I would appreciate some advice how to avoid this but if so then please be specific and as literal as possible.  Otherwise I will work it out on my own eventually.

It's my understanding that the local branch I'm working with has been fully caught up with `master` upstream.  My fork is also showing all the latest changes.  I need to understand why this is happening to get fully functional here. 😕

rphair:
this was necessary to manhandle my fork back into place (I will keep going over this until I understand exactly why it was required in my case):
```
git reset --hard upstream/master
git push --force
```

PR #293, opened 2022-07-08
CIP-0055 correction: constant overhead in babbageMinUTxOValue

JaredCorduan:
CIP-55 was an information CIP regarding the Babbage era protocol parameters. In it, it was explained how the new parameter
`coinsPerUTxOByte` is used. The explanation contained a mistake, namely not accounting for the constant overhead.

Thank you @gitmachtl for pointing this out! https://github.com/cardano-foundation/CIPs/pull/265#issuecomment-1179338845

### DIFF
--- a/CIP-0055/README.md
+++ b/CIP-0055/README.md
@@ -73,10 +73,11 @@ the current value of `coinsPerUTxOWord` will be converted to
 In the Babbage era, unspent transaction outputs will be required to contain _at least_
 
 ```
-|serialized_output| * coinsPerUTxOByte
+160 + |serialized_output| * coinsPerUTxOByte
 ```
 
-many lovelace.
+many lovelace. The constant overhead of 160 bytes accounts for the transaction input
+and the entry in the UTxO map data structure (20 words * 8 bytes).
 
 ## Rationale
 


rphair (approved):
I can't confirm this technically, but from the other thread this seems like a well considered & agreed-upon change.

JaredCorduan:
> I can't confirm this technically, but from the other thread this seems like a well considered & agreed-upon change.

thanks @rphair ! @KtorZ will be able to confirm that the change:
https://github.com/input-output-hk/cardano-ledger/blob/2505b7f103c78ee5a230b3e302e31d9607fffbd5/eras/babbage/impl/src/Cardano/Ledger/Babbage/Rules/Utxo.hs#L247-L265

KtorZ (approved):

gitmachtl:
@JaredCorduan i figured out, that for a mary/alonzo transaction in babbage era you get along with an overhead of only 158 bytes. what is causing the 2 extra bytes for an babbage formated transaction? is it the change in the legacy/babbage transaction output (https://github.com/input-output-hk/cardano-ledger/blob/2505b7f103c78ee5a230b3e302e31d9607fffbd5/eras/babbage/test-suite/cddl-files/babbage.cddl#L77-L88) ?

PR #294, opened 2022-07-12
CIP-0066 | NFT Identity

Padierfind:
The current CIP ID is set to 066, but this can be changed to any arbitrary number.


shawnim:
Excellent proposal!
Well designed and presented.
Two small nitpicks:
Line 44 "platofrms" -> "platforms"
Line 215 "The focous of this standard is on easy-of-use & flexibility." -> "The focus of this standard is on ease-of-use & flexibility."
Nice work!
I think this will benefit both creators and NFT platforms.

rphair:
thanks; looking forward to the pending material 😎

BadrulAlomIOG:
Feedback based on Zoom call.
1. Think the unnamed token should be called something like "Collection Metadata token" to make it easier to understand

2. Need to safeguard against the following scenario:
- Bad actor creates and NFT colleciton with their actual details
- Later they go an modify the social media data attached to their unnamed token to a fake social media profile
- This reduces problem down to still having to verify the DID or policy ID itself. (Capturing the actual metadata like this may be redundant?).

3. Does the solution require Twitter to agree to be an credential validator? If so this would limit adoption and is it better than the social media account owner publicly stating their public ID/policy key

Padierfind:
Okay, I added the images explaining the structure & changed the title from unnamed token to "Collection Token" like @BadrulAlomIOG suggested. I think that's a very good idea!

@BadrulAlomIOG regarding your third point: No, it does not require Twitter to be a credential validator. Twitter has a login function through the OAuth standard, that's all which is required for issuing the identity.

CIP-066/README.md
```diff
@@ -0,0 +1,426 @@
+---
+CIP: 066
+Title: NFT Identity
```
rphair, line 3:
```suggestion
Title: NFT Project Identity
```
Making the title more specific will be helpful, since _NFT Identity_ is too general for people to know what it's really about.  _NFT Project Identity_ would match what you describe in the Abstract & Motivation, or perhaps _NFT Collection Identity_ if that's not too specific.

CIP-066/README.md
```diff
@@ -0,0 +1,426 @@
...
+
+Each of these data points stores information about a direct connection between the NFT collection and some sort of identity medium. This identity medium could be a scoial account, a KYC process, a phone number or anything else where an identity is directly linked to an account or something similar.
+
+The data points connected to the NFT collection are publicly visible and can be audited by anyone. Either manually or programmatically.
+
+The creator can choose which identity data points he wants to connect to the collection and the auditor can then choose if he trusts these data points enough to interact with the colleciton or not.
+
+## Practical Example
+
+Let's imagine Paris Hilton wants to release an NFT collection on Cardano.
```
rphair, line 33: Use of the particular celebrity Paris Hilton doesn't add anything to the substance of the CIP.  If the "story" were a bit shorter and maybe used a more generic term "celebrity" then there would be room in the reader's attention for at least one other use case: that of the unknown artist who's less followed on social media & has less likelihood of being authenticated through a conventional platform.

rphair (commented):
I believe this is promising & worth discussion with some other NFT projects & platforms that are currently debating a new CIP here (https://github.com/cardano-foundation/CIPs/pull/299).  As soon as some preliminaries are met I would be happy to invite them over to review this CIP if they haven't discovered it already in the meantime. 😎

Maetti79:
> Feedback based on Zoom call.
> 
> 1. Think the unnamed token should be called something like "Collection Metadata token" to make it easier to understand
> 2. Need to safeguard against the following scenario:
> 
> * Bad actor creates and NFT colleciton with their actual details
> * Later they go an modify the social media data attached to their unnamed token to a fake social media profile
This is not possible, you need to update your identity using the Issuer Service you used in the first place, in order to have the data Signed again.
> * This reduces problem down to still having to verify the DID or policy ID itself. (Capturing the actual metadata like this may be redundant?).
> 
> 3. Does the solution require Twitter to agree to be an credential validator? If so this would limit adoption and is it better than the social media account owner publicly stating their public ID/policy key

rodolfomiranda:
It's a good idea to tie a DID to an NFT, however it is not specified how those social identities can be linked to the DID. I'd assume it will be some Verifiable Credentials issued by some party to that DID. If  that's the case, the issuer of the VC became the root of trust as are the NFT platforms. I'm not understanding how you solve the problem of the centralization effort. I'm probably missing something.
Or is it the stored data in the DID as shown on the example?. DIDs are controlled by the owner, not by a central identity provider, that's one of the main principles of SSI. It would be good if you can add definitions for  _Identity Provider_ and _Identity Data Points_ and how these can be a trusted sources.

CIP-066/README.md
phillewis: I would even suggest _NFT Verification Standard_ as it is the verifier's identity being used, not the NFT's identity.

phillewis:
As per my response on Twitter, I would also (as per @rodolfomiranda) suggest considering Verifiable Credentials.  My suggestion on Twitter was to explore using VCs in place of the proposed Identity Token.  By using an existing standard it would open the Cardano NFT space to cross chain interoperability.

I have proposed in the past that we should even consider using DIDs as NFTs.  While that would be a bigger change, linking the verifier's DID to the NFT via a VC could act as a stepping stone towards a true cross chain standard for NFTs.

sean118:
What's the benefit of having a separate "Collection Token" instead of simply adding a DID reference or identity token reference field directly to the existing NFT metadata?

CIP-066/README.md
```diff
@@ -0,0 +1,426 @@
...
+
+The verification process of the authenticity of NFT projects is currently a centralized effort, that is completely in the hands of the NFT platforms. 
+
+Most of these platforms create a centralized database of verified projects and have arbitrary rules that define which NFTs get verified on the respective platform. Not only is this not scalable because it relies on a case-to-case verification process often done manually and requires the NFT creators to go through the verification process on an increasing amount of different platforms, but it is also highly centralized and doesn't utilize a global cross-platform verification method.
+
+## High Level Overview
+
+By connecting an NFT collection to one or multiple Digital Identifiers, the creator of the NFT collection has the option to attach verification data points to the collection. 
+
+Each of these data points stores information about a direct connection between the NFT collection and some sort of identity medium. This identity medium could be a scoial account, a KYC process, a phone number or anything else where an identity is directly linked to an account or something similar.
```
KtorZ, line 25:
```suggestion
Each of these data points stores information about a direct connection between the NFT collection and some sort of identity medium. This identity medium could be a social account, a KYC process, a phone number or anything else where an identity is directly linked to an account or something similar.
```

CIP-066/README.md
KtorZ: Agreed. Alternatively, could pick a nameless name like _John Doe_ / _Jane Doe_ for the sake of the example.

CIP-066/README.md
```diff
@@ -0,0 +1,426 @@
...
+    "type": "Ed25519VerificationKey2020",
+    "controller": "did:example:123456789abcdefghi",
+    "publicKeyMultibase": "zH3C2AVvLMv6gmMNam3uVAjZpfkcJCwDwnZn6z3wXmqPV"
+  }]
+}
+```
+
+#### 2. Collection Token
+
+This is an unnamed token, as first described in the [CIP-0027](hhttps://github.com/cardano-foundation/CIPs/tree/master/CIP-0027).
```
KtorZ, line 106:
```suggestion
This is an unnamed token, as first described in the [CIP-0027](https://github.com/cardano-foundation/CIPs/tree/master/CIP-0027).
```

CIP-066/README.md
```diff
@@ -0,0 +1,426 @@
...
+
+It stores information about the location of either the DID or the Identity Token in the metadata of its minting transaction.
+
+```
+{
+  "725": {
```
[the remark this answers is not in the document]
KtorZ: Agreed. Ideally, we make it a self-contained file separate from the rest (e.g. `725.cddl`)

CIP-066/README.md
```diff
@@ -0,0 +1,426 @@
...
+![NFT Identity Structure Simple](./NFTIdentityDiagramSimple.jpg)
+
+Version 2: Complex & more flexible
+
+![NFT Identity Structure Complex](./NFTIdentityDiagramComplex.jpg)
+
+#### 1. DID
+
+This is an example DID taken from the official [W3C specification](https://www.w3.org/TR/did-core/#did-syntaxg).
+To keep the scope reasonable, this CIP will not explain the DID specification deeper. Please refer to the W3C specifation as linked above.
```
KtorZ, line 81:
```suggestion
To keep the scope reasonable, this CIP will not explain the DID specification deeper. Please refer to the W3C specification as linked above.
```

CIP-066/README.md
```diff
@@ -0,0 +1,426 @@
...
+
+
+### Structure
+
+There are three different key components to this structure.
+
+| Component                   | description                                                                                                                |
+| --------------------------- | -------------------------------------------------------------------------------------------------------------------------- |
+| DID                         | The actual Decentralized Identity Document where the information about the connected identity data points is stored.       |
+| Collection Token            | This token is minted under the policy ID of the NFT collection. It defines the connecton to the DID or the Identity Token. |
```
KtorZ, line 64:
```suggestion
| Collection Token            | This token is minted under the policy ID of the NFT collection. It defines the connection to the DID or the Identity Token. |
```

CIP-066/README.md
```diff
@@ -0,0 +1,426 @@
...
+| src                         | A link to the DID file or an asset_fingerprint to an Identity Token | no       |
+| other_properties            | Extensions of the standard                                          | yes      |
+| version                     | Version of the NFT Identity standard                                | no       |
+
+
+
+#### 3. Identity Token
+
+This token can be created under a seperate policy ID, so that it can be used for multiple collections at once.
+The policy can also be left unlocked in case the identity information requires changes at some later date.
```
KtorZ, line 148: How is this "token" defined and updated in practice? Is it assumed that the _last minting transaction_ would prevail? In which case, are previous identity tokens meant to be burnt?

CIP-066/README.md
```diff
@@ -0,0 +1,426 @@
...
+#### 3. Identity Token
+
+This token can be created under a seperate policy ID, so that it can be used for multiple collections at once.
+The policy can also be left unlocked in case the identity information requires changes at some later date. 
+
+```
+{
+  "725": {
+    "<policy_id>": {
+        "<collection_name>": {
```
KtorZ, line 154: How is the `collect_name` defined? Is there any restrictions beyond those of the protocol? Can an identity token's metadata carry multiple collections?

CIP-066/README.md
```diff
@@ -0,0 +1,426 @@
...
+
+## Cross-Chain Compatibility
+
+To make this standard cross-chain compatible it is possible to set the value of the src parameter under files to a unique identifier on another blockchain, where the DID might be located.
+
+## Backwards Compatibility
+
+To keep NFT metadata compatible with changes coming up in the future, we use the **`version`** property.
+
+## Rationale
```
KtorZ, line 407: Some unanswered questions in the document that would deserve an explanation in this section:

- Why bother with two approaches (collection token vs identity token) instead of a single one. It seems that one is clearly a superset of the other, and thus, make the other redundant. If the claim is about offering a simple alternative for those who don't want to go the "complex" way then I'd argue that this is misleading at most as, most implementations will likely end up implementing both making the system ultimately more complex than if there was only one generic and capable way. 

- How does exactly are the metadata tied to an NFT collection (beyond the policy id)? Can anyone publish identity metadata at any time after the mint? Does this has to happen during the mint? If so, then, what's the justification w.r.t the current metadata spoofing issues affecting CIP-0025. 

- Why adopting this particular format? It seems like this is mainly backed by the W3C Recommendation but this is only vaguely mentioned in the specification. If this W3C recommendation is meant to become an industry-grade standard then it'd be nice to highlight this in the rationale section.

CIP-066/README.md
```diff
@@ -0,0 +1,426 @@
...
+The focus of this standard is on ease-of-use & flexibility. Using this standard it is possible to reference DIDs not only on Cardano but at any arbitrary location, even cross-chain. It is also relatively simple to understand and implement because it borrows heavily from pre-existing metadata standards.
+
+## Path to Active
+
+[IAMX](https://iamx.id/) and [NMKR](https://www.nmkr.io/) have created a reference implementation of the standard that can be viewed [here](https://cardanoscan.io/transaction/608dd4fba7df5c523a1dfdca8d48fd1d69e7e1995a342ceb01c8355c62a0845b?tab=metadata).
+
+[IAMX](https://iamx.id/) will be creating a tool to create the DIDs that will be integrated natively into [NMKR](https://www.nmkr.io/) studio.
+
+By displaying this solution at the Catalyst Guild Hall we hope to reach more creators that are willing to integrate this solution.
```
KtorZ, line 416: If a reference implementation exist, I'd strongly suggest to include some test vectors as examples in this section to help other implementations to also happen. As mentioned also in some other comments, the data is ultimately stored on-chain as binary (CBOR). Thus, test vectors should include **at least** a CBOR representation of the metadata, and possibly a JSON / YAML one for human-readability.

CIP-066/README.md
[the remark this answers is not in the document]
SebastienGllmt: Ideally this CIP would be split into two CIPs: one that proposes the double-indirection scheme for minting policies introduced in this CIP (and have it be as a competitor to CIP68) and one that leverages the first CIP for an identity standard

rphair (commented):
The issue of the CIP title came up again in the CIP meeting # 51 today.  Although some words that could be added like [NFT...] {Collection, Issuer, {Verifier, Verification}} indicating the agent or function, @KtorZ made the good point that this is really just the W3C DID standard implemented on Cardano & that's what should be reflected in the CIP title.

So it's important to quality and complete the title by including W3C, DID and/or whatever language they use to indicate what a DID is, plus the fact that this can be referenced & stored in Cardano native token metadata.

The meeting also suggested clarifying & cleaning up the presentation... @Padierfind I think that's just a matter of addressing all the review comments above. 🙏

essbante-io:
Adding data points into a DID may raise some eyebrows in the SSI world. DIDs have been designed to decouple the identifier from the identity. The proposed implementation is pushing the identity of the creator into the DID. I understand this is not the usual context for an SSI implementation and some oddness is to be expected. Theoretically, the identity pieces should be separated from the DID (issued as Verifiable Credentials). 

On the same topic, the payload field used to store the data points is not a core property of the DID document specification https://www.w3.org/TR/did-core/#did-document-properties. I guess it can be added as an extension, but in that case, the details of the property should be specified to know what to expect (see: https://www.w3.org/TR/did-core/#extensibility)

I see the trust in this solution is placed on the Identity Provider and the DID, but I don't see a mechanism to link the DID to the "trusted" Identity Provider. Is the Identity Provider going to function as a trust registry and publish the DIDs it has provided? 

Also, I'm missing the use of signatures. The power of the DIDs comes from using the keys to sign data or prove DID ownership, but I don't see signatures used. Most likely, the implementation can be done with a more straightforward Identifier structure. (meaning: most of the DID document properties are optional. Authentication and verificationMethod are they needed?)

Data protection regulations: Considering the implementation proposes publishing what can be considered PII into an immutable public ledger. It's recommendable to validate with an expert on data protection regulations the responsibility and liability of the Identity Provider.

lohanspies:
> I have proposed in the past that we should even consider using DIDs as NFTs.

@phillewis do you have anything written down regarding using DIDs as NFTs? That is in my opinion the ultimate solution.

phillewis:
> > I have proposed in the past that we should even consider using DIDs as NFTs.
> 
> @phillewis do you have anything written down regarding using DIDs as NFTs? That is in my opinion the ultimate solution.

@lohanspies I have provided some extended thoughts on this in my comment on the CIP-0068 PR at https://github.com/cardano-foundation/CIPs/pull/299#issuecomment-1220219849, as it could address the requirements that both this and CIP-0068 are trying to address, rather than as 2 separate bespoke solutions.

I would be happy to discuss further if you are interested.

Wolfy18:
Do we have a consensus over DIDs metadata? The key for decentralized identity are cryptographic keys which payment address natively depend on (payment.vkey and payment.skey [Key Pairs](https://developers.cardano.org/docs/operate-a-stake-pool/cardano-key-pairs/)  )

Following the web3 standard for DIDs ; A DID is a simple text string consisting of three parts: 1) the did URI scheme identifier, 2) the identifier for the DID method, and 3) the DID method-specific identifier. Also, a DID document contains information associated with the DID, such as ways to cryptographically authenticate a DID controller.

_DID method-specific identifier_ can be the CBORHex serialized wallet address, following this schema 

image

We could store a JSON package in IPFS, like the one below, and upload a DID document which can be an encrypted version of a **Identity Verification Certificate** issued by a credit bureau (Experian, Transunion, Equifax, Okta).

Ex:
_ipfs://QmPPrHGV5UeG7YB9YJ1sVMhvb3NCtjPPFg8ksCtHWzuFVh_
```
{
  "@context": [
    "https://www.w3.org/ns/did/v1",
    "https://w3id.org/security/suites/ed25519-2020/v1"
  ]
  "id": "did:example:123456789abcdefghi",
  "issuer": "did:example:456", # https://bakrypt.io DID 
  "authentication": [{
    
    "id": "did:example:123456789abcdefghi#keys-1",
    "type": "Ed25519VerificationKey2020",
    "controller": "did:example:123456789abcdefghi",
    "publicKeyMultibase": "5820d84d47b1eecc9efc6008a65a7e179a8da8d7d5fe050b6ca3bfffe360d19f88a7"
  }]
}
```

As I mentioned before, the **publicKeyMultibase**  can be represented as the cborHex value of verification key (Public Key).

```
cat payment.vkey

{
    "type": "PaymentVerificationKeyShelley_ed25519",
    "description": "Payment Verification Key",
    "cborHex": "5820d84d47b1eecc9efc6008a65a7e179a8da8d7d5fe050b6ca3bfffe360d19f88a7"
}
```

Therefore we could attach a data structure to a token by minting a transaction as the following.

Ex: 
Metadata Structure linked to an ADA token
```
{
  "<DID_metadatum_label>": {
    "subject": <CBORHEX_paymentaddress>,
    "document": "ipfs://QmPPrHGV5UeG7YB9YJ1sVMhvb3NCtjPPFg8ksCtHWzuFVh"
  }
}
```

The ownership of the **identity verification certificate** can be authenticated by using the signing key (Private key) of the current holder of the minted DID token. Only the legitimate wallet that was used to mint the token is able to decrypt the data of the **verification certificate** by using the secret key of this wallet (payment.skey). 

A _verifier_ can request the _issuer_ to validate the existence of the certificate via APIs, and if necessary the verifier can request the token holder (Wallet owner) to "sign" a transaction and decrypt the DID document on request, **only if its authorized by the wallet owner.**  

[https://www.w3.org/TR/did-core/#a-simple-example](
https://www.w3.org/TR/did-core/#a-simple-example)

Lets try to reach a consensus over this! There are so many use cases for this type of applications. Lets elaborate this idea which of course wasn't mine! 🤣 I'm just trying to push development efforts for this which many of us feels its imperative for the ecosystem.

KtorZ:
@Wolfy18 I don't there's any consensus yet. This proposal is pretty much on hold without much sign of activity for a while. There hasn't been much engagement from the authors regarding the feedback given so far so I am not quite sure what's the state of this proposal and whether it is to be withdrawn? 

cc @Padierfind

Wolfy18:
> @Wolfy18 I don't there's any consensus yet. This proposal is pretty much on hold without much sign of activity for a while. There hasn't been much engagement from the authors regarding the feedback given so far so I am not quite sure what's the state of this proposal and whether it is to be withdrawn?
> 
> cc @Padierfind

Thank you for taking the time to respond. Well I think this a really good feature that can enable Cardano to add more use cases in the real world. 

If anyone’s interested, take a look at this interview I did a couple of days ago during the Cardano Summit in NYC. I’m talking about the subject, some use cases and how to get there.

Fast forward to minute 14:40s 😜

https://youtu.be/3mIs9ROtpow

Again, these are only raw ideas of how to get there leveraging platforms and services that we currently use in our day to day. These ideas are changing constantly and keep in mind that can be wrong! That’s the whole point, let’s have a discussion on how to make this happen. I’m already working on two CIPs to kickstart the whole conversation

KtorZ:
Closing due to inactivity. Feel free to open again to revive the debate.

rphair:
FYI @KtorZ I think we have to add the number 0066 back into the title because there's other discussion that references this material by the pending CIP number: here e.g. https://github.com/cardano-foundation/CIPs/pull/299#issuecomment-1202234305 and in message threads.  I had a PM just last week that referred to it by number alone & all these links will be broken unless we allow the number to persist as an artefact.  If you feel strongly otherwise, just change it again & I'll leave it that way. 😎

PatrickTobler:
Hey, @rphair, I would like to re-open this pull request. And I'm not sure why it was closed in the first place. I mean yes, there hasn't been a lot of stuff happening here but is that a reason to simply close the request? The standard is live and being used by more than a hundred NFT collections out there.

PS: I'm currently locked out of my original Github account... So don't mind the new account please. Lost my Yubikey :'D

rphair (commented):
@PatrickTobler it wasn't just "we didn't hear from you for a while" that caused this proposal to be closed... it was mainly that the comments, from almost the first review, hadn't been answered.  It's good to hear that the specification has been progressing & getting implemented over the last 9 months, but if we can't document this in the PR or the CIP draft then we still don't know what to do with the pull request.

Here are the next things we need you or your team to do:

1 - the CIP format has changed since this draft was written; please review the updated format and put the header & outline in the form we are currently using.  Otherwise we would have to review & mark this up with the same 12 differences or so which have emerged for everyone who wrote proposals back in 2022 before the update:

https://github.com/cardano-foundation/CIPs/blob/master/CIP-0001/README.md

This includes most especially
- the header data, which in your case still has the wrong CIP number, as already reported (not matching the PR title which has the correct number `66`)-
- the directory name itself which (currently `CIP-066`) doesn't have the correct number of 4 digits so it can't be merged as it exists now
- the outline of section headers which match our required template only loosely: https://github.com/cardano-foundation/CIPs/blob/master/.github/CIP-TEMPLATE.md

2 - settle the question about the CIP title as per https://github.com/cardano-foundation/CIPs/pull/294#discussion_r921710463 (still unanswered).  The current header title `NFT Identity` has two words with huge overlap and are therefore much more general than what's specifically proposed here.  I wouldn't block it on this alone, but the editorial suggestion to make the title more specific needs a robust discussion.

3 - we've become more emphatic about documenting all on-chain data in CDDL format, since chain data is in CBOR; so unless I'm wrong about the target please update the Specification section accordingly.

4 - go back & finally address the old reviews.  If the questions have been answered in the meantime with your practical development & what you've learned since then, please update the CIP draft itself as necessary.  In either case please make a comment on each of those review points so either your or editors an officially resolve all of them.

5 - update Path to Active with your current implementation progress, and put in this format please: https://github.com/cardano-foundation/CIPs/blob/master/CIP-0001/README.md#path-to-active

---

If we can do this via GitHub in the next week or so, we can add this for Review at the next CIP meeting with the goal of getting it marked for `Last Check` once we can verify we have a current & properly formatted document, and that that the progression to `Active` status has been as documented.

Also please note here if your new GitHub ID is unable to modify this PR branch.  We would definitely want to preserve the review history rather than opening up a new PR. 🤯

- It is also generally helpful if you allow contributors on the CIP repo (i.e. editors) to also modify your branch: which we currently cannot; which increases our overhead fixing format problems quickly. 🙏

rphair:
@Padierfind @PatrickTobler if interested in pursuing this proposal, please respond from whatever GitHub account you are currently using.  I'm closing this based on a lack of response to the requests in https://github.com/cardano-foundation/CIPs/pull/294#pullrequestreview-1428331182 made over a year ago.

I take your point in https://github.com/cardano-foundation/CIPs/pull/294#issuecomment-1548295736 that "The standard is live and being used by more than a hundred NFT collections out there" but please note: unless this document is merged, none of those projects are following a CIP.

In other words, you might be following a "standard", but it's not a published, open, peer-reviewed Cardano standard like other CIPs... and there is no "CIP-0066" unless this or another PR merges this document into the CIPs repository.  More than one editor has volunteered to cooperate with you on this and you can reopen this PR, or create a new one, any time you are willing to do the required work from your end.

PR #296, opened 2022-07-14
CIP-0062? | Cardano dApp-Wallet Web Bridge Catalyst Extension

ehanoc:
This proposal can be seen as an "extension" of the existing [CIP-30 (Cardano dApp-Wallet Web Bridge)](https://cips.cardano.org/cips/cip30/) to add voting delegation functionality spec'ed in [CIP-36 (Catalyst/Voltaire Registration Transaction Metadata Format - Updated)](https://cips.cardano.org/cips/cip36/). 

This extension also defines the API required for implementation by wallet providers to submit votes to Jormungandr. 

---

[see rendered Markdown](https://github.com/ehanoc/CIPs/blob/cip-governance-wallet-connector/CIP-0062/README.md)


CIP-0062/README.md
```diff
@@ -0,0 +1,220 @@
...
+[CIP-36](https://cips.cardano.org/cips/cip36/).
+
+
+### **Purpose**
+
+```
+type enum Purpose = {
+  CATALYST = 0,
+  OTHER = 1
+}
```
KtorZ, line 52: Suggestion: reverse the order here. Keep `0` for other / misc, `1` for Catalyst and leave the door open for new purposes to be added after 1 :+1:

CIP-0062/README.md
```diff
@@ -0,0 +1,220 @@
...
+## **Namespace**
+
+### **cardano.{walletName}.governance.enable(): Promise\<API>**
+The `cardano.{walletName}.governance.enable()` method is used to enable the governance API. It should request permission from the wallet to enable the API. If permission is granted, the rest of the API will be available. The wallet should maintain a specific whitelist of allowed clients for this API. This whitelist can be used to avoid asking for permission every time.
+
+This api being an extension of [CIP-30 (Cardano dApp-Wallet Web Bridge)](https://cips.cardano.org/cips/cip30/), expects that `cardano.{walletName}.enable()` to be enabled and added to CIP-30 whitelist implicitly.
+
+When both this API and **CIP-30** is being enabled, is up to the wallet to decide the number of prompts requesting permissions to be displayed to the user.
+
+# **`Jormungandr API`**
```
KtorZ, line 78: Jörmungandr hasn't been mentioned so far, and comes a bit out of the blue here. I'd suggest to simply call this section "Governance API" or "API Extension"

CIP-0062/README.md
```diff
@@ -0,0 +1,220 @@
...
+
+## **api.submitVotes**(keyPath: KeyPath, votes: Vote[], spendingCounter: number): Promise\<hash32>
```
KtorZ, line 80: I would recommend to split that function, or more specifically, to remove responsibilities from that function to keep the CIP-0030 API as orthogonal as possible. There are already is a function for submitting transactions (i.e. `submitTx`). Thus all that is really needed here is to:

- (a) create a transaction "skeleton" that contains the rightful metadata
- (b) sign that transaction with the appropriate signing keys (i.e. the stake keys)

I think both could be done at the same time, and return a partially signed transaction (for which input witnesses may still be provided later by `signTx`). So that the transaction can be inspected before submission.

CIP-0062/README.md
```diff
@@ -0,0 +1,220 @@
...
+
+`InvalidVotingKeyError` - If the `keyPath` is not a valid voting key.
+
+`InvalidVotePlanError` - If the `votePlanId` is not a valid vote plan.
+
+`InvalidVoteOptionError` - If the `index` is not a valid vote option.
+
+# **`Delegation API`**
+
+## **api.getVotingKey**(path: KeyPath): Promise\<Bip32PublicKey>
```
KtorZ, line 145: This sounds like it could be a more generic API to get _any_ public key from the wallet.

CIP-0062/README.md
```diff
@@ -0,0 +1,220 @@
...
+
+## **Delegation Cert process**
+
+1. **`Get Voting Key`** - use the method **api.getVotingKey** to return a ed25519 32 bytes public key (x value of the point on the curve).
+
+2. **`Collect Voting Keys`** - Collect the keys to delegate voting power to.
+
+3. **`Craft delegation cert`** - Use **api.buildDelegation** to construct the object containing the key array set to delegate voting power to. Each value will express the `weight` of the voting powers given.
+
+4. **`Submit delegation`** - Submit the metadata transaction to the chain using **api.submitDelegation** which implicitly sign and/or can use the already existing **api.submitTx**,  available from [CIP-30](https://cips.cardano.org/cips/cip30/)
```
KtorZ, line 220: This proposal is missing a rationale section. As far as I can tell, a lot of those operations (like the creation of the metadata itself) can be achieved with client libraries such as the cardano-serialization-lib. Thus, what's the justification for adding them to the wallet connector?

CIP-0062/README.md
```diff
@@ -0,0 +1,217 @@
...
+### **cardano.{walletName}.governance.enable(): Promise\<API>**
+The `cardano.{walletName}.governance.enable()` method is used to enable the governance API. It should request permission from the wallet to enable the API. If permission is granted, the rest of the API will be available. The wallet should maintain a specific whitelist of allowed clients for this API. This whitelist can be used to avoid asking for permission every time.
+
+This api being an extension of [CIP-30 (Cardano dApp-Wallet Web Bridge)](https://cips.cardano.org/cips/cip30/), expects that `cardano.{walletName}.enable()` to be enabled and added to CIP-30 whitelist implicitly.
+
+When both this API and **CIP-30** is being enabled, is up to the wallet to decide the number of prompts requesting permissions to be displayed to the user.
+
+# **`Jormungandr API`**
+
+## **api.submitVotes**(votes: Vote[], spendingCounter: number): Promise\<hash32>
```
dkijania, line 69: IMHO method signature suggests that we are not using some Jormungandr node features:

- fail_fast mechanism - which can control what to do after nth fragment would be rejected ([example](https://github.com/input-output-hk/jormungandr/blob/master/testing/jormungandr-integration-tests/src/jormungandr/rest/v1/fragments/fail_fast.rs#L228))
- parallel_spending_counter - which allows to use 8 different lanes for spending counter. As a result each 8th fragments are independent from each other [example](https://github.com/input-output-hk/jormungandr/blob/master/testing/jormungandr-integration-tests/src/jormungandr/bft/counter.rs#L8)

is it omission?

CIP-0062/README.md
oldcryptogeek: resurrect the serpent

mkazlauskas: Agree. How about adding a `Purpose`arg to `api.submitVotes`? To me it seems like dapps just want to vote on Catalyst and Jormungandr is just an implementation detail that wallets can handle internally. Would probably need a new error type for unsupported voting purpose and/or a way to check if wallet supports it.

CIP-0062/README.md
kevinhammond: This reflects CIP-36.  0 was chosen for backwards compatibility with CIP-15.  The intention is indeed to use purposes from 1 onwards for various voting situations, in the same way that there is eg an RFC for port numbers, and a different RFC that defines the conventional uses of various ports.  At this point in time, we don't know what those purposes will be

CIP-0062/README.md
kevinhammond: Yes, including the purpose makes sense

CIP-0062/README.md
kevinhammond: You would need to specify what kind of key you wanted, but yes this could be extended

CIP-0062/README.md
```diff
@@ -0,0 +1,215 @@
...
+
+### `spendingCounter`: 
+The spending counter is used to prevent double voting. The current spending counter for the account should be provided and the implementation should increment it for each vote before submission and attach it to the vote according to [Jormungandr Voting] (https://input-output-hk.github.io/jormungandr/jcli/vote.html#voting). This needs to be in sequential order.
+
```
stevenj, line 73: How does the dApp get the "Current spending counter for the account"?

CIP-0062/README.md
```diff
@@ -0,0 +1,215 @@
...
+#### `expiration`: 
+chain epoch \& slot for when the vote will expire. The type used is: 
+
+```
+interface BlockDate {
+    epoch: number
+    slot: number
+}
+```
+
```
stevenj, line 92: How does the dApp calculate the expiration BlockDate?  How does it know the current Jormungandr BlockDate in order to calculate an appropriate expiry?

CIP-0062/README.md
stevenj: I think `OTHER = 1` should be removed from the enum, its confusing as its not a real or defined current "Purpose".
There should just be a comment that says, other purposes may be defined in future revisions to this CIP.

minikin: I'd suggest finding a better type name, more self-explanatory than `Purpose.`

CIP-0062/README.md
```diff
@@ -126,32 +126,9 @@ The vote options. This is the set of options we can vote for.
 
 # **`Delegation API`**
 
-## **api.getVotingKeys**(): Promise<cbor<PublicKey\>[]>
-Should return a list of all the voting keys for the current wallet. 
+## **api.getVotingKey**(): Promise\<cbor<PublicKey\>>
```
refi93, line 129: what should this call return if there's no voting key registered yet?

refi93: also, as per the CDDL, a user may register multiple voting keys in a single registration, and multiple of them may be theirs, in theory. In such case, I'd expect this call to return a list of voting keys, or is that case impossible/against specs?

refi93: > what should this call return if there's no voting key registered yet? I'd expect it to return null but the specs apparently don't mention that

ok depends on the definition of "in use", whether it means "registered" or "the next in line to be registered" or something else. But I'd expect to have a call that would be able to provide both keys - the latest registered voting key makes sense to be able to show voting stats/construct the voting transaction and the "the next in line to be registered" key makes sense for the dapp to build the voting delegation. I'd therefore expect actually two different calls to be able to retrieve any of these two keys, as needed, but maybe I'm missing something

CIP-0062/README.md
```diff
@@ -0,0 +1,192 @@
...
+  stakingPub: string // staking public key
+  rewardAddress: string // reward address
+  nonce: number //nonce used in the transaction
+  purpose: Purpose
+}
+```
+
+Errors: `APIError`, `TxSendError`
+
+## **Delegation Cert process**
```
refi93, line 186: I would expect a similar breakdown for the Voting process, it would make the motivation behind the available calls clearer to the reader

CIP-0062/README.md
```diff
@@ -0,0 +1,192 @@
...
+### **cardano.{walletName}.governance.enable(): Promise\<API>**
+The `cardano.{walletName}.governance.enable()` method is used to enable the governance API. It should request permission from the wallet to enable the API. If permission is granted, the rest of the API will be available. The wallet should maintain a specific whitelist of allowed clients for this API. This whitelist can be used to avoid asking for permission every time.
+
+This api being an extension of [CIP-30 (Cardano dApp-Wallet Web Bridge)](https://cips.cardano.org/cips/cip30/), expects that `cardano.{walletName}.enable()` to be enabled and added to CIP-30 whitelist implicitly.
+
+When both this API and **CIP-30** is being enabled, is up to the wallet to decide the number of prompts requesting permissions to be displayed to the user.
+
+# **`Jormungandr API`**
+
+## **api.submitVotes**(votes: Vote[], spendingCounter: number): Promise\<hash32>
```
refi93, line 69: It seems the responsibility of this call is to construct a signed vote transaction containing the votes, using the currently registered voting key and submit it, am I right? I'd clarify this with a comment. Also, if there isn't any voting key registered, I guess this call should throw some error as the vote wouldn't be valid but I see no error code for this.

CIP-0062/README.md
```diff
@@ -0,0 +1,192 @@
...
+
+Errors: `APIError`, `TxSendError`
+
+## **Delegation Cert process**
+
+1. **`Get Voting Key`** - use the method **api.getVotingKey** to return a ed25519 32 bytes public key (x value of the point on the curve).
+
+2. **`Collect Voting Keys`** - Collect the keys to delegate voting power to.
+
+3. **`Submit delegation`** - Submit the metadata transaction to the chain using **api.submitDelegation** which implicitly sign and/or can use the already existing **api.submitTx**,  available from [CIP-30](https://cips.cardano.org/cips/cip30/)
```
refi93, line 192:
```suggestion
3. **`Submit delegation`** - Submit the metadata transaction to the chain using **api.submitDelegation** which implicitly signs it
```
I think with the current CIP-0062 API it's not really possible/reasonable to use `api.submitTx` to submit the delegation transaction. The `api.submitDelegation` call doesn't return back the full signed registration transaction and even if it did, it seems counterproductive to submit the transaction again if the `submitDelegation` call is already required to do that, I'd even expect most implementation to await the tx being submitted before returning, making further submission by `api.submitTx` pointless

CIP-0062/README.md
```diff
@@ -0,0 +1,192 @@
...
+```
+
+#### `choice`: 
+The choice **index** we want to vote for. An `UnkownChoiceError` should be thrown is the value is not within the `proposal` option set.
+
+#### **`proposal`** : 
+proposal information. Include the range of options we can use to vote. This defines the allowed values in `choice`.
+
+```
+interface Proposal {
```
refi93, line 100: will there be an API where the wallet could fetch the proposal metadata to show something more user friendly in the confirmation prompt to the user?

CIP-0062/README.md
```diff
@@ -0,0 +1,411 @@
...
+
+### Returns
+
+The [Signed Delegation Metadata](#signeddelegationmetadata) of the voter registration delegation passed in the `delegation` parameter.
+
+# Examples of Message Flows and Processes
+
+## Delegation Cert process
+
+1. **`Get Voting Key`** - dApp call the method [**api.getCurrentVotingKey**](#apigetcurrentvotingkey-promisecborpublickey) to return a ed25519 32 bytes public key (x value of the point on the curve).
```
refi93, line 333: I can't find the `api.getCurrentVotingKey` method in this doc anymore, was it dropped accidentally?

KtorZ (commented):
I am a bit on the fence regarding this proposal and, I will reiterate [my comment from August 2nd](https://github.com/cardano-foundation/CIPs/pull/296#discussion_r935260754):

> This proposal is missing a rationale section. As far as I can tell, a lot of those operations (like the creation of the metadata itself) can be achieved with client libraries such as the cardano-serialization-lib. Thus, what's the justification for adding them to the wallet connector?

I do not understand what exactly motivates the proposal and what shortcomings it addresses. Even the signing of the vote is already covered by CIP-0030 as it stands today (through `signData`).

CIP-0062/README.md
ehanoc: @KtorZ i agree. A generic method to fetch this would warrant an update on CIP-30

CIP-0062/README.md
ehanoc: @dkijania this method was removed in favor of `signVotes`. Resolving. Please re-open if clarification is needed on the new method

KtorZ:
Also, as a side-note: CIP-0030 is still technically speaking a draft. Which means that methods are still open for adjustments and the whole point of getting it implemented through wallets was to gather  real-life feedback and see shortcomings. Plus, it was mostly entirely designed pre-Alonzo era, which by itself open up a slew of use cases.

 I can see how the `signData` endpoint may today not be in the most adequate form. Perhaps consider a slight modification of `signData` over a new proposal / set of endpoints?

kevinhammond:
> Also, as a side-note: CIP-0030 is still technically speaking a draft. Which means that methods are still open for adjustments and the whole point of getting it implemented through wallets was to gather real-life feedback and see shortcomings. Plus, it was mostly entirely designed pre-Alonzo era, which by itself open up a slew of use cases.

It was decided to propose this as a new CIP rather than a change to CIP-0030, so that this was clearly optional (and could be referred to independently).  Going forwards, we would expect various kinds of (possibly incompatible) DApp support

CIP-0062/README.md
```diff
@@ -0,0 +1,417 @@
...
+* UserDeclined - Raised when the user declined to sign the entire transaction, in the case of a vote submission, this would be returned if the user declined to sign ALL of the votes.
+* VoteRejected - On a vote transaction, where there may be multiple votes.  If the user accepted some votes, but rejected others, then this error is raised, AND `rejectedVotes` is present in the Error instance.
+
+## Governance Extension to CIP-30
+
+### cardano.{walletName}.governance.apiVersion: String
+
+The version number of the Governance Extension API that the wallet supports.
+
+### cardano.{walletName}.governance.enable(purpose: VotingPurpose[]): Promise\<API>
```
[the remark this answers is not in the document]
ehanoc: This is to remove ambiguity of in-what context is the dapp requesting connection to the wallet. The prompt to the user should be clarify for application, use case is requesting connection. Otherwise after the whitelist, potentially the dApp could use it for other means. This was a separate method call, but through discussions and suggestion of @stevenj , it was included in the `enable` method.

CIP-0062/README.md
```diff
@@ -0,0 +1,417 @@
...
+}
+```
+
+* `votingKey` - Ed25519 pubkey 32 bytes HEX string
+* `weight` - Used to calculate the actual voting power using the rules described
+  in [CIP-36](https://cips.cardano.org/cips/cip36/).
+
+### Voting Purpose
+
+```js
```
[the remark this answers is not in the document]
ehanoc: will update :+1:

CIP-0062/README.md
```diff
@@ -0,0 +1,417 @@
...
+Authors: Bruno Martins <bruno.martins@iohk.io>, Steven Johnson <steven.johnson@iohk.io>
+Status: Draft
+Type: Standards
+Created: 2021-06-11
+License: CC-BY-4.0
+---
+
+# Abstract
+
+This document describe the interface between webpage / web-based stack and cardano wallets. This specifies that API of the javascript object that need to be injected into the web applications in order to support all the Governance features.
```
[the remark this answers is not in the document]
ehanoc: Makes sense, will re-write :+1:

CIP-0062/README.md
```diff
@@ -0,0 +1,417 @@
...
+```
+
+* epoch - an epoch value on the voting block chain.
+* slot - a block slot number on the voting block chain.
+
+#### Proposal
+
+```js
+interface Proposal {
+  votePlanId: number
```
stevenj, line 86: This actually needs to be a hex string because the VotePlanId is a Hash of the voteplan, not a number.

CIP-0062/README.md
stevenj: A: the dApp queries the catalyst node to get the current spending counter.

CIP-0062/README.md
```diff
@@ -0,0 +1,417 @@
...
+### Vote
+
+```js
+interface Vote {
+  proposal: Proposal,
+  choice: number,
+  expiration: BlockDate
+  purpose: VotingPurpose,
+  spendingCounter: number
+}
```
stevenj, line 105: Need to ensure the vote can be declared as either private/encrypted or public because signing the vote differs between them.

CIP-0062/README.md
ehanoc: Makes sense. will update. Thanks!

CIP-0062/README.md
KtorZ: (don't say it to me, write it in the proposal!)

CIP-0062/README.md
```diff
@@ -0,0 +1,419 @@
...
+* epoch - an epoch value on the voting block chain.
+* slot - a block slot number on the voting block chain.
+
+#### Proposal
+
+```js
+interface Proposal {
+  votePlanId: string
+  voteOptions: number[]
+  votePublic: boolean
```
refi93, line 88: In the library to sign votes there's an additional proposalIndex parameter which I guess should be included here as well: https://github.com/input-output-hk/catalyst-core/blob/557808b7ceeeb35103abc2d7340b4be909b7ceb5/src/jormungandr/jcli/src/jcli_lib/certificate/new_vote_cast.rs#L22

Ryun1:
Please see digest of IOG Slack discussions as requested within CIP Editors Meeting #55:

### **API Connection**
Having `cardano.{walletName}.governance.enable()` implicitly enable CIP-30 along with the CIP-62 extension was ultimately chosen. The other proposed option was to extend the existing enable function (in CIP-30) to take in a list of API CIPs (30, 62, etc.) and have the wallet return back the list of accepted APIs. The pursued option was preferred because of it allows greater flexibility in variation of `.enable()` between APIs, allowing different payloads to be delivered. 
This is shown by the addition of an array of `votingPurpose` as a parameter to `.enable()`. This affords users good flexibility and protection, whilst allowing dApp developers to silo their dApps to particular governance purposes. This parameter gives wallet users the ability to select which dApps have access to their wallet per voting purpose.

### **Voting Key Derivation**
Past iterations of the specification proposed three functions for voting key derivation; get all used voting keys, rotate current voting key and get the current voting key. Discussions led to the removal of these functions, replacing them with `.getVotingKey()`.
Firstly, these functions may have been costly for wallets, secondly such key control should probably best be decided by wallets, not dApps and thirdly, such need for multiple voting keys and control seems limited. Allowing dApps to have these sorts of controls over keys seems somewhat dangerous and introduces complexity when attempting to engage with governance functionality.
Result of discussions was the removal of these functions, leaving only `.getVotingKey()` with the intention that each wallet account is assigned a singular voting key.

### **Delegation Flow**
The current specification combines the creation of the delegation certificate with the submission to chain in one call (`.submitDelegation()`). Older iterations proposed two separate functions, one to generate delegation certificates and the second to submit the delegation to the chain within a transaction.
	Having two functions gives the advantage that one wallet can create the delegation certificate whilst another is able to package this with a signature into a transaction. This was ultimately decided as undesirable since it would likely be a rare use case. Furthermore, this design would introduce significant complexity for hardware wallets users doubling their signing burden.

A second discussion was had around if dApps should be able to pass in the rewards address field to the delegation certificate, ultimately it was decided that the dApp should not be able to do this.
	Allowing users to select reward addresses via connected dApps would allow voting rewards to be passed to another wallet’s control. But this would also allow malicious dApps to substitute in their own rewards address.
	By passing the reward address burden onto the wallet limits the damage a malicious dApp is able to do. This does not prevent wallets from implementing their own methods to allow users to pass rewards addresses.

### **Voting Flow**
Originally it was proposed that wallets would solely bear the responsibility of packing vote objects into transactions and submitting these to the Jormungandr sidechain.
	This raised concerns as running Jormungandr nodes would introduce cost for wallet providers with considerable complexity for a narrow use case; Catalyst vote submission.
	It was concluded that it would be better for dApps to submit voting transactions themselves, this significantly reduces the complexity of wallet implementations. Only requiring wallets to convert vote objects into signed vote transactions. This has the added benefit of making the specification less Catalyst specific.
	Due to the nature of vote objects, dApps already have to be in contact with an instance of a Jormungandr node to be able to provide all the required information to the wallet.

To aid wallet implementations of this function, the Jormungandr team have added Javascript bindings for vote transactions (aka fragment) creation - in [chain wallet libs](https://github.com/input-output-hk/catalyst-core/tree/main/src/chain-libs).

CIP-0062/README.md
ehanoc: *keyPath* removed. Single Voting key generated now.

CIP-0062/README.md
ehanoc: @refi93 The latest iteration of this spec, expects a single key to be created / "in-use" . We discussed supporting a N keys to be created, but for now seems unnecessary. Doesn't mean we wouldn't expand it in the future. 

Does it makes sense?

CIP-0062/README.md
ehanoc: Clearing this, as discussed that we only support signing votes through this API and the dapp should be able to submit it.

CIP-0062/README.md
```diff
@@ -0,0 +1,192 @@
...
+`Purpose`: Defines the purpose of the delegations. This is used to limit the scope of the delegations.  For example, a purpose might be a subset of Catalyst proposals, a council election, or even some private purpose (agreed by convention).
+
+## **Namespace**
+
+### **cardano.{walletName}.governance.enable(): Promise\<API>**
+The `cardano.{walletName}.governance.enable()` method is used to enable the governance API. It should request permission from the wallet to enable the API. If permission is granted, the rest of the API will be available. The wallet should maintain a specific whitelist of allowed clients for this API. This whitelist can be used to avoid asking for permission every time.
+
+This api being an extension of [CIP-30 (Cardano dApp-Wallet Web Bridge)](https://cips.cardano.org/cips/cip30/), expects that `cardano.{walletName}.enable()` to be enabled and added to CIP-30 whitelist implicitly.
+
+When both this API and **CIP-30** is being enabled, is up to the wallet to decide the number of prompts requesting permissions to be displayed to the user.
```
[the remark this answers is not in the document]
ehanoc: That's correct @refi93 . I've updated this section, could you have a re-read?

CIP-0062/README.md
ehanoc: Oops. Correct. fixing

CIP-0062/README.md
ehanoc: just added @refi93 :+1:

ehanoc:
> @ehanoc - tagging this as `Waiting for Author` because of the number of unresolved conversations above (particularly the still-missing Rationale section) and the recent incoming references to this proposal (so visitors can more easily see its status).

Thanks @rphair , just addressed most of them and added the rationale section. 

Still need to add the update for Private Voting. Incoming

CIP-0062/README.md
refi93: makes sense :+1:

CIP-0062/README.md
```diff
@@ -0,0 +1,429 @@
...
+The `cardano.{walletName}.governance.enable()` method is used to enable the
+governance API. It should request permission from the wallet to enable the API for the requested purposes.
+
+If permission is granted, the rest of the API will be available. The wallet
+should maintain a specific whitelist of allowed clients and voting purposes for
+this API. This whitelist can be used to avoid asking for permission every time.
+
+This api, being an extension of
+[CIP-30 (Cardano dApp-Wallet Web Bridge)](https://cips.cardano.org/cips/cip30/),
+expects that `cardano.{walletName}.enable()` to be enabled and added to CIP-30
```
refi93, line 267:
```suggestion
expects `cardano.{walletName}.enable()` to be enabled and added to CIP-30
```

CIP-0062/README.md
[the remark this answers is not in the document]
refi93: I'm checking https://github.com/cardano-foundation/CIPs/pull/296/files#diff-d3ac75d45e4e621f6ad95567378e0dc2b671f44d3c32425dda9ca08210e18876R265-R272 but I didn't notice a difference, am I checking perhaps the wrong commit/branch?

CIP-0062/README.md
```diff
@@ -0,0 +1,429 @@
...
+
+### Vote
+
+```ts
+interface Vote {
+  proposal: Proposal
+  choice: number
+  expiration: BlockDate
+  purpose: VotingPurpose
+  spendingCounter: number
```
refi93, line 116: I think also wallet settings are missing here. The vote signer needs to know some basic network parameters to produce a valid vote. It is a stringified JSON in the vote signing lib: 

https://github.com/input-output-hk/catalyst-core/blob/main/src/chain-wallet-libs/bindings/wallet-js/js-test/test/vote_cast.js#L9

refi93: also, I'm not sure whether some parameter to manage the "lanes" wouldn't be needed in case of signing multiple votes with the `signVotes()` call: https://github.com/input-output-hk/catalyst-core/issues/131#issuecomment-1339195034

Ryun1:
One problem with the current design is that dApps will not be able find delegations to other voting keys a connect wallet has made without access to the wallet's stake credential.

To remedy this we could replace `.getVotingKey()` with `.getVotingCredentials()` whereby both a wallet's voting key and stake credential is returned. This allows dApps to index the chain by voting key and stake credential to find a wallet's delegation/registration history.

CIP-0062/README.md
```diff
@@ -0,0 +1,215 @@
...
+
+`address_index` - index of the key to use. 
+
+### **Returns**
+cbor hex encoded representation of the public key
+
+
+## **api.getCurrentVotingKey**(): Promise\<cbor<PublicKey\>>
+
+Should return the current in-use voting public-key. The wallet should maintain a reference to the current `adress_index` counter and return the public key for that index.
```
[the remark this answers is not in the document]
ehanoc: clearing as this method doesn't exist anymore

CIP-0062/README.md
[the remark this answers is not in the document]
ehanoc: > Doesn't this sentence imply that the dapp has to **always** enable first the CIP-30 API and only then request the governance one? On the other hand this sentence

Yes, it ties to explain that CIP-30 should be enabled as well, if not so already. 

Maybe Is there a reason why it shouldn't? 

> Is my understanding that CIP-30 always has to be enabled before CIP-62 correct?

Yes, that was the idea. 

Would it be more clear if i were to remove the section that states: 
```
When both this API and CIP-30 is being enabled, is up to the wallet to decide the number of prompts requesting permissions to be displayed to the user.
```

CIP-0062/README.md
ehanoc: I see; this would definitely be easier to implement on the client side. 

About some sort of Map<string,string> // <planId, proposalTitle> ?

ehanoc:
> One problem with the current design is that dApps will not be able find delegations to other voting keys a connect wallet has made without access to the wallet's stake credential. Stake credential is usually a Public Stake Key, but CIP-36 allows for other credentials to be used.
> 
> To remedy this we could replace `.getVotingKey()` with `.getVotingCredentials()` whereby both a wallet's voting key and stake credential is returned. This allows dApps to index the chain by voting key and stake credential to find a wallet's delegation/registration history.

Thanks @Ryun1  I'm quite a fan of this. 

Something like this?
```ts
## api.getVotingCredentials(): Promise<VotingCredentials>

interface VotingCredentials {
votingKey: PublicKey
stakingKey: PublicKey
}
```
@refi93 @stevenj @kevinhammond @gyorgy-kurucz-vacuum @dkijania thoughts?

CIP-0062/README.md
```diff
@@ -284,7 +284,7 @@ returned to the dApp with the following methods.
 
 ## Governance API
 
-All methods (all but the signing functionality) should not require any user
+Except `signVotes`, no other method should require any user
```
Ryun1, line 287: `.submitDelegation()` should also need the user to authorise signing and submission of delegation transaction. Also, dependent on wallet implementations, `.getVotingKey()` may need user interaction -- to authorize mnemonic exposure for voting key derivation.

CIP-0062/README.md
```diff
@@ -0,0 +1,443 @@
...
+* delegations - List of [Governance Keys](#governancekey) and their voting weight to delegate voting power to.
+* purpose - The [Voting Purpose](#voting-purpose) being delegated.  A voter may have multiple active delegations for different purposes, but only 1 active delegation per unique Voting Purpose.
+
+## DelegatedCertificate
+
+```ts
+interface DelegatedCertificate {
+  delegations: GovernanceKey[]
+  stakingPub: string
+  rewardAddress: string
```
refi93, line 162: I guess this should be renamed to match the latest CIP-36 specs: https://github.com/cardano-foundation/CIPs/pull/373

```suggestion
  paymentAddress: string
```

Ryun1: @refi93 

Yes good spot - I was going to reflect that change here once #373 is merged. 

Next week I will do a pre-emptive PR to this CIP to match the changes in #373 and #427 so when/if they are merged this CIP is aligned.

ehanoc:
> @ehanoc - tagging this as `Waiting for Author` because of the number of unresolved conversations above (particularly the still-missing Rationale section) and the recent incoming references to this proposal (so visitors can more easily see its status).

@rphair Rationale section in now and a few updates.

rphair:
> @ehanoc: @rphair Rationale section in now and a few updates.

Thanks: Rationale looks good.  There have been a few changes to the document format since you originally submitted your proposal and it would be helpful if you could edit to conform to them while this document is still fluid:

New header preamble (https://github.com/cardano-foundation/CIPs/tree/master/CIP-0001#header-preamble) has some fields added & others removed.

All top level headings like Abstract, Motivation, Specification, etc. should be H2s (markdown: `##`) as per https://github.com/cardano-foundation/CIPs/blob/master/.github/CIP-TEMPLATE.md , with the smaller headings descending from there (it would be cleaner if you were to edit the rank of your many minor section headings, rather than us possibility messing it up).

All proposals must have a Path to Active, even if already active & completed (in that case having a description of the path taken so far).  This section & the its two sections (acceptance & implementation) are described here (there are some live examples in the CIP archive already): https://github.com/cardano-foundation/CIPs/tree/master/CIP-0001#status-proposed

CIP-0062/README.md
```diff
@@ -0,0 +1,473 @@
...
+}
+```
+
+An individual raw Unsigned Vote record.
+
+* proposal - The [proposal](#proposal) being voted on.
+* choice - This value MUST match one of the available `voteOptions` in the `proposal` element. An `UnknownChoiceError` should be thrown if the value is not one of the valid `voteOptions` of the `proposal`.
+* expiration - Voting Chain epoch \& slot for when the vote will expire. This value is supplied and maintained by the dApp, and forms a necessary component of a [Vote](#vote)
+* purpose - The [voting purpose](#voting-purpose) being voted on. (Currently always 0).
+* spendingCounter - Optional, but if present, the spending counter is used to prevent double voting. The spending counter for the vote transaction will be supplied and maintained by the dApp. The dApp will manage supplying this in the correct order, and this should not be enforced by the Wallet.
```
refi93, line 141: if not provided, what value, if any should this default to?

CIP-0062/README.md
```diff
@@ -0,0 +1,473 @@
...
+
+#### Proposal
+
+```ts
+interface Proposal {
+  votePlanId: string;
+  proposalIndex: number;
+  voteOptions?: number[];
+  voteEncKey?: string;
+}
```
refi93, line 112: since voteEncKey and voteOptions are required for private votes, I'd consider rewriting it like this to make that requirement explicit at type level:

```suggestion
interface Proposal =
  | {
      votePlanId: HexString
      voteOptions: number[]
      votePublic: false
      proposalIndex: number
      voteEncKey: HexString
    }
  | {
      votePlanId: HexString
      voteOptions: number[]
      votePublic: true
      proposalIndex: number
    }
```

CIP-0062/README.md
```diff
@@ -0,0 +1,473 @@
...
+The API chosen here is for the minimum API necessary for dApp <-> Wallet
+interactions without convenience functions that don't strictly need the wallet's
+state to work.
+
+## api.signVotes(votes: Vote[], settings: string): Promise\<Bytes>[]
+
+Errors: [`APIError`](#extended-apierror), [`TxSignError`](#extended-txsignerror)
+
+* `votes` - an array of up to 10 votes to be validated with the wallet user and, if valid, signed.
+* `settings` - *Optional*. Settings which are universally applicable to all `votes` being signed.  This is a json string.  The fields within the settings json string depend on the format of the vote record.  The wallet does not need to interpret this field and should pass it unchanged to the logic used to format the vote transaction.  However, to support multiple future vote transaction formats, the `settings` must contain at least:
```
refi93, line 323: if the wallet doesn't validate the purpose in settings, can't a malicious actor abuse that to pass a different purpose than the one whitelisted by the wallet? If so, I'd consider removing the purpose from the settings and passing it separately, or at least specifying that the wallet should validate that field in the settings

refi93: though looking at the js bidings lib to sign votes, I don't see the `purpose` field being used at all: https://github.com/input-output-hk/catalyst-core/blob/f5865be9a9d227e975689e32e6c03b5b34ff5d96/src/chain-wallet-libs/bindings/wallet-wasm-js/js-test/test/vote_cast.js#L10 - what's the benefit of passing `purpose` in the `settings` parameter then?

CIP-0062/README.md
Ryun1: This is sensible - I will make this change.

CIP-0062/README.md
Ryun1: These fields should default to 0 - thanks for raising this I will make it explicit in the specification.

CIP-0062/README.md
Ryun1: The intent of including the purpose here is to allow wallets to validate. Wallets should be able to show the supplied purpose of a vote to user at signing time for them to validate.

Ryun1:
Motivation for last two commints:

CIP-0062/README.md
Ryun1: Will do.

CIP-0062/README.md
```diff
@@ -0,0 +1,409 @@
...
+
+1. **Get Voting Key** - dApp calls the method `api.getVotingCredentials()` to return the connected wallet account's public `voteKey`.
+
+2. **Construct Delegation** - The dApp constructs `Delegation` using the Wallet's public `voteKey`, `weight` of 1 and choice of `VotingPurpose`.
+
+3. **Submit Delegation** - The dApp passes the `Delegation` object to the Wallet to build a metadata transaction and submit this to Cardano blockchain. Wallets are able employ the already existing [`api.submitTx()`](https://cips.cardano.org/cips/cip30/#apisubmittxtxcbortransactionpromisehash32), available from [CIP-30](https://cips.cardano.org/cips/cip30/).
+
+#### Voting
+
+TODO
```
SebastienGllmt, line 383: todo?

CIP-0062/README.md
```diff
@@ -0,0 +1,409 @@
...
+
+## Rationale
+
+To provide [CIP-36](https://cips.cardano.org/cips/cip36/) style governance specific functionality for Catalyst to wallet's and expose such API to the dApps (i.e Catalyst Voting Centers).
+
+This also addresses some short-comings of [CIP-30](https://cips.cardano.org/cips/cip30/); which signData can only be done by known an address; This signature is not relevant to a specific address, nor the dApp will know an address attached to the voting key. The voting key derivation is defined in [CIP-36](https://cips.cardano.org/cips/cip36/).
```
SebastienGllmt, line 389: typo? "by known an address"

PR #298, opened 2022-07-14
CIP-0067? | Asset Name Label Registry

alessandrokonrad:
This proposal defines a standard to classify Cardano native assets by the asset name.

---

[see rendered Markdown](https://github.com/berry-pool/CIPs/blob/asset_name_registry/CIP-0067/README.md)


rphair (commented):
thanks @alessandrokonrad and although it's a much different proposition I'll try to link this with the proposal https://github.com/cardano-foundation/CIPs/pull/137 which has been stalled for about 3 months: not to compare them but to get some of the same community contributing to this review.

Keeping the token registry centralised (ugh) does address the security concerns raised by the earlier proposal, by leaving new inclusions up to manual verification by CIP editors (ugh) and requiring pull requests to this repo each time a new asset is added.  I suppose that is the only alternative now to an agreed-upon, secure standard for keeping those records on chain... if so then **1)** could your CIP please discuss the alternative possibility of an on-chain token registry & why your solution is better?

As you say in the Motivation "As more assets are minted" it's becoming harder for "third parties" to know what to do with them.  Maybe _all_ Cardano assets don't need to be included in this registry, and therefore the overhead of manual work on this centralised DB would be OK?  If so then **2)** perhaps you would explain more in the CIP text _which assets_ the new registry would be applicable, and which would not? 🧐

alessandrokonrad:
> thanks @alessandrokonrad and although it's a much different proposition I'll try to link this with the proposal #137 which has been stalled for about 3 months: not to compare them but to get some of the same community contributing to this review.
> 
> Keeping the token registry centralised (ugh) does address the security concerns raised by the earlier proposal, by leaving new inclusions up to manual verification by CIP editors (ugh) and requiring pull requests to this repo each time a new asset is added. I suppose that is the only alternative now to an agreed-upon, secure standard for keeping those records on chain... if so then **1)** could your CIP please discuss the alternative possibility of an on-chain token registry & why your solution is better?
> 
> As you say in the Motivation "As more assets are minted" it's becoming harder for "third parties" to know what to do with them. Maybe _all_ Cardano assets don't need to be included in this registry, and therefore the overhead of manual work on this centralised DB would be OK? If so then **2)** perhaps you would explain more in the CIP text _which assets_ the new registry would be applicable, and which would not? 🧐



> thanks @alessandrokonrad and although it's a much different proposition I'll try to link this with the proposal #137 which has been stalled for about 3 months: not to compare them but to get some of the same community contributing to this review.
> 
> Keeping the token registry centralised (ugh) does address the security concerns raised by the earlier proposal, by leaving new inclusions up to manual verification by CIP editors (ugh) and requiring pull requests to this repo each time a new asset is added. I suppose that is the only alternative now to an agreed-upon, secure standard for keeping those records on chain... if so then **1)** could your CIP please discuss the alternative possibility of an on-chain token registry & why your solution is better?
> 
> As you say in the Motivation "As more assets are minted" it's becoming harder for "third parties" to know what to do with them. Maybe _all_ Cardano assets don't need to be included in this registry, and therefore the overhead of manual work on this centralised DB would be OK? If so then **2)** perhaps you would explain more in the CIP text _which assets_ the new registry would be applicable, and which would not? 🧐

The purpose of this CIP is to register token standards. This is not about registering specific assets, so it's not a central token registry.
Initially we only had CIP-0025 and the Cardano foundation off-chain registry. For a 3rd party it was fine to to work with that: Check if the asset has on-chain metadata following CIP-0025, if not then check the off-chain registry. But as new standards emerge a 3rd party needs to go through all options until one fits, but then it's still not clear if the correct method was used.
So I thought wouldn't it be cool to have a way to determine the token standard and type (NFT, FT, etc.) just from the assetname. The 3rd party knows exactly what to do with the asset and what type it is.
For instance my other proposal [Datum Metadata Standard](https://github.com/cardano-foundation/CIPs/pull/299) makes use of this CIP. There are 3 registered `asset_name_labels`:
- 100: Reference NFT
- 222: NFT => Look up Reference NFT and retrieve specific metadata from output datum
- 333: FT => Look up Reference NFT and retrieve specific metadata from output datum

rphair:
thanks @alessandrokonrad then there would not be very much overhead & also not related very much to the other proposal.  I appreciate your clearing up my misunderstanding 😎

CIP-0067/README.md
```diff
@@ -0,0 +1,49 @@
...
+As more assets are minted and different standards are used to query data for these assets, it's gettinger harder for 3rd parties to determine the asset type and how to proceed with it. This standard is similar to [CIP-0010](https://github.com/cardano-foundation/CIPs/tree/master/CIP-0010), but focuses on the asset name of an asset.
+
+
+## Specification
+
+To classify assets the `asset_name` needs to be prefixed with an opening and closing parentheses and the label in betwen: `({Label})`.
+ 
+For example:
+
+UTF-8 encoded: `(123)TestToken`\
```
SebastienGllmt, line 28: Hmm, (333) ends up taking 5 bytes of ascii which is more than coming up with a binary specification for this. I guess the advantage is that this is human readable.

alessandrokonrad: Yeah that's true, I think the compromise is worth it. With 5-6 bytes you still have 25-26 bytes free in the asset name. 
I also thought about encoding it in binary, but you won't be able to ever utf-8 encode the asset name again. The only solution I have in mind here is to split the asset name in two parts if a label was detected in the asset name. Decode the label separately and the remaining asset name to make both parts human readable if applicable.

michaelpj:
This CIP needs a Rationale section that explains how it achieves its goals, and in particular whether it is secure.

In particular, there is no way to know whether the creator of a token knew about this standard or intended to adhere to it (without some other metadata channel), so it is potentially dangerous to assume that these labels have these meanings. An oblivious token issuer might use the `(X)` prefix for all kinds of things, leading to misinterpretation by clients that follow this standard. For example, in https://github.com/cardano-foundation/CIPs/pull/299#issuecomment-1192399200 this manifests as metadata spoofing attacks.

Ways to improve the security would be to use a more obscure encoding of the data so it's less likely to be used by accident, but otherwise I think this is a fundamental security problem with this proposal. It might be safe enough to use anyway, but this should be discussed explicitly in the CIP.

CIP-0067/README.md
```diff
@@ -0,0 +1,53 @@
...
+This proposal defines a standard to classify Cardano native assets by the asset name.
+
+## Motivation
+
+As more assets are minted and different standards emerge to query data for these assets, it's getting harder for 3rd parties to determine the asset type and how to proceed with it. This standard is similar to [CIP-0010](https://github.com/cardano-foundation/CIPs/tree/master/CIP-0010), but focuses on the asset name of an asset.
+
+
+## Specification
+
+To classify assets the `asset_name` needs to be prefixed with an opening and closing parentheses and the label in between: `({Label})`.
```
alessandrokonrad, line 24: Alternatives to parentheses to make it more obscure:
- Single colon, e.g. `:222:`
- Double colon, e.g. `::222::`

BlakeBrown:
> This CIP needs a Rationale section that explains how it achieves its goals, and in particular whether it is secure.

@alessandrokonrad I read the rationale section, but "Asset name labels make it easy to classify assets." is kind of vague. If I understand correctly, you want this proposal to easily detect the metadata type in a Plutus validator?

> In particular, there is no way to know whether the creator of a token knew about this standard or intended to adhere to it (without some other metadata channel), so it is potentially dangerous to assume that these labels have these meanings. An oblivious token issuer might use the `(X)` prefix for all kinds of things, leading to misinterpretation by clients that follow this standard. For example, in [#299 (comment)](https://github.com/cardano-foundation/CIPs/pull/299#issuecomment-1192399200) this manifests as metadata spoofing attacks.

> 
> Ways to improve the security would be to use a more obscure encoding of the data so it's less likely to be used by accident, but otherwise I think this is a fundamental security problem with this proposal. It might be safe enough to use anyway, but this should be discussed explicitly in the CIP.

I appreciate the security input @michaelpj but I think in practice these aren't huge issues? `(XXX)` or `:XXX:` are both fairly obscure patterns. The onus is on the token issuer (NMKR, Anvil, JPG) to ensure that minted assets are following the latest CIPs. It would be exceptionally rare that an individual would mint an asset on their own AND accidentally classify their metadata incorrectly against this standard.

What might be more common is an attacker using the lack of verifiability to attack a smart contract, but that can be defended against.

BlakeBrown (approved):
Can we import the [registry](https://github.com/cardano-foundation/CIPs/blob/master/CIP-0010/registry.json) from CIP-10 to ensure backwards compatibility?

Would also appreciate an expansion in the rationale section.

Otherwise, LGTM! 👍

perturbing:
Regarding the encoding of the standard in the token name. Since most 3rd parties do not rely on the actual token name for their displayed name but rather the metadata, the visual impact of the encode matters less. Besides the possible UTF-8 encoding `«222»` [mentioned](https://github.com/cardano-foundation/CIPs/pull/299#issuecomment-1210494663) in CIP 68, we might choose to do it differently and encoded it in binary to reach the entropic limit of the encoding. This also opens the door to implement a checksum for extra obscurity. What are your thoughts? @michaelpj @SebastienGllmt @KtorZ

CIP-0067/registry.schema.json
```diff
@@ -0,0 +1,62 @@
+{
+  "$schema": "http://json-schema.org/draft-07/schema",
+  "$id": "https://github.com/cardano-foundation/CIPs/blob/master/CIP-?",
```
KtorZ, line 3: https://cips.cardano.org/cips/cip67/registry.schema.json should (eventually) work.

e.g. https://cips.cardano.org/cips/cip10/registry.schema.json

CIP-0067/registry.schema.json
```diff
@@ -0,0 +1,62 @@
...
+            "$id": "#/items/anyOf/0/properties/asset_name_label",
+            "type": "integer",
+            "title": "The asset_name_label number",
+            "default": 0,
+            "examples": [1967]
+          },
+          "class": {
+            "$id": "#/items/anyOf/0/properties/class",
+            "type": "string",
+            "title": "The asset class",
```
KtorZ, line 46: This is nowhere mentioned in the specification. If this can only be `FT` or `NFT`, I suggest to add an extra validation to this subschema: 

```
"enum": [ "FT", "NFT" ]
```

CIP-0067/registry.schema.json
```diff
@@ -0,0 +1,62 @@
...
+          },
+          "description": {
+            "$id": "#/items/anyOf/0/properties/description",
+            "type": "string",
+            "title": "The asset name label description",
+            "default": "",
+            "examples": ["CIP-0025 - NFT Metadata Standard"]
+          }
+        },
+        "additionalProperties": true
```
KtorZ, line 58: Why allow additional properties?

CIP-0067/README.md
```diff
@@ -0,0 +1,53 @@
...
+
+UTF-8 encoded: `(123)TestToken`\
+HEX encoded: `283132332954657374546F6B656E`
+
+
+These are the reserved `asset_name_label` values
+
+`asset_name_label`            | description
+----------------------------  | -----------------------
+0 - 15                        | reserved\*
```
KtorZ, line 36: What does the asterisk points to?

CIP-0067/README.md
```diff
@@ -0,0 +1,53 @@
...
+65536 - 131071                | reserved - private use
+
+For the registry itself, please see [registry.json](./registry.json) in the machine-readable format. Please open your pull request against
+this file.
```
KtorZ, line 40: One problem we had / have with CIP-0010 is that there's no particular "rule" that defines what can go in the registry. As editors we try to do some basic sanity check and ask people to pitch / justify a bit their project; but it would be nice / preferable if these rules were specified in the specification itself. For example: 

---

### Adding an entry to the registry

To propose an addition to the registry edit the `registry.json` with your details, open a pull request against the CIPs repository and give a brief description of your project and how you intend to use metadata associated with the label entry.

CIP-0067/README.md
KtorZ: Or, what about prefixing every thing with `CIP67:{label}:` ? A bit longer, but it leaves 24-20 bytes for the asset name. One can also find a more compact binary encoding which would take less bytes but loose the "readability" aspect of it. 

For example, consider the prefix to be `0xC067`, that's only 2 bytes and reads well once hex-encoded.

KtorZ, line 24: Also, additional suggestion: whatever the choice of format it, perhaps give an [ABNF](https://en.wikipedia.org/wiki/Augmented_Backus–Naur_form) syntax to describe it so that symbols aren't misinterpreted (one person reading `({label})` might think for its label 42, the prefix would be: `({42})`. 

```abnf
asset-name = asset-name-label asset-name-body

asset-name-label = "(" 1-5DIGIT ")"

asset-name-body = *OCTET ; exact length depends on the asset-name-label's length
```

CIP-0067/README.md
```diff
@@ -0,0 +1,53 @@
...
+License: CC-BY-4.0
+---
+
+## Abstract
+
+This proposal defines a standard to classify Cardano native assets by the asset name.
+
+## Motivation
+
+As more assets are minted and different standards emerge to query data for these assets, it's getting harder for 3rd parties to determine the asset type and how to proceed with it. This standard is similar to [CIP-0010](https://github.com/cardano-foundation/CIPs/tree/master/CIP-0010), but focuses on the asset name of an asset.
```
KtorZ, line 19:
```suggestion
As more assets are minted and different standards emerge to query data for these assets, it's getting harder for 3rd parties to determine the asset type and how to proceed with it. This standard is similar to [CIP-0010](../CIP-0010), but focuses on the asset name of an asset.
```

CIP-0067/README.md
```diff
@@ -0,0 +1,53 @@
+---
+CIP:
```
KtorZ, line 2:
```suggestion
CIP: 67
```

CIP-0067/README.md
```diff
@@ -0,0 +1,53 @@
...
+Title: Asset Name Label Registry
+Authors: Alessandro Konrad <alessandro.konrad@live.de>, Thomas Vellekoop <thomas.vellekoop@iohk.io>
+Comments-URI:
+Status: Draft
+Type: Informational
+Created: 2022-07-13
+Post-History:
```
KtorZ, line 9:
```suggestion
```

CIP-0067/README.md
```diff
@@ -0,0 +1,53 @@
...
+----------------------------  | -----------------------
+0 - 15                        | reserved\*
+65536 - 131071                | reserved - private use
+
+For the registry itself, please see [registry.json](./registry.json) in the machine-readable format. Please open your pull request against
+this file.
+
+## Rationale
+
+Asset name labels make it easy to classify assets. It's important to understand that a registered label standard itself doesn't provide any security off-chain nor on-chain as they can be spoofed. Only in combination with the Policy ID security can be derived from the minting policy.
```
KtorZ, line 44: Answers I'd like to see in the Rationale section:

- What are the 0-15 labels reserved for and how is this different from the 65536 - 131071 range?
- How big is the risk of collision for a given choice of prefix? (i.e. how likely is it that a "randomly generated" -- or a hash digest -- asset name may be misinterpreted as a CIP-0067's label. (though the current rationale hints in that direction by recommending to always consider the asset label in conjunction with the policy id)
- Is there any consideration regarding the size of the prefix? It seems to me that keeping the size of the prefix / asset name label under 4 bytes is preferable, because it allows to still embed 28-byte hash digests in the asset name.

CIP-0067/README.md
KtorZ: *I am only seeing / reading this comment now: https://github.com/cardano-foundation/CIPs/pull/298#issuecomment-1233421621

KtorZ:
@perturbing

Thanks for the detailed walkthrough on checksums options that you've considered. I have mainly two questions from reading your comment: 

(1) I understand the willingness to avoid collision but, since this registry exists off-chain anyway, wouldn't it make sense to also bundle the policy id that supports some of the identified tokens? So, if a certain project wants to use or define a new label, they can do so by adding an entry to this registry including their policy id. Downstream components that support the standard can interpret only policies that are included in the registry. Since an asset name is always to be seen within the context of a policy, the checksum altogether becomes even redundant. Doesn't it?

(2) It seems to me that a checksum as large as the data payload is overkill. In this scenario, the role of the checksum is really to make it harder to people to unknowingly abide by the standard, then I think that a _best effort_ is sufficient; given that there are also specific prefix and suffix that already make it less likely to happen. The prefix/suffix by themselves already makes the chance of accidental use of the standard less than 0.5%. Adding even a checksum on 4 bits brings that down to ~0.025%, if my calculations are correct. And, that isn't even considering that the label itself must match an existing label. Say the standard becomes really popular and ends up with 1000s of labels, you'd still need to get an equality on the 20 remaining bits (though you have 1000s of possible cases, so that's ~0.1% probability of collision, which brings the overall collision probability to 0.000025% (2.5e-7). I think this is largely sufficient as a "best effort".

SebastienGllmt:
I'm not sure I understand why we're prefixing and suffixing with a bunch of 0s instead of just using a larger checksum. Is it meant for human readability by looking at the binary encoding?

One thing that might be interesting is instead of coming up with a new checksum algorithm to instead use bech32 with `CIP67:` as the prefix. Note that in bech32, the prefix is included in the checksum mechanism so other people can use the same concept in the future to change CIP67 to something else if they want to

The problem with bech32 is that it operates on 5-bit chunks which may be tricky and also that it only has 32 separate characters for display despite the fact cip67 encodes a utf8 string (so it means these assets would have multiple representations -- the bech32 representation and the name encoded inside. Although this is also kind of true with your  custom 0-padding approach as well)

KtorZ:
> @SebastienGllmt: One thing that might be interesting is instead of coming up with a new checksum algorithm to instead use bech32 

 I thought about bech32 as well although, it comes with probabilistic error-detection and as you said, operates over 5 bits already. So it feels like an overkill for this particular purpose. 
 
I do agree with the "not re-inventing the wheel" statement though which is why I tend to be in favor of a relatively simple checksum solution. I would also consider a CRC before bech32. 

> @alessandrokonrad: Overall are you in favor of a binary encoded or utf-8 encoded label version?

The binary-encoded label is a much more robust approach. Especially because the asset name isn't meant to be a direct user-facing piece of information in principle. This is why we have metadata after all. Thus, for a standard which is ultimately about providing such metadata, I find it even ironic to make any effort to have the label somewhat human-readable 😬 ! 
I would also not over do it, to not hinder adoption of the standard. As I explained above, a best-effort checksum can already drastically reduce the chance of collision. If anything though, I'd choose a prefix / suffix that relates to the CIP a bit more (e.g. `0067....0067`). 

All-in-all, regardless of the solution chosen, the CIP will need to include test vectors and a reference implementation to ease development of different solutions.

CIP-0067/registry.schema.json
alessandrokonrad: Logically we can only have `FT` and `NFT`, but I think there can be more categories in the end. E.g. Semi fungible tokens (`SFT`), etc. Whatever establishes in the future.

CIP-0067/README.md
```diff
@@ -0,0 +1,93 @@
...
+
+To classify assets the `asset_name` needs to be prefixed the following `4 bytes` binary encoding:
+```
+[ 0000 | 2 bytes label_num | 1 byte checksum | 0000 ]
+```
+- The leading and ending four 0s are brackets
+- `label_num` has a fixed size of 2 bytes (`Label range in decimal: [0, 65535]`). 
+If `label_num` < 2 bytes the remaining bits need to be padded with 0s.
+- `checksum` has a fixed size of 1 byte. The checksum is calculated by applying the CRC-8 algorithm on the `label_num (including the padded 0s)`. 
+
```
perturbing, line 30: > the CRC-8 algorithm

There is no one canonical CRC 8 algorithm, each one uses a specific irreducible polynomial over GF(2) (a finite field over a binary field). We should specify which irreducible polynomial we take for the standard. Preferably, we denote this polynomial in its binary form, so as an example
```
x^8 + x^4 + x^3 + x^2 + 1 = 1*x^8 + 0*x^7 + 0*x^6 + 0*x^5 + 1*x^4 + 1*x^3 + 1*x^2 + 0*x^1 + 1*x^0
```
becomes `100011101` in its binary representation. I found online the following primitive polynomials of degree 8 for GF(2)
```
x^8 + x^4 + x^3 + x^2 + 1
x^8 + x^5 + x^3 + x^1 + 1
x^8 + x^6 + x^4 + x^3 + x^2 + x^1 + 1
x^8 + x^6 + x^5 + x^1 + 1
x^8 + x^6 + x^5 + x^2 + 1
x^8 + x^6 + x^5 + x^3 + 1
x^8 + x^7 + x^6 + x^1 + 1
x^8 + x^7 + x^6 + x^5 + x^2 + x^1 + 1
```

alessandrokonrad: What about providing the lookup table also?

CIP-0067/README.md
```diff
@@ -0,0 +1,128 @@
...
+0 - 15                        | -     |  private use             |
+
+### Adding an entry to the registry
+
+To propose an addition to the registry edit the [registry.json](./registry.json) with your details, open a pull request against the CIPs repository and give a brief description of your project and how you intend to use the label for assets.
+
+## Rationale
+
+Asset name labels make it easy to classify assets. It's important to understand that an oblivious token issuer might use the prefix X for all kinds of things, leading to misinterpretation by clients that follow this standard. We can minimize this attack vector by making the label format obscure. Brackets, checksum and fixed size binary encoding make it unlikely someone follows this standard by accident.
+
```
KtorZ, line 99:
```suggestion
## Path to Active
```

CIP-0067/README.md
```diff
@@ -0,0 +1,128 @@
...
+## Reference Implementation(s)
```
KtorZ, line 100:
```suggestion
### Reference Implementation(s)
```

CIP-0067/README.md
KtorZ: + there should be similar criteria as for CIP-0068 here w.r.t to the path to active. Namely, there should be visible adoption in a few major explorer and/or wallets. Two or more DApps making use of the asset's labels would also be a good sign of activity.

CIP-0067/README.md
```diff
@@ -0,0 +1,128 @@
...
+## Motivation
+
+As more assets are minted and different standards emerge to query data for these assets, it's getting harder for 3rd parties to determine the asset type and how to proceed with it. This standard is similar to [CIP-0010](../CIP-0010), but focuses on the asset name of an asset.
+
+
+## Specification
+
+To classify assets the `asset_name` needs to be prefixed the following `4 bytes` binary encoding:
+```
+[ 0000 | 2 bytes label_num | 1 byte checksum | 0000 ]
```
KtorZ, line 24:
```suggestion
[ 0000 | 16 bits label_num | 8 bits checksum | 0000 ]
```

Since the bracket is given in bits, I find it more logical to speak about bits for the inner parts as well.

CIP-0067/README.md
```diff
@@ -0,0 +1,128 @@
...
+```
+- The leading and ending four 0s are brackets
+- `label_num` has a fixed size of 2 bytes (`Label range in decimal: [0, 65535]`). 
+If `label_num` < 2 bytes the remaining bits need to be padded with 0s.
```
KtorZ, line 28:
```suggestion
If `label_num` < 2 bytes the remaining bits need to be left-padded with 0s.
```

CIP-0067/README.md
```diff
@@ -0,0 +1,128 @@
...
+- `checksum` has a fixed size of 1 byte. The checksum is calculated by applying the [CRC-8](#CRC-8) algorithm on the `label_num (including the padded 0s)`. 
+
+### CRC-8
+
+- Check: `0xF4`
+- Poly: `0x07`
+- Init: `0x00`
+- RefIn: `false`
+- RefOut: `false`
+- XorOut: `0x00`
```
KtorZ, line 38: I am not sure what these refer to exactly (I mean, I have only seen this notation in one particular implementation of the crc-8 algorithm in Go after searching for those specific terms). I seems that the literature mostly refer to either the lookup table and/or the polynomial representation. So it's probably sufficient to simply mention:

- Polynomial representation (normal): `0x07`
- Lookup table: ...

CIP-0067/README.md
```diff
@@ -0,0 +1,133 @@
+---
+CIP: 67
+Title: Asset Name Label Registry
+Authors: Alessandro Konrad <alessandro.konrad@live.de>, Thomas Vellekoop <thomas.vellekoop@iohk.io>
+Status: Draft
```
rphair, line 5:
```suggestion
Status: Proposed
```
@alessandrokonrad I think if we are going to merge this with your recently added Path to Active then it should be `Proposed` like CIP-0068 as well (cc @KtorZ)

KtorZ (approved):
@rphair @SebastienGllmt @crptmppt 

As discussed in editors meeting #54 (iirc), this proposal has now addressed the different points that were raised. I am approving, and happy to merge as `Proposed` while implementation are being worked on.

rphair (approved):
great stuff 🤩

KtorZ:
We're going _too fast_, that's a premiere for the CIP process 😄

PR #300, opened 2022-07-15
CIP-0381 | post-merge update top level README

rphair:
post merge https://github.com/cardano-foundation/CIPs/pull/220

(BTW this also shows some removal of extraneous white space in the diff)

### DIFF
--- a/README.md
+++ b/README.md
@@ -57,7 +57,8 @@ The entire process is described in greater detail in [CIP1 - "CIP Process"](./CI
 | 36 | [Catalyst/Voltaire Registration Transaction Metadata Format](./CIP-0036) | Proposed | 
 | 40 | [Collateral Output](./CIP-0040) | Proposed | 
 | 42 | [New Plutus Builtin: serialiseBuiltinData](./CIP-0042) | Proposed | 
-| 52 | [Cardano Audit Best Practice Guidelines](./CIP-0052) | Proposed |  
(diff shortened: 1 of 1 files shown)


KtorZ:
> what procedure did you guys work out for editing the top level table

I do grep with my eyes. It's cumbersome but relatively effective :grimacing: ... I typically do this before every bi-weekly to keep the top-level README updated. That usually takes me about 2-3 min and thus so far, I haven't felt the urge to automate this.

rphair:
@KtorZ @crptmppt @mangelsjover @SebastienGllmt what do you think of this process: immediately before merging each CIP, we make a quick edit to the top level README in the author's branch to update the table entry?

As we talked about here (https://github.com/cardano-foundation/CIPs/pull/216#issuecomment-1149147528) this is something we generally have to be diplomatic about, but shouldn't be a problem in this case because we're not changing the CIP author's documents.

This way, the top level table gets updated immediately, and doesn't get out of date (so doesn't need to be date-stamped) or need visual update passes periodically.

Also there was a merge conflict here yesterday during the meeting from something else updating a nearby entry in the table & so I lost my opportunity to ask someone to review this while we were all together... this would be less likely to happen if the top level README updates were always done immediately.

In any case can someone please approve this now so it can be merged? 🤓

KtorZ:
@rphair I'll include this one as part of a bigger update.

PR #305, opened 2022-07-25
updates to CIP meeting links and text since Discord migration

rphair:
Some things have gone out of date on the top-level `README` mostly due to the migration from Crowdcast to Discord:

- The link to the meeting forum itself has changed.
- Since this document was written, meetings are transcribed as well as summarised.
- We aren't making the recordings of the Discord meetings available yet (right?) so there's nothing to link to under "recorded".
- Some minor cleanup: grammar and the index character 🛈 before the blockquote: which sometimes renders as "info" (i-in-a-circle) but shows up as tofu on my own system so isn't universally valid.

Please let me know if we should still include the Crowdcast link because of the old meeting recordings.  I've not included it because there was talk in one of our last process meetings about dropping the Crowdcast subscription, which would drop that archive along with it (?).  If necessary I'll work that link back into the text as appropriate.

### DIFF
--- a/README.md
+++ b/README.md
@@ -70,7 +70,7 @@ The entire process is described in greater detail in [CIP1 - "CIP Process"](./CI
 
 ### Proposals Under Review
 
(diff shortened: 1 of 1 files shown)


KtorZ (approved):

SebastienGllmt (approved):
LGTM

PR #308, opened 2022-07-27
CIP-2551? | Ed25519 Elliptic Curve Group Primitives in Plutus Core

morganthomas:
This proposal adds Plutus primitives for efficiently computing with Ed25519 curves.

### DIFF
--- a/CIP-2551/CIP-2551.md
+++ b/CIP-2551/CIP-2551.md
@@ -1,0 +1,271 @@
+---
+CIP: ?
+Title: Ed25519 elliptic curve group primitives
+Authors: Morgan Thomas <morgan.thomas@platonic.systems>
+Discussions-To: <morgan.thomas@platonic.systems>
+Status: Draft
+Type: Standards Track
+Created: 2022-07-27
+* License: Apache-2.0
+---
+
+## Preamble
+
+RFC 822 style headers containing metadata about the CIP, including the CIP
+number, a short descriptive title (limited to a maximum of 44 characters),
+the names, and optionally the contact info for each author, etc.
+
+## Simple Summary
+
+This CIP proposes that Plutus be extended with native code
+implementations of certain mathematical operations which are useful for
+cryptography. These mathematical operations are used in the implementation
+of the Ed25519 public/private key signing protocol. Cardano already
+supports a native code implementation of verifying signatures created by
+the Ed25519 signing protocol, but it does not enable Plutus to scripts
+perform the underlying mathematical operations directly. Being able to do
+this would be useful for implementing other cryptographic protocols based
+on the Ed25519 mathematical operations.  Applications include verifying
+zero-knowledge proofs, such as in the context of implementing zk-rollups.
+
+## Abstract
+
+Extend Plutus with primitives for performing calculations over the
+Ed25519 curve and the corresponding finite field.  These operations are
+to include the basic group and field operations, as well as multiscalar
+multiplication.
+
+## Motivation
+
+CIP-0381 proposes inclusion of the Bls12-381 curve in Plutus. Bls12-381
+is a pairing-friendly curve. Not all applications of elliptic curves
+require a pairing-friendly curve. Choosing a non-pairing-friendly
+curve where a pairing-friendly curve is not required can lead to better
+usage of computing power.  The Ed25519 curve allows for more efficient
+operations as compared to the Bls12-381 curve.  Pairing-friendly curves
+require the use of more bits in order to achieve security (255 bits
+for Ed25519 vs 381 bits for Bls12-381). Therefore it would be helpful,
+for those applications which do not require a pairing-friendly curve,
+to include a non-pairing-friendly curve in Plutus.
+
+An example application of the Ed25519 curve is in verifying
+zk-SNARKs on-chain. Halo 2, for example, can be used with the Ed25519
+curve as the underlying curve. In its standard incarnation, Halo
+2 uses the Pasta curves (Pallas and Vesta). This is a 2-cycle of
+non-pairing-friendly curves.  The Pasta curves are specialized to a
+particular recursive proof strategy, as described in [Section 4.6 of the
+Halo 2 book](https://zcash.github.io/halo2/background/recursion.html/).
+For non-recursive proofs, the Pallas and Vesta curves are not required,
+and the Ed25519 curve suffices.
+
+Halo 2 over the Ed25519 curve offers short proofs and relatively low
+proof computational complexity of proof verification. These are the
+characteristics we are looking for in a zk-SNARK theory optimized for
+on-chain proof verification.
+
+Recursive proofs are very important in some applications of zk-SNARKs:
+zk-rollups, in particular.  Recursive proofs can be homogeneous (verifying
+a zk-SNARK produced by theory A in a zk-SNARK produced by theory A), or
+heterogeneous (verifying a zk-SNARK produced by theory A in a zk-SNARK
+produced by theory B).
+
+Halo 2's recursive Pasta curves proof strategy is heterogeneous: it
+uses proofs over the Pallas curve to verify proofs over the Vesta curve,
+and proofs over the Vesta curve to verify proofs over the Pallas curve.
+
+Orbis Labs has been researching ways of building recursive proofs
+off-chain and verifying them on-chain, for the purposes of zk-rollups. One
+way to do this is to use FRI-based zk-SNARKs for heterogeneous recursive
+proofs. A downside of this is that proof sizes are expected to be
+substantially larger than the Cardano transaction size limit when using
+FRI, necessitating multiple transactions to verify a proof. Another
+downside of this is that for recursive FRI proofs, it makes sense to use
+a circuit-friendly hash (e.g., Poseidon or Rescue), but Cardano does not
+currently support these natively. Cardano-friendly hashes are those which
+are implemented as Plutus primitives: SHA2-256, SHA3-256, and Blake2b-256.
+
+The issue is that currently there is no hash function which efficient
+both in-circuit and on-chain.  A relatively simple workaround for this
+issue with homogeneous FRI recursion is to use heterogeneous recursive
+proofs, where an on-chain proof is generated using FRI with over a
+Cardano-friendly hash function, and it recursively verifies a proof
+generated using FRI over a circuit-friendly hash function. With this
+change of hash function heterogeneous recursive proving pattern, we can
+have efficient hashes both in-circuit and on-chain.
+
+More generally, we can decouple the choice of zk-SNARK theory for on-chain
+proof verification from the choice of zk-SNARK theory for off-chain
+recursive proof building, as long as we can reasonably efficiently
+convert one form of zk-SNARK to the other using recursive proving.
+
+For efficient off-chain proving, it makes sense to use a FRI-based
+zk-SNARK theory, since this leads to relatively fast proving times,
+at the cost of larger proof sizes, compared to Halo 2 for example. This
+also has the advantage that recursive proofs are relatively simpler to
+implement, because they do not require an accumulation scheme over a
+2-cycle of curves. FRI allows for homogeneous recursive proving.
+
+For efficient on-chain proof verification, it would make sense to use
+an elliptic curve based zk-SNARK theory, such as Halo 2, since this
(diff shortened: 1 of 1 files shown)


L-as:
I do wonder if we could somehow support arbitrary curves, e.g. some kind of curve point type that is parameterised by the properties of the curve, but I'm not sure how efficient such an implementation could be.

It seems like we can possibly reuse the approach, and possibly also the implementation, of https://hackage.haskell.org/package/elliptic-curve.

CC @michaelpj @iquerejeta

L-as:
For pairings: https://hackage.haskell.org/package/pairing

https://github.com/orgs/adjoint-io has a lot of work in this field, there's no need for us to reinvent the wheel in a bad way.

iquerejeta:
> https://github.com/orgs/adjoint-io has a lot of work in this field, there's no need for us to reinvent the wheel in a bad way.
We would need to have a closer look to these implementations. For instance, we would need to know the reliability of these implementations. Have they been used elsewhere? Are they audited and by who? Have they been battle tested? Also, are they efficient? 

Nonetheless, even if some of the questions above are answered negatively, we could use the same ideas to make the API generic, which uses more efficient backends (and not the haskell one).

michaelpj:
I think the main barrier to a generic implementation is working out what the interface would look like in PLC _and how it would be costed_.

I'm a Haskell programmer, I love genericity as much as anyone. But sometimes it's best to do things the dumb way.

L-as:
We can probably figure out some relation between curve parameters and cost. We should probably at least figure out whether it's feasible or not before adding more curves (after BLS381).

CIP-2551/CIP-2551.md
```diff
@@ -0,0 +1,271 @@
...
+
+## Abstract
+
+Extend Plutus with primitives for performing calculations over the
+Ed25519 curve and the corresponding finite field.  These operations are
+to include the basic group and field operations, as well as multiscalar
+multiplication.
+
+## Motivation
+
```
iquerejeta, line 39: We also need a motivation of why we use a curve with cofactor 8 (such as the edwards curve) instead of one with cofactor 1 which is usually safer. This is for efficiency, as edwards curves are much more efficient. Nonetheless, I think it is necessary to explain the danger of having a group with cofactor 8. This makes me wonder if we also need built-in functions to check if the points are part of the prime order group or not. What do you think?

CIP-2551/CIP-2551.md
```diff
@@ -0,0 +1,271 @@
...
+b0 denote the least significant bit, b1 the second least significant,
+and b2 the third least significant.
+
+ * b0 is 1 if and only if the group element is in compressed form. 
+ * b1 is 1 if and only if the element is the point at infinity.
+   (The rest of the encoding after b1 should be zero in case b1 1 is 1.)
+ * b2 is 1 if and only if the element is in compressed form and not the
+   point at infinity and the point is "positive," meaning that its y
+   coordinate is lexicographically greater than the y coordinate of the
+   other point on the curve with the same x coordinate.
```
iquerejeta, line 210: I now see where the "extra" byte comes in from above, but it still should be 33 in compressed form and 65 in uncompressed form. However, I don't think this is the most efficient representation. We can use a representation that requires only 32 and 64 bytes. Also, due to the fact that we need to cost functions, I would stick to a single serialised representation (possibly the compressed one). Otherwise it would be hard to cost the function as the cost would differ for compressed and uncompressed group elements.

CIP-2551/CIP-2551.md
```diff
@@ -0,0 +1,271 @@
...
+byte containing the three flags, followed by two field element encodings,
+first the x coordinate and then the y coordinate.
+
+## Rationale
+
+The inclusion of `Ed25519_scale_and_add` is to allow for an efficient
+implementation of taking linear combinations of group elements. This
+is important for efficient proof verification for Halo 2's inner
+product argument. For example, this operation can be implemented using
+[Pippenger's algorithm](https://jbootle.github.io/Misc/pippenger.pdf).
```
iquerejeta, line 224: or Strauss algorithm would also be an option, depending on the size of the vector size.

CIP-2551/CIP-2551.md
```diff
@@ -0,0 +1,271 @@
...
+```
+
+and that for all integers `x`,
+
+```
+Ed25519_F_to_integer (Ed25519_F_from_integer x) ≡ x mod 2^255 - 19.
+```
+
+Provide property tests which check that the equality comparisons are
+equivalence relations (relexive, transitive, and symmetric).
```
iquerejeta, line 259:
```suggestion
equivalence relations (reflexive, transitive, and symmetric).
```

CIP-2551/CIP-2551.md
```diff
@@ -0,0 +1,271 @@
...
+are that
+
+```
+Ed25519_F_from_integer . Ed25519_F_to_integer = id,
+```
+
+and that for all integers `x`,
+
+```
+Ed25519_F_to_integer (Ed25519_F_from_integer x) ≡ x mod 2^255 - 19.
```
iquerejeta, line 255: So from integer computes the modular reduction? If that is the case, it should be mentioned.

iquerejeta (commented):
I like the idea of introducing elliptic curve operations in plutus, as this would allow some interesting scripts. Thanks for opening the discussion. I've left some comments to be addressed, but I would also like the feedback from @davidnevadoc or @CPerezz on the motivation around the SNARKs. 

On that note, I think it is also important to provide estimates of Halo2 proofs over Edwards25519. It is said that using pasta curves would be too inefficient. Are we sure this is not the case for Halo2 proofs over edwards25519? Do we have numbers of proof sizes and verification times for certain circuit sizes?

L-as:
What's up with the horrible CIP enumeration scheme?

iquerejeta:
The numbers are only settled once the PR is accepted and merged. See [here](https://github.com/cardano-foundation/CIPs/pull/220#issuecomment-1148515170). As far as we are concerned, the numbering of this CIP is just temporal while we are doing the PR.

CIP-2551/CIP-2551.md
```diff
@@ -0,0 +1,271 @@
...
+curve where a pairing-friendly curve is not required can lead to better
+usage of computing power.  The Ed25519 curve allows for more efficient
+operations as compared to the Bls12-381 curve.  Pairing-friendly curves
+require the use of more bits in order to achieve security (255 bits
+for Ed25519 vs 381 bits for Bls12-381). Therefore it would be helpful,
+for those applications which do not require a pairing-friendly curve,
+to include a non-pairing-friendly curve in Plutus.
+
+An example application of the Ed25519 curve is in verifying
+zk-SNARKs on-chain. Halo 2, for example, can be used with the Ed25519
```
michaelpj, line 52: How many EC operations are needed to implement Halo proof verification? This was a concern already with the Bls primitives: the primitives themselves are already expensive, and if we need many of them to actually perform Halo verification, then we may still not have enough resources to do so. It would be sad to implement these primitives and then find out we still couldn't do the thing we actually wanted to do.

I'd also be interested in knowing how long it takes to verify a typical Halo 2 proof on typical consumer hardware. If it's in the order of milliseconds then we are in trouble.

CIP-2551/CIP-2551.md
```diff
@@ -0,0 +1,271 @@
...
+Add the following functions:
+
+ * Basic group operations:
+    * `Ed25519_G_add :: Ed25519GElement -> Ed25519GElement -> Ed25519GElement`
+    * `Ed25519_G_neg :: Ed25519GElement -> Ed25519GElement`
+ * Basic field operations:
+    * `Ed25519_F_add :: Ed25519FElement -> Ed25519FElement -> Ed25519FElement`
+    * `Ed25519_F_mul :: Ed25519FElement -> Ed25519FElement -> Ed25519FElement`
+    * `Ed25519_F_neg :: Ed25519FElement -> Ed25519FElement`
+    * `Ed25519_F_recip :: Ed25519FElement -> Maybe Ed25519FElement`
```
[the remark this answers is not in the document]
perturbing: My take on the `recip` function, it's the reciprocal (the multiplicative inverse) of the field. Why this is a `Maybe` is not clear to me, given an `a` in `F_q` the element `a^(q-2) mod q` is its inverse.

CIP-2551/CIP-2551.md
```diff
@@ -0,0 +1,271 @@
...
+
+ * `Ed25519GElement`: the type of elements of the Ed25519 curve.
+ * `Ed25519FElement`: the type of elements of the finite field of order
+    2^255 - 19.
+
+Add the following functions:
+
+ * Basic group operations:
+    * `Ed25519_G_add :: Ed25519GElement -> Ed25519GElement -> Ed25519GElement`
+    * `Ed25519_G_neg :: Ed25519GElement -> Ed25519GElement`
```
perturbing, line 152: These functions could fail, right? So, shouldn't they use a Maybe? Above, you mention that the `Ed25519GElement` is an Ed25519 curve element, but in practices this is just a wrapper of two field elements, it could be that one or both of these tuples do not lie on the curve.

L-as:
@KtorZ I don't think @morganthomas is interested in pushing this further along.

perturbing:
For fun and profit, I made a naive Ed25519 [implementation](https://github.com/perturbing/plutus-crypto) in plutus from the other builtins as described in this proposal.

In general, I made the field as a `Ring` instance in plutus and the EC curve as an `additiveGroup` instance (this does not add the field `recip` function natively). I might add some El Gamal stuff with some zk proofs. The proofs I want to try onchain are: proof of correct encryption (DLEQ), and proof of knowing a private key of a public key (DLOG). I think these are viable since these require only comparing and adding EC points and hashes. And if not, they might be only viable in a hydra head, it has some application for games.

I implemented most tests but stopped at the benchmarking of the implementation since it is not clear to me what operations will be used onchain for a Halo 2 proof. I hope that the `scale` function (adding the base point n times to itself) is not needed in these proofs, as this is a costly function onchain! Furthermore, I suppose this is not the case since it is only used for key generation? To reiterate, it's not a good idea to use this *naive* implementation for key generation.

I might improve the implementation by considering the projective representation of the points (I read that this is computational better). @iquerejeta maybe you know if this true/significant?

iquerejeta:
That's supper cool. Curious to see the outcome of the benchmarks. 

With respect to the point representation for scalar multiplication, current implementations do not use a single representation, but instead mix coordinate systems for more efficiency. Projective points allow for more efficient doubling of points, while 'extended representation' allow for more efficient addition. Therefore, when performing a scalar multiplication (computed out of a series of doubling and additions), one switches between the different representations. Another efficiency trick that one can find in current implementations, is that scalar multiplication with variable base (same holds with the base point), one first calculates a lookup table, and then proceeds with a series of point doublings and additions (see [here](https://github.com/dalek-cryptography/curve25519-dalek/blob/main/src/backend/serial/scalar_mul/variable_base.rs) for example). Some of these tricks are mentioned in the original [ed25519 paper](https://ed25519.cr.yp.to/ed25519-20110926.pdf), but you can also find more details in some of the implementations (I usually look [dalek's implementation](https://doc-internal.dalek.rs/curve25519_dalek/backend/serial/curve_models/index.html), which is very well documented, or [libsodium](https://github.com/jedisct1/libsodium), which is based in DJB's reference implementation). These tricks should give you a considerable improvement, and some of these tricks are also useful for computing multiscalar multiplication efficiently (`a_1 * P_1 + a_2 * P_2 + ... + a_n * P_n`, where `a_i` and `P_i` are scalars and points respectively). 

With respect to the scale function (`n * P`, or adding `P` `n` times to itself), these operations are required in both Halo2 and simpler zk proofs (such as DLEQ or DLOG). When you verify a DLOG proof, for `H = x * B`, the final check is `s * B ?= c *  H + R`, where `R, c, s` are the announcement, challenge and response of the zk proof (or sigma-protocol), and `?=` is an equality check.

KtorZ:
Closing this PR for the sake of housekeeping. @morganthomas if you have any interest in pursuing this please re-open.

KtorZ:
@iquerejeta you may also have an interest in championing this CIP instead?

rphair:
Note @iquerejeta @KtorZ @imikushin there's some interest expressed on the Cardano Forum in reviving this proposal: https://forum.cardano.org/t/cip-elliptic-curve-operations-and-other-math-useful-for-zk-proof-verification-in-plutus/113624/6

PR #314, opened 2022-08-12
CIP-0041? | UPLC Serialization Optimizations

HarmonicPool:
this CIP proposes some changes to the UPLC AST serialization in order to reduce the size of the serialized output

### DIFF
--- a/CIP-optimized-uplc-serialization/README.md
+++ b/CIP-optimized-uplc-serialization/README.md
@@ -389,7 +389,6 @@ using the new algorithm the ```ByteString``` serialization space complexity goes
 
 ### ```data``` serialization
 
-
 All the effort of minimizing the size of on-chain scripts by prefering ```flat``` over ```CBOR``` serialization are ignored when it comes to ```data``` serialization.
 
 citing the june 2022 plutus core specification ( Appendix B, Section B.2, Note 1; page 23 )
@@ -413,17 +412,8 @@ data Data =
     | B ByteString
 ```
 
-a more efficient implementation could be obtained by introducing some tags for the various construcors, each taking 3 bits:
-
-data constructor | binary | decimal
------------------|--------|--------
-Constr           | 000    | 0
-Map              | 001    | 1
-List             | 010    | 2
-I                | 011    | 3
-B                | 100    | 4
-
-directly followed by the respective serialized value; in particoular
+a more efficient implementation could be obtained by introducing some tags for the various construcors, each taking 3 bits ( option 1 in specification );
+directly followed by the respective serialized value; in particular
 
 - ```Constr``` is followed by an **unsigned** integer and the a list of ```Data```
 - ```Map``` is followed by a list of ```Data``` pairs; where the paris are serialized as in the constant pair specification (only the value)
@@ -431,39 +421,13 @@ directly followed by the respective serialized value; in particoular
 - ```I``` is followed by a **signed** integer
 - ```B``` is followed by a ```ByteString``` 
 
-An important **alternative design** that should be considered and discussed with the comunity would see the ```Data``` type requiring only 4 tags which would then be:
+An important **alternative design** is the one described in the option 2.
 
-data tag  | binary | decimal
-----------|--------|--------
-data-pair | 00     | 0
-data-list | 01     | 1
-data-int  | 10     | 2
-data-bs   | 11     | 3
+this alternative allows for 2 bits data tags for the ```Constr``` constructor and the simple values constructors ```I``` and ```B``` containitng the actual data;
 
-and the constructors would map to:
+and a 3 bit tag for ```List``` and ```Map``` constructors;
 
-- ```Constr``` becomes a ```data-pair``` expecting an unsigned integer and a list of ```Data```
-- ```Map``` becomes a ```data-list``` containing 2 ```Data``` per element
-- ```List``` becomes a ```data-list``` containing 1 ```Data``` per element
-- ```I``` becomes a ```data-int``` tag expecting a signed integer
-- ```B``` becomes a ```data-bs``` tag expecting a ```ByteString```
-
-this design is unambigous since ```data-list``` that maps back to two of the constructors contains different values depending on what Constructor is intended.
-
-the only edge case for the ```data-list``` constructor is the empty list, but the problem can be easly solved by expecting one single bit only in the case of the empty list indicating the intended constructor.
-
-so the empty ```Map``` would serialize as
-```
-01 # data-list tag
-0  # list nil construcor ( empty list )
-0  # bit indicating Map data constructor
-```
-and the empty ```List``` as
-```
-01 # data-list tag
-0  # list nil construcor ( empty list )
-1  # bit indicating List data constructor
-```
+this is done in order to minimize the number of bits per constructor without introducing unused tags (which is the case of a 3 bit tag for all constructors; tags 5, 6 and 7 would be unused) but still being capable of encoding all 5 constructors.
 
 the second design also implies no more ```Data``` constructors will be added in the future, or that for those added it will be possible to find an unambigous representation using the ones specified.
 
--- a/CIP-optimized-uplc-serialization/README.md
+++ b/CIP-optimized-uplc-serialization/README.md
@@ -1,0 +1,327 @@
+---
+CIP: "???"
+Title: optimized UPLC serialization  
+Authors: Michele Nuzzi <michele.nuzzi.2014@gamil.com> 
+Discussions-To: TBD
+Comments-Summary: TBD  
+Comments-URI: TBD  
+Status: Draft  
+Type: Standards Track  
+Created: <date created on, in ISO 8601 (yyyy-mm-dd) format>  
+License: CC-BY-4.0  
+License-Code: Apache-2.0  
+Post-History: <dates of postings to Cardano Dev Forum, or link to thread>  
+---
+
+## Abstract
+
+this document describes the parts of the current serialization algorithm that can be improved and provides the specification and documentation needed in order to implement an optimized version of this one.
+
+## Motivation
+
+all Untyped Plutus Core Programs do need to be serialized in order to be included in a transaction (therefore stored on-chain);
+
+the current serialization algorithm is based on the [flat](http://quid2.org/docs/Flat.pdf) algorithm;
+
+this allows for a bit-oriented serialized program; but there are aspects that can be improved to further reduce the size of the serialized program;
+
+the major benefit of a uplc-optimized algorithm is the possibility to store more complex logic (or more in general, more complex transactions) on chain without the need to increase the transaction size limit.
+
+## Specification
+
+### Table of contents
+
+- [```pad0``` should add nothing](#pad0)
+- [constant's types list should work as Integer ones](#constT)
+- [remove the "type application" tag for constant types](#tyApp)
+- [```ByteString```s encoded as unsinged integer followed by the bytes](#ByteStrings)
+- [```data``` serialization
+](#data)
+
+<a name="pad0"></a>
+
+### ```pad0``` should add nothing rather than an entire byte
+
+sometimes it is necessary to allign the serialized program (partial or total) to fit into a byte;
+
+in order to do so, some padding is added based on the current bit-length of the program;
+
+the needed padding can be obtained as follows
+```ts
+pad( 8 - (currBitLength % 8) )
+```
+where the ```pad``` function could be defined as follows in typescript
+```ts
+pad( missingBits: 0 | 1 | 2 | 3 | 4 | 5 | 6 | 7 ): string
+{
+    switch( missingBits )
+    {
+        case 0: return "00000001";
+        case 1: return "1";
+        case 2: return "01";
+        case 3: return "001";
+        case 4: return "0001";
+        case 5: return "00001";
+        case 6: return "000001";
+        case 7: return "0000001";
+    }
+}
+```
+the case in which the ```missingBits``` is 0 implies that the current serialized program is already byte-alligned and, since this padding carries no usefull informations, the current ```pad( 0 )``` adds a useless byte each time a padding is needed and the number of used bits is a multiple of 8.
+
+An alternative implementation would be
+```ts
+pad( missingBits: 0 | 1 | 2 | 3 | 4 | 5 | 6 | 7 ): string
+{
+    switch( missingBits )
+    {
+        case 0: return "";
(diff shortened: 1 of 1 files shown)


CIP-optimized-uplc-serialization/README.md
```diff
@@ -0,0 +1,327 @@
...
+1  # bit indicating List data constructor
+```
+
+the second design also implies no more ```Data``` constructors will be added in the future, or that for those added it will be possible to find an unambigous representation using the ones specified.
+
+## Rationale
+
+This proposal suggest ways to reduce serialized scripts size.
+
+the changes where designed to keep the bit oriented style of flat minimizing the number of required bits for values where possible.
```
KtorZ, line 313: The rationale section is a bit _thin_, though there's a bit of rationale sprinkled in various places of the specification. It'd be nice to have a summary of what motivates each design choices. This can most likely be done by reworking a bit how the specification is written (that is, write the specification without justifications, and move the justifications to the rationale section). 

For example, most of the "data serialization" section is about explaining what is currently wrong with the `Data` CBOR serialization. Instead, leave the specification focused to what the proposal is about, and move the rationale to this section. 
  
It'd be nice to consider some benchmarks? Taking some reference UPLC and serializing them side-by-side with the two methods to see how much gain one can expect.

HarmonicPool: Thank you @KtorZ; will refactor as suggested ( and correct the various typos 😄 )

CIP-optimized-uplc-serialization/README.md
```diff
@@ -0,0 +1,327 @@
...
+Type: Standards Track  
+Created: <date created on, in ISO 8601 (yyyy-mm-dd) format>  
+License: CC-BY-4.0  
+License-Code: Apache-2.0  
+Post-History: <dates of postings to Cardano Dev Forum, or link to thread>  
+---
+
+## Abstract
+
+this document describes the parts of the current serialization algorithm that can be improved and provides the specification and documentation needed in order to implement an optimized version of this one.
```
L-as, line 18: You probably want to clean up the formatting, e.g. use capital case at the beginning of sentences.

CIP-optimized-uplc-serialization/README.md
```diff
@@ -0,0 +1,327 @@
...
+        case 0: return "00000001";
+        case 1: return "1";
+        case 2: return "01";
+        case 3: return "001";
+        case 4: return "0001";
+        case 5: return "00001";
+        case 6: return "000001";
+        case 7: return "0000001";
+    }
+}
```
L-as, line 68: What language is this? Why not describe this in Haskell?

HarmonicPool: typescript just because I feel the description is more intuitive using it

it translates to haskell as
```hs
pad :: Int -> String
pad 0 = "00000001"
pad 1 = "1"
pad 2 = "01"
pad 3 = "001"
pad 4 = "0001"
pad 5 = "00001"
pad 6 = "000001"
pad 7 = "0000001"
pad _ = undefined
```

CIP-optimized-uplc-serialization/README.md
```diff
@@ -0,0 +1,446 @@
+---
+CIP: "???"
+Title: optimized UPLC serialization  
+Authors: Michele Nuzzi <michele.nuzzi.2014@gamil.com>
```
KtorZ, line 4:
```suggestion
Authors: Michele Nuzzi <michele.nuzzi.2014@gmail.com> 
```

CIP-optimized-uplc-serialization/README.md
```diff
@@ -0,0 +1,446 @@
...
+
+this is done in order to minimize the number of bits per constructor without introducing unused tags (which is the case of a 3 bit tag for all constructors; tags 5, 6 and 7 would be unused) but still being capable of encoding all 5 constructors.
+
+the second design also implies no more ```Data``` constructors will be added in the future, or that for those added it will be possible to find an unambigous representation using the ones specified.
+
+## Backward Compatibility
+
+the proposed changes to the algorithm will cause the same UPLC Abstract Syntaxt Tree to serialize in a different way based on the algorithm used;
+
+In order to allow the deserializaton process to handle the old serialization algorithm the version of the program sholud be chcked first.
```
KtorZ, line 438:
```suggestion
In order to allow the deserializaton process to handle the old serialization algorithm the version of the program should be checked first.
```

KtorZ:
cc @michaelpj

michaelpj:
This could probably have started life as a `plutus` issue, but this is also fine!

@kwxm can you take a look?

KtorZ:
> **Note** 
> 
> We'll be reviewing this proposal briefly in today's editor meeting but since this change concerns Plutus internals, the ultimate decision and approval is up to the current Plutus core team in accordance with [CIP-0035](https://github.com/cardano-foundation/CIPs/tree/master/CIP-0035#evolution-of-plutus-core-in-the-cardano-ledger). 
>
> cc @michaelpj @kwxm

michaelpj:
Apart from anything else, this could do with an impact assessment that assess how much of an improvement this is: actual numbers are very relevant.

CIP-optimized-uplc-serialization/README.md
```diff
@@ -0,0 +1,446 @@
...
+        case 4: return "0001";
+        case 5: return "00001";
+        case 6: return "000001";
+        case 7: return "0000001";
+    }
+}
+```
+the case in which the ```missingBits``` is 0 implies that the current serialized program is already byte-alligned
+
+since this padding carries no usefull informations, the current ```pad( 0 )``` adds a useless byte each time a padding is needed and the number of used bits is a multiple of 8.
```
kwxm, line 253: I think that the reason for this is so that the decoder can easily discard padding at the end without having to check whether it's at a byte boundary, and this makes it faster.  The proposed change would on average save one bit per bytestring, so I'm not sure if it's worth it.

CIP-optimized-uplc-serialization/README.md
```diff
@@ -0,0 +1,446 @@
...
+
+### remove the "type application" tag for constant types
+
+as per specification constant terms are typed using tags.
+
+these tags are serialized as 4 bits and currently so 16 options ( 0 to 15 inclusive ) of wich 9 are used as follows
+
+tags from ```integer``` to ```bool``` and the ```data``` one are directly followed by the respective value encoding;
+
+tags ```list``` and ```pair``` are the only tags that do require some other tag in order to be a defined type; since the twos always require some other type in order to be valid, the type application is implcit and it should be removed.
```
kwxm, line 341: Yes, that's a bit annoying: it's like that because it reflects how the internal representations of types work.  The extra generality might permit us to do some more complex things in future, but it's not clear if we'll ever actually need that.  Simplifying this might be a good thing to do, since it complicates the encoding/decoding process as well as taking up extra room.

CIP-optimized-uplc-serialization/README.md
```diff
@@ -0,0 +1,446 @@
...
+```
+1) Unsigned (indefinite length) Integer indicating how many bytes ( or chunks of 8 bits ) are following
+2) as many bytes as specified in the Unsigned Integer at step 1
+```
+
+using the new algorithm the ```ByteString``` serialization space complexity goes form ```O(n)``` to ```O(log n)``` where ```n``` is the number of bytes in the ```ByteString``` 
+
+### ```data``` serialization
+
+All the effort of minimizing the size of on-chain scripts by prefering ```flat``` over ```CBOR``` serialization are ignored when it comes to ```data``` serialization.
```
kwxm, line 392: Yes, it would be good to have a more efficient encoding of `Data` rather than just wrapping the CBOR representation.  However, I think the reason for this is that `Data` is used elsewhere on the chain (including things that are passed to validator scripts as parameters), and CBOR is the preferred format there.  It might be difficult to switch because of this.

CIP-optimized-uplc-serialization/README.md
HarmonicPool: if the proposed changes are accepted the only place where padding will be used is at the end of the script; since bytestring will no longer require to be byte alligned

the check for ```pad( 0 )``` then becomes just ```lengthInBits `mod` 8 == 0```

CIP-optimized-uplc-serialization/README.md
HarmonicPool: conversion from CBOR to a specific data encoding can be done by the node prior to being passed as argument;

CBOR encoding is definitely to expansive to be included in a script

HarmonicPool:
> 



> Apart from anything else, this could do with an impact assessment that assess how much of an improvement this is: actual numbers are very relevant.

@michaelpj will work on a CIP implementation these days to have a comparison in terms of size

CIP-optimized-uplc-serialization/README.md
```diff
@@ -0,0 +1,446 @@
...
+```
+1) some padding that, depending on the context, may take up to 1 byte
+2) 1 (byte * chunk) indicating how many bytes are in the very next chunk
+3) 0 to 255 meaningful (bytes * chunk)
+4) 1 useless byte indicating no more chiunks are following
+```
+this implies ```(~1) + #chunks + 1``` meaningless bytes are added per ```ByteString```
+
+in the descripton above
+- step ```1``` allows for an easy serialization and deserialization but doesn't carries any meaningful informations; given the importance of ByteStrings it should be removed at the cost of an added shift while serializing/deserializing the value
```
kwxm, line 378: I think that might slow things down significantly, since I think we'd need to shift every byte in a bytestring if it didn't start at a byte boundary.  It's important that on-chain deserialisation be as fast as possible, so the extra expense might be problematic.  I'd like to see some benchmark results for this.

CIP-optimized-uplc-serialization/README.md
```diff
@@ -0,0 +1,446 @@
...
+- step ```4``` can be removed as it is simply useless
+- step ```3``` can be optimized using unsigned integer encoding
+
+the new algorithm for ```ByteString``` serialization would consist of two parts:
+```
+1) Unsigned (indefinite length) Integer indicating how many bytes ( or chunks of 8 bits ) are following
+2) as many bytes as specified in the Unsigned Integer at step 1
+```
+
+using the new algorithm the ```ByteString``` serialization space complexity goes form ```O(n)``` to ```O(log n)``` where ```n``` is the number of bytes in the ```ByteString```
```
kwxm, line 388: Hmmm.  I need to think more carefully about this.  It's a while since I've thought about bytestring encodings and I'll need to remind myself of the details of how we currently do it.

kwxm (commented):
The ideas here aren't nonsensical, but I'd like to see some benchmark results to see how they affect (a) the space taken up by serialised scripts, and (b) deserialisation times.  I think that the benefits would have to be quite big in order to accept this.

I'll add that we're aware that there's room for improving things, but the reason that things are the way that they currently are is because we're using a pre-existing library and that's the way that it does things.  We're definitely not averse to changing things, but then we'd need to maintain our own version of the `flat` library,

Apologies for a slightly hurried review: this had slipped off the bottom of my list.  I'll think a bit more about this, in particular the bytestring encoding.

CIP-optimized-uplc-serialization/README.md
HarmonicPool: It might be possible to implement deserialization without shifting bytes for efficiency

I'll leave here the [plu-ts implementation of the `readNBits`](https://github.com/HarmonicPool/plu-ts/blob/54332ac5716bcd0ddf51d512388a0cc74e904dde/src/onchain/UPLC/UPLCDecoder/index.ts#L142) function on which the deserialization process is based

I believe a very similar result can be achieved in Haskell using the [`Data.Binary.Get` monad](https://hackage.haskell.org/package/binary-0.10.0.0/docs/Data-Binary-Get.html)

Crypto2099:
Wondering if this is still an issue and needs to remain open given the advancements of Plutus v2/v3 and Plu.TS, Aiken, et al in the ensuing years since this CIP was opened. @michele-nuzzi what do you think?

michele-nuzzi:
Theoretically it could still be something worth exploring

Even with Aiken and plu-ts reducing size as much as possible Devs still tend to max out script size in an effort to reduce recursion and handle many inputs

However I or my team are not able to focus on this in the short term

If someone else wants to take over the CIP I'm ok with that

rphair:
thanks for update @michele-nuzzi ... @Ryun1 @Crypto2099 I'll remove this & others from the `Waiting for Authors` table in the `README` in the upcoming biweekly update.

colll78:
Improvements to the cost of deserialization are an extremely relevant topic now with the introduction of the cost per reference script byte fee which has heavily reduced the revenue of most major dApp protocols on Cardano. It would be good if either this CIP could be revisited (and reworked to optimize for deserialization speed instead of serialization size), or a more general CPS could be made about improving the serialization format with less emphasis on script size and more emphasis on deserialization speed (developers would be happy to pay a much much higher one-time min ada fee for deploying large reference scripts if it meant a lower feePerRefScriptByte because the latter is a reoccurring fee and eats into protocol revenue)

michele-nuzzi:
@colll78 if you want to take it over I'm ok with it

L-as:
Isn't the serialization already quite efficient? Merkelization could be useful though. I made a CIP before even.

colll78:
> Isn't the serialization already quite efficient? Merkelization could be useful though. I made a CIP before even.

The issue is the cost of deserialization (introduced with the feePerReferenceScriptByte param) has severely reduced dApp revenue because it disproportionally impacts honest actors in order to prevent cheap spam by dishonest actors. For certain scripts (mostly deliberately malicious scripts) 99% of the time is consumed by preparation/deserialization, but for any non-malicious script (except for some extreme edge-cases) preparation/deserialization should only account for a very small portion of the overall execution time. The current fee is set the way it is to prevent cheap spam from malicious actors who deliberately craft scripts such that 99% of the execution time is deserialization, but a by-product of this is that honest actors (who submit scripts where deserialization is only a small minority of execution time) are punished unfairly because the fees they must pay are disproportionate to the amount of work done (execution time of the plutus scripts).

L-as:
The point is you wouldn't need to deserialize the whole thing with merkelization.

L-as:
Anyway intuitively and naively I feel like you can probably make the situation 10-100x better without changing the serialization format by just improving the deserialization code. JSON isn't particularly optimized either but the parsers for JSON are incredibly fast.

Really 95% of the performance problems can probably be fixed by just improving the implementation rather than doing architectural changes.

PR #317, opened 2022-08-18
Add Marlowe to metadata label registry

paluh:


SebastienGllmt (approved):

SebastienGllmt:
Approved it assuming somebody checks this is legitimate (it does look like it judging by the author's Github history)

rphair:
@SebastienGllmt @paluh if only for my own understanding, I'm waiting for an answer to https://github.com/cardano-foundation/CIPs/pull/317#issuecomment-1253709234 before approving it.

OlofBlomqvist:
More details around this PR can be found here:  https://github.com/input-output-hk/MIPs/discussions/5

paluh:
Oh, I'm really sorry about the state of this PR and about my late response (I missed the notification). Please let me clarify the situation. We (Marlowe Team) decided to reserve the label for Marlowe because we have few use cases (the final schema specs are not ready yet but we are working on them):

* Marlowe can be "compilation target" meaning it can be generated from different languages (currently we have in TypeScript, Go, Haskell and our extended DSL as well) - we want to be able to encode a link to the original "source code" and other related to source information on the metadata layer.

* We do some space optimization for large contracts (through subtrees merkleization), in such a case the rest of the contract has to be stored offchain - we want to provide a standard for linking this continuation pieces and use metadata layer for that.

* We want to have ability to tag Marlowe version. We have option to deploy using reference scripts which directly implies the Marlowe version but still users can create parametrized version of the interpreter and create self contained contract.

* We want to allow probably also include some optional metadata information about the contract itself - it is not yet clear what should go there but we have a discussion around that here: https://github.com/input-output-hk/MIPs/discussions/7

Once again sorry for crappy PR and the delay - I somewhat lost track myself of this PR ;-)

rphair (approved):
thanks @paluh, that works for me & it's nice to have this stuff on record in each PR that adds a new label 🤓

PR #318, opened 2022-08-19
CIP-0047? | Hardfork safety mechanism

JaredCorduan:
This CIP replaces a manual safety check regarding the readiness of the network for a hardfork with an automatic check.

Ever since the Shelley ledger era, block headers have included a protocol version indicating the maximum supported protocol version that the block producer is capable of supporting (see section 13, Software Updates, of the [Shelley ledger specification](https://hydra.iohk.io/job/Cardano/cardano-ledger/shelleyLedgerSpec/latest/download-by-type/doc-pdf/ledger-spec)).

This (semantically meaningless) field in the header provides a helpful metric for determining how many blocks will be produced after a hardfork, since nodes that have not upgraded will no longer produce blocks. (Nodes that have not upgraded will fail the `chainChecks` check from Figure 74 of the Shelley ledger specification, since the major protocol version in the ledger state will exceed the node's max major protocol version value, and hence can no longer make blocks.)

If most of the blocks in the recent past (e.g. the last epoch) are broadcasting their readiness for a hardfork, we know that it is safe to propose an update to the major protocol version which triggers a hardfork.

This CIP proposes automating this process, and making the protocol version in the header semantically _meaningful_. The ledger state will determine the stake (represented as the proportion of the active stake) of all the block producers whose last block contained the next major protocol version. Moreover, a new protocol parameter `hfThreshold` will be used to reject any protocol parameter update that proposes to change the major protocol version but does not have enough backing stake.

### DIFF
--- a/CIP-hardforkSafeguard/README.md
+++ b/CIP-hardforkSafeguard/README.md
@@ -10,8 +10,11 @@ License: CC-BY-4.0
 
 ## Simple Summary / Abstract
 
-This CIP replaces a manual safety check regarding the readiness of the network
-for a hardfork with an automatic check.
+This CIP proposes lifting a specific, manual safety check to the protocol.
+In particular, a new on-chain mechanism will replace the step in the off-chain
+hardfork procedure where the governance body gathers information about which
+stake pool operators have upgraded to the version of the software which
+supports an upcoming hardfork.
 
 Ever since the Shelley ledger era, block headers have included a protocol
 version indicating the maximum supported protocol version that the block
@@ -31,14 +34,14 @@ broadcasting their readiness for a hardfork,
 we know that it is safe to propose an update to the major protocol version
 which triggers a hardfork.
 
-This CIP proposes automating this process,
-and making the protocol version in the header semantically meaningful.
+This CIP proposes automating this specific check,
+making the protocol version in the header semantically meaningful.
 The ledger state will determine the stake
 (represented as the proportion of the active stake) of all the block producers
 whose last block contained the next major protocol version.
-Moreover, a new protocol parameter `hardforkThreshold` will be used to reject any
-protocol parameter update that proposes to change the major protocol version
-but does not have enough backing stake.
+Moreover, a new protocol parameter `hardforkThreshold` will be used to reject
+any protocol parameter update that proposes to change the major protocol
+version but does not have enough backing stake.
 
 ## Motivation / History
 
@@ -78,9 +81,18 @@ is not ready for any hardfork.
 When a new release is introduced which can handle an upcoming hardfork,
 the node will be configured to use the next major protocol version in the
 block header.
-Note also that there is no ambiguity regarding what the endorsement in the
+
+Note that there is no ambiguity regarding what the endorsement in the
 block header is referring to, since the major protocol version is only allowed
 to increase by one.
+Moreover, regardless of what update proposals the governance keys have
+proposed, each block header indicates that the corresponding block producer is
+either prepared for the major protocol version to increase, or that it is not
+prepared for it to increase.
+This CIP does not address the problems that arise from multiple versions of
+the software (potentially with different semantics) broadcasting the same
+major protocol version. These problems will have to be addressed as progress
+is made towards a full decentralized governance.
 
 Whenever the major protocol version is updated, the set of endorsements is
 reset to the empty set.
@@ -203,4 +215,4 @@ hardfork is immanent.
 
 ## Copyright
 
-This CIP is licensed under Apache-2.0
+This CIP is licensed under [CC-BY-4.0](https://creativecommons.org/licenses/by/4.0/legalcode)
--- a/CIP-hardforkSafeguard/README.md
+++ b/CIP-hardforkSafeguard/README.md
@@ -1,0 +1,193 @@
+---
+CIP: ?
+Title: Hardfork Safeguard via Stake Representation
+Authors: Jared Corduan <jared.corduan@iohk.io>
+Status: Draft
+Type: Standards Track
+Created: 2022-06-30
+License: CC-BY-4.0
+---
+
+## Simple Summary / Abstract
+
+This CIP replaces a manual safety check regarding the readiness of the network
+for a hardfork with an automatic check.
+
+Ever since the Shelley ledger era, block headers have included a protocol
+version indicating the maximum supported protocol version that the block
+producer is capable of supporting (see section 13, Software Updates, of the
+[Shelley ledger specification](https://hydra.iohk.io/job/Cardano/cardano-ledger/shelleyLedgerSpec/latest/download-by-type/doc-pdf/ledger-spec)).
+This provides a helpful metric for determining how many blocks will be produced
+after a hardfork, since nodes that have not upgraded will no longer produce blocks.
+(Nodes that have not upgraded will fail the `chainChecks` check from Figure 74
+of the Shelley ledger specification, since the major protocol version in the
+ledger state will exceed the node's max major protocol version value,
+and hence can no longer make blocks.)
+
+If most of the blocks in the recent past (e.g. the last twenty-four hours) are
+broadcasting their readiness for a hardfork,
+we know that it is safe to propose an update to the major protocol version
+which triggers a hardfork.
+
+We propose automating this process.
+The ledger state will determine the stake
+(represented as the proportion of the active stake) of all the block producers
+whose last block contained the next major protocol version.
+Moreover, a new protocol parameter `hfThreshold` will be used to reject any
+protocol parameter update that proposes to change the major protocol version
+but does not have enough backing stake.
+
+## Motivation / History
+
+Currently, the governance key holders collectively agree to increase the major
+protocol version.  This allows them to make a human judgment as to the
+readiness of the network for a hardfork (using the mechanism described above).
+Since only a few, well-aligned parties are involved, this is currently easy to
+coordinate.  As we move into the Voltaire phase, where the governance of the
+network is decentralized,
+it is imperative that we codify this human judgment in the protocol itself.
+
+## Specification
+
+### New Protocol Parameter
+
+There will be a new protocol parameter named `hfThreshhold`,
+containing a rational number between zero and one.
+
+### Tracking Hardfork endorsements
+
+The ledger state will maintain a set of stake pool IDs corresponding to the
+block producers whose last block endorsed the next major protocol version.
+Endorsing here means that the major protocol version in the block header is
+exactly one more than the current major protocol version in the ledger state.
+Note that the protocol version in the block header is set by the particular
+cardano-node release being used by the block producer.
+When no hardfork is anticipated, the node will be configured to place the
+current major protocol version in the block header, indicating that the node
+is not ready for any hardfork.
+When a new release is introduced which can handle an upcoming hardfork,
+the node will be configured to use the next major protocol version in the
+block header.
+
+Whenever the major protocol version is updated, the set of endorsements is
+reset to the empty set.
+
+In order to track the endorsements, the `TICK` ledger rule will need two items
+added to the environment (since the Shelley era, the `TICK` rule has had an
+empty environment).
+In particular, it will need the following from the block header:
+* The pool ID of the block producer
+* The major protocol version
+
+### Rejecting Updates
+
+The main point of the safeguard introduced in this CIP is the ability to reject
+protocol parameter updates which propose to increase the major protocol version
+when not enough block producers are prepared.
+The rejection will happen in both the consensus layer and the ledger layer.
+
+The timing of the rejection is critical, and requires understanding a bit about
+the timing of the hardfork combinator (see the diagram below).
+Ouroboros (Praos and Genesis) have a notion of a stability window,
+corresponding to the duration of slots after which the consensus mechanism will
+no longer roll back a block.
+The stability window is currently three tenths of the epoch length
+(36 hours on mainnet).
+The hardfork combinator requires that the changes to the ledger state which
+enact a hardfork (confirmed proposals to increase the major protocol version)
+be stable two stability windows before the end of the epoch.
+See section 17.4, Ledger restrictions, of the
+[consensus report](https://github.com/input-output-hk/ouroboros-network/tree/314845c4087bc6e662d7df0d376ab1910a5b5476/ouroboros-consensus/docs/report).
+Therefore protocol parameter updates for the next epoch boundary must be
+submitted during the first four tenths of the epoch.
+Call this first four tenths of each epoch the "proposal window" for the
+purposes of this document.
+The consensus layer
+[analyzes](https://github.com/input-output-hk/ouroboros-network/blob/314845c4087bc6e662d7df0d376ab1910a5b5476/ouroboros-consensus-shelley/src/Ouroboros/Consensus/Shelley/Ledger/Inspect.hs#L77-L96)
+the ledger state one stability window after the proposal window has ended to
+determine if the major protocol version will be increased at the next epoch
+boundary.
+The ledger itself does not apply the protocol parameter update until the
+epoch boundary.
+
+To apply the new safeguard, the consensus layer will now use new logic for
+determining if the major protocol will be increased, and the ledger will
+need to use the exact same logic on the epoch boundary.
+The new logic will take the same parameters that are currently being
+used to make the determination.
+See [protocolUpdates](https://github.com/input-output-hk/ouroboros-network/blob/314845c4087bc6e662d7df0d376ab1910a5b5476/ouroboros-consensus-shelley/src/Ouroboros/Consensus/Shelley/Ledger/Inspect.hs#L106-L110),
+and note that the set of endorsements, the stake distribution, and the
+protocol parameters will all included in what the consensus layer calls the
+`LedgerState`, and what the ledger layer calls the `NewEpochState`
+(the endorsements will be added to `LedgerState`, but the pool stake
+distribution and the protocol parameters are already included).
+
+The new logic for determining if the major protocol version will change is:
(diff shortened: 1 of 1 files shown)


rphair (commented):
I am 100% behind this, pending the other editors being satisfied with the technical details.  It represents a huge improvement in the decentralisation of network control by the governance key holders: a challenging problem to say the least.  From my own point of view I can say that stake pool operators will feel positively about being included in the governance process as described. 😎

kieransimkin:
Gets my support - maybe should be combined with a bit of a rethink of the release tagging process on github too? I'd like to see a community vote before a release candidate is tagged

CIP-hardforkSafeguard/README.md
```diff
@@ -0,0 +1,193 @@
...
+Since only a few, well-aligned parties are involved, this is currently easy to
+coordinate.  As we move into the Voltaire phase, where the governance of the
+network is decentralized,
+it is imperative that we codify this human judgment in the protocol itself.
+
+## Specification
+
+### New Protocol Parameter
+
+There will be a new protocol parameter named `hfThreshhold`,
```
SebastienGllmt, line 54: Note that making `hfThreshhold` a protocol parameter means that the quorum required for a change to protocol parameters has to be equal to or greater than the requirement to trigger a hardfork
Otherwise, you could first make a protocol parameter vote to lower `hfThreshhold` and then do the hardfork afterwards

CIP-hardforkSafeguard/README.md
```diff
@@ -0,0 +1,193 @@
...
+containing a rational number between zero and one.
+
+### Tracking Hardfork endorsements
+
+The ledger state will maintain a set of stake pool IDs corresponding to the
```
SebastienGllmt, line 59: One problem this doesn't address is what if you have competing proposals? The scheme described in this CIP only allows for sequential votes. The way this CIP is structure, I don't think this even allows you to do one vote after another because the current structure doesn't have an expiry on upgrade proposals (other than maybe skipping version numbers if a version is deemed to have failed)

CIP-hardforkSafeguard/README.md
JaredCorduan: I completely get and empathize with this concern, but it does come at the cost of extra complexity (to an otherwise pretty bare-bones change). I'm not sure I understand what "has to be equal to or greater than the requirement" means (the quorum and the threshold have different units, they are not comparable), but I can guess that you would like the SPOs to be able to signal support for changing the `hfThreshhold` (the way the protocol version is done today). This would require a change to the block header, and we'd need to think about how it would be configured. The broadcast protocol version is baked into the node, and requires a software update to change (which is natural, given that it relates to a fork). Would we require a software upgrade to endorse a change to the `hfThreshhold`, or make it configurable?

---

Without addressing this concern, it makes `hfThreshhold` more like all the other protocol parameters that also greatly effect the SPOs, like `k`, the fees, etc. The governance structure has to openly, on chain for everyone to see, declare that a change is being made. And it is the SPOs who ultimately run the network. Decentralized governance is a huge topic that we'll be getting into next, but in the short term we can think about where to draw the line on small changes that we could implement sooner rather than later. My bias is toward being as minimal as possible, but I'm happy for folks to disagree!

CIP-hardforkSafeguard/README.md
JaredCorduan: There are strict rules on how the protocol version can be increased:

https://github.com/input-output-hk/cardano-ledger/blob/7e2f674d2a2d14752d4c2d5abf60b26ae015b9e2/eras/shelley/impl/src/Cardano/Ledger/Shelley/PParams.hs#L519-L520

```
(m + 1, 0) == (m', n') || (m, n + 1) == (m', n')
```

Either the major is increased by exactly one (and the minor reset to zero), or the minor is increased by exactly one (and the major remains unchanged).

Moreover, this proposal is only putting in a safeguard for hardforks (major number increase). So it is always clear what the broadcasted protocol version in the block header is referring to.

Maybe this is only clear if I also explain how the existing protocol parameter update system works? During the voting window, each goverance key can propose a change (they can submit multiple proposals, but the latest one overrides the previous) for the end of the current epoch. If quorum is met, the change happens, otherwise nothing happens and the voting state resets. After the voting window, the each goverance key can stage a vote for the next epoch, which behaves exactly as though they waited until the next epoch and placed a vote during the next voting window.

---

> current structure doesn't have an expiry on upgrade proposals

I think I've explained this above as well. The current structure has harsh expirations. or am I misunderstanding what you meant?

CIP-hardforkSafeguard/README.md
psychomb: Intuitively, `hfThreshold` has a different feel than any other parameter. To me, it resembles an article of a constitution. Because of that, it would require a larger majority to get modified.

Without an protective mechanism (`hfThresold2`?), its value could go down over time, and there is surely a point where it starts getting dangerously low.

michael-liesenfelt:
If threshold is a parameter community voting could lower a super-majority to a simple majority. To maintain the security of this parameter it should not be subject to simple majority governance. It may be wise to require a hard fork, not a vote, to change the hard fork threshold. Maybe it becomes a global constant instead of a parameter?

On programming style I oppose the abbreviation legacy of punch cards and tapes. We are not memory space or character limited. It is ideal to make all software less 'coded' and more human readable. Type out the words and use: hardfork_threshold.

beriardas:
Great proposal

JaredCorduan:
> If threshold is a parameter community voting could lower a super-majority to a simple majority. To maintain the security of this parameter it should not be subject to simple majority governance. It may be wise to require a hard fork, not a vote, to change the hard fork threshold. Maybe it becomes a global constant instead of a parameter?

If we decide that we prefer to make this value hard to change, hard-coding it is a nice and simple solution. I'll just reiterate that nearly all of the [protocol parameters](https://github.com/input-output-hk/cardano-ledger/blob/7e2f674d2a2d14752d4c2d5abf60b26ae015b9e2/eras/babbage/impl/src/Cardano/Ledger/Babbage/PParams.hs#L107-L152) are capable of having a dramatic change to the network, and that what we really need is a holistic solution for governing _all_ of them.
> 
> On programming style I oppose the abbreviation legacy of punch cards and tapes. We are not memory space or character limited. It is ideal to make all software less 'coded' and more human readable. Type out the words and use: hardfork_threshold.

I agree that spelling it out is a much better idea. I picked up this habit after a few years of writing latex specs for the Cardano ledger, where horizontal space is often at a premium. If we are discussing variable names in code, however, you should know that there is a very strong convention in Haskell to use camel case.

michael-liesenfelt:
> If we decide that we prefer to make this value hard to change, hard-coding it is a nice and simple solution. I'll just reiterate that nearly all of the [protocol parameters](https://github.com/input-output-hk/cardano-ledger/blob/7e2f674d2a2d14752d4c2d5abf60b26ae015b9e2/eras/babbage/impl/src/Cardano/Ledger/Babbage/PParams.hs#L107-L152) are capable of having a dramatic change to the network, and that what we really need is a holistic solution for governing _all_ of them.

Agreed. For this parameter, hardforkThreshold = 0.75 with a hard-coded range bound of ( 0.51 , 0.95 ) would be wise. It would be very wise to hard-code values of a safe reasonable minimum and a safe reasonable maximum for _all_ of the parameters. Governance could be used to adjust the parameter within the safe range and only a hardfork could change the bounds per parameter.

JaredCorduan:
> Agreed. For this parameter, hardforkThreshold = 0.75 with a hard-coded range bound of ( 0.51 , 0.95 ) would be wise.

I think this is an excellent suggestion! Make it a protocol parameter that can be changed, but restrict the range to safe values.

CIP-hardforkSafeguard/README.md
JaredCorduan: I've added a sentence to the end of this paragraph, let me know if it's clear.

CIP-hardforkSafeguard/README.md
```diff
@@ -0,0 +1,206 @@
...
+which triggers a hardfork.
+
+This CIP proposes automating this process,
+and making the protocol version in the header semantically meaningful.
+The ledger state will determine the stake
+(represented as the proportion of the active stake) of all the block producers
+whose last block contained the next major protocol version.
+Moreover, a new protocol parameter `hardforkThreshold` will be used to reject any
+protocol parameter update that proposes to change the major protocol version
+but does not have enough backing stake.
```
dirkhh-cf, line 41: Can you clarify was "automating the process" means in this context?
I can see in your description how this could be used to prevent a hardfork if there is insufficient backing - but what isn't explained is how such a hardfork would be initiated. Who could make such a change proposal? Is that also automated? Is it anyone who can produce blocks (so any SPO for example)? Or just some specific entity / entities?

CIP-hardforkSafeguard/README.md
dirkhh-cf: I don't think you actually addressed the real question - and this is related to my comment above.
Who gets to make such proposals. Let's say entity A wants to make a change to parameter X and therefore submits a proposal to change the protocol version to (m+1, 0) - and entity B wants to leave parameter X unchanged, but wants to change parameter Y instead. The also submit a proposal. Which protocol version would be assigned to that? How would a determination be made, which of the changes to proceed with? How would the SPOs indicate which of the proposals they endorse?

CIP-hardforkSafeguard/README.md
```diff
@@ -0,0 +1,206 @@
...
+    Note over s7: Consensus to<br>determine if<br>hardfork will occur
+    Note over sA: Non-rejected<br>updates are applied<br>to ledger state
+```
+
+## Rationale
+
+The safeguard presented in this CIP aligns very closely with the manual check
+currently performed
+today before any hardfork.
+Moreover, we have strived to make the minimal changes needed to automate
```
dirkhh-cf, line 187: This is actually only part of the criteria currently being used. SPO block creation ratio, defi TVL criterion, exchange adoption criterion. I'm not saying that the other two should be encoded here - but it would be reasonable to mention them in the Rationale

dirkhh-cf (commented):
I really like the direction this is going, but I have some questions... 😁

CIP-hardforkSafeguard/README.md
```diff
@@ -0,0 +1,206 @@
...
+
+A hardfork is required for these changes.
+A new ledger era is needed, containing the changes described.
+The consensus layer will require minimal changes, namely
+support for the new ledger era and adopting the new logic for determining if a
+hardfork is immanent.
+
+## Copyright
+
+This CIP is licensed under Apache-2.0
```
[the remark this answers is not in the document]
JaredCorduan: Thank you for catching this! I will ask about what I am supposed to use, I admit I just copied those from an [existing](https://github.com/cardano-foundation/CIPs/tree/master/CIP-0028) CIP created by an IOG employee.

CIP-hardforkSafeguard/README.md
JaredCorduan: > Let's say entity A wants to make a change to parameter X and therefore submits a proposal to change the protocol version to (m+1, 0) - and entity B wants to leave parameter X unchanged, but wants to change parameter Y instead. The also submit a proposal. Which protocol version would be assigned to that?

I guess there are two issues here:
* how the protocol parameter update system works
* how stake pool operators endorse a hard fork

The answer to the first question is:

Whether or not the governance system will move to change the protocol version to  `(m+1, 0)` depends not just on A and B, but also on the other five governance entities (the quorum is 5 of 7 on mainnet). The current system is very basic: at least five of the keys must agree on the entire set of changes. So if entities A - G all want to change Y to 42, but only four of the entities want to change the protocol version to `(m+1, 0)`, nothing is changed at all, not even Y.

The answer to the second question is:

Regardless of what the governance body is doing, if you are a stake pool operator and you are aware of a software update that prepares a hard fork, let's say introducing protocol version to  `(m+1, 0)`, you can:
* signal your willingness to enact the harfork by placing `m+1` in your block headers (the new software will actually do this for you)
* signal your unwillingness to enact the harfork by placing `m` in your block headers (the old software will actually do this for you)

If not enough stake is backed by the bolck producers posting `m+1`, no update proposal can occur which changes the major version to `m+1`, even if quorum is met and even if other protocol parameters we also slated to change.

---

I can try to summarize this in the CIP, since clearly it's still not clear.

CIP-hardforkSafeguard/README.md
JaredCorduan: Let me see if I can explain it well enough here, and if it makes sense I'll add it to the CIP.

The current manual process is: around the time of a hard fork, humans use a tool like db-sync to see which stake pools are posting blocks with the current major protocol version in the block header, and which ones have the the current major protocol version plus one. If the holders of the goverance keys decide "yea, looks good", and they are otherwise ok with the hard fork (I do not myself know all that goes into the operational side if this, and am not proposing automating anything besides this single check), then they submit update proposals. If 5 of 7 agree on the exact proposals, the change occurs on the epoch boundary.

CIP-hardforkSafeguard/README.md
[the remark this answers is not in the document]
dirkhh-cf: speaking with my "I am not a lawyer but I have spent way too much time working on licenses" hat on... I'd suggest going with the license in line 8 (CC-BY-4.0) as that is a great license for documents. Apache v2 is a very good choice for code (as it makes it easy to reuse and clarifies patent concerns), but doesn't really make a lot of sense if you apply it to documents.

CIP-hardforkSafeguard/README.md
JaredCorduan: Maybe I can make it more clear that this CIP is only aiming to automate one very specific check? I myself do not know the whole process.

CIP-hardforkSafeguard/README.md
dirkhh-cf: Yup, that makes total sense... I guess what I asked for was some clarification on what exactly is meant by "automating the process" - so in the brave new world, how would this work. Is it still one of the seven proposing, at least five of seven agreeing, but THEN (after that) there is ALSO the technical requirement proposed here?
Or does this change the first part of the process as well, i.e. can others propose that change? Who needs to approve before this technical requirement comes into play.

I am thinking (but could be totally wrong) that right now you are only focused on the negative case, i.e. this CIP _only_ deals with creating a hurdle that prevents a change and doesn't try to change any other part of the process. But I think this could be somewhat clearer in this paragraph of the explanation.

CIP-hardforkSafeguard/README.md
dirkhh-cf: Thanks for that clarification. Yes, that would be very appreciated. I clearly read way too much into what you were intending to change.

CIP-hardforkSafeguard/README.md
JaredCorduan: > I am thinking (but could be totally wrong) that right now you are only focused on the negative case, i.e. this CIP only deals with creating a hurdle that prevents a change and doesn't try to change any other part of the process.

That is exactly right! This CIP does not intend to change the off-chain process in any way. It's just a final blocker to stop a hard fork that is not properly endorsed by the SPOs.

JaredCorduan: I just realized I lied :laughing: 

> This CIP does not intend to change the off-chain process in any way.

This CIP intends to lift one very specific off-chain process to the protocol itself, namely looking through block headers to see who is signaling their readiness.

CIP-hardforkSafeguard/README.md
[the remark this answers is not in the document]
JaredCorduan: done!

CIP-hardforkSafeguard/README.md
JaredCorduan: I've tried to make this clear now.

CIP-hardforkSafeguard/README.md
JaredCorduan: I've tried again to make this point more clear. Let me know if it is still murky.

WarriorField: What happens if 20% of SPO's upgrade, then a bug is found.  50% of the original upgraded SPO then install the new fixed major release and enough other SPO's do as well so the hard fork is successful.  But 10% of the SPO's have the buggy prefix running but it has the same version number?

JaredCorduan: That's an excellent point @WarriorField , and I'm embarrassed I did not think to address it in this CIP. It's far from just a theoretical concern, this came up quite recently, and this is what we did:

https://github.com/input-output-hk/cardano-node/blame/8832f86728ef6a425452b44f2f269acde149448c/cardano-node/src/Cardano/Node/Protocol/Cardano.hs#L201-L206

We fiddled with the minor version. And of course this only worked since the process is still manual. This CIP should address this, thank you!

JaredCorduan:
I am going to close this PR, since it is replaced by [CIP-1694](https://github.com/cardano-foundation/CIPs/pull/380). yea?

rphair:
that makes sense @JaredCorduan but can you please add this URL `https://github.com/cardano-foundation/CIPs/pull/318` to the `Discussions:` header of your draft in https://github.com/cardano-foundation/CIPs/pull/380? (unless you'd like to declare it irrelevant)

JaredCorduan:
> that makes sense @JaredCorduan but can you please add this URL `https://github.com/cardano-foundation/CIPs/pull/318` to the `Discussions:` header of your draft in #380? (unless you'd like to declare it irrelevant)

I think I'd prefer to declare it irrelevant, since all the discussion here was about the nitty gritty details of the mechanism here that we are abandoning. but I'll happily defer to y'all's guidance!

michael-liesenfelt:
I agree.
I don't believe there are any good ideas in CIP-0047 which haven't been incorporated into the scope of CIP-1694.

PR #58, opened 2021-01-05
Catalyst Registration transaction metadata format

SebastienGllmt:
This CIP just described the format that was used during the round 2 registration for Catalyst. No guarantee IOHK will continue to use this system going forward, but since many explorers, wallets & tools have hardcoded support for this specific registrartion format, I felt it was appropriate to document this properly as a CIP

PR #321, opened 2022-08-23
CIP-0069? | Plutus Script Type Uniformization

zygomeb:
The aim of the CIP is to allow more secure designs, streamline the developer experience and increase the design space of Cardano validators. 

All three purposes are achieved by removing the `Datum` out of a spending script's signature, to be the exact same as every other type of script. Datum is made available in `ScriptPurpose` instead. 

This change lets us use one script as both a spending and a minting validator, meaning that we resolve a well known protocol design problem of 'mutually dependent' validators, where a minting validator needs as parameter a spending validator, and the spending validator needs as parameter the minting one. 
With this CIP, they can be merged into a single script, enabling us to enable the two-way (or more) dependency where necessary. 

---

[see rendered Markdown](https://github.com/zygomeb/CIPs/blob/master/CIP-0069/README.md)

### DIFF
--- a/CIP-0069/README.md
+++ b/CIP-0069/README.md
@@ -2,7 +2,7 @@
 CIP: 69
 Title: Script Signature Unification 
 Authors: Maksymilian 'zygomeb' Brodowicz <zygomeb@gmail.com>
-Status: Draft
+Status: Proposed
 Type: Standards Track
 Created: 2022-08-23
 License: CC-BY-4.0
--- a/CIP-0069/README.md
+++ b/CIP-0069/README.md
@@ -1,5 +1,5 @@
 ---
-CIP: 0069
+CIP: 69
 Title: Script Signature Unification 
 Authors: Maksymilian 'zygomeb' Brodowicz <zygomeb@gmail.com>
 Status: Draft
--- a/CIP-0069/README.md
+++ b/CIP-0069/README.md
@@ -28,49 +28,21 @@ Use Case 2 (taken from Optim's Liquidity Bonds): Unique NFT is minted to give a 
 
 Use Case 3 (taken from Minswap's Dex V1): NFT is minted for the same reason as above. It allows a minting policy to later mint LP tokens with that unique id token name.
 
-We see a similar pattern repeating over and over again, as I bid you, the developer reading this to provide witness. By allowing the multi-purpose policies (spending and any other) we increase the security of Cardano by giving us more confidence and allowing to design protocols that have their design driven by cardano's features, not limited by cardano's language. 
+We see a similar pattern repeating over and over again as witnessed by dapp developers and auditors alike. By allowing the multi-purpose policies (spending and any other) we increase the security of Cardano by giving us more confidence and allowing to design protocols that have their architecture driven by Cardano's features, not limited by Cardano's language. 
 
 ## Specification
 
 ### Removing the datum argument
 
-All the sctipt purposes have a form of `Redeemer -> ScriptContext -> ()` except the `Spending` one.
-It has form of `Datum -> Redeemer -> ScriptContext -> ()`. This is enforced by the cardano node. 
+All the sctipt purposes have a form of `Redeemer -> ScriptContext -> exists a. a` except the `Spending` one.
+It has form of `Datum -> Redeemer -> ScriptContext -> exists a. a`. This is enforced by the Cardano node. 
 
-We make it conform to the general pattern by having it also have a 'signature' of `Redeemer -> ScriptContext -> ()`
-
-### (Optional) Providing the Datum in ScriptPurpose
-
-The datatype in question defined as
-
-```
-data ScriptPurpose
-  = Minting CurrencySymbol
-  | Spending TxOutRef
-  | Rewarding StakingCredential
-  | Certifying DCert
-```
-
-to be redefined to
-
-```
-data ScriptPurpose
-  = Minting CurrencySymbol
-  | Spending TxOutRef Datum
-  | Rewarding StakingCredential
-  | Certifying DCert
-```
-
-which will allow a script to easily match on its purpose and pick the datum out as it would do so normally.
-
-If this change is to not be added, the Datum can be picked out of `ScriptConext`. Adding unnecessary boilerplate and clock cycles. 
-
-Cardano node would naturally verify that the correct `Datum` is given, as it does currently. 
+We make it conform to the general pattern by having it also have a type of `Redeemer -> ScriptContext -> ()`
 
 ## Rationale
 
 Unifying of the script signature is a very elegant solution to the problem, streamlining the experience of developing on cardano.
-Given that accessing the datum is almost always made by a spending script, it makes sense to intdoduce that argument back to the `ScriptPurpose` that now plays a more important role.
+Given that accessing the datum is almost always made by a spending script, it makes sense to introduce that argument back to the `ScriptPurpose` that now plays a more important role.
 It begs the question if it should be added as an argument to all validators, to further emphasize that fact. 
 
 ## Backwards compatibility
--- a/CIP-0069/README.md
+++ b/CIP-0069/README.md
@@ -16,11 +16,9 @@ For a while now every builder, myself included have struggled with the mutual de
 
 The exact change would be to have every validator take as argument the redeemer and then the script context. Datums, only relevant to locking validators would be able to be provided by either looking them up in the ScriptContext or by extending the `Spending` constructor of `TxInfo` to carry `(TxOutRef, Datum)`. 
 
-If this CIP were to be accepted and implemented, we'd be able to resolve a mutual dependency by merging two scripts into one, that can act as both say minting and spending depending on the purpose given. 
-
 ## Motivation / History
 
-As it stands the scripts being made on cardano often suffer this problem, and the tokens usually are made to be able to be minted at any time. This leads to further checks being made on the frontend and further fragilitiy of the systems we create.
+As it stands the scripts being made on cardano often suffer this problem, and the tokens usually are made to be able to be minted at any time. This leads to further checks being made on the frontend and further fragilitiy of the systems we create. When a mutual dependency arises we are forced to choose which script gets to statically know what's the hash of the other, and which has to be provided 'during runtime'.
 
 Use Case 1: Minting validator checks datum given to spending validator. The spending validator requires the token be present as witness of the datum's correctness.
 
@@ -28,16 +26,24 @@ Use Case 2 (taken from Optim's Liquidity Bonds): Unique NFT is minted to give a 
 
 Use Case 3 (taken from Minswap's Dex V1): NFT is minted for the same reason as above. It allows a minting policy to later mint LP tokens with that unique id token name.
 
-We see a similar pattern repeating over and over again as witnessed by dapp developers and auditors alike. By allowing the multi-purpose policies (spending and any other) we increase the security of Cardano by giving us more confidence and allowing to design protocols that have their architecture driven by Cardano's features, not limited by Cardano's language. 
+We see a similar pattern repeating over and over again as witnessed by dapp developers and auditors alike. By allowing the multi-purpose policies (spending and any other) we increase the security of Cardano by giving us more confidence and allowing to design protocols that have their architecture driven by Cardano's features, not limited by Cardano's language.
+
+This primarily manifests in the ability to use a single validator for both minting and spending but the proposed solution makes it possible to use one validator for any and all purposes at once.
 
 ## Specification
 
 ### Removing the datum argument
 
-All the sctipt purposes have a form of `Redeemer -> ScriptContext -> exists a. a` except the `Spending` one.
-It has form of `Datum -> Redeemer -> ScriptContext -> exists a. a`. This is enforced by the Cardano node. 
+All the script purposes have a form of `Redeemer -> ScriptContext -> exists a. a` except the `Spending` one.
+It has the following form: `Datum -> Redeemer -> ScriptContext -> exists a. a`. This is enforced by the Cardano ledger. 
 
-We make it conform to the general pattern by having it also have a type of `Redeemer -> ScriptContext -> ()`
+We modify the general signature to be `ScriptArgs -> ScriptContext -> exists a. a`.
+
+```hs
+data ScriptArgs =
+    RedeemerOnly Redeemer
+  | RedeemerAndDatum Redeemer Datum
+```
 
 ## Rationale
 
--- a/CIP-0069/README.md
+++ b/CIP-0069/README.md
@@ -1,30 +1,34 @@
 ---
 CIP: 69
-Title: Script Signature Unification 
-Authors: Maksymilian 'zygomeb' Brodowicz <zygomeb@gmail.com>
+Title: Script Signature Unification
+Category: Plutus
+Authors:
+  - Maksymilian 'zygomeb' Brodowicz <zygomeb@gmail.com>
+Implementors: N/A
 Status: Proposed
-Type: Standards Track
+Discussions:
+  - https://github.com/cardano-foundation/CIPs/pull/321
 Created: 2022-08-23
 License: CC-BY-4.0
 ---
 
-## Simple Summary / Abstract
+## Abstract
 
 This CIP unifies the arguments given to all types of Plutus scripts currently available (spending, certifying, rewarding, minting) by removing the argument of a datum.
 
-For a while now every builder, myself included have struggled with the mutual dependency issue (two validators that need to know each other's hash) when designing dapps and it is widely considered a substantial barrier to safe protocols and one that limits our design space considerably. 
+For a while now every builder, myself included have struggled with the mutual dependency issue (two validators that need to know each other's hash) when designing dapps and it is widely considered a substantial barrier to safe protocols and one that limits our design space considerably.
 
-The exact change would be to have every validator take as argument the redeemer and then the script context. Datums, only relevant to locking validators would be able to be provided by either looking them up in the ScriptContext or by extending the `Spending` constructor of `TxInfo` to carry `(TxOutRef, Datum)`. 
+The exact change would be to have every validator take as argument the redeemer and then the script context. Datums, only relevant to locking validators would be able to be provided by either looking them up in the ScriptContext or by extending the `Spending` constructor of `TxInfo` to carry `(TxOutRef, Datum)`.
 
-## Motivation / History
+## Motivation: why is this CIP necessary?
 
 As it stands the scripts being made on cardano often suffer this problem, and the tokens usually are made to be able to be minted at any time. This leads to further checks being made on the frontend and further fragilitiy of the systems we create. When a mutual dependency arises we are forced to choose which script gets to statically know what's the hash of the other, and which has to be provided 'during runtime'.
 
-Use Case 1: Minting validator checks datum given to spending validator. The spending validator requires the token be present as witness of the datum's correctness.
+- Use Case 1: Minting validator checks datum given to spending validator. The spending validator requires the token be present as witness of the datum's correctness.
 
-Use Case 2 (taken from Optim's Liquidity Bonds): Unique NFT is minted to give a unique identifier to a loan, that then gets reused by Bond Tokens. The spending validators require that NFT be present. 
+- Use Case 2 (taken from Optim's Liquidity Bonds): Unique NFT is minted to give a unique identifier to a loan, that then gets reused by Bond Tokens. The spending validators require that NFT be present.
 
-Use Case 3 (taken from Minswap's Dex V1): NFT is minted for the same reason as above. It allows a minting policy to later mint LP tokens with that unique id token name.
+- Use Case 3 (taken from Minswap's Dex V1): NFT is minted for the same reason as above. It allows a minting policy to later mint LP tokens with that unique id token name.
 
 We see a similar pattern repeating over and over again as witnessed by dapp developers and auditors alike. By allowing the multi-purpose policies (spending and any other) we increase the security of Cardano by giving us more confidence and allowing to design protocols that have their architecture driven by Cardano's features, not limited by Cardano's language.
 
@@ -34,10 +38,10 @@ This primarily manifests in the ability to use a single validator for both minti
 
 ### Removing the datum argument
 
-All the script purposes have a form of `Redeemer -> ScriptContext -> exists a. a` except the `Spending` one.
-It has the following form: `Datum -> Redeemer -> ScriptContext -> exists a. a`. This is enforced by the Cardano ledger. 
+All the script purposes have a form of `Redeemer -> ScriptContext -> a` except the `Spending` one.
+It has the following form: `Datum -> Redeemer -> ScriptContext -> a`. This is enforced by the Cardano ledger.
 
-We modify the general signature to be `ScriptArgs -> ScriptContext -> exists a. a`.
+We modify the general signature to be `ScriptArgs -> ScriptContext -> a`.
 
 ```hs
 data ScriptArgs =
@@ -45,11 +49,11 @@ data ScriptArgs =
   | RedeemerAndDatum Redeemer Datum
 ```
 
-## Rationale
+## Rationale: how does this CIP achieve its goals?
 
 Unifying of the script signature is a very elegant solution to the problem, streamlining the experience of developing on cardano.
 Given that accessing the datum is almost always made by a spending script, it makes sense to introduce that argument back to the `ScriptPurpose` that now plays a more important role.
-It begs the question if it should be added as an argument to all validators, to further emphasize that fact. 
+It begs the question if it should be added as an argument to all validators, to further emphasize that fact.
 
 ## Backwards compatibility
 
@@ -58,7 +62,13 @@ Node code must be modified.
 
 ## Path to Active
 
-The appropriate changes need to be made to the node code and a new language major version be introduced. 
+### Acceptance Criteria
+
+- [ ] The change has been implemented in the Plutus codebase, integrated in the ledger and released through a hard-fork.
+
+### Implementation Plan
+
+None.
 
 ## Copyright
 
--- a/CIP-0069/README.md
+++ b/CIP-0069/README.md
@@ -1,0 +1,87 @@
+---
+CIP: 0069
+Title: Script Signature Unification 
+Authors: Maksymilian 'zygomeb' Brodowicz <zygomeb@gmail.com>
+Status: Draft
+Type: Standards Track
+Created: 2022-08-23
+License: CC-BY-4.0
+---
+
+## Simple Summary / Abstract
+
+This CIP unifies the arguments given to all types of Plutus scripts currently available (spending, certifying, rewarding, minting) by removing the argument of a datum.
+
+For a while now every builder, myself included have struggled with the mutual dependency issue (two validators that need to know each other's hash) when designing dapps and it is widely considered a substantial barrier to safe protocols and one that limits our design space considerably. 
+
+The exact change would be to have every validator take as argument the redeemer and then the script context. Datums, only relevant to locking validators would be able to be provided by either looking them up in the ScriptContext or by extending the `Spending` constructor of `TxInfo` to carry `(TxOutRef, Datum)`. 
+
+If this CIP were to be accepted and implemented, we'd be able to resolve a mutual dependency by merging two scripts into one, that can act as both say minting and spending depending on the purpose given. 
+
+## Motivation / History
+
+As it stands the scripts being made on cardano often suffer this problem, and the tokens usually are made to be able to be minted at any time. This leads to further checks being made on the frontend and further fragilitiy of the systems we create.
+
+Use Case 1: Minting validator checks datum given to spending validator. The spending validator requires the token be present as witness of the datum's correctness.
+
+Use Case 2 (taken from Optim's Liquidity Bonds): Unique NFT is minted to give a unique identifier to a loan, that then gets reused by Bond Tokens. The spending validators require that NFT be present. 
+
+Use Case 3 (taken from Minswap's Dex V1): NFT is minted for the same reason as above. It allows a minting policy to later mint LP tokens with that unique id token name.
+
+We see a similar pattern repeating over and over again, as I bid you, the developer reading this to provide witness. By allowing the multi-purpose policies (spending and any other) we increase the security of Cardano by giving us more confidence and allowing to design protocols that have their design driven by cardano's features, not limited by cardano's language. 
+
+## Specification
+
+### Removing the datum argument
+
+All the sctipt purposes have a form of `Redeemer -> ScriptContext -> ()` except the `Spending` one.
+It has form of `Datum -> Redeemer -> ScriptContext -> ()`. This is enforced by the cardano node. 
+
+We make it conform to the general pattern by having it also have a 'signature' of `Redeemer -> ScriptContext -> ()`
+
+### (Optional) Providing the Datum in ScriptPurpose
+
+The datatype in question defined as
+
+```
+data ScriptPurpose
+  = Minting CurrencySymbol
+  | Spending TxOutRef
+  | Rewarding StakingCredential
+  | Certifying DCert
+```
+
+to be redefined to
+
+```
+data ScriptPurpose
+  = Minting CurrencySymbol
+  | Spending TxOutRef Datum
+  | Rewarding StakingCredential
+  | Certifying DCert
+```
+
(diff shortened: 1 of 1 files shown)


CIP-0069/README.md
```diff
@@ -0,0 +1,87 @@
...
+  | Spending TxOutRef Datum
+  | Rewarding StakingCredential
+  | Certifying DCert
+```
+
+which will allow a script to easily match on its purpose and pick the datum out as it would do so normally.
+
+If this change is to not be added, the Datum can be picked out of `ScriptConext`. Adding unnecessary boilerplate and clock cycles. 
+
+Cardano node would naturally verify that the correct `Datum` is given, as it does currently.
```
L-as, line 68: Perhaps mentioning the ledger rules is more correct? Albeit I'm not sure whether this sentence is relevant. Rather than verifying that the correct `Datum` is given, this is how the `ScriptContext` is generated. The `ScriptContext` isn't part of the transaction.

CIP-0069/README.md
```diff
@@ -0,0 +1,87 @@
...
+which will allow a script to easily match on its purpose and pick the datum out as it would do so normally.
+
+If this change is to not be added, the Datum can be picked out of `ScriptConext`. Adding unnecessary boilerplate and clock cycles. 
+
+Cardano node would naturally verify that the correct `Datum` is given, as it does currently. 
+
+## Rationale
+
+Unifying of the script signature is a very elegant solution to the problem, streamlining the experience of developing on cardano.
+Given that accessing the datum is almost always made by a spending script, it makes sense to intdoduce that argument back to the `ScriptPurpose` that now plays a more important role.
```
L-as, line 73: intdoduce

CIP-0069/README.md
```diff
@@ -0,0 +1,87 @@
...
+It begs the question if it should be added as an argument to all validators, to further emphasize that fact. 
+
+## Backwards compatibility
+
+This change is not backwards compatible; it must be introduced in a new Plutus language version.
+Node code must be modified.
```
L-as, line 79: Ledger?

L-as (commented):
This is great

CIP-0069/README.md
```diff
@@ -0,0 +1,87 @@
...
+
+As it stands the scripts being made on cardano often suffer this problem, and the tokens usually are made to be able to be minted at any time. This leads to further checks being made on the frontend and further fragilitiy of the systems we create.
+
+Use Case 1: Minting validator checks datum given to spending validator. The spending validator requires the token be present as witness of the datum's correctness.
+
+Use Case 2 (taken from Optim's Liquidity Bonds): Unique NFT is minted to give a unique identifier to a loan, that then gets reused by Bond Tokens. The spending validators require that NFT be present. 
+
+Use Case 3 (taken from Minswap's Dex V1): NFT is minted for the same reason as above. It allows a minting policy to later mint LP tokens with that unique id token name.
+
+We see a similar pattern repeating over and over again, as I bid you, the developer reading this to provide witness. By allowing the multi-purpose policies (spending and any other) we increase the security of Cardano by giving us more confidence and allowing to design protocols that have their design driven by cardano's features, not limited by cardano's language.
```
SebastienGllmt, line 31: Keep in mind CIPs not only serve to request changes to the protocol, but also help Cardano developers of the future understand features that were added to the protocol. Therefore, I would recommend against language directly targeted at reviewers like this

michaelpj:
I don't hate this, it's the most plausible variant of this proposal that I've seen. I think the fundamental idea of using a sum-type to carry the arguments to the script (whatever they are) is sound. However, I don't agree with putting this information in `ScriptPurpose`. `ScriptPurpose` serves to _identify_ a script execution, that's why it's used (for example) in the redeemers map as a key. The datum is not needed for that, so adding it is redundant and will lead to including expensive extra data in various places.

So I think we would need a new type, say `ScriptArgs`. Then we could have something like:
```
data ScriptArgs = 
  RedeemerOnly Redeemer
  RedeemerAndDatum Redeemer Datum

type ScriptType = ScriptArgs -> ScriptContext -> ()
```
or even (un-tupling `ScriptContext`)
```
type ScriptType = ScriptPurpose -> ScriptArgs -> TxInfo -> ()
``

That leaves an ugly unchecked dependence between `ScriptPurpose` and `ScriptArgs`, since only certain combinations of those are possible. But I think that's still nicer than pushing the `ScriptArgs` information into `ScriptPurpose`.

Having a separate `ScriptArgs` type also opens up the possibility of having scripts that take _other_ kinds of argument in future. For example, we could easily support scripts that don't even take a redeemer.

-----

Regarding the motivation for this proposal, I'm not 100% convinced that the desire to use the very same script in multiple contexts is compelling enough on its own. But I think something like what I sketched above has other advantages that make it compelling, in particular making us more flexible in how we pass arguments in future.

**Comment by WhatisRT** [comment]
I'm not sure this would actually require a change to the ledger or Plutus. If you think of it from the perspective of Plutus core I'm sure you would be able to write a script that works whether there's a datum present or not, and I think you should also be able to properly look at all the arguments (but maybe my understanding of how Plutus core is executed is wrong here). If that's the case, then this could be done with any changes in the node (and thus no HF).

**Comment by michaelpj** [comment]
> I'm not sure this would actually require a change to the ledger or Plutus. If you think of it from the perspective of Plutus core I'm sure you would be able to write a script that works whether there's a datum present or not

It's pretty awkward, since the datum comes first. You don't get the thing that tells you what kind of script you are until last.

**Comment by mchakravarty** [comment]
@michaelpj I totally agree with that you are writing and also think that having something like `ScriptArgs` is the better approach.

**Comment by polinavino** [comment]
I agree in general with `ScriptArgs` being a good way to go about having a type that allows for passing different sets of arguments, but a nice way of doing this would probably require some kind of HF-type changes. So, if this is happening, better to plan to wrap it up with an upcoming HF - I figure?

File: CIP-0069/README.md [hunk 5]
```diff
@@ -0,0 +1,87 @@
...
+
+## Simple Summary / Abstract
+
+This CIP unifies the arguments given to all types of Plutus scripts currently available (spending, certifying, rewarding, minting) by removing the argument of a datum.
+
+For a while now every builder, myself included have struggled with the mutual dependency issue (two validators that need to know each other's hash) when designing dapps and it is widely considered a substantial barrier to safe protocols and one that limits our design space considerably. 
+
+The exact change would be to have every validator take as argument the redeemer and then the script context. Datums, only relevant to locking validators would be able to be provided by either looking them up in the ScriptContext or by extending the `Spending` constructor of `TxInfo` to carry `(TxOutRef, Datum)`. 
+
+If this CIP were to be accepted and implemented, we'd be able to resolve a mutual dependency by merging two scripts into one, that can act as both say minting and spending depending on the purpose given.
```
KtorZ, line 19: The last sentence is unnecessary, or at least, needs reformulation. Write if from the perspective of the CIP being adopted. If it is rejected, it will be marked as such and if it is accepted, it should be read from that point of view.

CIP-0069/README.md
```diff
@@ -0,0 +1,87 @@
...
+Use Case 3 (taken from Minswap's Dex V1): NFT is minted for the same reason as above. It allows a minting policy to later mint LP tokens with that unique id token name.
+
+We see a similar pattern repeating over and over again, as I bid you, the developer reading this to provide witness. By allowing the multi-purpose policies (spending and any other) we increase the security of Cardano by giving us more confidence and allowing to design protocols that have their design driven by cardano's features, not limited by cardano's language. 
+
+## Specification
+
+### Removing the datum argument
+
+All the sctipt purposes have a form of `Redeemer -> ScriptContext -> ()` except the `Spending` one.
+It has form of `Datum -> Redeemer -> ScriptContext -> ()`. This is enforced by the cardano node.
```
KtorZ, line 38:
```suggestion
It has the following form: `Datum -> Redeemer -> ScriptContext -> ()`. This is enforced by the Cardano ledger. 
```

CIP-0069/README.md
```diff
@@ -0,0 +1,87 @@
...
+
+Use Case 3 (taken from Minswap's Dex V1): NFT is minted for the same reason as above. It allows a minting policy to later mint LP tokens with that unique id token name.
+
+We see a similar pattern repeating over and over again, as I bid you, the developer reading this to provide witness. By allowing the multi-purpose policies (spending and any other) we increase the security of Cardano by giving us more confidence and allowing to design protocols that have their design driven by cardano's features, not limited by cardano's language. 
+
+## Specification
+
+### Removing the datum argument
+
+All the sctipt purposes have a form of `Redeemer -> ScriptContext -> ()` except the `Spending` one.
```
KtorZ, line 37:
```suggestion
All the script purposes have a form of `Redeemer -> ScriptContext -> ()` except the `Spending` one.
```

KtorZ:
I think I failed to understand the motivation behind this proposal. The abstract mentions the problem of mutually dependent scripts and then focuses on the datum (?). However, scripts are only parameterized by datums, which only are populated at runtime. Hence, a datum itself does not change a script (and its corresponding hash). In case a policy needs to specifically access a datum, it can embeds it in its code, and a spending script can also refer to the policy without issue (for the datum is external to the spending script). 

The only case where this doesn't play out well is when the datum isn't known upfront. But I don't see how this proposal addresses this shortcoming as well. Am I missing something?

KtorZ:
- After discussion during the CIP editors meeting, @L-as kindly clarified the motivation behind the CIP -- which I thereby suggests to make more apparent: the ability to use a script as both a minting policy and a spending condition; effectively pattern-matching on the script purpose to distinguish it. 
- I am also suggesting a title change: 'Plutus Script Type Uniformization' since the word 'Signature' is a bit overloaded in our space. 
- Overall, I'd suggest to also rewrite some parts of the proposal from the point of view of it being adopted; and not using conditional tone or addressing reviewers. Think about developers reading the proposals years from now.

L-as:
@michaelpj
> So I think we would need a new type, say `ScriptArgs`. Then we could have something like:

I think it's not great to represent the same choice multiple times.
Rather than `type ScriptType = ScriptPurpose -> ScriptArgs -> TxInfo -> ()`, it seems more sensible to me to take what's inside `ScriptPurpose` right now and put it into the hypothetical `ScriptArgs`, essentially, renaming `ScriptPurpose` to `ScriptArgs` and including the `Datum` in the `Spending` constructor. This would be cleaner IMO.

michaelpj:
@L-as I discussed that earlier: 

> However, I don't agree with putting this information in ScriptPurpose. ScriptPurpose serves to identify a script execution, that's why it's used (for example) in the redeemers map as a key. The datum is not needed for that, so adding it is redundant and will lead to including expensive extra data in various places.

We need `ScriptPurpose` to be just `ScriptPurpose`. I guess we could have a `ScriptArgs` that partially duplicates the information in `ScriptPurpose`, but that feels quite ugly.

L-as:
@michaelpj Thanks, that makes sense. If we think in terms of efficiency, just putting `ScriptPurpose` first is also a viable solution, but means we can't type a unified minting policy and validator in PLC unless we introduce pi types or equality types/GADTs.

zygomeb:
Thank you all for your constructive and kind responses. I'll reword the proposal to make the tone match the expectation, and provide more context if it is still needed. 

@michaelpj would it make more sense to instead of introducing the unchecked dependency with these types
```
data ScriptArgs = 
  RedeemerOnly Redeemer
  RedeemerAndDatum Redeemer Datum

type ScriptType = ScriptArgs -> ScriptContext -> ()
```
and instead go and push further for unified types and make it 
```
data ScriptArgs = 
  RedeemerAndDatum Redeemer Datum

type ScriptType = ScriptArgs -> ScriptContext -> ()
```

What this does is feed scripts that did not require prior arguments of Datum, a datum that is equal to `()`. This could be a better API and if this were to make it in say PlutusV3 we'd be able to discuss if it were possible to make use of this parameter for purposes more general than just a datum. In spirit making it a unification in signature proper and while there is a one way dependency (i.e. if datum is not a unit then it cannot be Minting, Certifying or Withdrawing) it is much better, I believe, than having as much duplication as we do with the original datatype. 

An alternative route here to unified API with less dependencies which I describe in my [(optional) paragraph](https://github.com/zygomeb/CIPs/blob/master/CIP-0069/README.md#optional-providing-the-datum-in-scriptpurpose) is to move the datum into the ScriptPurpose which may be even more elegant. It would make the proposed types 

```
data ScriptArgs = 
  RedeemerOnly Redeemer
  
type ScriptType = ScriptArgs -> ScriptContext -> ()
```
As the `Datum` would be provided in script purpose. No dependencies at all.

michaelpj:
> What this does is feed scripts that did not require prior arguments of Datum, a datum that is equal to ()

This is basically isomorphic to what I suggested except that you can't distinguish the case where you're not supposed to have a datum, and adds a special case for the datum that you give to things that don't have one. I don't find this simpler than my proposal, which just represents the two cases with two cases. I don't think it introduces any more duplication, just a sum type.

> An alternative route here to unified API with less dependencies which I describe in my [(optional) paragraph](https://github.com/zygomeb/CIPs/blob/master/CIP-0069/README.md#optional-providing-the-datum-in-scriptpurpose) is to move the datum into the ScriptPurpose which may be even more elegant.

As I said earlier, I don't like this: https://github.com/cardano-foundation/CIPs/pull/321#issuecomment-1259429865

zygomeb:
I've changed the proposal to be aligned with the suggestions of @michaelpj and incorporated the typo and style fixes as suggested

CIP-0069/README.md
```diff
@@ -0,0 +1,65 @@
+---
+CIP: 0069
```
KtorZ, line 2:
```suggestion
CIP: 69
```

CIP-0069/README.md
```diff
@@ -0,0 +1,65 @@
...
+Title: Script Signature Unification 
+Authors: Maksymilian 'zygomeb' Brodowicz <zygomeb@gmail.com>
+Status: Draft
```
KtorZ, line 5:
```suggestion
Status: Proposed
```

CIP-0069/README.md
```diff
@@ -0,0 +1,65 @@
...
+It begs the question if it should be added as an argument to all validators, to further emphasize that fact. 
+
+## Backwards compatibility
+
+This change is not backwards compatible; it must be introduced in a new Plutus language version.
+Node code must be modified.
+
+## Path to Active
+
+The appropriate changes need to be made to the node code and a new language major version be introduced.
```
KtorZ, line 61: This section needs to contain an implementation plan approved by the Plutus team. Once this is covered, we can proceed and merge it as _Proposed_.

CIP-0069/README.md
```diff
@@ -0,0 +1,65 @@
...
+
+Use Case 3 (taken from Minswap's Dex V1): NFT is minted for the same reason as above. It allows a minting policy to later mint LP tokens with that unique id token name.
+
+We see a similar pattern repeating over and over again as witnessed by dapp developers and auditors alike. By allowing the multi-purpose policies (spending and any other) we increase the security of Cardano by giving us more confidence and allowing to design protocols that have their architecture driven by Cardano's features, not limited by Cardano's language.
+
+This primarily manifests in the ability to use a single validator for both minting and spending but the proposed solution makes it possible to use one validator for any and all purposes at once.
+
+## Specification
+
+### Removing the datum argument
```
michaelpj, line 35: This is not yet a sufficient specification. The ledger does not know about any of these types. The redeemer and datum are passed as values of type `Data`. So: what argument _exactly_ is going to be passed instead?

michaelpj:
An objection that occurs to me. This means that it is no longer as simple to run a script as it used to be: you can't just apply it to the redeemer and the datum, you now need to construct an additional structure. (Indeed, writing this I realised that I don't even know what that structure is, hence my comment in the text). I don't think this is a killer argument: it's already non-trivial to run a script since you need the `ScriptContext`, which really only the ledger can produce easily. But it's worth pointing out as a consequence.

I think this is broadly a good idea, but it's a big change to the interface and I'm a bit nervous that there are problems I haven't thought of. So I'd probably like to think about it more before I'm wholly in favour.

I also don't think the Plutus team will implement this any time soon - it doesn't seem high enough priority.

L-as:
> to run "a script"

Isn't it the other way around? Before you had to special-case validators.

> I also don't think the Plutus team will implement this any time soon

Maybe MLabs can take this on? I imagine it will only need changes to cardano-ledger's Alonzo, Babbage eras and also plutus-ledger-api? Seems simple.

CIP-0069/README.md
zygomeb: Let's discuss this in the CIP editors meeting

CIP-0069/README.md
zygomeb: I'm not sure exactly what is the answer here. Would you have any suggestions?

L-as: You need to look at it from the persective of cardano-ledger

L-as: not plutus-ledger-api (which doesn't do anything)

CIP-0069/README.md
rphair: FYI @zygomeb the next meeting, after having been delayed 1 week, has now been postponed 1 additional day ([CIP Editors Meeting #58 (30 Nov, 9:30am UTC)](https://discord.com/events/971785110770831360/1036940834932793363))

KtorZ:
As discussed in today's meeting, we're postponing merging this proposal until we've gotten a chance to discuss with the author (@zygomeb) here: https://github.com/cardano-foundation/CIPs/pull/321#discussion_r1031891589

fallen-icarus:
The discussion here seems to only be considering removing the datum from spending scripts in order for it to match the format of other scripts. Why not go the other way? For example, why not make both spending and minting scripts take a `Maybe Datum`? Then whenever the universal script is used to mint, `Nothing` would be passed for the datum automatically. This option would also make it possible to spend script utxos that have missing/incorrect datums as well as likely be a much smaller change than what is being discussed here.

Based on my understanding of de bruijn levels, as far as the low level encoding for the `Maybe Datum` goes, an extra de bruijn level could be added to the datum before passing it to the script. The script would then immediately unwrap this extra layer with pattern matching. For example, if you had the unit datum:
``` Haskell
unitDatum :: Data
unitDatum = Constr 0 []
```
and you wanted to pass this in for `Just ()`, you would add an outer de bruijn level like:
``` Haskell
maybeDatumPassedToScript :: Data
maybeDatumPassedToScript = Constr 1 [Constr 0 []]  -- ^ Unit datum is inside the list
```
The `Nothing` encoding would be:
``` Haskell
nothing :: Data
nothing = Constr 0 []  -- ^ This is not the same as the unit datum since it is a different de bruijn level.
```

The script would immediately unwrap these like:
``` Haskell
mkUniversalScript :: Maybe Datum -> Redeemer -> ScriptContext -> Bool
mkUniversalScript (Just datum) r ctx = ...
mkUniversalScript Nothing r ctx = ...
```

This idea is discussed in my proposal #418 which seeks to allow executing spending scripts once per transaction (as opposed to once per utxo) just like minting and staking scripts. If others agree that this `Maybe Datum` approach could work, I believe we should try combining this proposal with mine.

michaelpj:
> The discussion here seems to only be considering removing the datum from spending scripts in order for it to match the format of other scripts. Why not go the other way? For example, why not make both spending and minting scripts take a Maybe Datum?

The version [I proposed](https://github.com/cardano-foundation/CIPs/pull/321#issuecomment-1250053504) would pretty much allow that: we could simply pass `RedeemerOnly red` to validator scripts if there was no datum. In fact I guess that's a virtue of it being more loosely typed.

> Based on my understanding of de bruijn levels...

I'm sorry, this section didn't make much sense to me. If you're proposing to use the builtin Plutus `Data` type to encode the alternatives, then yes, I think that's the most plausible option at the moment.

> If others agree that this Maybe Datum approach could work, I believe we should try combining this proposal with mine.

I haven't read your proposal yet but looking at the motivation it seems fairly orthogonal.

fallen-icarus:
@michaelpj 

> The version [I proposed](https://github.com/cardano-foundation/CIPs/pull/321#issuecomment-1250053504) would pretty much allow that: we could simply pass RedeemerOnly red to validator scripts if there was no datum.

I may have misinterpreted your proposal. However, I do not believe the (hopefully) correct interpretation changes my stance. Just to make sure we are on the same page, I am proposing having all scripts written like this:
``` Haskell
mkUniversalScript :: Maybe Datum -> Redeemer -> ScriptContext -> Bool
```
If you are proposing this for all script types:
``` Haskell
mkUniversalScript :: ScriptArgs -> ScriptContext -> Bool
```
then I agree that our proposals are functionally the same. (The first time I read it, I interpreted it as only changing spending scripts to this.)

However, despite being functionally the same, you also later pointed out
> An objection that occurs to me. This means that it is no longer as simple to run a script as it used to be: you can't just apply it to the redeemer and the datum, you now need to construct an additional structure. (Indeed, writing this I realised that I don't even know what that structure is, hence my comment in the text).

As you noted, the complexity involved with implementing your version is an issue. My version (which you have suggested using in other unrelated threads) seems like it would be much simpler to implement since it can make use of most tooling as is.

> I'm sorry, this section didn't make much sense to me. If you're proposing to use the builtin Plutus Data type to encode the alternatives, then yes, I think that's the most plausible option at the moment.

I may be misusing terms... But yes, that is how I am suggesting implementing `Maybe Datum` for all scripts.

> looking at the motivation it seems fairly orthogonal.

My proposal is mostly, but not entirely orthogonal to this one. The issue in my proposal would also be solved using `Maybe Datum` which means both issues can be solved in parallel.

michaelpj:
> As you noted, the complexity involved with implementing your version is an issue. My version (which you have suggested using in other unrelated threads) seems like it would be much simpler to implement since it can make use of most tooling as is.

The version that requires using `Maybe` still requires additional processing of the arguments. I think we lose some simplicity when we go from "just apply the function to the arguments" to "you have to do something to the arguments first", and I don't it's much better if the thing you do to the arguments is slightly simpler.

fallen-icarus:
@michaelpj I agree that the `Maybe Datum` idea is still more complicated than the current design. However, I believe the point of CIPs like this (whether explicitly stated or not) is that the current simple design IS the problem.

It seems that the current design of "just apply the function to the arguments" is only possible because it makes the strong assumption that **spending scripts always directly use the datum attached to the utxo being spent**. This assumption does not match the real world use of spending scripts and this mismatch results in:

1. Being unable to efficiently use a spending script that validates based off of the transaction as a whole instead of any individual utxo. The eUTxO model already allows creating such scripts but the current simple design forces these "transaction level" scripts to be executed once for each utxo being spent from the corresponding address. This means that, if you spend 3 utxos from this address in the same transaction, this script will be executed 3 times even though the transaction context is exactly the same for each execution. A real world use case is an atomic swap where all that matter is, within one transaction, the value leaving the address has to be some proportion of the value entering the address. While datums are used in the atomic swap scenario, they are used collectively to calculate the weighted average price; they are not used individually like assumed.
2. Being unable to spend utxos that do not have a datum or have incorrect datums. There are plenty of use cases where a datum is just not necessary. Right now, a place-holder datum is required just to prevent the funds from getting locked. A real world use case is a DAO where users just need to deposit funds into the script address. I think it is also worth mentioning that regulators are not going to like any application where it is easy to accidentally lock users' funds. With the current design, this ease of locking is the default on Cardano.
3. Being unable to use the same script for both minting and spending since minting scripts do not take a datum. This makes it harder to secure certain scenarios where minting and spending are inextricably linked. A real world use case is to only mint a token if a deposit is made to the corresponding script address.

Developers have come up with workarounds for each of these situations but all of the workarounds have their own drawbacks:

1. To get around the redundant executions of "transaction level" scripts, you can use a staking script for the actual logic and only have the spending script check that the required staking script is executed in the transaction. This does not stop the redundant executions since the spending script will still redundantly check that the required staking script is executed. Further, this technique now means that two scripts are needed for every application instead of just one.
2. To help prevent accidental locking due to missing/incorrect datums, proxies are generally used. However, it is currently very hard to use these proxies without the proxies themselves being centralized. In other words, the Dapp loses some of its decentralization.
3. To link minting and spending, the spending script hash can be hard coded into the minting policy and the minting policy hash (Currency Symbol) can be passed to the spending script in the datum. However, given the mutability of the datum, this use of the datum makes this approach less secure by default than if the same script could do both minting and spending.

I understand the reluctance to add complexity to the ways things are currently done. However, if my understanding of the situation is correct, then the real world use of Cardano smart contracts really needs the above strong assumption to be relaxed. Intuitively, I don't see how it is possible to relax that assumption without adding a little more complexity. As far as whether this should be a priority, there have now been (at least) 6 proposals created whose motivations can all be boiled down to "the above assumption is too strong for the real world." (#364, #423, #309, #418, #310, and this one)

zygomeb:
I was notified that the CIP was waiting for me. I was unaware that I was to do something -- aside from the discussions that I think are of now best moved to the Maybe Datum PR, do we need to do something before approving this? 

Wholeheartedly I do believe that making datum not a mandatory arguments via maybe datum is the simpler and likely better solution however this is a simple compromise that'll allow us to get a lot of work done.

nielstron:
> I'm not sure this would actually require a change to the ledger or Plutus. If you think of it from the perspective of Plutus core I'm sure you would be able to write a script that works whether there's a datum present or not, and I think you should also be able to properly look at all the arguments (but maybe my understanding of how Plutus core is executed is wrong here). If that's the case, then this could be done with any changes in the node (and thus no HF).

Yes its possible to emulate the described behavior already but makes some ugly assumption about the Redeemer, I.e. the way it is done in eopsin: https://github.com/ImperatorLang/eopsin/blob/master/ARCHITECTURE.md#minting-policy---spending-validator-double-function

I'd be interested in aligning this "hack" more with the proposed result of this discussion.

nielstron:
@fallen-icarus Actually with the current design, the same script can do spending and minting (on a UPLC level only though). See my comment above.

fallen-icarus:
@nielstron  It's cool that you found a way to make it work but

> The single only drawback of this approach is that the second argument to a validator may never be PlutusData with constructor id 0

this is a pretty hefty cost IMO. Currently most tooling does not help a developer avoid using the constructor id 0 which means it now falls on the developer to remember. Indeed, most tooling encourages using the 0 constructor. This makes it another potential place where normal human error will cause issues. The hack seems to be a "chicken or egg" situation: the cost is only "bearable" if the tooling is already there to guide developers into using it correctly.

> It can also nicely be emulated using the hack described above.

I'm not sure how you handle plutus versions with eopsin but if the Maybe Datum proposal is implemented, AFAICT the hack wouldn't be necessary at all for plutus v3 scripts since all plutus v3 scripts would have the same signature.

nielstron:
@fallen-icarus Eopsin guides you through this. And sensible tooling that doesn't disclose the contructor ID to the developer (which should be the case for very clean tools) can simply start counting at 1 when assigning ids.

Yes for plutus v3 its not necessary to do this workaround. But one could already start developing with the maybe datum schema in eopsin and PlutusV2 using the hack. the tool will then transparently upgrade to PlutusV3 and removes the hack from the compilation chain when the new plutus version is availalable.

CIP-0069/README.md
```diff
@@ -0,0 +1,65 @@
...
+data ScriptArgs =
+    RedeemerOnly Redeemer
+  | RedeemerAndDatum Redeemer Datum
+```
+
+## Rationale
+
+Unifying of the script signature is a very elegant solution to the problem, streamlining the experience of developing on cardano.
+Given that accessing the datum is almost always made by a spending script, it makes sense to introduce that argument back to the `ScriptPurpose` that now plays a more important role.
+It begs the question if it should be added as an argument to all validators, to further emphasize that fact.
```
michaelpj, line 52: I found these last two sentences difficult to understand.

michaelpj:
I still want to do something like this.  

Problems:
- This is still lacking a proper specification. But I think that will work well with https://github.com/input-output-hk/plutus/pull/5099, because then we can use native Plutus Core sums-of-products to encode the `ScriptArgs` argument, instead of relying on `Data`.
- There are no acceptance criteria. I propose: a) `plutus` and `cardano-ledger` have been updated with the new design, b) the new design has been incorporated into a new ledger language
- There is no implementation plan. _Maybe_ the Plutus team might do it, it depends on how much time we have before the next HF and Plutus ledger language.
- I am still generically worried about the design. I am sorry that this is so vague, but these sorts of design decision are very sticky so I would really like to be confident that we're moving to somewhere good. The existing design is a clear point in the space (about as simple as you can get), if we make things more complex we have far more options.
    - For example, the question of whether the `ScriptPurpose` should be part of the `ScriptContext` argument or its own argument.

Nonetheless, I think we should merge this CIP as Proposed. Ideally after the above have been addressed (except for the last one).

L-as:
Seems to me like scripts should all take only one argument, which is a SOP. One index/constructor per script type, which includes per-script data and info:
```haskell
data ScriptInfo = Minting Redeemer CurrencySymbol TxInfo | Spending Redeemer Datum TxOutRef TxInfo | ...
```
This seems simpler than the current design while being more powerful IMO.

Thoughts? @michaelpj

michaelpj:
That's another possibility. This is what I mean when I say the design space is bigger now...

Note that I now think that having a separate `ScriptPurpose` may actually be desirable: that allows us to, say, run a spending script with no datum by running `script (RedeemerOnly redeemer) (Spending txOutRef) txInfo`. Your version can't do that because we have to link the script purpose to the expected arguments. There are quite a few desires to be satisfied here!

However, we could still have a single SOP for all arguments, and just have it have a `ScriptPurpose` field.

The main advantage I can see to your proposal is that it would make it easier to add new call patterns in future that change how the script context is passed, e.g. not including the `TxInfo`. I'm not sure if that's desirable or not...

imikushin:
I've been pointed to this discussion after I commented here: https://github.com/cardano-foundation/CIPs/pull/418#issuecomment-1440744894

Is what I'm describing feasible? If yes, would it be a simpler and smaller change to make, than proposed in this CIP?

Re-posting it here for convenience:

---
@fallen-icarus 

I really like the change to ScriptPurpose, suggested in https://github.com/cardano-foundation/CIPs/pull/418#issuecomment-1427046710

I'd perhaps want to use `TxInInfo` instead of `TxOutRef`:

```hs
data TxInInfo = TxInInfo
    { txInInfoOutRef   :: TxOutRef
    , txInInfoResolved :: TxOut
    }
```

```hs
data SpendingRef
  = SingleUTxORef TxInInfo
  | GroupedUTxORef [TxInInfo]

data ScriptPurpose
    = Minting CurrencySymbol
    | Spending SpendingRef
    | Rewarding StakingCredential
    | Certifying DCert
```

This would be great, because then we wouldn't need to spend extra cycles to get the `TxOut`. This could be further improved by adding `txOutDatumResolved` to `TxOut`:

```hs
data TxOut = TxOut {
    txOutAddress       :: Address,
    txOutValue         :: Value,
    txOutDatumHash     :: Maybe DatumHash
    txOutDatumResolved :: Maybe Datum
    }
```

With these simple changes, we'd be able to eliminate the datum argument (being passed to validation scripts) as it'd be trivial to access it via ScriptContext.

This feels (to me personally) as a fairly incremental and simple change. Also, what do you think, @michaelpj ?

michaelpj:
Please don't cross-post like this, now I don't know where to respond. I'm going to respond in the other CIP where you brought it up.

rphair (approved):
@zygomeb @michaelpj @L-as @KtorZ - this was reviewed in the CIP meeting today which had no Plutus representatives present.  Present editors' feeling based on this proposal's age and acknowledgement as at least a working alternative (via comments above) is that should be merged as `Proposed` pending Last Check... or deprecated in the equally near future if any irreconcilable problems.

L-as:
Thank you!

rphair (approved):
The last changes for CIP-0001 compliance, plus the round of agreement at the last meeting, I think make this ready to merge at the meeting coming up in a few minutes.  In fact @KtorZ @Ryun1 @SebastienGllmt I think we should treat this as a `Last Check` item on the agenda and be ready to merge it on the spot unless there are any objections.

KtorZ:
I missed the last editor meeting so I'll try to catch up in the coming one but .... it seems to me that the CIP is far from being in a "proposable" state? As far as I can tell, MPJ had valid comments regarding how the specification isn't detailed enough. 

While the motivation is clear, I don't think the proposed solution is sufficiently specified?

KtorZ (approved):
Given the following comment: https://github.com/cardano-foundation/CIPs/pull/321#issuecomment-1418993992

I am in the end _okay_ merging this CIP for the sake of recording a proposed solution to #497. The specification is still incomplete and insufficient for someone to implement this, and there's clearly no implementer. but it has the merit of being a well-formulated idea and motivation to a known problem.

PR #322, opened 2022-08-23
CIP-0053? | Multi-Script Hashes

blamario:
This is a proposal to generalize the script hashing functionality, so that multiple mutually-depended scripts can be assigned fixed addresses at the same time.

---

[see rendered Markdown](https://github.com/blamario/CIPs/blob/MultiScriptHashes/CIP-MultiScriptHashes/README.md)

### DIFF
--- a/CIP-MultiScriptHashes/README.md
+++ b/CIP-MultiScriptHashes/README.md
@@ -1,0 +1,95 @@
+---
+CIP: "???"
+Title: Multi-Script Hashes
+Authors: Mario Blažević <mario@mlabs.city>
+Discussions-To: TBD
+Comments-Summary: TBD
+Comments-URI: TBD
+Status: Draft
+Type: Standards Track  
+Created: <date created on, in ISO 8601 (yyyy-mm-dd) format>  
+License: CC-BY-4.0  
+License-Code: Apache-2.0  
+Post-History: [2022-06-09](https://forum.cardano.org/t/cip-draft-multi-script-hashes/102845)
+---
+
+## Abstract
+
+This is a (draft) CIP to introduce a simple way to statically identify a collection of scripts that depend on each
+other, so that each script can detect if a transaction tries to substitute an outside script address for an expected
+one.
+
+## Motivation
+
+Most non-trivial DApps will be distributed not only in the usual sense, but will also have their on-chain logic spread
+across multiple validator scripts. Some types of transactions will then be expected to consume and create UTxOs
+belonging to different scripts.
+
+To take an abstracted example, consider a DApp with only two validators scripts, `S₁` and `S₂`. There are transactions
+`T₁` that deal only with the first validator script, transactions `T₂` that deal only with the second, and
+transactions `T₁₂` that involve both. A `T₁₂` might for example simply move some funds from the first script to the
+second. The question is, how can the `S₁` ensure that the recipient is in fact `S₂`?
+
+The simplest solution is that we first compile `S₂`, then supply its `ValidatorHash` or its `Address` as a parameter
+to `S₁` which can trivially compare it to the recipient's address.
+
+If, however, there is also a valid transaction `T₂₁` that moves funds from `S₂` to `S₁`, we can't apply the same
+solution: there is currently no way to resolve a mutual dependency between scripts.
+
+There are workarounds to guard against the substitution. We can mint a limited amount of a token, for example, supply
+its `CurrencySymbol` as a parameter to all scripts, and then ensure that every UTxO locked by any script in our DApp
+carries one of these tokens as an identification badge. This however introduces unnecessary complexity and increases
+all transaction costs. It would be better to solve the root issue of mutual dependency.
+
+## Specification
+
+The minimum required functionality can be provided with just two additional functions in Plutus API, and a new field
+in `TxInfo`:
+
+```
+validatorHashes :: [Validator] -> [ValidatorHash]
+allScripts :: [Validator] -> ScriptLookups Any
+txInfoValidators :: TxInfo -> [ValidatorHash]
+```
+
+The new `validatorHashes` function is a generalization of the existing
+
+```
+validatorHash :: Validator -> ValidatorHash
+```
+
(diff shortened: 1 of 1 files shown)


L-as:
#321 is IMO a better solution, albeit not perfect

CIP-MultiScriptHashes/README.md
```diff
@@ -0,0 +1,95 @@
...
+
+If, however, there is also a valid transaction `T₂₁` that moves funds from `S₂` to `S₁`, we can't apply the same
+solution: there is currently no way to resolve a mutual dependency between scripts.
+
+There are workarounds to guard against the substitution. We can mint a limited amount of a token, for example, supply
+its `CurrencySymbol` as a parameter to all scripts, and then ensure that every UTxO locked by any script in our DApp
+carries one of these tokens as an identification badge. This however introduces unnecessary complexity and increases
+all transaction costs. It would be better to solve the root issue of mutual dependency.
+
+## Specification
```
michaelpj, line 44: This specification is not specific enough. I can't figure out whether you're just proposing changes to off-chain code or what. It seems like you _must_ want some changes for on-chain code but I can't figure out what, and that's the most important thing.

michaelpj (commented):
I'm finding this CIP very hard to follow, sorry.

CIP-MultiScriptHashes/README.md
blamario: The new field of `TxInfo`, namely 
```
txInfoValidators :: TxInfo -> [ValidatorHash]
```
is the proposed on-chain API modification. How should I make it clearer?

michaelpj: What does it do? Can you just explain specifically what is in that field and how it solves the problem?

blamario: That's in there as well, but I certainly could have made the explanation more prominent:

> The `[ValidatorHash]` list calculated by `allScripts` would be made available as the `txInfoValidators` accessor so each validator triggered by the transation could check it against the addresses of transaction's UTxOs.

In other words, the on-chain `txInfoValidators` field would contain `validatorHashes vals` where `vals` is the argument supplied off-chain to `allScripts` as part of the transaction lookups. The resulting list of validator hashes would become a part of `TxInfo`.

michaelpj: I'm sorry, that doesn't make any sense to me. The fields of `TxInfo` must be computed from the transaction itself. How is this field computed? How does the function `mkTxInfo :: Tx -> TxInfo` change?

blamario: As we build the transaction, the `allScripts vals` lookup will presumably result in a new constructor of `ScriptLookups`, say
```
slValidators :: [Validator]
```
The same value would be carried in `Tx` as a field: 
```
txValidators :: [Validator]
```
To compute the `txInfoValidators` field, we just apply `validatorHashes` to this, as in `TxInfo{..., txInfoValidators= validatorHashes (txValidators tx)}`. Is this clear enough?

KtorZ:
I feel that this proposal is not "general" enough; it currently works from the plutus-tx API which is one of the possible interfaces to deal with Plutus -- which leads, I believe, to some confusion as to what the proposal is trying to achieve. 

I'd suggest to re-formulate this from a more general point of view, taking Plutus core concepts irrespective of the framework used to represent those concepts. In essence, I understand that you are proposing a way for inter-dependent scripts to have ways to reference one another. How is that already not possible when providing configuration-as-datums to the respective script (would love to see this argued somewhere)?

CIP-MultiScriptHashes/README.md
michaelpj: Okay, so you're adding a new field to transactions. That's a big deal! I suggest you drop the stuff about the Plutus API and focus on that: how are you changing transactions and transaction validation. The rest is important additional information, but not the core change.

blamario:
Storing the dependent script hashes in datums is an alternative solution, but it's not trivial to get right because datum on its own can't be trusted. If dApp is not properly designed, attackers can perform a two-step attack:
1. send a UTxO to a script's address with a datum containing other scripts hashes of their own choosing, and then
2. submit a transaction that consumes the UTxO from step 1, other outputs from their own scripts, and performs other effects that escalate their access.

The "other effects" part may or may not exist for any particular dApp, but the point is that this may not be easy to prove. In contrast, the `txInfoValidators` would carry a static, compiled list of script hashes that can't be manipulated by any attacker. The only check that any validator would need to perform is
```
all (all (`elem` txInfoValidators info) . scriptHashAddress) . txOutAddress . txInInfoResolved) (txInfoInputs info)
```

NetWalker108:
> Storing the dependent script hashes in datums is an alternative solution, but it's not trivial to get right because datum on its own can't be trusted. If dApp is not properly designed, attackers can perform a two-step attack:
> 
> 1. send a UTxO to a script's address with a datum containing other scripts hashes of their own choosing, and then
> 2. submit a transaction that consumes the UTxO from step 1, other outputs from their own scripts, and performs other effects that escalate their access.
> 
> The "other effects" part may or may not exist for any particular dApp, but the point is that this may not be easy to prove. In contrast, the `txInfoValidators` would carry a static, compiled list of script hashes that can't be manipulated by any attacker. The only check that any validator would need to perform is
> 
> ```
> all (all (`elem` txInfoValidators info) . scriptHashAddress) . txOutAddress . txInInfoResolved) (txInfoInputs info)
> ```
> 
> [EDIT] ... besides, obviously, the checks specific to their own input.

Can't the use of Native Assets serve as guards against this sort of attack? Where the only appropriate UTXOs + Datums the script/dApp should ever interact with will be ones explicitly marked with specific Native Assets, provided by dApp provider/author.

blamario:
The major problem with this proposal is the transaction size. Every transaction that needs to invoke *any* validator script that's interconnected with others would need to supply *all* interconnected scripts, otherwise the hash won't match.

CIP-33 could help with this problem. Unfortunately I wrote up the proposal before reference scripts were available, so that part is lacking.

blamario:
I've decided I have no further interest in this proposal, and it doesn't look like anybody else does.

PR #329, opened 2022-09-17
CIP-0036 - Map keys are unsigned int values not strings

gitmachtl:
removed the quotes from the maps, because keys are unsigned values and not strings.


rphair (approved):
it's true; the CIP itself shows an integer data type:in the example:
```
61284: {
  1: [["0xa6a3c044...
```
... but a string in the test vector, corrected in this PR.

CIP-0015 has the same issue... @gitmachtl although that one might be considered deprecated, can you make the correction there as well: for completeness' sake & for the historical record?

gitmachtl:
@rphair jup... https://github.com/cardano-foundation/CIPs/pull/330

KtorZ:
Although... the examples / test vectors are Javascript. All object keys in Javascript are strings (or symbols but let's not get there).

Writing `{ 1: 'foo' }` is equivalent to `{ '1': 'foo' }` in Javascript, the key just get coerced to a string.

rphair:
> @KtorZ: test vectors are Javascript. All object keys in Javascript are strings

yes, I realised that as well... I just think it makes sense to honour this request for the nomenclature to be consistent, especially with the integer declaration in CBOR.  People might not use the text vectors except to illustrate the data & types, so making this change would be helpful for that purpose.

PR #330, opened 2022-09-18
CIP-0015 - Map keys are unsigned int values not strings

gitmachtl:


rphair:
My suggestion (https://github.com/cardano-foundation/CIPs/pull/329#pullrequestreview-1111493718) was to include this with #329, but maybe I didn't suggest it literally enough.  Anyway now that we have two PR's for the same purpose please let me suggest that all discussion of this PR continue in https://github.com/cardano-foundation/CIPs/pull/329... since we would implement either both these changes or neither, and the considerations are identical for both.

rphair (approved):
@KtorZ shall we then also approve & merge this one, since it's a twin of the merged https://github.com/cardano-foundation/CIPs/pull/329 for CIP-0036?

KtorZ (approved):

PR #332, opened 2022-09-23
CIP-0034 | Update registry.json to reflect Preview and Pre-Production test networks

Ryun1:


rphair (approved):
I don't have experience in this area but I confirm the hashes match the configs in the reference document.  @SebastienGllmt can you confirm the values are correct?

CIP-0034/registry.json
```diff
@@ -1,14 +1,20 @@
 {
+    "Pre-Production": {
```
SebastienGllmt, line 2: I would avoid dashes in key names since messes with Javascript dot notation

KtorZ:
Orthogonal to this proposal, we should probably seek to disambiguate what "genesis hash" refers to. There are now technically 3 genesis hash per network (Byron, Shelley & Alonzo) and maybe more in the future. 

Might be relevant to #323

KtorZ (approved):

PR #333, opened 2022-09-23
CIP-0059 | Add dates and slot numbers to the table

Quantumplation:
Based on the block here:

https://cexplorer.io/block/eea1247726ababb0b15ef7068b6917ceb6ebe3021c40fe44608585bba44e24b6

Amusingly, the first block in the babbage era has `babb` as a substring :)

### DIFF
--- a/CIP-0059/feature-table.md
+++ b/CIP-0059/feature-table.md
@@ -9,4 +9,4 @@
 | 2021/03 | Goguen  | Mary    | 23068800    | 251          | 4,0              | TPraos          | Ouroboros Praos     |                    |
 | 2021/09 | Goguen  | Alonzo  | 39916975    | 290          | 5,0              | TPraos          | Ouroboros Praos     |                    |
 | 2021/10 | Goguen  | Alonzo  | 43372972    | 298          | 6,0              | TPraos          | Ouroboros Praos     | intra-era hardfork |
-|         | Goguen  | Babbage |             |              | 7,0              |  Praos          | Ouroboros Praos     | Vasil HF           |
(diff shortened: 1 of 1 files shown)


rphair:
numbers confirmed accurate ✔️ ... @JaredCorduan what do you say?

> Amusingly, the first block in the babbage era has` babb` as a substring :)

seems like a little less than 1 chance in 1000 of this happening, unless I've gotten the maths wrong 🤔

rphair (approved):

JaredCorduan:
> There was some ambiguity (from my limited point of view) with full functionality not being available for exactly 1 more epoch, so maybe you could settle this fine point. monocle_face

Looks perfect to me! Even though the cost model for Plutus V2 won't be present until the next epoch, the Babbage ledger era has begun! thanks @Quantumplation !

Quantumplation:
@JaredCorduan I'm curious, I don't remember Alonzo going through the "5 days until the cost models are available" dance. Am I misremembering, or why was that different?

JaredCorduan:
good question @Quantumplation

No, Alonzo did not have to do this song and dance since it was the first ledger era with a cost model (or more specifically, since it got the cost model from the Alonzo genesis file, and Babbage did not have a genesis file). The root cause is an inflexible serialization scheme, which is described here: https://github.com/input-output-hk/cardano-ledger/issues/2902

The good news is that we have a plan for how to avoid this is the future (the plan is described in the ticket). We completed most of the plan already, and just need https://github.com/input-output-hk/cardano-ledger/issues/3014 still.

KtorZ (approved):

PR #335, opened 2022-09-24
CIP-0036 | Not allow duplicated vote-key entries

gitmachtl:
Added lines to make it clear that duplicated voting key entries (delegating weighted voting power) are not allowed in the registration metadata format and should be handled as invalid if such entries are included. This keeps away confusion about how the weight for each voting key should be calculated and its also mandatory for hw-wallets showing the right amount of weight for each voting key entry. @janmazak

### DIFF
--- a/CIP-0036/README.md
+++ b/CIP-0036/README.md
@@ -71,6 +71,8 @@ Recall: Cardano uses the UTXO model so to completely associate a wallet's balanc
 
 Given the above, we choose to associate staking credentials with voting keys. At the moment, the only supported staking credential is a staking key. Since most Cardano wallets only use base addresses for Shelley wallet types, in most cases this should perfectly match the user's wallet.
 
+No duplicated voting key entries are allowed in the voting key delegegation array, to remove weighting confusion and to support hw-wallet handling. HW-Wallets with there limited mem space, have to display the proper weight values for a voting key entry. Registration metadata containing duplicated voting key entries should be handled as invalid.
+
 The voting power that is associated with each delegated voting key is derived from the user's total voting power
 as follows.
 
@@ -79,6 +81,7 @@ as follows.
 3. The voting power associated with each voting key in the delegation array is calculated as the weighted fraction of the
    total voting power (rounded down);
 4. Any remaining voting power is assigned to the last voting key in the delegation array.
+
 
 This ensures that the voter's total voting power is never accidentally reduced through poor choices of weights,
 and that all voting powers are exact ADA.
@@ -113,6 +116,7 @@ considered valid if the following conditions hold:
 - The reward address is a Shelley address discriminated for the same network
   this transaction is submitted to.
 - The delegation array is not empty
+- The delegation array does not contain multiples of the same public voting key
 - The weights in the delegation array are not all zero
 
 
--- a/CIP-0036/README.md
+++ b/CIP-0036/README.md
@@ -116,6 +116,7 @@ considered valid if the following conditions hold:
   this transaction is submitted to.
 - The delegation array is not empty
 - The delegation array does not contain multiples of the same public voting key
+- The stake public key is different from all public voting keys in the delegation array (derived from a different path)
 - The weights in the delegation array are not all zero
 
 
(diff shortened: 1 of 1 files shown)


rphair:
cc @Zeegomo @kevinhammond @mark-stopka

gitmachtl:
Also, why is the weight starting at 0? That makes no sense and would only lead to divison/0 errors. There is no reason someone would include a vote public key in the delegation array with weight 0 and other entries with weight>0. So the smallest weight value should be set to 1 imo.

Scitz0:
Having clear rules defined in CIP for how to deal with edge cases like this is beneficial for wallet implementations. 
The two proposed changes to clarify that there should not be duplicate voting key entries and that vote weigt need to be > 0 sounds like a good addition to me.

SebastienGllmt (approved):
LGTM if the Catalyst team signs off

gitmachtl:
So, we had a discussion in the last days on Slack in the iog-governance channel. Outcome is to allow duplicated keys. Signing-/Generation-Tools can still decline to do it that way, but the Catalyst receiving side will allow duplicated key in the delegation array. Also, its ok to show each individual entry of the delegation on the gui of the hw-wallet, and not doing a sum-up. So there is no need to check for duplicated keys during the hw-wallet signing process. Also, vote-weight=0 is a valid value. Because Catalyst with ID=0 is only one usecase, and there might be other voting where 0 could be an important value. 

Another thing came up, that there should be an info that using an empty delegation-array equals to a deregistration. So an empty delegation array is also allowed. An additional check i came up with was that none of the vote-public-keys are the same as the public-signing-key. The vote keys are derived from a new path, the signing key is the stake key. So if any vote-public-key is equal to the public key of the signing key -> something is wrong. Either that specific vote-key is wrong, or the signing key is.

With all these changes, i think i have to form a new PR...

gitmachtl:
ok, so when a vote-weight=0 is a valid value in the delegation array, why is than the condition 
`The weights in the delegation array are not all zero` in CIP-0036 a needed condition?

We allow to set the vote-weight to zero, but not for all vote-public-keys? An empty delegation array is now seen as a deregistration, wouldn't be a delegation array with all vote-weights set to zero also just result in the same outcome?

gitmachtl:
This is now a valid registration, which would remove all delegated voting power:
```json
{
  "61284": {
    "1": [],
    "2": "0x57758911253f6b31df2a87c10eb08a2c9b8450768cb8dd0d378d93f7c2e220f0",
    "3": "0xe0c13582aec9a44fcc6d984be003c5058c660e1d2ff1370fd8b49ba73f",
    "4": 1234567,
    "5": 0
  },
  "61285": {
    "1": "0xb8382668e6e65c42718412f297a3c35d3beabbf85d57560f9c9658e1e2edb40a3f036b07afffa31745242f9b02cf89a59e9ba6391e83da2fe7b8c484b913640c"
  }
}
```

And this would also be valid if we allow vote-weight=0 and duplicated entries:
```json
{
  "61284": {
    "1": [
      [
        "0x51f117d26e29aea7db3d1f2f874ab5f585f619a95aed6d71d31a7404cb6557b5",
        0
      ],
      [
        "0x51f117d26e29aea7db3d1f2f874ab5f585f619a95aed6d71d31a7404cb6557b5",
        0
      ]
    ],
    "2": "0x57758911253f6b31df2a87c10eb08a2c9b8450768cb8dd0d378d93f7c2e220f0",
    "3": "0xe0c13582aec9a44fcc6d984be003c5058c660e1d2ff1370fd8b49ba73f",
    "4": 1234567,
    "5": 0
  },
  "61285": {
    "1": "0xdcd2a450f4a9a6790a4e0147b26a7be3b75d297d18a99405be6f15b07bf8f45df98d26a2ecd9d4d0a118d9f003ffd6226bc5fcf2fe48c68a8e045d2439ede600"
  }
}
``` 

Its a bit inconsistent. On one hand we're not saying that the vote-weight should be > 0 to avoid "invalid" delegations, on the other hand its not allowed to have vote-weight=0 for all vote-public-keys?

gitmachtl:
also related to allow exporting vote-public-keys on the wallets with a known prefix:
https://github.com/cardano-foundation/CIPs/pull/342

gitmachtl:
closed, will be done via a new PR

PR #331, opened 2022-09-21
CIP-0001 | Rework to reflect reality

KtorZ:
This is the first step in a broader work aiming at pushing the CIP in the forefront of _every_ Cardano changes. Before addressing current shortcomings of the process, it is necessary to capture what the process _actually is_ at present.

This rework will serve as a base for a retrospective discussion amongst the editors and, shall be proposed to the 
community for approval and review.

---

[see rendered Markdown](https://github.com/cardano-foundation/CIPs/blob/cip-0001-rework/CIP-0001/README.md)

PR #336, opened 2022-09-26
CIP-0070? | Add rejected CIP for spending policies

michaelpj:
This adds a rejected CIP recording a possible design that has been discussed in the past for "spending policies". I wanted to have somewhere to point when it comes up, and also to publicly record the reasons why designs like this are difficult.

---

[see rendered Markdown](https://github.com/michaelpj/CIPs/blob/mpj/spending-policies/CIP-0070/README.md)

### DIFF
--- a/CIP-0070/README.md
+++ b/CIP-0070/README.md
@@ -1,0 +1,135 @@
+---
+CIP: 70
+Title: Spending policies
+Authors: Michael Peyton Jones <michael.peyton-jones@iohk.io>
+Status: Rejected 
+Type: Standards Track 
+Created: 23/09/2022
+License: CC-BY-4.0
+---
+
+## Abstract
+
+IMPORTANT: this CIP exists to record a design which has been considered in the past and rejected.
+No version of this CIP is being proposed for inclusion in Cardano currently.
+
+This CIP proposes to allow native tokens on Cardano to have a _spending_ policy in addition to their _minting_ policy.
+The spending policy would be run whenever those tokens were spent, allowing control over how they are spent and moved.
+
+## Motivation
+
+Native tokens on Cardano today are associated with a minting policy that controls how the tokens can be minted or burned.
+This allows user control of the _supply_ of tokens.
+
+However, once tokens have been minted, they follow the normal accounting rules of Cardano's ledger. 
+This is one of the things that justifies calling them "native", but it makes it difficult to implement some use cases that really _do_ want to change the basic accounting rules that their tokens follow.
+
+Today, it is possible to work around this problem by using Cardano's main feature for controlling the spending of outputs: script-locked outputs.
+However, using this systematically typically means that such tokens must always be locked by a script.
+This adds significant complexity to the implementation, and negates a good deal of the benefit of the tokens being _native_: tokens locked by a script are not naturally "in" a user's wallet, nor can they be locked with _other_ scripts, which prevents using them in other "smart contracts".
+
+### Examples
+
+#### Blacklisting addresses
+
+A common requirement for tokens that are regulated externally by governmental authorities (e.g. USDC) is to be able to blacklist addresses, preventing them from taking certain actions with their tokens (sometimes spending them, sometimes receiving them).
+
+This has two challenges:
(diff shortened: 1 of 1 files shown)


rphair (commented):
I understand this proposal at a high level but would want @KtorZ and/or @SebastienGllmt to agree that it should be put on record with these details.

@michaelpj I also think it's appropriate to have a `Rejected` status and have argued for this in https://github.com/cardano-foundation/CIPs/pull/331#issuecomment-1258047610... expecting your earlier conversation there will be resolved in favour of adding the `Rejected` status back into the CIP workflow 😎

michaelpj:
Yeah, this can have whatever the "not gonna happen" status is, when you decide on one :)

CIP-0070/README.md
```diff
@@ -0,0 +1,135 @@
+---
+CIP: 70
+Title: Spending policies
+Authors: Michael Peyton Jones <michael.peyton-jones@iohk.io>
+Status: Rejected 
+Type: Standards Track 
+Created: 23/09/2022
```
KtorZ, line 7:
```suggestion
Created: 2022-09-23
```

ISO-8601 please 🙏

michaelpj:
This should probably actually become a CPS.

zygomeb:
@michaelpj 

I've recently discussed that this could be implemented in a way that I don't believe was discussed before -- that we can modify the event that triggers a spending policy to only happen when we move tokens between two different spending addresses (no invocation on any other change). 

I believe that would be sufficient to create tokens that 
1) blacklist addresses 
2) have custom fees 
3) have claw back (more annoying to implement but possible) 

This would also allow us to safely use them in wallets as at any time the user may isolate the token to a 'garbage' utxo.

michaelpj:
> I've recently discussed that this could be implemented in a way that I don't believe was discussed before -- that we can modify the event that triggers a spending policy to only happen when we move tokens between two different spending addresses (no invocation on any other change).

1. Can you be more specific about the triggering condition? Bear in mind that many tokens are fungible, so there is not necessarily a clear answer about which tokens moved where.
2. Are you suggesting that spending policies would _only_ trigger in this context, or that they can somehow _choose_ not to trigger if the transfer is to the same address?
3. I don't think this fundamentally makes things easier: wallets still become largely unable to reliably transfer tokens to other people, which is one of the main functionalities of a wallet!

zygomeb:
> 1. Can you be more specific about the triggering condition? Bear in mind that many tokens are fungible, so there is not necessarily a clear answer about which tokens moved where.

On: transfer of token from a locking/spending address A to a spending address B. A change of stake key and change of datum (and any other utxo-carried information) does not trigger the Spending policy to run under this definition. 

> 2. Are you suggesting that spending policies would _only_ trigger in this context, or that they can somehow _choose_ not to trigger if the transfer is to the same address?

Only in this context. Choosing to not act IMO should be done inside the script. 

> 3. I don't think this fundamentally makes things easier: wallets still become largely unable to reliably transfer tokens to other people, which is one of the main functionalities of a wallet!

Yes, but that case is now largely silo'd out into outgoing transfers which may now be isolated and understood on a case by case basis. The feature is a burden for wallets but now we can build an off-chain knowledge base to be able to construct transactions in a way that this spending policy would find acceptable.

KtorZ:
Flagging this as "Waiting for author" from @michaelpj's earlier comment: 

> This should probably actually become a CPS.

Seems like a good idea to me indeed; we've taken quite some time processing that proposal and although we agreed with the general idea, CPS didn't exist at the time and could now be a better fit indeed. I am fine either way: merging as an inactive CIP or turning this to a CPS.

rphair:
Note @zygomeb @fallen-icarus (as you have shown some interested) that this indeed should have been a CPS... unfortunately we were not fortunate enough, as the CPS was evolving, to have the benefit of Michael's understanding in rewriting this as a CPS.

We have to close this now because editors don't have the skills to rephrase it as a CPS, but someone in the community might take that over: likely as a new PR since the authorship would be different.  Editors are standing by to cross-reference the old & new discussions if that ever happens.  cc @Ryun1 @Crypto2099

PR #337, opened 2022-09-26
changed to reflect babbage

kevinhammond:


kevinhammond:
This CIP reflects the changes in Babbage to define a Plutus V2 cost model

rphair (approved):

KtorZ:
This isn't the only change as far as I recall. There is also:

- coinsPerUTxOByte replacing lovelacePerUTxOWord
- removal of `extraEntropy` 
- removal of `decentralizationParameter`

(and I think that's all). 

In addition, we have created separate document by the past to cover protocol parameters (as [CIP-0009](https://github.com/cardano-foundation/CIPs/tree/master/CIP-0009) and [CIP-0028](https://github.com/cardano-foundation/CIPs/tree/master/CIP-0009)), and, @JaredCorduan did create an extension a while back for Babbage as [CIP-0055](https://github.com/cardano-foundation/CIPs/tree/master/CIP-0055) only covering the coinsPerUTxOByte change however. 

So I'd suggest to make this addition in CIP-0055 instead, and also mention the removal of the two params (extraEntropy + decentralizationParameter) compared to Alonzo with a line of rationale for each.

KtorZ:
@JaredCorduan whoops indeed, I forgot that the removal of `d` and `extraEntropy` was added to CIP-0055 in the end. So I guess, if we want to add a mention to the new cost model in there (CIP-0055), that's fine but that's a bit out of topic for CIP-0028.

KtorZ (changes requested):
This should be added to CIP-0055, not CIP-0028

rphair:
@kevinhammond what @KtorZ says above looks right: the later CIP (@JaredCorduan https://github.com/cardano-foundation/CIPs/pull/265) has a very small list of changed parameters, which would be incomplete without this added list... as likewise this longer list is likely to get overlooked if it stands on its own.  Can you refactor your table into https://github.com/cardano-foundation/CIPs/tree/master/CIP-0055 instead - or have any objection?

kevinhammond:
I think it's better to merge CIP-0055 into this one.  It's not really standalone as far as I can see and we don't want to proliferate the parameter CIPs.

kevinhammond:
I have merged in CIP-55 @JaredCorduan and @KtorZ

KtorZ:
@kevinhammond , for the sake of clarity, as stated earlier, we want the opposite; this addition to be merged into CIP-0055 and not the other way around. It's best to have one proposal per era since we started doing this. There's one for Shelley, one for Alonzo and CIP-0055 is the one partially covering Babbage.

kevinhammond:
Unlike Alonzo, Babbage doesn't really have its own set of parameters, @KtorZ

KtorZ:
What do you mean? It surely does! While there's no new genesis configuration for Babbage, it still defines some new parameters and removes some existing ones. The goal of CIP-0055 was to highlight those changes. For PV8, I'd suggest to re-use a similar approach to what was done before with a Changelog in the `Path to Active` section (e.g. https://github.com/cardano-foundation/CIPs/tree/master/CIP-0009#change-log)

PR #338, opened 2022-09-27
Update top-level README tables & status for a few now-active proposals.

KtorZ:


rphair (approved):
good to put those changes into the documents themselves, especially the ones activated with the Vasil HF 😎

rphair:
(oops, didn't double check the table itself before review... hence the last commit) 😖

KtorZ:
Ah! I had actually updated that table but somehow committed without saving 🤦. Well, I am glad we are two pairs of eyes on that!

PR #340, opened 2022-09-30
Create Meeting Notes 54

AntMeyer1:


rphair:
@AntMeyer1 just please make a note when all your commits are in, so we can start proofreading without duplication of effort 🙏

rphair:
p.s. also in general I think this semantic would work: when happy with the document yourself (and all commits are pushed), just tag us in a review request 😎

rphair (approved):
Judging by time that's gone by, the last question is answered by default; though generally it should be considered essential to declare when a document is ready for review when the last posting in the PR is more than one commit in succession (otherwise how would reviewers know when the author is really done?)

I corrected mostly just the proper names in last commit (not reviewing line by line this time) but there are so many little problems it would take forever to fix them all.  Probably people will get enough of the gist without trying to make these documents perfect each time: so I'm letting that go.

Machine translations still need to be closely supervised.  It cannot be completely blamed on @KtorZ's accent that his use of the word _standard_ was transcribed as "*standout*" _every single time_.  This was never an issue before... has their AI somehow been trained to transcribe the _wrong_ thing? 🤪

Another idea might be to "crowdsource" the minutes by posting each meeting section link (once merged) to the PR it relates to: knowing the biweekly review generally _hasn't_ caught every error.  That way if any PR author feels like something sloppy, incorrect or misleading has been recorded for their PR (especially if authors themselves were present) they can point it out & submit another corrective PR:

1. I don't think that would happen often enough to really increase the workflow on our part.
2. This would have the additional benefit of linking easily back to the minutes so we & the community can reference the discussions more easily.

This last suggestion would have been a good thing to mention at last month's process meeting but it still hadn't sunk in back then how important this linkage might be.... and it certainly would avoid us having to spend more time correcting the minutes than attending the meeting itself. 🙏

cc @SebastienGllmt @crptmppt

KtorZ:
@rphair I haven't reviewed the transcript yet but looking at your comment I am hesitant to 😅 ... 
Maybe the transcripts do not have that much of an added value in the end -- or at least, the ratio of time spent on them versus how useful they are may be unbalanced. 

Since we do record calls and upload them on youtube for a few months now, I'd be inclined to just stop with the full transcript altogether; those interested in the full detail can watch the recording (with automatically generated subtitles). 

What I think would be more valuable are short summaries of the key decisions taken during a meeting. This is mostly the only reason why I go back to meeting notes from time to time; to remember what we said exactly on this or that CIP. I tend myself not to have time to take proper notes during the meetings because I am often the one talking.

rphair:
yes @KtorZ I think that would be a good format: each item that appears on the agenda, with the resolution(s) listed after each item.

@AntMeyer1 the transcription service will still help (if _not_ posted to the public, to check those resolutions for the published document & ideally post to each GitHub thread where they will do the most good for authors, other reviewers, users & advocates.  The same time spent editing the full, publicly posted minutes would be better applied that way.

The header at the beginning which lists attendees is still important because we have 2 funded Catalyst projects now which use meeting attendance as a criteria for success.  Also, showing which guests have attended is a small acknowledgement which could be valuable as we try to get more people to participate & eventually recruit other editors.

AntMeyer1:
test

rphair:
This PR is not really useful anymore except to illustrate why we stopped machine-transcribing the CIP meeting minutes. 🤪

PR #346, opened 2022-10-05
CIP-0036 | Changed terminology in schema to match README

Ryun1:
Removed the notion of "proportion" from the schema replacing it with the term "weight" since "weight" is used within the specification. cc: @kevinhammond


Ryun1:
Used the incorrect branch

PR #347, opened 2022-10-05
CIP-0036 | Changed terminology in schema to match README specification

Ryun1:
Removed the notion of "proportion" from the schema replacing it with the term "weight" since "weight" is used within the specification. cc: @kevinhammond


kevinhammond:
Thanks, Ryan - looks good!

rphair (approved):

gitmachtl:
can we add the deregistration scenario (empty delegation array) also in this PR?

PR #348, opened 2022-10-05
CIP-0036 | Added deregistration functionality

gitmachtl:

PR #349, opened 2022-10-08
CIP-0036 | Adding the deregistration metadata format

gitmachtl:
- Adding the deregistration metadata format to the README.md and the schema.cddl
- The deregistration metadata uses the key 61286 and is independent from the registration metadata which uses key 61284
- Updating CIP-0010 (metadata datum registry), description for key 61285(witness) and 61286(deregistration)


rphair:
please everybody see, for prior discussion:

- https://github.com/cardano-foundation/CIPs/pull/348#issuecomment-1268721923
- https://github.com/cardano-foundation/CIPs/pull/335#issuecomment-1263991634

gitmachtl:
so, this should now be the final PR hopefully 😄 
we had a discussion, and @kevinhammond agreed, that the until now unused but reserved key 61286 could be used to add a deregistration metadata format. this is independent from the registration metadata format and that part of CIP-0036 is untouched. the specification is currently only for the voting_purpose=0 (Catalyst). other voting_purpose values (voting-chains) may use a different way to specify a deregistration. the goal was that to always produce valid registration metadata and not abuse the use of an invalid registration metadata to act like a deregistration. this addition in the PR should now give a clear format if users/tools want to do a de-registration of the stake credentials on the catalyst voting chain.

kevinhammond:
> we had a discussion, and @kevinhammond agreed, that the until now unused but reserved key 61286 could be used to add a deregistration metadata format. 

Yes

> the specification is currently only for the voting_purpose=0 (Catalyst). other voting_purpose values (voting-chains) may use a different way to specify a deregistration.

Technically, this is specific to Catalyst voting on the corresponding Jormungandr voting chains (it is possible that purpose 0 might also be collated and used by a third party, for example, as a generic/non-specific vote delegation).

kevinhammond:
The main reason for not wanting to use eg an empty delegation array or second metadata format for deregistration purposes is that this creates complication for hardware wallets (perhaps two apps are needed).

gitmachtl:
There is no need imo that hardware wallets have to support all features, thats why i wrote the line:
https://github.com/cardano-foundation/CIPs/blame/6e2cf66230bf662b04065ec75098f27695045fcc/CIP-0036/README.md#L154

I don't see a problem if deregistraton would not be supported on hardware wallets.

SebastienGllmt (approved):
LGTM

rphair (approved):
That makes sense 👍 @gitmachtl are you ready for this to be merged?

rphair:
OK, this all seems in order then 😎

PR #351, opened 2022-10-13
CIP-0071? | Non-Fungible Token (NFT) Proxy Voting Standard

thaddeusdiamond:
This is the initial commit for a proposed Cardano Improvement Proposal (CIP) that would enable NFT projects that did not want to use a governance or other native fungible asset to perform verifiable on-chain votes.

[ Documentation: README.md, example/ ]

---

[see rendered Markdown](https://github.com/thaddeusdiamond/CIPs/blob/cip-71/CIP-0071/README.md)

### DIFF
--- a/CIP-0071/README.md
+++ b/CIP-0071/README.md
@@ -12,7 +12,7 @@ License: CC-BY-4.0
 
 ## Abstract
 
-This proposal defines a standard mechanism for performing verifiable on-chain voting in NFT projects that do not want a governance token using inline datums, plutus minting policies, and smart contracts.
+This proposal defines a standard mechanism for performing verifiable on-chain voting in NFT projects that do not have a governance token by using datums, plutus minting policies, and smart contracts.
 
 ## Motivation
 
@@ -29,22 +29,22 @@ It is not intended for complex voting applications or for governance against fun
 
 ### A Simple Analogy
 
-The basic analogy here is that of a traditional state or federal vote.  Imagine a citizen who has a state ID (e.g., Driver's License) and wants to vote.  That citizen would:
-1. Go to a precinct and show their ID to the appropriate authority
-2. Receive a ballot with choices for the current vote
-3. Mark their selections on the ballot
-4. Sign their ballot with their name
-5. Submit their ballot into a single "ballot box"
-6. Authorized vote counting authorities process the vote after polls close
-7. Await the election results through a trusted news outlet
+The basic analogy here is that of a traditional state or federal vote.  Imagine a citizen who has a state ID (e.g., Driver's License) and wants to vote, as well as a central voting authority that counts all the ballots.
+1. Citizens go to to a precinct and show their ID to the appropriate authority
+2. Citizens receive a ballot with choices for the current vote
+3. Citizens mark their selections on the ballot
+4. Citizens sign their ballot with their name
+5. Citizens submit their ballot into a single "ballot box"
+6. Central voting authorities process the vote after polls close
+7. Citizens await the election results through a trusted news outlet
 
 This specification follows the same process, but using tokens:
 1. A holder of a project validates their NFT by sending it to self
-2. The user signs a Plutus minting policy to create a proxy NFT that represents the current ballot
-3. The holder marks their vote selections off-chain
-4. The holder signs a new transaction that sends the proxy "ballot" NFT to a smart contract ("the ballot box") with a datum representing the vote
-5. Authorized vote counting wallets process the UTxOs' datum in the "ballot box" smart contract after the polls close
-6. Project creators report the results in a human-readable off-chain format to their holders.
+2. The holder signs a Plutus minting policy to create a "ballot" NFT linked to their unique NFT
+3. The holder marks their desired vote selections off-chain
+4. The holder signs a tx that sends the "ballot" NFT to a "ballot box" (smart contract) with their "vote" (datum)
+5. Authorized vote counting wallets process UTxOs and their datums in the "ballot box" smart contract after polls close
+6. Authorized vote counters report the results in a human-readable off-chain format to holders
 
 Because of the efficient UTxO model Cardano employs, steps #1, #2, and #4 can occur in a single transaction.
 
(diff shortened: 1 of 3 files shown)


thaddeusdiamond:
This is not ready for review yet I am just putting it up here so I have a pointer to get the process started.  When I am ready for formal reviews, I will move it out of Draft status.

thaddeusdiamond:
This is now ready for community review.  I will engage key participants to determine how best to get the ball rolling!

rphair:
thanks @thaddeusdiamond - I've added a direct link to the README.md file in your branch in the form we use when new CIP PR's are submitted... please modify the text as you like & keep the link updated if your CIP draft changes path (I see your "example" is linked from the CIP text).

If you can indicate here which individuals or groups you are already addressing as "key participants" we can ask for feedback from any others who have submitted or reviewed proposals about voting, governance, and/or authentication by NFT.

The proposal looks technically sound to me, and complete with all formalities including the Path to Active that we would expect to eventually merge this with `Proposed` status,  but I'm not qualified to review the technical details: therefore hoping that @KtorZ @SebastienGllmt and perhaps @michaelpj will be interested in reading & reviewing soon.

CIP-0071/README.md
```diff
@@ -0,0 +1,246 @@
+---
+CIP: 71
+Title: Non-Fungible Token (NFT) Proxy Voting Standard
+Authors: Thaddeus Diamond <support@wildtangz.com>
+Comments-URI:
+Status: Proposed
+Type: Informational
```
rphair, line 7: I am just curious why you chose to label this as `Informational` (rather than `Process`) because what's described here seems to depend on a fair bit of interoperability... from [latest CIP-0001 draft](https://github.com/cardano-foundation/CIPs/blob/cip-0001-rework/CIP-0001/README.md):

Type          | Description
--- | ---
Process       | A process defines a common standard to follow to achieve a certain degree of interoperability between various actors. It can be a formal document describing a specific procedure or a set of recommendations/guidelines on one particular topic.
Informational | An informational CIP documents a particular design decision. These are often written a posteriori to address a gap in understanding and knowledge sharing.

CIP-0071/README.md
```diff
@@ -0,0 +1,246 @@
...
+        vote: voteResult.vote,
+        count: voteCount
+      }
+    }
+  }
+  return votesByAsset;
+}
+```
+
+There is no requirement that the "ballot counter" redeem all "ballots" from the "ballot box" and send them back to the respective voters, but we anticipate that this is what will happen in practice.  We encourage further open-sourced code versions that enforce this requirement at the smart contract level.
```
yHSJ, line 160: I believe it makes more sense to burn the "ballots" instead of returning them to their respective voters. There are a few reasons for this:
First, in the motivation section, you discuss the issues with governance tokens. Although the ballots are no longer usable (defined by the time in the policy), it seems unnecessarily confusing for users to receive what seems like a governance token. That could result in confusion and even scams. 
Secondly, burning the asset will make it much easier for third parties to tabulate votes or at on the results. While all the data is provided to tabulate votes as a third party already, it is quite simple, and already done for other applications, to track the -1 mint transactions of all assets under a policy ID. Any on chain voting should allow vote results to be verified as simply as possible, which would be done through burning the assets.

Of course, once you burn the asset, there is the minUTXO that was previously attached to the asset. It is important to do SOMETHING with that minUTXO, so we don't end up with up to ~10.5K ADA locked in the contract address. I propose that it is all sent back to the voters, so the only cost associated with voting is the transaction fee, but I would also support splitting it with the developer(s) in some capacity to pay for their time.

CIP-0071/README.md
thaddeusdiamond: Great question! I honestly did not know about this categorization I was just copying from prior CIPs.  Will update to be "Process".

CIP-0071/README.md
thaddeusdiamond: I think the ideal here is to have the default ballot box perform:

* Burn ballot
* Send minUTxO back to address indicated by datum
* Only do the above after verifying on chain that polls are closed

I would still like to leave this as optional if possible, though it will definitely be an initial implementation in the [voting.js](https://github.com/thaddeusdiamond/cardano-nft-mint-frontend/blob/master/src/js/pages/voting.js) file in cardano-nft-mint-frontend.

thaddeusdiamond:
> thanks @thaddeusdiamond - I've added a direct link in your OP to the README.md file in your branch, in the form we use when new CIP PR's are submitted... please modify the text as you like & keep the link updated if your CIP draft changes path (I see your "example" is linked from the CIP text).
> 

What is OP? I can update any forms you need I just was not sure of them.

> If you can indicate here which individuals or groups you are already addressing as "key participants" we can ask for feedback from any others who have submitted or reviewed proposals about voting, governance, and/or authentication by NFT.
> 

I have talked to several NFT projects that want this kind of functionality (especially for newer/smaller projects), but they may not be technical.  I've reached out to encourage them to comment here.

> The proposal looks technically sound to me, and complete with all formalities including the Path to Active that we would expect to eventually merge this with `Proposed` status, but I'm not qualified to review the technical details: therefore hoping that @KtorZ @SebastienGllmt and perhaps @michaelpj will be interested in reading & reviewing soon.

rphair:
it's a common Internet abbreviation for Original Post, i.e. your [comment here](https://github.com/cardano-foundation/CIPs/pull/351#issue-1407160556).  When new proposals come in I generally link them at the bottom to the working proposal to make it easy for me & others to review, especially when on the spot at our biweekly CIP meetings:
> [see rendered Markdown](https://github.com/thaddeusdiamond/CIPs/blob/cip-71/CIP-0071/README.md)

Since you already mentioned the `README.md` in your own text I also thought you might want to merge these.

CIP-0071/README.md
yHSJ: > I would still like to leave this as optional if possible, though it will definitely be an initial implementation in the [voting.js](https://github.com/thaddeusdiamond/cardano-nft-mint-frontend/blob/master/src/js/pages/voting.js) file in cardano-nft-mint-frontend.

I disagree with this. I think that if there is going to be a standard, it should be just that -- standard. A 3rd party should not have to change their tabulation system in order to verify votes. Ballots should always be burned once the polls have closed. If a voting organizer wants to give out a commemorative voting NFT (think the "I voted" sticker), they should then mint assets on a new policy ID and send those to the voters.

Leaving it up to choice in this situation seems like the worst choice, it removes the benefits of standardizing this.

thaddeusdiamond: As we discussed in the chat, because the minting of the "ballot" HAS TO occur in the same transaction as the "send-to-ballot-box" to prevent a "mint now, vote later" attack, I think this is less of a concern.  Monitoring for a burn transaction would actually duplicate the cost to viewers who would have to find the burn transaction per the standard.

yHSJ: Upon further discussion with @thaddeusdiamond in a Discord server (and re-reading the proposal), I now tend to agree that the burning shouldn't be required, at least not for the sake of a developer. I still do believe that burning the token makes the most sense (release the minUTXO, prevent confusion on the side of users), but it certainly isn't mandatory.

rodrigomd94: I am not sure about it being optional or not but I do agree that burning the token makes the most sense instead of sending it back to the voter.

CIP-0071/README.md
```diff
@@ -0,0 +1,246 @@
...
+- Airdrops of governance tokens require minUTxO ADA attached, costing the NFT project ADA out of pocket
+- Fungible tokens do not have a 1:1 mechanism for tracking votes against specific assets with voting power
+- Sale of the underlying NFT is not tied to sale of the governance token, creating separate asset classes and leaving voting power potentially with those who are no longer holders of the NFT
+
+This standard provides a simple solution for this by using the underlying NFT to mint a one-time "ballot" token that can be used only for a specific voting topic. Projects that adopt this standard will be able to integrate with web viewers that track projects' governance votes and will also receive the benefits of verifiable on-chain voting without requiring issuance of a new native token.
+
+It is not intended for complex voting applications or for governance against fungible tokens that cannot be labeled individually.
```
KtorZ, line 26: I'd love to see perhaps a concrete use case unfold in this section. Do you have a particular project as an example?

CIP-0071/README.md
```diff
@@ -0,0 +1,246 @@
...
+
+```ts
+func main(redeemer: Redeemer, ctx: ScriptContext) -> Bool {
+  tx: Tx = ctx.tx;
+  minted_policy: MintingPolicyHash = ctx.get_current_minting_policy_hash();
+  redeemer.switch {
+    Mint => {
+      polls_are_still_open(tx.time_range)
+        && assets_were_spent(tx.minted, minted_policy, tx.outputs)
+        && assets_locked_in_script(tx, tx.minted)
```
KtorZ, line 80: Isn't the minting policy expected to check that the wallet has at least one NFT of the required policy in order to mint a ballot?

CIP-0071/README.md
```diff
@@ -0,0 +1,246 @@
...
+    },
+    // Burn code elided for space...
+  }
+}
+```
+
+#### "Vote" -> eUTxO Datum
+
+To cast the vote, the user sends the ballot NFT just created to a "ballot box".  Note that for reasons specified in [the "attacks" section below](#Creation-of-Ballot-Without-Casting-a-Vote) this needs to occur during the same transaction that the ballot was minted in.
```
KtorZ, line 89: What enforces that? The minting policy?

CIP-0071/README.md
```diff
@@ -0,0 +1,246 @@
...
+The datum is a simple object representing the voter who cast the vote and the vote itself:
+
+```ts
+type VoteDatum = {
+  voter: PubKeyHash,
+  vote: object
+};
+```
+
+The `voter` element is extremely important in this datum so that we know who minted the ballot NFT and who we should return it to.  At the end of the ballot counting process, this user will receive their ballot NFT back.
```
KtorZ, line 100: Why not burn the ballot?

CIP-0071/README.md
```diff
@@ -0,0 +1,246 @@
...
+### The Vote Counting Process
+
+#### "Ballot Counter" -> Authorized Wallet
+
+Given the flexible nature of the ["ballot box" smart contract](#Ballot-Box---Smart-Contract) enumerated above, we propose a simple algorithm for counting votes and returning the ballots to the user:
+
+1. Ensure the polls are closed (can be either on or off-chain)
+2. Iterate through all UTxOs in forward-time-order locked in the "ballot box" and for each:
+	* Determine which assets are inside of that UTxO
+	* Mark their most recent vote to match the `vote` object in the UTxOs datum
```
KtorZ, line 132: I think this last point answer a question I was about to ask about multiple votes. What happens when a voter mints multiple ballots and votes multiple time? This suggests only the most recent vote is counted.

CIP-0071/README.md
KtorZ: > I believe it makes more sense to burn the "ballots" instead of returning them to their respective voters. 

Yes. 

>  I think that if there is going to be a standard, it should be just that -- standard.

Also.

> Upon further discussion with @thaddeusdiamond in a Discord server (and re-reading the proposal)

I'd love to see the outcome of that discussion in the 'Rationale'. In particular, the rationale for burning/not burning token needs to be stated clearly as this is clearly a point everyone seems to be tripping on.

CIP-0071/README.md
```diff
@@ -0,0 +1,246 @@
...
+
+### Returning the "Ballot" NFTs to the Wrong User
+
+During the construction of the ballot NFTs we allow the user to specify their vote alongside a `voter` field indicating where their "ballot" NFT should be returned to once the vote is fully counted.  Unfortunately, this is not strictly checked inside the Plutus minting policy's code (largely due to CPU/memory constraints).  So, we rely on the user to provide an accurate return address, which means that there is the potential for someone who has not actually voted to receive a commemorative NFT.  This does not impact the protocol though, as the "ballot" NFT was legally minted, just returned to the incorrect location.  That user actually received a gift, as they can now burn the ballot and receive some small amount of dust.
```
KtorZ, line 206: > Unfortunately, this is not strictly checked inside the Plutus minting policy's code (largely due to CPU/memory constraints).

I am not sure that this is a valid justification. Nor do I understand actually how one would check the voter field? Surely, there isn't much that the minting policy can do besides verifying that it's indeed a well-formed public key.

CIP-0071/README.md
```diff
@@ -0,0 +1,246 @@
...
+        }
+      })
+    }
+  }
+}
+```
+
+The Helios code above simply checks that during a burn (as indicated by the Plutus minting policy's `redeemer`), the user is not attempting to mint a positive number of any assets.  With this code, *any Cardano wallet* can burn *any ballot* minted as part of this protocol.  Why so permissive? We want to ensure that each vote is bringing the minimal costs possible to the user.  In providing this native burning mechanism we can free up the minUTxO that had been locked with the ballot, and enable the user to potentially participate in more votes they might not have otherwise.  In addition, users who really do not like the specific commemorative NFTs or projects that choose to skip the "commemorative" aspect of ballot creation now have an easy way to dispose of "junk" assets.
+
+## Potential Attacks and Mitigations
```
KtorZ, line 190: Please nest this under a `Rationale` section.

CIP-0071/README.md
```diff
@@ -0,0 +1,246 @@
...
+- The "ballot" NFT should have some sort of unique metadata (if using [CIP-25](https://github.com/cardano-foundation/CIPs/tree/master/CIP-0025)), datum (if using [CIP-68](https://github.com/cardano-foundation/CIPs/tree/master/CIP-0068)) or other identification so that the users can engage with the ballot in a fun, exciting way and to ensure there is no confusion with the reference NFT.
+- The "vote" represented by a datum will be easier to debug and analyze in real-time if it uses the new "inline datum" feature from Vasil, but the protocol will still work on Alonzo era transactions.
+- The "ballot box" smart contract should likely enforce that the datum's "voter" field is respected when returning the ballots to users after voting has ended to provide greater transparency and trust for project participants.
```
KtorZ, line 224: One thing I am missing from the justifications here is why this solution addresses the problem stated in the motivation. As far as I understood, one of the core motivation was the cost of minting governance tokens (min ada value), but this proposal also proposes to mint ballot NFT and only, _optionally_ return the ballot NFT to the voter. So arguably, the proposal only _optionally_ addresses the motivation. Since this is stated as a motivation, I think the proposal should better describes how it solves this issue systematically.

KtorZ (commented):
Overall, the idea looks okay but I am missing a bit some perspective with existing similar voting systems such as SundaeSwap's (cc @Quantumplation) or Voteaire's (cc @codefly) from whom I'd appreciate their review / comments on this proposal. 

I am also not fully sure to understand the need for a Ballot NFT and counting ballot contract altogether. Indeed, the counting seems to happen entirely off-chain, by fetching and ordering UTxO entries, and counting assets within them. The contract merely states the rule by which the ballot can be returned to their original voters. But, why bother minting and sending ballots in the first place then? 

What is the added benefit compared to a more simple vote done by sending a transaction to oneself with the associated metadata? An off-chain process can easily fetch transactions with the appropriate metadata tag and count votes in a very similar fashion. If there's a clear benefit over doing that, it should be stated in the motivation (what problem this approach has) and the rationale (how this specification solves issues of that simple approach);

CIP-0071/README.md
thaddeusdiamond: It's mostly about the use case around commemorative NFTs.  Will add a Rationale section with our offline discussion.

CIP-0071/README.md
thaddeusdiamond: Added 3.  I can name names or leave it generic just let me know which is preferred in CIPs.

CIP-0071/README.md
thaddeusdiamond: yes that is `assets_were_spent` let me rename the pseudocode better

CIP-0071/README.md
thaddeusdiamond: Yes that is the `assets_locked_in_script` function above (will rename to `assets_locked_in_ballot_box`).

CIP-0071/README.md
thaddeusdiamond: See the new Rationale section around commemorative NFTs.

CIP-0071/README.md
thaddeusdiamond: Correct.  Only the most recent vote is counted, although all votes are technically "valid". Basically we are allowing voters to change their mind.

CIP-0071/README.md
thaddeusdiamond: It's fair, we can remove it (or make the standard be that the user's return address is checked).  Basically you would have to check that the `datum.vote.address` field matches the pubkeyhash of one of the signatories of ballot creation.  Should we change it to that?

CIP-0071/README.md
thaddeusdiamond: Done.

CIP-0071/README.md
thaddeusdiamond: That's fair.  I think we can then enforce that the ballot has to be either:

a) Returned to the user
b) Burned and the ADA unlocked returned to the user

Thoughts?

thaddeusdiamond:
Thank you so much for the review.  A couple of your comments I have follow-up questions on.  I am also pushing a new revision that hopefully addresses some of your thoughts.  More feedback welcome (especially @codefly and @Quantumplation seeing as they have governance experience.

> Overall, the idea looks okay but I am missing a bit some perspective with existing similar voting systems such as SundaeSwap's (cc @Quantumplation) or Voteaire's (cc @codefly) from whom I'd appreciate their review / comments on this proposal.
> 

I am not familiar with SundaeSwap but Voteaire uses metadata (correct me if I am wrong @codefly I am going off of https://github.com/voteaire/voteaire-onchain-spec).  I would like to move the procedure on-chain to allow for on-chain verification of vote authenticity (e.g., enforcing RCV/vote selection in a Smart Contract).

> I am also not fully sure to understand the need for a Ballot NFT and counting ballot contract altogether. Indeed, the counting seems to happen entirely off-chain, by fetching and ordering UTxO entries, and counting assets within them. The contract merely states the rule by which the ballot can be returned to their original voters. But, why bother minting and sending ballots in the first place then?
> 

The creation of the ballot and simultaneously sending in the vote is the entire process.  Counting can occur off-chain (just like NFT viewers retrieve IPFS images off-chain), but the validity of the votes are enforced on-chain.  I am going to add a few use cases in the "Motivation" section that hopefully clear up why you need to mint the ballots on-chain.

> What is the added benefit compared to a more simple vote done by sending a transaction to oneself with the associated metadata?

For dApps, it is much simpler (and more cost effective against service providers) to look for UTxOs at a single address (the ballot box).  Otherwise we would have to snapshot the addresses who held and look at all their txn's during a voting period to find the specific vote.  We also could not enforce vote authenticity on-chain.

> An off-chain process can easily fetch transactions with the appropriate metadata tag and count votes in a very similar fashion. If there's a clear benefit over doing that, it should be stated in the motivation (what problem this approach has) and the rationale (how this specification solves issues of that simple approach);

I will update the motivation and rationale for why this needs to be on-chain.

victormundi:
Thank you @thaddeusdiamond for writing up this proposal. We would certainly be adopting and using this standard at [ApeWatch](https://apewatch.app/) in order to let holders of our NFT - to be released - vote on various topics, e.g. which feature to build next.

CIP-0071/README.md
KtorZ: I'd have a slight preference for "Burned and ADA returned to the user" because most users wouldn't probably know how to do that in the first place. Plus, it makes the protocol complete this way by cleaning up after oneself.

KtorZ (approved):
As outlined in the editors meeting today, we're happy to see this proposal merged as _Proposed_ while implementations are being worked on. 

Please come back with updates on the proposal (might it simply be changing the status to 'Active') as the implementation progresses.

CIP-0071/README.md
```diff
@@ -0,0 +1,268 @@
...
+---
+
+## Abstract
+
+This proposal uses plutus minting policies to create valid "ballots" that are sent alongside datum "votes" to a centralized smart contract "ballot box" in order to perform verifiable on-chain voting in NFT projects that do not have a governance token.
+
+## Motivation
+
+This proposal is intended to provide a standard mechanism for non-fungible token (NFT) projects to perform on-chain verifiable votes using only their NFT assets. There are several proposed solutions for governance that involve using either a service provider (e.g., Summon) with native assets or the issuance of proprietary native assets.  However, there are several issues with these approaches:
+- Airdrops of governance tokens require minUTxO ADA attached, costing the NFT project ADA out of pocket
```
rphair, line 20:
```suggestion
- Airdrops of governance tokens require minUTxO ada attached, costing the NFT project ada out of pocket
```

CIP-0071/README.md
```diff
@@ -0,0 +1,268 @@
...
+      }
+    }
+  }
+  return votesByAsset;
+}
+```
+
+There is no requirement that the "ballot counter" redeem all "ballots" from the "ballot box" and send them back to the respective voters, but we anticipate that this is what will happen in practice.  We encourage further open-sourced code versions that enforce this requirement at the smart contract level.
+
+### Reclaiming ADA Locked by the Ballot NFTs
```
rphair, line 173:
```suggestion
### Reclaiming Ada Locked by the Ballot NFTs
```

CIP-0071/README.md
```diff
@@ -0,0 +1,268 @@
...
+
+Even if the ballot NFT is returned to the user, this will leave users with ADA locked alongside these newly created assets, which can impose a financial hardship for certain project users.
```
rphair, line 175:
```suggestion
Even if the ballot NFT is returned to the user, this will leave users with ada locked alongside these newly created assets, which can impose a financial hardship for certain project users.
```

CIP-0071/README.md
```diff
@@ -0,0 +1,268 @@
...
+
+### Commemorative NFTs with Optional Token Burning
+
+There is a question as to whether we should enforce the requirement that votes be burned when they are counted by the vote counter.  However, we do not want that to be a standard as many users of NFT communities have expressed an interest in receiving commemorative NFTs (similar to an "I Voted" sticker).  Instead, we propose that the ballot Plutus minting policy be burn-able by anyone who holds the NFT in their wallet.  This way, locked ADA can be reclaimed if the user has no further use for the commemorative NFT (see an example of this in the [Implementation](./example/)).
```
rphair, line 209:
```suggestion
There is a question as to whether we should enforce the requirement that votes be burned when they are counted by the vote counter.  However, we do not want that to be a standard as many users of NFT communities have expressed an interest in receiving commemorative NFTs (similar to an "I Voted" sticker).  Instead, we propose that the ballot Plutus minting policy be burn-able by anyone who holds the NFT in their wallet.  This way, locked ada can be reclaimed if the user has no further use for the commemorative NFT (see an example of this in the [Implementation](./example/)).
```

CIP-0071/example/voting.html
```diff
@@ -0,0 +1,149 @@
...
+          </div>
+      </div>
+
+      <div id="vote-count-output" class="row padded">
+          Vote results will appear here...
+      </div>
+
+      <div id="vote-burn-container" class="row padded">
+          <div class="col-12 centered">
+            <h4>Done With Your Ballots? Want to Reclaim Your ADA?</h4>
```
rphair, line 73:
```suggestion
            <h4>Done With Your Ballots? Want to Reclaim Your Ada?</h4>
```

rphair (commented):
I only noticed this from a recent review comment but there are several misspellings of _ada_, using instead the trading symbol _ADA_ (according to Cardano Foundation long standing documentation, a style error):

thaddeusdiamond:
Thank you @rphair.  I have updated the CIP to to have the appropriate spelling for ADA -> ada.  Can you advise on how to complete path to Active?  So far I have:

- Considerations for ranked-choice voting if projects wish to have it
- Minimal reference implementation making use of [Lucid](https://github.com/spacebudz/lucid) (off-chain), [Plutus Core](https://github.com/input-output-hk/plutus) [using Helios](https://github.com/Hyperion-BT/Helios) (on-chain): [Implementation](./example/)
- Open-source implementations from other NFT projects that make use of this CIP

1. RCV is listed in this README as an add-on that can be verified through the datum
2. Reference implementation is checked in here
3. I open-sourced the full example (which we conducted and completed at wildtangz.com/vote) as part of the [cardano-nft-js-toolkit](https://github.com/thaddeusdiamond/cardano-nft-mint-frontend) repository I have.

Should I add those references somehow?

rphair (approved):
@thaddeusdiamond I'll go ahead & merge this since it already satisfies the requirements laid out at yesterday's meeting and more.  Since you'll be updating this to `Active` soon enough, perhaps you'd link to the "full example" (`cardano-nft-js-toolkit`) then (when you list your active implementations). 😎

nftguild:
In the 20th NFT Guild Roundtable we explored CIP 71 with its author @thaddeusdiamond 

Watch the entire Roundtable:
https://youtu.be/Yyx4EEFXRh4

Or read through the transcript:
https://drive.google.com/file/d/1gz9z0Gp0nh2uK6OgdmhmQHg-8OzAJbvS/view?usp=sharing

Issue #345, opened 2022-10-03
CIP-0025: Why is the image tag marked as required?


Jack-0:
CIP 25's title is  `NFT Metadata Standard`  which implies it's the standard for Non Fungible Tokens on Cardano. 

However the `image` property tag is required in the current specification. Implying all 721 based NFTs require an image. 
Is 721 metadatum metadata for all assets or only image focused assets? If so should the title be `NFT Image Metadata standard` for consistency and to avoid confusion.

This might seem pedantic but it matters when parsing 721 metadata to ensure it applies to the standards proposed in CIP25. The proposal currently implies without an image tag the NFT metadata is invalid.

rphair:
The reason I remember was that image thumbnails both:
1) can be made from data types that are not images
2) are needed according to the semantic of NFT wallets so they have _something_ to show for _any_ NFT.

The dialogue establishing this on the Cardano Forum begins, as far as I can tell, with this post: https://forum.cardano.org/t/cip-nft-metadata-standard/45687/125

... and is the subject of the first review of the originally posted CIP here: https://github.com/cardano-foundation/CIPs/pull/85

I remember some discussion of this at the CIP editor meetings: though I was only attending as an observer, not an editor, back then.  If you need more detail you can look in the `BiweeklyMeetings` section of the repo around the date the comments above were posted, since I remember there was vigorous discussion in the meeting about it.  Though @KtorZ @SebastienGllmt @crptmppt may remember otherwise, it was the best way to handle representation in the wallets to the owner without a lot of subjectivity about what the wallets should do or how dynamically generated thumbnails would look.

Jack-0:
Thank for the link to the discussion, it looks to be required to have something to represent the NFT. Though marking as required seems to me as incorrect behavior. At worse if an NFT has no `image` then a wallet / 'user' could fall back on the asset name (centered in a div or something), this seems to be the default behavior with most 'wallets'. I would suggest altering CIP 25 to change `image` from required to **suggested**. 

I've replied to that discussion. I'll update or close this issue depending on the outcome generated.

SmaugPool:
CIP 25 is for NFTs with media files. You can make an NFT without CIP 25, and without any metadata actually. There are a lot of such NFTs used by DEXes for example.

If you don't need media associated with the NFT, maybe you simply don't need CIP 25.

Jack-0:
[...]

PR #353, opened 2022-10-16
CIP0025 | retitle to Media NFT Metadata Standard

Jack-0:


rphair (commented):
@SmaugPool this seems consistent with what you recently said here: https://github.com/cardano-foundation/CIPs/issues/345#issuecomment-1267363884

My understanding is that retitling it would be consistent with discussion so far, in which more than one person used media support as a rationale for making the NFT images mandatory (because all media were presumed to be representable by an image).

So if @alessandrokonrad has no objections then I would approve this change.

I do need to retitle this for _brevity_ and eliminate the question of whether this is being done for "clarity" (since the existing title isn't really unclear) rather than "precision", "accuracy", or "specificity" which were the words I would have used.

SmaugPool:
Yes, at some point the idea was to split the CIP in 2, one for the policy/asset structure that could be used for example for fungible tokens too, and one for the media specific metadata:

https://github.com/cardano-foundation/CIPs/pull/85#discussion_r629120254

This did not happen so I guess it makes sense to rename the whole CIP instead.

KtorZ:
Seems fine indeed, we also need to update the top-level README's table.

KtorZ (approved):

PR #355, opened 2022-10-18
CIP-0072? | DApp Registration & Discovery

ehanoc:
This proposal aims to standardize the process of metadata claims, dApp registration, **discovery** and verification. and to provide a way for DApp developers to make a set of claims when registering their Dapps and adding a way for other actors to index / crawl and **discovery** registration of DApps in the eco-system. 

It attempts to create a fair standard for actors to discover dapps being registered in the ecosystem and no boundaries to participate and collect information about dapps, authors, scripts, audits. Also by design tries to give options to the different actors of the system: 
- **Stores & Auditors** : from where to read metadata from (i.e off-chain sources)
- **Developers**:  Perform targeted releases (Where to write metadata so it can be found by stores, auditors & certifiers) 

Validation mechanisms:
- **integrity checks** from what is **anchored** on-chain vs off-chain metadata sources (i.e cip-26, ipfs, others)
- **trust** - each other can maintain and choose it's own list of trusted entities / public keys to assert trust in the signed certificates.  
  - This can be expanded by using **identity** concepts and including extra information as part of the metadata that is registered (i.e DIDs)

---

([rendered proposal in branch](https://github.com/ehanoc/CIPs/blob/cip-72_dapp-registration/CIP-0072/README.md))


matiwinnetou:
1. Great starting point(!)
2. Our experience building DappsOnCardano.com is that a dapp has formal or informal version, version is nothing else than collection of scripts. Scripts can  be only Plutus V1, only Plutus V2 or mixture of both. Some dApps are one script based but many dApps collectively are a list of scripts working together.
3. Is this needed for lace wallet only or will there be other use cases?
4. Are you also planning within the same spec for certification companies issue a dapp certificate (based on certification results) or that would have to be a different CIP?

matiwinnetou:
@ehanoc would love to chat about this. We managed to iterate over domain model for dApps at: https://github.com/Cardano-Fans/crfa-offchain-data-registry, while this is probably richer than what you wanted to achieve there could be some valueable learnings there for this CIP.

ehanoc:
> 1. Great starting point(!)
> 2. Our experience building DappsOnCardano.com is that a dapp has formal or informal version, version is nothing else than collection of scripts. Scripts can  be only Plutus V1, only Plutus V2 or mixture of both. Some dApps are one script based but many dApps collectively are a list of scripts working together.
> 3. Is this needed for lace wallet only or will there be other use cases?
> 4. Are you also planning within the same spec for certification companies issue a dapp certificate (based on certification results) or that would have to be a different CIP?

@matiwinnetou Thanks!

2) Would love to explore these and trying to consider them into a possible solution.
3) Overall use cases. But maybe not just relevant to wallets, i think this is a general metadata problem
4) I have some ideas for dapp certification, which are not metadata related; mostly identity & DID based. 

But i've set this to draft because i believe it's important to describe the problem first before coming up with the solution. In order with the new guideliens, i've open a draft [CPS](https://github.com/cardano-foundation/CIPs/pull/371) for this purpose. I think if we nail the problem first in a concise way, we can come up with a (or set of) solutions for it that make sense.

CIP-0072/README.md
```diff
@@ -0,0 +1,107 @@
...
+
+This proposal aims to standardise the process of dApp registration and verification, and to provide a set of claims that dApp developers can use to register their dApp(s).
+
+## **Motivation**
+DApps can express a plethora of information. Some of this information could be claims about who the developer is, what is the dApp's associated metadata are, and more. This data lacks standarisation, persistence, and immutabulity. Data without these features, means that dApp users cannot verify if the dApp's expressed information is consistent across time. The goal of this CIP is to formalise how dApps register their information with a new transaction metadata format that can record the dApp's metadata, ownership, and potentially developer's identity. This formalisation means dApp users can verify if the data expressed by a dApp is consistent with what was registered on-chain.
```
matiwinnetou, line 17: typo: immutability

CIP-0072/README.md
```diff
@@ -0,0 +1,107 @@
...
+		"r": "5114674f1ce8a2615f2b15138944e5c58511804d72a96260ce8c587e7220daa90b9e65b450ff49563744d7633b43a78b8dc6ec3e3397b50080",
+		"s": "a15f06ce8005ad817a1681a4e96ee6b4831679ef448d7c283b188ed64d399d6bac420fadf33964b2f2e0f2d1abd401e8eb09ab29e3ff280600",
+		"algo": "Ed25519−EdDSA",
+		"pub": "b7a3c12dc0c8c748ab07525b701122b88bd78f600c76342d27f25e5f92444cde"
+	}
+}
+```
+
+### Properties
+*`subject`*: Identifier of the claim subject (i.e dapp). A UTF-8 encoded string. This uniquess of this property cannot be guaranteed by the protocol and multiple claims for the same subject may exist therefore it is required to exist some mechanism to assert trust in the *veracity* of this property. This can be done by knowning verifying the signature against a trusted / known public key or by using Decentralized Identifiers (DIDs) as described in the [DID Specification](https://www.w3.org/TR/did-core/).
```
matiwinnetou, line 41: I discussed similar topic with George from mLabs. Creator of a dApp needs to create some key and then all updates to dApp metadata (separate CIP coming from mLabs on this) should be with a subkey. By checking subkey of a key we can actually verify if the owner is the same.

One of the things we are hoping to work on is CIP-26 compatible dApp registry. Generally it is not that hard to do (read only access) but tricky part is testing whether dApp update has happened via the creator of the project or somebody else. Now if it is somebody else they could be malicious and update information maliciously, e.g. even pointing to the scam website.

CIP-0072/README.md
```diff
@@ -0,0 +1,107 @@
...
+
+*`cip26`*: An array of URLs that point to the location of the dApp's metadata. The client should try to fetch the metadata from the first URL in the array. If the client fails to fetch the metadata from the first URL, it should try the second URL, and so on. If the client fails to fetch the metadata from all URLs, it should consider the metadata to be invalid.
+
+The format of the metadata tree object is described in the [CIP-26](https://github.com/cardano-foundation/CIPs/tree/master/CIP-0026) specification.
+
+*`did`*(optional): The DID of the dApp's developer. The DID is used to verify the authenticity of the certificate. The client should use the DID to fetch the developer's public key from the Cardano blockchain. The client should then use the public key to verify the signature of the certificate. This field is *optional* and can be used by identity centric dApps to verify the authenticity of the developer.
```
matiwinnetou, line 49: 👍

CIP-0072/README.md
```diff
@@ -0,0 +1,107 @@
...
+
+## **Specification**
+
+### **On-chain dApp Registration certificate**
+
+```json
+{
+	"subject": "d684512ccb313191dd08563fd8d737312f7f104a70d9c72018f6b0621ea738c5b8213c8365b980f2d8c48d5fbb2ec3ce642725a20351dbff9861ce9695ac5db8",
+	"rootHash": "8c4e9eec512f5f277ab811ba75c991d51600c80003e892e601c6b6c19aaf8a33",
+    "cip26": ["http://somelocation.io/", "http://somelocation2.io/"],
```
matiwinnetou, line 29: @aleks-djuric

CIP-0072/README.md
ehanoc: Thank you @matiwinnetou. 

I've added an extra `type` (register, update), perhaps this would help. Over time per subject we can trace the actions of the owner with different registrations. We can add extra actions if necessary. Open to discuss alternatives

CIP-0072/README.md
ehanoc: thank you :+1: typo fixed

CIP-0072/README.md
```diff
@@ -0,0 +1,126 @@
...
+
+*`rootHash`*: The hash of the metadata entire tree object. This hash is used by clients to verify the integrity of the metadata tree object. When reading a metadata tree object, the client should calculate the hash of the object and compare it with the `rootHash` property. If the two hashes don't match, the client should discard the object. The metadata tree object is a JSON object that contains the dApp's metadata. The metadata tree object is described in the next section.
+
+This hash is calculated by taking the entire metadata tree object, ordering the keys in the object alphanumerically, and then hashing the resulting JSON string using the blake2b hashing algorithm. The hash is encoded as a hex string.
+
+*`cip26`*: An array of URLs that point to the location of the dApp's metadata. The client should try to fetch the metadata from the first URL in the array. If the client fails to fetch the metadata from the first URL, it should try the second URL, and so on. If the client fails to fetch the metadata from all URLs, it should consider the metadata to be invalid.
```
flyerman, line 50: ❓ Would it be preferable to have an array of `{subject}` instead of URLs?  This would keep the spirit of CIP-26 `We therefore believe that it is critical that we allow users to choose which metadata server (or servers) they refer to.`.  In this context, clients can be dapps registrants or dapps certifiers.  Dapps certifiers could publish the list of metadata servers they trust and dapps developers seeking for certification would have to target the list of metadata servers used by the desired dapps certifiers.

Moreover, in the event that a metadata server operator need to update its base URL, developers could have to re-issue registrations and certifiers would have to certify again.

CIP-0072/README.md
```diff
@@ -0,0 +1,126 @@
...
+  "cip26": ["http://somelocation.io/", "http://somelocation2.io/"],
+	"did": "did:ada:f2019bd31a8530fb67c6d81da758dfa9a65be09d835cf2cd361d595a8858301d",
+	"signature": {
+		"r": "5114674f1ce8a2615f2b15138944e5c58511804d72a96260ce8c587e7220daa90b9e65b450ff49563744d7633b43a78b8dc6ec3e3397b50080",
+		"s": "a15f06ce8005ad817a1681a4e96ee6b4831679ef448d7c283b188ed64d399d6bac420fadf33964b2f2e0f2d1abd401e8eb09ab29e3ff280600",
+		"algo": "Ed25519−EdDSA",
+		"pub": "b7a3c12dc0c8c748ab07525b701122b88bd78f600c76342d27f25e5f92444cde"
+	}
+}
+```
```
flyerman, line 39: 🚧 I suggest to add a schema version field, something like `dapps_registration_version: 1.0`.  We can anticipate that there will be newer schema versions and Dapps Stores will need to support multiple formats over time as well as potentially deprecate or ignore older ones.

CIP-0072/README.md
```diff
@@ -0,0 +1,126 @@
...
+  ]
+}
+```
+
+## Metadata Label
+
+When submitting the transaction metadata picks one of the following values for `transaction_metadatum_label`:
+
+- `666`: DApp Registration
+- `667`: DApp Certification
```
flyerman, line 126: ❔ Why `666` and shall we chose a less conotated number?

flyerman (commented):
👍 Great proposal.  I think it's a good step forward for the user experiences and the ecosystem.  There is still details to iron out but I like the general concept and the proposed solution.

See my comments and questions below:

CIP-0072/README.md
ehanoc: Thanks @flyerman . I think that's a great idea. In that way we won't be "marrying" the registration / certifications to a particular spec (I.e CIP-26) and let the different actors advertise which off-chain storage they pay attention to. 

I've updated spec to move torwards this direction!

CIP-0072/README.md
ehanoc: Fair :+1:  ; Picked a less conotated number that was available according to the https://github.com/cardano-foundation/CIPs/blob/master/CIP-0010/registry.json

CIP-0072/README.md
```diff
@@ -0,0 +1,412 @@
...
+            "auditId": "z5L90f",
+            "scripts": [
+                {
+                    "id": "PmNd6w",
+                    "version": 1
+                }
+            ]
+        }
+    ],
+    "scripts": [
```
ehanoc, line 372: @matiwinnetou As we discuss last i'm including a similar structure of the suggested metadata structure for review and more collaboration.

CIP-0072/README.md
```diff
@@ -0,0 +1,413 @@
...
+- [CIP-26](https://github.com/cardano-foundation/CIPs/tree/master/CIP-0026) compliant servers
+- IPFS
+
+##### **Targetted Releases**
+Each developer and publisher can choose to where to write metadata based on the information available from known stores & auditors. This gives **developers** and **publishers** the ability to perform targeted releases. (i.e to which stores and auditors)
+
+#### **Suggested Validations**
+
+- **`integrity`**: The dapp's metadata off-chain should match the metadata anchored on-chain.
+- **`trust`**: The dApp's certificate should be signed by a trusted entity. It's up to the store/auditor to decide which entities are trusted and they should maintain their own list of trusted entities. Although this entities might be well known, it's not the responsibility of this CIP to maintain this list.
```
simonjohnthompson, line 51:
```suggestion
- **`trust`**: The dApp's certificate should be signed by a trusted entity. It's up to the store/auditor to decide which entities are trusted and they should maintain and publish their own list of trusted entities. Although this entities might be well known, it's not the responsibility of this CIP to maintain this list.
```

CIP-0072/README.md
```diff
@@ -0,0 +1,413 @@
...
+		"r": "5114674f1ce8a2615f2b15138944e5c58511804d72a96260ce8c587e7220daa90b9e65b450ff49563744d7633b43a78b8dc6ec3e3397b50080",
+		"s": "a15f06ce8005ad817a1681a4e96ee6b4831679ef448d7c283b188ed64d399d6bac420fadf33964b2f2e0f2d1abd401e8eb09ab29e3ff280600",
+		"algo": "Ed25519−EdDSA",
+		"pub": "b7a3c12dc0c8c748ab07525b701122b88bd78f600c76342d27f25e5f92444cde"
+	}
+}
+```
+
+### Properties
+*`subject`*: Identifier of the claim subject (i.e dapp). A UTF-8 encoded string. This uniquess of this property cannot be guaranteed by the protocol and multiple claims for the same subject may exist therefore it is required to exist some mechanism to assert trust in the *veracity* of this property.
```
simonjohnthompson, line 75: Note that there might be multiple valid claims, e.g. for different audits of the same (version of a) DApp, presumably by different auditors; and indeed there might be multiple claims by _the same claimant_, by way of updates.

CIP-0072/README.md
```diff
@@ -0,0 +1,413 @@
...
+
+*`type`*: The type of the claim. This is a JSON object that contains the following properties: 
+- *`action`*: The action that the certificate is asserting. It can take the following values: 
+  - *`REGISTER`*: The certificate is asserting that the dApp is being registered for the first time. 
+  - *`UPDATE`*: The certificate is asserting that the dApp is being updated.
+
+*`rootHash`*: The hash of the metadata entire tree object. This hash is used by clients to verify the integrity of the metadata tree object. When reading a metadata tree object, the client should calculate the hash of the object and compare it with the `rootHash` property. If the two hashes don't match, the client should discard the object. The metadata tree object is a JSON object that contains the dApp's metadata. The metadata tree object is described in the next section.
+
+This hash is calculated by taking the entire metadata tree object, ordering the keys in the object alphanumerically, and then hashing the resulting JSON string using the blake2b-256 hashing algorithm. The hash is encoded as a hex string.
```
simonjohnthompson, line 84: Is this specific enough to be canonical? Discussion on Stack overflow suggests that need to handle e.g. whitespace as well? https://stackoverflow.com/questions/4670494/how-to-cryptographically-hash-a-json-object

CIP-0072/README.md
```diff
@@ -0,0 +1,413 @@
...
+
+*`signature`*: The signature of the certificate. The signature is done over the blake2b hash of the certificate. The client should use the public key to verify the signature of the certificate.
```
simonjohnthompson, line 86: Throughout uses both `blake2b` and `blake2b-256`; should all uses prescribe the size?

CIP-0072/README.md
```diff
@@ -0,0 +1,413 @@
...
+		"pub": "b7a3c12dc0c8c748ab07525b701122b88bd78f600c76342d27f25e5f92444cde"
+	}
+}
+```
+
+### Properties
+*`subject`*: Identifier of the claim subject (i.e dapp). A UTF-8 encoded string. This uniquess of this property cannot be guaranteed by the protocol and multiple claims for the same subject may exist therefore it is required to exist some mechanism to assert trust in the *veracity* of this property.
+
+*`type`*: The type of the claim. This is a JSON object that contains the following properties: 
+- *`action`*: The action that the certificate is asserting. It can take the following values:
```
simonjohnthompson, line 78: Will we also add `CERTIFICATION` here? (Though there's a horrible ambiguity in usage here; not your fault.)

CIP-0072/README.md
```diff
@@ -0,0 +1,413 @@
...
+                "publicKey": "0x1234567890abcdef"
+              }
+          ]
+        }
+    ]
+}
+```
+
+### **Scripts Audits**
+As part of the metadata we will be including a list of audits for each script. This will be a list of audits that have been performed on the script. The audits will be performed by a third party auditor and will be signed by the auditor.
```
simonjohnthompson, line 402:
```suggestion
As part of the metadata we will be including a list of audits for each DApp. This will be a list of audits that have been performed on the deployed scripts comprising the DApp. The audits will be performed by a third party auditor and will be signed by the auditor. 
```

simonjohnthompson, line 402: Note that in practice this information might be provided by a _separate_ DApp registration, that simply links the audit to the deployed set of scripts comprising a DApp version.

ehanoc: I guess here that's definitely an option. That the auditors / certification issuers emit their own cert (maybe some structure) and they are responsible for going through every  single script and metadata registered. 

The other option could be that they only look at scripts and if the developer can include the audit of every script they use; we could consider that as "certified"?

CIP-0072/README.md
```diff
@@ -0,0 +1,413 @@
...
+As part of the metadata we will be including a list of audits for each DApp. This will be a list of audits that have been performed on the deployed scripts comprising the DApp. The audits will be performed by a third party auditor and will be signed by the auditor. 
+
+Some of the properties included in the audit are:
+
+- **auditType** property will be included that can take one of the following values: MANUAL, AUTOMATED, or MIXED.
+- **auditId** property will be included that will be a unique identifier for the audit.
+- **auditLink** property will be included that will be a link to the audit report.
+- **signature** property will be included that will be a signature of the audit report.
+- **publicKey** property will be included that will be the public key of the auditor.
+
```
simonjohnthompson, line 411:
```
# Certification

This section describes how certification metadata can be included in CIP-72-compliant metadata for DApp registration. It first describes the requirements for DApp certification, and then sets out the options for its inclusion in CIP-72.

Note: In this section ‘certification’ is used to mean the process of providing various kinds of assurance of DApps through the provision of verifiable, trustworthy metadata linking to relevant evidence of assurance, such as audit reports, test run data etc.

It is expected that evidence of various kinds of assurance of DApps is recorded in an immutable and verifiable way on the Cardano blockchain; these forms of assurance are: testing, including property-based testing (level 1), audit (level 2) and formal verification (level 3). Certification is to be provided by certification providers including: testing services (level 1), auditors (level 2), and verification services (level 3).

## Requirements

Certification metadata should be related to the set of scripts that constitute the deployed version of the DApp that has been certified. 

Certification metadata should contain the following information.:

* the set of deployed scripts to which the metadata pertains
* the certification provider
* the hash of the offchain meta-data by the certification provider
* links to relevant evidence off chain, including, as appropriate,
  * source code repository
  * compiler and version used
  * parameter values for parametric validators 
  * audit report	
  * audit summary (max 100 words)
  * test run data
  * formal proof objects. 
  * certified compilation proof object, etc.

The certification metadata should be linked to DApp registration, as outlined in this CIP.  This requires the registration of any certified DApp prior to, or simultaneously with, the submission of the certification metadata. While audit and testing may take place before a DApp is registered or released, certification is only significant once the DApp is released, and so it can take place once the DApp is registered.

Certification metadata should be signed by the certification provider. It is assumed that certification providers make available their public keys for signature verification.

The metadata should be discoverable by all certification stakeholders, including end-users, DApp developers, and ecosystem components, such as light wallets and DApp stores. Information should be indexable by certification provider, DApp developer, DApp and DApp version.

It should be possible for there to be multiple versions of metadata published by the same certification provider for the same version of a DApp, particularly when a (minor) update of the evidence is necessary. In such a case it should be possible to identify the most recent version of the report for that DApp version. It will also be the case that the same DApp may have certification metadata provided by multiple different certification providers.

It should be possible for wallets to identify to users the certification status of a DApp when they are signing a transaction that is being submitted to a deployed DApp.

## Mechanisms

Evidence of certification is to be posted as explained here. Line references here relate to [CIP-0072? | DApp Registration & Discovery](https://github.com/cardano-foundation/CIPs/pull/355/files#top). 


Evidence is to be published as a separate on-chain item, but as a new `type` of CIP-72 registration.
* Use the same mechanism as for certification (lines 88–142), with a new `type`, `CERTIFICATION` (lines 77–80).
* Assumes that the `subject` will be common with the registration and update of the DApp.
* Assume that it has the same value for `transaction_metadatum_label`, 1667 (lines 143–147).
* This option requires a mechanism to link the certification data to the DApp: this is done via the subject field of the metadata (“the DApp”) and the set of deployed scripts (“the DApp version”).

Advantages
* This item can be managed by the certificate provider independently of other parties. 
* This item can be concerned with certification data only.
* This item is identified by the same label as registration data.
* The metadata will only require update if there is a change in the certification information for the given set of deployed scripts.

Disadvantages
* Requires DApp stores to implement the linking mechanism described above.

Question 
* Will using this typing mechanism require the type `UPDATE` to be split into `REG-UPDATE` and `CERT-UPDATE`?


## General discussion 
1. The model of update outlined in CIP-72, using an explicit `type` for updates, seems to be a particular version of the more general mechanism described in [CIP-26](https://cips.cardano.org/cips/cip26/). It would be clearer to adopt the CIP-26 model for managing metadata, rather than being mechanism agnostic but having to reinvent some aspects of it,
1. From the certification perspective, it makes no sense to audit single contracts. All audit and certification activity is related to DApps, that is to sets of deployed scripts. This needs to be reflected in CIP-72, which links audits to scripts, individually. Of course, it is possible to link from a single script to an audit via the set(s) of scripts to which it belongs (giving the context for the audit).
1. Certificates in CIP-72 are classified as `MANUAL`, `AUTOMATED`, `MIXED`. Preferably this should be levels 1, 2 and 3, fitting with the widely discussed approach to assurance.
1. There needs to be a canonical mechanism for identifying a set of scripts. Two options:
  * Order the script hashes, concatenate them, and then hash the result.
  * Order the scripts according to script names, hash them, concatenate and hash.

```

simonjohnthompson: Suggest certification of assurance evidence as outlined above.

CIP-0072/README.md
```diff
@@ -0,0 +1,413 @@
...
+##### **Targetted Releases**
+Each developer and publisher can choose to where to write metadata based on the information available from known stores & auditors. This gives **developers** and **publishers** the ability to perform targeted releases. (i.e to which stores and auditors)
+
+#### **Suggested Validations**
+
+- **`integrity`**: The dapp's metadata off-chain should match the metadata anchored on-chain.
+- **`trust`**: The dApp's certificate should be signed by a trusted entity. It's up to the store/auditor to decide which entities are trusted and they should maintain and publish their own list of trusted entities. Although this entities might be well known, it's not the responsibility of this CIP to maintain this list.
+
+### **On-chain dApp Registration certificate**
```
Ryun1, line 53: Is there any use case for attaching an expiration to metadata registrations? such as "after X epochs this metadata is out of date and should be updated" - Although this may introduce problems with updates being missed by chain followers who are waiting for expirations before checking chain.

CIP-0072/README.md
```diff
@@ -0,0 +1,413 @@
...
+- [CIP-26](https://github.com/cardano-foundation/CIPs/tree/master/CIP-0026) compliant servers
+- IPFS
+
+##### **Targetted Releases**
+Each developer and publisher can choose to where to write metadata based on the information available from known stores & auditors. This gives **developers** and **publishers** the ability to perform targeted releases. (i.e to which stores and auditors)
+
+#### **Suggested Validations**
+
+- **`integrity`**: The dapp's metadata off-chain should match the metadata anchored on-chain.
+- **`trust`**: The dApp's certificate should be signed by a trusted entity. It's up to the store/auditor to decide which entities are trusted and they should maintain and publish their own list of trusted entities. Although this entities might be well known, it's not the responsibility of this CIP to maintain this list.
```
Ryun1, line 51: Could this be linked to the auditors list from [CIP-0052](https://github.com/cardano-foundation/CIPs/tree/master/CIP-0052)?

CIP-0072/README.md
```diff
@@ -0,0 +1,413 @@
...
+            "s": {
+				"type": "string",
+				"description": "hex representation of the S component of the signature"
+			},
+              
+		},
+		"required": ["r", "s", "algo", "pub"]
+      }
+    },
+    "required": ["subject", "rootHash","type", "signature"]
```
Ryun1, line 139: Add `schema_version` to the required? also I suggest changing `schema_version` to match how versioning was done in [CIP-25](https://github.com/cardano-foundation/CIPs/blob/master/CIP-0025/README.md?plain=1#L67) - that it is just an integer.

CIP-0072/README.md
```diff
@@ -0,0 +1,413 @@
...
+
+Some of the properties included in the audit are:
+
+- **auditType** property will be included that can take one of the following values: MANUAL, AUTOMATED, or MIXED.
+- **auditId** property will be included that will be a unique identifier for the audit.
+- **auditLink** property will be included that will be a link to the audit report.
+- **signature** property will be included that will be a signature of the audit report.
+- **publicKey** property will be included that will be the public key of the auditor.
+
+### **Stores Custom fields**
```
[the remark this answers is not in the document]
ehanoc: I think that each store should advertise on their own these. Adding them here would make this CIP some sort of database for available stores and somewhat permission ed based on what get's merged or not.

CIP-0072/README.md
```diff
@@ -0,0 +1,413 @@
...
+
+### **On-chain dApp Registration certificate**
+
+```json
+{
+	"subject": "d684512ccb313191dd08563fd8d737312f7f104a70d9c72018f6b0621ea738c5b8213c8365b980f2d8c48d5fbb2ec3ce642725a20351dbff9861ce9695ac5db8",
+	"rootHash": "8c4e9eec512f5f277ab811ba75c991d51600c80003e892e601c6b6c19aaf8a33",
+  "schema_version": "0.0.1",
+  "type": { 
+    "action": "REGISTER",
```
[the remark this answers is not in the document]
ehanoc: Making it explicit, gives more information to stores when indexing updates. For example, "have i seen a previous version of this dapp, did i trusted the previous update?". Making this ambiguous, will make it more difficult for an operator to decide it's there's history of this dapp / developer

mgajda:
Will it include a description of `Datum` format, including constructors and documentation link?

rphair:
thanks @ehanoc ... looks like ball is back in our court to review this again (@Ryun1 @simonjohnthompson @flyerman @matiwinnetou 🙏)

CIP-0072/README.md
```diff
@@ -0,0 +1,413 @@
...
+    "releaseNumber": "1.0.0",
```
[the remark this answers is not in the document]
ehanoc: I think the stores / auditors and all other actors should include it as part of their own validation. Any other ideas?

CIP-0072/README.md
ehanoc: Interesting. Could you suggest a change to the CIP? 

CC: @matiwinnetou

ehanoc:
> Will it include a description of `Datum` format, including constructors and documentation link?

@mgajda That could be a nice add. 

Could you suggest an update to the metadata to include the Datum format / schema?

CC: @matiwinnetou @simonjohnthompson @flyerman @refi93

ehanoc:
> thanks @ehanoc ... looks like ball is back in our court to review this again (@Ryun1 @simonjohnthompson @flyerman @matiwinnetou pray)

Thank you @rphair . 

We've been having weekly meeting to discuss, align and refine; ended up being much simplified after de-scoping trying to fit Certification into it (likely a separate proposal). 

We're open to further review and discussion.

fallen-icarus:
I do not think using metadata labels is a good idea. There is no way to cryptographically guarantee the validity of the metadata. Instead, you should look to use [beacon tokens](#466). By tagging transactions with a native token that **only the dapp creators can mint**, the metadata is then easily queryable off-chain and the information is cryptographically guaranteed to be from the dapp creators. Here is a similar reference implementation I created for personal payment address books on cardano ([cardano-address-book](https://github.com/fallen-icarus/cardano-address-book)). The address book example can be generalized to any kind of information. IMHO, all metadata CIPs should migrate to using beacon tokens instead of metadata labels. The specific beacon token itself is the label.

fallen-icarus:
Also I think you should use datums instead of metadata because the format can be strictly enforced. If a required field is missing, registration is guaranteed to fail. Here is a reference implementation that does something similar using beacon tokens ([cardano-datum](https://github.com/fallen-icarus/cardano-datums)). Also, with beacon tokens, the datum utxos can be consumed while still leaving the datum information accessible.

CIP-0072/README.md
```diff
@@ -0,0 +1,382 @@
...
+  "category": "GAMING",
+  "subCategory": "RPG",
+  "description": {
+    "short": "A story rich game where choices matter"
+  },
+  "releases": [
+    {
+      "releaseNumber": 1,
+      "releaseName": "V1",
+      "auditId": "z5L90f",
```
matiwinnetou, line 343: I thought we agreed to remove audits?

ehanoc: Ah yes; missed this one. Thanks @matiwinnetou

ehanoc:
> I do not think using metadata labels is a good idea. There is no way to cryptographically guarantee the validity of the metadata. 

We achieve the same and explain how to verify correctness and assert trust of the metadata.  There's also precedence for similar flows (i.e stake pool registrations). We also keep this one to expand into certification of dapps (not just the scripts, datums) and identity concepts (i.e DIDs).  

I'm all up for competing proposals and if that one becomes the accepted one, we're happy to implement accordingly. We had several community members engaging in this one and we're happy to submit to review.

CIP-0072/README.md
```diff
@@ -0,0 +1,384 @@
...
+Each store might have their own requirements for the metadata. For example, some stores might require a field for logo, or screenshots links. The store's should adviertise what fields they require in their documentation so that developers are aware and they can include them in the metadata. 
+
+### **Offchain Metadata Storage**
+
+There are multiple options to store metadata offchain. The most common options are:
+- [CIP-26](https://github.com/cardano-foundation/CIPs/tree/master/CIP-0026) compliant servers
+- [IPFS](https://ipfs.tech/)
+- [Bitbucket](https://bitbucket.org/)
+- Any REST JSON API
+
```
Ryun1, line 384:
```suggestion

## Rationale: how does this CIP achieve its goals?
<!-- The rationale fleshes out the specification by describing what motivated the design and what led to particular design decisions. It should describe alternate designs considered and related work. The rationale should provide evidence of consensus within the community and discuss significant objections or concerns raised during the discussion.
It must also explain how the proposal affects the backward compatibility of existing solutions when applicable. If the proposal responds to a CPS, the 'Rationale' section should explain how it addresses the CPS, and answer any questions that the CPS poses for potential solutions.-->

## Path to Active

### Acceptance Criteria
<!-- Describes what are the acceptance criteria whereby a proposal becomes 'Active' -->

### Implementation Plan
<!-- A plan to meet those criteria. Or `N/A` if not applicable. -->

## Copyright
<!-- The CIP must be explicitly licensed under acceptable copyright terms. -->
```

Please add these sections.

SebastienGllmt:
> Will it include a description of Datum format, including constructors and documentation link?

No, CIP57 is what people would use for that

My understanding is that this CIP is ready to merge as long as:
1. We add the rationale and other editorial changes
2. Lace wallet launches in production with this feature to have at least 1 production app using this

matiwinnetou:
**Open points:** / TODO
- schema_version is required (double check if it should be underscore or camel case) / is it string, is it integer?
- Make changes according to: https://github.com/cardano-foundation/CIPs/blob/master/.github/CIP-TEMPLATE.md?plain=1 including **rationale**
- canonical json for offchain CIP-26 part

CIP-0072/README.md
Ryun1: This is outdated idea from me, not worth pursuing.

rphair (commented):
last changes look good to me, thanks @matiwinnetou; FYI the H1 title will be auto-generated in web contexts so we're recommending authors remove it here...

CIP-0072/README.md
matiwinnetou: I am working on this, thanks

CIP-0072/README.md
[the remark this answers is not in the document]
matiwinnetou: @Ryun1 - at first I thought that this is a good comment and we should put that versions should be always increasing for new updates but then I realised that somebody may have missed an update and wants to retrospectively fill it. 

I think it is best not to stipulate such things albeit it looks like a very good idea at first.

CIP-0072/README.md
```diff
@@ -0,0 +1,384 @@
...
+---
+
+# CIP-0072: Cardano DApp Registration & Discovery
+
+## Abstract
+DApp developers do not have a standardised method to record immutable, persistent claims about their dApp(s) that their users can verify. A dApp developer needs to "register" their dApp by providing a set of claims about their dApp(s) that can be verified by the user. This CIP describes a standardised method for dApp developers to register their dApp(s) and for users to verify the claims made by dApp developers.
+
+This proposal aims to standardise the process of dApp registration and verification, and to provide a set of claims that dApp developers can use to register their dApp(s).
+
+## Motivation
```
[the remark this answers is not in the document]
Ryun1: Thanks!

CIP-0072/README.md
Ryun1: Thanks for adding!

CIP-0072/README.md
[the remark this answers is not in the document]
Ryun1: @matiwinnetou - Makes sense to me :)

CIP-0072/README.md
```diff
@@ -0,0 +1,560 @@
...
+some stores simply do not require dApps to add their scripts in order to be listed.
+
+### Schema Version or CIP Version
+We want to take schema version further and actually attest the whole CIP version so we are introducing a top level field: version. Any changes to the document will require bumping the version. CIP version needs to be also
+propagated in the on-chain JSON itself.
+
+### Tags
+We briefly discussed tags and we will likely introduce tags in the near future. An array of tags to help stores / dApp developers categories where their dApp should show. This will complement `categories` field.
+
+### DE-REGISTER
```
Ryun1, line 536: This addition raises a question for me:

- What should happen if a publisher registers then de-registers then registers again? is this a valid flow?

rphair:
Tx metadatum label addition under way in https://github.com/cardano-foundation/CIPs/pull/505.

CIP-0072/README.md
```diff
@@ -0,0 +1,603 @@
...
+### **On-chain dApp Registration Certificate**
+
+The on-chain dApp registration certificate MUST follow canonical JSON and be serialised according to RFC 8785 (https://www.rfc-editor.org/rfc/rfc8785). This stipulation is to avoid any ambigiutines in the signature calculation.
+
+```json
+{
+  "subject": "b37aabd81024c043f53a069c91e51a5b52e4ea399ae17ee1fe3cb9c44db707eb",
+  "rootHash": "7abcda7de6d883e7570118c1ccc8ee2e911f2e628a41ab0685ffee15f39bba96",
+  "metadata": [
+    "https://cip26.foundation.app/my_dapp.json"
```
kolesarp1, line 55: I feel like **cip26.** prefix can be removed from the link @matiwinnetou

CIP-0072/README.md
```diff
@@ -0,0 +1,614 @@
...
+on-chain_metadata = {
+  subject: string,
+  rootHash: sig_256,
+  metadata: [+ string] / [+ string / [+ string]],
+  type: registration / de-registration,
+  signature: signature,
+}
+
+registration = {
+  action: "REGISTRATION",
```
marcin-mazurek, line 103:
```suggestion
  action: "REGISTER",
```

CIP-0072/README.md
```diff
@@ -0,0 +1,614 @@
...
+         }
+       },
+       "type":{
+          "type":"object",
+          "description":"Describes the releases, if they are new or an updates.",
+          "properties":{
+             "action":{
+               "type":"string",
+               "enum":["REGISTER", "DE_REGISTER"],
+               "pattern": "^(REGISTER|DE_REGISTER)$",
```
marcin-mazurek, line 171: If this is defined as enum, do we additionally need the regex?

CIP-0072/README.md
```diff
@@ -0,0 +1,614 @@
...
+              "releaseNumber": {
+                "type": "string",
+                "pattern": "^(0|[1-9][0-9]*)\\.(0|[1-9][0-9]*)\\.(0|[1-9][0-9]*)(-((0|[1-9][0-9]*|[0-9]*[a-zA-Z-][0-9a-zA-Z-]*)(\\.(0|[1-9][0-9]*|[0-9]*[a-zA-Z-][0-9a-zA-Z-]*))*))?(\\+([0-9a-zA-Z-]+(\\.[0-9a-zA-Z-]+)*))?$",
+                "description": "Semver compatible release number (major.minor.patch) or (major.minor.patch-some_text) e.g. 1.2.3 or 1.1.1-alpha"
+              },
+              "releaseName": {
+                "type": "string",
+                "description": "An optional human readable release name, e.g. V1"
+              },
+              "security_vulnerability": {
```
marcin-mazurek, line 333: Small suggestion - we are naming other properties with camelCase:
```suggestion
              "securityVulnerability": {
```

CIP-0072/README.md
matiwinnetou: probably not, good point

CIP-0072/README.md
matiwinnetou: I already removed it, thanks

rphair (approved):
Review at last meeting highlights already live & reference implementations, so should be merged as Proposed to facilitate further adoption.  Therefore marking Last Check and we should also get @KtorZ to review and @Ryun1 will also add comments.

I've also been reading this periodically and I find it quite satisfactory today & acknowledge it's also well received by apparently unaffiliated parties and therefore @ehanoc our merging as `Proposed` would solve the chicken-egg problem of merging as CIP and evidence of further adoption depending upon each other.

CIP-0072/README.md
```diff
@@ -0,0 +1,613 @@
...
+                "type":"string",
+                "description": "An optional script name usually related to it's function."
+              },
+              "purposes":{
+                "type":"array",
+                "items": {
+                    "type": "string",
+                      "enum":["SPEND", "MINT"]
+                },
+                "description": "Purpouses of the script, SPEND or MINT (notice it can be both for some modern Cardano languages)."
```
Ryun1, line 387:
```suggestion
                "description": "Purposes of the script, SPEND or MINT (notice it can be both for some modern Cardano languages)."
```
This misspelling happens a few times :)

CIP-0072/README.md
```diff
@@ -0,0 +1,613 @@
...
+We are introducing a field in the on-chain JSON only, which allows dApp development teams to provide a free text field
+comment about changes they are making in a given (re-)registration request.
+
+## Path to Active
+We will evaluate for a few months if we have not missed any details, collect feedback from dApp developers, stores. We reserve right to make small changes in this time, while the proposal
+is in `PROPOSED` status. After certain period (no longer than 6 months), from time of merging PR in `PROPOSED` status to the main / master branch, we will
+update the proposal to be in `ACTIVE` state.
+
+### Acceptance Criteria
+- IOG and CF approval
```
Ryun1, line 604: How is this measured?

CIP-0072/README.md
```diff
@@ -0,0 +1,613 @@
...
+- Community representative approval (Santiago / TxPipe and Marcel / Eternl wallet)
+- PR to https://github.com/cardano-foundation/CIPs/blob/master/CIP-0010/registry.json
+
+### Implementation Plan
+- Reference implementation: https://github.com/Cardano-Fans/crfa-dapp-registration-and-certification-service (incubator)
+
+## Copyright
+[CC-BY-4.0]: https://creativecommons.org/licenses/by/4.0/legalcode
```
Ryun1, line 612:
```suggestion
[CC-BY-4.0](https://creativecommons.org/licenses/by/4.0/legalcode)
```

CIP-0072/README.md
```diff
@@ -0,0 +1,613 @@
...
+[Apache-2.0]: http://www.apache.org/licenses/LICENSE-2.0
```
Ryun1, line 613:
```suggestion
```
Removed to match copyright in the header

Ryun1 (commented):
Great work on this proposal @matiwinnetou, for me this proposal is well written and has a solid technical approach and thus is a good candidate to merge.

I only have a few minor points - see below. 🤠

CIP-0072/README.md
KtorZ: More importantly, this CANNOT be an acceptance criteria. IOG and CF are not authorities when it comes to approving CIPs. Acceptance criteria must reflect actually measurable data points and show "proof of activity / usage".

CIP-0072/README.md
```diff
@@ -0,0 +1,613 @@
...
+
+The standard doesn't prevent anyone from making a claim, so it's up to the different operator to their diligence work and make their own choices of whom they trust. The signature should give the most confidence as anyone can collect known public keys from known development companies. Future CIP revisions can include DID's and Verifiable Credentials to tackle this problem in a more elegant way.
+
+### DIDs
+Since DIDs / Verifiable Credetials are not yet widely used in Cardano ecosystem, usage of them in this initial CIP version has been de-scoped.
+
+### Categories
+`Categories` is a predefined enum with values defined in the CIP / schema, it is *NOT* a free text field, rationale for this is that dApp developers will have no idea what ontology / classification to use, which will likely result in many duplicates of the same thing.
+
+### Purpouse Field As an Array or as a Single Item?
```
KtorZ, line 549:
```suggestion
### Purpose Field As an Array or as a Single Item?
```

CIP-0072/README.md
```diff
@@ -0,0 +1,613 @@
...
+It may have been visible that we have a `purpose` field, which can be: "SPEND" or "MINT", those fields directly map to what is allowed by a Cardano Smart Contract. As of the time of writing CIP - PlutusTx does not allow a script to be both of type: "SPEND" and "MINT", however, there are new
+languages on Cardano being worked on where they already allow one validator to be both spending UTxOs and minting tokens - all with the same script hash. To be open for the future it has been agreed to turn `purpouse` field into `purpouses` and make it a JSON array.
```
KtorZ, line 551:
```suggestion
languages on Cardano being worked on where they already allow one validator to be both spending UTxOs and minting tokens - all with the same script hash. To be open for the future it has been agreed to turn `purpose` field into `purposes` and make it a JSON array.
```

CIP-0072/README.md
```diff
@@ -0,0 +1,613 @@
...
+comment about changes they are making in a given (re-)registration request.
+
+## Path to Active
+We will evaluate for a few months if we have not missed any details, collect feedback from dApp developers, stores. We reserve right to make small changes in this time, while the proposal
+is in `PROPOSED` status. After certain period (no longer than 6 months), from time of merging PR in `PROPOSED` status to the main / master branch, we will
+update the proposal to be in `ACTIVE` state.
+
+### Acceptance Criteria
+- IOG and CF approval
+- Community representative approval (Santiago / TxPipe and Marcel / Eternl wallet)
```
KtorZ, line 605: If these people are meant to be implementors of the standard, then we should list them as implementors. And, it goes without saying that **they** should confirm having such role.

CIP-0072/README.md
```diff
@@ -0,0 +1,613 @@
...
+### Comment Field (on-chain JSON)
+We are introducing a field in the on-chain JSON only, which allows dApp development teams to provide a free text field
+comment about changes they are making in a given (re-)registration request.
+
+## Path to Active
+We will evaluate for a few months if we have not missed any details, collect feedback from dApp developers, stores. We reserve right to make small changes in this time, while the proposal
+is in `PROPOSED` status. After certain period (no longer than 6 months), from time of merging PR in `PROPOSED` status to the main / master branch, we will
+update the proposal to be in `ACTIVE` state.
+
+### Acceptance Criteria
```
KtorZ, line 603: As a concrete acceptance criteria, I'd suggest to have at least 2-3 non-trivial dApps **from different entities** registered through this method. That would serve as a good measure of its acceptance.

CIP-0072/README.md
```diff
@@ -0,0 +1,613 @@
...
+as having security vulnerability. We are intoducing an optional field in the offchain json on the release level: `securityVulnerability": true.
+
+### Comment Field (on-chain JSON)
+We are introducing a field in the on-chain JSON only, which allows dApp development teams to provide a free text field
+comment about changes they are making in a given (re-)registration request.
+
+## Path to Active
+We will evaluate for a few months if we have not missed any details, collect feedback from dApp developers, stores. We reserve right to make small changes in this time, while the proposal
+is in `PROPOSED` status. After certain period (no longer than 6 months), from time of merging PR in `PROPOSED` status to the main / master branch, we will
+update the proposal to be in `ACTIVE` state.
```
KtorZ, line 601: This can only happen if the proposal meets its acceptance criteria though.

CIP-0072/README.md
```diff
@@ -0,0 +1,613 @@
+---
+CIP: 72
+Title: Cardano dApp Registration & Discovery
+Status: Proposed
+Category: Metadata
+Authors: 
+  - Bruno Martins <bruno.martins@iohk.io>
+  - Mateusz Czeladka <mateusz.czeladka@cardanofoundation.org>
+Implementors: []
```
KtorZ, line 9: Given that there's an implementation plan, this shouldn't be empty.

rphair:
@ehanoc the `Last Check` label that I put on after the CIP meeting discussion 3 weeks ago was pending a response to other editors' direct review, which still hasn't happened yet.  We're still lacking your response to these 2 reviews, and some of these issues are definitely blocks to merging (e.g. "IOG and CF approval" as an acceptance criteria):
- https://github.com/cardano-foundation/CIPs/pull/355#discussion_r1209814169
- https://github.com/cardano-foundation/CIPs/pull/355#pullrequestreview-1429004121

These points are probably addressable within the hour; perhaps you can make it to the meeting and, if they've been fixed before then, perhaps we can merge it.  Otherwise I will suggest we label this `Waiting for Author` until you can respond to these reviews. 🙏

rphair:
@matiwinnetou thanks for the discussion in the last hour's CIP editors' meeting, after which we have agreed to leave on the `Last Check` label with the intention of merging at the next biweekly meeting, once all the review comments above are resolved especially the updates you are making to `Path to Active` (cc @ehanoc).

We also noted that this question might be addressed by leaving the list of implementors empty, but also that this could negatively impact adoption of the standard (this is a common difficulty that affects a number of CIPs).  If you need more time to work out that detail please let us know.

CIP-0072/README.md
matiwinnetou: They didn't, I removed them.

CIP-0072/README.md
matiwinnetou: done

CIP-0072/README.md
matiwinnetou: makes sense, changed

CIP-0072/README.md
matiwinnetou: done

matiwinnetou:
@rphair @rphair - I hope this is it.. I added changed according to suggestions...

KtorZ (approved):
Thanks for the prompt update @matiwinnetou 🙏

PR #56, opened 2020-12-17
CIP-1853 - HD Stake Pool Cold Keys

refi93:
Creating this PR to have a wider discussion and standardize how should the stake pool cold keys be deterministically derived, to e.g. allow their management on hardware wallets

Not sure if right now there are other, possibly not yet published, CIPs that are meant to have this number as a follow-up of CIP-1852, but if they don't introduce a new derivation path purpose, it IMHO makes sense to have 1853 for this CIP which does so

PR #362, opened 2022-10-25
CIP editor meeting #56's housekeeping

KtorZ:


rphair (approved):
very good... I was going to give "Smart Contract Licenses" a day or so to respond, but in hindsight that wasn't very efficient 😎

PR #366, opened 2022-10-26
(new) CIP-0001 & CIP-9999: Cardano Problem Statements

KtorZ:
This is a proposal to change CIP-0001 and rework how the CIP process currently function. This includes the introduction of a new type of document: Cardano Problem Statements (CPS); specified as part of CIP-9999 (read as _"CIP minus 1"_).

View [CIP-0001 Markdown](https://github.com/cardano-foundation/CIPs/blob/cip-cps-rework/CIP-0001/README.md)

View [CIP-9999 Markdown](https://github.com/cardano-foundation/CIPs/blob/cip-cps-rework/CIP-9999/README.md)

View [updated top level README](https://github.com/cardano-foundation/CIPs/blob/cip-cps-rework/README.md)


CIP-9999/README.md
```diff
@@ -0,0 +1,189 @@
...
+CIP: 9999
+Title: Cardano Problem Statements
+Status: Active
+Category: Meta
+Authors:
+    - Matthias Benkort <matthias.benkort@cardanofoundation.org>
+    - Michael Peyton Jones <michael.peyton-jones@iohk.io>
+Implementors: N/A
+Discussions:
+    - https://github.com/cardano-foundation/CIPs/pulls/365
```
jehrhardt, line 11:
```suggestion
    - https://github.com/cardano-foundation/CIPs/pull/366
```

It looks like this is the correct link to the pull request for discussion.

KtorZ: Indeed, I picked 365 in anticipation but there was a PR opened in the meantime!

rphair (approved):
Brilliant work all around 🤓

The proper lists (rather than comma-separated as before) of authors & discussions look much better in the header.  We should be on the lookout to be sure this different syntax also works when GitHub CIP `README.md` files are scraped onto `cips.cardano.org` and the Developer Portal.

I'm letting go of the comments I made before about having a template or example file visible on the front page for new visitors, writers, etc. to land on.  The links to each template are still difficult to spot, but people will find them if they know they're there.  The structure is also clear enough from the Specification > CIPs > [Structure](https://github.com/cardano-foundation/CIPs/blob/cip-cps-rework/CIP-0001/README.md#structure) section that finding the CIP template isn't vital.

@KtorZ to facilitate agenda preparation for meetings, and quick review of relevant PRs during the meetings, can we have GitHub PR labels corresponding to these [Categories](https://github.com/cardano-foundation/CIPs/blob/cip-cps-rework/CIP-0001/README.md#categories)?  Preferably with the same prefix so they group together (e.g. `Category: Meta` instead of just `Meta`)?

In any case I think we would have to add an `RSS` (Rewards Sharing Schemes) category to this list.  Regardless of how any institution decides to respond to these proposals, after over 2 years of RSS proposals there are now several of them which will repeatedly need to be seen as a group, and for which none of the other listed Categories are applicable.

Also, did you have a process in mind for people with CIPs who decide to convert their CIP into a CPS?  For example CIP-0050 could be a good candidate for this (though I would happily leave that decision entirely to the author).

CIP-9999/README.md
```diff
@@ -0,0 +1,189 @@
...
+    - Alice <alice@domain.org>
+    - Bob <bob@domain.org>
+Proposed Solutions: []
+Discussions:
+    - https://forum.cardano.org/t/solving-the-blockchain-trilemma/107720
+    - https://github.com/cardano-foundation/cips/pulls/9999
+Created: 2009-02-14
+---
+```
+</details>
```
michaelpj, line 85: same comment as before

CIP-9999/README.md
```diff
@@ -0,0 +1,189 @@
...
+
+<details>
+  <summary><span id="repository-organization"><strong>Repository Organization</strong><span></summary>
+
+A CPS must be stored in a specific folder named after its number and in a file called `README.md`. Before a number is assigned, use `????` as a placeholder name.
```
michaelpj, line 90: same comment about folder name

CIP-9999/README.md
```diff
@@ -0,0 +1,189 @@
...
+> **Note** A reference template is available in [.github/CPS-TEMPLATE.md][CPS-TEMPLATE.md]
+
+#### Statuses
+
+From its creation onwards, a problem statement evolves around the following statuses.
+
+Status       | Description
+---          | ---
+Open         | Any problem statement that is fully formulated but for which there still exists no solution that meets its goals. Problems that are only partially solved shall remain _'Open'_ and list proposed solutions so far in their header's preamble.
+Solved       | Problems for which one or more solutions have been found and implemented. When solved via a CIP, the solved status should indicate it as such: `Solved: by <CIP-XXXX>[,<CIP-YYYY>,...]`.
```
michaelpj, line 112: Given my previous comment about CPSs probably having multiple CIPs that "solve" parts of them, I wonder again about having something like "Addressed-by" instead. And perhaps the CPS could move to "Resolved" when it is completely addressed? Because that could also happen if it ceases to be a problem for some other reason.

CIP-9999/README.md
michaelpj: Or just say "same as CIP-001" here also

CIP-9999/README.md
```diff
@@ -0,0 +1,189 @@
...
+
+#### 2. Editors' role (same as CIP-0001)
+
+##### [2.a. Triage in bi-weekly meetings](https://cips.cardano.org/cips/cip1##triage-in-bi--weekly-meetings)
+
+##### [2.b. Reviews](https://cips.cardano.org/cips/cip1#reviews)
+
+#### 3. Merging CPSs in the repository
+
+A statement must be well-formulated (i.e. unambiguous) and demonstrate an existing problem (for which use cases exist with no suitable alternatives). When related to a current project, the problem statement must also have been acknowledged by its respective code owners. In some cases, problem statements may be written after the facts and be merged directly as _'Solved'_ should they document in more depth what motivated an existing solution.
```
michaelpj, line 137: "code owner" I think another new name for "project maintainer"?

CIP-9999/README.md
```diff
@@ -0,0 +1,189 @@
...
+
+Problem statements deemed unclear, for which alternatives exist with no significant drawbacks or establish unrealistic goals, shall be rejected (i.e. pull request closed without merge) with justifications or withdrawn by their authors.
```
michaelpj, line 139: Do we want to explicitly allow for the possibility of having multiple competing presentations of a problem? I guess we can leave it up to the editors.

michaelpj (commented):
Looking good!

CIP-0001/README.md
```diff
@@ -1,185 +1,374 @@
...
+Rationale      | The rationale fleshes out the specification by describing what motivated the design and what led to particular design decisions. It should describe alternate designs considered and related work. The rationale should provide evidence of consensus within the community and discuss significant objections or concerns raised during the discussion. <br/><br/>It must also explain how the proposal affects the backward compatibility of existing solutions when applicable. If the proposal responds to a [CPS][], the 'Rationale' section should explain how it addresses the CPS and answer any questions that the CPS poses for potential solutions.
+Path to Active | Organised in two sub-sections:<br/><h5>Acceptance Criteria</h5>Describes what are the acceptance criteria whereby a proposal becomes _'Active'_.<br/><h5>Implementation Plan</h5>A plan to meet those criteria. Or `N/A` if not applicable.
 Copyright      | The CIP must be explicitly licensed under acceptable copyright terms ([see below](#licensing)).
 
 <details>
```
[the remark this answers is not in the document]
rphair: I also thing the collapsible sections don't work as well as straightforward markup would.  It's clear enough from the header division: while collapsing the text under each header can make it seem like the following text is associated with that header (e.g. the links to the template files are easily missed because they appear under "Licensing").

More of a problem: the collapsed sections _aren't searchable_.  People will be looking for keywords in that collapsed text (e.g. header field names or ecosystem tags) and won't be able to find them unless those sections are expanded.

CIP-0001/README.md
```diff
@@ -1,185 +1,374 @@
...
 License: CC-BY-4.0
 ---
 ```
 
 </details>
 
+<details>
+  <summary><span id="repository-organization"><strong>Repository Organization</strong><span></summary>
+
+A CIP must be stored in a specific folder named after its number (4-digit, left-padded with `0`) and in a file called `README.md`. Before a number is assigned, use `????` as a placeholder name.
```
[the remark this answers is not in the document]
rphair: I've been asking authors to use a directory name of `CIP-` followed by a very abbreviated name of their proposal.

We should also tell submitters to provide a link to this file in the OP of the PR thread.  Any time we create the link for them (as we have after submission or during triage), we're inviting them to change the README pathname and have this link become invalid.  If they create it, generally they will maintain it.

The place to tell them to create this link is probably here, though; not in the text above: https://github.com/cardano-foundation/CIPs/blob/cip-cps-rework/CIP-0001/README.md#1a-authors-open-a-pull-request

CIP-0001/README.md
```diff
@@ -1,185 +1,374 @@
...
+Field          | Description
+---            | ---
+`CIP`          | The CIP number (without leading 0), or "\?" before being assigned
+`Title`        | A succinct and descriptive title
+`Status`       | Proposed \| Active \| Inactive (.._reason_..)
+`Category`     | One of the registered categories covering one area of the ecosystem.
+`Authors`      | A list of authors' real names and email addresses (e.g. John Doe <john.doe@email.domain>)
+`Implementors` | A list of implementors committed to delivering an implementation of the proposal, when applicable. `N/A` when not applicable and `[]` when there's currently no implementor.
+`Discussions`  | A list of links where major technical discussions regarding this CIP happened. Links should include any discussion before submission, and _must_ include a link to the pull request that created the CIP and any pull request that modifies it.
+`Solution-To`  | A list of [CPS][] that this CIP addresses, if any. Omitted when not applicable.
```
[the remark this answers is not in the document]
KtorZ: I originally put addresses there, and thought it might be confusing with the other meaning of addresses in our space. People might misinterpret that as a kind of donation section or whatever. I preferred to avoid the term altogether.

CIP-0001/README.md
```diff
@@ -1,185 +1,374 @@
...
+##### Status: Proposed
+
+A _'Proposed'_ CIP is any CIP that meets the essential CIP criteria but is not yet _'Active'_. The criteria that must meet a CIP to be merged as _'Proposed'_ are:
+
+- It must contain all the sections described in [Structure](#structure).
+- The quality of the content must be to the Editors’ satisfaction. That means it must be grammatically sound, well-articulated and demonstrates noticeable efforts in terms of completeness and level of detail.
+- Its technical soundness should have been established. Where necessary, this may require review by particular experts and addressing their concerns. Note that the requirement is that the proposal makes sense (i.e. be technically sound), yet no consulted experts need to think it is a good idea.
+- It must have a valid _'Path to Active'_. This particular section must be subdivided into two sub-sections:
+
+  - _'Acceptance Criteria'_
```
michaelpj, line 146: Pondering this in the case of Plutus builtins in particular, and I was wondering about insisting that CIPs proposing Plutus builtins be required to include test cases, or more generally a test plan. Because often the proposer is the one who actually knows what it's supposed to do, so they should include them here.

That made me wonder whether it's something that we want to include in _any_ CIP: the acceptance criteria SHOULD include test cases that must pass, and the implementation plan SHOULD include a testing plan.

WDYT?

CIP-0001/README.md
```diff
@@ -1,185 +1,374 @@
...
 Status: Active
-Type: Process
-Created: 2020-03-21
+Category: Meta
+Authors:
+    - Matthias Benkort <matthias.benkort@cardanofoundation.org>
+    - Michael Peyton Jones <michael.peyton-jones@iohk.io>
+Implementors: N/A
+Discussions:
+    - https://github.com/cardano-foundation/cips/pulls/365
```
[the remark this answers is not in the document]
crptmppt: having been the one who put CIP-0001 together and shouldered much of the initial weight of the CIP framework, I want to make sure I remain listed as author, thanks!

crptmppt (dismissed):
as noted in the comment, changing the authors of a CIP is not appropriate. I kindly requests the original authors be kept and that Michael be added as alternative.

rphair:
Despite practically every sentence being changed in the last draft, I would still support retaining @crptmppt (and maybe others) as an author in the header because of the practical work that went into establishing and documenting the original process.

I don't see how the current evolution of the CIP process which @KtorZ has documented much more rigorously today could never have happened except as experimental revisions from the old process: both in practice and in the earlier revisions of the document.

The paragraphs & sections in common between old & new documents may have changed completely in text, but they still serve the same purpose as the original.  Having expressed those concepts in the first place is just as much a part of this living document as the current text itself.

CIP-0001/README.md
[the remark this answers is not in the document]
crptmppt: +1 for removing the collapsing sections, prefer the entire document scrolleable

.github/CIP-TEMPLATE.md
```diff
@@ -1,26 +1,44 @@
 ---
 CIP: ?
 Title: ?
-Authors: John Doe <john.doe@email.domain>
-Status: Draft
-Type: Core | Process | Informational
+Category: ?
+Status: Proposd
```
crptmppt, line 5: "Proposed"

CIP-0001/README.md
```diff
@@ -1,185 +1,374 @@
...
+Created: 2022-10-14
```
crptmppt, line 12: While this expands on the original CIP, I think the date should be that of the initial proposal here. It might be confusing to some to see dates all over the place since the process is functionally the same

CIP-0001/README.md
```diff
@@ -1,185 +1,374 @@
...
+Plutus     | Changes or additions to Plutus, following the process described in [CIP-0035][]
+
+
+#### Project Engagement With The CIP Process
+
+In this context, we define _a project_ as a git repository owned by a team of maintainers. Besides, projects of the Cardano ecosystem that intend to follow the CIP process must explicitly enlist themselves and commit to the following:
 
-### Statuses
+- a) allocating time to **review** proposals from actors of the community when solicited by editors (i.e. after one first round of reviews);
+- b) defining additional rules and processes whereby external actors can engage with their project;
```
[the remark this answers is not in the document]
KtorZ: I do. Will clarify.

CIP-0001/README.md
```diff
@@ -1,185 +1,374 @@
...
+
+Discussions and comments shall mainly happen on Github in pull requests. When discussed on other mediums, we expect authors or participants to report back a summary of their discussions to the original pull request to keep track of the most critical conversations in a written form and all in one place.
+
+A dedicated Discord channel may also be created for some long-running discussions to support quick chats and progress on particular topics (while still being regularly summarised on the repository).
+
+As much as possible, commenters/reviewers shall remain unbiased in their judgement and assess proposals in good faith. Authors have the right to reject comments or feedback but are strongly encouraged to address concerns in their 'Rationale' section. Ultimately, CIP editors shall make the last call concerning the various statements made on a proposal and their treatment by the author(s).
+
+By opening pull requests or posting comments, commenters and authors agree to our [Code of Conduct][CoC]. Any comment infringing this code of conduct shall be removed or altered without prior notice.
```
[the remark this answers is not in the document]
KtorZ: Had to be in some places, thought it was the least worse.

CIP-9999/README.md
KtorZ: There already is a `Proposed Solutions` field in the preamble. Having another `Addressed-By` field feels redundant since, any `Proposed Solutions` in an `Active` would be effectively addressing (even partially) the problem. If you're suggesting to rename `Proposed Solutions` to `Addressed-By`, that's an option. I have no strong feeling about naming here.

rphair:
I'm still happy to go ahead with this, especially now that https://github.com/cardano-foundation/CIPs/pull/366#pullrequestreview-1170652027 has been tentatively resolved by restoring @crptmppt  to the author list.  I am guessing the red flag might still remain because @SebastienGllmt @dcoutts  haven't been similarly restored?

If this is still a potential blocker, I saw on [EIP-1](https://eips.ethereum.org/EIPS/eip-1) the [EIP Editors](https://eips.ethereum.org/EIPS/eip-1#eip-editors) section has a list of "Emeritus EIP editors" to indicate substantial prior contributions.  Maybe @KtorZ that would work for CIP-0001 (e.g. "Emeritus authors") if that's a way to go forward that places everyone where they want to be.  This could work for other evolving CIPs as well.

CIP-9999/README.md
michaelpj: I guess it just seems weird to me to put the CPS in the Solved state if only part of the problem has been addressed. That seems misleading. But then, leaving it in Open when it has been partially solved is perhaps also misleading. So I don't know what to do.

CIP-0001/README.md
michaelpj: I guess more generally I'm thinking that Acceptance Criteria usually involve some expectations of behaviour change, and it's not unreasonable to expect the proposer to at least broadly sketch out the expected behaviour changes.

e.g. for reference inputs we could have asked me to provide some transaction submission scenarios that would now work. That then is very helpful for generating test cases.

michaelpj:
Personally, I lean towards the liberal attitude towards authorship: offering credit and thanks is cheap, so the bar for doing it should be low. IMO it costs us nothing to have everyone who ever made a significant contribution to the document listed as an author.

KtorZ:
Not sure why everyone keeps arguing about authorship, as I've said, I have no particular wish to _remove_ authors. The document was simply written from scratch and I "scrapped" the old one, with its authors. The rationale being that I don't want to put words in other people's mouth; and none of the previous CIP-0001 authors have contributed to this new document. 

@crptmppt indicated he would like to be preserved in the authors list for the original work on CIP-0001, so I added him back. I will add back @dcoutts and @SebastienGllmt on this new document if they wish to.

CIP-9999/README.md
KtorZ: Then I need to clarify something because, `Solved` is only assigned to problems that are **fully** solved. A problem that is only partially solved remains in `Open`, but is updated accordingly to show what's left to solve. 

As @abailly-iohk also hinted once, it's probably better to have CPS be focused on a single "atomic" problem as much as possible to avoid this type of situation.

michaelpj: Well, the text says just:

> Problems for which one or more solutions have been found and implemented. 

Which certainly has left me confused about the partial solution state.

> As @abailly-iohk also hinted once, it's probably better to have CPS be focused on a single "atomic" problem as much as possible to avoid this type of situation.

I don't think I agree with this. I think it's very unlikely that the "problem view" lines up with the "solution view" so neatly. In practice we often need to do many things to fully address a user-visible problem!

abailly-iohk: The important part is 
> as much as possible
Of course, real life is messy and problems do not usually align neatly with solutions, but we should encourage definitions of problems that are actionable and "compact", eg. instead of a CPS trying to "Solve world hunger", write a CPS saying "How to salvage France's hypermarkets food trashed?" 
Not sure this is a great example but I guess the idea is there

ccgarant:
**CIP/CPS part numbering scheme** 
Since this is a new revision, just something to maybe consider. I haven't seen this done in crypto land, but in engineering land we have easily identifiable drawing and part numbers, like XXX-YYY-ZZZ-AAA format.
E.g. 901-301-001-101
 - 901 is the platform
 - 301 is a hydraulic part
 - 001 is the specific part
 - 101 is the revision number (e.g. 103 is the next rev)

Maybe a new, helpful, easy to visually check numbering scheme could be:
- CIP-META-0021B, where B is rev 2
- CPS-WALLET-0003A
- MBO-RSS-0001D

This might cause renumbering/mapping of old CIPs which might be problematic, so maybe:
 - CIP-0068-W-1, where W is for Wallet, and 1 is for Rev 1? (or CIP-0068-WALLET-01, MBO-0001-RSS-02, etc.)

Just food for thought and consideration...

abailly-iohk:
I have a meta  question about CIPs: I am having a hard time distinguishing the _Motivation_ section from the _Rationale_ section. According to dictionaries:
> **Motivation**: a reason or reasons for acting or behaving in a particular way.
> **Rationale**: a set of reasons or a logical basis for a course of action or belief.

These seems pretty much synonymous to me :thinking_face:

abailly-iohk:
I am really having a hard time wrapping my head around what `Active` really means, and how to get there. I wonder if having some form of  state machine diagram in the body of the CIP document would not help?

KtorZ:
@abailly-iohk there used to be a state-machine diagram, but it's basically as simple as:

```mermaid
stateDiagram-v2
    [*] --> Proposed
    Proposed --> Active
    Proposed --> Inactive
```

Hence why I don't much see the value to be frank. Is there anything I could add to make the definition of what `Active` means clearer? We've iterated several times on that and I thought that last draft was pretty clear:

> An _'Active'_ CIP has taken effect according to the criteria defined in its _'Path to Active'_ section.
> What that means depends on the nature of the CIP, typically:
>
> - For CIPs that relate to particular projects or pieces of technology, it becomes _'Active'_ by being implemented and released;
> - For changes to the Cardano protocol, a CIP will become _'Active'_ by being live on the Cardano mainnet;
> - For ecosystem standards, it means gaining sufficient and visible adoption in the community.

abailly-iohk:
Yeah, I agree it's a bit silly... I don't know why but it cannot stick to my mind, perhaps because as you mention it depends on the actual CIP 🤔

KtorZ:
Perhaps we have focused too much on how to become active but should also explain what are the implications of being 'Active'.

'Active' means completed. It's kind of analogous to a release for a software. The proposal has matured, was challenged and has demonstrated to be solid enough to be followed. 

On the CIP process itself, it means that we would generally not substantially change any 'Active' CIP but instead, ask people to make a new one. While still in 'Proposed

KtorZ:
(sorry, pressed the 'Close with comment' button by accident).

abailly-iohk:
But it seems to me there's a missing step between `Proposed` and `Active` which should be something like `Approved` or `Accepted` or `Planned` or whatever. You can make a CIP which is structurally sound, well-written, is vouched by CIP editors, becomes `Proposed` but ultimately receives more attention, leads to prototypes, and is rejected because it's too complicated, or noone actually cares to implement this. 
Sorry, it seems I am reenacting an old debate here so feel free to ignore my comment and confusion.

CIP-9999/README.md
michaelpj: > "How to salvage France's hypermarkets food trashed?"

... which is hardly going to have one solution, no? :) I just think it's unrealistic to suppose that every CPS will have one solution. In fact, my bet would be that that's the less common case.

abailly-iohk: That a problem stated in a CPS has more than one solution is certainly desirable and that's actually what I was, awkwardly, trying to express.

abailly-iohk: The idea I was trying to convey is that a CPS should express a problem that has  _at least_ one solution which itself could be expressed as a CIP. We want to rule out problems whose solution required something that would not fit in a single CIP (I think).

KtorZ:
@abailly-iohk -> I have slightly reworded the 'Status: Active' section as such: 

---

##### Status: Active

An _'Active'_ CIP has taken effect according to the criteria defined in its _'Path to Active'_ section. Said differently, each CIP defines by which criterias it can become _'Active'_ and those criteria are included in the review process. Exact criteria thereby depends on the nature of the CIP, typically:

- For CIPs that relate to particular projects or pieces of technology, it becomes _'Active'_ by being implemented and released;
- For changes to the Cardano protocol, a CIP becomes _'Active'_ by being live on the Cardano mainnet;
- For ecosystem standards, it means gaining sufficient and visible adoption in the community.

A proposal that is _'Active'_ is considered complete and is synonymous with "production readiness" when it comes to the maturity of a solution. _'Active'_ CIPs will not be updated substantially (apart from minor edits, proofreading and added precisions). They can, nevertheless, be challenged through new proposals if need be.

rphair:
> @abailly-iohk ... distinguishing the Motivation section from the Rationale section...
>
>  - Motivation: a reason or reasons for acting or behaving in a particular way.
>  - Rationale: a set of reasons or a logical basis for a course of action or belief.
>
> These seems pretty much synonymous to me...

They are: but only without respect to the progression of _time_.  In our process, the motivation _precedes_ the evolution of the proposal: so whatever past set of conditions that proposal hopes to improve will otherwise remain in force.  _After_ we establish that we're doing something about it (as outlined in the proposal), the rationale describes what we'd be doing in the present & future.

Also: in general they do mean the same thing if they are _about_ the same thing.  But in the case of CIP's, the "motivation" is for the proposal itself, while the "rationale" is for _the method described in_ the proposal.

crptmppt (approved):
I changed the authors list to the original authors + added MPJ (for the changes going into 0001). As the CIP process hardly changes beside the CPS addition I feel it is fine to merge with the added change. If there are issues regarding the addended authorship on this PR, I suggest this be submitted as a different CIP #, which can override 0001 (this could then be noted in the body of 0001) - "convenience" should not override facts, having created 0001 and the entire original CIP process, I request original authorship of 0001 be noted in the PR.  Other CIP fine as-is.

PR #367, opened 2022-10-28
Add more optional fields. Update state to Active

AndrewWestberg:


CIP-0060/cddl/version-1.cddl
```diff
@@ -97,6 +101,11 @@ metadata_details =
     ? parental_advisory: string,
     ? explicit: bool,
     ? isrc: string,
+    ? isrc: string,
```
rphair, line 104: I'm guessing this line is duplicated because `isrc:` was already there before the whole group was recently added?

rphair (commented):
fixes a typo as well, I can see 😎

CIP-0060/cddl/version-1.cddl
AndrewWestberg: Nope. this is wrong. I'll fix...

AndrewWestberg: fixed

rphair: OK, I see what happened 👍

rphair (approved):
These are industry standard tags so this addition seems in order.

@KtorZ it seemed from the original discussion that this was about to go `Active` - if not active already - but still is it recommended that active CIP's mention some example implementations that are currently live?

KtorZ:
Because we missed it when accepting the proposal, @AndrewWestberg may you clarify the acceptance criteria by which we now consider this proposal active? I reckon it is adoption by a certain number of parties / projects; This would need to go as a `Path to active` top-level section under `Rationale` (cf: https://github.com/cardano-foundation/CIPs/blob/master/.github/CIP-TEMPLATE.md). 

A small list of checked-checkbox will do 🙏

AndrewWestberg:
@KtorZ I've added the `Path to active` section.

KtorZ (approved):

PR #381, opened 2022-11-18
CIP-0081? | Tiered Pricing Protocol

abailly-iohk:
This CIP introduces a _Tiered Pricing_ mechanism to Cardano network as a solution to manage network congestions.
While related to the [Ouroboros Leios](https://github.com/cardano-foundation/CIPs/pull/379) proposal, its implementation can happen separately from it. 

Here is the [rendered CIP](https://github.com/abailly-iohk/CIPs/tree/tiered-pricing-protocol/CIP-XXXX).


CIP-XXXX/README.md
```diff
@@ -0,0 +1,125 @@
...
+
+More specifically, tiers are introduced and modified to achieve at minimum a target ratio between consecutive prices and delays; moving from tier `i` to tier `i+1` both the price must be substantially lower and the delay higher than that of the previous tier. The first tier is always available and its delay is set to the minimum level.
+Additional tiers are introduced, if the demand on the last (slowest) tier increases, i.e., its price becomes high enough. Similarly, if the demand of the last tier falls below a certain level, the tier gets deleted and other tiers are resized accordingly, to avoid leaving the allocated space unused.
+
+The price of each tier is updated in similar fashion to [EIP-1559](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-1559.md), disregarding other tier prices.
```
L-as, line 40: Please specify what this means _precisely_.

L-as: i.e., don't rely on the EIP

L-as (changes requested):

CIP-XXXX/README.md
```diff
@@ -0,0 +1,125 @@
...
+
+
+### Are we departing from a low-cost system?
+While this proposal departs from the low fixed fees approach for the reasons explained earlier, by appropriately setting the relevant parameters it can be guaranteed that a relatively low-cost service option will always be available to users.
+This option may come with a high expected delay when the system is congested. Note, that this is also the case in the current system. Moreover, tiered pricing improves in that it offers users a clear view of the delay expected from each tier, compared to the current system where the expected delay can only be estimated by off-chain channels.
```
michaelpj, line 63: > it offers users a clear view of the delay expected from each tier

Is this true? Can we guarantee that the expected delays for a tier won't be exceeded under conditions of high congestion?

CIP-XXXX/README.md
```diff
@@ -0,0 +1,125 @@
...
+
+
+
+### Why not EIP-1559?
+While our approach bares similarities to that of EIP-1559 regarding the way prices are updated, our design is a lot more diverse in that it allows different types of use-cases to be served by the system in a satisfactory manner. We highlight this point further through simulation.
```
michaelpj, line 68: This makes it sound like you have a precise formulation of how the tiers work and when they change... could we include that in the CIP? :) It's impossible to know what these simulations are telling us if we don't know what they're doing!

michael-liesenfelt:
A governance system may collectively decide to raise the lowest price tier to accumulate more treasury spending power. To protect the end-users it may be wise to limit the lowest price tier to some reasonably small fraction of the minimum UTXO parameter.

rphair:
@abailly our agreement at the CIP meeting today is to put this up for review... in the meantime can you please respond to some of the reviews & comments above so we have the basis for a well rounded review? 🙏

abailly-iohk:
> @abailly-iohk our agreement at the CIP meeting today is to put this up for review... in the meantime can you please respond to some of the reviews & comments above so we have the basis for a well rounded review? 🙏

I pinged @plazos who is the original author, and we'll make sure to address your comments.

CIP-XXXX/README.md
```diff
@@ -0,0 +1,125 @@
+---
+CIP: ????
```
```suggestion
CIP: 81
```

CIP-XXXX/README.md
```diff
@@ -0,0 +1,125 @@
...
+Unfortunately such an approach is ill-suited to handle traffic congestion, as it does not provide any means for users to signify their urgency and accommodate them based on their needs. Throughput scaling solutions, e.g., Ouroboros Leios, may only temporarily solve the problem, until peak traffic grows larger than the available throughput. Traffic congestion is also a possible attack vector that malicious actors may try to exploit to increase the average delay of the system at a moderate cost. Thus, Cardano needs a better way to handle traffic congestion.
+
+Ideally, we would like the system to offer a multitude of options, and have users decide the price/delay trafe-off that suits them. These options should change dynamically to reflect current traffic levels. We would like to have a mechanism that informs users of the current congestion status, and takes into account their preferences in prioritizing transactions.
+
+
+## Specification <!-- The technical specification should describe the syntax and semantics of any new feature. The specification should be detailed enough to allow competing, interoperable implementations. -->
```
SebastienGllmt, line 30: I think one of the biggest challenges that doesn't seem to be mentioned this CIP is how to implement this without hurting determinism.

Notably, in Cardano now (before this CIP), when executing a Plutus script you specify a fee for the script execution. The way Cardano is designed, if you transaction passes phase 1 validation, then the fee for the script should be deterministic and so you should never run out of gas in the real world.

However, if you have a dynamic fee, it's possible that a transaction was created when the fee was low, then the transaction runs on chain when the fee is higher, causing it to run out of gas in a phase-2 validation

Tiered fees could be a way around this issue if you define a tier as having a fixed price and if the tier disappears when the price changes, as it would allow the phase 1 validation to detect this tx was generated for a tier that no longer exists and fail-fast. However, figuring out the exact specifics of how that would work is I feel the main challenge a fee system CIP has to solve

CIP-XXXX/README.md
```diff
@@ -0,0 +1,125 @@
...
+Additional tiers are introduced, if the demand on the last (slowest) tier increases, i.e., its price becomes high enough. Similarly, if the demand of the last tier falls below a certain level, the tier gets deleted and other tiers are resized accordingly, to avoid leaving the allocated space unused.
+
+The price of each tier is updated in similar fashion to [EIP-1559](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-1559.md), disregarding other tier prices.
+On the other hand, delays are updated much less frequently than prices and depend on them. In particular, delays observe the average prices between each update and adjust up or down in small steps accordingly, to ensure that prices of consequent tiers are separated enough. Finally, tier additions and deletions happen even less frequently.
+
+Transactions are allowed to specify higher fees than those determined by the tier selected. In the end, they are only going to pay the actual tier price, and get back the change as a reward at the end of the epoch. The reward mechanism should be adjusted accordingly.
```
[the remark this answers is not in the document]
plazos: We did not get into such details for this CIP, because our plan was to involve the community with discussing the  overarching principles and concepts of the tiered approach design overall. The simplest approach however would be to reimburse users according to their staking key.

CIP-XXXX/README.md
plazos: Indeed, we have a precise formulation that was used for the simulations. We were concerned that it contains too many details that would be distracting in the discussion (which we thought it should focus on the merits as well as any potential downsides we haven’t thought about but the community can identify concerning the tiered pricing philosophy).

Since you think it is helpful, I will share that bit of the code, specifically how prices, delays and tier sizes are updated given the past.

CIP-XXXX/README.md
plazos: The expected delays, yes. However, it is possible (e.g. after a sharp and sustained increase in demand) that these delays are exceeded or that the transactions never make it on-chain without the user paying a higher fee (or selecting another tier). Unfortunately this is true of any design aiming to satisfy some notion of ease of use from, honest behaviour from SPO's and immunity to user-SPO collusion.

CIP-XXXX/README.md
plazos: I can add this to the CIP.

Roughly, for every tier, we would measure how ‘full’ the input blocks where over a small time period and compare it to a target threshold T. This is slightly more complicated because input blocks can overlap, but calling S the fraction of the throughput that was actually used, the new price for tier i would be p_i^{t+1} = p_i^{t} * 2^(a(S - T)/T), for a > 0. The higher a is, the more aggressive the update.

CIP-XXXX/README.md
L-as: Or highest priority tier is chosen for the fee given automatically?

CIP-XXXX/README.md
michaelpj: > We were concerned that it contains too many details that would be distracting in the discussion

I think there's a meta-point here. I tend to assume that CIPs are _actual, concrete proposals_, and should be expected to contain all the details. I certainly agree there's a place for higher-level discussion of ideas, but maybe that's not a CIP? Perhaps @KtorZ has thoughts?

But at any rate, if you're going to do that it's weird to include arguments like this that rest on the specific details without actually saying what they are! This argument that you're better than EIP-1559 rests on the _specific_ proposal (which you're not telling us), and the simulation of it. I just found it very confusing.

CIP-XXXX/README.md
michaelpj: This is essentially why the per-script budgets are there. They act as a firewall between the script-execution part and the rest of transaction validation.

CIP-XXXX/README.md
plazos: I apologise for the confusion, we may have had the wrong impression about the level of detail required for a CIP. I added some pseudocode to show exactly how prices, delays and tiers would be updated under this proposal. This is the same algorithm used in the simulations.

CIP-XXXX/README.md
```diff
@@ -0,0 +1,125 @@
...
+### Acceptance Criteria <!-- Describes what are the acceptance criteria whereby a proposal becomes 'Active' -->
+
+* Tiered Pricing is available on Cardano mainnet
+
+### Implementation Plan <!-- A plan to meet those criteria. Or `N/A` if not applicable. -->
+
+While Tiered Pricing will have a deeper impact when deployed on top of Ouroboros Leios, due to the high level of throughput that can be directed to different use-cases, Cardano running Ouroboros Praos could still benefit from it during high congestion periods. A Tiered Pricing implementation should thus be as much as possible agnostic of the actual protocol run.
+
+Should this CIP be accepted, the high-level implementation plan would be:
+1. Publish a detailed pricing algorithm
```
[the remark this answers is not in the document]
plazos: I added this to the CIP. The algorithm contains details and corner cases which we should discuss, but let's not lose focus of our main point: is having different levels of service for different prices a better alternative than current congestion control solutions?

kevinhammond:
A few quick points/questions. Some of these might arise because I haven't studied the text in enough detail?  Some comparison with standard queuing solutions would be helpful.  I'd expect it to be a fairly common problem for distributed systems.

1. It may be necessary to have a low-delay but low-priced tier to allow prioritisation of certain kinds of transactions (eg system-level ones).  This is a fairly common requirement.  The proposal mentions different types of block, but doesn't say how these are distinguished - that mechanism might work for this purpose?

2. Separating this - what are the different types of block/transaction that need to be separated and what are their properties?   This will help drive the requirements.  I think some of Michael's comments might be related to this - concretisation would help.

3. The proposed mechanism seems to suggest increasing tiers in a uniform way (specific gaps between the tiers) - it's not obvious to me that this is the most effective approach - you might want very low cost tiers with potentially long delays, for example, so I'd expect non-uniform gaps to be more useful.  It would be useful to model these alternatives against real/potential traffic pattens.

4.  The way the delay mechanism worked wasn't completely clear to me - generally, we can't guarantee specific delays (even long ones!) - you might get reasonable certainty if you know all the in-flight transactions, but there will still be some variations.   This might be because I haven't read the proposal clearly enough, but is the idea that some delay is always imposed (which might lead to periods when the chain is unused), that there is a maximum guaranteed delay for each tier (which could not be promised in general), or something else?

5.   Picking up on this - as defined, is it necessary to reserve some chain slots for each tier?  That would lead to inefficient utilisation if the demand was low, of course.  Basically, the equivalent of leaving first class seats unoccupied even if passengers are standing in second class. 

6. Are you expecting the delay to be eg linked with the timeout mechanism so that transactions have some window during which they will definitely be enacted (i.e. this transaction happens no sooner than X and no later than Y) - again, hard to achieve with real confidence.

7.  The speed of change may be an issue - you want the system to react sufficiently quickly to be effective as a throttling mechanism (also so that people can't block the chain with junk), but slowly enough to be predictable.  One problem, though is that you can't predict transaction arrival reliably - you can only observe the current chain load and the local transaction queue [mempool], and perhaps historic usage patterns

I need to check through the pseudocode.  Generally, I'd prefer a more declarative specification if that was possible - it can be hard to know how a system will behave if you need to work through timing/state concerns.  I'd prefer to see something that said "under these conditions, the system will behave in the following way" - that would also make it easier to verify an implementation

rphair:
@abailly there was a lot of press about this & a blog entry from IOG about this roughly 2 years ago.  Is this effort still under way?  Does it require a protocol overhaul to proceed?  Or is it already integrated somehow into a future protocol revision?

Without status updates or even any commentary in >18 months we would have to consider at least the _document_ abandoned even if the tiered pricing effort is not.  If you could please give us some kind of road map with community expectations about how & when this CIP could be qualified then I could scale this back to `Waiting for Author` rather than `Likely Abandoned`.

PR #382, opened 2022-11-20
CPS-0003? | Smart tokens

szist:
The CPS standard is coming from this proposal: https://github.com/cardano-foundation/CIPs/tree/master/CIP-9999

The topic of smart tokens at this stage is not ready for a CIP, so I deemed a CPS to be a better fit for discussions. Since the CPS standard is not set yet and also this topic would require proper discussions potentially elsewhere, I also made this as a DRAFT.

The need to extend the functionality around tokens and their utility was raised in the Interoperability working group - https://github.com/input-output-hk/Interoperability-working-group/issues/16

Disclaimer:
I don't think I'm a good fit to define the full requirements for what smart-tokens should fulfill, neither have I the overview of the development direction of Cardano. I also don't have the connections to push this forward. I think experts in regulations, compliance definitely should have a strong say, with dapps and wallets defining how they would work from the UX perspective. My aim here is simply to raise the topic.

---

([rendered proposal in branch](https://github.com/zoeldevapps/cardano-CIPs/tree/cps/smart-tokens/CPS-%3F%3F%3F%3F))

### DIFF
--- a/CPS-????/README.md
+++ b/CPS-????/README.md
@@ -1,0 +1,141 @@
+---
+CPS: ?
+Title: Smart Tokens
+Status: Open
+Category: ?
+Authors:
+    - Istvan Szentandrasi <szist@zoeldev.com>
+Proposed Solutions: []
+Discussions:
+    - https://github.com/cardano-foundation/cps/pulls/?
+Created: 2022-11-20
+---
+
+# DRAFT - smart tokens
+
+## Abstract
+
+The aim of this problem statement is to raise the issue around the ability to control a native token's lifecycle at more points in time.
+Currently only forging of the tokens can be controlled, but not interactions with them. Extending the blockchain to support other interactions is non-trivial and might require many tradeoffs.
+
+## Problem
+
+The Mary era introduced native tokens, that could be minted/burned using native scripts. These native scripts for example could allow defining basic conditions to allow forging like required signatures or time validity range. The native tokens are identified with `policyId` and `assetName`. The `policyId` identifies the "publisher" or "publishers" while `assetName` can be an arbitrary.
+
+With the Alonzo era, plutus scripts were introduced. They allowed extending the functionality of the native scripts with more generic logic scoped at the transaction level. The format and shape of native tokens didn't change - the `policyId` became the hash of the plutus scripts pushing the "publisher" role responsibility to smart contracts.
+
+Skipping ahead, although Babbage didn't add any additional functionality directly, the addition of reference inputs indirectly opened additional use-cases to the the token's lifecycle. To be specific an example: imagine an decentralized bridge with a wrapped token that gets dynamically minted or burned. As a safeguard they would introduce a global UTxO (e.g. with a datum including signature by a multisig wallet or validity token) controlled by recognized members or governance voting mechanism that includes an `enable` flag. The enable flag could be checked by the wrapped token minting policy through a reference input. When a governance vote would pass, the UTxO datum could be changed and the flag could be set to freeze the token.
+
+Even in the Babbage era, after the token is minted, there is no way to control the token after it leaves a smart-contract environment. Users can freely exchange tokens and transfer among each other. This allows cheaper, more efficient and more open ecosystem. In contrast with this, ERC-20 tokens are smart-contracts that fully control any interaction with a token.
+
+Similar to how UTxO addresses can be optionally fully controlled by smart-contracts, there could be potential also to optionally control other lifecycle points of a token than just forging on Cardano.
+
+## Use cases
+
+There are multiple use-cases that would benefit from having the ability to validate token interactions.
+
+From the **NFT space**, having royalties for tokens is currently a social contract where the publisher provides the royalty information while minting the tokens. The marketplaces can choose to honor this information and transfer part of the token sales towards the author. The problem is, this is not enforced and "black market" can thrive. As an author of an art NFT, I would like to benefit from any token transfers that involve smart contracts (a marketplace platform).
+
+From the **dapp-space**, consider validity and authorization tokens. For example, liquidity pools the liquidity provider token minting can be authorized by a validity token. This token _should_ be fully controlled by the liquidity pool smart contracts. If it were to leak out of the liquidity pool smart contract, it would give the ability to anyone to mint LP tokens and drain liquidity pools of the actual funds (as seen in the Minswap security vulnerability). IF the authorization token was itself controlled by a smart contract, even after it left the liquidity pool, the vulnerability could be avoided.
+
+Another use-case comes up around **bridged wrapped tokens**. Due to regulations, some publishers might require the ability to freeze or selectively allow specific flow directions for the token at times. These issues were raised also at the `interoperability working group`. Bridges bridging stables from Ethereum would like more control over the wrapped bridged tokens to be in-line with regulations and help with security.
+
+When it comes to regulations, also **native stable coins** are another case, where the publisher might want to control the ability to freeze or require authorization with interactions of their tokens.
+
+**TBD: add more examples**
+
+## Goals
+
+The first main goal is to **achieve a consensus** if having more control over the tokens is actually something that would be in-line with the values and direction of Cardano. There are multiple tradeoffs from the usability perspective. Having validated lifecycle for tokens would have an impact on all dapps and all wallets.
+
+The second main goal is **create a CIP** with the technical direction how smart tokens could function. Below is a potential direction and possible issues and roadblocks that would need to be addressed in the CIP.
+
+### Validating token transfers
+
+A direct way to support the use-cases above, would be to require a validator minting policy to be executed, whenever a token is included in a transaction. Currently plutus smart contract have 4 different usages:
+
+- SPEND - used by utxo validator scripts
+- MINT - used minting policy scripts
+- REWARD and CERTIFY - for staking use-cases
+
+UTxOs have basically 2 operations associated with them:
+
+- READ - by reference inputs
+- SPEND - whenever the utxo is an input
+
+From these only the SPEND is controlled by smart-contracts. Currently reference inputs cannot control the transactions they are in (which is a topic for another CPS).
+
+In comparison, the operations with tokens are 3 fold:
+
+- mint
+- burn
+- transfer
+
+The MINT plutus scripts control mint and burn, but not transfer. There are 2 possible directions how the transfer could be validated:
+
+1. Call the token policy with the `MINT`, also whenever a smart token is transferred. The minting policy then could decide based on the redeemer for how it should validate the transaction.
+2. Introduce a `TRANSFER` operation and a separate script type that would be used for this exact use-case.
+
+Both of these options would strain wallets and dapps, as they would need additional specific logic to handle smart tokens. The token validation scripts might not be generic enough and might be costly to do in every interaction. As the blockchain evolves, some interactions couldn't be predicted by the token and evolving the scripts would be problematic.
+
+A workaround might be, if some smart-token policies delegated the validation to some validations scripts. Unfortunately, without adding the ability to execute validation on referenced scripts, even this would require either: minting/forging a token in the transaction or spending an utxo with the desired validation logic. This approach doesn't seem scalable-enough, either.
+
+### Identifying smart tokens
+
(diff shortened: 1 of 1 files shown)


rphair (commented):
thanks @szist - I believe this is a valid subject for a CPS: between the high profile use cases and the reasons we might _not_ want to do this (centralisation, complexity, breaking changes from existing methods).  There is enough of an inventory of risks & benefits for the community itself to start processing this, after some ledger / core experts & NFT veterans can help refine it (cc @JaredCorduan @alessandrokonrad)

A similar discussion began immediately after this CIP was submitted (https://github.com/cardano-foundation/CIPs/pull/378), addressing your NFT use case.  You might incorporate some of this discussion in your CPS and/or contribute to that discussion.

I'm also going to follow what I believe is your intention and convert this officially to a GitHub draft; more information here: [GitHub Docs > Converting a pull request to a draft](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/proposing-changes-to-your-work-with-pull-requests/changing-the-stage-of-a-pull-request)

michaelpj:
You might be interested in https://github.com/cardano-foundation/CIPs/pull/336

L-as:
I am _very_ sure you don't want this. This entire issue stems from a conflation with what tokens are considered to be in Ethereum and Cardano. You **don't** want to use tokens for these use cases.

You can effectively implement ERC-20 on top of Cardano, in several ways:
- Single UTXO with Merkle Tree of (owner, amount) pairs.
- One UTXO per "coin" with helper token to authenticate coin. Datum would include amount, owner, etc.

"Smart tokens"/spending policies don't offer any benefit over the above in terms of fees, efficiency, simplicity, or similar.

CPS-????/README.md
```diff
@@ -0,0 +1,141 @@
...
+
+From the **dapp-space**, consider validity and authorization tokens. For example, liquidity pools the liquidity provider token minting can be authorized by a validity token. This token _should_ be fully controlled by the liquidity pool smart contracts. If it were to leak out of the liquidity pool smart contract, it would give the ability to anyone to mint LP tokens and drain liquidity pools of the actual funds (as seen in the Minswap security vulnerability). IF the authorization token was itself controlled by a smart contract, even after it left the liquidity pool, the vulnerability could be avoided.
+
+Another use-case comes up around **bridged wrapped tokens**. Due to regulations, some publishers might require the ability to freeze or selectively allow specific flow directions for the token at times. These issues were raised also at the `interoperability working group`. Bridges bridging stables from Ethereum would like more control over the wrapped bridged tokens to be in-line with regulations and help with security.
```
matiwinnetou, line 41: USDC use case?

CPS-????/README.md
```diff
@@ -0,0 +1,141 @@
...
+
+When it comes to regulations, also **native stable coins** are another case, where the publisher might want to control the ability to freeze or require authorization with interactions of their tokens.
+
+**TBD: add more examples**
+
+## Goals
+
+The first main goal is to **achieve a consensus** if having more control over the tokens is actually something that would be in-line with the values and direction of Cardano. There are multiple tradeoffs from the usability perspective. Having validated lifecycle for tokens would have an impact on all dapps and all wallets.
```
matiwinnetou, line 49: I personally think this is a good idea from the point of view that this is sometimes needed. They should be in addition to native assets because majority of cases do not require "smart tokens".

It needs to be said also that since they would be like kinda ERC-20 tokens, which means that they would be ultimately more expensive for the blockchain to handle. Reason we have native assets is because they are secure and cheap for the leader (kinda like colour coins back in the days).

matiwinnetou:
> I am _very_ sure you don't want this. This entire issue stems from a conflation with what tokens are considered to be in Ethereum and Cardano. You **don't** want to use tokens for these use cases.
> 
> You can effectively implement ERC-20 on top of Cardano, in several ways:
> 
> * Single UTXO with Merkle Tree of (owner, amount) pairs.
> * One UTXO per "coin" with helper token to authenticate coin. Datum would include amount, owner, etc.
> 
> "Smart tokens"/spending policies don't offer any benefit over the above in terms of fees, efficiency, simplicity, or similar.

It would be nice to have PoC for this / some examples we could play with.

szist:
> I am _very_ sure you don't want this. This entire issue stems from a conflation with what tokens are considered to be in Ethereum and Cardano. You **don't** want to use tokens for these use cases.
> 
> You can effectively implement ERC-20 on top of Cardano, in several ways:
> 
> * Single UTXO with Merkle Tree of (owner, amount) pairs.
> * One UTXO per "coin" with helper token to authenticate coin. Datum would include amount, owner, etc.
> 
> "Smart tokens"/spending policies don't offer any benefit over the above in terms of fees, efficiency, simplicity, or similar.

could you elaborate please? Depending on how exactly you want to implement this I see issues in every direction. If you have a single utxo, you run into spend/interaction conflicts and tx chaining won't solve for you. If you have try to start emulating a "blockchain with tokens" with coins on utxos, to me it seems you are just moving security risks to the authentication token. In either case you would need direct full smart contract support from wallets anyhow. So I don't see how this is more efficient or simpler.

L-as:
Supporting this in wallets is neuded anyway; wallets need to in general be able to show user funds locked by protocols.

Wrt. the single UTXO solution, transaction chaining can still solve this, wrt. the multi UTXO solution, I'm not sure what you mean about moving risks. It's waaaay simpler than any proposed solution I've seen for spending policies.

szist:
> Supporting this in wallets is neuded anyway; wallets need to in general be able to show user funds locked by protocols.
> 
> Wrt. the single UTXO solution, transaction chaining can still solve this, wrt. the multi UTXO solution, I'm not sure what you mean about moving risks. It's waaaay simpler than any proposed solution I've seen for spending policies.

Could you please actually elaborate :) rather than just refuting my points. If you have a single utxo and you want to do chaining with 100x people + smart contracts trying to interact with the same token at the same time, how are you doing transaction chaining? 

As for the auth coin. If the auth coin ever gets leaked, the token practically becomes worthless, as you could create a utxo with an auth token with arbitrary amount. This way you increase the surface of potential vulnerabilities from the validator script that would track the coins and datums, to also the liquidity policy that handles minting/burning the auth coins.

rphair:
Looking to the future about how we might ensure this subject is tracked properly: since [as said above](https://github.com/cardano-foundation/CIPs/pull/382#pullrequestreview-1187299482) this CPS shares considerations with #378... which is currently defined as a CIP but might be more useful as a CPS (https://github.com/cardano-foundation/CIPs/pull/378#issuecomment-1322152149)... could separate authors @szist and @MartinSchere post any ideas about combining these proposals into a single CPS?

rphair:
p.s. also since mentioned in https://github.com/cardano-foundation/CIPs/pull/378#issuecomment-1322336214 this should be related to https://github.com/cardano-foundation/CIPs/pull/336 (primarily in terms of reasons to _not_ do this... or at least some reasons to be careful about considering it)

CPS-????/README.md
```diff
@@ -0,0 +1,141 @@
+---
+CPS: ?
```
KtorZ, line 2:
```suggestion
CPS: 3
```

szist:
I will try to incorporate the comments and considerations from the enforceable royalties into this CPS till eow. I wasn't aware of that CIP before writing this CPS. Thanks KtorZ for the updates, i'll fix the version.

colll78:
Programmable Tokens is one of the worst components of the Ethereum ecosystem. It opens up a Pandora's box of issues that would be super hard to justify for the functionality that they offer.

approve() & transferFrom() are probably the two most desirable features introduced by Smart Tokens, but they are also immeasurably dangerous and have a hand in practically every one of the 100+ Ethereum exploits totaling to billions of dollars in damages. 

Freezing tokens also comes with its own set of issues and introduces either a lot of complexity (making sure entire UTXOs are not locked just because they have 1 programmable token in them, forcing the inclusion of some type of Spending policy in every single transaction involving UTXOs that contain smart tokens even if the smart tokens are just part of the change UTXO), or a lot of spam UTXOs (ie you cannot have programmable tokens in the same UTXO as ADA / native assets). 

If this is implemented I'd suggest it be a very very small subset of vetted features of programmable tokens to avoid these issues. 

"Supporting this in wallets is needed anyway; wallets need to in general be able to show user funds locked by protocols." 
This should be a priority for so many reasons outside of just supporting mock ERC tokens.

"Single UTXO with Merkle Tree of (owner, amount) pairs.
One UTXO per "coin" with helper token to authenticate coin. Datum would include amount, owner, etc."

I'm not particularly fond of either of those solutions. 

I propose the following solution:
Smart tokens are minted to a programmableTokenValidator that enforces that those tokens cannot be sent to anywhere outside the validator address. Each user possessing smart tokens has their own UTXOs locked at the smart contract containing their smart tokens with a datum identifying them as the owner. If Bob wants to send Alice 20 of his smartCoin, he simply spends his UTXO from the script to the script adjusting the datum accordingly.

programmableTokens

Then any arbitrary logic can be tagged onto the validator (thus giving us programmable tokens). For instance to add the ability to freeze tokens, we just enforce that the validator checks for the inclusion of a reference Input identified by some auth token managed by the creator of the programmable token. The reference input contains a datum with the field {freeze :: Bool}, if freeze = false the validator does not allow spending of the programmable token, if freeze = true spending is allowed.

michele-nuzzi:
I'm working on a CIP that proposes a solution with logarithmic complexity to add a new public key hash (account) and keeps the balance unrelated to the Merkle tree so that transactions with ERC20-like tokens can happen with constant complexity

CharityDA0:
Here is an example of token functionality requested for Cardano native tokens. Perhaps it can be included in the Smart Token proposal?

Title: CPS 101? - Cardano Improvement Proposal for Customizable Fees for Native Tokens

Abstract: CPS 101? proposes an innovative solution to introduce a customizable automated fee system for Cardano native tokens that can be easily executed by smart contracts on all exchanges. This proposal outlines two alternative methods for implementing the fee system - one utilizing a smart contract that controls the supply of tokens, and the other leveraging metadata of a special initial token during the mint.

Motivation: Automated fee systems have become increasingly popular in the cryptocurrency space as they incentivize token holders to hold onto their tokens and discourage frequent pump and dumps. They are also used to fund development and charitable giving. This functionality has been present on every major chain for years. However, Cardano’s current structure of native tokens makes it difficult to implement such fee systems, which rely on smart contracts to govern fees and distribution. This proposal seeks to introduce a method for implementing a similar automated fee system for Cardano native tokens, enabling greater flexibility and customization for token issuers.

Specification: Option 1 - Smart Contract-Based Fee System: To implement a smart contract-based fee system, we propose the following steps:

    Create a smart contract that controls the supply of tokens and governs the behavior of the token.

    Define the logic for deducting the fee from each transaction and distributing those fees to the appropriate parties within the smart contract.

    Deploy the smart contract to the Cardano blockchain and associate it with the native token.

    The smart contract will automatically execute and handle the fees and fee distribution whenever a transaction involving the token occurs.

Option 2 - Metadata-Based Fee System: To implement a metadata-based fee system, we propose the following steps:

    Mint a special initial token that includes metadata defining the fee structure for the native token, much like the royalties done with NFTs on CIP-27.

    When a user sends a transaction involving the native token, DEXs can easily implement the fee like done on NFT marketplaces.

    Deduct the fee from the transaction amount, and distribute it to the appropriate parties.

    Since the fee is specified in the metadata of the initial token, it does not require a smart contract to be associated with the native token.

Alternatively, the metadata could even point to a smart contract that handles the details automatically, making it more modular.

Rationale: By introducing a method for implementing a customizable automated fee system for Cardano native tokens, this proposal would enable greater flexibility and customization for token issuers. This proposal aligns with the overall goal of the Cardano ecosystem to provide a robust and flexible platform for building decentralized applications and tokens. Moreover, it will make it easier for communities to fund their projects democratically and incentivize token holders to hold onto their tokens. The proposal will make it possible for the Cardano ecosystem to attract a wide range of projects, from nation-states seeking to create their own digital currencies to charitable organizations wanting to fund their operations using the generated fees. The possibilities are endless, and the potential for innovation is vast.

szist:
As you might've deduced by now, I'm on a hiatus since December with no end in site. Life happens. Unless there is anybody else who would want to take up this proposal, @KtorZ feel free to close this.

rphair:
thanks for the update @szist: I've already referred a handful of people to this CPS and thread... if there is any potential interest in carrying on the work you have done, I (we editors) will certainly refer them here. 😎

GeorgeFlerovsky:
@colll78 

If you squint, your proposed solution is an application of CIP-68.

Issue a user NFT to the freezing authority.

Specify the token amount and current owner in the ref NFT.

Disallow changing the amount but allow the user NFT holder to freeze.

Current owner can assign new owner.

PR #383, opened 2022-11-24
CIP-1775? | Delegated Spending Authorizations

imikushin:
Allow users to delegate authorized spending of outputs in their wallets to smart contracts or other users, without sending said outputs out of their wallets.

This brings benefits to Cardano users when interacting with existing dapps and opens up new use cases, like noncustodial asset management, concurrent use of dapps and seamless integration with sidechains.

### DIFF
--- a/CIP-1775/README.md
+++ b/CIP-1775/README.md
@@ -15,26 +15,34 @@ License: CC-BY-4.0
 
 ## Abstract
 
-We propose to allow users to delegate authorized spending of outputs in their wallets to smart contracts or other users, without sending said outputs out of their wallets.
+We propose to allow users to delegate authorized spending of outputs in their wallets to smart contracts or other users,
+without sending said outputs out of their wallets.
 
 ## Motivation
 
-In order to use any dApp on Cardano (e.g. a decentralized exchange or an NFT marketplace), a user has to **deposit tokens** to a script address, thereby **giving up custody** of the tokens.
+In order to use any dApp on Cardano (e.g. a decentralized exchange or an NFT marketplace), a user has to **deposit
+tokens** to a script address, thereby **giving up custody** of the tokens.
 
 This can be problematic for a number of reasons.
 
-- The user loses control of the assets they send out to a smart-contract, unless the smart contract is specifically designed to allow for such control (e.g. giving a "Cancel trade" or "De-list NFT" button in the UI).
+- The user loses control of the assets they send out to a smart-contract, unless the smart contract is specifically
+  designed to allow for such control (e.g. providing a "Cancel trade" or "De-list NFT" button in the UI).
 - Assets sent out from a user's wallet:
-  - are no longer  **delegated** to the stake pool of the user's choice.
-  - cannot participate in **governance voting** on behalf of the user.
-- Any output, accidentally sent to a script address without a datum is effectively lost now (because of how transaction validation scripts are invoked).
-- Requiring users to deposit assets into a dapp may subject the dapp's owners to regulatory scrutiny.
+    - are no longer **delegated** to the stake pool of the user's choice, unless sent to a specially constructed
+      _hybrid_ address with the user's StakeKeyHash as the delegation part,
+    - cannot participate in **governance voting** on behalf of the user, unless hybrid addresses are used and the voting
+      system counts assets at these hybrid addresses based on the stake key hash.
+- Any output, accidentally sent to a script address without a datum is effectively lost now (because of how transaction
+  validation scripts are invoked).
+- Requiring users to deposit assets into a dapp may subject the dapp to regulatory scrutiny.
 
-We believe, this is not strictly necessary and would like a way to allow users to delegate spending authorization for some outputs **within** their wallets to other users (smart contracts, multisigs, wallet addresses).
+We believe, this is not strictly necessary and would like a way to allow users to delegate spending authorization for
+some outputs **within** their wallets to other users (smart contracts, multisigs, wallet addresses).
 
 ### Use Cases
 
-Instead of sending tokens to the dapp smart-contract, the user creates an output in their own wallet that delegates spending authorization to a smart-contract (or other entity).
+Instead of sending tokens to the dapp smart-contract, the user creates an output in their own wallet that delegates
+spending authorization to a smart-contract (or other entity).
 
 Some **existing** use cases, where this can be beneficial:
 
@@ -42,63 +50,107 @@ Some **existing** use cases, where this can be beneficial:
 2. Providing liquidity on a decentralized exchange.
 3. Listing an NFT on a marketplace.
 
-Some **New** use cases, enabled by this proposal:
+Some **new** use cases, enabled by this proposal:
 
-1. On-chain noncustodial asset management. A user can delegate managing (some of) their assets to a manager, another individual user or organization.
-2. Concurrent use of dapps. For example, a user can provide liquidity and, at the same time, have a trade open at a (different) DEX, with the same output. When the trade is executed, it looks just as though the user has withdrawn the output from the liquidity pool.
-3. Seamless interaction with L2 (sidechains or Hydra). The user doesn't need to send assets to the L2 entrypoint script/multisig, nor do they need to explicitly take them out.
+1. On-chain noncustodial asset management. A user can delegate managing (some of) their assets to a manager, another
+   individual user or organization.
+2. Concurrent use of dapps. For example, a user can provide liquidity and, at the same time, have a trade open at a (
+   different) DEX, with the same output. When the trade is executed, it looks just as though the user has withdrawn the
+   output from the liquidity pool.
+3. Seamless interaction with L2 (sidechains or Hydra). The user doesn't need to send assets to the L2 entrypoint
+   script/multisig, nor do they need to explicitly take them out.
 4. Combinations of all of the above, both existing and new 🙃.
+
+Note, that we are **not** proposing to eliminate DApps with full control of outputs. Traditional custodial ("deposit and
+pray") smart contracts are preserved in full force.
 
 ### Other Benefits
 
-- Users' assets continue to participate in delegated proof of stake and on-chain governance.
+- Users' assets continue to participate in delegated proof of stake and on-chain governance without special effort from
+  DApps or users.
 - Assets cannot be locked forever in noncustodial (delegated) smart-contracts, by definition.
 
 ## Specification
 
 ### Delegated Spending Authorizations
 
-A solution is to introduce _delegated spending authorizations_: instead of **sending** a output to an address, a user could **delegate** a spending authorization of the output to the address.
+A solution is to introduce _delegated spending authorizations_: instead of **sending** an output to an address, a user
+could **delegate** a spending authorization of the output to the address.
 
 ### CDDL
 
-To enable this, a **new** field is added to the transaction output (the current spec can be found here: https://github.com/input-output-hk/cardano-ledger/blob/master/eras/babbage/test-suite/cddl-files/babbage.cddl):
+To enable this, a **new** field is added to the transaction output (the current spec can be found
+here: https://github.com/input-output-hk/cardano-ledger/blob/master/eras/babbage/test-suite/cddl-files/babbage.cddl):
 
 ```cddl
 post_alonzo_transaction_output =
   { 0 : address
   ; ...
-  , ? 4 : { * address => ( datum_option / nil ) }   ; New; delegated spending authorizations
+  , ? 4 : { * address => ( datum_option / nil ) }   ; New; authorized spenders
   }
 ```
 
-The new field is a map from `address` to `datum_option` or `nil`. In this map, `address` can be any address: Plutus v1 and v2 scripts, wallets, etc. The value (`datum_option`) only makes sense for Plutus script addresses, so it is optional (can be `nil`).
+The new field is a map from `address` to `datum_option` or `nil`. In this map, `address` is a Shelley address (
+see [CIP-19](https://github.com/cardano-foundation/CIPs/tree/master/CIP-0019)). The value (`datum_option`) only makes
+sense for Plutus script addresses, so it is optional (can be `nil`).
 
 ### Spending of Outputs
 
-When the output's `address` (field 0) is a **wallet** address, a transaction can spend the output if it satisfies spending conditions:
-- for the output's address (i.e. signed by the wallet's private key), or 
+Delegated spending authorizations **only** work for:
+
+- wallets (Shelley addresses Type 0, 2, 4 or 6, according
+  to [CIP-19](https://github.com/cardano-foundation/CIPs/tree/master/CIP-0019)),
+- Plutus v3 (or higher) scripts (Shelley addresses Type 1, 3, 5 or 7, according
+  to [CIP-19](https://github.com/cardano-foundation/CIPs/tree/master/CIP-0019)).
+
+These addresses can:
+
+- own outputs with delegated spending authorizations,
+- be a delegated authorized spender of an output.
+
+A transaction can spend the output if it satisfies spending conditions:
+
+- for the output's owner address (i.g. signed by the wallet's private key), or
 - for any address in the output's _delegated spending authorizations_.
 
-If the output's `address` is of **any other kind** (e.g. native or Plutus script), the _delegated spending authorizations_ field contents **do not affect** spending behavior. Existing scripts' assumptions must be **unaffected** by the proposed ledger change: spending of outputs at a script address are controlled by the script only.
+Delegated spending authorizations are **ignored** for addresses of other kinds (both as output owners and delegated
+authorized spenders).
 
-With the above rules, we believe, the change does not require Plutus language version bump.
+### Script Context
+
+Plutus v3 (or higher) scripts can participate in delegated spending authorizations, both as owners of the output and as
+authorized spenders. The Plutus version bump is required because:
+
+- assumption of exclusive control of outputs should be preserved for prior script versions - to not break things,
+- authorized spenders need to be added to the script context - to make the script aware of its "competition".
 
 ## Implications
 
-While providing the benefits listed in the **Motivation** section, there is a certain cost: users need to be made aware of _delegated spending authorizations_ on the outputs in their wallets. 
+While providing the benefits listed in the **Motivation** section, there is a certain **cost**: users need to be made
+aware of _delegated spending authorizations_ on the outputs in their wallets.
 
-Assets that arrive into a user's wallet with delegated spending authorizations attached, kind of come with a form of remote control (which can be exploited). Obviously, the user can remove it, but **only** if they know of its existence.
+Assets arriving into a user's wallet with delegated spending authorizations attached, come with a form of remote
+control (which can be exploited). Obviously, the user can remove it, but **only** if they know it exists.
 
-Therefore, when this change is implemented, wallet UX **needs** to reflect it.
+Therefore, when this change is implemented, wallet UX **needs** to reflect it. Users should be warned when:
+
+- they are _about to sign_ a transaction resulting in UTXOs with delegated spending authorizations in their wallets,
+- have UTXOs with delegated spending authorizations attached (perhaps, having received them from a dapp or a sidechain).
 
 ## Related Work
 
-The idea for this CIP came about when thinking of ways to simplify user experience and reduce L1 transaction load when interacting with sidechains. 
+The idea for this CIP came about when thinking of ways to simplify user experience and reduce L1 transaction load when
+interacting with sidechains.
 
-To work with a sidechain, assets need to be deposited into its "gate script" -- a script, that will hold the assets for the duration of their usage in the sidechain. When the sidechain transactions are settled, the assets can be withdrawn by their new owner. Of course, when/if the new owner needs to use them on the sidechain, they need to send to the gate script again.
+To work with a sidechain, assets need to be deposited into its "gate script" -- a script, that will hold the assets for
+the duration of their usage in the sidechain. When the sidechain transactions are settled, the assets can be withdrawn
+by their new owner. Of course, when/if the new owner needs to use them on the sidechain, they need to send to the gate
+script again.
 
-These "deposit / withdraw" steps may seem unavoidable, but in fact, to some extent, they defeat the purpose of a sidechain (provide superior user experience, higher speed, lower cost, rely on L1 for security). For example, if a sidechain powers a decentralized CashApp equivalent, these extra "deposit / withdraw" steps would mean added load on the L1 and additional complexity for the user and the developer.
+These "deposit / withdraw" steps may seem unavoidable, but in fact, to some extent, they defeat the purpose of a
+sidechain (provide superior user experience, higher speed, lower cost, rely on L1 for security). For example, if a
+sidechain powers a decentralized CashApp equivalent, these extra "deposit / withdraw" steps would mean added load on the
+L1 and additional complexity for the user and the developer.
 
 _Delegated spending authorizations_ solve these problems:
 
--- a/CIP-1775/README.md
+++ b/CIP-1775/README.md
@@ -1,0 +1,108 @@
+---
+CIP: 1775
+Title: Delegated Spending Authorizations
+Authors: Ivan Mikushin <sorrel_cantos.0@icloud.com>
+Comments-URI:
+Status: Proposed
+Type: Standards Track
+Created: 2022-11-17
+License: CC-BY-4.0
+---
+
+# Delegated Spending Authorizations
+
+"Give me custody, or give me death!" - [@ArmySpies](https://twitter.com/ArmySpies/status/1592586212816355328)
+
+## Abstract
+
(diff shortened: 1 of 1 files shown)


L-as:
I get the motivation, but I don't think this is a good idea.
There are two unrelated issues here:
- Protocols some times don't respect staking. Those protocols are buggy. Don't use them.
- Wallets don't show where your funds are in custom protocols. This is harder, but essentially entirely a wallet issue. Wallets need to be fixed.

You could imagine there being some wallet-protocol interoperation ABI specification language, such that the wallet can provide limited information about where your funds are.

KtorZ (commented):
While I like the motivation behind the proposal, I don't think that the proposed designed would work for it allows multiple actors to get a hold of some funds without actually ensuring that funds will be available when they need to spend them.

Consider even only the case where there's a single delegated authorization. There's nothing preventing a user from withdrawing their funds into a different UTxO, leaving any contract relying on the original funds screwed. While I can think of use cases where this makes sense, I can also think of problematic ones (e.g. An auction). Moving (and thus, locking) funds into contract is a fundamental part that makes many DApps work. 

Having said that, I think it would be possible to achieve a similar set of goals from the one highlighted in this document by working on a common conventionnal definition of a 'managed account' via contract. Said differently, instead of modifying the core structure of an output, we could imagine designing a script whose sole purpose is to be a façade for other DApps. DApps would have to explicitly embrace working with these special script-locked inputs as part of their contracts, but it would give nice extra benefits to users (and consequently, alleviate DApps from having to consider things like cancelation or reimbursement). This is very vague but, feels like a track worth exploring 🤔

imikushin:
@L-as: 

> I get the motivation, but I don't think this is a good idea. There are two unrelated issues here:
> 
> * Protocols some times don't respect staking. Those protocols are buggy. Don't use them.

Can you name a few that do? I used a few DEXes (MuesliSwap, SundaeSwap, Minswap, WingRiders), and none of them offered me my staking rewards back for the time my ADA was in their custody.

> * Wallets don't show where your funds are in custom protocols. This is harder, but essentially entirely a wallet issue. Wallets need to be fixed.

Is there a uniform way to reflect ownership of assets in Cardano? (For the assets sent to dApps, I mean.) Genuinely curious. But even if it exists, it **cannot** be enforced.

> You could imagine there being some wallet-protocol interoperation ABI specification language, such that the wallet can provide limited information about where your funds are.

Would make a nice CIP idea 😉

TerminadaPool:
@L-as
> Can you name a few that do? I used a few DEXes (MuesliSwap, SundaeSwap, Minswap, WingRiders), and none of them offered me my staking rewards back for the time my ADA was in their custody.

Indigo protocol uses franken-addresses to let you keep your staking rewards on your Ada locked in its CDPs.  Also I believe that AXO's "programmable swaps" intends to do the same - though it is not released yet.

imikushin:
> Indigo protocol uses franken-addresses to let you keep your staking rewards on your Ada locked in its CDPs. 

Curious about franken-addresses. Can you provide any links?

> Also I believe that AXO's "programmable swaps" intends to do the same - though it is not released yet.

Axo's Jarek talking about this was actually an inspiration to write this CIP. And wouldn't it be nice to be able to use **any** DApp this way?

TerminadaPool:
Franken addresses video by @AndrewWestberg https://youtu.be/KULzovfWn-M

TerminadaPool:
Ledger specifications for Cardano addresses on page 114 of https://hydra.iohk.io/build/7918420/download/1/ledger-spec.pdf

L-as:
> Is there a uniform way to reflect ownership of assets in Cardano? (For the assets sent to dApps, I mean.) Genuinely curious. But even if it exists, it **cannot** be enforced.

Shitty protocols will always exist.

> > You could imagine there being some wallet-protocol interoperation ABI specification language, such that the wallet can provide limited information about where your funds are.
> 
> Would make a nice CIP idea 😉

Yes, definitely. I might make one in the future after my work on our Plutus Specification Language.

CIP-1775/README.md
```diff
@@ -0,0 +1,108 @@
...
+
+## Motivation
+
+In order to use any dApp on Cardano (e.g. a decentralized exchange or an NFT marketplace), a user has to **deposit tokens** to a script address, thereby **giving up custody** of the tokens.
+
+This can be problematic for a number of reasons.
+
+- The user loses control of the assets they send out to a smart-contract, unless the smart contract is specifically designed to allow for such control (e.g. giving a "Cancel trade" or "De-list NFT" button in the UI).
+- Assets sent out from a user's wallet:
+  - are no longer  **delegated** to the stake pool of the user's choice.
```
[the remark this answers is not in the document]
AndrewWestberg: Some DEXes like WingRiders allow those in a liquidity pool to vote for the stakepool where all the ada gets delegated. It's definitely a shortcoming of the DApp design, not a protocol limitation.

CIP-1775/README.md
```diff
@@ -0,0 +1,108 @@
...
+We believe, this is not strictly necessary and would like a way to allow users to delegate spending authorization for some outputs **within** their wallets to other users (smart contracts, multisigs, wallet addresses).
+
+### Use Cases
+
+Instead of sending tokens to the dapp smart-contract, the user creates an output in their own wallet that delegates spending authorization to a smart-contract (or other entity).
+
+Some **existing** use cases, where this can be beneficial:
+
+1. Swapping tokens on a decentralized exchange.
+2. Providing liquidity on a decentralized exchange.
```
AndrewWestberg, line 42: How does the DEX monitor these thousands of delegated utxos to ensure they haven't been spent yet? In the current case where the DEX has custody of the funds, they only have to monitor 1 or 2 UTxOs to know what the state of the liquidity pool is. In the case of thousands of UTxOs to monitor, the smart contract becomes un-runnable due to size. 

If there's no common liquidity pool to swap with, how do you take just a sliver of tokens from thousands of utxos without making them no longer spendable. The whole concept of a DEX doesn't work unless custody is given. Without custody, there's no guaranteed-available liquidity there.

CIP-1775/README.md
```diff
@@ -0,0 +1,108 @@
...
+3. Listing an NFT on a marketplace.
+
+Some **New** use cases, enabled by this proposal:
+
+1. On-chain noncustodial asset management. A user can delegate managing (some of) their assets to a manager, another individual user or organization.
```
AndrewWestberg, line 47: This is achievable already with multi-sig. You could easily send funds to an address that multiple individuals are allowed to spend from. It's just a matter of properly showing these funds in a wallet interface.

The funds could remain fully controllable by you and fully controllable by the other individual or organization. You could do a 1 of 2 multisig to achieve it.

CIP-1775/README.md
```diff
@@ -0,0 +1,108 @@
...
+2. Concurrent use of dapps. For example, a user can provide liquidity and, at the same time, have a trade open at a (different) DEX, with the same output. When the trade is executed, it looks just as though the user has withdrawn the output from the liquidity pool.
```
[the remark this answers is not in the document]
AndrewWestberg: The ledger rules prevent double-spending, but yes, there's likely going to end up being failed contracts and lots of collateral collecting happening by the protocol in this scenario.

CIP-1775/README.md
```diff
@@ -0,0 +1,108 @@
...
+3. Seamless interaction with L2 (sidechains or Hydra). The user doesn't need to send assets to the L2 entrypoint script/multisig, nor do they need to explicitly take them out.
```
AndrewWestberg, line 49: There's ways to achieve this with off-chain code. You can monitor a given multi-sig address for receipt of assets and mint duplicate assets inside an already running hydra head. This way, you can keep the hydra head running and doing work without explicitly starting or stopping the head. The off-chain code can operate as the on/off ramp for the head and handle continuous settlement with the main chain.

imikushin:
@KtorZ Thank you for taking a look!

> While I like the motivation behind the proposal, I don't think that the proposed designed would work for it allows multiple actors to get a hold of some funds without actually ensuring that funds will be available when they need to spend them.

We can address that.

It is specifically called out that **only** wallet addresses will support delegated spending authorizations. So, assets **sent** to a script are still **only** controlled by the script.

> Consider even only the case where there's a single delegated authorization. There's nothing preventing a user from withdrawing their funds into a different UTxO, leaving any contract relying on the original funds screwed. While I can think of use cases where this makes sense, I can also think of problematic ones (e.g. An auction). Moving (and thus, locking) funds into contract is a fundamental part that makes many DApps work.

I can see your concern for **existing** Plutus scripts expecting to have exclusive control of an output, while the output might be in a wallet and could be spent without script's "consent". 

Maybe we could require that **only Plutus v3** (or higher) scripts can participate in DSAs, and make DSAs **available via script context**. Scripts relying on asset locking would easily confirm exclusive control of them. 

As an added benefit, we can also allow DSAs to work the same way for outputs **located at** Plutus v3 script addresses. This would open up even more use cases: imagine being able to add functionality while keeping existing scripts.

> Having said that, I think it would be possible to achieve a similar set of goals from the one highlighted in this document by working on a common conventionnal definition of a 'managed account' via contract. Said differently, instead of modifying the core structure of an output, we could imagine designing a script whose sole purpose is to be a façade for other DApps. DApps would have to explicitly embrace working with these special script-locked inputs as part of their contracts, but it would give nice extra benefits to users (and consequently, alleviate DApps from having to consider things like cancelation or reimbursement). This is very vague but, feels like a track worth exploring 🤔

Not a bad idea. Yet, more complicated: 
- for the end user - yet another script to trust, 
- for the DApps - changes to script context, perhaps even not as simple as adding a field with a map of authorized spenders.

CIP-1775/README.md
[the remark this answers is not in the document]
imikushin: The problems are: 
- It is hard: you have to create a personal script address for each user.
- They don't have to. And actually incentivized to do the opposite: stake users' assets locked in their smart contract and keep the rewards.

And how would users withdraw staking rewards from all these protocols? They sure can't simply press a button in their wallet to do that.

And what about governance? Can I vote with my funds locked in a LP or orders on DEXes?

CIP-1775/README.md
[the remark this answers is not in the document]
imikushin: Value transfers still work the same, consensus protocol still works the same, therefore there's no double spending.

CIP-1775/README.md
```diff
@@ -0,0 +1,108 @@
...
+  - are no longer  **delegated** to the stake pool of the user's choice.
+  - cannot participate in **governance voting** on behalf of the user.
+- Any output, accidentally sent to a script address without a datum is effectively lost now (because of how transaction validation scripts are invoked).
+- Requiring users to deposit assets into a dapp may subject the dapp's owners to regulatory scrutiny.
+
+We believe, this is not strictly necessary and would like a way to allow users to delegate spending authorization for some outputs **within** their wallets to other users (smart contracts, multisigs, wallet addresses).
+
+### Use Cases
+
+Instead of sending tokens to the dapp smart-contract, the user creates an output in their own wallet that delegates spending authorization to a smart-contract (or other entity).
```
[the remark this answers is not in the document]
imikushin: We could look into extending native scripts: 
- add `datum_option` to any Plutus script address referred by the native script,
- include the native script into the script context passed to the referred Plutus scripts.

This doesn't address staking and voting. Integration with sidechains (and Hydra) will not be as seamless as it could be with the proposed CIP.

CIP-1775/README.md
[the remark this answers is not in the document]
imikushin: @AndrewWestberg,

> Some DEXes like WingRiders allow those in a liquidity pool to vote for the stakepool where all the ada gets delegated.

What about orders? For example, I want to buy some WRT tokens at some low price, send a limit order and wait for weeks for my order to get executed (if at all).

CIP-1775/README.md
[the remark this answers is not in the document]
imikushin: @AndrewWestberg 

> lots of collateral collecting happening by the protocol in this scenario

What do you mean? And is it bad?

CIP-1775/README.md
imikushin: The point isn't to take away full custody from DApps.

The point is to introduce an additional, non-custodial, way to interact with DApps that support it.

> How does the DEX monitor these thousands of delegated utxos to ensure they haven't been spent yet? In the current case where the DEX has custody of the funds, they only have to monitor 1 or 2 UTxOs to know what the state of the liquidity pool is. 

Can't they run an indexer off-chain to know the size? I didn't know that a liquidity pool can only be 1 or 2 UTXOs. (And still don't know 🙂). If I'm one of thousands of people providing liquidity, how can I withdraw my assets? My ownership has to be encoded in these 1 or 2 UTXOs. No wonder people complain about concurrency. It doesn't have to be **only** this way, though.

> In the case of thousands of UTxOs to monitor, the smart contract becomes un-runnable due to size.

Aren't smart-contracts in Cardano designed to validate spending of a **single** output? Which one to spend can be decided off-chain, I imagine.

> If there's no common liquidity pool to swap with, how do you take just a sliver of tokens from thousands of utxos without making them no longer spendable. 

I'd imagine, an off-chain process matches swap orders with these thousands of UTXOs and constructs transactions. A transaction for each swap order, spends one or more (just a few, not thousands) such UTXOs. UTXOs with the remaining and incoming tokens are created back in the LP users' wallets.

If, by the tx execution time, some of the LP UTXOs are already spent, the transaction is re-constructed, until executed. This is classic [Optimistic concurrency control](https://en.wikipedia.org/wiki/Optimistic_concurrency_control).

> The whole concept of a DEX doesn't work unless custody is given. Without custody, there's no guaranteed-available liquidity there.

Nothing in this world is guaranteed (except for taxes and death, of course). But! Optimistic concurrency does work, and I don't see why not make use of it for the benefit of Cardano users.

CIP-1775/README.md
imikushin: Cool. Can I vote in Catalyst with this multisig? How do I delegate to an SPO with the multisig?

CIP-1775/README.md
imikushin: The point is to **not** make users to send to a multisig. And, symmetrically, to not make their recipients pull the assets back (when the interaction is over).

CIP-1775/README.md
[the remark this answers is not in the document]
TerminadaPool: DEX's can use franken-addresses to solve the problem of allowing smart contract locked Ada to still earn staking rewards.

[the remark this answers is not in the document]
imikushin: Fixing the text. Thanks for your patience!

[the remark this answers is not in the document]
michaelpj: > If a user's StakeKeyHash is used as the delegation part of the script address, the outputs at this personalized script address are staked just as the user's wallet.

Given that this is the case, I'm not sure why the text continues to say that you can't stake to your chosen pool? You absolutely can!

CIP-1775/README.md
```diff
@@ -0,0 +1,108 @@
...
+
+We propose to allow users to delegate authorized spending of outputs in their wallets to smart contracts or other users, without sending said outputs out of their wallets.
+
+## Motivation
+
+In order to use any dApp on Cardano (e.g. a decentralized exchange or an NFT marketplace), a user has to **deposit tokens** to a script address, thereby **giving up custody** of the tokens.
+
+This can be problematic for a number of reasons.
+
+- The user loses control of the assets they send out to a smart-contract, unless the smart contract is specifically designed to allow for such control (e.g. giving a "Cancel trade" or "De-list NFT" button in the UI).
```
michaelpj, line 26: I'm confused about why this is a bad thing, and I think that's a crux of disagreement.

At the moment, an output which has an address A is _exclusively_ spendable according to that address. This is a useful property! Consider e.g. the fact that many dapps rely on collateral or deposits of some kind - most of these systems _do not work_ if the depositor can simply reclaim their collateral by using an alternative spending condition.

As you point out, today you can _deliberately_ remove exclusivity by building it into your script.

IMO, the main effect of this CIP would be to _flip the default_: addresses would by default become non-exclusive. But as we saw, exclusivity is _important_ to the functioning of many applications. So you would need a way to ensure you _did_ have exclusivity, which seems both tricky and an annoying footgun for application developers (didn't add the check to ensure exclusivity? Oops, users went off with all their deposits).

So:
- Today: exclusivity is default and useful, applications can build in non-exclusivity easily
- With this CIP: non-exclusivity is default, applications can build in exclusivity (?) but it's harder

This doesn't look like a good trade to me.

CIP-1775/README.md
michaelpj: Yes, the staking credential and the spending credential are entirely independent, which is a very important point for this proposal!

CIP-1775/README.md
michaelpj: This seems like exactly a case where the application _needs_ exclusivity! What you're suggesting would allow me to send some funds to a Hydra head, perform transactions on there that transfer that money to someone else... and then _also_ spend the money on the mainchain. I'm not sure what would happen there but probably we'd have to cancel the whole head? Worst case it lets you steal money. This does _not_ seem desirable.

CIP-1775/README.md
```diff
@@ -0,0 +1,108 @@
...
+
+If the output's `address` is of **any other kind** (e.g. native or Plutus script), the _delegated spending authorizations_ field contents **do not affect** spending behavior. Existing scripts' assumptions must be **unaffected** by the proposed ledger change: spending of outputs at a script address are controlled by the script only.
+
+With the above rules, we believe, the change does not require Plutus language version bump.
+
+## Implications
+
+While providing the benefits listed in the **Motivation** section, there is a certain cost: users need to be made aware of _delegated spending authorizations_ on the outputs in their wallets. 
+
+Assets that arrive into a user's wallet with delegated spending authorizations attached, kind of come with a form of remote control (which can be exploited). Obviously, the user can remove it, but **only** if they know of its existence.
```
michaelpj, line 91: This is pretty nasty, tbh, and another ugly case for wallet implementors.

CIP-1775/README.md
```diff
@@ -0,0 +1,108 @@
...
+
+The new field is a map from `address` to `datum_option` or `nil`. In this map, `address` can be any address: Plutus v1 and v2 scripts, wallets, etc. The value (`datum_option`) only makes sense for Plutus script addresses, so it is optional (can be `nil`).
+
+### Spending of Outputs
+
+When the output's `address` (field 0) is a **wallet** address, a transaction can spend the output if it satisfies spending conditions:
+- for the output's address (i.e. signed by the wallet's private key), or 
+- for any address in the output's _delegated spending authorizations_.
+
+If the output's `address` is of **any other kind** (e.g. native or Plutus script), the _delegated spending authorizations_ field contents **do not affect** spending behavior. Existing scripts' assumptions must be **unaffected** by the proposed ledger change: spending of outputs at a script address are controlled by the script only.
```
michaelpj, line 83: I don't particularly like that this is asymmetrical. I haven't thought of somewhere where that goes wrong, but there often is such a place.

CIP-1775/README.md
michaelpj: Okay, I hadn't quite realised that the addition is asymmetrical. So instead we'd have a situation where applications will have to decide whether to do anything with outputs that have delegated spending conditions or not. That is indeed less unsafe although weirdly asymmetrical, and I need to think about it.

CIP-1775/README.md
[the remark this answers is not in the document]
imikushin: @michaelpj I'm fixing the text.

CIP-1775/README.md
imikushin: Yes, I realize that. To fix this, I'm updating the text so that only beginning with Plutus v3, scripts will be able to participate in DSAs or receive outputs with DSAs.

CIP-1775/README.md
imikushin: I'm updating the text so that only beginning with Plutus v3, scripts will be able to participate in DSAs or receive outputs with DSAs. This will restore the symmetry.

The version bump is necessary in order to preserve the exclusivity assumptions of existing scripts and to be able to pass DSAs via script context.

CIP-1775/README.md
imikushin: Updated the text.

imikushin:
@L-as @KtorZ @TerminadaPool @AndrewWestberg 

Thank you all for the feedback! The text should be ready for another look.

CIP-1775/README.md
```diff
@@ -0,0 +1,160 @@
...
+# Delegated Spending Authorizations
+
+"Give me custody, or give me death!" - [@ArmySpies](https://twitter.com/ArmySpies/status/1592586212816355328)
+
+## Abstract
+
+We propose to allow users to delegate authorized spending of outputs in their wallets to smart contracts or other users,
+without sending said outputs out of their wallets.
+
+## Motivation
```
SebastienGllmt, line 21: I think this has an almost total overlap with https://github.com/cardano-foundation/CIPs/pull/309 in motivation. The main difference is that instead of using native scripts and proxy contracts to define the delegated authorization, this CIP instead proposes inlining this system directly into the UTXO definition

imikushin: @SebastienGllmt The outcome seems to be similar, but **only if** the datum (or `()` if not present in the output) could be passed to the script referenced by the `RequireScript` clause. This would indeed make CIP-38 equivalent: for every user of a DApp, a proxy script could be constructed which would enable both the user and the DApp to spend the output.

Wallets would then find such outputs and show as still being in their custody, but still manageable by other entities.

I do see a couple issues to iron out before I could get behind CIP-38:
1. We should specify what the `RequireScript` script is passed as its arguments. If it's all the same arguments as a regular Plutus validation script, this works for me. This also means, datum should be present in the output or implied as `()`.
2. Looks like this can only work starting with Plutus v3: we should make the Plutus script aware (via script context) of its non-exclusive control of the output.

CIP-1775/README.md
imikushin: @michaelpj It really depends on how the L2 treats concurrency. In Optimistic Concurrency control, this works. Where exclusive locking is needed, depositing to smart contracts is still available 🙂.

CIP-1775/README.md
imikushin: Well, progress does require some work.

CIP-1775/README.md
imikushin: @michaelpj I've updated the text.

imikushin:
> Having said that, I think it would be possible to achieve a similar set of goals from the one highlighted in this document by working on a common conventionnal definition of a 'managed account' via contract. Said differently, instead of modifying the core structure of an output, we could imagine designing a script whose sole purpose is to be a façade for other DApps. DApps would have to explicitly embrace working with these special script-locked inputs as part of their contracts, but it would give nice extra benefits to users (and consequently, alleviate DApps from having to consider things like cancelation or reimbursement). This is very vague but, feels like a track worth exploring 🤔

@KtorZ, did you have a chance to take a look at #309 yet (posted by @SebastienGllmt)? It does look pretty much just like what you were suggesting. Interpretation of the referred scripts needs clarification (which I'm suggesting to add [here](https://github.com/cardano-foundation/CIPs/pull/309/files#r1037768135)), but otherwise looks promising.

fallen-icarus:
@imikushin @AndrewWestberg In my experience, the loss of custody is more of a consequence of the way Dapps are designed rather than the "law". As a counter example, I created a DEX proof-of-concept where users DO NOT give up custody of their assets nor do they pool assets together into a liquidity pool. They also maintain full delegation control of their assets at all times. I open-sourced it and it can be found [here](https://github.com/fallen-icarus/cardano-swaps).

Everything is explained in the README but the main idea is to give all users their own personal smart contract instance which they uniquely control. This alone is a major paradigm shift in the way Dapps are designed. Then finding all of the other users is done using Beacon Tokens. There is still room for improvement but given your perspectives, I am curious of your thoughts.

SebastienGllmt:
I just added the desired behavior of this CIP as one of the motivation in CIP-38  https://github.com/cardano-foundation/CIPs/pull/309

rphair:
@imikushin this was never disqualified in the review process but was never pushed forward either (in a long time now).  The idea seems no less relevant today than it did 1½ years ago but we have to close this PR if nothing further is going to happen with the proposal... I'm marking it `Abandoned` based on this assumption in the meantime.

Please let us know if you'd like to press this forward; if so I'll put on the next CIP meeting agenda to reconfirm the need, confirm that any remaining issues are addressed, and assign a properly sequential CIP number.  Failing that (or any reviewer wanting to take up this proposal in your stead) this would be closed some time later when tagged `Abandoned` proposals are cleaned up.

cc (reviewers so far): @L-as @KtorZ @TerminadaPool @AndrewWestberg @SebastienGllmt @fallen-icarus

imikushin:
@rphair glad to hear this still seems relevant to people. I'd love to resume working on this if there's interest. I have no desire to spend time on something people don't want. So, if there's anyone still interested in this, please react to this comment.

rphair:
Thank you @imikushin - without confirmation of such interest we will have to close this for now.  If similar issues come up in the future I'll try to make a note here so you might better gauge the community need to reopen this proposal submission.

PR #388, opened 2022-11-30
Update top-level README and CIP/CPS number allocations.

KtorZ:
I've been a bit more "drastic" than usual about what is moved into the "waiting for author" section but, I think -- moving forward -- we should be a bit more proactive in declaring CIPs inactive. 

We're seeing a lot of proposals that are often submitted without much follow-up from their authors; probably because of a misunderstanding about the process and the role of editors altogether. So I am thinking that moving some inactive proposals to "Waiting for author" once they reach a certain level of inactivity may actually wake up some authors to "do their parts". 

And, it helps us clearing up a bit proposals that aren't actually championed anymore.


rphair (approved):
(commit above was to resolve merge conflict from section title change)

yes @KtorZ I think the new section will have exactly the desired effect & create some positive peer pressure either to advance the stalled proposals or to close/disavow them.

PR #391, opened 2022-12-01
Adjust preamble and structure of CIP-0002 to abide by new CIP-0001

### LINKED ISSUE #390: Adjust preamble and structure of CIP-0002 to abide by new CIP-0001

KtorZ:
Fixes #390.

### DIFF
--- a/CIP-0002/README.md
+++ b/CIP-0002/README.md
@@ -1,13 +1,19 @@
 ---
 CIP: 2
 Title: Coin Selection Algorithms for Cardano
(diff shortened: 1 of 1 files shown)


KtorZ:
Yes, I realized that this one wasn't quite following the original template in the first place; so I had to re-order a few sections to make them fit in the new template. Nothing substantial but necessary I believe.

KtorZ:
(Just removed the 'markdown' from the header; I had put links in a markdown format but I am not sure whether this will play well down-the-line so, I am reverting this to just plain URLs).

PR #392, opened 2022-12-01
CPS-???? | Properly burning NFTs / tokens

matiwinnetou:
([rendered version in branch](https://github.com/matiwinnetou/CIPs/blob/CPS--burn-tokens/CPS-%3F%3F%3F/README.md))

### DIFF
--- a/CPS-???/README.md
+++ b/CPS-???/README.md
@@ -19,16 +19,25 @@ The aim of this problem statement is to raise the issue around the ability to bu
 
 It is understood that minUTxO has been introduced as a way to protect ledger from tokens and NFTs spam, however, the only currently supported methods of "burning NFTs" do not address the core of the problem, namely releasing ADA 'locked' behind a token or NFT. Currently recommended methods of 'burning' are either sending to another wallet owned by the person sending or send to another wallet owned by somebody else. Neither of those methods actually releases underlying 'locked' minUTxO ADA.
 
+It is worth mentioning that since the deposit scaling is affine, there is some incentive to keep bundling the tokens into bigger and bigger transaction outputs, potentially making the situation worse. It's unclear what the expectation should be regarding accepting payments in ADA that may be mired in tokens.
+
 ## Use cases
 
 - As a Cardano user I have many NFTs that I do not need anymore and I would like to relaim ADA 'locked' by those NFTs. If a Cardano user has many useless NFTs / tokens this could sum up to significant amount of 'locked' minUTxO ADA.
+- A typical use case of this would be going to your wallet of choice and bundling one or multiple tokens in one transaction and burning them somehow. Upon such a burn, ADA locked in those UTxOs would be reclaimed.
+
+- Proposal is not a feature request to wallets only, it actually requires ledger changes first, which allow such reclaiming of ADA locked by tokens
 
 ## Goals
 - Add ability for users to burn tokens and NFTs to be able to reclaim 'locked' minUTxO ADA.
 
 ## Open Questions
-- How can we actually allow this if blockchain is supposed to be immutable? One there was a token and now there is no token anymore (dangling tokens?).
+- How can we actually technically burn tokens?
+- What changes would there need to be on ledger level to support such burning and make sense in the current ledger architecture?
 
+## Bad solutions
+- a potential solution of sending to zero address is not a good solution because not only it does not give you a full deposit back but also
+zero addresses can be used as a reference input which could cause unexpected issues for somebody who thinks a token was burned
 
 ##  Possible solution
 - smart contract / native script being able to burn token after minting policy is closed
--- a/CPS-???/README.md
+++ b/CPS-???/README.md
@@ -1,0 +1,34 @@
+---
+CPS: ?
+Title: Burning NFTs / tokens
+Status: Open
+Category: Tokens
+Authors:
+    - Mateusz Czeladka <mateusz.czeladka@cardanofoundation.org>
+Proposed Solutions: 
+    - smart contract / native script that is able to burn token
+
+Created: 2022-12-01
+---
+
+## Abstract
+
+The aim of this problem statement is to raise the issue around the ability to burn NFT / tokens and release minUTxO (ADA) backing the NFT / token.
+
+## Problem
+
+It is understood that minUTxO has been introduced as a way to protect ledger from tokens and NFTs spam, however, the only currently supported methods of "burning NFTs" do not address the core of the problem, namely releasing ADA 'locked' behind a token or NFT. Currently recommended methods of 'burning' are either sending to another wallet owned by the person sending or send to another wallet owned by somebody else. Neither of those methods actually releases underlying 'locked' minUTxO ADA.
+
+## Use cases
+
+- As a Cardano user I have many NFTs that I do not need anymore and I would like to relaim ADA 'locked' by those NFTs. If a Cardano user has many useless NFTs / tokens this could sum up to significant amount of 'locked' minUTxO ADA.
+
+## Goals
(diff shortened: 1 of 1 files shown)


rphair:
Thanks for posting this @matiwinnetou - marked officially as in Draft stage according to the document header & the rudimentary stage of development.  When ready for review it can be [marked as such](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/proposing-changes-to-your-work-with-pull-requests/changing-the-stage-of-a-pull-request).

rphair:
@matiwinnetou since today's CIP meeting acknowledged that this proposal is still a bit "thin" - a bit like a feature request for wallets - could you include some use cases for what NFT / token burns would look like, as a way of filling out the proposal and identifying components that might exist beyond the wallets themselves?

CPS-???/README.md
```diff
@@ -0,0 +1,34 @@
...
+
+## Goals
+- Add ability for users to burn tokens and NFTs to be able to reclaim 'locked' minUTxO ADA.
+
+## Open Questions
+- How can we actually allow this if blockchain is supposed to be immutable? One there was a token and now there is no token anymore (dangling tokens?).
+
+
+##  Possible solution
+- smart contract / native script being able to burn token after minting policy is closed
```
SebastienGllmt, line 34: I would add something about sending to the zero-address not being a good solution because not only does it not give you your full deposit back, but also the zero-address can still be used as a reference input which could cause unexpected issues for somebody who thinks a token was burned

CPS-???/README.md
```diff
@@ -0,0 +1,34 @@
...
+
+## Use cases
+
+- As a Cardano user I have many NFTs that I do not need anymore and I would like to relaim ADA 'locked' by those NFTs. If a Cardano user has many useless NFTs / tokens this could sum up to significant amount of 'locked' minUTxO ADA.
+
+## Goals
+- Add ability for users to burn tokens and NFTs to be able to reclaim 'locked' minUTxO ADA.
+
+## Open Questions
+- How can we actually allow this if blockchain is supposed to be immutable? One there was a token and now there is no token anymore (dangling tokens?).
```
JaredCorduan, line 30: In my view, the immutability of the blockchain is the combination of two things:
* an immutable, sequential, time-stamped sequence of blocks
* the (implicit, human) agreement that the semantics of the blocks will not change after they are accepted.

In particular, the _ledger state_ is very much _not_ immutable. Indeed, transaction inputs come and go, as do registered stake credentials, etc.

CPS-???/README.md
```diff
@@ -0,0 +1,34 @@
...
+Created: 2022-12-01
+---
+
+## Abstract
+
+The aim of this problem statement is to raise the issue around the ability to burn NFT / tokens and release minUTxO (ADA) backing the NFT / token.
+
+## Problem
+
+It is understood that minUTxO has been introduced as a way to protect ledger from tokens and NFTs spam, however, the only currently supported methods of "burning NFTs" do not address the core of the problem, namely releasing ADA 'locked' behind a token or NFT. Currently recommended methods of 'burning' are either sending to another wallet owned by the person sending or send to another wallet owned by somebody else. Neither of those methods actually releases underlying 'locked' minUTxO ADA.
```
JaredCorduan, line 20: Thank you for writing this CPS up. It might also be worth mentioning that since the deposit scaling is affine, there is some incentive to keep bundling the tokens into bigger and bigger transaction outputs, potentially making the situation worse. It's unclear what the expectation should be regarding accepting payments in ADA that may be mired in tokens.

nielstron:
I am willing to take this up and have a _solution_ suggestion:

Instead of native scripts, I suggest users normally use Smart Contract tokens. I will try to compile a set of sample codes to emulate native script behaviour (most importantly the signature stuff) or maybe even an automatic tool that converts a native script into a Smart Contract.

This smart contract would then enforce the native script conditions for _minting_, but always allow burning.

Tutorials for minting/burning NFTs should be rewritten to use this smart contract instead of native tokens to make sure that at least all _future_ tokens (that have no particular reason not to) are compatible


Extended reasoning: there are tokens that DO NOT want to be burned under any circumstances i.e. pool tokens certifying the correctness of a pool. Users should be able to opt-in into having their token burned at any time and it should be as simple as possible for them.

nielstron:
The referenced PR shows a sample script that would compile to only 750 bytes of Smart Contract and cost 0.05 ada to execute, IMO feasible (and much more powerful) as a new default for Cardano Native Tokens

nielstron:
@matiwinnetou I also don't know about ERC-721.

I know about the proposal to implement ERC-20 tokens by @michele-harmonic in https://github.com/cardano-foundation/CIPs/pull/444 . This is different in a sense that it would still make use of Cardano Native Tokens and not simulate them within contract datums.

Further I would suggest that we should not retrospectively change the guarantees made towards token creators (which only increases the potential for insecurity and problems with finite gain, only affecting tokens minted so far). Rather I suggest establishing a new standard that applies to all _new_ tokens with no additional risk and infinite gain (affecting all future tokens).

rphair:
Nice to see that https://github.com/cardano-foundation/CIPs/pull/479 has finally been presented as a solution for this.

michaelpj:
In the past we have discussed a much simpler solution: simply don't require the minting policy script to be run when burning tokens. That is: you can always do it. We considered this right from the beginning in Mary. 

The argument for why this is okay is quite simple: if you can send some tokens to an output, you can _effectively_ send those tokens to the bin by sending them to an unspendable output. An unspendable output is more-or-less equivalent to burning tokens as far as the rest of the system is concerned, since they can never interact with anything else ever again. The only difference is they still show up in the UTXO set (and eat the min-ada deposit!).

I have always been a bit uncertain about this proposal but I think it a) achieves the goal very directly, b) is very simple, c) is very easy to implement, and d) seems pretty safe.

JaredCorduan:
> In the past we have discussed a much simpler solution: simply don't require the minting policy script to be run when burning tokens. That is: you can always do it. We considered this right from the beginning in Mary.

I'd just like to chime in to say that I've always been a strong supporter of this idea.

The other idea that I really like (which perhaps I should write as a CIP) is @redxaxder's "claim check" idea. Think of a hash function as a way to compress data. If I have a large token bundle in a single UTxO, imagine a transaction that exchanges the bundle for the hash of the bundle, thus allowing the min utxo value to decrease. In this form, the tokens are unusable, but if ever you want them back, we would allow for the inverse action, namely exchanging data for the matching hash. And you could have hashes within hashes, an infinite bad of holding.

michaelpj:
And elaborating on a variant of the "claim check" idea that Jared and I discussed: just allow the elision of the "value" part of an output in favour of a hash, in exactly the same way that we do for the script and datum (script are always hashes, datums can be either). The preimage must be provided whenever needed, which means both when creating the output (to check balancing), and when spending the output (same). But crucially the size of the output in the UTXO would then be smaller and constant, so could have a smaller min-utxo requirement.

nielstron:
I found out that there is a problem with this specification in combination with CIP 25:
NFT and FT metadata is tagged by the latest minting transaction of the token. However if every user can burn a token, this is technically  a minting transaction, hence allowing every holder to arbitrarily change the metadata of their FT (for NFTs this is an obvious non-issue). However I would propose to adjust CIP 25 instead of changing any of the approaches here.

rphair:
thanks @nielstron - Even with https://github.com/cardano-foundation/CIPs/pull/527 submitted I think it would also be a good idea to add this reservation to the CPS here (as you say, without changing any of the recommended approaches)... since we don't know whether or how that requested modification to CIP-0025 will proceed and should have this issue documented in any case.

rphair:
@matiwinnetou this looks `Abandoned` so we'll close this soon unless you (or another advocate of this proposal) can post some updates & being carrying this forward.

PR #393, opened 2022-12-01
CPS-???? | On chain dApp and script audits

matiwinnetou:
([rendered version](https://github.com/matiwinnetou/CIPs/blob/CPS-dapp-certification/CPS-%3F%3F%3F/README.md))

### DIFF
--- a/CPS-???/README.md
+++ b/CPS-???/README.md
@@ -1,0 +1,25 @@
+---
+CPS: ?
+Title: On chain dApp / scripts audits
+Status: Open
+Category: ?
+Authors:
+    - Mateusz Czeladka <mateusz.czeladka@cardanofoundation.org>
+Proposed Solutions: []
+Created: 2022-12-01
+---
+
+# DRAFT - certification
+
+## Abstract
+
+Currently there is no way to check if a particular dApp version (release) or a given script has been audited from on-chain metadata.
+
+## Problem
+
+Our understand is that dApp on Cardano is nothing else that a collection of scripts, sometimes it could be even just one script. dApps are naturally evolving therefore they receive new versions and also new script hashes. We often find ourselves in the situation that we have to trust the people issuing scripts. In typical examples one can lookup a certification as PDF from e.g. certiK but there are cases where a team is claiming that a certification of a script is also valid for new version (e.g. when moving from Plutus V1 to Plutus V2). We don't want to trust anybody, we would like to build an indexer from on chain metadata and be able to verify that CertiK has signed off an audit. We don't want to have to trust teams / people building dApps that Plutus V1 certification / audit is also valid for Plutus V2.
+
+If you look at crfa-offchain-metadata registry (https://github.com/Cardano-Fans/crfa-offchain-data-registry) and more specifically dApps currently we maintain manually the fact that there is an audit of a given dApp or not. We would prefer to automate this via on chain indexer and this data should be automatically injested without human level oracle involvement.
+
+##  Possible solution
+Propose a CIP where various certification companies will be able to certify a "dapp release" or certain scripts so that we know that they have been properly audited.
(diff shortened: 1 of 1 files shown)


rphair:
@matiwinnetou sanitised the title from the `/` character since this has a way of throwing off scripts (and these documents are automatically formatted into other documents)

rphair:
Thanks for posting this @matiwinnetou - marked officially as in Draft stage according to the document header & the rudimentary stage of development.  When ready for review it can be [marked as such](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/proposing-changes-to-your-work-with-pull-requests/changing-the-stage-of-a-pull-request).

matiwinnetou:
@rphair I improved this PR and from my point of view it is ready for a proper review, let me know if there is something I am still missing for it to be accepted for a proper review, e.g. part of process?

CPS-???/README.md
```diff
@@ -0,0 +1,42 @@
+---
+CPS: ?
+Title: On chain dApp / scripts audits
+Status: Open
+Category: dApps
```
rphair, line 5:
```suggestion
Category: Metadata
```
since there's no `dApps` category on the initial list, I'd suggest `Metadata` because it should refer to script certification as well.

rphair (commented):
thanks @matiwinnetou - it seems like a good start for collective review, so the current PR status should be fine...

RSoulatIOHK:
I’d like to discuss your open questions:
1) I don’t think you can but people or DApp stores could always maintain a list of trusted auditors and not consider valid any “certificate” issued by others
2) Couldn’t the audit report refer to the different script hashes that have been audited and certified? We could even hash those hashes to link the certificate to a version of the DApp and not to each script individually.
3) I would expect that you don’t. I would expect new releases to require a new audit. It could be a lighter audit that only addresses the changes but it would be a new certificate issued.

matiwinnetou:
@RSoulatIOHK 

>
> I’d like to discuss your open questions:

> I don’t think you can but people or DApp stores could always maintain a list of trusted auditors and not consider valid any “certificate” issued by others

I like that it is a simple and alternative is that we need GPG or Keybase on Cardano. @rcmorano that now works for CF actually wanted to build it among a few other ideas he had :)

> Couldn’t the audit report refer to the different script hashes that have been audited and certified? We could even hash those hashes to link the certificate to a version of the DApp and not to each script individually.

Sounds like an idea. We have not explored that yet.

> I would expect that you don’t. I would expect new releases to require a new audit. It could be a lighter audit that only addresses the changes but it would be a new certificate issued.

This is what I thought but obviously reality is that dApp teams will want to save costs and shave you report for V1 which they claim is also valid for V2. This is one of the reasons for this CPS, we need some on chain dApp certification CIP.

RSoulatIOHK:
@matiwinnetou 

> This is what I thought but obviously reality is that dApp teams will want to save costs and shave you report for V1 which they claim is also valid for V2. This is one of the reasons for this CPS, we need some on chain dApp certification CIP.

I mean, yes they will want to do that but I'm not sure we should accept it.
Maybe we could try to link to an history of certification for example through metadata and use old version of a dApp and its links to certificates to show that this dApp has an history of certification and thus could be trusted even without a new certificate but I would not want to trust the developers to say that their update don't have to be recertified. As an auditor, I don't think I would want my audit and certificate to be reused without me auditing the changes. 

And I agree we need a CIP on certification.

KtorZ:
As discussed in today's editors meeting, the problem stated in this CPS is the light of open questions raised by CIP-0052. I would suggest @simonjohnthompson to also join this conversation and perhaps, reach out to the auditors involved in the writing of CIP-0052. There's perhaps already something being cooked behind the scene in that regard.

rphair:
... particularly @simonjohnthompson we mentioned that in the process of finishing (?) the updates begun in https://github.com/cardano-foundation/CIPs/pull/406 that any polling of the current list of auditors could also include querying them about how they feel about _this_ PR.

RSoulatIOHK:
There is a working group about certification where some auditors are present. I can add discussing both #406 and our current point to next week's agenda.

PR #395, opened 2022-12-01
CIP-0045? | Decentralized WebRTC dApp-Wallet Communication

fabianbormann:
This CIP aims to propose a decentralized communication method between dApps and wallets based on WebTorrent trackers and WebRTC.

---

([rendered CIP document in branch](https://github.com/fabianbormann/CIPs/blob/master/CIP-%3F%3F%3F%3F/README.md))


CIP-????/README.md
```diff
@@ -0,0 +1,247 @@
+---
+CIP: ?
+Title: Decentralized WebRTC dApp/Wallet Communication Using WebTorrent Tracker for Peer Discovery
```
```suggestion
Title: Decentralized WebRTC dApp-Wallet Communication
```
Likewise this title is also scraped out of the header so needs to be generally as long as others: for example when it appears in a long list of CIPs in a narrow sidebar on a web page.

rphair (commented):
@fabianbormann we have to shorten the GitHub title a bit (and remove a potentially offending special character) because of the limited space for the title we get in certain contexts.

@SebastienGllmt @KtorZ @crptmppt I wanted to check with you before also recommending that the code in `/demo` be moved elsewhere.  There are some accountability issues with having this code in this particular repository... so unless there are any objections I'm going to recommend this be moved into another repo controlled by the author (if not there already) & linked to from the CIP document.

I also believe the spirit of the CIP process is for each proposal to focus on a standard that could have many implementations: so to have specific code here for a particular implementation creates a false impression that the implementation _is_ the standard. 🧐

CIP-????/README.md
fabianbormann: Thanks @rphair I will create a demo repository under my Github profile and link it instead of providing the code directly. You are right: the intention of this CIP is to provide a standard and not a concrete implementation. It's just a poc to show that this architecture actually works 😊  ..

CIP-????/README.md
```diff
@@ -0,0 +1,247 @@
...
+    TURN/STUN-->>dApp: You are <ip:port>
+    Wallet-->>+TURN/STUN: Who am I?
+    TURN/STUN-->>Wallet: You are <ip:port>
+    dApp->>Wallet: messages
+    Wallet->>dApp: messages
+```
+
+The data will be send peer to peer via a WebRTC data channel once the ip discovery has been finished. E.g. WalletConnect expects/provides a relay server URL to initialize the connection. While this method allows dApps to communitcate peer-to-peer with wallets it also leads to a [possible SPOF](https://twitter.com/walletconnect/status/1407279853943001088?lang=en).
+
+### Establish a WebRTC Connection Using WebTorrent Tracker (Our approach)
```
matiwinnetou, line 44: Is there any possibility that some malicious actor may inject compromised website instead of legitimate one and by this scam users? Centralised servers solution from TrustWallet is far from ideal but had-shaking to malicious / scammers cannot happen then I guess.

fabianbormann: @matiwinnetou that's the main reason why I reversed the communication as described in the security section. The dApp acts as a server with an "api" rpc method allowing the wallet (client) to inject it's api. So the wallet will inject the api and the webpage will not inject any code into the wallet. And you need to explicitly share the address (public key) with the wallet app. Using a QR scanner in the wallet app or deeplink/universal link (will ask you in a mobile system ui dialog "open with .."). So it would still be possible that a user inject it's wallet api into a scam website by scanning a malicious QR code from a fake dApp. or e.g. connecting the wallet to voting.summit.cadano.org instead of voting.summit.ca **r** dano.org

fabianbormann: If you call api functions from the dApp that requires user interactions within the wallet app the wallet app will check if the address of the caller is equal to the servers (dApp) address to protect you and bring the wallet app to foreground to then perform e.g. the signing request manually.

rphair (commented):
I would look forward to another more technical review but I can say for my own part that this has all the basics of a good design.  It's relevant to Cardano community needs and the `Path to Active` is robust and detailed.

@fabianbormann FYI this is going to be introduced in "Triage" at the biweekly CIP meeting on Discord in case you are there to say a few words about it (cc @KtorZ).  In any case I would recommend that this be promoted to a candidate CIP for more active consideration.

KtorZ:
As discussed in today's meeting: the proposal looks well articulated, with a PoC and a clear motivation. It would be interesting to get some wallet providers involved in the discussion. 

In the meantime, we'll assign a number to that proposal and mark it for review after some more reviews have happened.

fabianbormann:
@hazae41 good idea. It would be great to support different methods to push the decentralization. Meanwhile I reimplemented bugout into a typescript lib that supports react (webpack 5) without eject (called [Meerkat](https://github.com/fabianbormann/meerkat)) so that the next step would be a reference implementation of the api injection in boostwallet (@jimcase is currently working on this). But I will look into libp2p pubsub in parallel and if it makes sense, add it as alternative peer discovery method.

CIP-????/README.md
```diff
@@ -0,0 +1,247 @@
...
+    - https://github.com/cardano-foundation/CIPs/pull/395
+Created: 2022-11-29
+License: CC-BY-4.0
+---
+
+# CIP-????: Decentralized WebRTC dApp-Wallet Communication
+
+## Abstract
+
+We want to introduce a decentralized communication method between dApps and wallets based on WebTorrent trackers and WebRTC. This CIP also contains a proof of concept implementation injecting the wallet rpc methods into the dApps global window object similar to [CIP-0030](https://github.com/cardano-foundation/CIPs/tree/master/CIP-0030).
```
L-as, line 22: I'm honestly having a hard time understand what this is supposed to mean.
What is a "dApp" here? What does communication between the dApp and wallet mean?
Do you mean dApp as in a website that hosts an interface to one?
Assuming so, what is the motivation? To connect to the website without using DNS?
It's not clear to me.

fabianbormann: Hi @L-as, 
yes the dApp is a website (I'm currently working on providing the cip-0030-like interface https://github.com/fabianbormann/cardano-peer-connect). The method described within this CIP aims to allow those web applications to connect peer to peer with the mobile wallet to call rpc functions like getRewardAddresses/signTx/signData etc. (without having so much implementation overhead) just by sharing the identifier (address) of the dapp (e.g. by using a QR Code) and without the need of a "central" relay server for the ip discovery. The method behind it can also be used to implement chats between peers (apps) and is not fixed to wallet<-->dapp communication (feel free to play around with it https://github.com/fabianbormann/meerkat).  

This is the demo repository that maybe can manage some expectations https://github.com/fabianbormann/WebRTC-WebTorrent-tracker-communication-demo/blob/main/dApp.html

CIP-????/README.md
```diff
@@ -0,0 +1,247 @@
...
+sequenceDiagram
+    dApp-->>+TURN/STUN: Who am I?
+    TURN/STUN-->>dApp: You are <ip:port>
+    Wallet-->>+TURN/STUN: Who am I?
+    TURN/STUN-->>Wallet: You are <ip:port>
+    dApp->>Wallet: messages
+    Wallet->>dApp: messages
+```
+
+The data will be send peer to peer via a WebRTC data channel once the ip discovery has been finished. E.g. WalletConnect expects/provides a relay server URL to initialize the connection. While this method allows dApps to communitcate peer-to-peer with wallets it also leads to a [possible SPOF](https://twitter.com/walletconnect/status/1407279853943001088?lang=en).
```
cleanerm5, line 42: Is there a chance that WalletConnect [v2 Client APIs](https://docs.walletconnect.com/2.0/specs/) can be implemented "on top" of this CIP? Basically I agree on that the CIP has some advantages over WalletConnect v2 in terms of decentralization and flexibility but in terms of adoption and interoperability it would be an asset to be compatible with those APIs so that DApps from other blockchains could more easily migrate to Cardano.

fabianbormann: Sure! This is the (work in progress) interface for a cip-0030-like api https://github.com/fabianbormann/cardano-peer-connect/blob/main/lib/wallet.ts but of course we can also add an interface for an api that matches the requirements of wallet connect, or that "translates" the wallet connect api to cip-0030 if needed .. and on the same time they could also just add the torrent tracker method additionally to their current peer discovery approach. The goal of this cip is not to replace wallet connect and also not to "challenge" their approach. It's more like an additional puzzle piece that can be used in different ways.

CIP-????/README.md
L-as: wdym central relay server?

fabianbormann: I mean [this one](https://webrtc.org/getting-started/turn-server) Please take a look at [the first figure](https://github.com/fabianbormann/CIPs/tree/master/CIP-%3F%3F%3F%3F#establish-a-webrtc-connection-using-a-stunturn-server-state-of-the-art-approach) right under the specification section within the CIP there I try to explain it also graphically. Just to break it down to a simple explanation: Both peers that wants to talk to each other are need a server component to establish the connection. It locates both peers (resolves the IPs behind [NATs](https://en.wikipedia.org/wiki/Network_address_translation)) and exchange this information to  both peers to "let them find each other" and start from this point to talk directly peer to peer without the server.

L-as: Sorry, what's better compared to the current situation where raw IP and DNS is used?

fabianbormann: Can you explain me your mental model of how you think that a dApp + wallet communication currently works? Because I dont understand the point with raw IP and DNS (sounds a bit like you are thinking of REST)

CIP-????/README.md
```diff
@@ -0,0 +1,247 @@
+---
+CIP: ?
```
```suggestion
CIP: 45
```

fabianbormann: Nice CIP-0045 sounds wonderful to me 😊 ! Meanwhile the following points along the implementation plan has been succeeded: 

1. The [bugout reimplementation](https://github.com/fabianbormann/meerkat) into a typescript and webpack 5 friendly lib has been done 
2. The [cardano-peer-connect lib](https://github.com/fabianbormann/cardano-peer-connect) has been created to simplify the implementation of CIP-0045 for dApp and wallet app developers
3. The [example implementation](https://github.com/fabianbormann/cip-xxxx-demo-implementation) has been refactored so that it now uses the peer-connect lib 

During the implementation it turned out that a we need to invest some more time into:

1. Testing different scenarios. For example the communication with a network setup using a symmetric NAT
2. If the dapp and the wallet  app runs on the same device or in the same subnet (without NAT) we can simplify the communication establishment
3. Heartbeats, reconnection and a dynamic update of the [WebTorrent tracker list](https://github.com/ngosang/trackerslist) would be useful to make the communication more solid

The next steps would be to provide a reference implementation within a cip-0030 compliant wallet app.

jehrhardt:
@fabianbormann I think the CIP is not really correct on the used term TURN server.

A [TURN server](https://webrtc.org/getting-started/turn-server) is required to do hole punshing. Let's say your mobile phone is connected via 5g and your wallet is on your laptop connected via WIFI. Then your wallet cannot be reached by your phone directly over the internet. That's why both will connect to a TURN server first to open a connection to the internet, which will then be used for p2p traffic.
Actually WebRTC without TURN server would only work within the same network or with devices open to public internet.

What is meant in this CIP is a [Signaling server](https://datatracker.ietf.org/doc/html/rfc8827#section-4), which you replace with a WebTorrent tracker. It is used to exchange the connection information and is independent of using a TURN server for hole punshing.

While this CIP is about establishing a connection via WebRTC, I wonder where the RPC protocol that is used between dapp and wallet is specified. I guess, it will be derived from CIP-30. But maybe it is worth to make it another CIP?

@hazae41 I thought about libp2p as well, when I read the title. But in libp2p you still need a way how wallet and dapp can establish a p2p connection. So a WebTorrent tracker could be used as well.

OlofBlomqvist:
> @fabianbormann I think the CIP is not really correct on the used term TURN server.
> 
> A [TURN server](https://webrtc.org/getting-started/turn-server) is required to do hole punshing. Let's say your mobile phone is connected via 5g and your wallet is on your laptop connected via WIFI. Then your wallet cannot be reached by your phone directly over the internet. That's why both will connect to a TURN server first to open a connection to the internet, which will then be used for p2p traffic. Actually WebRTC without TURN server would only work within the same network or with devices open to public internet.
> 
> What is meant in this CIP is a [Signaling server](https://datatracker.ietf.org/doc/html/rfc8827#section-4), which you replace with a WebTorrent tracker. It is used to exchange the connection information and is independent of using a TURN server for hole punshing.
> 
> While this CIP is about establishing a connection via WebRTC, I wonder where the RPC protocol that is used between dapp and wallet is specified. I guess, it will be derived from CIP-30. But maybe it is worth to make it another CIP?
> 
> @hazae41 I thought about libp2p as well, when I read the title. But in libp2p you still need a way how wallet and dapp can establish a p2p connection. So a WebTorrent tracker could be used as well.

To be a bit more precice, STUN servers are used for NAT hole-punching where ICE is not enough and TURN is used as last resort proxy relays between peers that cannot otherwise establish direct peer-to-peer connections.

The Bugout library mentioned uses Google's public STUN servers (via webtorrent->peer-connect) by default but also requires TURN servers in order to work for all cases.

Afaict this CIP does as you said only propose replacing custom signaling servers with webtorrent trackers, but we'd still rely on both STUN and TURN to exist or some peers will not be able to connect with each other. 

Looking at the [POC implementation](https://github.com/fabianbormann/cip-0045-demo-implementation) for the CIP, it does indeed look like it uses webtorrent (thru [meerkat](https://github.com/fabianbormann/meerkat)).

If so it sounds a bit misleading to say in the CIP that WebTorrent trackers are used **instead** of STUN/TURN.. unless im missing something?

fabianbormann:
Hey @OlofBlomqvist and @jehrhardt, thanks for coming up with your feedback. We are 98% done with the implementation of the p2p wallet connector (thru meerkat as @OlofBlomqvist mentioned). We tested wifi+cellular, wifi+wifi & cellular+cellular (browser<->browser) across different countries and all of our tests worked to inject the cip30 api with webrtc and webtorrent trackers from the wallet into the dapp. 

If I understood the bugout explanation correctly, then they use webtorrent trackers for the peer discovery instead of a single server which is the STUN or TURN (I thought) but if I understood it incorrectly (and if I used an incorrect wording) let's just have a quick sync meeting and iterate over the markdown file of this CIP and use the proper terms and words. 

I guess the overall goal to have a lightweight wallet <-> dapp communication for mobile wallets + dapps should not be affected by that discussion as we already see that it works in our different setups.

jehrhardt:
Hey @fabianbormann, I totally agree that it does not affect the overall goal. Just the wording in the CIP is misleading.

Your wifi+cellular test works, because meerkat is using Google's public TURN under the hood to establish the connection after using WebTorrent for the discovery. Otherwise it would technically not be possible to establish a connection between devices in both networks.

fabianbormann:
@jehrhardt & @OlofBlomqvist would be great to have a meeting with you to go a little bit deeper. Could you drop me a mail to fabian.bormann@cardanofoundation.org so that we can align on a date+time ? 😊 Or text me on Discord (abundzugamer#4587)

fabianbormann:
@jehrhardt it took a while, but I updated the CIP as discussed. Thank you very much for the productive conversation, and for your valuable suggestions! 😊

CIP-????/README.md
rphair: @L-as @fabianbormann without any further challenges I think the original question has been answered.

CIP-????/README.md
rphair: Good question @cleanerm5 which I trust has been answered.

CIP-????/README.md
```diff
@@ -224,24 +224,36 @@ The purpose of this CIP mainly consists of two parts. It addresses the current l
...
     1. A check for mobile/desktop environment
     2. Depending on the environment provide interfaces for CIP-0030 / and / or CIP-?
     3. Add a full implementation of the server/client side code above to define a communication standard similar to CIP-0030 (getRewardAddresses, signData, signTx, ...)
 - [ ] Start discussions about security gaps within the proposed method with various developers and also look for research papers
 - [x] Check if the wallet app also reacts to rpc calls in background mode on Android
-- [ ] Check if the wallet app also reacts to rpc calls in background mode on iOS
+- [x] Check if the wallet app also reacts to rpc calls in background mode on iOS
 - [ ] Implement the library within an example dApp and [boost wallet](https://github.com/boost-pool/boost-wallet)
 
+### Updates
```
rphair, line 245: These are good updates but editorially we can't allow a section for it here, since the comments would be out of flow with the standardised sections (and therefore difficult to process for readers & in "diffs")
- Some of these items would be notes for the Specification;
- notes about Implementations should be ticked boxes in the Implementation Plan above (I think they already are?)
- notes about the wording don't need to be referenced at all... we will see them from the PR's commit history... and you can also call attention to them in the PR comments to help reviewers focus on your new material 😎

rphair (commented):
@fabianbormann it looks like https://github.com/cardano-foundation/CIPs/pull/395/commits/e7ccca1a44e35312b632c7a0ea26863aec5a152c completes the recent requests for clarification so I'm adding this to the CIP Editors' meeting agenda (https://hackmd.io/@cip-editors/70 - currently scheduled 01 August 16:00 UTC) for re-review.  In the meantime:

CIP-????/README.md
```diff
@@ -0,0 +1,259 @@
...
+Implementors:
+    - Fabian Bormann <fabian.bormann@cardanofoundation.org>
+    - Jaime Caso <jaime.caso@cardanofoundation.org>
+Discussions: 
+    - https://github.com/cardano-foundation/CIPs/pull/395
+Created: 2022-11-29
+License: CC-BY-4.0
+---
+
+# CIP-????: Decentralized WebRTC dApp-Wallet Communication
```
Ryun1, line 18:
```suggestion
```
Remove title to fit updated CIP template :)

CIP-????/README.md
```diff
@@ -0,0 +1,259 @@
...
+```
+
+### Security Aspects
+
+We decided to spawn the server within the dApp to force the user to manually scan a QR code (using a wallet app) or accept an "Open with `<WalletAppName>`" ui dialog (in case of Universal Links or Deeplinks). This prevents the user from connecting the wallet to an unwanted dApp. Additionally we need to add  a few security checks to prevent a misusage of this method.
+
+- The wallet app needs to verifiy the origin (address) of the RPC call
+- dApps should ask the user for permission to inject the wallet names into the window.cardano object to prevent XSS attack (Maybe using a graphical representation of the wallet app address e.g. blockies)
+
+## Rationale
```
Ryun1, line 219: To me this rationale section is a bit light, id like to see answers too:
- Were there any alternate designs considered? Why did the current design succeed? 
  - maybe [#395 comment](https://github.com/cardano-foundation/CIPs/pull/395#issuecomment-1346314526) applies.
- What are the benefits and drawbacks of this approach when compared to the connection via CIP-30?

Ryun1 (commented):
Hey @fabianbormann, 

Very interesting proposal, im quite a fan!

Please see the editorial couple comments in this review. I do have one large question:
- How should this proposal be versioned? so that it could be iterated on in the future.

CIP-????/README.md
rphair: As I remember @fabianbormann you/we talked about some alternative design considerations in the CIP meetings around the time this was originally presented... so perhaps you could recall & include an inventory of those that came up?

Crypto2099:
Saw a message from Eternl about providing support for this standard so I decided to find it and learn more about it. Seems very interesting and a lot of exciting potential use cases. Primary amongst them being the ability to connect to a desktop or kiosk session from your mobile phone that has your wallet details.

Based on this "multi-device" bridging, I wonder if it would be good for an extension/new authority to CIP-13 here using the `web+cardano:` standard so that wallets integrating other QR information for things like staking or payments can easily connect and interact.

Just some rough thoughts but what about something like:


| [scheme] | [authority] | [version] | [data] |
|---|---|---|---|
| web+cardano: | //connect | /v1 | ?hash=<HASH provided by the dApp> |

`web+cardano://connect/v1?hash=ABC123`

(@rphair for visibility) :)

rphair:
@fabianbormann since you seemed to endorse the suggestion in https://github.com/cardano-foundation/CIPs/pull/395#issuecomment-1658797311 we agreed at the CIP editors' meeting today to wait for you to approve or refuse the suggestion to add some kind of URI scheme to what you already have here.

If making no further additions, the consensus is that this is ready to merge as-is... but we'll wait on any further editorial review until that last question is answered.

fabianbormann:
@rphair I agree and it makes 💯% sense. I don't know if I will manage to look into it this week (as I'm on holiday with the family 😉), but I will invest some time next week and add it to the cip.

stevenj:
> While this CIP is about establishing a connection via WebRTC, I wonder where the RPC protocol that is used between dapp and wallet is specified. I guess, it will be derived from CIP-30. But maybe it is worth to make it another CIP?
I came here to ask the same thing :)
I think we should have a way of standardizing CIP-30 + Extension APIs using Json RPC or similar serialization.  Which any wallet interface can use, so that we don't need to continuously redefine the wallet API's for different interface methods.

stevenj:
I would be concerned about a potential MITM attack here.
What would prevent the Webtorrent tracker from directing you to an adversary,  which then forwards your traffic where you expect it to go?
https://security.stackexchange.com/questions/84875/how-can-mitm-be-performed-in-webrtc

fabianbormann:
@hazae41 @stevenj We (@marcuspuchalla and I) decided to go with the solution of using an identicon in the center of the qr code. It's generated from both ids (dapp and wallet) https://github.com/fabianbormann/cardano-peer-connect/blob/main/src/DAppPeerConnect.ts#L572 .. I tried to play around with a few attack scenarios and as the communication itself goes via p2p and the qr code or identifier you are scanning/coping is the public key, it would not be able for a man in the middle to connect you to a different wallet. The identifier you've scanned/copied and the public key coming from the signed messages would be different. The mitm could not fake the signature matching to the public key that the dapp would expect from the wallet and therfore not start the handshake communication that would be necessary to establish the connection. But it would be nice if someone else could also try to attack the connection to make sure to think of as much attack vectors as possible.

rphair:
thanks @fabianbormann ... marking `Waiting for Author` due to https://github.com/cardano-foundation/CIPs/pull/395#issuecomment-1660836696 so we can give you all the time you might need for that & also for others to review security scenarios as per the more recent discussion.  I'll keep a close eye on this in any case & please also know when you think it's ready for review again 😎

fabianbormann:
Thanks @rphair 🙏 I started to look into **CIP13** and `web+cardano://connect/v1?identifier=ABC123` (as suggested by @Crypto2099) makes a lot of sense. I will add it to the Readme.md and to the cardano-peer-connect + demo application as soon as possible 😊

fabianbormann:
@rphair, I've included additional details about the identifier and the information relating to CIP13. Moving forward, my next steps will involve updating the [cardano-peer-connect repository](https://github.com/fabianbormann/cardano-peer-connect) and the [demo repository](https://github.com/fabianbormann/cip-0045-demo-implementation) regarding the URI scheme. Additionally, I'll be reaching out to the Eternl team to discuss these minor changes. As of now, the CIP itself should be unblocked and won't be impacted by these activities. 😊

rphair (approved):
@fabianbormann that looks good to me.  At first it appeared a requirement to submit an ABNF grammar for any CIP-13 URI form, but @SebastienGllmt @KtorZ it seems a bit harsh in hindsight to inflict that on everyone... I would rather leave it up to those extending the URI scheme (cc @Crypto2099) especially as it would depend on the complexity of the extended syntax.

So unless there are any arguments for an ABNF syntax (as there could be eventually even in the simple cases, if there were many different `//authority` possibilities, to help generate a general parser for URI links), in the meantime your addition passes my own review pending any objections on that subject.

KtorZ (approved):
👌 Not sure why we haven't merged that one already. Agree with @rphair last comment.

CIP-????/README.md
```diff
@@ -0,0 +1,276 @@
+---
+CIP: 45
+Title: Decentralized WebRTC dApp-Wallet Communication
+Status: Proposed
```
KtorZ, line 4:
```suggestion
Status: Active
```

rphair:
@fabianbormann @KtorZ I'll update the status to `Active` in the top level `README` in another PR since the version of that file here is stale.  Good to see this one finally on the books.

Issue #109, opened 2021-07-29
Broken cross-CIP links on derived website


qwertyuu:
Hello! Just browsing on this CIP 3 page: https://cips.cardano.org/cips/cip3/

When going through the "History" table, the names are clickable but they all go 404. When browsing the .md files here, they seem fine. There must be a way to properly link/host them!

Thanks :D

rphair:
@qwertyuu this is because the markup refers to items in the same Github directory as relative pathnames, as you can see here: https://raw.githubusercontent.com/cardano-foundation/CIPs/master/CIP-0003/CIP-0003.md

@crptmppt what's the best way of handling this?

1. try to have the build script parse out relative pathnames & prepend the CIP GitHub directory containing the CIP itself to those pathnames?
2. require absolute pathnames in the CIPs?

I would vote for (2) since it can be done retroactively for the small number of cases it would affect (like this one) and because it would make the raw GitHub CIP files more portable in general beyond the built environment.

crptmppt:
This issue was touched to  last week's Editor meeting (27)- see [notes](https://github.com/cardano-foundation/CIPs/blob/master/BiweeklyMeetings/08-03-2021.md#issue).
As noted, there are a lot of dependancies and this might have to wait until later this year (as there are a few cross links) 
next week's [8/17 Editor meeting 28](https://www.crowdcast.io/e/cip-editors-meeting-28). 
If the issue is of relevance to you, consider attending the meeting and speaking up, or adding here in the thread!

mangelsjover:
@KtorZ please find below the broken links which I've been able to identify:

Site: https://cips.cardano.org/cips/cip35/#specification
Broken Links:
https://github.com/cardano-foundation/CIPs/wiki/Comments:CIP-0001
[usual CIP process](https://cips.cardano.org/cips/cips/cip1/README.md)

Site: https://cips.cardano.org/cips/cip35/
Broken Link:
[usual CIP process](https://cips.cardano.org/cips/cips/cip1/README.md)

rphair:
the original path problems as reported in the OP were fixed by an earlier PR which fixed a lot of relative pathnames).  But another link on this CIP needed to be updated: 
https://input-output-hk.github.io/adrestia/cardano-wallet/concepts/cryptography-and-encoding

(doc being moved from Adrestia) to:
https://input-output-hk.github.io/cardano-wallet/concepts/cryptography-and-encoding

[...]

PR #399, opened 2022-12-03
CIP/CPS template & categories tweaks 

KtorZ:
Found that these were missing as I updated a few past CIPs.


rphair:
This is definitely an improvement but I'm still hoping (as posted in https://github.com/cardano-foundation/CIPs/issues/109#issuecomment-1333771465) there will be a way to link from Categories reference on CIP-9999 to that section on CIP-0001, without breaking links on the generated `cips.cardano.org`, so that CPS authors don't submit documents with invalid categories (as we've seen recently).

rphair (approved):

PR #272, opened 2022-06-06
[paradiso.app] reserve transaction metadatum labels

aarondunnington:
Greetings Cardano Foundation team,

The team at [paradiso.app](https://paradiso.app) would like to reserve our transaction metadatum entries in the CIP-0010 registry.

We are using the entries below for storing our onchain metadata.

```
{
    "transaction_metadatum_label": 1188,
    "description": "paradiso.app marketplace metadata"
},
{
    "transaction_metadatum_label": 1189,
    "description": "paradiso.app services metadata"
}
```

PR #405, opened 2022-12-06
[paradiso.app] remove transaction metadatum labels

aarondunnington:
Requesting removal of paradiso.app metadatum labels.

We are shutting down this month, and would like to free these labels up.

Original PR which reserved these labels is listed below for reference:
https://github.com/cardano-foundation/CIPs/pull/272

### DIFF
--- a/CIP-0010/registry.json
+++ b/CIP-0010/registry.json
@@ -32,14 +32,6 @@
     "description": "finitum.io token bridge transactions metadata"
   },
   {
-    "transaction_metadatum_label": 1188,
-    "description": "paradiso.app marketplace metadata"
-  },
-  {
-    "transaction_metadatum_label": 1189,
-    "description": "paradiso.app services metadata"
-  },
-  {
     "transaction_metadatum_label": 1564,
     "description": "Marlowe - a DSL for writing and executing financial contracts"
   },


rphair (commented):
Also authenticated by https://github.com/cardano-foundation/developer-portal/pull/874.

@aarondunnington (cc @KtorZ @SebastienGllmt) since this is the first metadatum label removal I've seen: is there not an advantage to leaving this label in place so the metadata trails already left can be traced back to Paradiso?

(**edit, p.s.**) if so, could this PR be changed so it leaves the label in place but adds a `Defunct` tag to the label text?

rphair:
@aarondunnington I'm tagging this as Waiting for Author since it's up for "triage" (initial review) at tomorrow's CIP meeting, but really we only need to identify the consequences of removing the label.  I don't see why it helps to remove the label if it already tags a number of `paradiso.app` transactions which would then be tagged by an unassigned number: even worse if as you say "freeing up" the labels allows them to be used by another transaction label in the future.

If you could progress this question before the meeting then we might be able to decide what to do with this request at the meeting, or soon after.  cc @KtorZ

aarondunnington:
Thanks for the question @rphair 

No material live transaction volume ended up using these labels.

Only two live minting transactions (single NFT each) utilized labeled 1188 in production.

All other transactions utilizing 1188 on mainnet were for internal paradiso.app prod validation testing.

No transactions utilized the 1189 tag.

rphair (approved):
Thanks, that helps us qualify this request since it doesn't seem like those transactions being orphaned will be a problem. 😎

KtorZ (approved):

PR #406, opened 2022-12-06
CIP-0052 | Add public keys to auditor table

simonjohnthompson:
Added field for public key information for auditors. This to be used to verify signed audit certification metadata.

### DIFF
--- a/CIP-0052/README.md
+++ b/CIP-0052/README.md
@@ -216,13 +216,13 @@ DApps are used by *DApp users*, and built by *DApp developers*. Audit is perform
 
 ### 3. Cardano auditors
 
-| Audit company   | URL                                     | Contact email      |
-| -----------     | -----------                             | -----------        |
-| FYEO Inc.       | https://gofyeo.com/blockchain-security  | sales@gofyeo.com  |
-| Hachi           | https://hachi.one                       | team@hachi.one        |
-| MLabs           | https://mlabs.city                      | info@mlabs.city      |
-| Runtime Verification           | https://www.runtimeverification.com | contact@runtimeverification.com  |
-| Tweag                | https://www.tweag.io                  | sales@tweag.io                  |
+| Audit company   | URL                                     | Contact email      | Public key      |
+| -----------     | -----------                             | -----------        | -----------        |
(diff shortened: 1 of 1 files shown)


rphair (commented):
@simonjohnthompson it's good to have that extra field in the table for sure... but is this PR complete without those keys filled in?  If not, should this PR be kept in abeyance until those keys are added in further commits?

L-as:
We should definitely add the public keys first. I suppose we can just comment in this PR.

rphair:
> I suppose we can just comment on this PR.

sure @L-as if they're posted here in a way that allows editors to verify their validity (e.g. from the auditor's GitHub account), we can commit them if the PR author hasn't done so already.

rphair:
@L-as @simonjohnthompson is [the last commit](https://github.com/cardano-foundation/CIPs/pull/406/commits/97cdc70d1c652a95136afd6030982de4f9c31f31) clear enough according to how people will be using the table?

L-as:
I'd say so.

KtorZ:
Suggestion: public keys here may be added over time by a representative of each relevant party, with a commit signed by the respective private key and from an account that belongs to the party's organization. 

How does that sound?

SebastienGllmt (approved):
Makes sense. Not sure there is a reason to block this PR waiting for other entities to add something. They can just make a followup PR if they want to

KtorZ (approved):

L-as:
> Suggestion: public keys here may be added over time by a representative of each relevant party, with a commit signed by the respective private key and from an account that belongs to the party's organization.
> 
> How does that sound?

It's not sound. There's a cyclic dependency. s/signed commit/PR by member of org/.

KtorZ:
> There's a cyclic dependency. 

What do you mean? 

- Alice puts key in the file. 
- Alice commits and signs the change with corresponding private key
- Alice opens a PR

where's the cycle exactly?

L-as:
That only tells us that the owner of the key consented, not that the owner is the correct one.

KtorZ:
Hence the second part: "and from an account that belongs to the party's organization"

PR #418, opened 2022-12-20
CPS-0004? | Spending Script Redundant Execution

fallen-icarus:
([rendered proposal in branch](https://github.com/fallen-icarus/CIPs/tree/full-eUTxO-potential/CPS-%3F%3F%3F%3F#readme))


rphair:
@fallen-icarus we need to reduce the amount of non-essential words from the title while still covering exactly what this is about.  Would it be accurate & complete to title this _Spending Script Redundant Execution_?  If not, then something else with a small number of meaningful words?

fallen-icarus:
@rphair *Spending Script Redundant Execution* works for me. I'll change the title.

L-as:
I'm not sure this is an actual problem? If you have this problem, your script should rather than doing the computation, check that you withdraw 0 ADA as rewards from a helper script that checks the computation.

fallen-icarus:
@L-as I'm not sure what you mean. This has nothing to do with rewards. Can you please elaborate on what you mean?

L-as:
> @L-as I'm not sure what you mean. This has nothing to do with rewards. Can you please elaborate on what you mean?

It's a trick to solve exactly the issue you describe. You can always withdraw 0 ADA.

fallen-icarus:
> It's a trick to solve exactly the issue you describe. You can always withdraw 0 ADA.

My understanding of this trick is the following:
| Script A | Script B|
|--|--|
| checks whole transaction | only checks if Script A was executed |
| has ADA that is never actually withdrawn | has any number of utxos |
| executed once per transaction | executed once per utxo from Script B |

If this is correct then this trick does not solve the issue. Script B will still redundantly check if Script A was executed. **This does not stop redundant script executions; it only minimizes the cost of those redundant executions.**

Also this trick only works well with centralized smart contracts. I have created a DEX proof-of-concept (that I will open source soon) where users maintain full delegation control of their assets at all times. As a design requirement for this feature, all users have their own swap contracts. Users can then lookup all available swap contracts to remotely interact with other users' swaps. I have already extensively tested the DEX and it works as expected. It also has other nice properties such as being naturally concurrent (there are at least as many utxos as there are users) and therefore doesn't require batchers. The only issue is having to tolerate the redundant executions.

If I were to use this trick with the DEX, it would literally double the number of smart contracts required. Not only would this dramatically over-complicate the DEX design, but it would also create bottlenecks in the lookup process of finding all the necessary information for users to interact with the DEX. It is also worth pointing out that each user needs to store their reference scripts in each of their swap contract address. Right now this is like a deposit of 24 ADA in order to open a swap position. Using this trick would increase the necessary deposit each user needs to make per swap position.

The above drawbacks aren't an issue for centralized smart contracts. But these centralized smart contracts also have the drawbacks of:
1. Users must give up delegation control.
2. Batchers (middle-men) are required to overcome the lack of concurrency.

It is my strong opinion that, given the redundant executions incentivize this trick (in the pursuit of lower fees) and this trick strongly favors centralized smart contract designs, this actually further highlights why these redundant executions are a problem.

rphair:
@fallen-icarus I really hope you will include this last explanation in your CPS somehow.  Cardano has a great potential (compared to other L1's) to support fully distributed architectures and the DEX example is an excellent way to bring this problem statement into active discussion & solution.

fallen-icarus:
@rphair I will look to edit the pull request this weekend and incorporate it. I was not aware of this trick before so thank you @L-as for mentioning it.

L-as:
I'm not sure I understand your issue, but you definitely shouldn't have a script per user. I'm not sure how you came to that conclusion. I'm 'aso not sure how this pertains to batchers, or delegating,  or centralisation.

In any case, I don't think the fee overhead from this trick is worth messing so heavily with how validators work.

fallen-icarus:
@L-as Imagine you have the following setup:
| Script A | Script B |
|--|--|
| checks whole transaction | checks if script A is executed |
| has 5 utxos at the address | the number of utxos is user dependent |
| executed once per utxo from this address | executed once per utxo from this address |
| script is used as a reference script | script is used as a reference script |

Now imagine that Alice and Bob both have 5 utxos locked by Script B (10 utxos in total). What would happen in the following contexts?

1. Alice and Bob want to send all their utxos together in the same transaction
2. Alice and Bob want to send each of their utxos individually (10 txs in total)

In the first context, there is no major problem:

1. One utxo is used from Script A to force its execution 
2. Script B redundantly checks if Script A is executed once for every utxo (10 times) but the cost of each check is minor.

In the second context, there is now a bottleneck problem: each of the 10 txs need to use one utxo from Script A to force its execution but there are only 5 utxos available. In this context, batchers are required to use the limited number of Script A utxos efficiently. 

You can increase the number of utxos at Script A but the concurrency will always be limited to the number of utxos at Script A. Given the world that blockchain hopes to achieve where everyone is using DeFi, would 1000 utxos be enough? What about 10000? If you guess wrong, bottlenecks will cause issues that only batchers can solve.

*The reason for this bottleneck is due to Script A not being a real reference script. It is basically a frankein-script: half-reference, half-real.* By this I mean that a completely unrelated (to the transaction) utxo needs to be consumed in order to "reference" Script A.

---
Now imagine this counter setup:

| Script Alice | Script Bob |
|--|--|
| checks whole transaction | checks whole transaction |
| the number of utxos is up to Alice | the number of utxos is up to Bob |
| executed once per utxo from this address | executed once per utxo from this address |
| script is used as reference script | script is used as reference script |

Imagine Alice has 5 utxos locked by Script Alice and Bob has 5 utxos locked by Script Bob. What would happen in the same two contexts before?

In the first context, there are annoying redundant checks that cost about 0.3 ADA each but it works:

1. 5 utxos reference Script Alice and 5 utxos reference Script Bob
2. The total fee after the redundant checks is 3 ADA. If there were no redundant checks, the fee would be 0.6 ADA (0.3 for Script Alice + 0.3 for Script Bob)

In the second context, there are no problems at all (there are no redundant checks since they are separate txs):

1. 5 transactions reference Script Alice for a total fee of 1.5 ADA.
2. 5 transactions reference Script Bob for a total fee of 1.5 ADA.

These 10 txs can go through at the same time because their is no concurrency bottleneck. Script Alice and Script Bob are always  available since they are being used as true reference scripts.

---
The second setup's maximum concurrency is directly tied to the the number of utxos each user wants to use. In other words, the more people want to use it, the more concurrency is naturally available.

The first setup (which is the trick) has the concurrency "hardcoded" by the number of utxos at Script A. This concurrency does not naturally grow or shrink with the demand. In low demand contexts, there are likely to be too many utxos at Script A which results in inefficient use of the ADA locked at Script A. While in high demand contexts, there are likely to be too few utxos at Script A which means batchers will need to be used to compensate. The backlog that SundaeSwap suffered is a perfect example. Backlogs only exist when the concurrency is too low for the demand. This doesn't even deal who decides how many utxos are available from Script A which is another whole "can of worms".

In short, this trick sacrifices concurrency in order to minimize the cost of redundant executions. To compensate for this drop in concurrency, centralizing solutions are used, such as batchers. That is how this pertains to concurrency and centralization.

fallen-icarus:
@L-as 
> you definitely shouldn't have a script per user.

I have officially open sourced my proof-of-concept which can be found [here](https://github.com/fallen-icarus/cardano-swaps). Given your stance, I would like to know your thoughts on it.

L-as:
@fallen-icarus 
1) Liquidity pools can be designed such that they are inherently decentralised, concurrent, and stakeable (to a liquidity provider chosen SPO).
2) Your design is sensible, but
  a) I don't see why you have a script per user. This also messes with reference scripts. Just put the information in a datum.
  b) Your design seems not to need to be on-chain, because there is no global state. You could do the swaps using state channels instead, or at the very least just do normal atomic swaps. There is the question of how to broadcast available swaps, for this you might want a sidechain (easy but nonoptimal), but you could also have a simple system where a P2P network disperses trades by each node broadcasting an open trade to its peers. You can make it such that it's profitable to pass on an open trade by cryptographically securing a fee if the trade is closed (through you). This is much more useful and practical on Cardano than e.g. Ethereum because assets are native and we have UTXOs.

L-as:
Anyway, I don't get how this pertains to the proposal, and would say that maybe what you want is to optimise the script for checking the 0 ADA withdrawal. AFAICT though this can be a script of around ~100 bytes if not smaller. #309 is also relevant.

fallen-icarus:
@L-as
>  I don't get how this pertains to the proposal

From my perspective, your view that redundant executions aren't a real issue stems from your set beliefs about how Dapps can and cannot be created. I thought if I showed you concrete proof of another (and in my option better) way, I could change your mind about this. I actually built the DEX so people can try it out. It has all the following desirable features:

1. Full user delegation control
2. No impermanent loss
3. Naturally concurrent
4. Easily upgradable
5. Easily integratable into all wallet software

Again in my opinion, these features make this proof-of-concept probably the best DEX that so far exists in DeFi. **The only thing holding it back is the redundant executions.** Without the redundant executions, it would also be much cheaper than any DEX using the "withdraw 0 ADA" trick.

As for your points:

Point 2 is addressed in its entirety in the README. Both the reference scripts and the available swaps are easily broadcasted by what I call [Beacon Tokens](https://github.com/fallen-icarus/cardano-swaps#beacon-tokens). There is no need for any of the complicated techniques that were mentioned. It easily works on Cardano as is. As a bonus, these Beacon Tokens can be generalized to any Dapp. This means broadcasting any information to other users is trivial for any application. Also the need for a global state is not the only reason to use a blockchain. The blockchain is also meant for any peer to peer application where users do not trust each other. This latter reason is why my DEX is on-chain. There is no need for state-channels; there is plenty of room on layer 1.

With the current design of the DEX, there is basically a deposit required for opening swaps which comes from having the user store his/her reference script on-chain to then be found with the Beacon Tokens. This deposit also disincentivizes opening empty swap addresses. However, if the deposit is deemed to be too high (currently about 20 ADA), I also mention a potential adaptation of the design at the [end](https://github.com/fallen-icarus/cardano-swaps#potential-adaptation-of-the-design) of the README that does not have this high of a deposit requirement. I am mentioning this to point out that, combined with the Beacon Tokens, there are no drawbacks from giving users their own scripts/addresses.

As for point 1, I have looked at every DEX I could find that uses liquidity pools and none of them meet the gold standard of users maintaining full delegation control. By this I mean that, while both Alice and Bob use the DEX:

1. Alice can delegate her assets to ANY pool she wants.
2. Bob is not forced to delegate to the same pool as Alice. He too can delegate to ANY pool.
3. No one is forced to pick from a chosen subset of pools.

> Liquidity pools can be designed such that they are inherently decentralised, concurrent, and stakeable (to a liquidity provider chosen SPO).

This still sounds like it violates the third requirement of the standard. From a first principle perspective, I am very skeptical of the idea that liquidity pools can ever be designed in a way to match the level of decentralization obtained from full user delegation control. Unless you can directly point to an actual liquidity pool based DEX that meets the above standard (including the third requirement), we will just have to agree to disagree.

In short, it seems to me that there are three consequences (that I know of so far) to having these redundant executions:

1. Wasted computational resources that result in higher fees.
2. It incentivizes a "trick" that creates concurrency bottlenecks which requires batchers to handle.
3. It holds back, in my strong opinion, better designed Dapps that can't compete with the fees of Dapps using the "trick" in consequence 2.

You keep suggesting to use the "withdraw 0 ADA" trick and I keep explaining why I do not believe this trick is a good solution: it leads to more centralized designs. I strongly disagree with the view "we should just use the trick instead of fixing the redundant executions". Also the problem is not that the execution costs are too high; the problem is that there are redundant executions being done so the fee compounds. 0.3 ADA is already extremely cheap to check a whole transaction. If I wanted to, converting my script to be used like the trick suggested is not an issue.

Therefore, #309 would not be useful. However, I think my proposal would actually solve #309  issue. My proposed solution allows a plutus script to be able to handle both the case where a datum is missing and the case where a datum is present; no proxy-scripts are necessary. It does this without having to create an on-chain encoding for a missing datum. I know there are other proposals motivated by the no datum issue. I will try to aggregate them and add to my proposal how they are easily solved by my proposed solution.

I do not mean to come across as shilling the DEX proof-of-concept. There is actually no way for me to personally profit from others using the DEX, aside from new people coming to Cardano to use it. (Cheers to being properly incentivized lol). It is for this reason that I think it is okay for me to discuss it. I personally believe it adds weight to the need for this proposal but if others feel the DEX does not help explain why the fix is needed, I can drop it from the discussion.

rphair:
> if others feel the DEX does not help explain why the fix is needed, I can drop it from the discussion.

In my opinion, the divergence of opinion is exactly why it needs to remain in the discussion and certainly in the CPS itself (I said that above, but would say so even more emphatically now).  @fallen-icarus - others are going to want to follow your more "granular" (I don't know if that's the right word?) approach even if trade practice suggests otherwise.

Even if it were only an academic consideration it would still have a place in this CPS: editorially I can say that our overall process is about academic concepts finding implementations: sometimes gradually.

L-as:
I have a lot of thoughts, but I will focus on the biggest issue:
- I don't see how this trick pertains to batching **at all**. Using this trick has nothing to do with batching, centralisation, or protocol design. Given a script s, you can turn it into s' that checks that 0 ADA is withdrawn from s'' with the same logic as s. This is a transformation that is valid for any script.

fallen-icarus:
@L-as You are not accounting for the utxo ID changing every time the "withdraw 0 ADA" script is executed. On Cardano, a script will NOT execute unless a utxo from that script's address is consumed. It doesn't matter if there was no value taken. The old utxo will be destroyed and a new utxo with a different utxo ID will be produced. This will happen even if 0 ADA is actually taken from that script's address. It is NOT possible to successfully execute a script WITHOUT changing the utxo ID of the utxo that was used to force its execution. 

This is completely different to the way actual reference scripts work. Actual reference scripts assume all utxos actually being used are at the address that belongs to the reference script. To have two different scripts means you have two different addresses. In order to execute both scripts in each transaction, at least one utxo must be consumed from each address. This means the utxos IDs at the "withdraw 0 ADA" script will be constantly changing.

L-as:
@fallen-icarus OK, I think I understand: You mistakenly believe that you need a UTXO to use the withdraw 0 trick. This is not the case. There is no UTXO at all locked by the script (or using the script as staking address, but even if there were, it doesn't matter).
The trick simply involves checking that a stake script is withdrawn from. The amount doesn't matter (but it's important to note that it can be 0).

fallen-icarus:
@L-as What is the `cardano-cli transaction build` option for this? Is it `--auxilliary-script-file`?

fallen-icarus:
So I have been testing using the `--auxiliary-script-file` option with the dex but I can't seem to find a way to get the auxiliary script to be seen by the plutus script. It doesn't seem to be in the `ScriptContext` that plutus scripts see.

I don't believe I mentioned this before but the dex uses composable atomic swaps. Do you want to convert 10 ADA into 5 WMT and 5 DUST? No problem; you can do this in one tx. Do you want to convert 5 ADA and 5 WMT to 3 AGIX and 7 DUST? Again no problem. This can also be done in the same tx.

Using composable atomic swaps means, in one tx, you can easily do:
``` Haskell
many assets -> many assets
```
In order to maintain composability, there is a separate auxiliary script for each possible trading pair (easily created on the fly by users). Without the atomic swap contracts being able to verify that the required auxiliary script was executed, the security of the composition is completely broken.

@L-as Am I missing something else that would allow me to do this?

fallen-icarus:
@L-as After thinking about it more, I'm starting to think the auxiliary script is not what you mean.

> your script should rather than doing the computation, check that you withdraw 0 ADA as rewards from a helper script that checks the computation.

> The trick simply involves checking that a stake script is withdrawn from. The amount doesn't matter (but it's important to note that it can be 0).

> Given a script s, you can turn it into s' that checks that 0 ADA is withdrawn from s'' with the same logic as s.

Do you mean:
1. The "helper script" is literally a staking script, not a spending script.
2. The logic I need executed is done by the staking script.
3. The spending script then checks the `txInfoWdrl` part of the `ScriptContext` to see if the staking script was executed and 0 ADA was withdrawn from the reward address.

Is this correct?

L-as:
That's what I've been saying from the start. I have not misspoken.

fallen-icarus:
@L-as 
> I have not misspoken.

I know. The misunderstanding was my fault. I didn't realize you could literally use a staking script as a spending script like this. So whenever you mentioned rewards or staking, I thought you just didn't understand my issue. Apologies for this misunderstanding.

I will test this today and comment with what I find.

fallen-icarus:
Okay so I was able to get everything working using the trick. However, despite the dex technically still working, I believe this CPS should still be pursued for the following reasons:

1. This trick means everything requires two scripts now instead of just one. While the spending script can be made quite small, this is still a drawback IMO.
2. This trick does not stop the redundant executions since the spending script will still redundantly check if the required staking script was executed. This trick just minimizes the cost of those redundant executions.
3. This is not how staking scripts were intended to be used.
4. My proposed solution to the redundant executions also fixes the datum issues other developers have been struggling with: it would make it possible to spend a utxo at a script address despite the utxo missing a datum or having the wrong datum. I believe the solution does this without opening up any security holes or requiring a change to the way things are stored on-chain. The only thing that would change would be the way plutus scripts are executed.

> I don't think the ... overhead from this trick is worth messing so heavily with how validators work.

@L-as Maybe reasons 1, 2, and 3 alone are not enough of a reason to change how validators work. But when you also consider the possibility of reason 4, I believe it is worth doing. Do you?

L-as:
Why would you want 4.?

fallen-icarus:
Having a utxo locked forever due to a missing or incorrect datum is overly stringent. If this can be relaxed, it should be.

- There are a number of CIPs and stack exchange questions dedicated to exactly this restriction. 
  1. #309
  2. #364
  3. #310
  4. [stack exchange](https://cardano.stackexchange.com/questions/9257/is-pure-ada-put-in-a-script-address-locked-up-forever/9270#9270)
  5. [stack exchange](https://cardano.stackexchange.com/questions/4313/how-to-handle-datum-in-a-decentralized-way/4320#4320)
  6. [stack exchange](https://cardano.stackexchange.com/questions/337/validators-for-asset-type)
- Reading the comments in #309 seems to strongly suggest a consensus that datum-less execution of plutus scripts is desired. The only issue is on how to do this. My proposal does not seem to suffer from any of the issues mentioned with the previous suggestions.
- Humans are always liable to make mistakes. A mistake like forgetting a datum is very easy to make. In the future, regulators are not going to like this. This could literally end up being an obstacle for mainstream adoption.

As Dapps get more complicated, the necessity for simple building blocks becomes even more important for properly reasoning about security. Any design feature that unnecessarily adds complexity is a huge risk. Therefore it is not a good idea to "settle" for more complicated solutions instead of just putting in the work to allow for the simplest solution.

fallen-icarus:
I am going to update the pull request today to reflect what has come about from this discussion. That way it can potentially be discussed at the next CIP meeting and we can get more opinions.

L-as:
As noted in https://github.com/cardano-foundation/CIPs/pull/364, choosing an arbitrary default is not a good design decision and is a workaround for bad tooling. I am very doubtful that such a change will happen. Humans should never be sending to scripts manually.

fallen-icarus:
@L-as 
> As noted in https://github.com/cardano-foundation/CIPs/pull/364, choosing an arbitrary default is not a good design decision and is a workaround for bad tooling.

My proposal is not looking to use an arbitrary default. The idea is to use
``` Haskell
mkSpendingScript :: Maybe Datum -> Redeemer -> ScriptContext -> Bool
```
Others have mentioned this to be a valid potential solution as long as the representation can be solved.

`Nothing` would be used instead of a valid `Datum`. I am currently trying to find a way to represent that `Nothing` on a low level if it is actually necessary. I do not know how `Nothing` works on a low level in Haskell but if possible, I do not believe a valid `Datum` should be used if a low level representation is necessary. An option I am currently trying to check is if it is possible is to use an invalid de bruijn index like (-1). This would guarantee a valid `Datum` is not removed from the space of possible `Datum`.

I am open to other ideas if you have any.

> Humans should never be sending to scripts manually.

This seems to just be arbitrarily limiting the possible applications for scripts. If the proper safe guards were put in place, this should be fine and would vastly improve decentralized usage without adding complicated proxies.

fallen-icarus:
I came up with another low level specification for `Maybe Datum` that I believe is better so I will be changing the proposal to use it.

fallen-icarus:
@rphair Do I need to mark this as "Ready for review" in order for it to be discussed at a CIP meeting?

rphair:
@fallen-icarus in my observation your proposal already *is* being reviewed and actively incorporating community suggestions: so it could be promoted here on GitHub to reflect this reality, and should (unless you want to incubate it longer).

We've presented still marked as "Draft" pull requests at CIP meetings, but so far only in the "Triage" phase where a new proposal is introduced for brief comments & not a detailed review.  In case there is spontaneous discussion during the Triage phase (which sometimes happens, time permitting) you could invite more comment both at the next & future meetings by taking this out of "Draft" phase.  (cc @KtorZ in case he wants to add anything)

BTW (I just found this out myself) the CIP meeting scheduled for tomorrow 10 January has been postponed exactly 1 week: same day & time on 17 January.

fallen-icarus:
@L-as I have added the low level specification for how I think the `Maybe Datum` should work. It does not use a default datum value and should not require changing much about the way datums are used. When you have time, can you please review this section (the very last section titled "How would the `Maybe Datum` be represented at a low level?") and let me know your thoughts?

rphair:
@fallen-icarus @L-as at our CIP meeting tomorrow (if nothing else happens in the meantime) I'm going to suggest:
- we take this proposal about of Draft mode (mark "Ready for review")
- that authors, advocates & antagonists try to contribute to the debate above about how & if the suggested workarounds are necessary.

fallen-icarus:
@rphair I'm not sure if you got a chance to see but this conversation is also taking place in #321. I believe #364, #423, #309, #310, #321, and this proposal are all related. While I am fine with taking this CPS out of Draft mode, ultimately I think these 6 proposals should be merged if this is going to be pursued. My last comment in #321 explains my view. Unfortunately, I will not be able to make the meeting but I can answer any questions here or in #321.

rphair:
thanks @fallen-icarus - I didn't have enough technical knowledge to relate them in that way.  Let's leave it the way it is then for now & I'll reference your latest comment in the meeting.  All our agenda items are "Triage" and many aren't new CIPs, so we should have time to check & talk a bit about how these items relate.  cc: @KtorZ

CPS-????/README.md
```diff
@@ -0,0 +1,242 @@
...
+  --spending-reference-tx-in-redeemer-file $swapRedeemerFile \
+  --tx-out "$(cat ../assets/wallets/01.addr) 2000000 lovelace + 300 c0f8644a01a6bf5db02f4afe30d604975e63dd274f1098a1738e561d.54657374546f6b656e0a + 0 c0f8644a01a6bf5db02f4afe30d604975e63dd274f1098a1738e561d.4f74686572546f6b656e0a" \
+  --tx-out "$(cat ${swapScriptAddrFile}) + 4000000 lovelace + 0 c0f8644a01a6bf5db02f4afe30d604975e63dd274f1098a1738e561d.54657374546f6b656e0a + 250 c0f8644a01a6bf5db02f4afe30d604975e63dd274f1098a1738e561d.4f74686572546f6b656e0a" \
+  --tx-out-inline-datum-file $swapDatumFile \
+  --tx-in-collateral bc54229f0755611ba14a2679774a7c7d394b0a476e59c609035e06244e1572bb#0 \
+  --change-address $(cat ../assets/wallets/01.addr) \
+  --protocol-params-file "${tmpDir}protocol.json" \
+  --testnet-magic 1 \
+  --out-file "${tmpDir}tx.body"
+```
```
KtorZ, line 40: I don't think that showing this particularly lengthy command excerpt really serves the purpose of highlighting the problem. I'd suggest to simply go with prose of a succinct diagram showing how multiple inputs locked by the same script results in the script being executed multiple times.

KtorZ, line 40: I don't think that showing this particularly lengthy command excerpt really serves the purpose of highlighting the problem. I'd suggest to simply go with prose of a succinct diagram showing how multiple inputs locked by the same script results in the script being executed multiple times.

CPS-????/README.md
```diff
@@ -0,0 +1,242 @@
...
+
+Being that the atomic swap script checks all the inputs and outputs in the transaction to determine validity, below is the transaction fee estimation with the current limitation:
+
+``` Txt
+tx fee = 0.3 ADA * #swap_ref_script_executions + 0.02 ADA * ( #input_utxos + #output_utxos)
+``` 
+
+As the equation shows, when the swap script must be exected once per UTxO, the fee can grow quite large very quickly. Meanwhile, if the redundant exectutions were not there, the fee would grow very slowly.
+
+## The Trick To Minimize This Limitation
```
KtorZ, line 54:
```suggestion
### Current alternative
```

CPS-????/README.md
```diff
@@ -0,0 +1,242 @@
...
+  --tx-out-inline-datum-file $swapDatumFile \
+  --withdrawal "$(cat ${stakingScriptAddrFile})+0" \
+  --withdrawal-script-file $stakingScriptFile \
+  --withdrawal-redeemer-file $swapRedeemerFile \
+  --tx-in-collateral bc54229f0755611ba14a2679774a7c7d394b0a476e59c609035e06244e1572bb#0 \
+  --change-address $(cat ../assets/wallets/01.addr) \
+  --protocol-params-file "${tmpDir}protocol.json" \
+  --testnet-magic 1 \
+  --out-file "${tmpDir}tx.body"
+```
```
KtorZ, line 86: Same remark as above. The cardano-cli command is rather opaque and doesn't highlight well what the workaround is about. I'd suggest to reformulate as prose or a diagram.

CPS-????/README.md
```diff
@@ -0,0 +1,242 @@
...
+3. Try to use the ledger as is. The way scripts are stored onchain should not change unless absolutely necessary.
+4. Try to leave the developer experience as untouched as possible. We don't want to make developers re-learn plutus.
+5. Do not add too much new overhead to already resource constrained plutus scripts.
+
+## Open Questions
+1. Would this require a hardfork?
+2. Would this require a new version of plutus?
+
+---
+## Possible Solution
```
KtorZ, line 120: I'd advise moving this into a new CIP document.

KtorZ (commented):
The problem makes sense and is well formulated. I'd advise however to split the current document into two: the current CPS and a CIP containing the proposed solution.

Indeed, most of the debate seems to be more around the proposed solution than around the problem statement.

CPS-????/README.md
fallen-icarus: I eliminated all extraneous parts of the transaction. Hopefully this makes it more approachable.

CPS-????/README.md
fallen-icarus: I removed the proposal and will look to create a new CIP.

CPS-????/README.md
```diff
@@ -0,0 +1,81 @@
...
+I tried to test Cardano simple scripts to determine if this problem also impacts them. However, given that simple scripts are so small, it was hard to tell from just looking at the transaction fee. Since the interface for using Cardano simple scripts is the same as for using plutus scripts, I have to assume that this problem also impacts simple scripts. 
+
+If simple scripts are indeed impacted by this problem, the above technique would not work for them. The technique only works for plutus scripts.
+
+## Use Cases
+Any time a spending script's validation depends on the entire transaction and not on any particular UTxO.
+
+1. P2P Atomic Swaps - within one transaction, the value entering the swap address must be some proportion of the value leaving the swap address
+2. P2P Lending - within one transaction, the amount borrowed must be some proportion of the amount posted as collateral. Likewise, the amount of collateral reclaimed depends on how much of the loan is repaid.
+3. DAOs
```
michaelpj, line 68: This is far too loose. It seems false for many examples that I can think of? In particular, any application that relies on the script knowing its own address relies on the ability to know which script is executing.

CPS-????/README.md
```diff
@@ -0,0 +1,81 @@
...
+## Goals
+1. Stop redundantly executing transaction level spending scripts.
+2. Do not remove the option to allow script execution per UTxO. There are still use cases where the script should validate based off individual UTxOs. Also, there could be cases where this option is more efficient than having to traverse the transaction's context to get the relevant datum. An example of this latter case is when only one UTxO is consumed from a script address.
+3. Try to use the ledger as is. The way scripts are stored onchain should not change unless absolutely necessary.
+4. Try to leave the developer experience as untouched as possible. We don't want to make developers re-learn plutus.
+5. Do not add too much new overhead to already resource constrained plutus scripts.
+
+## Open Questions
+1. Would this require a hardfork?
+2. Would this require a new version of plutus?
```
michaelpj, line 81: The answer to both of these questions is almost certainly yes.

CPS-????/README.md
```diff
@@ -0,0 +1,81 @@
...
+  the weighted average asking price of all UTxOs being spent from the swap address
+```
+
+This brings us to the problem: **there is currently no way to efficiently use a plutus spending script that validates based off of the transaction as a whole**. (These scripts will be called *transaction level spending scripts* for the rest of this CPS). With the current design of Cardano, if Bob consumed UTxO 2 and UTxO 3 in the same transaction to satisfy his desire for 25 ADA, the atomic swap contract will be executed twice. Since the transaction context doesn't change between executions, the second execution is completely redundant.
+
+Any time a developer creates a transaction level spending script, it will be redundantly executed for every additional UTxO consumed from the corresponding script address. These redundant executions are a waste of scarce resources and result in much higher fees for end users. Not only that, but the inability to efficiently use these transaction level spending scripts significantly handicaps the use of the eUTxO model. The eUTxO model already allows for creating such spending scripts; developers are just unable to properly use them.
```
michaelpj, line 39: I'm not sure you've quite argued that it's a _significant_ handicap :)

CPS-????/README.md
```diff
@@ -0,0 +1,81 @@
...
+1. This technique now means that everything requires two plutus scripts instead of just one. While the spending script can be made extremely small, this doesn't change the fact that two are now needed.
+2. This technique DOES NOT stop the redundant executions; the spending script will still redundantly check if the required staking script was executed. This technique just minimizes the cost of the redundant executions by moving the heavy computation into a script that is only executed once per transaction.
+3. This is not the intended use for staking scripts.
+
+### Simple Scripts
+I tried to test Cardano simple scripts to determine if this problem also impacts them. However, given that simple scripts are so small, it was hard to tell from just looking at the transaction fee. Since the interface for using Cardano simple scripts is the same as for using plutus scripts, I have to assume that this problem also impacts simple scripts. 
+
+If simple scripts are indeed impacted by this problem, the above technique would not work for them. The technique only works for plutus scripts.
+
+## Use Cases
```
michaelpj, line 63: I'd like to see a lot more detail here, of the level that there is for the swap example. At the moment it looks to me like there is _one_ convincing usecase, which isn't much.

CPS-????/README.md
```diff
@@ -0,0 +1,81 @@
...
+
+Any time a developer creates a transaction level spending script, it will be redundantly executed for every additional UTxO consumed from the corresponding script address. These redundant executions are a waste of scarce resources and result in much higher fees for end users. Not only that, but the inability to efficiently use these transaction level spending scripts significantly handicaps the use of the eUTxO model. The eUTxO model already allows for creating such spending scripts; developers are just unable to properly use them.
+
+### The Current Alternative
+The alternative involves taking advantage of the fact that staking scripts do not suffer from the same limitation as spending scripts.
```
michaelpj, line 42: I think the explanation for why this works is wrong. Staking scripts _do_ get information about what script is running and _do_ get redeemers. 

The reason this works is because you get all your spending scripts to coordinate on delegating to a _single_ other script execution. You could do exactly this trick with another spending script: spend an additional UTXO locked with a script that does the "real work", and have all the other UTXO spending scripts just assert that that script is being spent in the transaction. Same effect!

michaelpj (commented):
I somewhat disagree with the framing of the problem. The key thing you need in order to share evaluation of scripts is that they must not have access to any information that directly pins down the script being evaluated ("indexical information"). Specifically, the script context tells you this, and the redeemer and datum are also script-specific. So I think what you want is something like "non-indexical scripts", which are precisely scripts that don't get passed that information.

I think this is more helpful than the current framing which focusses on the fact that the scripts just look at the transaction (which of course they can already do), which it makes it sound a bit like we need new functionality instead of to restrict existing functionality.

----

I think the problem described in this CPS is a real problem, although I'm not sure how important it is. In particular, the benefit accrues only when you have scripts that really are identical and don't require any indexical information. I'm not convinced that those are _that_ common, and I'd like to see much more discussion of other usecases that would benefit.

CPS-????/README.md
```diff
@@ -0,0 +1,81 @@
...
+
+But what if he wants to swap his DUST for 25 ADA? That would require using at least two of Alice's UTxOs. Doing this is trivial for the atomic swap contract because it validates based off of the transaction as a whole: 
+
+``` Txt
+The value being deposited to the swap address must be >= 
+  the value withdrawn from the swap address * 
+  the weighted average asking price of all UTxOs being spent from the swap address
+```
+
+This brings us to the problem: **there is currently no way to efficiently use a plutus spending script that validates based off of the transaction as a whole**. (These scripts will be called *transaction level spending scripts* for the rest of this CPS). With the current design of Cardano, if Bob consumed UTxO 2 and UTxO 3 in the same transaction to satisfy his desire for 25 ADA, the atomic swap contract will be executed twice. Since the transaction context doesn't change between executions, the second execution is completely redundant.
```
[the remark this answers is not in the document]
fallen-icarus: No, I specifically mean the `TxInfo` part of the `ScriptContext`.

CPS-????/README.md
fallen-icarus: Your reason doesn't not account for everything in the actual implementation. Using another spending script was discussed [here](https://github.com/cardano-foundation/CIPs/pull/418#issuecomment-1367309018). It is not a viable solution because it creates a bottleneck problem: you are only as concurrent as the number of utxos available at the spending script being delegated to. This bottleneck problem does not exist for staking scripts since no utxos are needed to force execution of the script.

fallen-icarus:
@michaelpj Thanks for the review. Can you please explain why you believe tx-level scripts mean the script can no longer know which script is being executed? Staking and minting scripts are both executed at the tx-level and yet they both know when they are being executed. This is the exact behave I believe spending scripts should have. Knowing which script is being executed is also how scripts know "themselves" which allows finding only the utxos locked by it (or staked by it). This "proximity knowledge" is very important for only using the right utxos in the computation. If the tx-level script cannot use this information anymore, it is basically pointless.

From my perspective, the fact that staking scripts can be used in the exact way I would like spending scripts to be used (the alternative technique in the CPS), strongly suggests that spending scripts can be used the same way while still using redeemers and script purpose.

CPS-????/README.md
michaelpj: The point is not that it's a workaround, my point is that your explanation for why it works is wrong.

michaelpj:
> Staking and minting scripts are both executed at the tx-level

Staking and minting scripts are executed once per "use site", it's just that there happens to only ever be one use site. My point above is that the staking and minting scripts get evaluated once for precisely the same reason that a spending script would get evaluated once if you had all your other scripts delegate to it. There isn't anything else special about them.
-----

I'm sorry, but I think the whole distinction of "tx-level" versus otherwise is just wrong and confusing.

Let's take a step back. At the end of the day the ledger runs script evaluations which look like `data ScriptEvaluation = ScriptEvaluation { script :: Script, args :: [Data]}`. We can skip some of those iff some of the `ScriptEvaluation`s are strictly the same. So the issue is: what parts of the `ScriptEvaluation` are preventing us from having them be the same.

Obviously the script itself has to be the same, so the answer is that the arguments must be different. So we get differences from:
1. Scripts being provided redeemers as arguments when the redeemers are different
2. Scripts being provided datums as arguments when the datums are different
3. Scripts being provided the script context as arguments when the script context is different. In practice the only part of the script context that varies is the `ScriptPurpose`, and within that the script hash is fine, but the full output reference is problematic.

So we have three sources of difference. 1&2 can be avoided by giving the scripts identical datums and redeemers, or if we made datums and redeemers optional, by omitting them.

My understanding of your [proposal](https://github.com/cardano-foundation/CIPs/pull/440) is to add two mechanisms to make script evaluations more similar:
1. Do _not_ provide the datum _even if_ it is present in the transaction.
2. Do _not_ provide the output reference in the `ScriptPurpose`.

(You don't provide a way to deal with the redeemers being different, so I assume you are just going to rely on them being the same, but in principle they're just as problematic as the datums and they should also be handled like 1)

And you propose to control these via a single flag.

I think this presentation is much clearer than the "tx-level" presentation. There is just a question of how to make more script evaluations look the same, and we can enumerate the cases that make them different.

fallen-icarus:
Okay thank you for clarifying that. I see now that "tx-level" is not accurate for staking and minting scripts. I will need to think about some of the things you said.

fallen-icarus:
> My understanding of your https://github.com/cardano-foundation/CIPs/pull/440 is to add two mechanisms to make script evaluations more similar:
> 1. Do not provide the datum even if it is present in the transaction.
> 2. Do not provide the output reference in the ScriptPurpose.

The second mechanism is not what I was suggesting. I suggested that the `ScriptPurpose` type take a sum type for the `Spending` constructor. However, if the goal is just to make `ScriptEvaluation` look different, then I now think this would be better:
``` Haskell
data ScriptPurpose
    = Minting CurrencySymbol
    | Spending ValidatorHash   -- ^ This is different.
    | Rewarding StakingCredential
    | Certifying DCert
```
I think this would address how to make the `ScriptEvalutation` look the same. Now, unless the utxos are using different datums and/or different redeemers, they would look like the same `ScriptEvaluation` and would only be executed once. Yet it would still be possible to execute the scripts individually by using different datums and/or different redeemers despite using the same script.

This should work with either `Maybe Datum` or `ScriptArgs`.

fallen-icarus:
@michaelpj What do you think about changing `ScriptPurpose` to this:
``` Haskell
-- | Script runtime representation of a Digest SHA256.
newtype ScriptArgsHash = ScriptArgsHash BuiltinByteString

data ScriptExecRef = ScriptExecRef
  { scriptExecHash :: ValidatorHash    -- ^ The hash of the script currently being executed.
  , scriptExecArgsHash :: ScriptArgsHash  -- ^ The hash of the ScriptArgs passed to this script.
  }

data ScriptPurpose
    = Minting CurrencySymbol
    | Spending ScriptExecRef
    | Rewarding StakingCredential
    | Certifying DCert
```
---
1. The `ScriptPurpose` is still only used to *identify* the execution.
2. This makes `ScriptEvaluation` differences only depend on `ScriptArgs`.
3. The cost of the duplication of `ScriptArgs` is minimized by hashing it.
4. `ScriptPurpose` can still be used in the `txInfoRedeemer` map.
5. More constructors can be added to `ScriptArgs` without having to change this again.

michaelpj:
> I think this would address how to make the ScriptEvalutation look the same. Now, unless the utxos are using different datums and/or different redeemers, they would look like the same ScriptEvaluation and would only be executed once. 

Yes, but it means the ScriptPurpose doesn't identify the script being evaluated any more, which users might want. Even if you have a unique datum and redeemer, it means that if you want to identify the value in "your" input, you would have to loop over all the inputs testing their datums and redeemers for identity with yours until you found it! Terrible!

> What do you think about changing ScriptPurpose to this <hashing>

Again, it doesn't help scripts identify their output any more. And you're wrong about it being usable in the redeemers map: it creates a circular data dependency. You want to know the redeemer to use for an input -> you need to create the `ScriptPurpose` to look it up -> that needs the redeemer.

fallen-icarus:
@michaelpj AFAICT the idea of making the `ScriptPurpose` optional ([here](https://github.com/cardano-foundation/CIPs/pull/440#discussion_r1097259633)) suffers from a similar limitation. The `ScriptPurpose` is currently the way that a plutus script can know its own `ValidatorHash` at run-time. This is important for knowing which inputs/outputs are its own. If the `ScriptPurpose` is not included, how would the plutus script be able to do this? Any solution to this problem must still allow a plutus script to know its own `ValidatorHash` at run-time.

fallen-icarus:
Instead of trying to remove duplicate `ScriptEvaluation`s after the tx is created, what about making it possible for users to create txs without any duplicate `ScriptEvaluation`s in the first place. Currently, a tx input is only a single utxo. But what if we changed this so that it could be either a sinlge utxo or a group of utxos like this:
``` Haskell
data TxInput 
  = SingleUTxO TxInInfo
  | GroupedUTxOs [TxInInfo]

data TxInfo
  { txInInfo :: [TxInput]
  ...
  }
```
The requirement for this is that all utxos in a group must use the same spending script to unlock them. Then the `ScriptPurpose` could become:
``` Haskell
data SpendingRef
  = SingleUTxORef TxOutRef
  | GroupedUTxORef [TxOutRef]

data ScriptPurpose
    = Minting CurrencySymbol
    | Spending SpendingRef
    | Rewarding StakingCredential
    | Certifying DCert
```
A group probably wouldn't take a datum (or it can choose which utxo in the group to use for the datum) and it would only use one redeemer (or none) for the entire group.

@michaelpj Am I misunderstanding something else that would make this approach infeasible?

michaelpj:
That might work, but it's a much more heavyweight set of changes for a small benefit.

fallen-icarus:
> it's a much more heavyweight set of changes for a small benefit.

I'm not sure if you mean "small" relative to the overall benefit of fixing this problem or "small" relative to the deduplication method.

If the former, I believe this problem impacts more than just the atomic swap scenario explained in the *Problem* section. I will focus on explaining the use cases more thoroughly when I rewrite the CPS. Also, the grouped input could easily allow ignoring datums which could be very useful, two birds one stone. 

If the latter, I'm not convinced making `ScriptPurpose` optional is that small of a change either since some other means would need to be introduced so that plutus scripts can know their own hashes at run-time. Without it, the deduplication approach wouldn't even solve the described problem since it would be unusable for this context.

imikushin:
@fallen-icarus:
> A group probably wouldn't take a datum (or it can choose which utxo in the group to use for the datum)

Let's suppose, in the atomic swap scenario, all participants put their assets into their "own" personal script addresses, created from the same script (with their own staking key part). Other than using the output's datum, how would the script know who owns each output?

@michaelpj:
> a much more heavyweight set of changes for a small benefit

So, two transactions instead of one is fine?

fallen-icarus:
@imikushin 
>  how would the script know who owns each output?

In the context of the atomic swaps, it is not necessary to use datums for this.
 >  all participants put their assets into their "own" personal script addresses, created from the same script (with their own staking key part).

This is actually how v2 of my DEX PoC works. The plutus script is able to get the address credentials (spending + staking) directly from the utxo being validated. This address credential is unique for every actual address. My DEX uses this address credential to check whether the swap conditions are met for that particular address. In essence, it traverses the list of inputs and list of outputs, and only cares about the utxos with an address that matches the address credential of the utxo being validated.

If grouped inputs were possible, I would alter the plutus script to enforce that all utxos in a group are from the same address (in addition to other checks), not just using the same plutus script. This would mean you would still need to execute the script once per address but dramatically simplifies the logic since each execution only needs to concern itself with one address.

fallen-icarus:
@michaelpj 
> all participants put their assets into their "own" personal script addresses, created from the same script (with their own staking key part)

I was going to explain this when I rewrite the CPS but I believe it is worth mentioning here. I am going to refer to all Dapps designed this way as *Distributed Dapps*. The *distributed* part comes from all users having their own address while using the Dapp. I believe all distributed dapps are impacted by this problem. 

For concentrated Dapps, care could be taken to maximize the value unlocked with each execution of the script. Simply consolidating all of the Dapp's value into just a few utxos would suffice. This approach is not possible for distributed dapps since at the very least, there will be one utxo per user. Also, unless that user is a whale, the amount of value in each utxo will be relatively small.

AFAICT my DEX PoC is the very first distributed dapp. But while this may be the case, I believe distributed dapps should be given higher priority than concentrated dapps. The reason for this belief is that, in the absence of atomic delegation, **distributed dapps are the only dapp design that can allow users to maintain full delegation control of their assets while using them.** This is existentially important for proof of stake blockchains. To my knowledge, my DEX PoC is the very first Defi application that has full delegation control for its users.

Side note: Finding all of the addresses for a distributed Dapp is not a problem. I solved this issue using what I call beacon tokens. Here is a link to my DEX PoC for convenience. [HERE](https://github.com/fallen-icarus/cardano-swaps)

imikushin:
@fallen-icarus 

> > how would the script know who owns each output?

> The plutus script is able to get the address credential (spending + staking) directly from the utxo being validated.

I'm _really curious_ as to where exactly (which part of [utxo](https://github.com/input-output-hk/cardano-ledger/blob/c7c63dabdb215ebdaed8b63274965966f2bf408f/eras/babbage/test-suite/cddl-files/babbage.cddl#L75)) this information could be encoded. I thought, the datum was _the only_ possible place. 

Note: I'm talking about the case where exactly **one** (reference) script is used to form the script addresses. So that (if tx-level validation script execution existed) it'd be possible to call the script exactly one time per transaction (if the transaction only involves outputs at addresses created from that one single script).

Even if that's possible, let's suppose, _your app_ doesn't need a datum for that. But in other apps, utxos may still need their own datums to store additional information per output, meaningful to the application (e.g. what the utxo's owner wants in return).

fallen-icarus:
@imikushin 
> I'm really curious as to where exactly (which part of [utxo](https://github.com/input-output-hk/cardano-ledger/blob/c7c63dabdb215ebdaed8b63274965966f2bf408f/eras/babbage/test-suite/cddl-files/babbage.cddl#L75)) this information could be encoded.

It is in the address variable. Check out this [part](https://github.com/input-output-hk/cardano-ledger/blob/c7c63dabdb215ebdaed8b63274965966f2bf408f/eras/babbage/test-suite/cddl-files/babbage.cddl#L142-L150). Here is what it looks like in plutus (slightly simplified):
``` Haskell
data Credential
  = PubKeyCredential PubKeyHash -- ^ The transaction that spends this output must be signed by the private key
  | ScriptCredential ValidatorHash -- ^ The transaction that spends this output must include the validator script and be accepted by the validator.

-- | Address with two kinds of credentials, normal and staking.
data Address = Address{ 
  addressCredential :: Credential, 
  addressStakingCredential :: Maybe Credential 
  }

-- | A transaction output (utxo), consisting of a target address, a value,
-- optionally a datum/datum hash, and optionally a reference script.
data TxOut = TxOut {
    txOutAddress :: Address,
    txOutValue :: Value,
    txOutDatum :: OutputDatum,
    txOutReferenceScript :: Maybe ScriptHash
    }
```
So when a dapp gives each user his/her own address, the address variable itself can be used as a user ID. 

> > A group probably wouldn't take a datum (or it can choose which utxo in the group to use for the datum) and it would only use one redeemer (or none) for the entire group.

I think you misunderstood what I meant here. I am talking about the plutus script here, not utxos. Currently, a plutus script takes three arguments: a datum, a redeemer, and the script context. The datum comes directly from the utxo being validated. But what if a group of utxos is being validated and each utxo has its own datum? Which datum should be passed to the script here? As of right now, only one datum can be passed to the script. The ScriptArgs proposal could potentially enable passing all of the datums in the group to the plutus script as a list.

It is important to recognize that the script context argument has both the datum and redeemer in it. [Here](https://github.com/input-output-hk/cardano-ledger/blob/c7c63dabdb215ebdaed8b63274965966f2bf408f/eras/babbage/test-suite/cddl-files/babbage.cddl#L49-L67) is basically what it is in the cddl file. The ledger-script interface extracts the datum and redeemer from the script context and passes them separately for convenience. If the plutus script is not passed any datum directly, it can still get all of the necessary datums from the script context argument by looking at the datums attached to the important utxos. My DEX PoC largely ignores the datum passed to the plutus script and instead gets pretty much all of its information from the script context argument.

My DEX PoC would still use datums. But my DEX PoC would still just get them directly from the script context argument. It does not need any datums passed as separate arguments to the plutus script for execution. I probably wouldn't even use the ScriptArgs datum list since assuming all of Bob's utxos are in the group being validated may not be a safe assumption to make. The grouping is just meant to group all of those utxos into one execution.

imikushin:
> I am talking about the plutus script here, not utxos. Currently, a plutus script takes three arguments: a datum, a redeemer, and the script context. The datum comes directly from the utxo being validated. But what if a group of utxos is being validated and each utxo has its own datum? Which datum should be passed to the script here?

@fallen-icarus I think I'm beginning to understand now. I like the idea of grouping UTXOs being spent in a `SpendingRef` (as an argument of the `Spending` ScriptPurpose) you suggested above!

Let's suppose we've unified script signature across all use cases: spending, minting, rewarding, etc. What's still unclear to me is how a script would determine what it's called for (even if we have an explicit ScriptPurpose to Redeemer map in [`TxInfo`](https://github.com/input-output-hk/plutus/blob/master/plutus-ledger-api/src/PlutusLedgerApi/V2/Contexts.hs#L71-L87)).

Would it make sense to _explicitly_ pass the ScriptPurpose to the script? (instead of making it figure it out from the context). This way, a single script could be used for several different purposes in the same transaction without any problem, e.g. to validate spending of a bunch of UTXOs (either as a group or individually - as encoded in the `Spending` script purpose) _and_ to mint/burn some tokens. Each such usage would translate in a call with its distinct ScriptPurpose value passed to the script. 

It's actually pretty weird that ScriptPurpose isn't being passed to the script directly.

imikushin:
> I didn't think mentioning the ScriptPurpose part of the script context argument was relevant before so I left it out. My apologies for the confusion.

My apologies for not doing the RTFM thing hard enough. Also, Cardano documentation is quite... decentralized. (Decentralization is a good thing, right?).

imikushin:
@fallen-icarus 

I really like the change to ScriptPurpose, suggested in https://github.com/cardano-foundation/CIPs/pull/418#issuecomment-1427046710

I'd perhaps want to use `TxInInfo` instead of `TxOutRef`:

```hs
data TxInInfo = TxInInfo
    { txInInfoOutRef   :: TxOutRef
    , txInInfoResolved :: TxOut
    }
```

```hs
data SpendingRef
  = SingleUTxORef TxInInfo
  | GroupedUTxORef [TxInInfo]

data ScriptPurpose
    = Minting CurrencySymbol
    | Spending SpendingRef
    | Rewarding StakingCredential
    | Certifying DCert
```

This would be great, because then we wouldn't need to spend extra cycles to get the `TxOut`. This could be further improved by adding `txOutDatumResolved` to `TxOut`:

```hs
data TxOut = TxOut {
    txOutAddress       :: Address,
    txOutValue         :: Value,
    txOutDatumHash     :: Maybe DatumHash
    txOutDatumResolved :: Maybe Datum
    }
```

With these simple changes, we'd be able to eliminate the datum argument (being passed to validation scripts) as it'd be trivial to access it via ScriptContext.

This feels (to me personally) as a fairly incremental and simple change. Also, what do you think, @michaelpj ?

fallen-icarus:
@imikushin Something like this was suggested in another PR [discussion](https://github.com/cardano-foundation/CIPs/pull/321#issuecomment-1250053504).

michaelpj:
I'm sorry, I'm a bit lost. Are you proposing changes to the `ScriptPurpose` independently of the grouped inputs proposal? The grouped inputs proposal is definitely _not_ simple, and needs a lot more details worked out (and TBH I doubt would be accepted without a lot more motivation).

imikushin:
@michaelpj 

> I'm sorry, I'm a bit lost. Are you proposing changes to the `ScriptPurpose` independently of the grouped inputs proposal? 

In my comment https://github.com/cardano-foundation/CIPs/pull/418#issuecomment-1440744894 I was suggesting to @fallen-icarus that `TxInInfo` would be nice to use instead of `TxOutRef` in his grouped inputs proposal. Sorry if it's confusing.

CPS-????/README.md
```diff
@@ -0,0 +1,81 @@
...
+
+### Simple Scripts
+I tried to test Cardano simple scripts to determine if this problem also impacts them. However, given that simple scripts are so small, it was hard to tell from just looking at the transaction fee. Since the interface for using Cardano simple scripts is the same as for using plutus scripts, I have to assume that this problem also impacts simple scripts. 
+
+If simple scripts are indeed impacted by this problem, the above technique would not work for them. The technique only works for plutus scripts.
+
+## Use Cases
+Any time a spending script's validation depends on the entire transaction and not on any particular UTxO.
+
+1. P2P Atomic Swaps - within one transaction, the value entering the swap address must be some proportion of the value leaving the swap address
```
rphair, line 66: Would it be fair to ask for a more detailed explanation of this use case?  This may be naive (and I'm not technically capable of joining the technical debate of the proposal itself) but if a treatment of this issue is helpful in implementing atomic swaps on Cardano, then I believe this use case should be covered in detail in the document... to progress this CPS and any CIP that facilitates that treatment.

I'm trying to get more reviewers based on interest in the Atomic Swap idea as you've described it on GitHub.  If in agreement with what I've said above, I also think this repository should be linked from the CPS itself.  Here is what I posted about [fallen-icarus/cardano-swaps](https://github.com/fallen-icarus/cardano-swaps) (please correct me if I've described the issue inaccurately or naïvely) in the Cardano MBO on Matrix

> I've been watching this concept develop from https://github.com/cardano-foundation/CIPs/pull/418 and I don't think third party reviewers have been giving it enough credit, or even enough attention.  I don't know every technical underpinning of every current & proposed DEX on Cardano but in the meantime this idea seems unique to me.  I do wish I could do more to promote this idea since my own technical shortcomings make it impossible for me to convincingly push ahead any proposed standards that would be required to make this happen.  I can only hope key people at IOG, CF, and Cardano dev agencies can take a serious look at how significant it would be for the industry & Cardano's own marketplace to support a true non-custodial DEX. 😯

https://matrix.to/#/!QvtUklhUfmEuCBIexz:matrix.org/$lU0OwYOAQXpNYVFJl1I33YPhhQ_wH14Jzu5etQP9Drk?via=matrix.adamantium.online&via=matrix.org&via=forum.balanceanalytics.io

rphair (commented):
> @fallen-icarus ([on Dec 31, 2022](https://github.com/cardano-foundation/CIPs/pull/418#issuecomment-1368059524)): I have officially open sourced my proof-of-concept which can be found [here](https://github.com/fallen-icarus/cardano-swaps). Given your stance, I would like to know your thoughts on it.

imikushin:
Just to add a little more color to this discussion: transaction-level scripts would _trivially_ address problems like [double satisfaction](https://plutus.readthedocs.io/en/latest/reference/writing-scripts/common-weaknesses/double-satisfaction.html) — just by being able to look at all sides of the trade at once.

I do realize that, of course, a per-output script execution _also_ has access to all spent outputs, but: if, in general, we may have to look at all participating outputs anyway, why do per-output calls at all?

imikushin:
I would perhaps even go as far as to suggest that _any_ script doesn't _need_ to be run more than once per transaction. 

This means, that even if a script could be used for multiple purposes within a single transaction (e.g. mint and spend), it should only be called once.

This would improve both security and scalability.

L-as:
> Just to add a little more color to this discussion: transaction-level scripts would _trivially_ address problems like [double satisfaction](https://plutus.readthedocs.io/en/latest/reference/writing-scripts/common-weaknesses/double-satisfaction.html) — just by being able to look at all sides of the trade at once.

Trivially not true given that two different scripts can also cause the problem.

fallen-icarus:
> two different scripts can also cause the problem.

I don't see how this invalidates what he said. Since the script can see all the other inputs, you can either 1) disallow transactions that have inputs from other scripts or 2) keep track of the value from each input and track where the value goes.

imikushin:
> Trivially not true given that two different scripts can also cause the problem.

@L-as Well, now we have this problem even for one script. What I'm suggesting would eliminate it (for outputs spent from the same script), wouldn't it? 

It would also make multi-script transaction logic (if you're into that) _a lot_ simpler.

CPS-????/README.md
```diff
@@ -0,0 +1,81 @@
+---
+CPS: ?
```
KtorZ, line 2:
```suggestion
CPS: 4
```

SebastienGllmt:
Although this CIP is for a different use-case than CIP38 there is some overlap and I just added this to the CIP38 motivation section https://github.com/cardano-foundation/CIPs/pull/309

CPS-????/README.md
rphair: Committing so we can become more ambitious about cross-referencing this with other CIPs e.g. https://github.com/cardano-foundation/CIPs/pull/466

rphair (commented):
@KtorZ @fallen-icarus @zhekson1 I'm trying to follow up on increased [interest on the Cardano MBO on Matrix](https://matrix.to/#/!QvtUklhUfmEuCBIexz:matrix.org/$pc9_V9arFOVMdpwYseaLyXkr3p65jAq_KP7dSvFjHh0?via=matrix.adamantium.online&via=matrix.org&via=forum.balanceanalytics.io) in [Cardano Swaps](https://github.com/fallen-icarus/cardano-swaps) and whatever it would take to implement them.  I've tried to stay ahead of this idea and see it's fallen quiet for about a month along with some of the CIP documented components that would be required.

Can we summarise here, so the community can follow potential implementations and so we can keep moving this CPS forward:

1 - Is there a "best" place on the CIP repository for discussing Cardano Swaps?  Or given that it's given [a full heading](https://github.com/fallen-icarus/CIPs/blob/beacons/CIP-%3F%3F%3F%3F/README.md#cardano-swaps-a-l1-dex-with-full-delegation-control) in https://github.com/cardano-foundation/CIPs/pull/466, would that be a better choice?

 - On that note, if Cardano Swaps is a major use case for this CPS, shouldn't it have its own section rather than a single line in a bulleted list of [Use Cases](https://github.com/fallen-icarus/CIPs/tree/full-eUTxO-potential/CPS-%3F%3F%3F%3F#use-cases) it has now? 

2 - Whether we can designate such a place, or need to create a new one (like a GitHub Discussion, if that's not a step backward), can we then use that place to gather public discussions like the ones I started collecting in https://github.com/cardano-foundation/CIPs/pull/466#discussion_r1135428703 when they relate to community interest & potential implementations.. both for the Cardano Swaps and related use cases?

3 - The [SPO Digest today](https://mail.iogservices.io/your-monthly-recap-is-here) mentioned [Cardano Working Groups: how to get involved](https://www.essentialcardano.io/article/cardano-working-groups) which lists an IOG group "Developer experience" which would cover applications for DEX and this CPS's other Use Cases... but without any more specific (sub-?) groups for these areas.

 - @KtorZ  how would we go about suggesting that a working group be formed to address this similar set of use cases, primarily the arguably most urgent case of Cardano Swaps?
 - If IOG hasn't formed or can't form a group or sub-group for it, is this something the CF could do?

Whether there are Plutus changes that need to happen first, or this CPS needs to be changed in scope, or an additional CPS needs to be created for Cardano Swaps or other DeFi components, etc... then such a Working Group might finally consolidate all this potential into better documentation of who needs to do what & in what order.

fallen-icarus:
I do not use matrix so I cannot see the activity there. Here are my thoughts though:
1. zhekson1 and I are currently working on `cardano-loans`. As the name suggests, it is a lending and borrowing platform using a similar distributed dApp design as `cardano-swaps`. It has gone through several iterations but is in the final steps. I am cleaning up the code and zheckson1 is working on the readme. We are doing this in a private repo because we thought that would be cleaner since it has gone through so many iterations; we didn't want people thinking it was finalized before it actually was. That is why the last month "appears" quiet. `cardano-loans` is relevant to this CPS but I think it is better to explain how after you see how it works.
2. Both `cardano-swaps` and `cardano-loans` are written using IOG's PlutusTx. While great for testing and prototyping, it is not very efficient. @Opshin did some [benchmarks](https://github.com/OpShin/plutus-bench) with the various smart contract languages and found PlutusTx can be 100x less efficient that something like aiken. With that in mind, I think both `cardano-swaps` and `cardano-loans` should be rewritten using aiken to determine just how efficient this design currently is, despite the redundant executions. This is totally arbitrary, but IMO `cardano-swaps` probably only needs to compose 10-20 swaps in a single tx. There are diminishing returns for more than that. Given the aforementioned benchmarks, I think this is already achievable if it is written in aiken. I fully intend to rewrite the code using aiken once `cardano-loans` and one other PoC are finished. That being said, `cardano-swaps` was open sourced and explained so that someone else can take a stab at it. The same is true for `cardano-loans`. I don't think this CPS can be completed without first seeing how far another smart contract language can push the designs while still suffering from the redundant executions. Unfortunately, these PoCs are taking longer than I expected. I am already working on them full-time.
3. As for where to discuss these designs, I can look to add a discussion option to each PoC repo. I've never done something like that before but I'm willing to see how it goes. Perhaps the Issue section can be used like this, too? I am open to other ideas.

CPS-????/README.md
```diff
@@ -0,0 +1,81 @@
...
+  - fallen-icarus <modern.daidalos@gmail.com>
+Proposed Solutions: []
+Discussions:
+  - https://github.com/cardano-foundation/CIPs/issues/417
+Created: 2022-12-20
+---
+# CPS-XXXX: Spending Script Redundant Executions
+
+## Abstract
+Plutus spendings scripts are currently executed once for every UTxO being consumed from the script's address. However, there are use cases where the validity of the transaction depends on the transaction as a whole and not on any individual UTxO. The eUTxO model allows a spending script to see the entire transaction context which means it is already possible to create scripts that validate based on the transaction as a whole. As of now, spending scripts that do this are forced to be executed once per UTxO even if these extra executions are completely redundant. Not only does this waste computational resources, but it can also result in the transaction fee growing quickly based on the number of UTxOs being consumed from the relevant script address. There is a technique that can be used to minimize this limitation but it has its own drawbacks.
```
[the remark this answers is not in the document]
matiwinnetou: @fallen-icarus what technique can be used already to minimise this drawback?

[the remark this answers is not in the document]
fallen-icarus: @matiwinnetou  The [Current Alternative](https://github.com/fallen-icarus/CIPs/tree/full-eUTxO-potential/CPS-%3F%3F%3F%3F#the-current-alternative) lays it out. In a nut shell, since you *can* make a plutus staking script execute only once for a given transaction, you off-load the heavy computation to that script. Then the spending plutus script just checks if the staking script was executed in the tx.

The idea is similar to what is being discussed in #309.

fallen-icarus:
After writing both cardano-swaps and cardano-loans in aiken, I am no longer concerned about the redundant executions. cardano-swaps can handle 8+ swaps in a single transaction, and cardano-loans can accept 6 loans or make payments on 8+ loans in a single transaction. I think these benchmarks are more than sufficient for right now. To be clear, redundant executions are still present but aiken is efficient enough to make them negligible. Therefore, I do not see a need to pursue this CPS.

rphair:
That's very good to hear @fallen-icarus and so closing this would be consistent with others' suggestions that more general methods would be applicable, and thanks for the update.

Is there anything that would be helpful in standards, documentation, developer support, or visibility to Cardano official agencies that would help with [`cardano-swaps`](https://github.com/fallen-icarus/cardano-swaps) as I was hoping to find in https://github.com/cardano-foundation/CIPs/pull/418#pullrequestreview-1398256071?

If so, l'd be happy if you would open a [CIPs issue](https://github.com/cardano-foundation/CIPs/issues) about it to help provide whatever support you might need for the Cardano Swaps, Loans, and other use cases you've documented here.

matiwinnetou:
@fallen-icarus I would disagree with this actually. I have been playing with Hydra and L2 and some more complex use cases we totally blew up Aiken and Plutus.

This still makes sense and it all depends on complexity of what one is doing.

fallen-icarus:
> Is there anything that would be helpful in standards, documentation, developer support, or visibility to Cardano official agencies that would help with [cardano-swaps](https://github.com/fallen-icarus/cardano-swaps) as I was hoping to find in https://github.com/cardano-foundation/CIPs/pull/418#pullrequestreview-1398256071?

@rphair I would like to finish the feature set of these protocols before personally revisiting CIPs. The ideascale proposal pretty much covers what I would like to do. I never thought these POCs would take so long... There are a few issues where CIPs could help but I think the arguments for them would be stronger once there are fully featured protocols to point to. The issues mostly revolve around maintenance of these protocols. Will the official location for the code be my repo, the CIPs repo, or an entirely new repo like a "Cardano Protocols" repo? Should an entity like the cardano foundation hold the license to them? What will be the process for making changes to the protocols? I think these questions relate to the ones you asked in the linked comment. Perhaps the community can start thinking about these questions while the protocols are finalized.

fallen-icarus:
@matiwinnetou Considering I have not personally dealt with the issues you mention, I do not think I am the right person to head this CPS if it should still be pursued. Should this one be re-opened or should a new PR be made where someone else is the author?

rphair:
> @fallen-icarus:
> Will the official location for the code be my repo, the CIPs repo, or an entirely new repo like a "Cardano Protocols" repo?
So far what we've determined is that the code should not be in the CIPs repo for a few reasons already highlighted
- code changes (even for instructive examples) require PRs that have to funnel through editors who might not be the best to approve or improve them.
- it's easier to subscribe, star, file issues, etc. for a project in its own repository that blur this development with the CIP repo divited among a lot of different documentation effors.
- the CIP process is focused on documentation rather than development & a lot of other subjective reasons as well as that one.

> Should an entity like the cardano foundation hold the license to them?

I can think of no good in particular that would be accomplished by that.  Anything the community does to promote Cardano P2P markets can rely entirely upon open source licenses published by developers themselves... and anything the CF does to promote these would not depend upon its holding license over either the code or its documentation in any CIPs.

> @matiwinnetou Considering I have not personally dealt with the issues you mention, I do not think I am the right person to head this CPS if it should still be pursued. Should this one be re-opened or should a new PR be made where someone else is the author?

What do you say Mateusz?  Upon your request we can open this up again and declare you the advocate of this CPS, with you eventfully adding yourself a CPS author whenever that might be appropriate. 🤔

matiwinnetou:
@rphair - I am happy to take over this CIP as an author. I personally don't think having Aiken (which is very performant indeed) is an excuse not to purse this work in future versions of Plutus.

rphair:
@matiwinnetou coincidentally it's been exactly a year since you offered to take over authorship / advocacy for this CPS.  Due to that time interval I'm marking this `Abandoned` (generally we would look for updates every 3 months unless dependencies are indicated) but we can change to `Waiting for Author` or a more active status if you just give us an update of what you might still intend to do & what we would be "waiting" for.

Failing that, we might close this PR if nothing further happens with it ("Abandoned" mainly refers to the CIP submission rather than the idea).  @fallen-icarus @MicroProofs @colll78 you might also give us your perspective about whether or not the recent update (rolling out with Chang) update to CIP-0069 addresses this issue (based on @fallen-icarus https://github.com/cardano-foundation/CIPs/pull/418#issuecomment-1384251009).

colll78:
> @matiwinnetou coincidentally it's been exactly a year since you offered to take over authorship / advocacy for this CPS. Due to that time interval I'm marking this `Abandoned` (generally we would look for updates every 3 months unless dependencies are indicated) but we can change to `Waiting for Author` or a more active status if you just give us an update of what you might still intend to do & what we would be "waiting" for.
> 
> Failing that, we might close this PR if nothing further happens with it ("Abandoned" mainly refers to the CIP submission rather than the idea). @fallen-icarus @MicroProofs @colll78 you might also give us your perspective about whether or not the recent update (rolling out with Chang) update to CIP-0069 addresses this issue (based on @fallen-icarus [#418 (comment)](https://github.com/cardano-foundation/CIPs/pull/418#issuecomment-1384251009)). 

No Chang does not, there is still no way to guard the spending of a UTxO without a spending script that is executed once per UTxO. However, https://github.com/cardano-foundation/CIPs/tree/master/CIP-0112 the observe CIP that was merged does address this.

fallen-icarus:
The observer script CIP only partially addresses this issue; the spending scripts are still redundantly checking if the observer script is executed. The [Spend Many CIP](https://github.com/cardano-foundation/CIPs/pull/858) is the only CIP I am aware of that would entirely eliminate the redundant executions.

rphair:
@fallen-icarus if you agree with this:
> However, [CIP-0112] the observe CIP that was merged does address this.

then it can (now that it's merged) be added under `Proposed Solutions:` in the header.

colll78:
> The observer script CIP only partially addresses this issue; the spending scripts are still redundantly checking if the observer script is executed. The [Spend Many CIP](https://github.com/cardano-foundation/CIPs/pull/858) is the only CIP I am aware of that would entirely eliminate the redundant executions.

Not exactly, because in the observer CIP extends native scripts such that they can check for the presence of an observer, and native scripts have no impact on Ex-Units. This means that no spending scripts are necessary, and no redundant executions occur.

rphair:
OK @fallen-icarus from the above conversation last month (not progressed since then) we have to assume this submission has been abandoned & that you or another advocate will either reopen this or submit a different iteration of the same idea if & when there is progress to be made.

PR #421, opened 2023-01-03
CIP-0075 | Fair Stakepool Rewards

TobiasFancee:
New pull request for CIP-0075 because I renamed branch. I also made changes to meet new CIP-0001 requirements.

PR #387, opened 2022-11-30
CIP-???? | Tune Parameters minPoolCost and K

ADARobinHood:
(see [rendered proposal in branch](https://github.com/ADARobinHood/CIPs/blob/master/CIP%20-%20%3F/README.md))

PR #422, opened 2023-01-03
CIP-0082? | Improved Rewards Scheme Parameters

TobiasFancee:
This is a new CIP for improving rewards scheme parameters. This proposal includes removing minPoolCost, adding minPoolRate, and increasing k. These are some of the most requested parameter changes discussed by the community. These changes will level the playing field for all stakepools while ensuring established community pools receive sufficient revenue via minPoolRate. This CIP does not deprecate CIP-0075 and is meant to be an alternative. While I still believe CIP-0075 to be useful, it simply makes too many changes to be implemented anytime soon (would require additional research). This CIP offers smaller changes that do not alter the incentives scheme. Rewards calculation for stakepools and pledge benefit are not changed (other than the addition of minPoolRate). In that way, I believe this CIP has a better chance of being implemented.

Forum Post:
https://forum.cardano.org/t/cip-improved-rewards-scheme-parameters/112409

---
([current proposal, rendered from branch](https://github.com/TobiasFancee/CIPs-1/blob/CIP-Improved-Rewards-Scheme-Parameters/CIP-0082/README.md))


CIP-Improved_Rewards_Scheme_Parameters/README.md
```diff
@@ -0,0 +1,238 @@
...
+    - Input Output Global
+    - Cardano Foundation
+    - Emurgo
+Discussions:
+    - https://forum.cardano.org/t/cip-improved-rewards-scheme-parameters/112409
+Created: 2022-01-03
+License: CC-BY-4.0
+---
+
+# Abstract
```
rphair, line 18:
```suggestion
# Improved Rewards Scheme Parameters
## Abstract
```
The H1 header level is generally used for a document title, while the top-level sections are H2's, and so on to H3 as those sections are broken up.  Therefore for proper outline you should add an H1 title here (duplicating the `Title:` header field if necessary).

CIP-Improved_Rewards_Scheme_Parameters/README.md
```diff
@@ -0,0 +1,238 @@
...
+
+The current parameter settings of Cardano's rewards sharing scheme leave much to be desired in terms of fairness and promoting decentralization. minPoolCost puts small stakepools at a significant disadvantage. Replacing minPoolCost with a minPoolRate will ensure a level playing field for stakepools while providing sufficient sybil attack resistance. Additionally, the current setting of k, the optimal number of stakepools, is too low to provide an adequate pledge benefit. Increasing k will make the pledge benefit more effective and get delegations moving in hopes of helping single pool operators gain delegations.
+
+The parameter changes in this proposal are an optimization of the current settings and are meant to improve the fairness and decentralization of Cardano. Furthermore, all changes suggested in this proposal have been specifically voiced by the Cardano community.
+
+# Motivation
+
+## Definitions
```
rphair, line 26:
```suggestion
### Definitions
```
... etc

CIP-Improved_Rewards_Scheme_Parameters/README.md
```diff
@@ -0,0 +1,238 @@
...
+
+1.	Propose changes voiced by the community.
+2.	Make Cardano staking fairer by eliminating aggressive anticompetitive features.
+3.	Increase the effect of the pledge benefit on staking rewards.
+4.	Get stale delegations moving and allow users to reconsider their delegation.
+5.	Maintain sufficient sybil attack protection.
+
+## Explanation
+
+### Stage 1: minPoolCost is decreased to 170 ADA
```
rphair, line 113:
```suggestion
#### Stage 1: minPoolCost is decreased to 170 ADA
```
... etc

rphair (commented):
thanks; I read this one on the Cardano Forum already & in my opinion it's explicit and well-written except for the rank of the section headers.

I already wrote something about whether multiple CIPs should be submitted by the same author to accomplish the same thing (https://github.com/cardano-foundation/CIPs/pull/387#pullrequestreview-1206808868).  I believe this is another one of those cases where the proposals should be merged, but also that this is a moot point without an engagement process from IOG for considering RSS change proposals.

These titles (including https://github.com/cardano-foundation/CIPs/pull/421) are also more compelling without the subjective language.  If there were a universal definition of "fair" or "improved" then the RSS system mightn't have been disputable in the first place.  I've tried to come up with acceptable synonyms but that's not easy either: "Decentralising" is hard to prove; perhaps "distributed" or "favouring distribution"?

TobiasFancee:
I've made the suggested changes to the headers. Let me know if they are good now.

As for IOG considering RSS changes, I think we need to write the CIPs regardless. The community and SPOs need to be able to point to a CIP when asked about what they want. 

As for subjective language, I agree that it is difficult to come up with better titles. An alternative might be "Updated Rewards Scheme Parameters" but that is quite generic. I actually think it's okay to have subjective titles in order to get community support. The community should understand it's a proposal and that they must form their own opinions. As for the title "Improved Rewards Scheme Parameters", I don't think it's that bad. This is an "Improvement Proposal" after all. If I did not consider it an improvement, I would not have posted it. Also, important to note CIP-0023 "Fair Min Fees" has a subjective title and it has been merged. Overall, I don't think the title is that bad, but I am open to suggestions for a new title.

Maybe a more objective title might be "Pledge Benefit Optimization" as this CIP makes pledge more important by removing minPoolCost and increasing k? But it doesn't change the pledge benefit, so I wouldn't want there to be confusion.

rphair:
> changes to the headers. Let me know if they are good now

Thanks: the header outline structure looks fine now 🤓

rphair:
Some notes for community consideration of this CIP have been made by the author here; most of this will already be known to RSS reviewers already but it is a pretty good review of this proposal's Rationale: https://twitter.com/TobiasFancee/status/1613949612766760960

prometheus-pool:
I believe minPoolRate would be a huge improvement over minPoolCost for many of the reasons outlined in this CIP. At the same time I am not convinced a minPoolRate is necessary, and I believe it suffers from one of the major problems that minPoolCost suffers from, that it hides the most rewarding pools from delegators. Specifically, there will be pools that don't need to set any fixed cost because the minPoolRate covers their operating expenses and then some, and some of these pools would charge less than minPoolRate were they allowed to do so because they would still be profitable. Yet to delegators these pools will look the same as many other pools that are not being run as efficiently, meaning minPoolRate will make it difficult for delegators to select the best run pools. Such pools may rightfully feel frustrated that they are unable to leverage their competitive advantage to attract delegation, and the network may end up again paying people who are the loudest rather than the best at operating the network infrastructure.

An additional comment, I don't think it is necessarily true that "the pledge benefit on its own has no means to combat zero fee sybil pools" as you say in the CIP, it is certainly possible for a pool to leverage its pledge benefit to offer greater rewards than a 0 fee pool while still charging fees to cover its operating costs, how common this is comes down to the current parameter settings as well as the distribution of ada across the ecosystem. It seems like it would be valuable to test and adjust the protocol's researched parameters (after removing minPoolCost) before attempting to pre-empt potential problems by adding a new parameter such as minPoolRate to the protocol.

rphair:
Notes from CIP meeting today (cc @KtorZ)

https://github.com/cardano-foundation/CIPs/pull/421#discussion_r1071872077 is also applicable here because the companies listed as Implementors ("Implementors include IOG, CF, and Emurgo") haven't agreed to implement or engage with this.

Also this would have to be merged as `Inactive` rather than `Proposed` because there's no framework for IOG to discuss or implement these proposals.  It's been resolved at the meeting that we will have to do this for all RSS proposals, even ones that have already been merged.

CIP-Improved_Rewards_Scheme_Parameters/README.md
```diff
@@ -0,0 +1,239 @@
+---
+CIP: ????
```
KtorZ, line 2:
```suggestion
CIP: 82
```

TobiasFancee:
So I guess these CIPs really are useless. At least they are educational. Let me know if there is anything else I need to do to get it merged.

rphair:
@TobiasFancee lots of people value your contributions and in my opinion all the RSS proposals support governance goals which have long term use beyond pure education.

CIP-Improved_Rewards_Scheme_Parameters/README.md
```diff
@@ -0,0 +1,239 @@
+---
+CIP: 82
+Title: Improved Rewards Scheme Parameters
+Status: Proposed
```
rphair, line 4:
```suggestion
Status: Inactive (no IOG support)
```
... or some such "justification" which indicates the problem.  @KtorZ we may need your help choosing a justification tag that will be (grimly) applied to all RSS proposals.  Mentioning the specific company name seems undiplomatic but it is not inaccurate or open to interpretation either so I believe it should be stated baldly as such.

CIP-Improved_Rewards_Scheme_Parameters/README.md
```diff
@@ -0,0 +1,239 @@
...
+3. Genesis delegates IOG, CF, and Emurgo sign the transaction.
+4. Transaction is submitted.
+5. Hardfork is accepted by majority of the network.
+6. Hardfork and changes to ledger rules are confirmed.
+
+### Implementation Plan
+
+Implementors include IOG, CF, and Emurgo. All three entities must approve the changes to the network. Each stage will be an individual update. Stages 1, 3, and 4 will be parameter updates. Stage 2 will require a hardfork.
+
+Before implementation, IOG's engineering and research teams must review the feasibility and potential consequences of the proposal. IOG will create the implementation for each update and decide on the time interval between updates.
```
rphair, line 228:
```suggestion
Each stage will be an individual update. Stages 1, 3, and 4 will be parameter updates. Stage 2 will require a hardfork.

Before implementation, engineering and research teams must review the feasibility and potential consequences of the proposal, create the implementation for each update, and decide on the time interval between updates.
```
Like we said in an earlier comment (maybe for your other proposal... having to duplicate everything makes me uncertain), only companies who've agreed to participate in the implementation should be mentioned explicitly here.

It makes the language a bit awkward and we'll see if @KtorZ has any feedback that might relax the requirement to avoid mentioning these companies explicitly.  My feeling is that the `Inactive` status should allow authors to say anything they want here.

rphair (commented):
It looks like this is therefore ready to merge but based on prior discussion (I can dig up some links if needed) it should have a state of `Inactive` because there is no way to activate the RSS proposals.

The same would be true for https://github.com/cardano-foundation/CIPs/pull/421 with the same modifications, and also fixing the header levels as for this one: https://github.com/cardano-foundation/CIPs/pull/421#pullrequestreview-1250388369

CIP-Improved_Rewards_Scheme_Parameters/README.md
KtorZ: I'd simply go for: "project not enlisted for CIP process", which goes back to CIP-0001 and the need for a project to explicitly states their engagement to the CIP process.

TobiasFancee:
@KtorZ @rphair I've removed any mention of entities with genesis keys and changed the status to "Project Area Not Enlisted for CIP Process" Let me know if this can be merged now or if further changes are necessary. Sorry I haven't been active in making these changes. I've had a busy month. My goal is to at least have this CIP ready for Voltaire.

CIP-Improved_Rewards_Scheme_Parameters/README.md
```diff
@@ -0,0 +1,237 @@
...
+    - Tobias Fancee <tobiasfancee@gmail.com>
+Implementors:
+    - TBD
+Discussions:
+    - https://forum.cardano.org/t/cip-improved-rewards-scheme-parameters/112409
+Created: 2022-01-03
+License: CC-BY-4.0
+---
+
+# Improved Rewards Scheme Parameters
```
rphair, line 16:
```suggestion
```
Clarification from `CIP-0001` in meantime since you submitted your proposal precludes the H1 title (note in [CIP-0001 > Structure section](https://github.com/cardano-foundation/CIPs/tree/master/CIP-0001#structure)) because this will generally be generated in software.

CIP-Improved_Rewards_Scheme_Parameters/README.md
```diff
@@ -0,0 +1,237 @@
+---
+CIP: 82
+Title: Improved Rewards Scheme Parameters
+Status: Project Area Not Enlisted for CIP Process
+Category: Reward-Sharing Scheme
+Authors:
+    - Tobias Fancee <tobiasfancee@gmail.com>
+Implementors:
+    - TBD
```
rphair, line 9:
```suggestion
Implementors: []
```
Unless I'm reading [CIP-0001 > Header Preamble](https://github.com/cardano-foundation/CIPs/tree/master/CIP-0001#header-preamble) wrong myself, this is the correct syntax for not having any implementors yet (cc @KtorZ - not sure since we don't have this syntax on any merged CIPs yet).

CIP-Improved_Rewards_Scheme_Parameters/README.md
```diff
@@ -0,0 +1,237 @@
...
+---
+
+# Improved Rewards Scheme Parameters
+## Abstract
+
+The current parameter settings of Cardano's rewards sharing scheme leave much to be desired in terms of fairness and promoting decentralization. minPoolCost puts small stakepools at a significant disadvantage. Replacing minPoolCost with a minPoolRate will ensure a level playing field for stakepools while providing sufficient sybil attack resistance. Additionally, the current setting of k, the optimal number of stakepools, is too low to provide an adequate pledge benefit. Increasing k will make the pledge benefit more effective and get delegations moving in hopes of helping single pool operators gain delegations.
+
+The parameter changes in this proposal are an optimization of the current settings and are meant to improve the fairness and decentralization of Cardano. Furthermore, all changes suggested in this proposal have been specifically voiced by the Cardano community.
+
+## Motivation
```
rphair, line 23:
```suggestion
## Motivation: why is this CIP necessary?
```
pedantic perhaps: but necessary according to current convention.

CIP-Improved_Rewards_Scheme_Parameters/README.md
```diff
@@ -0,0 +1,237 @@
...
+|-----------------------|---------------------|-------------------------|----------------|------------------------------|
+| stakePoolTargetNum    | N                   | N                       | 750            | See Rationale Section.       |
+
+### Stage 4: k increased to 1000
+
+| Name of the Parameter | New Parameter (Y/N) | Deleted Parameter (Y/N) | Proposed Value | Summary Rationale for Change |
+|-----------------------|---------------------|-------------------------|----------------|------------------------------|
+| stakePoolTargetNum    | N                   | N                       | 1000           | See Rationale Section.       |
+
+## Rationale
```
rphair, line 100:
```suggestion
## Rationale: how does this CIP achieve its goals?
```
as with _Motivation_ also indicated by current convention.

CIP-Improved_Rewards_Scheme_Parameters/README.md
```diff
@@ -0,0 +1,237 @@
+---
+CIP: 82
+Title: Improved Rewards Scheme Parameters
+Status: Project Area Not Enlisted for CIP Process
```
rphair, line 4:
```suggestion
Status: Inactive (project area not enlisted for CIP process)
```
This is a qualifying phrase for the status of `Inactive`, rather than the status itself... lowercase according to current example in [CIP-0001 > Status: Inactive](https://github.com/cardano-foundation/CIPs/tree/master/CIP-0001#status-inactive).

rphair (commented):
on the contrary @TobiasFancee I am happy you are following through with this.  I would prefer to see this on the books during this period as well, but there are a couple more format nitpicks.

Also you will have to rename the project folder from `CIP-Improved_Rewards_Scheme_Parameters` to `CIP-0082` before we merge it. 🤓

TobiasFancee:
@KtorZ @rphair I've made the requested changes. Let me know if there is anything else needed to merge. Thanks.

KtorZ (approved):

rphair (approved):
you're welcome @TobiasFancee - this looks ready to merge at or before the next CIP meeting.

PR #427, opened 2023-01-12
CIP-0005, CIP-0036 | Added clarity to voting key definitions

Ryun1:
## Motivation
- Concerns were raised around the definition of CIP-36's voting keys and how these should be represented in supporting tooling (good spot from @janmazak and @gitmachtl). 
- Avoid confusion with current and future protocol governance.

## Proposed Changes
CIP-05
- [825d54d](https://github.com/cardano-foundation/CIPs/pull/427/commits/825d54da6487fa84eb398604df335928881f4d19): Added `cvote_sk` and `cvote_vk` prefixes to be used when Bech32 encoding CIP-36's voting keys. As per [gitmachtl's suggestion](https://github.com/cardano-foundation/CIPs/pull/427#issuecomment-1383960821).

CIP-36 
- Added clarity for voting / vote key terminology. Linking to CIP-05 prefixes.
- Gave examples of potential naming supporting tooling could use.
- [4f6b879](https://github.com/cardano-foundation/CIPs/pull/427/commits/4f6b8794e51ce1f04dc808199b4223fcd7bc59ef): Added preference for implementations to use term "vote".
- [825d54d](https://github.com/cardano-foundation/CIPs/pull/427/commits/825d54da6487fa84eb398604df335928881f4d19): Added more example keyTypes with descriptions, as per [gitmachtl's suggestion](https://github.com/cardano-foundation/CIPs/pull/427#issuecomment-1383960821).
- [825d54d](https://github.com/cardano-foundation/CIPs/pull/427/commits/825d54da6487fa84eb398604df335928881f4d19): Added explanation for how other projects can add themselves to current keyType descriptions, as per [gitmachtl's suggestion](https://github.com/cardano-foundation/CIPs/pull/427#issuecomment-1383960821).

## Rationale
With these changes we are trying to avoid confusion of CIP-36's vote keys with current and future protocol governance keys and processes.

CIP-05
- The previously added `gov_sk` and `gov_vk` which were intended to relate to such keys, are likely to cause confusion, so I have added new prefixes.
- By leaving `gov_sk` and `gov_vk` - we intend to reserves these prefixes for potential use in the future protocol governance.
- [825d54d](https://github.com/cardano-foundation/CIPs/pull/427/commits/825d54da6487fa84eb398604df335928881f4d19): By leaving `vote_sk` and `vote_vk` - we intend to reserves these prefixes for potential use in the future protocol governance.

CIP-36
- By associating both the terms vote and voting, we attempt to avoid future confusion by not allowing other standards to utilize one term.
-  Adding explicit `keyType`'s was the level of clarity that was requested, I believe.
- [825d54d](https://github.com/cardano-foundation/CIPs/pull/427/commits/825d54da6487fa84eb398604df335928881f4d19): By encouraging the use of terms such as "CIP-36 vote keys" we leave the base "vote key" for use by future standards.
- [825d54d](https://github.com/cardano-foundation/CIPs/pull/427/commits/825d54da6487fa84eb398604df335928881f4d19):  Adding a fuller list of `keyType` with descriptions better satisfied the concerns raised in comments on this PR.
- [825d54d](https://github.com/cardano-foundation/CIPs/pull/427/commits/825d54da6487fa84eb398604df335928881f4d19): By allowing future projects to update `keyType` descriptions we cater for future use of these CIP-36 vote keys. This was suggested by gitmachtl [here](https://github.com/cardano-foundation/CIPs/pull/427#issuecomment-1383960821).

**Note**: Submitted as one PR so the related changes are kept together, but if its preferable to have 2 PR's please advise.

### DIFF
--- a/CIP-0036/README.md
+++ b/CIP-0036/README.md
@@ -55,11 +55,27 @@ Each delegation therefore contains:
   - a voting key: simply an ED25519 public key. This is the spending credential in the sidechain that will receive voting power from this delegation. For direct voting it's necessary to have the corresponding private key to cast votes in the sidechain. How this key is created is up to the wallet.
   - the weight that is associated with this key: this is a 4-byte unsigned integer (CBOR major type 0, The weight may range from 0 to 2^32-1) that represents the relative weight of this delegation over the total weight of all delegations in the same registration transaction.
 
-### Voting key derivation path
+### Voting key 
+
+The terms voting keys and vote keys should be used interchangeably to indicate the keys described in this specification. The term governance should not be associated with these keys.
+
+#### Derivation path
 
 To avoid linking voting keys directly with Cardano spending keys, the voting key derivation path must start with a specific segment:
 
 `m / 1694' / 1815' / account' / chain / address_index`
+
+We recommend that implementors only use `address_index`=0 to avoid the need for voting key discovery.
+
+#### Tooling
+
+Supporting tooling should clearly define and differentiate this as a unique key type. Using the `vote_sk` and `vote_vk` Bech32 prefixes when encoding, as described in [CIP-05](https://github.com/cardano-foundation/CIPs/tree/master/CIP-0005).
+
+Examples of acceptable `keyType` naming for supporting tools:
+- `VotingSigningKey_ed25519`
+- `VotingVerificationKey_ed25519`
+- `VoteSigningKey_ed25519`
+- `VoteVerificationKey_ed25519`
 
 ### Associating voting power with a voting key
 
--- a/CIP-0036/README.md
+++ b/CIP-0036/README.md
@@ -57,9 +57,9 @@ Each delegation therefore contains:
 
 ### Voting key 
 
-The terms voting keys and vote keys should be used interchangeably to indicate the keys described in this specification. But it should be made clear that implementations should favour the term "vote" and that the association of both terms aims to reduce the possibility of confusion.
+The terms "(CIP-36) voting keys" and "(CIP-36) vote keys" should be used interchangeably to indicate the keys described in this specification. But it should be made clear that implementations should favour the term "(CIP-36) vote" and that the association of both terms aims to reduce the possibility of confusion.
 
-The term governance should not be associated with these keys.
+The term governance should not be associated with these keys nor should these keys be associated with other vote or voting keys used in the ecosystem. When discussing these keys in a wider context they should be specified by such terminology as "CIP-36 vote keys" or "CIP-36 style vote keys".
 
 #### Derivation path
 
@@ -71,11 +71,24 @@ We recommend that implementors only use `address_index`=0 to avoid the need for 
 
 #### Tooling
 
-Supporting tooling should clearly define and differentiate this as a unique key type, describing such keys as "vote" keys. Using the `vote_sk` and `vote_vk` Bech32 prefixes when encoding, as described in [CIP-05](https://github.com/cardano-foundation/CIPs/tree/master/CIP-0005).
+Supporting tooling should clearly define and differentiate this as a unique key type, describing such keys as "CIP-36 vote keys". When utilizing Bech32 encoding the appropriate `cvote_sk` and `cvote_vk` prefixes should be used as described in [CIP-05](https://github.com/cardano-foundation/CIPs/tree/master/CIP-0005)
 
-Examples of acceptable `keyType` naming for supporting tools:
-- `VoteSigningKey_ed25519`
-- `VoteVerificationKey_ed25519`
+Examples of acceptable `keyType`s for supporting tools:
+
+| `keyType` | Description |
+| --------- | ----------- |
+| `CIP36VoteSigningKey_ed25519` | Catalyst Vote Signing Key |
+| `CIP36VoteExtendedSigningKey_ed25519` | Catalyst Vote Signing Key |
+| `CIP36VoteVerificationKey_ed25519` | Catalyst Vote Verification Key |
+| `CIP36VoteExtendedVerificationKey_ed25519` | Catalyst Vote Verification Key |
+
+For hardware implementations:
+| `keyType` | Description |
+| --------- | ----------- |
+| `CIP36VoteVerificationKey_ed25519` | Hardware Catalyst Vote Verification Key |
+| `CIP36VoteHWSigningFile_ed25519` | Hardware Catalyst Vote Signing File |
+
+The intention with this design is that if projects beyond Catalyst implement this specification they are able to add to themselves `keyType` **Description**s.
 
 ### Associating voting power with a voting key
 
--- a/CIP-0036/README.md
+++ b/CIP-0036/README.md
@@ -57,7 +57,9 @@ Each delegation therefore contains:
 
 ### Voting key 
 
-The terms voting keys and vote keys should be used interchangeably to indicate the keys described in this specification. The term governance should not be associated with these keys.
+The terms voting keys and vote keys should be used interchangeably to indicate the keys described in this specification. But it should be made clear that implementations should favour the term "vote" and that the association of both terms aims to reduce the possibility of confusion.
+
+The term governance should not be associated with these keys.
 
 #### Derivation path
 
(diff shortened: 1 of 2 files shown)


rphair (commented):
Looks like an essential improvement to me, as far as I can understand it.

cc @kevinhammond @Zeegomo @KtorZ @mark-stopka

CIP-0036/README.md
```diff
@@ -55,12 +55,28 @@ Each delegation therefore contains:
   - a voting key: simply an ED25519 public key. This is the spending credential in the sidechain that will receive voting power from this delegation. For direct voting it's necessary to have the corresponding private key to cast votes in the sidechain. How this key is created is up to the wallet.
   - the weight that is associated with this key: this is a 4-byte unsigned integer (CBOR major type 0, The weight may range from 0 to 2^32-1) that represents the relative weight of this delegation over the total weight of all delegations in the same registration transaction.
 
-### Voting key derivation path
+### Voting key 
+
+The terms voting keys and vote keys should be used interchangeably to indicate the keys described in this specification. The term governance should not be associated with these keys.
```
janmazak, line 60: Is the vote/voting dichotomy worth it? For talking about in natural language it's OK, but for code it's not great if you have two ways of naming the same thing. So in the section "Tooling" I'd propose to only have one thing, or one of them strongly preferred (I'd go for the shorter version if other reasons are absent).

Otherwise, anyone working with two tools that use different names, will have to choose one of the ways, and then his code will be inconsistent with the other tool (and everyone dealing with that code will have to wonder "why are we using vote while the tool uses voting ". And people will needlessly get stupid bugs when integrating tools/libraries and mistakenly use one name instead of the other.

[A mess of this kind was created by reward account / reward address / stake address / staking address for that particular type of Cardano address. Every time someone new comes to the ecosystem, he has to think about whether these different names refer to the same thing or not, and ask around for clarification.]

Ryun1: @janmazak 

Thanks for your suggestion - you raise a good point, I agree. I will change the wording to prefer the term "vote" for tooling.

gitmachtl:
As talked in the Discord-Chat, we should also add `VoteExtendedSigningKey_ed25519` as a possible key-type for the json key-format. Because its an extended key if derived from path `1694H/1815H/0H/0/0` BIP32 with BIP39 mnemonic, if not a simple ed25519 keypair is used. 

So currently we would have the following key types for the **json** format:
|type|description|
|-----|-----|
|`VoteVerificationKey_ed25519`|`Vote Verification Key`|
|`VoteSigningKey_ed25519`|`Vote Signing Key`|
|`VoteExtendedSigningKey_ed25519`|`Vote Signing Key`|

The **hardware** related key/files via cardano-hw-cli would than be:
|type|description|
|-----|-----|
|`VoteVerificationKey_ed25519`| `Hardware Vote Verification Key`|
|`VoteHWSigningFile_ed25519`|`Hardware Vote Signing File`|

disassembler:
I agree with the thought we want to have types/description to identify these key types so you can use them across multiple tools and don't shoot yourself in the foot, however; I think we should be very specific in the purpose for the keys rather than using a generic "Vote" context. Ideally, keys used by Catalyst would have a Catalyst Prefix and keys used for on-chain would have a separate prefix (e.g. DRep), and the the derivation paths would not be the same for security reasons. It's a bad idea in my opinion to use the same keys for multiple purposes. This is primarily why we created the type system in the first place to prevent people from accidentally using a payment key as a cold key for their stake pool. While it is possible at a cryptographic level, we wanted people to not accidentally do it.

gitmachtl:
Going with the suggestion of @disassembler, we could do the following than for catalyst voting specific?

CLI based **json** format:
|type|description|bech-prefix|
|-----|-----|-----|
|`CatalystVoteVerificationKey_ed25519`|`Catalyst Vote Verification Key`|cvote_vk|
|`CatalystVoteExtendedVerificationKey_ed25519`<sup>1</sup>|`Catalyst Vote Verification Key`|cvote_vk|
|`CatalystVoteSigningKey_ed25519`|`Catalyst Vote Signing Key`|cvote_sk|
|`CatalystVoteExtendedSigningKey_ed25519`|`Catalyst Vote Signing Key`|cvote_sk|

<sup>1</sup>: Not sure if we need an extended version here, tools can strip the last bytes of an extended one and the result would be a normal verification key.


The **hardware** related .vkey/.hwsfile via cardano-hw-cli:
|type|description|bech-prefix|
|-----|-----|-----|
|`CatalystVoteVerificationKey_ed25519`| `Hardware Catalyst Vote Verification Key`|cvote_vk|
|`CatalystVoteHWSigningFile_ed25519`|`Hardware Catalyst Vote Signing File`|-|

janmazak:
@gitmachtl @disassembler 
HW wallets can't distinguish keys. There's the same API and the same code to process general voting keys and catalyst keys, and the keys are given from the outside, a HW wallet has no idea if it is catalyst or not. So it cannot choose the appropriate prefix (vote_vk/cvote_vk). The prefix cannot be given together with the key, that would be just a security hazard (a HW wallet must only display info that it understands / can verify, everything else comes with a big warning and is displayed as-is in raw form, and under no circumstances _wrong_ info should be displayed, e.g. a key with a false prefix).

On the other hand, it's possible to make the distinction on the level of cardano-hw-cli, but the cli tool must know what is catalyst and what is not --- so far the key derivation path is the same AFAIK, so it would have to just be told when writing keys obtained from a HW wallet to a file.

gitmachtl:
we can for now change the `gov_vk` in the cardano-app into `cvote_vk` right? we have the voting_purpose indicator, that is 0 for catalyst. would it be possible to just do a simple
``` js
if ( vote_purpose == 0 ) { prefix = 'cvote_vk'; }
```
?

as we have only catalyst for now, i think we can code it for that atm right?

janmazak:
@gitmachtl No. The voting purpose comes in the last APDU, so it is not known at the moment delegations are processed. (This could be changed.)

Ryun1:
@disassembler @gitmachtl @janmazak

Thanks everyone for your insights on this, I really appreciate the input.

I see two options here:
1. Follow Catalyst naming.
2. Follow generic CIP-36 naming.

I believe intent of CIP-36's design was to create a generic governance standard which could be used for different projects. For this, a derivation path was chosen for keys so that these keys could be used across projects. So although Catalyst is currently the only user of `1694H/1815H/0H/0/0` this does not stop other projects from using `1694H/1815H/0H/0/0` in their use of CIP-36. Remember to do this other projects would have to register this intent in this CIP and their delegations/registrations transactions would be distinguished from Catalysts' by the `voting_purpose` field. I believe the idea of `voting_purpose` allows for the reuse of the same keys  across projects.

With this in mind and reading everyone's thoughts I see three options; 1. We name these keys after Catalyst, which does make sense as Catalyst is the only current implementer. This does not stop other projects from using these same keys in the future as per CIP-36's design. 2. We name these keys after the CIP itself, making these keys not tied to Catalyst which may make this standard more attractive to other projects at the risk of making things confusing.

gitmachtl:
@Ryun1 @disassembler @janmazak 

ok, so if we go with your type suggestion @Ryun1  but keep the purpose also in the description, we could do it like:

CLI based .vkey/.skey **json** format:
|type|description|bech-prefix|
|-----|-----|-----|
|`CIP36VoteVerificationKey_ed25519`|`Catalyst Vote Verification Key`|cvote_vk|
|`CIP36VoteExtendedVerificationKey_ed25519`|`Catalyst Vote Verification Key`|cvote_vk|
|`CIP36VoteSigningKey_ed25519`|`Catalyst Vote Signing Key`|cvote_sk|
|`CIP36VoteExtendedSigningKey_ed25519`|`Catalyst Vote Signing Key`|cvote_sk|

The **hardware** related .vkey/.hwsfile via cardano-hw-cli in **json** format:
|type|description|bech-prefix|
|-----|-----|-----|
|`CIP36VoteVerificationKey_ed25519`| `Hardware Catalyst Vote Verification Key`|cvote_vk|
|`CIP36VoteHWSigningFile_ed25519`|`Hardware Catalyst Vote Signing File`|-|

So thats is clear by the type how the keys are derived (via CIP36 specification), but the usecase for a specific purpose is reflected in the description. And the bech prefix is a generic (cardano) cvote_vk/sk.

Tools like hw-cli, or the spo scripts, cntools, can write out the files with the right type & description depending on the purpose/path. HW-Wallets would only show the generic cvote_ prefixed bech string for the voting keys.

@JaredCorduan what do you think?

JaredCorduan:
> @JaredCorduan what do you think?

I think what you have proposed sounds great! I appreciate the effort to reduced any confusion that could arise from the voting keys used in Catalyst and those related to CIP-1694.

gitmachtl:
> FYI @JaredCorduan @gitmachtl @Ryun1 @janmazak this is on the CIP Meeting agenda for tomorrow & when @KtorZ is also there we can assess whether there is a consensus about [#427 (comment)](https://github.com/cardano-foundation/CIPs/pull/427#issuecomment-1383960821), and if so then we could be ready to approve & merge this (cc @disassembler).

sounds pretty good to me :-)

disassembler (approved):

rphair (approved):
With unilateral approval after the last discussion & changes it looks like this is ready to merge at today's CIP meeting.

SebastienGllmt (approved):

Issue #420, opened 2022-12-28
CIP-35 and the Draft status


JaredCorduan:
The `Draft` status of CIPs has been removed. The Plutus category of CIPs, however, still makes use of the distinction between `Draft` and `Proposed` (see CIP-35).

rphair:
@KtorZ - @JaredCorduan is right!  [Additions to the Plutus Core Builtins](https://github.com/cardano-foundation/CIPs/tree/master/CIP-0035#additions-to-the-plutus-core-builtins) requires progression through 3 stages: if not {`Draft`, `Proposed`, `Active`} then something else would need to replace `Draft` if they weren't allowed to use that term.

So I would think the solution would be updating `CIP-0001` to indicate that the `Draft` status is _only acceptable for the `Plutus` category_... which removes the inconsistency as far as I can tell.

cc @michaelpj @SebastienGllmt @crptmppt

michaelpj:
CIP-35 needs some work to make it match the new setup better. I think probably I'll just merge the Draft and Proposed criteria.

rphair:
As far as I can tell this is fixed by https://github.com/cardano-foundation/CIPs/pull/428 (thanks @michaelpj); any last words @JaredCorduan before closing?

JaredCorduan:
looks great, thanks all!

PR #428, opened 2023-01-13
CIP-0035 | Redraft to handle Plutus Core language versions

michaelpj:
- Include discussion of Plutus Core language versions and how they change.
- Use "ledger language" instead of "language version" or "ledger language version" to avoid confusion
- Remove the reference to the 'Draft' status.
- Minor prose improvements.

The actual Plutus Core language version was not discussed previously, as it hadn't really seen much use. However we might want to change Plutus Core itself in the medium-term future, so we should make it clear how to do that.

I also tried to clear up some confusing language - the use of 'language version' for the ledger languages was just too confusing, especially in the presence of _another_ language version! So I've opted to just refer to 'ledger languages', which I think is clearer.

I'm planning to do another PR to do more adjustments to make this fit the new CIP-001 better, but I wanted to do this first as it has some substantive changes.

### DIFF
--- a/CIP-0035/README.md
+++ b/CIP-0035/README.md
@@ -27,7 +27,6 @@ The Plutus Core language, its builtins, and its interface to the ledger are all 
 - We may wish to remove elements which have been deprecated due to the addition of improved versions.
 - … and more
 
-At the moment there is no process for making such changes other than the discretion of the core developers. 
 This CIP gives a taxonomy of changes, explains how such changes might be implemented in Cardano, and prescribes processes for proposing such changes.
 
 ## Background
@@ -36,11 +35,13 @@ This CIP assumes general familiarity with Plutus Core and the Cardano ledger, bu
 
 ### Plutus Core
 
-_Plutus Core_ is a script language used in the Cardano ledger. For the purposes of this document, Plutus Core consists of various _language constructs_, and also _builtin types and functions_.
+_Plutus Core_ is a script language used in the Cardano ledger. 
+For the purposes of this document, Plutus Core consists of various _language constructs_, and also _builtin types and functions_.
 
 Plutus Core has a number of builtin types, such as integers, and builtin functions, such as integer addition. 
-Builtin functions provide access to functionality that would be difficult or expensive to implement in Plutus Core code. 
-Builtin functions can operate only over builtin types. Builtin types come with a _size metric_ which is used by costing functions. 
+Builtin functions provide access to functionality that would be difficult or expensive to implement using the basic constructs of Plutus Core, which is otherwise little more than the untyped lambda calculus.
+Builtin functions can operate only over builtin types or arbitrary Plutus Core terms treated opaquely. 
+  * [ ] Builtin types come with a _size metric_ which is used by costing functions. 
 For example, the size metric for integers returns the bit-size of the integer.
 
 The performance of Plutus Core scripts has two components: how expensive the script actually is to run (_real performance_) and how expensive we say it is to run in the ledger (_model performance_). 
@@ -189,13 +190,17 @@ In order to move to Proposed status, it MUST include:
     - For new builtin functions: a costing function for the builtin function.
 - In the Rationale:
     - An argument for the utility of the new builtins.
-    - If an external implementation is provided: an argument that it is trustworthy.
+    - If an external implementation is provided: an argument that it satisfies the following non-exhaustive list of criteria:
+        - It is trustworthy 
+        - It always terminates
+        - It (or its Haskell bindings) never throw any exceptions
+        - Its behaviour is predictable (e.g. does not have worst-case behaviour with much worse performance)
     - Discussion of how any measures and costing functions were determined.
 
 It SHOULD also include:
 - In the Rationale
     - Examples of real-world use cases where the new additions would be useful.
-    - A comparison with an implementation using the existing features, and an argument why the builtin is preferable (e.g. better performance).
+    - If feasible, a comparison with an implementation using the existing features, and an argument why the builtin is preferable (e.g. better performance).
 
 In order to move to Active status, the following must be true:
 - The external implementations MUST be available.
@@ -232,7 +237,7 @@ For example, if a bug fix changes behaviour, it will have to wait for a new Plut
 ### Implementing and releasing changes
 
 This CIP does not cover the process of implementing changes.
-As usual, the CIP process covers the design phase, and it is up to the implementor to ensure that their proposal is implemented, which may require additional work to meet the requirements of the maintainers of the Cardano code repositories (testing, implementation quality, approach), and so on.
+As usual, the CIP process covers the design phase, and it is up to the implementer to ensure that their proposal is implemented, which may require additional work to meet the requirements of the maintainers of the Cardano code repositories (testing, implementation quality, approach), and so on.
 
 Changes can be released after their CIPs have reached Active status. 
 Different changes will require different releases as described in "Types of release".
@@ -268,8 +273,11 @@ For example, it means that we cannot just fix bugs in the semantics: we must rem
 
 It is tempting to think that if we can show that a particular behaviour has never been used in the history of the chain, then changing it is backwards-compatible, since it won’t change the validation of any of the actual chain. 
 However, this is sadly untrue. 
-The behaviour could be triggered (potentially deliberately) in the interval between the update proposal being accepted and it being implemented, which is extremely dangerous. 
-So even if a bug is obscure, we cannot just fix it.
+
+1. The behaviour could be triggered (potentially deliberately) in the interval between the update proposal being accepted and it being implemented. This is extremely dangerous and could lead to an un-managed hard fork.
+2. The behaviour could be triggered in a script which has not yet been executed on the chain, but whose hash is used to lock an output. This could lead to that output being unexpectedly un-spendable, or some other change in behaviour. Moreover, since we only have the hash of the script, we have no way of telling whether this is the case.
+
+So even if a behaviour is obscure, we cannot just change it.
 
 ### Are backwards-compatible binary format changes really safe?
 
--- a/CIP-0035/README.md
+++ b/CIP-0035/README.md
@@ -52,31 +52,61 @@ For example, the costing function for addition says that the CPU and memory cost
 
 Determining costing functions is done empirically, by running the function in question against a large number of inputs and choosing a costing function that fits the data well.
 
+Plutus Core has a _language version_ (LV).
+This is the version of the Plutus Core programming language itself, and it controls e.g. which constructs are available in the language.
+Changing any of the features which are guarded by the language version requires a new language version to be supported by the chain.
+Note that changing the set of builtin types or functions does _not_ require a new language version; any individual Plutus Core language version is compatible with any set of builtin types and functions.
+
+Depending on the type of change, a major or minor bump to the language version may be required.
+The following table shows typical examples.
+
+| Change                                                | Type  | Notes                     |
+|-------------------------------------------------------|-------|---------------------------|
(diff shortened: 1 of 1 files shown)


rphair (commented):
As far as I can tell this also completes https://github.com/cardano-foundation/CIPs/issues/420.  It all looks good to me but @michaelpj could you tag a couple more reviewers who would be willing to go over the text before we merge?

michaelpj:
I'll ask a couple of the Plutus team to check it over.

CIP-0035/README.md
```diff
@@ -156,26 +181,22 @@ So, for example, a single CIP could propose multiple new builtin functions, or a
...
+    - For new builtin functions: a costing function for the builtin function.
+- In the Rationale:
     - An argument for the utility of the new builtins.
+    - If an external implementation is provided: an argument that it is trustworthy.
+    - Discussion of how any measures and costing functions were determined.
 
 It SHOULD also include:
 - In the Rationale
     - Examples of real-world use cases where the new additions would be useful.
     - A comparison with an implementation using the existing features, and an argument why the builtin is preferable (e.g. better performance).
```
kwxm, line 198: In some cases an implementation using the existing features might be difficult or impossible. For example, Implementing a complex cryptographic function using the basic integer operations might be a very large amount of work for something that would clearly be inferior to a C implementation.

Maybe something like "If possible/feasible, a comparison ..."?

CIP-0035/README.md
```diff
@@ -240,10 +261,10 @@ However, this becomes less so as the platform starts to mature and is neither su
...
-### Do removals and changes really need a new language version?
+### Do removals and changes really need a new ledger language?
 
-Not being able to make removals or changes to behaviour without a LV is quite painful. 
-For example, it means that we cannot just fix bugs in the semantics: we must remain bug-for-bug compatible with any given LV.
+Not being able to make removals or changes to behaviour without a LL is quite painful. 
+For example, it means that we cannot just fix bugs in the semantics: we must remain bug-for-bug compatible with any given LL.
 
 It is tempting to think that if we can show that a particular behaviour has never been used in the history of the chain, then changing it is backwards-compatible, since it won’t change the validation of any of the actual chain. 
 However, this is sadly untrue.
```
kwxm, line 270: ... also because there could be an un-triggered validator on the chain that relies on the old behaviour?  Maybe the code is law and that's the owner's problem, but they'd have a reasonable grievance if the behaviour changed from what was advertised when the code was written.

CIP-0035/README.md
```diff
@@ -156,26 +181,22 @@ So, for example, a single CIP could propose multiple new builtin functions, or a
...
+In order to move to Proposed status, it MUST include:
 - In the Specification:
     - Names and types/kinds for the new functions or types.
     - A source for the implementation (e.g. a library which can be linked against); or a generic description of the functionality which is implementable in any programming language.
-    - In the Rationale:
+    - For new types: a precise description of the measure used for the size of a value of that type.
+    - For new builtin functions: a costing function for the builtin function.
+- In the Rationale:
     - An argument for the utility of the new builtins.
+    - If an external implementation is provided: an argument that it is trustworthy.
```
kwxm, line 192: There are certain extra conditions that we'll require to hold: for example the builtin must always terminate and must never throw any exceptions, and its behaviour should be predictable (there shouldn't be any worst-case behaviour that's significantly worse than the costing function would predict).  I don't think we have a formal list of such things yet, but we'll need one eventually and it's not clear to me whether it should go in this CIP or somewhere else.  Maybe we need a builtin implementer's guide.

CIP-0035/README.md
```diff
@@ -240,10 +261,10 @@ However, this becomes less so as the platform starts to mature and is neither su
 Furthermore, while many changes to Cardano are obscure or not of interest to many community members, there is a much larger community who have a keen interest in changes to Plutus Core: dApp developers. 
 Hence it is especially important to have a clear way for this community to be able to propose changes and see how they are progressing.
 
-### Do removals and changes really need a new language version?
+### Do removals and changes really need a new ledger language?
 
-Not being able to make removals or changes to behaviour without a LV is quite painful. 
-For example, it means that we cannot just fix bugs in the semantics: we must remain bug-for-bug compatible with any given LV.
+Not being able to make removals or changes to behaviour without a LL is quite painful. 
+For example, it means that we cannot just fix bugs in the semantics: we must remain bug-for-bug compatible with any given LL.
```
kwxm, line 267: What do we do if we discover that some builtin contains a bug that can compromise the chain, like a division by zero error in a C library?

Also, what do we do if there's a change in the behaviour of the (Haskell or C) library function underlying a builtin at some point in the future? Would we have to maintain a version of the library function reproducing the old behaviour in perpetuity and use it for the execution of existing scripts depending on the old behaviour?

kwxm (approved):
I've added a few comments, maybe not all of which are directly related to this CIP.

Also, "implementer" or "implementor"?  I'm never sure about that myself, but Google and my dictionary seem to favour the former.

All in all I think this does a pretty good job of explaining the complexities of how the different versions of different things all fit together: that's quite confusing so it's very helpful to have it all clearly explained in one place.

kwxm:
A couple of bonus comments.  Near the beginning it says

> Plutus Core has a number of builtin types, such as integers, and builtin functions, such as integer addition. Builtin functions provide access to functionality that would be difficult or expensive to implement in Plutus Core code.

That's kind of self-contradictory.  Maybe it could say something like "... expensive to implement using the basic constructs of Plutus Core"?  It might also be worth having a sentence saying that the non-builtin part of Plutus Core is something like the untyped lambda calculus.

Just below that it says

> Builtin functions can operate only over builtin types.

That's not strictly true (cf `ifThenElse`, where the branches are just terms), but I'm not sure how to explain the true situation concisely.

CIP-0035/README.md
```diff
@@ -52,31 +52,61 @@ For example, the costing function for addition says that the CPU and memory cost
...
+
+Depending on the type of change, a major or minor bump to the language version may be required.
+The following table shows typical examples.
+
+| Change                                                | Type  | Notes                     |
+|-------------------------------------------------------|-------|---------------------------|
+| Adding a construct to the language                    | Minor | Backwards-compatible.     |
+| Removing a construct from the language                | Major | Not backwards-compatible. |
+| Changing the behaviour of a construct in the language | Major | Not backwards-compatible. |
+| Changing the binary format of the language in a backwards-compatible way | Minor | Safe even if it makes previously non-deserializable scripts deserializable.[^backwards-safe] |
```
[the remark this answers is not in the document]
michaelpj: I blame markdown. It's hard to do reliable section links so I just didn't try. I guess I can try again.

CIP-0035/README.md
```diff
@@ -156,26 +181,22 @@ So, for example, a single CIP could propose multiple new builtin functions, or a
...
 - In the Specification:
     - Names and types/kinds for the new functions or types.
     - A source for the implementation (e.g. a library which can be linked against); or a generic description of the functionality which is implementable in any programming language.
-    - In the Rationale:
+    - For new types: a precise description of the measure used for the size of a value of that type.
+    - For new builtin functions: a costing function for the builtin function.
+- In the Rationale:
     - An argument for the utility of the new builtins.
+    - If an external implementation is provided: an argument that it is trustworthy.
+    - Discussion of how any measures and costing functions were determined.
```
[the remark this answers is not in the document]
michaelpj: I think I'm deliberately leaving that a little vague because I'm not sure. I think I want them to _talk_ about it but I think we might want more or less depending on the proposal :thinking:

CIP-0035/README.md
michaelpj: I'll add some of that.

CIP-0035/README.md
michaelpj: I think that's all out of scope of this CIP (since I would say that's "emergency measures"), but
1. No script can ever depend on behaviour that crashes the node. So we can deal with such things. If there's wrong but non-crashing behaviour... we need to preserve that.
2. Yes, if there's a change in the behaviour of the underlying function we'll need to maintain a version with the old behaviour. I guess that's another criteria for builtins: the behaviour should be as clearly specified as possible, so it won't change.

CIP-0035/README.md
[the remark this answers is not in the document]
michaelpj: Also: I think those benchmarks would be part of the implementation, which is separate from the CIP.

thealmarty (approved):

SebastienGllmt (approved):

rphair (approved):
Looks like @michaelpj last commit addresses last round of feedback but will wait for @KtorZ to approve / merge.

KtorZ (approved):
Happy with the update, although this raises an interesting follow-up question:

Why is there a Plutus-core language version if it's always set to 1.0.0 and, according to this new update, is never going to change?

michaelpj:
> Why is there a Plutus-core language version if it's always set to 1.0.0 and, according to this new update, is never going to change?

Hmmm, the whole point of this update was to convey that it _will_ change if we change any of the constructs of Plutus Core itself. Which bit made it sound like it won't change?

CIP-0035/README.md
```diff
@@ -99,37 +130,32 @@ Changes to Plutus Core can be released onto Cardano in four ways, with ascending
...
-| Adding a new builtin function or type                                      | HF (rarely LV[^2]) | Backwards-compatible. Requires a binary format change.                                              |
-| Removing a builtin function or type                                        | LV                 | This will cause scripts which use this builtin to be rejected, so is not backwards compatible.      |
-| Changing the behaviour of a builtin function or type                       | LV                 | This changes the behaviour of existing scripts, so is not backwards compatible.                     |
-| Changing the interface between the ledger and the interpreter              | LV                 | The ledger must provide scripts with exactly the right interface. New interface means new language. |
-| Improving model performance                                                | PP                 | _Must_ strictly follow an improvement in real performance.[^3]                                      |
+| Adding a new Plutus Core language version                                 | HF                 | Backwards-compatible since the new behaviour is guarded by the new LV. |
```
zliu41, line 144: Doesn't this depend on what changed in the new LV? Imagine a new LV that removes builtin type `Data` (by the way, do we still need `Data` once we have sums of products?). This would lead to an unavoidable change in the ledger-script interface, necessitating a new LL.

KtorZ:
> @michaelpj Which bit made it sound like it won't change?

Every mention of the Plutus language being removed 😅? There's only mention of the ledger language version now (`LL`), which as I understand it refers to the language tag that the ledger uses to distinguish between different Plutus version. It is thus different from the on-chain program version that encapsulate every UPLC programs, and that has stayed to `1.0.0` so far.

michaelpj:
> Every mention of the Plutus language being removed sweat_smile? There's only mention of the ledger language version now (LL), which as I understand it refers to the language tag that the ledger uses to distinguish between different Plutus version. It is thus different from the on-chain program version that encapsulate every UPLC programs

Now we have the LV for the "program version" and LL for the different Plutus ledger languages. And both can change - there's an explicit type of change for adding a new LV.

CIP-0035/README.md
michaelpj: Sure, but then you're also including a change to the ledger-script interface. If you do multiple changes obviously you need to biggest kind of release.

zliu41: Fair enough.

kwxm:
There seems to be something wrong with some of the links in the [version](https://cips.cardano.org/cips/cip35/) that's currently publicly visible.

CIP35

rphair:
thanks @kwxm - this is a known issue (footnotes not working) (https://github.com/cardano-foundation/CIPs/issues/435) which we'll be able to resolve when most CIPs are edited for the new format & we can overhaul the web site generator.

Issue #389, opened 2022-12-01
Update older CIPs for revised CIP-0001


KtorZ:
The CIP-0001's rework has been merged. In this rework, we've changed the preamble header structure and renamed some of the sections. Thus, for the sake of consistency, we'll need to update past CIPs to reflect this new structure. This issue is meant to help organizing the work around that.  For each CIP, we need to:

- Convert `Authors` to an actual bullet-point list (one name per line, starting with `-`)
- Remove any field that is no-longer supported (e.g. `type`)
- Add a `Discussions` field with a list (same as authors) of links to past discussions. At the very least, this should contain a link to the original PR (`git log` comes in handy to figure this out).
- Add any other new fields specified by CIP-0001 (`Category`, `Implementors`...).
- Some sections have been renamed and need to follow the structure given by the new CIP template. In particular, the _path to active_ needs to divided into two sub-sections. It may very much be the case for older CIPs that the _path to active_ is incomplete; which should be reflected in the new structure (CIPs without implementors or implementation plan shall be explicitly labelled as such).
- Update the top level `README.md` if there is a status change for this CIP in the table.

> **Warning** We may introduce categories as we see fit; however, categories that would refer to specific areas of the ecosystem (e.g. `Ledger` which also applies to RSS proposals) need to be endorsed by the respective teams of maintainers. Until this happens, we should refrain from using these categories and simply use `N/A` in the meantime.

In terms of process; here below is a list of all (currently) merged CIPs; Each item can be converted into issue and assigned to one person. Thus, if you intend to help, make an issue out of one of the point, assign it to you and make a follow-up PR with the changes. Once all CIPs have been updated, we should be able to repair the website build.

- [x] #720
- [x] #390
- [x] #397
- [x] #400
- [x] #451
- [x] #644
- [x] #532
- [x] #646
- [x] #649
- [x] #651
- [x] #653
- [x] #655
- [x] #402
- [x] #657
- [x] #603
- [x] #659
- [x] #661
- [x] #642
- [x] #663
- [x] #410
- [x] #665
- [x] #667
- [x] #533
- [x] #534
- [x] #669
- [x] #671
- [x] #673
- [x] #675
- [x] #677
- [x] #681
- [x] #682
- [x] #684
- [x] #686
- [x] #689
- [x] #691
- [x] #604
- [x] #693
- [x] #695
- [x] #697
- [x] #699
- [x] #701
- [x] #704
- [x] #706
- [x] #708
- [x] #710
[...]

PR #429, opened 2023-01-15
CIP-0013 | Adjust preamble and structure w.r.t CIP-0001

rphair:
Aside from general reformatting as per https://github.com/cardano-foundation/CIPs/issues/389:

- Standard CIP sections were H1s and other subheadings were too deep.
- There is no section "Read More" (which was a bare list of related URLs) in the standard format so I've embedded those links inline. @SebastienGllmt @v-almonacid feel free to suggest any corrections to this for the links you've chosen.


KtorZ (approved):
Thank you @rphair

Issue #433, opened 2023-01-17
Allow multiple categories?


michaelpj:
I'm updating CIP-35  to be more in line with the new CIP-1. I notice that lots of Plutus changes are _also_ Ledger changes, e.g. changes to the ledger-script interface, protocol parameter changes. The obvious thing for me to say is that in these situations a CIP should be in both categories and get reviewed by both sets of reviewers. But right now CIPs can only have one category.

KtorZ:
I recall us discussing the case of multiple categories in the rework of CIP-0001 and, the point was that having only a single category makes it easier for classification. Plus, there's almost always a main category, even though the work may indeed impact other areas of the ecosystem. I am afraid that by allowing multiple categories, we'll soon end up with CIP referencing half a dozen of categories to which they barely relate. 

I find it easier to only consider one main category; even if some CIPs may actually fit multiple on occasions.

michaelpj:
Okay, that's reasonable. I think it may cause it's own confusion, but fine.

michaelpj:
Actually, I'm still unsure. Consider the case of the Ledger and Plutus categories. Suppose that both of these require some process, e.g. that the submitter propose changes to the corresponding specifications. A CIP in the overlap of the two _should_ have to follow both of the processes, i.e. we _do_ need changes to both specifications. I can say something like "a CIP which would also fall under the Ledger category should be categorized under Plutus but must also follow the process of the Ledger category", but that seems clunky? :thinking:

fallen-icarus:
I couldn't find any proposals (merged, closed, or open) that use the *Ledger* category. The cips [website](https://cips.cardano.org/all/) doesn't even have the option for *Ledger* sorting. Plus, looking at CIP-0035, it seems to me that all plutus changes, except for real performance changes, require a hardfork. AFAIK this means the ledger team must be involved for all plutus changes except for real performance changes. Even CIPs 31,32, and 33 use the *Plutus* category despite making changes to the ledger.

Perhaps the *Plutus* and *Ledger* category should just be merged into a new *Plutus/Ledger* category and the software update changes should go into a new *Node* category. If possible, the processes for the original *Plutus* and *Ledger* categories should just be merged into one process.

[...]

PR #437, opened 2023-01-17
CIP-0035 | Update for new CIP-1

michaelpj:
This updates CIP-35 to be in line with the new CIP-1.

The main changes are:
- Remove some sections that are IMO implied by CIP-1, e.g. the necessity of a clear implementation plan.
- Delete the CIP registry, since you can just list the category
- Explicitly identify the 'Plutus' category and list the reviewers. At the moment it's just me, maybe it should be more people?
- Attempt to clarify the prose a bit.

There's one very questionable thing that I've done, which is to assert that CIPs in the crossover between the Ledger and Plutus should be in both categories. I just went ahead and wrote that here, but I defer to the resolution of https://github.com/cardano-foundation/CIPs/issues/433 and I'll change this depending on what gets decided there.

### DIFF
--- a/CIP-0031/README.md
+++ b/CIP-0031/README.md
@@ -2,10 +2,13 @@
 CIP: 31
 Title: Reference inputs
 Authors: Michael Peyton Jones <michael.peyton-jones@iohk.io>
-Comments-Summary: No comments
-Comments-URI:
+Implementors: 
+    - Michael Peyton Jones <michael.peyton-jones@iohk.io>
+    - Jared Corduan <jared.corduan@iohk.io>
 Status: Active
-Type: Standards Track
+Categories: 
+    - Plutus
+    - Ledger
 Created: 2021-11-29
 License: CC-BY-4.0
 ---
--- a/CIP-0031/README.md
+++ b/CIP-0031/README.md
@@ -6,9 +6,7 @@ Implementors:
     - Michael Peyton Jones <michael.peyton-jones@iohk.io>
     - Jared Corduan <jared.corduan@iohk.io>
 Status: Active
-Categories: 
-    - Plutus
-    - Ledger
+Category: Plutus
 Created: 2021-11-29
 License: CC-BY-4.0
 ---
--- a/CIP-0035/README.md
+++ b/CIP-0035/README.md
@@ -1,6 +1,6 @@
 ---
 CIP: 35
-Title: Plutus Core Evolution
+Title: Changes to Plutus Core
 Authors: Michael Peyton Jones <michael.peyton-jones@iohk.io>
 Comments-Summary: No comments
 Comments-URI: 
@@ -10,11 +10,13 @@ Created: 2022-02-09
 License: CC-BY-4.0
 ---
 
-# Evolution of Plutus Core in the Cardano ledger
+# Changes to Plutus Core
 
 ## Abstract
 
-This CIP proposes a process for proposing changes to Plutus Core, its builtins, and its interface to the Cardano ledger.
+This CIP specifies a process for proposing changes to Plutus Core, its builtins, and its interface to the Cardano ledger.
+It gives a taxonomy of typical changes, and explains how these changes may be made, which in some cases requires a CIP.
+It introduces the 'Plutus' CIP category for tracking these.
 
 ## Motivation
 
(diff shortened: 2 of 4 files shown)


CIP-0035/README.md
```diff
@@ -1,6 +1,6 @@
 ---
 CIP: 35
-Title: Plutus Core Evolution
+Title: Changes to Plutus Core
 Authors: Michael Peyton Jones <michael.peyton-jones@iohk.io>
 Comments-Summary: No comments
 Comments-URI:
```
rphair, line 6: This CIP preamble needs the same makeover as CIP's 31 to 33 (some obsolete fields here).  Note we will also be double checking this in #389.

CIP-0031/README.md
```diff
@@ -2,10 +2,13 @@
...
-Comments-Summary: No comments
-Comments-URI:
+Implementors: 
+    - Michael Peyton Jones <michael.peyton-jones@iohk.io>
+    - Jared Corduan <jared.corduan@iohk.io>
 Status: Active
-Type: Standards Track
+Categories: 
+    - Plutus
+    - Ledger
```
rphair, line 11: The duplicate category seems perfectly understandable to me but this syntax still awaits confirmation via issue https://github.com/cardano-foundation/CIPs/issues/433 and probably also (@KtorZ) an update to https://github.com/cardano-foundation/CIPs/tree/master/CIP-0001#categories.

rphair (commented):
@michaelpj this is great and also helps for compliance with the new standardisation proceeding via #389.  At first view the document structures all look parseable (headers at the right levels) but the new form also needs:

a _Path to Active_ for each one, even ones that are currently `Active`.
- This is undocumented (or implied) so far because https://github.com/cardano-foundation/CIPs/tree/master/CIP-0001#status-proposed only lists these criteria for `Proposed` proposals.  
- But we are adding this section for all merged CIPs as per https://github.com/cardano-foundation/CIPs/issues/389#issue-1470899761 and if adding a history of acceptance & implementation retroactively then nobody understands these better than you, so it makes the most sense for you to write it.
- @KtorZ if my understanding after several GitHub exchanges is that we must have _Path to Active_ for every single proposal is incorrect, now would be a good time to correct it (and also to confirm it).

a section at the bottom with a link to the License quoted in the header (I / other editors can edit these later in if it gets missed here).

another nitpick which we might also handle in concierge style: each of the Motivation and Rationale sections now have slightly longer titles:
- Motivation: why is this CIP necessary?
- Rationale: how does this CIP achieve its goals?

michaelpj:
Thanks, I'll fix those things.

I can add the "Path to Active" although I'm not sure what I'd say. By its nature this CIP is Active in virtue of us endorsing it, so... what?

rphair:
> - Motivation: why is this CIP necessary?
> - Rationale: how does this CIP achieve its goals?

@michaelpj since you were also editing CIP-0001 when the longer titles were assigned to these 2 sections maybe you could please also contribute your opinion about the latter part of https://github.com/cardano-foundation/CIPs/issues/439.

CIP-0031/README.md
KtorZ: There's actually no `Ledger` category.

KtorZ: I mean, it still is not registered. We've added it for some existing CIPs; but I would refrain to add it to any new ones..

KtorZ:
> @michaelpj By its nature this CIP is Active in virtue of us endorsing it

I'd be happy to see just this as a path to active :)

CIP-0031/README.md
michaelpj: I'm assuming there will be one, though.

michaelpj:
I've addressed the comments, and removed the multiple categories. I've attempted to say something sensible about what to do in cases where things fall in the overlap, but we can fix it up later if it proves problematic I guess.

rphair (commented):
thanks @michaelpj - this all looks both compliant & well explained to me.

@KtorZ are we strictly requiring that the H1 document titles contain the CIP-XXXX identifier as well?
- I ask mainly because they've been edited into CIPs 2 through 4 (and I followed suit with 13) in the course of #389.
- It seems to me this could be left up to the author considering the number will be in the published URL and also just above in the header.

Sorry for another nitpick from the format nanny, but also trying to make sure whatever CI we might invent to normalise the presentation doesn't have to be more like an AI. 🤓

CIP-0031/README.md
KtorZ: I've been assuming for a couple of months too now 🙃

KtorZ:
> @KtorZ are we strictly requiring that the H1 document titles contain the CIP-XXXX identifier as well?

Perhaps not. I am on the fence with that. It reads better on Github, but it's currently redundant on the website (see e.g. https://cips.cardano.org/cips/cip1/). Plus, there's the "problem of section heading" that you outlined in #439 ... which makes me think we might just consider dropping it and bumping all section headers to `h1` level 🤷

michaelpj:
Okay, sounds like you're happy with this overall then?

rphair (approved):
Yes 😎

KtorZ (approved):

CIP-0031/README.md
rphair: just noting for cross-reference: there is soon going to be one 🙂 https://github.com/cardano-foundation/CIPs/pull/456/files

PR #438, opened 2023-01-17
Fix website build

### LINKED ISSUE #436: Website is not updating
The website doesn't seem to have updated recently, e.g. the new CIP-001 is not present there.

KtorZ:
Fixes #436 

- :round_pushpin: **Fix editors parsing + provide better debugging on malformed input data.**
  
- :round_pushpin: **Display CIPs by category, if it exists.**
    This also sorts categories such that 'All' is always first and 'Unclassified' is always last.

### DIFF
--- a/CIP-0049/README.md
+++ b/CIP-0049/README.md
@@ -1,25 +1,15 @@
 ---
-CIP: 49 
-Title: ECDSA and Schnorr signatures in Plutus Core  
-Authors: Koz Ross (koz@mlabs.city), 
-         Michael Peyton-Jones (michael.peyton-jones@iohk.io), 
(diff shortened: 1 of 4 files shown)


rphair:
Just going to try buidling locally using this new thing.  Even if doesn't work will approve this anyway, but would prefer to give some feedback if possible.  I'm guessing that the CIPs edited are the ones with some kind of breaking syntax.

rphair (approved):
hey, that's nice... in `scripts` running `node build.js` creates a `public` directory at top level with everything apparently in it (looks like familiar web site, but complete with new material, when running `python3 -m http.server` there).  That looks like a complete test to me but let me know if I missed anything.

Issue #423, opened 2023-01-06
CIP-27 | Smart contract royalty addresses


Tenancio:
CIP-27 is the current royalty standard for Cardano NFTs however it isn't mentioned anywhere how a marketplace should deal with a royalty address which is a smart contract address. The problem with this is that if a marketplace sends royalties to a smart contract without a datum then these funds are locked permanently.

I think it should either be put in the standard that royalties should be sent with any basic datum or a tag should be added to the royalty token metadata saying the address is a SC so that marketplaces can add a datum to only those transactions. This would then prevent any funds being locked in a SC.

rphair:
@Tenancio the best I can do with this is to tag the original authors & contributors to the discussion of https://github.com/cardano-foundation/CIPs/pull/116: @huths0lo @matiwinnetou @gitmachtl @SebastienGllmt @Crypto2099 @KtorZ

Maybe after discussing a bit here you could submit either

- an update CIP for CIP-0027 (if there's a convergence on your suggestion or other straightforward alternative), or
- a CPS if it's a problem without a particular solution and/or a wider variety of approaches.

huths0lo:
This is a valid point.  And I think its probably time to start adding handles to the royalty token as well.  I'll probably look to make an update in the near term, where a few new tags are added:
'version': 2
'handle': $yourhandle #superceeds the addr if the marketplace supports it
'sc_datum': {your needed datum}

mgajda:
Option would be to add an `Acceptor` script that only accepts valid payments.
Every eUTxO to a smart contract would have to call it in order to confirm that datum is acceptable.

huths0lo:
> Option would be to add an `Acceptor` script that only accepts valid payments. Every eUTxO to a smart contract would have to call it in order to confirm that datum is acceptable.

I doubt there is a practical way to make that work at the moment.  The marketplaces would need to be able to provide the necessary datum when sending to a smart contract.  And theres no details within the existing standard that could provide that info to the marketplaces.

I'm not a plutus expert, but if there was a way to provide the needed "datum" within the 777 metadata, then it might be doable.  I also think we need to add a "handle" tag, that would allow the usage of a wallet handle (obviously Ada Handle being the currently well known handle).  The idea of using an asset as the identifier of the royalty wallet was proposed by @nicolasayotte way back when the original 777 standard was developed.  This was well before Ada Handle existed.  So I have to give him credit where credit is due.

[...]

PR #440, opened 2023-01-18
CIP-0087? | Maybe Datum

fallen-icarus:
A CIP that looks to address the concerns of #364, #423, #309, #310, and #321 in one go.

See proposal [here](https://github.com/fallen-icarus/CIPs/tree/maybe-datum/CIP-%3F%3F%3F%3F).

### DIFF
--- a/CIP-????/README.md
+++ b/CIP-????/README.md
@@ -15,13 +15,15 @@ License: CC-BY-4.0
 # CIP-XXXX: Maybe Datum
 
 ## Abstract
-Currently all plutus spending scripts take a datum directly from the UTxO being spent. While this allows for creating a very simple ledger-script interface, there are a number of use cases where this feature is not just undesired but can also be an obstical to Dapp developement. I propose changing plutus scripts so that they optionally take the datum directly from the UTxO being spent. This single change would solve a number of issues currently motivating different CIPs/CPSs. 
+Currently all plutus spending scripts take a datum directly from the UTxO being spent. While this allows for creating a very simple ledger-script interface, there are a number of use cases where this feature is not just undesired but can also be an obstacle to Dapp developement. I propose changing all plutus scripts so that they optionally take the datum directly from the UTxO being spent. This single change would solve a number of issues currently motivating different CIPs/CPSs. 
 
 ## Motivation: why is this CIP necessary?
 As of now, plutus spending scripts take the datum directly from the UTxO being spent. While it allows the ledger-script interface to be very simple, it causes three major problems.
 
 ### Problem 1: Spending Script Redundant Executions
 Using the eUTxO model, it is currently possible to create spending scripts that validate based off of the transaction as a whole instead of any individual UTxO. Imagine if three UTxOs were spent from such a script's address within the same transaction. Which datum should be passed to the spending script? Currently, the ledger-script interface passes each UTxO's datum to the script individually and executes this "transaction level" spending script three times. Since the transaction context does not change between executions, these additional executions are completely redundant. There is currently no way to use a transaction level spending script so that these redundant executions do not happen. This limitation significantly handicaps Cardano's usage of the eUTxO model. While there is a potential workaround, the workaround has its own major drawbacks. You can read more about it in the related CPS [here](https://github.com/fallen-icarus/CIPs/tree/full-eUTxO-potential/CPS-%3F%3F%3F%3F#readme).
+
+Note: While Cardano simple scripts do not use a datum, they may also suffer from this limitation.
 
 ### Problem 2: Wrong/Missing Datums Result in Permanently Locked Script UTxOs
 When a script's UTxO is missing a datum or has the wrong datum, the plutus spending script is unable to parse the datum attached to the UTxO and is therefore guaranteed to fail. This results in permanently locked UTxOs at the script's address. In order to help prevent accidental locking of script UTxOs, proxies are generally used. The idea is for users to send their funds to this proxy first. The proxy then attaches the required datum for the users and pass the funds onto the desired script address. However, it is very difficult to use such proxies in a decentralized way. In short, by being forced to use proxies, the Dapp loses some of its decentralization.
@@ -95,11 +97,14 @@ Since there is no need to parse the datums attached to each UTxO, the datum flag
 
 Users **MUST** explicitly state which level to use. Building a transaction should fail if the level is not explicitly stated.
 
+#### Simple Scripts
+Given that Cardano simple scripts are only for multisig and time locking, and both scenarios are ones that validate based off of the transaction as a whole, simple scripts are only usable as transaction level scripts. If the simple scripts language is going to be extended in the future, perhaps it would "future proof" the ledger-script interface to still have the option for using simple scripts at the UTxO level even though there are no use cases right now.
+
 #### Ensuring All Script Witnesses Are Present
 The node is already capable of detecting if all required witnesses are present. This tooling will just need to be adapted to allow transaction level scripts to witness all UTxOs while being executed only once. The idea is similar to only one pubkey signature being required no matter how many UTxO are being spent from the corresponding address.
 
 #### Does This Require A Hardfork?
-According to [CIP-0035](https://github.com/michaelpj/CIPs/blob/8e296066c0afc7d2ed46db88eca43f409830e011/CIP-0035/README.md#scripts-in-the-cardano-ledger), what is being suggested here is a change to the ledger-script interface. This means a new Plutus Core ledger language (LL) is required. The CIP states that this requires a hardfork.
+According to [CIP-0035](https://github.com/michaelpj/CIPs/blob/8e296066c0afc7d2ed46db88eca43f409830e011/CIP-0035/README.md#scripts-in-the-cardano-ledger), what is being suggested here is a change to the ledger-script interface. This means a new Plutus Core ledger language (LL) is required. The CIP states that this requires a hardfork. However, being that the way things are stored on-chain will not change, this should be a smaller hardfork than the past few.
 
 ## Rationale: how does this CIP achieve its goals?
 The key idea of this CIP is to address multiple developer concerns as simply as possible. There have been (at least) 6 proposals created so far whose motivations can all be addressed by the `Maybe Datum` ([#364](https://github.com/cardano-foundation/CIPs/pull/364),[#423](https://github.com/cardano-foundation/CIPs/issues/423),[#309](https://github.com/cardano-foundation/CIPs/pull/309),[#418](https://github.com/cardano-foundation/CIPs/pull/418),[#310](https://github.com/cardano-foundation/CIPs/pull/310),[#321](https://github.com/cardano-foundation/CIPs/pull/321)). 
--- a/CIP-????/README.md
+++ b/CIP-????/README.md
@@ -1,5 +1,5 @@
 ---
-CIP: ?
+CIP: 87
 Title: Maybe Datum
 Category: Plutus
 Status: Proposed
--- a/CIP-????/README.md
+++ b/CIP-????/README.md
@@ -1,0 +1,112 @@
+---
+CIP: ?
+Title: Maybe Datum
+Category: Plutus, Ledger
+Status: Proposed
+Authors:
+    - fallen-icarus <modern.daidalos@gmail.com>
+Implementors: []
+Discussions:
+Created: 2023-01-18
+License: CC-BY-4.0
+---
+# CIP-XXXX: Maybe Datum
+
+## Abstract
+Currently all plutus spending scripts take a datum directly from the UTxO being spent. While this allows for creating a very simple ledger-script interface, there are a number of use cases where this feature is not just undesired but can also be an obstical to Dapp developement. I propose changing plutus scripts so that they optionally take the datum directly from the UTxO being spent. This single change would solve a number of issues currently motivating different CIPs/CPSs. 
+
+## Motivation: why is this CIP necessary?
+As of now, plutus spending scripts take the datum directly from the UTxO being spent. While it allows the ledger-script interface to be very simple, it causes three major problems.
+
+### Problem 1: Spending Script Redundant Executions
+Using the eUTxO architecture, it is currently possible to create spending scripts that validate based off of the transaction as a whole instead of any individual UTxO. Imagine if three UTxOs were spent from such a script's address within the same transaction? Which datum should be passed to the spending script? Currently, the ledger-script interface passes each UTxO's datum to the script individually and executes this "transaction level" spending script three times. Since the transaction context does not change between executions, these additional executions are completely redundant. There is currently no way to use a transaction level spending script so that these redundant exections do not happen. This limitation significantly handicaps Cardano's usage of the eUTxO architecture. While there is a potential workaround, the workaround has its own major drawbacks. You can read more about it in the related CPS [here](https://github.com/fallen-icarus/CIPs/tree/full-eUTxO-potential/CPS-%3F%3F%3F%3F#readme).
+
+### Problem 2: Wrong/Missing Datums Result in Permanently Locked Script UTxOs
+When a script's UTxO is missing a datum or has the wrong datum, the spending script is unable to parse the datum attached to the UTxO and is therefore guaranteed to fail. This results in permanently locked UTxOs at the script address. In order to help prevent accidental locking of script UTxOs, proxies are generally used. However, it is very difficult to use such proxies in a decentralized way. In short, by being forced to use proxies, the Dapp loses some of its decentralization.
+
+As an addition point, regulators are not going to like any application where it is easy to accidentally lock (lose) users' funds permanentally. Regulators may force certain certifications for Dapp developers as a consequence. The motivation is exactly the same for why finanical advisors need to be licensed. Currently, this ease of accidental locking is the default on Cardano. This needs to be changed if we do not want regulators forcing developer certifications that are inherently centralizing.
+
+### Problem 3: No Universal Plutus Script
+Since spending scripts take a datum while minting and staking scripts do not, it is currently not possible to create a universal script (one that can do all three). There are use cases where the Dapp would be more secure if the same script could be used for both spending and minting. An example is to only mint a token if a deposit is made to the relevant script address. 
+
+Right now, if a developer wanted to do this, he/she would need to hardcode the spending script hash into the minting policy AND the minting policy id needs to be passed to the spending script as a datum. If a universal script could be used, the spending script hash would be the same as the minting policy id. In the former case, the security of this process depends on how well the datum usage can be guarded. In the latter case, the process is secure by default since the hash cannot be faked/wrong.
+
+## Specification
+### `Data` Specification for `Maybe Datum`
+Before the ledger-script interface passes the datum to the plutus script, the datum will be wrapped in a `Maybe` data type. Seen as type `Data` this wrapping is:
+
+``` Haskell
+Nothing = Constr 0 []
+Just datum = Constr 1 [datumAsData]
+```
+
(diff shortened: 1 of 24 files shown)


CIP-????/README.md
```diff
@@ -0,0 +1,121 @@
+---
+CIP: ?
+Title: Maybe Datum
+Category: Plutus, Ledger
```
rphair, line 4:
```suggestion
Category: Plutus
```
The current editorial view is that we should only have one category per CIP: https://github.com/cardano-foundation/CIPs/issues/433#issuecomment-1386614231

and specifically (via precedent of https://github.com/cardano-foundation/CIPs/pull/437) Plutus proposals which also affect the Ledger should be given a category of `Plutus` (unless maybe needing more involvement from the Ledger team): https://github.com/cardano-foundation/CIPs/issues/433#issuecomment-1386885858

rphair (commented):
Looks good at first view @fallen-icarus & thanks for drawing all these discussions together.

nielstron:
I approve this CIP. The issue of accidentally locking UTxOs for good is widespread and problematic and the use cases for multipurpose scripts are clear enough to make them easy to achieve by design.

nielstron:
Note that the motivation/problem 3 is not accurate. A solution to this problem and the corresponding drawback is outlined here: https://github.com/ImperatorLang/eopsin/blob/master/ARCHITECTURE.md#minting-policy---spending-validator-double-function

fallen-icarus:
@nielstron I updated the problem 3 section to acknowledge `eopsin`'s workaround. My apologies for not doing it sooner.

nielstron:
Thanks!

ch1bo:
PR #309 has a related discussion to this. 

https://github.com/cardano-foundation/CIPs/pull/309#discussion_r939157351

fallen-icarus:
@ch1bo What are your thoughts about this CIP? It is meant to address the concerns of 309 as well. Do you agree that it does?

imikushin:
@fallen-icarus In this model, can there be a situation when there's one script address, and:
- one transaction would spend from it running the script as a tx-level script,
- another transaction would spend from it running the script per-UTxO?

Another question: can a transaction run several tx-level scripts? (Perhaps, when spending outputs from _M_ addresses, created from _N_ scripts, where _M_ > _N_) How would this work?

CIP-????/README.md
```diff
@@ -0,0 +1,168 @@
...
+``` Bash
+  --tx-in <script_utxo1> \
+  --tx-in <script_utxo2> \
+  --tx-in <script_utxo3> \
+  --tx-level-spending-tx-in-reference <ref_script_utxo_id> \
+  --plutus-script-v3 \
+  --tx-level-spending-reference-tx-in-redeemer-file <script_redeemer> \
+```
+
+For transaction level scripts, since there is no need to parse the datums attached to each UTxO, the datum flags are not needed. Any datums present will still appear in the `ScriptContext` that gets passed to the plutus script. All UTxOs being validated by the transaction level script are assumed to be using the same redeemer which is why it is okay to place the redeemer flag at the end.
```
imikushin, line 98: What if we're building a transaction involving multiple tx-level scripts each expecting their own redeemer data (not compatible with others)?

fallen-icarus:
@imikushin The same script can be used in either context. You can think of all scripts as having logic like
``` Haskell
if maybe_datum == Just datum
then utxo_level_logic
else tx_level_logic
```
> can a transaction run several tx-level scripts? ... How would this work?

You can use as many tx-level scripts as will fit into the tx. The following table shows some valid usages:

| Tx 1 | Tx 2 | Tx 3 | Tx 4 |
|--|--|--|--|
| Script A as tx-level | Script A as tx-level | Script A as utxo-level | Script A as utxo-level |
| Script B as tx-level | Script B as utxo-level | Script B as tx-level | Script B as utxo-level |

The only restriction is that, within the same tx, script A cannot be used as BOTH a tx-level script and a utxo-level script. As far as how this would work, the ledger is already capable of detecting if all the relevant scripts are present. Currently, every utxo needs an accompanying script to execute at the utxo-level. This detection tooling will need to be slightly altered to also allow using tx-level scripts to witness all utxos from that script's address in the same way that a pubkey can witness all utxos from the user's address.

> Perhaps, when spending outputs from M addresses, created from N scripts, where M > N

This is an interesting use case: *If Address AA and Address AB are protected by Script A, can script A be used at the tx-level for Address AA's utxos while being used at the utxo-level for Address AB's utxos in the same tx?* I would have to say no simply because this would likely dramatically complicate the ledger-script interface and I don't think this use case is ubiquitous enough to warrant adding such complexity. For most use cases, the script is probably meant to be used as only one level or the other, regardless of the address being spent from. If this usage is necessary, you can always break it up over multiple txs. Somebody on the plutus or ledger team will know better how complicated allowing this niche use case would be. The proposal, as it is now, does not allow for this use case.

CIP-????/README.md
fallen-icarus: Each tx-level script will get its own redeemer. So if Script A and Script B were used in the same tx, Script A will use redeemer A and Script B will use redeemer B. The `--tx-level` redeemer flag is meant to be paired up with a specific `--tx-level` script for this reason. This proposal does not make all scripts in the tx use the same redeemer.

imikushin:
@fallen-icarus, thanks for your reply. With clarifications about multiple tx-level scripts, I can definitely get behind this proposal. 

**Tx-level validation** (as opposed to per-UTxO) basically gives more **scalability for free** (besides, of course, implementation cost), i.e. we're not paying for it by sacrificing security or decentralization.

imikushin:
@fallen-icarus Just a thought: maybe a better name for this CIP could be Transaction Level Validation Scripts.

I think it communicates the solution much better. Sure, script re-usability is an issue. But a much bigger concern is scalability, which this proposal addresses. 

I mean, just think of a transaction dealing with multiple outputs that has to perform a ZK proof verification (a pretty intensive computation). How much easier it gets with per-transaction script execution vs per-output.

fallen-icarus:
@imikushin I had thought about that before but tx level scripts aren't the main focus of this CIP. While I personally agree with the tx level scripts being the most important consequence of this CIP (since you can't fully utilize the eUTxO model without it), the main focus is on changing things to use the `Maybe Datum`. @rphair Do you have an opinion on this?

rphair:
@fallen-icarus I agree the "Transaction" oriented name would be eye-catching for implementations, but the goal of any standard is to be applied as generally as possible.  Therefore if the current name "Maybe Datum" is the most generally applicable I'd be in favour of keeping it in the title unless there's a more explicit title that means the same thing (as a non Plutus programmer it means nothing to me except what people have said about this & related CIPs).

@imikushin's highlighted use case is a great one to keep focusing on, but from my (not technically expert on this subject) point of view you have to read deeply into the text to find how this phrase applies to the subject.  So maybe that use case could be included under a heading like "Use cases for Transaction Level Validation Scripts" in the Motivation or Rationale section.

michaelpj (commented):
This CIP contains two very different proposals:
- A fairly straightforward proposal to make datums optional. This is fine, I would like to do something like this, but overall I prefer https://github.com/cardano-foundation/CIPs/pull/321
- A complicated proposal to reduce how often scripts are executed in some cases.

I think it would be a stronger proposal if these were separated (from the title I assumed it would just be the first part). In particular the second needs _much_ more development, as it currently glosses over a lot of important details.

----

I think the overall proposal could be carved up and made simpler and more compositional.

- Proposal 1: when the ledger accumulates script evaluations to perform as `(Script, [Data])` tuples, it deduplicates this list before performing evaluations, so as not to do redundant work. This would need a CIP because it would need to specify the budget discount for the non-evaluated scripts, otherwise it's pointless.
- Proposal 2: optional datums or https://github.com/cardano-foundation/CIPs/pull/321
- Proposal 3: something with this special flag designed to make more script evaluations look identical

This would allow for proposal 1 to get implemented, while we're still working out what the best way to get more identical script evaluations is (whether it's proposal 3 or something else).

CIP-????/README.md
```diff
@@ -0,0 +1,168 @@
...
+
+Further, there are a number of use cases where a datum is not needed for the Dapp to function. In these situations, a dummy datum is still required just to prevent locking.
+
+As an addition point, regulators are not going to like any application where it is easy to accidentally lock (lose) users' funds permanentally. Regulators may force certain certifications for Dapp developers as a consequence. The motivation is exactly the same for why finanical advisors need to be licensed. Currently, this ease of accidental locking is the default on Cardano. This needs to be changed if we do not want regulators forcing developer certifications that are inherently centralizing.
```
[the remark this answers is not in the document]
fallen-icarus: I agree that funds can still be locked with bad logic. However, I think comparing locking due to arbitrary logic and locking due to missing a datum is an apples to orange comparison. The former can be extensively tested and patched while the latter cannot. Still, I can look to alter the wording to convey this better.

CIP-????/README.md
```diff
@@ -0,0 +1,168 @@
...
+
+### Problem 3: No Universal Plutus Script
+Since plutus spending scripts take a datum while plutus minting and staking scripts do not, it is currently not possible to create a universal plutus script (one that can do all three). There are use cases where the Dapp would be more secure if the same script could be used for both spending and minting. An example is to only mint a token if a deposit is made to the relevant script address.
```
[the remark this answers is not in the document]
fallen-icarus: Consider the following example (this is an actual example I have encountered). I have a Dapp spending script that can be paired with any staking key/script. However, I would like to be able to tag all addresses that were made using the spending script with a dedicated token in order to tell them apart from other addresses. The requirements:
1. The token cannot be minted to an address that is not using the Dapp's spending script.
2. Once the token is at the Dapp's address, the token cannot be withdrawn. It can only be burned.

The solution to (1) is to hard code the spending script hash into the minting policy. Now the minting policy can check that the token is never minted to an address not using the spending script. The solution to (2) is to require the minting policy ID to be attached to the token's output in the datum. Now the spending script can prevent that token from leaving since it knows the policy ID.

The problem is in number 2. What if a user deposits a utxo with a datum containing a different policy ID? While this does not allow misuse of the Dapp's token, if the token with the supplied policy ID is present and not burnable by the user, that utxo is now locked forever since the spending script will not allow withdrawing that token. The spending script can only go off of what is in the datum.

This issue would not exist if the same script could both mint and spend since the script would automatically know the proper policy ID. Saying it would be "more secure" may not be the most appropriate description for this but IMO neither is saying it would be "easier to write". Perhaps saying it would be "easier to guarantee proper usage" would be better.

CIP-????/README.md
```diff
@@ -0,0 +1,168 @@
...
+## Specification
+### `Data` Specification for `Maybe Datum`
+Before the ledger-script interface passes the datum to the plutus script, the datum will be wrapped in a `Maybe` data type. Seen as type `Data`, this wrapping is:
+
+``` Haskell
+Nothing = Constr 0 []
+Just datum = Constr 1 [datumAsData]
+```
+
+The datum will only be converted from `BuiltinData` if the `Just` is going to be used. No parsing should be attempted if the `Nothing` is going to be used.
```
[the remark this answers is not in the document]
fallen-icarus: I just realized those two sentences are wrong so I will definitely need to change them. Those two sentences should probably be something like: 

The ledger-script interface will only assume the script needs a datum if it is told to use the `Just` constructor. It will then look for the `BuiltinData` for the script to use and wrap it in the `Just` constructor before passing it to the script. If the ledger-script interface is told to use the `Nothing` constructor, it won't bother looking for a `BuiltinData` for the script to use (and therefore won't throw an error if it can't find one). It will simply pass the `Nothing` constructor to the script.

CIP-????/README.md
```diff
@@ -0,0 +1,168 @@
...
+#### Spending
+Using the universal scripts, there will now be two ways to use a spending script.
+
+1. **UTxO level spending scripts**: These scripts will have the datum attached to the UTxO being spent parsed and passed to the universal script as `Just datum`. These scripts are meant to be executed once per UTxO consumed from the script address.
+2. **Transaction level spending scripts**: These scripts will have `Nothing` passed to the universal script without attempting to parse the datums attached to the UTxOs being spent. These scripts are meant to be executed once per transaction. The datums will still be present in the `ScriptContext`.
+
+#### `cardano-cli`
+`cardano-cli` will need to be expanded to add `--utxo-level` and `--tx-level` flags so that users can explicitly say which level of the spending script to use.
+
+Example UTxO level spending script flag usage:
```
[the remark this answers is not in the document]
fallen-icarus: Technically, the developer decides what to do. If the script is only meant to be run as one or the other, an error can be thrown when the wrong level is used. If the developer would like the creator of the output to choose, the datum can be used to specify the wishes of the output creator. Since most tx-level scripts will be traversing all input datums anyway, the script can fail if all datums are not flagged for tx-level usage (if a datum isn't present or is wrong, it can be used in either level). Allowing the spender to decide the level is just the default.

fallen-icarus:
@michaelpj thanks for the review. I'll need to think about the rest of your comments here.

CIP-????/README.md
```diff
@@ -0,0 +1,168 @@
...
+
+### Problem 1: Spending Script Redundant Executions
+Using the eUTxO model, it is currently possible to create spending scripts that validate based off of the transaction as a whole instead of any individual UTxO. Imagine if three UTxOs were spent from such a script's address within the same transaction. Which datum should be passed to the spending script? Currently, the ledger-script interface passes each UTxO's datum to the script individually and executes this "transaction level" spending script three times. Since the transaction context does not change between executions, these additional executions are completely redundant. There is currently no way to use a transaction level spending script so that these redundant executions do not happen. This limitation significantly handicaps Cardano's usage of the eUTxO model. While there is a potential workaround, the workaround has its own major drawbacks. You can read more about it in the related CPS [here](https://github.com/fallen-icarus/CIPs/tree/full-eUTxO-potential/CPS-%3F%3F%3F%3F#readme).
```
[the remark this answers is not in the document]
fallen-icarus: I do not believe these requirements are accurate. I believe you are being overly restrictive. Staking scripts and minting scripts are already executed at the tx-level despite having redeemers and using the `ScriptPurpose`.

ShariffL:
@oversize @KtorZ @katomm adding you in this discussion as well. Is there a way to get this CIP approved and merged? It will improve scalability for multi output tx CC: @imikushin

fallen-icarus:
@ShariffL I agree with @michaelpj that this CIP should be broken up into a few composable CIPs. Currently, I am considering breaking it up into two CIPs:

1. Using `Maybe Datum` for when a utxo is missing one. This would also enable universal scripts (one that can mint, spend, or manage rewards). It is mutually exclusive to #321.
2. Adding a new tx input type for grouped inputs. See [here](https://github.com/cardano-foundation/CIPs/pull/418#issuecomment-1427046710). This would address the redundant executions and enable "ignoring" datums even if utxos have them (they will still be present in the `ScriptContext`).

CIP-????/README.md
```diff
@@ -0,0 +1,98 @@
+---
+CIP: ?
```
```suggestion
CIP: 87
```

CIP-????/README.md
```diff
@@ -0,0 +1,98 @@
...
+Authors:
+    - fallen-icarus <modern.daidalos@gmail.com>
+Implementors: []
+Discussions:
+    - https://github.com/cardano-foundation/CIPs/pull/321
+    - https://github.com/cardano-foundation/CIPs/pull/418
+Created: 2023-01-18
+License: CC-BY-4.0
+---
+# CIP-XXXX: Maybe Datum
```
KtorZ, line 15:
```suggestion
```

CIP-????/README.md
```diff
@@ -0,0 +1,97 @@
...
+    2. This increases the need to trust software that is being used to get the datum information from the extended address. A malicious frontend can alter the datum information so that it can steal funds from the user.
+- Using a proxy - users send their funds to a middle contract/entity that then attaches the required datum for the users and passes the funds onto the desired script address. Currently, the only way to use proxies in a trustless manner is to use a plutus script as the proxy. But then we are back to where we started: a missing datum will result in the permanent locking of the funds sent to the proxy contract. So the only real proxy workaround for completely preventing locking due to a missing datum is to use a multisig native script or another pubkey address and have the owner(s) attach the proper datum for the users. Both of these options force reliance on centralized parties.
+
+As a final point, there are use cases where a datum is not needed for the Dapp to function. In these situations, a dummy datum is still required just to prevent locking.
```
ch1bo, line 29: To give another use case: There are situations where the sender _should not need to be aware_ they are sending funds to a script.

An example to this are "Smart contract wallets". These wallets would lock funds in a script instead of a private key and can be used to implement do social recovery wallets or for-purpose debit accounts.

CIP-????/README.md
```diff
@@ -0,0 +1,97 @@
...
+
+With this change, UTxOs would no longer be permanently locked simply for missing a datum. Further, it would now be possible to use the same plutus script for spending, minting, and staking.
+
+### Backwards Compatibility
+Only plutus scripts that use the new plutus ledger language will have their datums handled in this way. All other plutus ledger languages will retain their usage.
+
+### `ScriptArgs` Alternative
+The `ScriptArgs` proposal ([here](https://github.com/cardano-foundation/CIPs/pull/321)) could also address the issues motivating this CIP. The `ScriptArgs` approach could allow for easily implementing future changes such as allowing scripts to take other kinds of arguments. However, all of the details have yet to be ironed out.
```
michaelpj, line 85: Since this is the major competitor to your proposal, I think you need to say a _lot_ more about why you think yours should be implemented instead!

fallen-icarus: @michaelpj Do you think there is value in implementing the `Maybe Datum` for now while we sort out the `ScriptArgs`? The main difference between the two is that the `ScriptArgs` is easily extendable, but there are still questions around it. The `Maybe Datum` proposal does not take us very far from the existing design. IMO the issues discussed in this CIP are pressing enough to prioritize fixing soon. If we implement the `Maybe Datum` proposal first, we can then take our time sorting out the `ScriptArgs`. If `ScriptArgs` is deemed to be better, how much different is the process of going from our current design to `ScriptArgs` than going from the `Maybe Datum` design to `ScriptArgs`?

Being able to sort out the `ScriptArgs` without the pressure caused by the discussed issues is my main argument for why we should do the `Maybe Datum`. Basically, the idea is to use this CIP as a potential stopgap.

michaelpj: No, absolutely not. Changes that require new Plutus ledger languages are very expensive because they require us to support old versions _forever_. I am extremely not keen to add something that we plan to remove later.

fallen-icarus: @michaelpj Understood. Then I don't have an argument for choosing this proposal over the `ScriptArgs` one since I personally believe that, if the `ScriptArgs` can be added safely, it should be used. However, I think there is value in documenting what has been considered at each step of evolution for the design - this has been my main motivation for continuing to steel man this CIP despite thinking that `ScriptArgs` would likely be better. I am totally fine with this proposal getting merged, but then marking its status as inactive once `ScriptArgs` is ready. I'm not sure if the CIP process is meant to be used this way.

Considering that you did something similar with #336, what are your thoughts about this? I'm willing to wait until the `ScriptArgs` proposal is ready and then try arguing against it.

fallen-icarus:
The others can correct me if I am wrong, but as the author of this CIP, I do not think this is being implemented in Conway. CIP-69 (#784) uses a `Maybe Datum` in a different way than this CIP was proposing. IMO, CIP-69 obsoletes the need for this one.

lehins:
@fallen-icarus Yes, you are right in that CIP-0069 takes a different approach to solve the same problem. In that sense, it does make this CIP obsolete.

rphair:
Thanks everybody especially @fallen-icarus for all the prior work on this.  Based on the last comments by @lehins @fallen-icarus @Ryun1 I'm closing this as superseded, but please feel free to reopen if that turns out not to be true.

Issue #413, opened 2022-12-11
[New CIP]Post-quantum crypto in Plutus


mgajda:
NIST Post Quantum Cryptography work group has selected three digital signature algorithms for standardization:

* DILITHIUM-CRYSTALS
* Falcon
* SPHINX+

Estimates for PQC migration are within next 15 years, so it those smart contracts that want PQC hardness would certainly benefit from option to check PQC signatures.

While PQC algorithms are expected to be more secure than previous generation, the main issue is the signature size (666 bytes for Falcon, 2420 bytes for Dilithium, 8080 bytes for Sphincs+).

For this reason, these signature algorithms may be chosen by long-term security applications.

Proposal would involve:

* implementing Falcon, Dilithium, and Sphincs+ as `Crypto` class instance,
* adding Plutus instructions to call these signature algorithms,
* providing a test "vesting" contract that releases funds after combination of signatures is given.

When additional PQC algorithms are standardized by NIST and enjoy shorter signature size, we may add these as well. (But we should not hold, since the process may take years, possibly more than estimated for migration.)

Future PQC-hardening of wallets and blockchain should be considered a sequel to this proposal.

rphair:
@mgajda this sounds like it would make an interesting CIP, but also keep in mind the [CPS format](https://github.com/cardano-foundation/CIPs/tree/master/CIP-9999) in case your problem is more general than a specific proposal.  There are some CIPs already for Plutus changes which maybe help to review the discussion on,  if you haven't already: including one for increased cryptographic support (https://github.com/cardano-foundation/CIPs/pull/250).

rphair:
That should be fine as a proposed CPS, which could also include what you mentioned specifically in #414.  The general procedure is to first submit the CPS draft on GitHub: later, when the new PR comes up for review in our bi-weekly meeting, it would be assigned the next CPS number if & when it's accepted as a candidate.

Another recommendation is to first post on the Cardano Forum where I can see you have an account: the Developers and CIPs groups might provide good initial discussion before you submit your draft.  It can also be marked literally as "Draft" in the GitHub PR if you'd like to leave it posted on GitHub while you keep working on it (pending review).

rphair:
NIST progress posted in last month: [NIST Releases First 3 Finalized Post-Quantum Encryption Standards](https://www.nist.gov/news-events/news/2024/08/nist-releases-first-3-finalized-post-quantum-encryption-standards)

rphair:
[...]

PR #441, opened 2023-01-19
CIP-???? | Post Quantum signatures and native wallets

mgajda:
This is preliminary draft for #413: PostQuantum signature primitives in Plutus and native wallets.

NIST Post Quantum Cryptography work group has [selected three digital signature algorithms for standardization](https://csrc.nist.gov/projects/post-quantum-cryptography/selected-algorithms-2022):

* DILITHIUM-CRYSTALS
* Falcon
* SPHINX+

Estimates for PQC migration are within next 15 years, so it those smart contracts that want PQC hardness would certainly benefit from option to check PQC signatures.

While PQC algorithms are expected to be more secure than previous generation, the main issue is the [signature size](https://sofiaceli.com/2022/07/05/pq-signatures.html) (666 bytes for Falcon, 2420 bytes for Dilithium, 8080 bytes for Sphincs+).

For this reason, these signature algorithms may be chosen by long-term security applications.

Proposal would involve:

* implementing Falcon, Dilithium, and Sphincs+ as `Crypto` class instance,
* adding Plutus instructions to call these signature algorithms,
* providing a test "vesting" contract that releases funds after combination of signatures is given.

When additional PQC algorithms are standardized by NIST and enjoy shorter signature size, we may add these as well. (But we should not hold, since the process may take years, possibly more than estimated for migration.)

Future PQC-hardening of wallets and blockchain should be considered a sequel to this proposal.

---

([rendered proposal in branch](https://github.com/mgajda/CIPs/blob/michal/post-quantum-plutus/CIP-%3F%3F%3F%3F/README.md))


mgajda:
@L-as The NIST has issues standardization recommendation already, and IBM already announced over 400-bit quantum computer. Recent papers have shown how to hack another signature algorithm with only ~260 qbits.

Waiting for further research sounds like waiting until the current signature algorithms are hacked.

L-as:
@mgajda I agree with this sentiment, disregard my previous comment (however, I still believe better algorithms will appear).

I however think we should support the algorithms noted natively everywhere signatures are used.

I think Cardano should as much as possible switch to a signature algorithm agnostic approach as it's likely we'll change again. Plutus should have a single verifySignature builtin, that takes an integer index to choose the algorithm. Every other part of the system should be changed similarly.

mgajda:
@L-as It may be good idea to add a general `verifySignature` interface, since I sincerely hope even better signature algorithms appear. I will add this to a proposal.

This should not delay the implementation, so it is great improvement.
Of course we will need to keep existing signature verification functions with both interfaces to keep compatibility.

SebastienGllmt:
These signatures are:
- relatively expensive computationally & space-wise
- costly to implement (not just in Haskell, but every client that appears in the future)
- make it harder to replace Plutus with a ZK compute layer in the future (if there is interest in that)
- have no guarantee not to be replace by something better

So if we really want to implement one in Plutus, I would prefer if we picked one instead of all of them. However, I'm more convinced that if we want to handle PQCs in the short term, it may be better to prioritize using them in overlay networks like Mithril which is the same thing Algorand did by having Falcon-signed state proofs periodically generated for their chain.

L-as:
I don't think it will affect verifying Cardano itself in a ZK manner. I don't think Mithril is an overlay network, it's just a stake-weighted multisignature scheme. It also wouldn't fix the issue. L1 needs to be resistant against the attacks.

SebastienGllmt:
> I don't think it will affect verifying Cardano itself in a ZK manner

Definitely it does in the same way any cryptography in a zk Plutus Core would need to be verifiable. Of course it doesn't stop you from building anything ZK that doesn't need to be isomorphic 

> I don't think Mithril is an overlay network, it's just a stake-weighted multisignature scheme

Sure, but this is like saying blockchains aren't peer-to-peer networks, but just a chain of cryptographic signatures

> L1 needs to be resistant against the attacks.

It does in the same way State Proofs on Algorand does (which uses the same ideas as Mithril). If you mean it doesn't in the sense that individual transactions won't be secure (since state proofs are just quantum-resistant checkpoints), then I agree but I think the checkpoints are good enough to be able to kick the can a bit further down the road

iquerejeta:
I think PQ security is something that should definitely be in the roadmap of Cardano. However, I don't see the benefit of having a post-quantum signature in plutus while the node uses traditional signatures. In my opinion it makes more sense to go the other way. First include PQ security at the node/consensus level, and then include it in Plutus.

I think the suggestion of @SebastienGllmt of doing it in Mithril is a good starting point, as it provides quantum safe checkpoints without the need of having to do massive changes in the node.

iquerejeta:
> Mithril isn't a thing yet
Agree, but we can say it's an ongoing effort. More info can be found here: https://mithril.network/doc/. But you are right, some design decisions are not yet established. 

> allowing all signatures everywhere in Cardano to use other algorithms by abstracting over the algorithms in question everywhere.
This would be a cool feature I think, and might facilitate the addition of new signature schemes for plutus in the future (or switching between signing algorithms in the ledger). But wouldn't you say we already have something like that with DSIGN? 

And as I said before, I don't think it makes sense to use quantum safe crypto in plutus if the ledger does not use it. Using quantum safe crypto in the ledger is non trivial, as it might affect network capabilities, diffusion times, computational requirements by the node, etc. I'm by no means suggesting that we shouldn't move forward with PQ crypto. I'm simply saying that maybe the first step is to make a proposal as to which algorithm could be used in the ledger, and how it would affect the network.

SebastienGllmt:
> However, I do think we should start from the other way around, allowing all signatures everywhere in Cardano to use other algorithms by abstracting over the algorithms in question everywhere.

Maybe the best path for this would be to add PQ crypto to native scripts and improve account abstraction using native scripts as well

mgajda:
> These signatures are:
> 
> * relatively expensive computationally & space-wise

That is why use of PQC will be opt-in at the smart contract level.

> * make it harder to replace Plutus with a ZK compute layer in the future (if there is interest in that)

That would be totally orthogonal, and such ZK layer would also have to be quantum-resistant.

> * have no guarantee not to be replace by something better

This is true for all cryptography. Especially classical cryptography in the face of Quantum Computing revolution.

> So if we really want to implement one in Plutus, I would prefer if we picked one instead of all of them. However, I'm more convinced that if we want to handle PQCs in the short term, it may be better to prioritize using them in overlay networks like Mithril which is the same thing Algorand did by having Falcon-signed state proofs periodically generated for their chain.

Note that Plutus initially only supported Ed25519 and Blake2. Now we see that Schnorr and EcDSA are necessary.
Why not to add all feasible implementations?

Falcon is indeed recommended, since signature length is shortest.

mgajda:
Answers to some comments above:

* @SebastienGllmt I have added native wallets as you suggested.
* @iquerejeta We now include DSIGN instance as part of this CIP.
* Security properties have been emphasised, since latest research brings uncertainty about how much time do we have to migrate.

mgajda:
@michael-liesenfelt Would you review too?

mgajda:
Since it may be easier to agree on the problem statement itself, I added CPS.
I ask reviewers to consider it a matter of possible urgency, given that IBM plans to double the quantum computing capacity by 2025.

Please let me know if you would prefer to accept CPS separately.

rphair:
Thanks @mgajda for updates about the relevance of this proposal.

I believe the CIP and CPS should be submitted in separate PRs mainly because, as you say above, the discussion & acceptance process will be different in each case.  cc @KtorZ

mgajda:
I will split them as soon as preliminary review is made.
It would be useful to cross-reference discussion between CIP and CPS.

abailly:
@mgajda I must admit my near total ignorance of PQC beside the basic principle so I would be hard pressed to qualify the urgency level, but I will check the references you provide and do some research to better understand the issue at stake.

Couple of more or less minor points: 
* There's some issue with markdown which breaks rendering from the `Hash algorithm` point on. 
* This proposal falls under the change process outlined in  [CIP-0035](https://github.com/cardano-foundation/CIPs/tree/master/CIP-0035#additions-to-the-plutus-core-builtins), it's not clear to me if all the requirements there are met, in particular w.r.t. providing some evidence on the feasibility of an implementation
* @michaelpj Should be the most relevant person to provide feedback on the integration path for this feature

@L-as Mithril _is_ a thing, you can try it _now_ to restore a full-node (Eg. Daedalus wallet) on `preview` and `preprod` networks, with `mainnet` support "around the corner."

mgajda:
@michaelpj I made my best with respect to putting both motivation (CPS, intro), and implementation (long term signature scheme).
But I badly need your review, since entire CIP part of this MR falls under [CIP-0035](https://github.com/cardano-foundation/CIPs/tree/master/CIP-0035#additions-to-the-plutus-core-builtins).

@abailly Fixed markdown rendering issue just now.
I tried my best to provide links to a [PQClean](https://github.com/PQClean/PQClean) implementation of crypto algorithms using standard reference code published by NIST.

While exact urgency level maybe disputed, US published an order to consider PQC transition for any information systems from July 2023 on. Conservative safety is a hallmark of cryptography, so I believe we should look into this sooner rather than later.

mgajda:
@SebastienGllmt I have added note on native wallets. Do you believe it is sufficient?

CIP-????/README.md
```diff
@@ -0,0 +1,294 @@
...
+so migration to post-quantum cryptography algorithms should start several
+years before they would become insecure.
+This summer NIST has recommended three PQC signature algorithms
+for standardization, publishing their reference implementations,
+and finishing initial comment period.
+
+Supporting PQC signatures will bring the following benefits:
+
+* smart contracts may use future standard algorithms today;
+* early support for PQC will simplify future PQC migration, if needed;
```
michaelpj, line 42: it's unclear to me what the PCQ migration is and how this enables it? I would like to see this elaborated somewhere. It seems that you have an idea of what it is and how we would do it, but it's not explained anywhere.

CIP-????/README.md
```diff
@@ -0,0 +1,294 @@
...
+```
+
+This API needs additional support from the off-chain applications in order to:
+
+* provide correct public key for the given algorithm
+* help users to transfer their euTxOs to new signature algorithms when old algorithms are obsoleted.
+
+Note that `LongTermHash` and `LongTermSignature` can be implemented as standard library functions.
+
+## Rationale
```
michaelpj, line 247: I would like to see some explanation of _specifically_ how this proposal achieves the benefits and use cases listed in the Motivation. It is currently not clear to me.

CIP-????/README.md
```diff
@@ -0,0 +1,294 @@
...
+* future PQC migration would be smooth and require less changes;
+* support for PQC may increase industry acceptance for long running projects;
+* users that want their wallets to survive over ten years may
+  use smart contracts verifying multiple algorithms today;
+* Cardano will become first blockchain to verify PQC standard signatures,
+  thus getting first-mover advantage.
+
+## Use cases
+
+Adding multiple PQC signature algorithms will allow the following:
```
michaelpj, line 52:
```suggestion
Adding multiple PQC signature algorithms to Plutus Core will allow the following:
```

CIP-????/README.md
```diff
@@ -0,0 +1,294 @@
...
+
+* opt-in migration of wallets and smart contracts to safer cryptographic protocol
+* pay-as-you-go for longer signatures required for post-quantum resistance
```
michaelpj, line 55: again, not sure what this means

CIP-????/README.md
```diff
@@ -0,0 +1,294 @@
...
+* opt-in resistance to _cryptocalypse_ - surprise breakage of a single major encryption algorithm -
+  by using multiple signature algorithms for redundancy.
+* compliance with US government regulations past July 2023, when all agencies
+  are expected to plan post-quantum migration.
+
+## Specification
+
+[`DSIGN`]() class instance should be provided for the post-quantum signature algorithms.
```
michaelpj, line 63:
```suggestion
### `cardano-base`

[`DSIGN`]() class instance should be provided for the post-quantum signature algorithms.
```

CIP-????/README.md
```diff
@@ -0,0 +1,294 @@
...
+
+Two new builtin functions would be provided:
+
+* A verification function for both pre-quantum and post-quantum resistant
+  signature schemes:
+     - CRYSTALS(DILITHIUM) signatures;
+     - FALCON signatures;
+     - SPHINCS signatures; and
+     - SEQUOA
+* Extensible interface for signature verification functions.
```
michaelpj, line 75: Same comment as in the CPS: it's not clear why this is entangled with the PQC stuff.

CIP-????/README.md
```diff
@@ -0,0 +1,294 @@
...
+This is a reference implementation of the three kinds of signature schemes in C. This 
+implementation would be called from Haskell using direct bindings to C. These
+bindings would be defined in `cardano-base`, using its existing `DSIGN`
+interface, with new builtins in Plutus Core on the basis of the `DSIGN`
+interface for both schemes.
+
+The builtins would be costed as follows: each signature verification has
+linear cost proportional to the number of message bits, and square with respect to key length.
+_Constants will be proportional to the ratio of time as benchmarked against
+Ed25519._
```
michaelpj, line 94: I think you are suggesting a very complex cost model, namely something that branches on the first argument and then is costed differently (at least wrt constants) depending on that. We could probably add such a thing, but it seems complicated.

CIP-????/README.md
```diff
@@ -0,0 +1,294 @@
...
+              | Blake2
+              | EcDSA
+              | SECP256k1
+              | Schnorr
+              | Falcon
+              | Sphincs
+              | Dilithium
+              | Sequoa
+```
+To facilitate extensibility, the name of algorithm should be rendered as case-insensitive text in JavaScript
```
michaelpj, line 110: I'm confused about what this has to do with the rest of the proposal?

CIP-????/README.md
```diff
@@ -0,0 +1,294 @@
...
+| Sphincs+                |                 32 |              8080^[https://sofiaceli.com/2022/07/05/pq-signatures.html#:~:text=A%20table%20of-,comparison%20of%20sizes,-%3A] |                                 2.8^[Benchmark was for Sphins\*128s] | [Selected](https://csrc.nist.gov/projects/post-quantum-cryptography/selected-algorithms-2022) |
+| sequoa863pq128          |                108 |               216 |                                    ? | Submitted^[Communication on ebats@cryp.to mailing list.] |
+
+  : Reference for signature, key sizes, and time cost multipliers [as suggested here](https://blog.cloudflare.com/nist-post-quantum-surprise/). In the order of preference, with comparison against current default signature algorithm (Ed25519).
+
+The following function should then be used to verify arbitrary binary signed
+with permitted signature scheme:
+
+```
+verifySignWithAlgo :: SignAlgo
```
michaelpj, line 127: I think this is the heart of the "extensibility" proposal. The obvious question is "why not just write this function in user-space, having it delegate to the appropriate single-algorithm builtin?"

Here's my guess at the scenario you're trying to support where that wouldn't be enough (please actually write it down yourself though, I'd prefer not to guess!):

1. Dapp creator writes script S using `verifySignWithAlgo`
2. S can dynamically (?) choose which signature algorithm it is using, maybe by doing something with a value in the datum (?)
3. We update the Plutus ledger language being used by S such that `verifySignWithAlgo` has additional algorithms (I'm actually not sure we can do this safely!)
4. S can be used as-is, only requiring a change to e.g. its datum to get it to use the new algorithm.

This has several tricky steps, notably 2 and 3. It's hard for me to say whether or not we could do that since I'm not sure what you're trying to do, but I think we might not be able to do those (esp 3) without e.g. making some scripts succeed that previously would have failed at runtime, which we shouldn't really do.

CIP-????/README.md
```diff
@@ -0,0 +1,294 @@
...
+
+1. A verification key (length depends on the algorithm chosen);
+2. An input to verify (either the message itself, and a hash);
+3. A signature.
+
+The two different schemes handle deserialization internally: specifically, there
+is a distinction made between 'external' representations, which are expected as
+arguments, and 'internal' representations, used only by the implementations
+themselves. This creates different expecations for each argument for both of
+these schemes; we describe these below.
```
michaelpj, line 146: where? I couldn't find it. Generally I'm confused about this

CIP-????/README.md
```diff
@@ -0,0 +1,294 @@
...
+
+For the [all signature
+schemes](),
+the keys are required to be given in the same format as NIST test vectors.
```
michaelpj, line 150: Please be specific or link to the specific description.

CIP-????/README.md
```diff
@@ -0,0 +1,294 @@
...
+
+The builtin operations will error with a descriptive message if given inputs
+that don't correspond to the constraints above, return `False` if the signature
+fails to verify the input given the key, and `True` otherwise.
+
+### Post-quantum wallets
+
+We should permit using all chosen signature algorithms to lock transactions.
```
michaelpj, line 158: :rotating_light: :rotating_light: :rotating_light: 

I assume you mean "to lock UTXOs".

This is A HUGE CHANGE and not a Plutus one. This is now veering into changing all the cryptography in Cardano as a whole. I _highly_ recommend making a separate CIP for that, as I think it will be extremely difficult and contentious and needs a lot of support.

CIP-????/README.md
```diff
@@ -0,0 +1,294 @@
...
+
+```
+data SignAlgo = SHA2_256
+              | SHA3_256
+              | Blake2
+
+hashWithAlgo :: HashAlgo -> BuiltinByteString -> BuiltinByteString
+```
+
+This will also allow for gradual _opt-in_ migration of scripts as the blockchain evolves.
```
michaelpj, line 200: Again, this is about extensibilty, and I still think it should be clearly separated out and argued for.

CIP-????/README.md
```diff
@@ -0,0 +1,294 @@
...
+mostSecureHash :: BuiltinByteString -> LongTermHash
+mostSecureHash = hashWithAlgo mostSecureHash
+```
+
+This API needs additional support from the off-chain applications in order to:
+
+* provide correct public key for the given algorithm
+* help users to transfer their euTxOs to new signature algorithms when old algorithms are obsoleted.
+
+Note that `LongTermHash` and `LongTermSignature` can be implemented as standard library functions.
```
michaelpj, line 245: So this is just a proposal to add some stuff to the `plutus` standard library? That doesn't need a CIP! I suggest removing it to keep things slim and opening an issue in `plutus`.

CIP-????/README.md
```diff
@@ -0,0 +1,294 @@
...
+functionality or affect any other builtins. Likewise, at levels above Plutus
+Core (such as `PlutusTx`), no existing functionality should be affected.
+
+On-chain, this requires a hard fork.
+
+## Path to Active
+
+An implementation will require the following steps:
+
+1. Linking to PQC library.
```
michaelpj, line 272: Which library? That is a big deal!

CIP-????/README.md
```diff
@@ -0,0 +1,294 @@
...
+2. Providing the patch implementing new Plutus primitives just like for [EcDSA and Schnorr](https://github.com/input-output-hk/plutus/pull/4368).
+3. Providing the PQC primitives using `Crypto` interface for ease of use.
+
+## Discussion full quantum resistance
+
+This CIP _does not_ provide for full quantum resistance of the Cardano blockchain yet,
```
michaelpj, line 278: I wouldn't expect it to, it says it's about Plutus.

michaelpj (commented):
Needs some work, there is a lot going on here and it needs to be disentangled. 

I'm not against adding more cryptography primitives if there's clear need for them. I'd prefer not to have to add several if we can avoid it, it's not _that_ easy (witness how long it took us to get the SECP stuff in). 

I'm sceptical about the "extensibility" thing.

CIP-????/README.md
```diff
@@ -0,0 +1,294 @@
...
+
+## Path to Active
+
+An implementation will require the following steps:
+
+1. Linking to PQC library.
+2. Providing the patch implementing new Plutus primitives just like for [EcDSA and Schnorr](https://github.com/input-output-hk/plutus/pull/4368).
+3. Providing the PQC primitives using `Crypto` interface for ease of use.
+
+## Discussion full quantum resistance
```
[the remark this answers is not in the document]
L-as: Even then, it will be obvious that it was done illegally, and will be trivial to revert.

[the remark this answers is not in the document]
mgajda: @iquerejeta Quantum and wallet scripts will need much more time to migrate.
We can do ledger and consensus migration as a last step, since they do not change API.

CIP-????/README.md
mgajda: @michaelpj Thanks Michael, but a lot of people (@iquerejeta, @rphair) asked about this more ambitious target.

CIP-????/README.md
```diff
@@ -0,0 +1,294 @@
...
+in order to decrease cost of implementation.
+
+SHA-256 hash is considered [_safe for a long-time yet_](https://cryptobook.nakov.com/quantum-safe-cryptography#quantum-safe-and-quantum-broken-crypto-algorithms), and any future upgrade would only need to consider SHA-384 as far as we know.
+The main concern would be the public key signatures used for entire wallet infrastructure.
+That is why we propose an _opt-in_ scheme, where users may gradually choose to migrate towards safer, post-quantum cryptography
+over the course of few years.
+That would make Cardano permit **quantum-resistant** smart contracts today at negligible cost,
+and assure **long-term viability** of those smart contracts that will be deployed in the near future.
+
+[Long-term document storage](https://adastamp.co) is a killer application for the blockchain.
```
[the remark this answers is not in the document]
mgajda: Long-term document storage (notarization) needs a long term migration to new signature protocols.

This CIP proposes how to do this in general, beside handling post-quantum signatures as a special case.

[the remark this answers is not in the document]
mgajda:
```suggestion
[Long-term document storage](https://adastamp.co) is a killer application for the blockchain. It greatly benefits from standardized protocol of migration to better digital signature schemes.
```

CIP-????/README.md
mgajda:
```suggestion
1. Linking to PQC library. We propose [PQClean](https://github.com/PQClean/PQClean) since it is a robust distribution of NIST-selected candidates. It also aims at cleanness and ease of formal verification.
```

CIP-????/README.md
```diff
@@ -0,0 +1,294 @@
...
+This API needs additional support from the off-chain applications in order to:
+
+* provide correct public key for the given algorithm
+* help users to transfer their euTxOs to new signature algorithms when old algorithms are obsoleted.
+
+Note that `LongTermHash` and `LongTermSignature` can be implemented as standard library functions.
+
+## Rationale
+
+We consider the implementation trustworthy after passing long comment period,
```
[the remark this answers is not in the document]
mgajda:
```suggestion
We consider the NIST candidates trustworthy after passing a long comment period,
```

CPS-????/README.md
```diff
@@ -0,0 +1,115 @@
...
+Adding multiple PQC signature algorithms will allow the following:
+
+* opt-in migration of wallets and smart contracts to safer cryptographic protocol
+* pay-as-you-go for longer signatures required for post-quantum resistance
+* opt-in resistance to _cryptocalypse_ - surprise breakage of a single major encryption algorithm -
+  by using multiple signature algorithms for redundancy.
+* compliance with US government regulations past July 2023, when all agencies
+  are expected to plan post-quantum migration.
+* allow fast migration of entire blockchain to post-quantum resistance when it becomes necessary.
+* [Long-term document storage](https://adastamp.co) is a killer application for the blockchain.
```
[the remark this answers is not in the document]
mgajda: The protocol for long-term signature storage proposed by AdaStamp means storing signatures on the Cardano blockchain, and "upgrading them", by re-signing with new cryptographic.
That means appending new signatures and hashes, and that requires implementation mentioned in this CIP.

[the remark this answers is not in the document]
mgajda:
```suggestion
* [Long-term document storage](https://adastamp.co) is a killer application for the blockchain. It requires long-term migration to new signature schemes, for the documents
to remain unforgeable.
```

CPS-????/README.md
```diff
@@ -0,0 +1,115 @@
...
+
+Adding multiple PQC signature algorithms will allow the following:
+
+* opt-in migration of wallets and smart contracts to safer cryptographic protocol
+* pay-as-you-go for longer signatures required for post-quantum resistance
+* opt-in resistance to _cryptocalypse_ - surprise breakage of a single major encryption algorithm -
+  by using multiple signature algorithms for redundancy.
+* compliance with US government regulations past July 2023, when all agencies
+  are expected to plan post-quantum migration.
+* allow fast migration of entire blockchain to post-quantum resistance when it becomes necessary.
```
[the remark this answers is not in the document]
mgajda: You are right. The "fast migration" item should be moved to CIP.

```suggestion
```

CPS-????/README.md
```diff
@@ -0,0 +1,115 @@
...
+
+# Use cases
+
+Adding multiple PQC signature algorithms will allow the following:
+
+* opt-in migration of wallets and smart contracts to safer cryptographic protocol
+* pay-as-you-go for longer signatures required for post-quantum resistance
+* opt-in resistance to _cryptocalypse_ - surprise breakage of a single major encryption algorithm -
+  by using multiple signature algorithms for redundancy.
+* compliance with US government regulations past July 2023, when all agencies
```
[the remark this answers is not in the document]
mgajda: Smart contract developers and dApp authors in general.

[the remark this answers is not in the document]
mgajda:
```suggestion
* compliance of smart contracts with US government regulations past July 2023, when all agencies
```

CPS-????/README.md
```diff
@@ -0,0 +1,115 @@
...
+[IBM has a goal of 4000-qbit processor by 2025](https://research.ibm.com/blog/ibm-quantum-roadmap-2025), while a group of researchers
+claimed that [372 physical qbits may be sufficient](https://arxiv.org/pdf/2212.12372.pdf) with few thousand layers to break RSA2048.
+If confirmed, then an existing [400-qbit](https://newsroom.ibm.com/2022-11-09-IBM-Unveils-400-Qubit-Plus-Quantum-Processor-and-Next-Generation-IBM-Quantum-System-Two) quantum processor could break the RSA2048 using [IBM public quantum computing service](https://quantum-computing.ibm.com/) **this year**.
+
+# Use cases
+
+Adding multiple PQC signature algorithms will allow the following:
+
+* opt-in migration of wallets and smart contracts to safer cryptographic protocol
+* pay-as-you-go for longer signatures required for post-quantum resistance
```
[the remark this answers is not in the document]
mgajda: It means that we do not need to immediately increase size of each Cardano block by forcibly upgrading the entire consensus and ledger layers to post-quantum signatures.
We start paying increased digital signature size within smart contracts that _want_ to be safe in the long term. API will remain unchanged.

CPS-????/README.md
```diff
@@ -0,0 +1,115 @@
...
+NIST plans to finalize standardization in 2023/2024 time frame, with US executive branch recommending
+all projects to plan for post-quantum resistance by mid-2023.
+
+[IBM has a goal of 4000-qbit processor by 2025](https://research.ibm.com/blog/ibm-quantum-roadmap-2025), while a group of researchers
+claimed that [372 physical qbits may be sufficient](https://arxiv.org/pdf/2212.12372.pdf) with few thousand layers to break RSA2048.
+If confirmed, then an existing [400-qbit](https://newsroom.ibm.com/2022-11-09-IBM-Unveils-400-Qubit-Plus-Quantum-Processor-and-Next-Generation-IBM-Quantum-System-Two) quantum processor could break the RSA2048 using [IBM public quantum computing service](https://quantum-computing.ibm.com/) **this year**.
+
+# Use cases
+
+Adding multiple PQC signature algorithms will allow the following:
```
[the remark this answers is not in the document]
mgajda: Problem: "pre-quantum signature algorithms will ultimately get hacked".

What other solution may there be than replacing them?

CPS-????/README.md
```diff
@@ -0,0 +1,115 @@
...
+# Abstract
+
+Rapid and ramping up progress of quantum computing threatens safety of traditional
+cryptographic protocols like Ed25519 and SHA256 used in Cardano blockchain.
+While migration of a blockchain itself within a single year is imaginable,
+migration of user wallets and smart contracts will certainly take much longer.
+We should prepare by allowing gradual, opt-in migration of wallets and smart contracts
+well in advance.
+
+# Problem
```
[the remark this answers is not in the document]
mgajda: Yes, ultimately we will need to migrate all signature schemes.

It is only Plutus that needs to migrate sooner, if we want our smart contracts to survive more than few years.

L-as:
Is this CIP worth it when users can lock their funds in a quantum-resistant way using Plutus?
It currently doesn't matter much for consensus, because all consensus does is prevent double-spending.
If someone takes over a majority of the stake and does a double spend, we'd notice and move swiftly without any issue.
SPOs would have their funds locked in a quantum-resistant way, even though the address they stake to isn't.
With Leios, now we rely on the stake to _validate_ logic too, however, there is very likely to be at least one node that notices something is off, meaning that the community will switch to a quantum-resistant Cardano, retroactively undoing any transaction that was illegal.

The issue remains that most people aren't storing their funds in a quantum-resistant way. Wallets should support it, and such a script needs to be developed. It would likely depend on bitwise operations (https://github.com/input-output-hk/plutus/pull/4733), which we don't have yet. A bigger problem is that Plutus discriminates scripts from PKHs, and most scripts only work with PKHs, meaning that most dApps will be inaccessible to such users. This is however not a reason to support quantum-resistant signature schemes, rather, we should fix Plutus.

MicroProofs:
I think if quantum resistance went to being in Plutus only the major issue that comes up is "what about collateral?"

Collateral requires a vkey signer and currently any ada held to be used as collateral would be susceptible to a quantum attack. I don't have an answer, but it's something that points toward a quantum vkey being essential unless we rework collateral.

SebastienGllmt:
@MicroProofs I agree, which is why I made this comment originally. I think if we were to introduce PQC in would have to be part of a larger discussion on account abstract

> > However, I do think we should start from the other way around, allowing all signatures everywhere in Cardano to use other algorithms by abstracting over the algorithms in question everywhere.
> 
> Maybe the best path for this would be to add PQ crypto to native scripts and improve account abstraction using native scripts as well

In an ideal world, we could solve the problem through https://github.com/cardano-foundation/CIPs/pull/749 because you could then encode the PCQ as a Plutus script and then have your account abstraction by using a native script whose execution depends on a Plutus script (that could delegate verification to a PCQ verifier).

It doesn't quite work because although currently native scripts can't pay collateral, there is no reason why they shouldn't be able to in the current implementation of Cardano. This possibility gets hampered a bit with #749 because if you native script depends on a Plutus script, it can't pay collateral anymore

That is to say, I think if we want proper abstraction while still keeping #749, we would need a version of #749 that works with ZK proofs of Plutus execution because that way you could have a native script paying collateral while using a ZK proof of PCQ verification (but now you run into a new problem where you need your ZK proof system to be post-quantum secure such as STARKs, but Cardano in its current form is more friendly to STARKs anyway)

CPS-????/README.md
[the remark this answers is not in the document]
mgajda: Yes, definitely!

CIP-????/README.md
mgajda: The reference C implementations of Post Quantum Crypto candidates from NIST are included in the library that facilitates their use.

mgajda:
@MicroProofs The goal was to have "soft integration" schedule for PQC into Plutus, which is easier than migrating entire blockchain.

First we integrate PQC code as an _option_ for smart contract, dealing with collateral and migration of user accounts later.

mgajda:
> Is this CIP worth it when users can lock their funds in a quantum-resistant way using Plutus? It currently doesn't matter much for consensus, because all consensus does is prevent double-spending. If someone takes over a majority of the stake and does a double spend, we'd notice and move swiftly without any issue. SPOs would have their funds locked in a quantum-resistant way, even though the address they stake to isn't. With Leios, now we rely on the stake to _validate_ logic too, however, there is very likely to be at least one node that notices something is off, meaning that the community will switch to a quantum-resistant Cardano, retroactively undoing any transaction that was illegal.
> 
> The issue remains that most people aren't storing their funds in a quantum-resistant way. Wallets should support it, and such a script needs to be developed. It would likely depend on bitwise operations ([IntersectMBO/plutus#4733](https://github.com/IntersectMBO/plutus/pull/4733)), which we don't have yet. A bigger problem is that Plutus discriminates scripts from PKHs, and most scripts only work with PKHs, meaning that most dApps will be inaccessible to such users. This is however not a reason to support quantum-resistant signature schemes, rather, we should fix Plutus.

Our goal is to enable choice of signature mechanism for Plutus Smart Contracts, so that migration from current signature schemes to PQC signature schemes can be done gradually.

Otherwise PQC would imply obsoleting old signatures overnight, without providing an alternative.

CIP-????/README.md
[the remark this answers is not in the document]
mgajda: @L-as Concern about old generation of signatures is that Quantum Computing can hack them in the foreseeable future, so we need PQC wallets to protect against it.

Migration to PQC wallets will take some time, so it would be nice to provide PQC wallets even before entire blockchain is PQC-protected.

CPS-????/README.md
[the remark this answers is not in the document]
mgajda: @michaelpj Is it clear why we need to start migration of smart contracts and wallets even before we start migration of the blockchain?

mgajda:
@rphair Is there any hope that this CIP will get a number assigned soon?
Given Mithril, is it something that will be supported within CIP process?

rphair:
@mgajda since it's coming up to a year since the last community review I would be in favour of adding it to the next CIP meeting agenda but there's still the pending question about separation between

👉 an implementation oriented CIP (which will be hairy: as @michaelpj points out (https://github.com/cardano-foundation/CIPs/pull/441#discussion_r1127752477) a complete overhaul of the cryptography chain) and

👉 a CPS for the general problem: especially if as @SebastienGllmt says in https://github.com/cardano-foundation/CIPs/pull/441#issuecomment-1399278437 there are multiple signature algorithms to consider, plus different approaches to handling this in Core and/or Plutus first, etc.

I would be in favour of giving this a number as soon as the material can be divided between a CIP and CPS since we have to set the CIP scope properly first.  You had offered to do this in https://github.com/cardano-foundation/CIPs/pull/441#issuecomment-1455380740 "as soon as a preliminary review is made" ... we have that in https://github.com/cardano-foundation/CIPs/pull/441#pullrequestreview-1328235919 and though some of @michaelpj's points have been might have been answered, many of these comments haven't been answered and I personally can't tell whether they've all been answered in the document.

The document actually has some CPS sections in it like _Use Cases_ so even editorially it needs to be "disentangled" between problem statement & the actual proposal for improvement.  Could you proceed by decomposing this into a separate CIP and CPS (with another PR for the separated material) that takes into account @michaelpj's last review?  

We have an editor's meeting coming up later tomorrow (https://hackmd.io/@cip-editors/82) and if the documents are ready, the agenda seems like so we should have time to discuss it there.  But as generally agreed there is no rush so feel free to take more time & we would make space for it at the next meeting.

Ryun1:
Hey @mgajda!

Thanks for this pull request
this was reviewed in CIP editors call # 82
In order for this proposal to progress we will require that this PR is split into two, one for the CPS and one for the CIP.
Furthermore, please do adhere to the template provided ([here](https://github.com/cardano-foundation/CIPs/tree/master/.github)) as well guidance outlined within [CIP-0001](https://github.com/cardano-foundation/CIPs/tree/master/CIP-0001).

rphair:
@mgajda are you still pursuing this?  Marking `Waiting for Author` — in the hope of keeping this moving forward — since an unmentionable political event this week, avoiding mention of either cryptocurrency or quantum computing, suggest that a crypto key apocalypse is coming in the near term 😉

rphair:
NIST progress posted in last month: [NIST Releases First 3 Finalized Post-Quantum Encryption Standards](https://www.nist.gov/news-events/news/2024/08/nist-releases-first-3-finalized-post-quantum-encryption-standards)

@mgajda after not hearing from you for a couple months, this submission might be considered `Abandoned`.  Even if it's closed that way (if continuing not to have any progress or updates from you, especially the consideration of this work as a CPS rather than a CIP due to the perceived time frame vs. current uncertainty), we would welcome you opening this up again or (you or other author) submitting a newer work once more is known about the subject.

PR #442, opened 2023-01-19
CIP-???? | Authenticated Web3 HTTP requests

jmagan:
([rendered CIP document in branch](https://github.com/jmagan/CIPs/blob/master/CIP-%3F%3F%3F%3F/README.md))

### DIFF
--- a/CIP-????/README.md
+++ b/CIP-????/README.md
@@ -1,0 +1,98 @@
+---
+CIP: ?
+Title: Authenticated Web3 HTTP requests
+Category: Tools
+Status: Proposed
+Authors:
+    - Juan Salvador Magán Valero <jmaganvalero@gmail.com>
+Implementors: []
+Discussions:
+    - https://github.com/cardano-foundation/cips/pulls/?
+Created: 2022-12-27
+License: CC-BY-4.0
+---
+
+# CIP-XXXX: Authenticated Web3 HTTP requests
+
+## Abstract
+
+The cardano wallets have the ability to sign arbitrary pieces of data as we can see in the [Message signing CIP-0008](/CIP-0008/README.md). All wallets implement the method ```api.signData(addr: Address, payload: Bytes): Promise<DataSignature>``` defined in [Cardano dApp-Wallet Web Bridge CIP-0030](/CIP-0030/README.md).
+
+An important use case for web3 sites is calling HTTP endpoints using authenticated requests signed by the private keys of the users' wallets. In this way, the servers can perform actions using onchain and offchain data. 
+
+
+## Motivation
+
+dApps generate arbritary payloads as byte arrays. These payloads are signed and included in the protected headers of the signatures. The wallets are responsible for showing the payload data to the user, who will proceed to sign or reject the payload. It's a common practice to encode a string or a JSON string but there isn't any standard for the way to construct and to show this data.
+
+The current implementations for web3 applications use static strings. This is dangerous because if a bad actor intercepts the signed message then it can be used in a replay attack by the bad actor. That's why it is very important to produce a dynamic payload rather a static string.
+
+Another problem with the current approach is how the wallets show the information contained in the payload. The payload is a enconded byte array and it could contain anything. If Alice want to call an endpoint and Bob has the ability to change the message before Alice gets it. Alice must be notified somehow that she is signing a potentially malicious payload. A simple hex-encoded representation of the payload isn't enough to ensure a safe interaction.
+
+## Specification
+
+This specification involves multiple parties. We split this specification in three parts: server processing, payload structure and wallet specification.
+
+### Requirement Levels 
+
+The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "SHOULD NOT", "RECOMMENDED", "NOT RECOMMENDED", "MAY", and "OPTIONAL" in this document are to be interpreted as described in [RFC 2119](https://tools.ietf.org/html/rfc2119).
+
+### Payload structure
+
+The payload MUST be encoded as a JSON string. JSON strings are semi-structured data that are human-readable. So that with a straightforward decoding into text, it will be understandable for the readers. This feature also allows the users to debug it in an easy manner, for example, with the browser debugger tools.
+
+The content of the payload will be included in the protected header of the COSESign1 signature, hence the content effects directly the behavior and security of the system. The payload MUST have the following fields: 
+
+1. The field ```url``` MUST contain the full path to the endpoint, where the payload will be processed.
+
+2. Sometimes, the endpoint ```url``` field is not enough to determine its purpose. The user should understand perfectly the objective of the payload which he or she is signing. That's why the payload MUST contain an ```action``` field with a descriptive text containing the purpose of the payload. 
+
+3. The payload MUST include a UNIX timestamp. The ```timestamp``` field is used as a nonce and a mark for the payload expiration. This field is very imporant in case the payload is compromised.
+
+Additional fields MAY be included in the payload. Depending on the process, it could be useful to include some aditional fields in the protected header of the signature. For example, the email information should be included in the protected header in a registration request. In that way, this payload can be used only to register that specific email and it can not be tampered. 
+
+```JSON
+{
+    "url": "http://example.com/signup",
+    "action": "Sign up",
+    "timestamp": 1673261248,
+    ["email": "email@example.com"] <- Optional
+}
+```
+
+### Wallet specification
+
+The wallets can improve the overall security implementing the folliwing guidelines. We RECOMMEND to show in a structured way the payload information for shake of clarity. This information should be well understanded by the users before the payload is singed.  
+
+The ```url``` field provides information about the hostname of the application. This hostname MUST be included in the wallet allow list. If a known domain A tries to sign a payload for an unknown domain B, you will be prompted with permission popup making more obvious the cross-domain interaction. When possible, the wallet SHOULD warn the user if a payload is for a different domain.
+
+The wallet SHOULD update the ```timestamp``` field to the current time just before the signature. This field ideally should match the moment just before the signature, thereby the server recieves the more updated payload possible. 
+
+### Server processing
+
+The server has ultimate responsibility of processing correctly the requests. We use the content to validate the payload. The request will be processed with the following steps:
+
+1. The server MUST check the action and the endpoint included in the request. Each route to an endpoint MUST have an associated action and a URL. The first step is to check that they match with the parameterized action.
+
+2. The server MUST check the expiration of the payload. The expiration SHOULD be enough to give time to the user to introduce the wallet password but it SHOULD NOT be too long, we RECOMMEND not more than 5 minutes.
+
+3. The server MUST validate the COSESign1 signature and check that the address inside the protected map of the signature corresponds to the public key in the COSEKey. 
+
+Additionally the server COULD extract the payload content and pass it through the server logic.
+
+## Rationale: how does this CIP achieve its goals?
+
+This specification provides the general guidelines and necessary recommendations for performing secure authenticated web3 requests in the Cardano ecosystem. It covers the two main desired characteristics for a secure payload: It must expire and it must be non-static.
+
+Another important aspect for security is how wallets process the payload. They can improve the security using the data inside the payload to warn the users about possible malicious interactions. This specification emphasizes the importance of informing users clearly about the purpose of the payloads and how wallets can use the URL field to apply allow-lists and/or cross-domain policies.
+
+Finally, it establishes the requirements and recommendations for server side processing. The server must also ensure the validity of the signature and the payload, as well as of its purpose in order to accomplish the authentication. 
+
+## Reference implementation
+
+* [jmagan/passport-cardano-web3](https://github.com/jmagan/passport-cardano-web3)
+* [jmagan/cardano-express-web3-skeleton](https://github.com/jmagan/cardano-express-web3-skeleton)
+
+## Copyright
+
+This CIP is licensed under [CC-BY-4.0]: https://creativecommons.org/licenses/by/4.0/legalcode


CIP-????/README.md
```diff
@@ -0,0 +1,98 @@
...
+    - https://github.com/cardano-foundation/cips/pulls/?
+Created: 2022-12-27
+License: CC-BY-4.0
+---
+
+# CIP-XXXX: Authenticated Web3 HTTP requests
+
+## Abstract
+
+The cardano wallets have the ability to sign arbitrary pieces of data as we can see in the [Message signing CIP-0008](/CIP-0008/README.md). All wallets implement the method ```api.signData(addr: Address, payload: Bytes): Promise<DataSignature>``` defined in [Cardano dApp-Wallet Web Bridge CIP-0030](/CIP-0030/README.md).
```
matiwinnetou, line 19: Small detail: it is actually CIP-8 combined with CIP-30.

CIP-????/README.md
```diff
@@ -0,0 +1,98 @@
...
+    "url": "http://example.com/signup",
+    "action": "Sign up",
+    "timestamp": 1673261248,
+    ["email": "email@example.com"] <- Optional
+}
+```
+
+### Wallet specification
+
+The wallets can improve the overall security implementing the folliwing guidelines. We RECOMMEND to show in a structured way the payload information for shake of clarity. This information should be well understanded by the users before the payload is singed.
```
matiwinnetou, line 65: folliwing -> following

CIP-????/README.md
```diff
@@ -0,0 +1,98 @@
...
+3. The payload MUST include a UNIX timestamp. The ```timestamp``` field is used as a nonce and a mark for the payload expiration. This field is very imporant in case the payload is compromised.
+
+Additional fields MAY be included in the payload. Depending on the process, it could be useful to include some aditional fields in the protected header of the signature. For example, the email information should be included in the protected header in a registration request. In that way, this payload can be used only to register that specific email and it can not be tampered. 
+
+```JSON
+{
+    "url": "http://example.com/signup",
+    "action": "Sign up",
+    "timestamp": 1673261248,
+    ["email": "email@example.com"] <- Optional
```
matiwinnetou, line 59: From GDPR (europe wise) point of view even putting email here as optional is a total disaster...

CIP-????/README.md
```diff
@@ -0,0 +1,98 @@
...
+
+Additionally the server COULD extract the payload content and pass it through the server logic.
+
+## Rationale: how does this CIP achieve its goals?
+
+This specification provides the general guidelines and necessary recommendations for performing secure authenticated web3 requests in the Cardano ecosystem. It covers the two main desired characteristics for a secure payload: It must expire and it must be non-static.
+
+Another important aspect for security is how wallets process the payload. They can improve the security using the data inside the payload to warn the users about possible malicious interactions. This specification emphasizes the importance of informing users clearly about the purpose of the payloads and how wallets can use the URL field to apply allow-lists and/or cross-domain policies.
+
+Finally, it establishes the requirements and recommendations for server side processing. The server must also ensure the validity of the signature and the payload, as well as of its purpose in order to accomplish the authentication.
```
matiwinnetou, line 89: Would a user have to sign a transaction for every http request or just like JWT there would be a way to cache and store token somewhere and use this until it expires. 

If this is each time then it could be very annoying, usually a webapp makes many requests to the server side, even for one page.

jmagan: Exactly. This specification isn't for each request. You can use it only for the important requests like Login, Register, etc... Afterwards a JWT token can take control of the session.

I developed an example using this methodology [cardano-express-web3-skeleton)](https://github.com/jmagan/cardano-express-web3-skeleton)

CIP-????/README.md
jmagan: Maybe it's a bad example. I don't know much about the GDPR. Could you please tell me more about why including the email in the payload can be a problem for the GDPR?

CIP-????/README.md
jmagan: Resolve, thanks!

CIP-????/README.md
rphair: Once the email address is in there, it's personally identifiable data that cannot be removed.  According to my understanding, even if an email address is added optionally it also has to be removable later.

- Technical recommendations (with lots of references): https://www.nature.com/articles/s41598-022-19341-y
- Bureaucratic view from EU itself & recommendations: https://www.europarl.europa.eu/RegData/etudes/STUD/2019/634445/EPRS_STU(2019)634445_EN.pdf

rphair: Once the email address is in there, it's personally identifiable data that cannot be removed.  According to my understanding, even if an email address is added optionally it also has to be removable later: even if you have to destroy the whole blockchain to efface it.

- Technical recommendations (with lots of references): https://www.nature.com/articles/s41598-022-19341-y
- Bureaucratic view from EU itself & recommendations: https://www.europarl.europa.eu/RegData/etudes/STUD/2019/634445/EPRS_STU(2019)634445_EN.pdf

CIP-????/README.md
```diff
@@ -0,0 +1,98 @@
...
+
+## Rationale: how does this CIP achieve its goals?
+
+This specification provides the general guidelines and necessary recommendations for performing secure authenticated web3 requests in the Cardano ecosystem. It covers the two main desired characteristics for a secure payload: It must expire and it must be non-static.
+
+Another important aspect for security is how wallets process the payload. They can improve the security using the data inside the payload to warn the users about possible malicious interactions. This specification emphasizes the importance of informing users clearly about the purpose of the payloads and how wallets can use the URL field to apply allow-lists and/or cross-domain policies.
+
+Finally, it establishes the requirements and recommendations for server side processing. The server must also ensure the validity of the signature and the payload, as well as of its purpose in order to accomplish the authentication. 
+
+## Reference implementation
```
rphair, line 91:
```suggestion
### Reference implementation
```
There isn't a main section heading for this () which could be fixed by demoting it to be part of the Rationale.

CIP-????/README.md
```diff
@@ -0,0 +1,98 @@
...
+
+* [jmagan/passport-cardano-web3](https://github.com/jmagan/passport-cardano-web3)
+* [jmagan/cardano-express-web3-skeleton](https://github.com/jmagan/cardano-express-web3-skeleton)
+
```
rphair, line 95: We also need a _Path to Active_ section with acceptance & implementation sub-sections: https://github.com/cardano-foundation/CIPs/tree/master/CIP-0001#statuses

There's a candidate proposal in a similar field with good checklists for these that might be helpful as a model (pending CIP-0045 if this gets merged & branch disappears): https://github.com/fabianbormann/CIPs/blob/master/CIP-%3F%3F%3F%3F/README.md#path-to-active

rphair (commented):
This is a sensible and useful proposal (thanks @matiwinnetou @jmagan for the proofreading):

CIP-????/README.md
jmagan: Thanks for the reply. Now I see your point but the payload is signed in the client and sent to the server. It isn't stored in the blockchain. So, the GDPR doesn't apply in this case. Right?

CIP-????/README.md
jmagan: I indent it as you propused. Thanks

CIP-????/README.md
jmagan: Thanks for the reference. It was very useful. I wrote the Path to Active.

CIP-????/README.md
rphair: > @jmagan: It isn't stored in the blockchain
As far as I'm concerned that takes care of it.  Any other situations where personal data might be exposed off-chain can be dealt with later if they're a problem. 😎

jmagan: Great. Then can I mark as resolve?

CIP-????/README.md
jmagan: Is good for you if I mark this as resolve?

CIP-????/README.md
rphair: Yes (I'm marking it) 😎

jmagan:
UPDATE: I added a new implementation with Next JS and Auth JS. 

[cardano-nextjs-web3-skeleton](https://github.com/jmagan/cardano-nextjs-web3-skeleton)

CIP-????/README.md
```diff
@@ -0,0 +1,119 @@
...
+The payload MUST be encoded as a JSON string. JSON strings are semi-structured data that are human-readable. So that with a straightforward decoding into text, it will be understandable for the readers. This feature also allows the users to debug it in an easy manner, for example, with the browser debugger tools.
+
+The content of the payload will be included in the protected header of the COSESign1 signature, hence the content effects directly the behavior and security of the system. The payload MUST have the following fields: 
+
+1. The field ```url``` MUST contain the full path to the endpoint, where the payload will be processed.
+
+2. Sometimes, the endpoint ```url``` field is not enough to determine its purpose. The user should understand perfectly the objective of the payload which he or she is signing. That's why the payload MUST contain an ```action``` field with a descriptive text containing the purpose of the payload. 
+
+3. The payload MUST include a UNIX timestamp. The ```timestamp``` field is used as a nonce and a mark for the payload expiration. This field is very imporant in case the payload is compromised.
```
KtorZ, line 50: How could the payload be "compromised" here? What is the assumed threat model?

CIP-????/README.md
```diff
@@ -0,0 +1,119 @@
...
+    "url": "http://example.com/signup",
+    "action": "Sign up",
+    "timestamp": 1673261248,
+    ["email": "email@example.com"] <- Optional
+}
+```
+
+### Wallet specification
+
+The wallets can improve the overall security implementing the following guidelines. We RECOMMEND to show in a structured way the payload information for shake of clarity. This information should be well understanded by the users before the payload is signed.
```
KtorZ, line 65:
```suggestion
The wallets can improve the overall security implementing the following guidelines. We RECOMMEND to show in a structured way the payload information for sake of clarity. This information should be well understood by the users before the payload is signed.  
```

CIP-????/README.md
```diff
@@ -0,0 +1,119 @@
...
+
+The current implementations for web3 applications use static strings. This is dangerous because if a bad actor intercepts the signed message then it can be used in a replay attack by the bad actor. That's why it is very important to produce a dynamic payload rather a static string.
+
+Another problem with the current approach is how the wallets show the information contained in the payload. The payload is a enconded byte array and it could contain anything. If Alice want to call an endpoint and Bob has the ability to change the message before Alice gets it. Alice must be notified somehow that she is signing a potentially malicious payload. A simple hex-encoded representation of the payload isn't enough to ensure a safe interaction.
+
+## Specification
+
+This specification involves multiple parties. We split this specification in three parts: server processing, payload structure and wallet specification.
```
KtorZ, line 34: It isn't clear to me what "the server" refers to in this proposal. I think the proposal would deserve a bit of an introduction regarding each actor involved and their role.

CIP-????/README.md
```diff
@@ -0,0 +1,119 @@
...
+
+The ```url``` field provides information about the hostname of the application. This hostname MUST be included in the wallet allow list. If a known domain A tries to sign a payload for an unknown domain B, you will be prompted with permission popup making more obvious the cross-domain interaction. When possible, the wallet SHOULD warn the user if a payload is for a different domain.
+
+The wallet SHOULD update the ```timestamp``` field to the current time just before the signature. This field ideally should match the moment just before the signature, thereby the server recieves the more updated payload possible. 
+
+### Server processing
+
+The server has ultimate responsibility of processing correctly the requests. We use the content to validate the payload. The request will be processed with the following steps:
+
+1. The server MUST check the action and the endpoint included in the request. Each route to an endpoint MUST have an associated action and a URL. The first step is to check that they match with the parameterized action.
```
KtorZ, line 75: What request are we talking about here and, how does the server "check the action"? What is being sent to the server, the wallet etc..? 

Perhaps a little diagram (even a plain ASCII one) could help better show the interaction between the actors.

CIP-????/README.md
```diff
@@ -0,0 +1,119 @@
...
+
+2. The server MUST check the expiration of the payload. The expiration SHOULD be enough to give time to the user to introduce the wallet password but it SHOULD NOT be too long, we RECOMMEND not more than 5 minutes.
+
+3. The server MUST validate the COSESign1 signature and check that the address inside the protected map of the signature corresponds to the public key in the COSEKey. 
+
+Additionally the server COULD extract the payload content and pass it through the server logic.
+
+## Rationale: how does this CIP achieve its goals?
```
KtorZ, line 83: One question I'd like to see answered in the 'Rationale' is what value does such an authorization has from a service provider standpoint (assumed to be 'the server' here)? Said differently, why would my-awesome-dapp care that a user authorized something with its wallet? Is the service provider assumed to have some knowledge of the key involved in the authorization and is only expecting a proof of ownership? How is such a proof better than a transaction witness proving the ability to spend associated UTxOs?

CIP-????/README.md
jmagan: You can call the endpoint, for example "http://my-awesome-dapp/users", but this endpoint can create or delete users. If someone gets the signed payload, he or she can access to both functions. This action field mitigates risk.

CIP-????/README.md
```diff
@@ -0,0 +1,119 @@
...
+License: CC-BY-4.0
+---
+
+# CIP-XXXX: Authenticated Web3 HTTP requests
+
+## Abstract
+
+The cardano wallets have the ability to sign arbitrary pieces of data as we can see in the [Message signing CIP-0008](/CIP-0008/README.md). All wallets implement the method ```api.signData(addr: Address, payload: Bytes): Promise<DataSignature>``` defined in [Cardano dApp-Wallet Web Bridge CIP-0030](/CIP-0030/README.md).
+
+An important use case for web3 sites is calling HTTP endpoints using authenticated requests signed by the private keys of the users' wallets. In this way, the servers can perform actions using onchain and offchain data.
```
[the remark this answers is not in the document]
jmagan: When I say the server, I mean the own dApp application server. I pursuit real decentralization, only the App server and the Blockchain is needed for this Authentication model. It can be also  used for authorization using NFTs.

I know that lately almost all Web3 Blockchain authorization and authentication models use external entities as identity authorities. But this CIP isn't about it.

I want to provide a simple and effective method for authentication without additional actors.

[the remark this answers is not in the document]
jmagan: @KtorZ I rewrote the abstract. Sorry for the delay. Could you please check it?

CIP-????/README.md
jmagan: @KtorZ I've already created an ASCII diagram. I hope it clarifies the process.

CIP-????/README.md
jmagan: I added dApp Server and the ASCII diagram. I believe now it's clearer.

CIP-????/README.md
jmagan: @KtorZ, I made some improvement in the Rationale. I think now it's a little bit more specific.

CIP-????/README.md
```diff
@@ -0,0 +1,147 @@
...
+3. The payload MUST include a UNIX timestamp. The `timestamp` field is used as a nonce and a mark for the payload expiration. This field is very important in case the payload is compromised.
+
+Additional fields MAY be included in the payload. Depending on the process, it could be useful to include some aditional fields in the protected header of the signature. For example, the email information should be included in the protected header in a registration request. In that way, this payload can be used only to register that specific email and it can not be tampered. 
+
+```JSON
+{
+    "uri": "http://example.com/signup",
+    "action": "Sign up",
+    "timestamp": 1673261248,
+    ["email": "email@example.com"] <- Optional
```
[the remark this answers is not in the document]
jmagan: @matiwinnetou Thank you for your suggestion! Providing the JSON schema instead of an invalid JSON is indeed a great idea.

jmagan:
I have changed the url property to uri as I believe uri is a more general term.

jmagan:
I have changed the url property to uri as I believe uri is a more general term.

CIP-????/README.md
rphair: @jmagan this is understandable, but still not explained in the CIP as such: can you add a little paragraph repeating or explaining what you've just said?  cc @KtorZ

CIP-????/README.md
rphair: @jmagan I can see you've added at the beginning of this section (https://github.com/jmagan/CIPs/blob/master/CIP-%3F%3F%3F%3F/README.md#specification) so @KtorZ marking this as resolved.

CIP-????/README.md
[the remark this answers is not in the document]
rphair: Disambiguated as per https://github.com/cardano-foundation/CIPs/pull/442#discussion_r1182508929 and @jmagan I can see how the Abstract now includes some functional description.  It's still a bit vague and redundant, so could be improved... will leave it to @KtorZ about whether it's currently acceptable.  I would like to see one clear sentence about what the proposed component is actually doing, with less general information about Cardano and CIPs.

CIP-????/README.md
rphair: @jmagan what's currently in the Rationale is more specific, with respect to currently available alternatives that are trying to do the same thing, but it doesn't go into as much detail as @KtorZ is requesting above.  If it answers the question about "How does the dApp trust the wallet in this method?" then can you please explain it here and/or in the CIP?  (@KtorZ please mark Resolved if you think it is)

rphair (commented):
@jmagan thanks for the updates.  I'm basing my own review, to get ready to promote this to a candidate, on @KtorZ's last review (https://github.com/cardano-foundation/CIPs/pull/442#pullrequestreview-1371060181) so my comments are above in that review.  Most issues have been progressed but a few are still unresolved in my opinion.

I think this is currently ready to be marked as a candidate because the pending review items deal with the explanations in the CIP document rather than the soundness of the proposal itself.

So I'd be ready to green-check this as soon as the nitpick below & items still marked unresolved in https://github.com/cardano-foundation/CIPs/pull/442#pullrequestreview-1371060181 are addressed and a candidate CIP number is assigned.

CIP-????/README.md
jmagan: @rphair  Okey I see the point. I though that this point would clarify it: 

>  Moreover, the signature method proposed in this CIP does not require users to spend funds in a transaction, which further lowers the cost and barriers to entry for users.

In relation to your question 

>How does the dApp trust the wallet in this method?

I can add a brief summary of or a referenfe to the CIP-0008 but I though It could be a little bit redundant.

rphair: I also thought that sentence was relevant to the question: so maybe a bit more detail, maybe even a subheading on that subject, would make the advantage more apparent to service providers... since generally they're the ones who will be implementing your proposal 🤓

CIP-????/README.md
jmagan: @KtorZ I added an example and how it's mitigated this threat model as commented by @rphair 

> 2. Sometimes, the endpoint `uri` field is not enough to determine its purpose. The user should understand perfectly the objective of the payload which he or she is signing. That's why the payload MUST contain an `action` field with a descriptive text containing the purpose of the payload. For example, if someone calls the endpoint `/users` without an action field, they can create or delete users. However, by including the action field in the payload, it not only provides additional clarity but also effectively limits the scope of the payload.

CIP-????/README.md
jmagan: @rphair I added some definition before the ASCII diagram:

> This specification involves multiple parties: Wallet/Client, dApp Server and Blockchain. 
> 
> 1. **Wallet/Client**: The Wallet/Client is responsible for managing the user's cryptographic keys. Anyone can create a wallet using the CIP-0030 API interface, but it may produce invalid or malicious data sent to the dApp. This CIP aims to validate the ownership and veracity of the data provided by wallets. Additionally, it establishes guidelines for mitigating common wallet attacks, improving security for user interaction.
> 
> 2. **dApp Server**: The dApp Server represents the server-side infraestructure that supports decentralized applications (dApps). It communicates with the blockchain to retrieve stored data and validate wallet status. It must enforce minimum payload requirements to ensure authenticity and protect users from malicious actors.
> 
> 3. **Blockchain**: The Blockchain is the underlying distributed ledger technology that forms the foundation of decentralized systems. It is a decentralized and immutable ledger that securely records all transactions and data in a chronological and transparent manner. The Blockchain can be utilized for authentication, providing user identity, and for authorization, tracking user history and current status.

I believe this gives more background and context to each actor.

CIP-????/README.md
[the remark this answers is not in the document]
jmagan: I was thinking about how to improve the Abstract part but I'm clueless. I can't figure out how to write it without referencing Cardano. Shouldn't it be general and easy to understand?

CIP-????/README.md
jmagan: I made a connection between CIP-0008 and this CIP at the beggining of the Rationale. I feel that previously this part was missing and now all make more sense.

> ## Rationale: how does this CIP achieve its goals?
> 
> CIP-0008 enhances authentication by enabling individuals to prove ownership of addresses, identities, and off-chain data through message signing. It provides a reliable means of authentication by allowing individuals to attach a public key to their data and sign messages associated with that data, thereby establishing ownership and ensuring the integrity of the authentication process.
> 
> Additionally, This specification provides the general guidelines and necessary recommendations for performing secure authenticated web3 requests in the Cardano ecosystem. It covers the two main desired characteristics for a secure payload: It must expire and it must be non-static. Moreover, the signature method proposed in this CIP does not require users to spend funds in a transaction, which further lowers the cost and barriers to entry for users.

CIP-????/README.md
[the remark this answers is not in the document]
jmagan: @matiwinnetou, great idea. It's already implmented:

> #### JSON Schema
> ```JSON
> {
>   "$schema": "http://json-schema.org/draft-07/schema#",
>   "type": "object",
>   "properties": {
>     "uri": {
>       "type": "string",
>       "format": "uri"
>     },
>     "action": {
>       "type": "string"
>     },
>     "timestamp": {
>       "type": "integer"
>     }
>   },
>   "required": ["uri", "action", "timestamp"],
>   "additionalProperties": {
>     "type": "string"
>   }
> }
> ```
> #### Minimum payload:
> 
> ```JSON
> {
>     "uri": "http://example.com/signin",
>     "action": "Sign in",
>     "timestamp": 1673261248,
> }
> ```
> #### Sign up payload example:
> ```JSON
> {
>     "uri": "http://example.com/signup",
>     "action": "Sign up",
>     "timestamp": 1673261248,
>     "email": "email@example.com"
> }
> 
> ```

jmagan:
@rphair I did my best in order to improve the quality of this CIP. I hope the changes are good enough. We could talk to include it in a CIP Editor Meeting when you consider it done. Thanks for your time and your valuable feedback. 🙂

CIP-????/README.md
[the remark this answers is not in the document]
rphair: no @jmagan by all means I wasn't suggesting you not mention Cardano: just that it doesn't have to read like an article for an open audience.  If you want to set a more conversational tone, that's up to you.

rphair (approved):
@jmagan thanks, I'm marking the issues above resolved because I think they've been addressed; @KtorZ @matiwinnetou please re-open if you think otherwise:

https://github.com/cardano-foundation/CIPs/pull/442#discussion_r1157306100
https://github.com/cardano-foundation/CIPs/pull/442#discussion_r1157323830
https://github.com/cardano-foundation/CIPs/pull/442#discussion_r1157331430
https://github.com/cardano-foundation/CIPs/pull/442#discussion_r1157335373
https://github.com/cardano-foundation/CIPs/pull/442#discussion_r1179238670

I've suggested it for our Review agenda for the meeting # 66 in 6 days to consider promotion to a CIP candidate.

rphair:
@jmagan this is up for Review on our CIP meeting agenda in 20 minutes... hope you can participate; will post feedback here in case you can't make it.

rphair:
Consensus at CIP meeting today is to promote this to a CIP candidate & assign a CIP number (cc @KtorZ).

CIP-????/README.md
```diff
@@ -0,0 +1,187 @@
...
+### Implementation Plan
+
+- [X] Create a library for processing payload according to this specification.
+- [ ] Open a conversation about this specification and its possible improvements.
+- [ ] Talk about further web3 standards and new specifications. 
+- [ ] Write the documentation for web3 developers.
+
+## Copyright
+
+This CIP is licensed under [CC-BY-4.0]: https://creativecommons.org/licenses/by/4.0/legalcode
```
Ryun1, line 187:
```suggestion
This CIP is licensed under [CC-BY-4.0](https://creativecommons.org/licenses/by/4.0/legalcode).
```
Small formatting clean up

Ryun1 (commented):
I think this is a well written and interesting proposal, I would be keen for this to see adoption.

I have two larger outstanding questions:
- Could (should?) this functionality be added to a CIP-30 extension?
- We often see issues around versioning of CIPs (or lack there of) how should this proposal handle future upgrades/expansions?

rphair:
> I don't have any strong opinion about that. Are there any editor's preferences regarding this topic? Or could you provide me with an example of another versioned CIP?

There's some history on the CIP Editors' Meeting Discord about whether CIP versions themselves should be part of the CIP framework.  The repeated resolution is that they should not.  (The question of how to handle API versioning for CIP-0030 and @Ryun1 may be the best resource about that.)

Today on Discord we (so far just me) reiterated that CIP authors are free to handle versioning as they see fit & here is the best example I can think of of this being done: https://github.com/cardano-foundation/CIPs/pull/502

jmagan:
Hi @rphair @Ryun1 ,

I have added a version history table at the end of the file, and I have created a directory containing the JSON schema for version 1. This directory will serve as an archive for all future versions. With these updates in place, I hope everything is now ready to obtain the CIP number.

rphair:
Versioning looks good to me.  FYI @KtorZ @Ryun1 I'm assigning the pending CIP numbers to this & #499 now.

CIP-????/README.md
```diff
@@ -0,0 +1,194 @@
...
+```JSON
+{
+    "uri": "http://example.com/signup",
+    "action": "Sign up",
+    "timestamp": 1673261248,
+    "email": "email@example.com"
+}
+
+```
+
```
stevenj, line 129: I have a concern with this approach.
It's using the pre-existing `api.signData` method, and these fields are pretty generic and not very identifying.  
Below it is defined for the wallet to unambiguously determine `It MUST detect when the payload is formatted using this specification.`.
How does the wallet do this, and how does the wallet/dApp guard against false positive/negative detection?

Eg, What should the wallet do if it sees `"uri"` and `"action"` but not `"timestamp"`?
1. Assume it is an Authenticated WEB3 HTTP Request that is in error? ; OR
2. Assume that it is NOT an Authenticated WEB3 HTTP Request?

And what should the wallet do in either case?

CIP-????/README.md
```diff
@@ -0,0 +1,194 @@
...
+### Acceptance Criteria
+
+- [X] At least one library should implement this authentication method.
+- [ ] The 80% users should have wallets implementing the following requirements:
+    1. It MUST detect when the payload is formatted using this specification.
+    2. The information contained in the payload MUST be parsed and formatted in the signing pop-up.
+    3. The wallet SHOULD update the timestamp just before the payload is signed.
+    4. The wallet MUST detect if the URI is in the allow list.
+    5. The wallet SHOULD warn the user against cross-domain requests.
```
stevenj, line 182: These act as an update to CIP-30.  A wallet that does this is not following the specification defined for CIP-30 for signData which does not include these rules for payload data processing.  There would be no way for a dApp to know if a wallet obeyed CIP-30 or CIP-93 because CIP-93 is not currently defined as an extension of CIP-30.

CIP-????/README.md
jmagan: Hi @stevenj! Apologies for the delay. This functionality seamlessly integrates with the existing implementation. To detect the payload, we simply retrieve a JSON object and check for the presence of the mandatory fields: "uri", "action", and "timestamp," as you mentioned earlier. It's important to note that it should not throw an error because CIP-0030 allows signing arbitrary payloads. The crucial aspect here is that the wallet effectively communicates to the user the specific endpoint and corresponding action associated with the request.

CIP-????/README.md
jmagan: I partially addressed this in the previous comment. The proposed changes do not modify the API, and the signData method retains the same arguments and return value. This CIP involves three key parties: the wallet, the payload, and the server processing. It would be challenging to cover all these aspects within a wallet-specific CIP alone, which is why it is presented as a comprehensive specification covering the entire process.

jmagan:
Hey @matiwinnetou, I've incorporated the suggested changes:

- Added an `actionText` field to enhance globalization capabilities.
- Included the option of including either the timestamp or slot field for time.
- Mentioned that additional properties can also be objects.
- Added a new section emphasizing the common usage of JWT tokens for efficient session management.

I hope this covers all the modifications you requested! Let me know if there's anything else I can assist you with.

CIP-????/README.md
rphair: @stevenj this looks resolved based on last comment but feel free to re-open (we've moving this forward & would like to know if any unresolved issues)

CIP-????/README.md
rphair: likewise resolved as per https://github.com/cardano-foundation/CIPs/pull/442#discussion_r1235142130.

rphair:
@jmagan thanks for attending today's CIP editors' meeting and allowing us to move this forward to `Last Check` in our meeting scheduled 2 weeks from today.  In the meantime:

There was a suggestion that the Cardano Foundation (mentioning @matiwinnetou) is using a part of this mechanism in an application, but we agreed today not to require it in _Path to Active_ which only requires implementation in wallets.  If it should be merged as `Active` on the basis of the CF implementation then please adjust the `Path to Active` accordingly and change the status to `Active` in the document header.

jmagan:
Hello @AndrewWestberg, I'm huge fan of your videos. Actually, I was talking about this today in the CIP Editor Meetings channel on Discord. The CIP-8 already approachs this issue and the wallet can pass to the hardware wallet a hash of the unprotected headers:

> #### Payload encoding
> 
> To solve `E3`, `signed_message` body header MUST contain `hashed: bool` as an `unprotected` header which defines whether or not we signed the `payload` OR the `Blake2b224` hash of the `payload`. The hash MUST be used in the following two cases
> 
> 1) The size of the raw `payload` would otherwise be too big to fit in hardware wallet memory (see E1). Note that the exact size for which this is the case depends on the device.
> 1) The payload characters (ex: non-ASCII) that cannot be displayed on the hardware wallet device (see E3)
> 
> We RECOMMEND showing the user the full payload on the device if possible because it lowers the attack surface (otherwise the user has to trust that the hash of the payload was calculated correctly).
> 
> `Blake2b224` was chosen specifically because `224` bits is already a long string for hardware wallets.

Maybe, some CIP-30 modification is necessary in order to enforce the hash version of the payload.

rphair:
@jmagan thanks for discussing this at today's CIP Editors' meeting.  My understanding of the reservations expressed is that we should take off the `Last Check` status until the questions are answered about
- whether CIP8 headers should be used instead
- whether alternatives to JSON (particularly CBOR) should be used in the specification.

I apologise for the rough description & if you can to go into detail about those reservations for the record here then please do so.

We also noted that the change to CIP8 would be nontrivial and time-consuming relative to the work you can put into this CIP.  Therefore I'm taking off the `Last Check` label & if you think it will take multiple meeting intervals to progress these issues then please say so and we'll mark this `Waiting for Author`.

jmagan:
Hi @rphair, thank you for your comments. I'm planning to address the reasons for not using CIP-8 in the CIP's rationale. Essentially, I will explain that adopting that approach would necessitate modifying both CIP-8 and CIP-30, as well as waiting for corresponding wallet implementations. In contrast, the current implementation is ready to be deployed, allowing us to provide immediate value to the Cardano Web3 ecosystem. However, this doesn't mean that this CIP is intended as a temporary solution while those CIPs are being modified, potentially leading to their deprecation or use as legacy in the future.

Regarding the choice of JSON over CBOR, the CIP already discusses this matter. The requirement for the string to be human-readable is the primary factor, which CBOR does not fulfill. Additionally, JSON is more user-friendly for web developers, who are the target users of this CIP.

rphair:
@jmagan I am happy to hear that the suggestions about CIP-8 don't need to block this proposal from being merged.  Your documenting that part of the rationale will help people decide whether to follow your suggested implementation or to use CIP-8 instead and perhaps write their own "competitive" CIP for it.

From my own experience I agree with your point about the human readability being an important factor for web implementations... similar decisions were made to favour human readable formats in our NFT related CIPs.

Given your last response I think it's appropriate to leave the `Last Check` label on this CIP because once you've documented the CIP-8 alternative then we'll have addressed the last of the comments that were brought up in yesterday's meeting (cc @Ryun1 @KtorZ @SebastienGllmt).

jmagan:
Hi @rphair. I had a discussion this morning with @matiwinnetou, and we concluded that the current version of this CIP holds value and can be implemented as is. We also carefully considered the concerns raised by @SebastienGllmt, and we agree that making this CIP idiomatic with Cardano using CBOR would be ideal. Our proposed solution is to proceed with this CIP while simultaneously working on modifying CIP-8 and CIP-30 to define the Payload format as either JSON or CBOR. This approach would allow sufficient time for browser developers to meet the requirements for human readability and make necessary adjustments to the wallet API. Consequently, we can introduce a version 2 of this CIP incorporating COSESign and CBOR, accommodating both realms and ensuring broad support.

jmagan:
Hi @rphair, I have made the requested modifications. I have added an extension to the rationale section to address the CBOR and JSON dilemma. Additionally, I have included a suggestion from @matiwinnetou to allow numeric strings in the timestamp and slot fields.

Extended rationale:

> ### Alternative implementations
> 
> During discussions about this specification, the possibility of modifying CIP-0008 to incorporate the standards defined here was considered. However, a thorough evaluation revealed that this approach would require extensive modifications to CIP-0008 and CIP-0030, leading to significant changes in the Cardano wallet API. Moreover, it would result in a lengthy waiting period for browser wallet developers to implement the necessary requirements. This could potentially bypass important security measures outlined in this CIP, such as the requirement for human readability.
> 
> While this alternative approach brings advantages, such as defining the payload in CBOR, which aligns well with Cardano, it also presents challenges. JSON and CBOR offer different levels of expressiveness, and the choice between the two depends on the specific needs of the application. JSON provides a more flexible and widely supported data format, whereas CBOR offers a more compact and efficient representation, particularly beneficial when working with the blockchain.
> 
> Considering these factors, it was concluded that deploying this standard as it currently stands, while coexisting with a future version that allows users to choose between JSON and CBOR payloads, would be the most practical approach. This would provide sufficient time for modifying CIP-0008 and CIP-0030, enabling browser developers to fulfill the requirements for human readability and make necessary adjustments to the wallet API. Consequently, a version 2 of this CIP can be introduced, incorporating COSESign and CBOR, accommodating both realms, and ensuring broad support.

New JSON Schema:

```JSON
{
  "$schema": "http://json-schema.org/draft-07/schema#",
  "type": "object",
  "properties": {
    "uri": {
      "type": "string",
      "format": "uri"
    },
    "action": {
      "type": "string"
    },
    "actionText": {
      "type": "string"
    },
    "timestamp": {
      "anyOf": [
        {
          "type": "integer"
        },
        { "type": "string", "pattern": "^\\d+$" }
      ]
    },
    "slot": {
      "anyOf": [
        {
          "type": "integer"
        },
        { "type": "string", "pattern": "^\\d+$" }
      ]
    }
  },
  "required": ["uri", "action"],
  "oneOf": [{ "required": ["timestamp"] }, { "required": ["slot"] }],
  "additionalProperties": {
    "type": ["string", "object"]
  }
}
```

rphair:
@jmagan thanks for taking care of this & I believe the new _Alternative implementations_ section satisfies the conversation that came up in this week's CIP meeting.

It also seems logical to me that you've also ticked off the two conversation & review items under "Implementation plan" because we've now done this extensively on GitHub and the meetings.

So I guess this is cleanly back to `Last Check` again scheduled for merge at the next meeting and will ping the other reviewers to confirm & to see if there are any other issues before then.

Ryun1 (approved):
Well written and discussed proposal, good work @jmagan!  🚀

rphair:
OK @KtorZ I think the proposer has done a great job adapting this proposal to challenges whenever necessary and so it's ready to hit the `merge` button at the CIP meeting today unless you or anyone express any reservations there (or here). 🚀

jmagan:
Thank you, @rphair and @Ryun1, for your kind words and the invaluable support provided throughout this journey. Your contributions have been greatly appreciated.

gitmachtl:
@jmagan is it possible that the libs you provide for cip93 reference implementation do not support hashed payload content?

jmagan:
Hi @gitmachtl, yes. I'll take a look to it. It's an important feature to be implemented. Can you give me further details about the packages you are using or some way to reproduce the error?

gitmachtl:
@jmagan turnes out that the used https://github.com/cardano-foundation/cardano-verify-datasignature functions did not support the `{ "hashed": true/false }` header to verify the correct response. but there are already PRs going on to fix it.

gitmachtl:
i have implemented the CIP8/30 specs pretty early on in https://github.com/gitmachtl/cardano-signer and therefore i went with the full specs for it. from time to time i stumble across implementations that do now have the information about a hashed payload in it. but its clearly stated in the CIP here:
https://github.com/cardano-foundation/CIPs/blob/f389cda520d578b10967cbbb7979358d38e030ff/CIP-0008/README.md?plain=1#L255

jmagan:
@gitmachtl I've spoken with @matiwinnetou  and the problem was addressed and there are multiple PRs almost ready to merge. It will be fixed soon.

PR #453, opened 2023-01-26
Rework CIP-0005's headers and sections + remove h1 headers from various CIPs

### LINKED ISSUE #451: CIP-0005 | Adjust preamble and structure w.r.t CIP-0001

KtorZ:
Fixes #451


rphair (approved):
@KtorZ can you add or change something in the PR title to say that it also affects 9 other CIPs besides CIP-0005?  We may want to find & reference this later e.g. if there are ongoing questions about the use of H1 tags.

PR #454, opened 2023-01-27
CPS-???? | Wallet UX

micahkendall:
[Rendered](https://github.com/micahkendall/CIPs/tree/master/CPS-%3F%3F%3F%3F)

## Abstract
Some Wallets have poor UX, causing headaches for users and developers. This CPS aims to improve the UX of wallets, particularly browser wallets implementing CIP-30. Security concerns should also be considered.

## Problem

e.g
- https://github.com/cardano-foundation/CIPs/pull/443
    - Txs must not be folded together into one, as that presents an attack vector.
    - Whether to present as many txs in a list, or sequentially, is a UX decision, affected by the number of txs and chaining.

- https://github.com/cardano-foundation/CIPs/pull/88
    - 'No collateral' warning should not be shown for a tx without smart contracts. @Quantumplation "I deal with literally 4 or 5 support tickets a day where people see that and get scared"

## Use cases
Wallet Users will have a better user experience. Wallet providers will have an easier time building user interfaces.

## Goals
Create and enforce UX standards.

## Open Questions
- What are the UX standards?
- How do we enforce them?

### DIFF
--- a/CPS-????/README.md
+++ b/CPS-????/README.md
@@ -23,7 +23,7 @@ e.g
     - Whether to present as many 
 
 - https://github.com/cardano-foundation/CIPs/pull/88
-    - 'No collateral' warning should not be shown for a tx without smart contracts. Cc @Quantumplation "I deal with literally 4 or 5 support tickets a day where people see that and get scared"
+    - 'No collateral' warning should not be shown for a tx without smart contracts. @Quantumplation "I deal with literally 4 or 5 support tickets a day where people see that and get scared"
 
 
 ## Use cases
--- a/CPS-????/README.md
+++ b/CPS-????/README.md
@@ -1,0 +1,37 @@
+---
+CPS: ?
+Title: Wallet UX
+Status: Open
+Category: Wallets
+Authors:
+    - Micah Kendall <micahk.me@icloud.com>
+Proposed Solutions: []
+Discussions:
+    - https://github.com/cardano-foundation/cips/pulls/?
+Created: 2023-01-27
+---
+
+## Abstract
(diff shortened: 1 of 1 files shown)


rphair (commented):
dear @micahkendall I'm putting this into Draft stage because right now it's only a CPS template wrapped around some tersely suggested behavioural changes to existing CIP implementations (?).

It could be a "stub" for further detail which you or other contributors provide later, but even the current title is too vague to leave it open for others to know what they'd be contributing to.  The problem of "poor UX" is too general to mean anything to any developer or to create a well defined set of CIPs within it: which is the ultimate purpose of the CPS.

So if you want to go through with this, please start by giving it a meaningful title... e.g. if it's about transactions and collateral you could call it _Wallet transaction and collateral management_.

Then you must provide a lot more background about how this _particular_ problem presents in the current state of the art as the "Problem" section, and then some really specific "Use cases" with some options of what a better user experience would be like.

Finally please remember there is never going to be an "enforced" UX standard.  Not even CIPs can be enforced.  It's not constructive to say "We want to force wallet developers to do better" when you haven't completely categorised or defined the problem(s) yet.

It may be that other editors want to close this PR, but I would be up for leaving it open as long as steps are being made regularly toward establishing a well-defined CPS here in the document.  Please use existing merged & candidate CPSs to gauge the level of consideration & detail this would require and post here (or on the Cardano Forum, if you want to poll more users & developers) if you need help.

KtorZ:
Stressing out @rphair's comment, the lack of activity on this proposal and its very draft state; we have decided in the last editor meeting to close the PR for the sake of keeping the repository tidy.

I'd encourage you @micahkendall  to participate in the following conversations which I believe addresses some of the problems you're raising here:

- #446 
- #462

PR #458, opened 2023-02-04
CIP-0068 | Extra field

alessandrokonrad:
Added extra field to datums to allow for user customized data. Additionally made it explicit that additional properties are alllowed in metadata.
@perturbing agrees on these changes as well.

### DIFF
--- a/CIP-0068/README.md
+++ b/CIP-0068/README.md
@@ -57,7 +57,7 @@ Some remarks about the above,
 1. The `user token` and `reference NFT` do not need to be minted in the same transaction. The order of minting is also not important.
 2. It may be the case that there can be multiple `user tokens` (multiple asset names or quantity greater than 1) referencing the same `reference NFT`.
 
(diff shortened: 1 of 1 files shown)


CIP-0068/README.md
```diff
@@ -115,9 +116,15 @@ metadata =
     ? mediaType : bounded_bytes, ; UTF-8
     ? description : bounded_bytes, ; UTF-8
     ? files : [* files_details]
+    ; ... Additional properties are allowed
```
KtorZ, line 119: Not sure to follow this line 🤔 ? Is this in reference to the new `extra` ? In which case, additional properties aren't really on the `metadata` record, right?

mmahut: This comment is just to clarify that additional proprieties are allowed within the `metadata` field.

alessandrokonrad: @KtorZ it's not really related to the extra field. Just thought I sneak it into the PR, wanted to make it more explicit that additional properties are allowed.

KtorZ (approved):

rphair (approved):

PR #460, opened 2023-02-06
Add the Plutus category to a couple of existing CIPs

michaelpj:


rphair (approved):
thanks @michaelpj - we were hoping to roll through this kind of thing in https://github.com/cardano-foundation/CIPs/issues/389 but aren't moving as quickly on it as we would like.

KtorZ (approved):

PR #461, opened 2023-02-09
CIP-0036 | Fix CDDL naming to match changes in spec from #427 

Ryun1:
### Motivation
Changes #427 introduced proposed that tool makers supporting this CIP should call the keys described within it "(CIP-36) vote keys" over "voting keys". This change was reflected in the specification and testvectors file, but not accurately reflected in the CDDL schema.

### Proposed Change
Rename `voting_pub_key` to be `cip36_vote_pub_key` in the CDDL schema file.

### Rationale
This change matches the CDDL schema to naming suggested within the CIP specification itself.

**Note**: Thank you @janmazak for spotting this.

### DIFF
--- a/CIP-0036/schema.cddl
+++ b/CIP-0036/schema.cddl
@@ -3,13 +3,13 @@ registration_cbor = {
   61285: registration_witness
 }
 
-$voting_pub_key /= bytes .size 32
+$cip36_vote_pub_key /= bytes .size 32
 $payment_address /= bytes
 $nonce /= uint
 $weight /= uint .size 4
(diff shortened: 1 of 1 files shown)


janmazak (approved):

rphair (approved):
approving because this simply looks like the nomenclature is being adjusted for consistency with recommendations added to CIP-0036 here: https://github.com/cardano-foundation/CIPs/blob/master/CIP-0036/README.md#tooling

KtorZ (approved):

PR #462, opened 2023-02-16
CIP-0090? | Extendable dApp-Wallet Web Bridge

Ryun1:
This document describes a *base* web-wallet communication bridge accompanied by a scheme from which functionality can be added through API extensions.

This specification defines a wallet-dApp connection standard using injected Javascript object, with an extendable API interface.

This aims to allow the creation of dApp connectors without the need to include the CIP-30 API as proposed in #446. 
Instead the CIP-30 API can act as an optional extension to this specification.

This draws direct inspiration from @mkazlauskas' [comment](https://github.com/cardano-foundation/CIPs/pull/446#discussion_r1103630573), utilizing @ktorz's design from #446. Thus the author accreditations.

### Todo
- [ ] Rewrite to align with current landscape of CIP30 extensions.

---

([rendered proposal in branch](https://github.com/Ryun1/CIPs/blob/extendable-dapp-connector/CIP-0090/README.md))


CIP-XXXX/CIP-XXXX.md
```diff
@@ -0,0 +1,180 @@
...
+
+#### cardano.{walletName}.isEnabled(): Promise\<bool>
+
+Errors: APIError
+
+Returns true if the dApp is already connected to the user's wallet, or if requesting access would return true without user confirmation (e.g. the dApp is whitelisted), and false otherwise. If this function returns true, then any subsequent calls to `wallet.enable()` during the current session should succeed and return the `API` object.
+
+#### cardano.{walletName}.supportedExtensions: Extension[]
+
+A list of extensions supported by the wallet. Extensions may be requested by dApps on initialization. Some extensions may be mutually conflicting and this list does not thereby reflect what extensions will be enabled by the wallet. Yet it informs on what extensions are known and can be requested by dApps if needed.
```
mkazlauskas, line 123: For backwards compatibility
```suggestion
#### cardano.{walletName}.supportedExtensions: Extension[] | undefined

A list of extensions supported by the wallet. Extensions may be requested by dApps on initialization. Some extensions may be mutually conflicting and this list does not thereby reflect what extensions will be enabled by the wallet. Yet it informs on what extensions are known and can be requested by dApps if needed.

`undefined` is equivalent to `{cip: 30}`.

```

CIP-XXXX/CIP-XXXX.md
```diff
@@ -0,0 +1,180 @@
...
+
+By stripping functionality from this API we facilitate dApp-Wallet connections with less functions than CIP-30 API. A use case for this is connections where minimal wallet information is wished to be shared with a dApp.
+
+Here we attempt to mitigate the impact of succeeding CIP-30 as *the* Cardano wallet connector, by enabling it's functionality by default at connection time. This should minimize the impact this proposal could have on established dApps who support CIP-30. By only adding what is necessary for extensions we aim to reduce the required load for wallets to adopt this standard.
+
+### Extensions
+
+Extensions provide an extensibility mechanism and a way to negotiate (possibly conflicting) functionality between a dApp and a wallet provider. There's rules enforced as for what extensions a wallet decide to support or enable. The current mechanism only gives a way for wallets to communicate their choice back to a dApp.
+
+We use object as extensions for now to leave room for adding fields in the future without breaking all existing interfaces. At this point in time however, objects are expected to be singleton.
```
mkazlauskas, line 157: Don't quite understand what 'singleton' means in this context (or what is the 'object'). Would be nice to elaborate.

CIP-XXXX/CIP-XXXX.md
```diff
@@ -0,0 +1,180 @@
...
+3. There's a need from an API consumer standpoint to clearly identify what features are supported by providers.
+
+By moving CIP-30 style modifications into a formalized extension scheme we aim to avoid fragmentation of functionality between different dApp-Wallet bridges.
+
+## Path to Active
+
+### Acceptance Criteria
+
+- [ ] The interface is implemented and supported by various wallet providers.
+- [ ] The interface is used by dApps to interact with wallet providers.
```
mkazlauskas, line 172: This is supposed to be backwards-compatible with cip30, so I believe this criteria is already satisfied.
```suggestion
- [x] The interface is implemented and supported by various wallet providers.
- [x] The interface is used by dApps to interact with wallet providers.

```

CIP-XXXX/CIP-XXXX.md
```diff
@@ -0,0 +1,180 @@
+---
+CIP: ????
+Title: Extendable dApp-Wallet Web Bridge
+Category: Wallets
+Status: Proposed
+Authors:
+    - Ryan Williams <ryan.williams@iohk.io>
+    - Martynas Kazlauskas <martynas.kazlauskas@iohk.io>
+    - Matthias Benkort <matthias.benkort@cardanofoundation.org>
```
mkazlauskas, line 9: This is based on cip30
```suggestion
    - Matthias Benkort <matthias.benkort@cardanofoundation.org>
    - rooooooooob
```

mkazlauskas (changes requested):

CIP-XXXX/CIP-XXXX.md
Ryun1: Wallets will still have to implement the additional extension scheme functions. Although this change should be transparent to CIP-30 compatible dApps.

CIP-XXXX/CIP-XXXX.md
Ryun1: This means that `extension` objects cannot pass additional fields at enable time, currently.

CIP-XXXX/CIP-XXXX.md
```diff
@@ -0,0 +1,180 @@
...
+### Full API
+
+Upon successful connection via `cardano.{walletName}.enable()`, a javascript object we will refer to as `API` (type) / `api` (instance) is returned to the dApp with the following method.
+
+This is where extensions are free to add functionality as they see fit. Furthermore extensions are free to include additional sections, naming their APIs as they see fit, an example could be an `Experimental API`.
+
+#### api.getExtensions(): Promise\<Extension[]>
+
+Retrieves the list of extensions enabled by the wallet. This may be influenced by the set of extensions requested in the initial `enable` request.
+
```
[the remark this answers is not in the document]
Ryun1: Calling enable repeatedly could result in the user's permission being asked multiple times.

CIP-XXXX/CIP-XXXX.md
```diff
@@ -0,0 +1,180 @@
...
+Yes. Extensions may introduce new endpoints or error codes, and modify existing ones. Extensions may even change the rules outlined in this very proposal. The idea being that wallet providers should start off implementing this CIP, and then walk their way to implementing their chosen extensions.
+
+##### Are wallet expected to implement all extensions?
+
+No. It's up to wallet providers to decide which extensions they ought to support.
+
+### Error Types
+
+#### APIError
```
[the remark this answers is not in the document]
Ryun1: This is a good spot, I agree, I will make this change.

CIP-XXXX/CIP-XXXX.md
```diff
@@ -0,0 +1,180 @@
...
+
+Yes. Extensions may have other extensions as pre-requisite. Some newer extensions may also invalidate functionality introduced by earlier extensions. There's no particular rule or constraints in that regards. Extensions are specified as CIP, and will define what it entails to enable them.
+
+##### Should extensions follow a specific format?
+
+Yes. They all are CIPs.
+
+##### Can extensions add their own endpoints and/or error codes?
+
+Yes. Extensions may introduce new endpoints or error codes, and modify existing ones. Extensions may even change the rules outlined in this very proposal. The idea being that wallet providers should start off implementing this CIP, and then walk their way to implementing their chosen extensions.
```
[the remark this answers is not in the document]
Ryun1: > This could be a mess.

I agree, we need to navigate this carefully. 

I see three options:
- we silo extensions, not allowing them to overlap or share anything (datatypes, error codes, etc.) only allowing them to extend this API. I believe this aligns with your suggestion.
-  as described currently - we allow extensions to share, replace and overlap with each other.
- somewhere in the middle, I think some novel design would be needed for this to be viable.

I would like to hear other's insights on how is best to handle this.

CIP-XXXX/CIP-XXXX.md
```diff
@@ -0,0 +1,180 @@
...
+APIError {
+	code: APIErrorCode,
+	info: string
+}
+```
+
+* InvalidRequest - Inputs do not conform to this spec or are otherwise invalid.
+* InternalError - An error occurred during execution of this API call.
+* Refused - The request was refused due to lack of access - e.g. wallet disconnects.
+* AccountChange - The account has changed. The dApp should call `wallet.enable()` to reestablish connection to the new account. The wallet should not ask for confirmation as the user was the one who initiated the account change in the first place.
```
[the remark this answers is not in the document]
Ryun1: This proposal only inherited this from CIP-30, I am not sure of how prevalent/useful this actually is.

CIP-XXXX/CIP-XXXX.md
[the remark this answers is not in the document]
Scitz0: I lean toward option 1 on your list. It could quickly become a mess with dependencies between extensions etc. If the same data types/error codes are defined in multiple extensions, it could be up to the implementer to combine and merge as it sees fit.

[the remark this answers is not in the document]
Ryun1: > I lean toward option 1 on your list.

I have been reflecting on this and I agree. We can always give extensions more freedom to share down the line, if needed. What we cannot do is try to impose restrictions on extensions once we discovery that granting flexibility has resulted in a dependency nightmare.

CIP-XXXX/CIP-XXXX.md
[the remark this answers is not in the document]
Scitz0: Agree, I think this function should exist. If I'm not mistaken it's needed in the discovery phase by the dapp to know if the wallet injected supports the required extensions to only show supported wallets in wallet connector?

CIP-XXXX/CIP-XXXX.md
Scitz0: Do we need to limit this in this text? Isn't it enough to say that this CIP only defines a single key in the object passed, and that it's up to Extensions to define additional fields in their respective CIP, if any?

CIP-XXXX/CIP-XXXX.md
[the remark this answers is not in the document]
Scitz0: Agree, AccountChange makes no sense for this PR, a CIP30-specific thing.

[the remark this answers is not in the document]
Scitz0: Unsure how I feel about this one. As the idea is to try to define the absolute base functionality for the dapp connector, knowing if the connected account has changed or not I feel could be part of this.

CIP-XXXX/CIP-XXXX.md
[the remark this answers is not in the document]
mkazlauskas: > Calling enable repeatedly could result in the user's permission being asked multiple times.

Yes, if dApp is asking for more permissions on the 2nd call.

> If I'm not mistaken it's needed in the discovery phase by the dapp to know if the wallet injected supports the required extensions to only show supported wallets in wallet connector

Discovery phase should be using [supportedExtensions](https://github.com/cardano-foundation/CIPs/pull/462/files#diff-4ffe445f8e72a6cee701a5a989aef62750d96c6296058bb612cad8ff305b49a5R130)

This method is returning "extensions enabled by the wallet" and I still think is quite useless.

CIP-XXXX/CIP-XXXX.md
mkazlauskas: > This means that extension objects cannot pass additional fields at enable time, currently

Thanks. Could be just me not understanding it from text, but consider rewording it @Ryun1 

> Do we need to limit this in this text? Isn't it enough to say that this CIP only defines a single key in the object passed, and that it's up to Extensions to define additional fields in their respective CIP, if any?

The simpler the base dApp connector cip the better imho. The scope of the extension CIPs can then be contained to just the API object that `enable()` returns. The extensions can have API methods (or arguments for those methods) instead. If we allow configuration at `enable()` time then we can expect to see extension designs that would require the dApp calling `enable` multiple times with different arguments. Do we want that? :sweat_smile:

CIP-XXXX/CIP-XXXX.md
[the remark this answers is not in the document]
Scitz0: You are totally correct, I skimmed through it too fast and missed [supportedExtensions](https://github.com/cardano-foundation/CIPs/pull/462/files#diff-4ffe445f8e72a6cee701a5a989aef62750d96c6296058bb612cad8ff305b49a5R130)

getExtensions() in the full API should indeed go. 👍

CIP-XXXX/CIP-XXXX.md
mkazlauskas: Yeah, it's probably up to CIP editors to set this criteria

CIP-XXXX/CIP-XXXX.md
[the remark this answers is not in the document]
mkazlauskas: There is no connected account from the perspective of the API of this CIP. It's only connected after you `enable()`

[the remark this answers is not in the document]
Ryun1: This change has been made.

CIP-XXXX/CIP-XXXX.md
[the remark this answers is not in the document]
Ryun1: This change has been made.

CIP-XXXX/CIP-XXXX.md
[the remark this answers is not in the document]
Ryun1: This change has been made.

CIP-XXXX/CIP-XXXX.md
Ryun1: > Do we want that?
You raise a good point, I think this could become messy quickly.

And I have removed this sentence to avoid the confusing wording.

CIP-XXXX/README.md
```diff
@@ -0,0 +1,190 @@
...
+{ "cip": 30 }
+```
+
+##### Known Extensions
+
+This should act as the definitive list of known and accepted extensions to this proposal. 
+
+| CIP      | Scope                           |
+| -------- | ------------------------------- |
+| CIP-30   | Generic interface               |
```
KtorZ, line 56:
```suggestion
| [CIP-0030](https://github.com/cardano-foundation/CIPs/tree/master/CIP-0030)   | Generic interface               |
```

CIP-XXXX/README.md
```diff
@@ -0,0 +1,190 @@
...
+
+Once a connection is established, this can be called multiple times by the dApp to request access to further extensions. User permission must be asked if new extensions are requested. If this is called without new extensions requests then no user permission is required.
+
+**Note**: In order to improve backwards compatibility an empty `extensions` field should supply CIP-30 functionality, as if a CIP-30 `extension` object was provided.
+
+#### cardano.{walletName}.isEnabled(): Extension[]
```
KtorZ, line 126: I believe the return type has to be a promise here (`Promise<Extension[]>`) because the API is asynchronous. It's been a while since I played with it though so I may be wrong.

KtorZ (commented):
Love it. Thanks for the pushing this @Ryun1. 

Any thoughts @Quantumplation ?

CIP-XXXX/README.md
Ryun1: @KtorZ 

Good spot - I have amended.

CIP-XXXX/README.md
```diff
@@ -0,0 +1,190 @@
...
+
+* InvalidRequest - Arguments do not conform to this specification or are otherwise invalid - e.g. requested extension is not supported by the wallet.
+* InternalError - An error occurred during execution of this API call.
+* Refused - The request was refused due to lack of access - e.g. user rejected connection to this dApp.
+
+### Initial API
+
+In order to initiate communication from webpages to a user's Cardano wallet, the wallet must provide the following javascript API to the webpage. A shared, namespaced `cardano` object must be injected into the page if it did not exist already. Each wallet implementing this standard must then create a field in this object with a name unique to each wallet containing a `wallet` object with the following methods. The API is split into two stages to maintain the user's privacy, as the user will have to consent to `cardano.{walletName}.enable()` in order for the dApp to read any information pertaining to the user's wallet with `{walletName}` corresponding to the wallet's namespaced name of its choice.
+
+#### cardano.{walletName}.enable(extensions: Extension[] = [{ cip: 30 }]): Promise\<API>
```
SebastienGllmt, line 114: Same comment about structural vs positionsl as I left on the other CIP30 PR

stevenj: Given this is defined in CIP-30 and is already implemented by wallets, this is asking for trouble on a backwards compatibility front.

It would be simpler if the wallets just `enable` ALL extensions they are aware of.

Alternatively, there could be:
```
cardano.{wallet}.reqExtensions(extensions: Extension[])
```
This could pre-informs the wallet what extensions the dapp wants to use.  The `enable` then just `enables` those.  The default if its not called should be enable all extensions the wallet supports.  However, in this case, I think its informative only (maybe the wallet can use it to display something meaningful to the user?).  The wallet should enable all the extensions that it supports, which were requested.  If an extension was requested but the wallet does not support it, that should not be an error.

CIP-XXXX/README.md
```diff
@@ -0,0 +1,190 @@
...
+
+Errors: `ConnectorError`
+
+Returns a list of extensions which have been enabled by the wallet. If no wallet access has been granted then a `ConnectorError` with  `Refused` error code should be thrown.
+
+#### cardano.{walletName}.supportedExtensions: Extension[] | undefined
+
+A list of extensions supported by the wallet. Extensions may be requested by dApps on initialization. Dapps can repeatably call `wallet.enable()` to add further extensions. Some extensions may be mutually conflicting, and this list does not thereby reflect what extensions will be enabled by the wallet. Yet it informs on what extensions are known and can be requested by dApps if needed.
+
+`undefined` is equivalent to `{cip: 30}`.
```
stevenj, line 136: I think this is redundant.

I do not see much utility in fine grain switching of extensions.  
If the process was just:

```
cardano.{walletName}.enable();
extensions = cardano.{walletName}.isEnabled();
```

That tells you everything you need as a dApp.
1. Does this wallet support this cip, or is it just obeying cip-30.
2. What extensions does the wallet support.
3. All supported extensions are enabled.

CIP-XXXX/README.md
```diff
@@ -0,0 +1,190 @@
...
+
+Once a connection is established, this can be called multiple times by the dApp to request access to further extensions. User permission must be asked if new extensions are requested. If this is called without new extensions requests then no user permission is required.
+
+**Note**: In order to improve backwards compatibility an empty `extensions` field should supply CIP-30 functionality, as if a CIP-30 `extension` object was provided.
+
+#### cardano.{walletName}.isEnabled(): Promise\<Extension[]>
+
+Errors: `ConnectorError`
+
+Returns a list of extensions which have been enabled by the wallet. If no wallet access has been granted then a `ConnectorError` with  `Refused` error code should be thrown.
```
stevenj, line 130: Again from a backward compatibility viewpoint, this should be allowed to return either a `bool` or a list of enabled extensions.  a bool set to true is equivalent to `CIP-30`, and should mean "This wallet does not support this CIP".  

Even though a wallet that implements this CIP should not return a bool, its a valid response from a CIP-30 wallet and it should be clear how a dapp reacts to it.

CIP-XXXX/README.md
```diff
@@ -0,0 +1,190 @@
+---
+CIP: ????
```
```suggestion
CIP: 90
```

CIP-XXXX/README.md
Ryun1: I do agree with this, I think I will do a revision of this proposal.

CIP-XXXX/README.md
Ryun1: I agree with this also, I think I had not considered backward compatibility enough... I plan to do a more significant rewrite when/if #486 is merged.

Ryun1:
I plan to do a significant rewrite once there is more community consensus around #486.

Ryun1:
With the advent of:
- #509
- #588 
- #589 
- #617 
- #619 
- #620 

I plan to work on this standard again

rphair:
@Ryun1 since this still has not moved forward (and I am not rushing you) I'm reapplying the `Waiting for Author` tag that you removed 9 months ago.  As with all those tagged as such, it will help to have a comment indicating what it is "Waiting for Author" _for_ 😅

Ryun1:
Leave me to it...

klntsky:
@Ryun1 how would the new proposal differ?

rphair:
@Ryun1 is this effectively deprecated by
- https://github.com/cardano-foundation/CIPs/pull/957

? (we wouldn't have to mark it as such, though I did want to get your assessment on record)

Ryun1:
> is this effectively deprecated by
> 
> https://github.com/cardano-foundation/CIPs/pull/957
> ? (we wouldn't have to mark it as such, though I did want to get your assessment on record)

Yup!
solutions like https://github.com/cardano-foundation/CIPs/pull/957 I feel are much better than what I had proposed here.

rphair:
thanks @Ryun1 - will tag `Deprecated` then & if you think that sends the wrong message please feel free to remove.

PR #371, opened 2022-11-09
CPS-0001 | Metadata Discoverability and Trust

ehanoc:
There seems to be multiple proposals and attempts to solve issues around metadata specific to certain contexts. Proposing this as a first CPS in order for us to agree what the general problem actually is and try to steer the process to tackle the root cause of the problem. This is my attempt at describing the problem.

---

([current draft in branch](https://github.com/ehanoc/CIPs/tree/cps-metadata-discoverability/CPS-0001))

PR #467, opened 2023-03-01
CIP-0088? | Token Policy Registration

Crypto2099:
A proposal to create a standardized format, similar to stake pool registration certificates and Catalyst voter registration certificates that will allow Native Asset policies and the projects/creators behind them to register the intent and top-level information pertaining to their project.

This is a very rough first draft of the proposal and some refinements and sub-specifications are already underway to formalize the payloads mentioned.

The proposal aims to be as forward and backwards compatible for all native script policies assuming project creators still have access to their keys. This standard should allow us to have dynamic, readily updateable metadata at a token policy level with minimal ledger bloat and maximum flexibility.

The **867** JSON metadata index is used in the examples here and would be acceptable as a CIP number for the sake of consistency. The goal would be to register, reserve, and utilize the JSON index that matches the CIP number for ease of reference.

---

([rendered proposal in branch](https://github.com/Crypto2099/CIPs/blob/token-registration/CIP-0088/README.md))


CIP-XXXX/README.md
```diff
@@ -0,0 +1,364 @@
...
+key hash of the signing key (as included in the native script) and the signed key witness. If a script requires multiple
+signatures, enough signatures to meet the criteria of the script should be included and required for proper validation
+of an updated token registration.
+
+**Example**
+
+```json 
+[
+  [
+    "e97316c52c85eab276fd40feacf78bc5eff74e225e744567140070c3j",
```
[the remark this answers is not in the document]
Crypto2099: Can you submit binary via on-chain metadata? Wouldn't it need to be string-encoded to match the JSON format?

CIP-XXXX/README.md
```diff
@@ -0,0 +1,364 @@
...
+ecosystem. At current, most DEXes pull information from the _Cardano Token Registry_ but there is no similar function
+for NFT projects. As such, much information must be manually provided to individual marketplaces by the token projects
+creating an undue burden on the project creators to provide a largely static amount of information via different web
+forms and authentication schemas rather than simply publishing this information to the blockchain directly.
+
+## Specification
+
+### Registration Metadata Format
+
+A Native Script registration is a regular Cardano transaction with specific metadata attached (similar to CIP-36
```
michele-nuzzi, line 69: Since we are exploring an on-chain solution for token metadata registration would it be possible to make available critical informations (such as FT decimals) using inline datums?

Crypto2099: Hrm, not a bad idea, we could include a reference UTXO that contains inline datum w/ FT token registry details via this method so that SCs could have an on-chain, in-SC method for reading the data.

CIP-XXXX/README.md
```diff
@@ -0,0 +1,364 @@
...
+    ]
+  }
+}
+```
+
+### Token Registration Witness Array
+
+The Witness Array included in the on-chain metadata should consist of an array of arrays with two elements, the public
+key hash of the signing key (as included in the native script) and the signed key witness. If a script requires multiple
+signatures, enough signatures to meet the criteria of the script should be included and required for proper validation
```
nullhashpixel, line 189: Can this lead to ambiguities when for a 2 out of [A, B, C] script A+B and B+C both individually push a different update with the same nonce?

Crypto2099: Hrm, that is a potential edge-case scenario that would need to be tested and developed against. Would potentially have to resort to accepting the "first seen" in that case?

CIP-XXXX/README.md
kiriakos: What's the downside of having just the datum in the tx/txo?
Afaic most (all?) chain indexers can deal with datums now so using a datum instead of tx metadata tags would allow us to have a single source of truth and be SC compatible.

CIP-XXXX/README.md
nullhashpixel: I guess it would not happen "by chance", more like in a situation with B as kind of a Byzantine assassin, so B would trick A and C individually to sign different multi-sig transactions and then submit them within the same block. Maybe enough to take block height + tx hash?

michele-nuzzi: If the nonce will be a hash then the id of any of the input utxos is guaranteed to be unique

CIP-XXXX/README.md
Crypto2099: The problem is that most of this data is not SC-relevant and needing to maintain a UTXO in perpetuity in order to provide things like social media handles or off-chain information is a waste of UTXO (and minUTXO) space. Appropriate "lensing" of data that is/should be referenced within a smart contract and referencing that via a simple transaction would be more apt.

CIP-XXXX/README.md
Crypto2099: As described, the nonce should be a "counter value" [uint] so you could always reference the most recent update (largest integer) and avoid a historic "replay" attack.

CIP-XXXX/README.md
fallen-icarus: @michele-nuzzi @Crypto2099 

> needing to maintain a UTXO in perpetuity in order to provide things like social media handles or off-chain information is a waste of UTXO

You do not need to maintain it in perpetuity.

> would it be possible to make available critical informations (such as FT decimals) using inline datums?

This is an area [beacon tokens](#466) can be used to trustlessly share this information. I actually created a reference implementation that does something similar called [cardano-datums](https://github.com/fallen-icarus/cardano-datums). I was curious if beacon tokens would work here and they do. **Using beacon tokens, that information is forever accessible even if the UTxO is consumed.** The Beacon CIP is a work in progress but it can work for all metadata. BTW using inline datums instead of metadata also means the format is strictly enforced - if a required field is missing, registration would fail.

fallen-icarus:
The [cardano-address-book](https://github.com/fallen-icarus/cardano-address-book) program I created does something similar. It cryptographically links metadata to a specific user in an easily queryable manner. If you are looking to tie keys/scripts to a minting policy, I think it would be more secure to use beacon tokens like in the address-book program.

Ryun1 (commented):
A couple small things.

CIP-XXXX/README.md
```diff
@@ -0,0 +1,364 @@
...
+      ]
+    ]
+  }
+}
+```
+
+## Rationale
+
+For this specification, I have drawn inspiration from 
+[CIP-30: Catalyst/Voltaire Registration Metadata Format](https://github.com/cardano-foundation/CIPs/tree/master/CIP-0036)
```
Ryun1, line 335:
```suggestion
[CIP-36: Catalyst/Voltaire Registration Metadata Format](https://github.com/cardano-foundation/CIPs/tree/master/CIP-0036)
```
typo

CIP-XXXX/README.md
```diff
@@ -0,0 +1,364 @@
...
+
+### Implementation Plan
+
+In order to foster and facilitate implementation of this CIP I will work on some open-source code highlighting how to
+create, submit, and index/validate these on-chain registrations to make it as easy as possible for third party consumers
+to utilize this data quickly and efficiently in their endpoints.
+
+## Copyright
+
+[CC-BY-4.0]: https://creativecommons.org/licenses/by/4.0/legalcode
```
Ryun1, line 364:
```suggestion
This CIP is licensed under [CC-BY-4.0](https://creativecommons.org/licenses/by/4.0/legalcode).
```
To match header? so this is viewable when the markdown is rendered.

CIP-XXXX/README.md
```diff
@@ -0,0 +1,364 @@
+---
+CIP: ?
+Title: Hybrid Token Policy Registration
+Category: Tokens
+Status: Proposed
+Authors:
+    - Adam Dean <adam@crypto2099.io>
+Implementors: []
+Discussions:
+    - https://github.com/cardano-foundation/cips/pulls/?
```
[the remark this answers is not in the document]
Crypto2099: I had to run out this weekend for a family affair but will get this sorted :) Thanks!

CIP-XXXX/README.md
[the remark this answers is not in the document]
AndrewWestberg: I would at least write the spec to allow for either. json is a limitation of building a transaction with cardano-cli and not the on-chain metadata itself. You can save 50% of the size by not using hex strings. CIP-25 v2 has been updated to allow binary strings.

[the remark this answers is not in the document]
Crypto2099: I will take a look and revise accordingly!

CIP-XXXX/README.md
Crypto2099: @michele-nuzzi added a reference to an inline datum in the CIP-26 (FT) specific registration section, would love to hear thoughts/feedback :) Can be found in the CIP26 folder/readme

cent-development:
Hi Adam, I'm leaving my comment / suggested addition to the CIP as we discussed in the NFT Guild Roundtable 18 (Wed March 8th), so it can be discussed further. (Link to recording will shortly be available at https://www.nft-guild.io/the-nft-roundtable)

It is specified in the section "Token Registration Witness Array" to include public key hash "as included in the native script". I agree that this will (as you say) be "proper validation of an updated token registration" for native scripts. As commented in said meeting, this will not be the case for smart contract minting policies who do not have an owner in the same sense as native scripts.

Plutus smart contract scripts only contain a script cbor hex. I am adding a minimal example here for reference.

```
{
    "type": "PlutusScriptV2",
    "description": "",
    "cborHex": "49480100002221200101"
}
```
This is the simplest / shortest possible Plutus smart contract, but the structure is the same for all contracts.

The "ownership" of a smart contract or "verification that the owner of a certain pub key hash can mint tokens using this policy" can not be proved directly by examining the script file. If the owner pub key hash is hard coded into the Plutus code, you could verify by manual examination of the code, but then you would need to also build the code to verify it is in fact the correct policy id. This would of course be too cumbersome to be usable.

I think small additions can be made to this CIP to make it work also for smart contract policies. An example could be to support that the same 867 metadata can be part of the tx metadata of special "identity token" minted with the policy. This would prove that you are able to mint tokens with the policy, or at least that someone that can mint tokens with it has named you to be an owner, and the pub key hash would be in the metadata of the token. The identity token would then enable automatic proof and verification for the relationship between the Plutus minting policy and pub key hash.

Or...is there another solution than the "identity token"?

I might add that if the logic of the plutus minting policy is programmed to be locked after some time similar to locked native scripts, then this identity token would not be a possibility as no new minting of identity token can be done. So this approach is not backwards compatible with "locked" plutus policy scripts.

Crypto2099:
Thank you for taking the time to write this up @cent-development! It is indeed tricky to apply this same logic to Plutus minting scripts, particular I believe impossible to do so via backwards compatibility as we cannot retroactively introduce the logic to mint a "beacon" or "identity" token. The other part of the problem (I suppose) would be having to introduce additional complexity/size to all Plutus minting scripts going forward for what may be minor benefit...

Just to spitball ideas... since most scripts should now take advantage of script references via UTXO that would be in the control of the deploying entity (or some sort of multisig DAO or something)… Maybe including this 867 metadata in the tx where the script is deployed would be sufficient?

Or, one could, optionally of course, somehow add a pub key when deploying a Plutus script reference (via inline datum perhaps?) that could then be used for signing and issuing updates?

cent-development:
Yes, backward compatibility seems impossible to achieve for Plutus minting policies if we cannot figure out a mechanism that doesn't include the need for identity token minting. 

I agree that the reference UTxOs may be nice feature to make use of in this solution. Throwing out a couple of ideas that also can be investigated (NFT guild is already looking into this and we can help of course as we are eager to participate in introducing token verification to the community:) ):

1 - Minting contract owner can mint token called ID (or something that can act as standard) that has the 867 metadata attached either as Datum or tx metadata. (This is perhaps the option you mentioned?)
2 - Mint token called `ID-<ownerpubkeyhash>` or some other prefix if owner wants another name `<random-name>-<ownerpubkeyhash>`. 

Information added to the 867 metadata is the token name as a value in the Witness Method (element 3) and the owner pub key hash as usual in Witness Array (element 7).

Verification in both options can be done by simply verifying the token is minted using the correct minting policy and that it is named according to the 867 metadata.

CIP-XXXX/README.md
```diff
@@ -0,0 +1,447 @@
...
+forward.
+
+**Example:**
+
+`[25, 27]`
+
+#### 3. Witness Method
+
+The witness method portion of the payload should describe the method used to generate the witness signature (CIP-8, etc)
+to aid in validation by third parties.
```
Ryun1, line 116: Might be nice to include an explanation of the flow for discovery and validation for clients.

CIP-XXXX/README.md
```diff
@@ -0,0 +1,447 @@
...
+          "assuming that the overall length of the witness is longer than the",
+          "64-character on-chain metadata limit for strings."
+        ]
+      ]
+    ]
+  }
+}
+```
+
+## Rationale
```
Ryun1, line 415: It appears to me that this proposal goes some way to address [CPS-0001](https://github.com/cardano-foundation/CIPs/pull/371), might be nice to mention this CPS here. 

Also it would be good to add more to this section to articulate the potential alternative designs.

rphair: yes: at least the first part of this suggestion was well received at today's CIP meeting (cc @ehanoc)

CIP-XXXX/README.md
```diff
@@ -0,0 +1,447 @@
+---
+CIP: ?
```
```suggestion
CIP: 88
```

nftguild:
See Adam Dean give a brief overview of his new CIP, what it's all about and the main issues the proposal is trying to address.
[https://youtu.be/8V1zfu76hyw](Intro to CIP 88 - Native Asset Policy Registration/Information Certificates)

CIP-XXXX/README.md
Crypto2099: Agreed, this discovery flow description is currently being worked on as I am working on getting some NFTs I control on preprod so I can use the various tools involved to highlight and show various process of discovery, indexing, and validation as part of the CIP.

CIP-XXXX/README.md
Crypto2099: Thank you, I will review that CPS and add references as appropriate.

Can you expand on what you mean by "articulate the potential alternative designs"?

rphair:
thanks @nftguild but I would suggest the video title be corrected, and the featured image touched up (not expecting a re-recording of video), to show the actual CIP number `88`.  I understand the fondness for the number `867` but in the long run the community will be fonder of not using 2 different numbers for this CIP. 🙏 (cc @Crypto2099)

Crypto2099:
@rphair hehe I was doing my best to keep emphasizing "working number, tbd" in all the context I've discussed the CIP. That video is property of @nftguild but I'm sure they'll be happy to correct the title at the very least. Updates to the repo to accommodate the newly assigned number coming soon. :)

CIP-XXXX/README.md
Ryun1: > Can you expand on what you mean by "articulate the potential alternative designs"?

What I mean is, what could be competing design elements to the ones in this CIP? if any? -- for example: is there any alternative to storing this information in transaction metadata?

rphair:
@Crypto2099 it seems to me that this proposal supersedes 2 other proposals, designed to work together & from the same author, to handle on-chain references to NFT / FT common data:

- https://github.com/cardano-foundation/CIPs/pull/249
- https://github.com/cardano-foundation/CIPs/pull/343

Perhaps you would give them a look & comment if you think there is anything this other framework could provide that is not handled (better in my opinion) by CIP-0088.  Otherwise we will probably deprecate those pending proposals in favour of this one.

CIP-XXXX/README.md
rphair: @Ryun1 there was an alternative framework that came along last year but it was not very popular & has been waiting in the wings now for a few months.  It may be useful to have a look at these as alternatives for comparison: https://github.com/cardano-foundation/CIPs/pull/467#issuecomment-1471481889

nftguild:
> thanks @nftguild but I would suggest the video title be corrected, and the featured image touched up (not expecting a re-recording of video), to show the actual CIP number `88`. I understand the fondness for the number `867` but in the long run the community will be fonder of not using 2 different numbers for this CIP. 🙏 (cc @Crypto2099)

Thanks for the heads up @rphair . We've chaged the title and the video thumbnail to CIP 88.
At the time we published the videos (both the full recording of the call and the short presentation) the CIP didn't have an official designation assigned to it so we went with the working title given by @Crypto2099 . We were aware we would probably have to go back and change it later, but we weren't expecting it to be this soon :-). But it's all done now.

CIP-XXXX/README.md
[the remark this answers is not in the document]
KtorZ: > Can you submit binary via on-chain metadata? Wouldn't it need to be string-encoded to match the JSON format?

Yes of course! And at this point I don't know why people keep submitting hex-encoded strings to the chain instead of plain byte strings. Hex-encoded data are twice the size and less efficient to decode...

CIP-XXXX/README.md
```diff
@@ -0,0 +1,447 @@
...
+policy on-chain. This CIP will aim to create a method that is backwards compatible and enable projects to declare, and
+update over time, the supported feature set and metadata details pertaining to their tokens.
+
+This CIP will aim to make use of a hybrid (on and off-chain) information schema to enable maximum flexibility and 
+adaptability as new and novel use cases for native assets (NA) expand and grow over time.
+
+## Motivation
+
+This CIP was borne out of a distaste for the lack of on-chain token policy intent registration that has been cited as
+both a centralization and security concern at various points over the preceding two years of NA history on Cardano.
```
KtorZ, line 27: NA -> Native Assets ?

CIP-XXXX/README.md
```diff
@@ -0,0 +1,447 @@
...
+### Example 1: The Cardano Token Registry
+
+Many Fungible Token (FT) projects require special treatment of their NAs such as decimal places for proper display and
+formatting, project information, and token logo. As it stands, these projects must currently register via a GitHub
+repository ([Cardano Token Registry](https://github.com/cardano-foundation/cardano-token-registry)) in order to have
+their token properly appear in wallets. This GitHub repository is currently managed by the Cardano Foundation (CF).
+While there is no reason to believe that the CF would take any malicious or nefarious action, forcing token projects to
+register in this fashion introduces a point of centralization, _potential_ gate keeping, and ultimately reliance on the
+interaction of a 3rd party (the human at CF responsible for merging pull requests) in order for projects to register, or
+update, their information.
```
KtorZ, line 38: As a side-note: the intent was never for the CF to be the sole actor to provide a token registry. When originally envisioned, the entire solution was meant to be run and provided by various actors. The CF would only be one of the possible actors. Yet, as time now shows, there hasn't been any actor stepping up to provide their own token metadata server and wallets implementators haven't given this flexibility (of choosing their own source of data) to users.

CIP-XXXX/README.md
```diff
@@ -0,0 +1,447 @@
...
+
+### Example 2: Token Metadata Insecurity
+
+One of the stated rationales for [CIP-68](https://github.com/cardano-foundation/CIPs/tree/master/CIP-0068) was that the
+"original" Cardano NFT Metadata Standard ([CIP-25](https://github.com/cardano-foundation/CIPs/tree/master/CIP-0025)) was
+insecure in some example use cases. This is due to the link between the transactional metadata and the minting of native
+assets. For example, a smart contract that cannot/does not validate against transaction metadata (i.e. Liquidity Pool
+tokens) could have a malicious user inject CIP-25 metadata into the transaction, potentially inserting illicit, illegal,
+or otherwise nefarious metadata information tied to the tokens which may be picked up and displayed to end users 
+unwittingly by explorers and wallets.
```
KtorZ, line 48: Yes, but not _only_. The real security concern regard metadata spoofing. CIP-0025 supposes that EVERY token minting will follow CIP-0025 as a metadata process. If a project decides not to, then there's nothing preventing some user to provide its own metadata as-if the project was CIP-0025 compliant. 

Which means that fundamentally, there's a need for a meta-metadata which informs what standard does a project actually follow. I am sensing that this is possibly CIP-88?

CIP-XXXX/README.md
```diff
@@ -0,0 +1,447 @@
...
+Catalyst Voter Registration). At the root level, the registration consists of two pieces of data:
+
+1. A Token Registration Payload Object [(TRPO)](#token-registration-payload-object)
+2. A Token Registration Witness array [(TRWA)](#token-registration-witness-array)
+
+### Token Registration Payload Object
+
+The Token Registration Payload Object (TRPO) consists of 4 required fields and then optional, additional fields to
+provide additional context and information as part of the registration payload. For the purposes of this draft CIP and
+in the examples below we will use the top-level JSON Metadata Index of **867** although this could be changed to the
```
KtorZ, line 79: May I suggest to not use JSON for the specification, but actually use CBOR diagnostic since it better reflects what is stored on-chain. I'll make a concrete code suggestion for the relevant bits.

CIP-XXXX/README.md
```diff
@@ -0,0 +1,447 @@
...
+If a script requires multiple signatures (multi-sig) then the same validation logic should be applied when verifying
+the authenticity of the registration.
+
+**Example:**
+
+`'00000002df633853f6a47465c9496721d2d5b1291b8398016c0e87ae'`
+
+#### 2. Feature Set
+
+The Feature Set should consist of an array of CIPs that should be applied to the tokens under the given policy. For
```
KtorZ, line 101:
```suggestion
The Feature Set should consist of an array of CIP numbers that should be applied to the tokens under the given policy. For
```

CIP-XXXX/README.md
KtorZ: I think this section needs to be expanded a bit indeed. Possibly settle on a few "well-known" methods and give concrete examples of using those authentication methods. 

One big question is how does one link that back to the original token metadata? How do we prevent someone from claiming being someone else token's owner? Do we rely on client applications to then perform manual verification of the witnesses for tokens they support?

CIP-XXXX/README.md
```diff
@@ -0,0 +1,447 @@
...
+CIP-36, and stake pool KES counter rotation, the nonce is expected to be an integer value that is incremented as new or
+updated registrations occur. The suggestion to use the current slot tip of the network at the time of submission from 
+CIP-36 is also recommended here.
+
+Validators and indexers observing token registrations should always honor the verified registration with the highest
+nonce value.
+
+**Example:**
+
+`12345`
```
KtorZ, line 134: Question: can the witness method change between two registrations? If yes, what prevents an attacker from registering with a higher nonce and a completely different witness method that they control?

CIP-XXXX/README.md
```diff
@@ -0,0 +1,447 @@
...
+
+##### CIP-26: Fungible Tokens / Monetary Policy
+
+See [CIP26](CIP-26) for a description of a fungible token specific registration ([CIP26.json](CIP-26/CIP26.json) as an 
+example, [CIP26.schema.json](CIP-26/CIP26.schema.json) for schema documentation). This information can replace the information
+currently housed in the [Cardano Token Registry](https://github.com/cardano-foundation/cardano-token-registry) and is 
+based on the format currently used in those registrations along with a few additional fields.
+
+##### CIP-27: Cardano NFT Royalty Information
+
```
KtorZ, line 180: Is this (☝️ ) an empty section or, is the following header "Inline Datum / Reference ..." meant to be nested under "CIP-27: ..." ?

CIP-XXXX/README.md
```diff
@@ -0,0 +1,447 @@
...
+##### Inline Datum / Reference UTXO Registration
+
+Depending on the needs of a specific project or CIP, the CIP-Specific registration should include a reference pointer to
```
KtorZ, line 183: What is the format for the pointer? From the example below, it seems to be an address from the outputs and.. a hash of some sort?

Wouldn't it be sufficient to simply store a hash of the inline datum as a pointer? Application digesting this will have to crawl through the list of outputs anyway and are very much likely capable of calculating hash digests.

CIP-XXXX/README.md
```diff
@@ -0,0 +1,447 @@
...
+        [
+          "this is the second witness for a 2 out of 3 multi-sig script.",
+          "it's also broken up into an array.",
+          "strings in the witness array should be concatenated as-is",
+          "withouts spaces or any formatting applied"
+        ]
+      ]
+    ]
+  }
+}
```
KtorZ, line 341:
```suggestion
```cbor
{
  867: {
    1: {
      1: h'00000002df633853f6a47465c9496721d2d5b1291b8398016c0e87ae',
      2: [
        25,
        27
      ],
      3: [
        "Ed25519"
      ],
      4: 12345,
      5: "https://oracle.mycoolnftproject.io/",
      6: {
        25: {
          1: {
            0: "Cool NFT Project",
            1: [
              "This is a description of my project",
              "longer than 64 characters so broken up into a string array"
            ],
            2: [
              "https://",
              "static.coolnftproject.io",
              "/images/icon.png"
            ],
            3: [
              "https://",
              "static.coolnftproject.io",
              "/images/banner1.jpg"
            ],
            4: 0,
            5: {
              "twitter": [
                "https://",
                "twitter.com/spacebudzNFT"
              ],
              "discord": [
                "https://",
                "discord.gg/spacebudz"
              ]
            },
            6: "Virtua Metaverse"
          }
        },
        27: {
          0: "0.05",
          1: h'0003ee65bb9e4ec90b99d6af6998fe94028a6817b6ea80ebb723bd6ef6f966d78e1be122a9305fbf963cf796ac4f0e0ada602335ad1f0c75e0'
        }
      }
    },
    2: [
      [
        h'e97316c52c85eab276fd40feacf78bc5eff74e225e744567140070c3',
        [
          "witness may be broken up into a sub-array of 64-character strings",
          "assuming that the overall length of the witness is longer than the",
          "64-character on-chain metadata limit for strings."
        ]
      ],
      [
        h'26bacc7b88e2b40701387c521cd0c50d5c0cfa4c6c6d7f0901395757',
        [
          "this is the second witness for a 2 out of 3 multi-sig script.",
          "it's also broken up into an array.",
          "strings in the witness array should be concatenated as-is",
          "withouts spaces or any formatting applied"
        ]
      ]
    ]
  }
}
```

This would be the CBOR diagnostic equivalent, which can be directly encoded as binary using appropriate tools (e..g https://cbor.me/?bytes=A1190363A201A601581C00000002DF633853F6A47465C9496721D2D5B1291B8398016C0E87AE02821819181B038167456432353531390419303905782368747470733A2F2F6F7261636C652E6D79636F6F6C6E667470726F6A6563742E696F2F06A21819A101A70070436F6F6C204E46542050726F6A65637401827823546869732069732061206465736372697074696F6E206F66206D792070726F6A656374783A6C6F6E676572207468616E203634206368617261637465727320736F2062726F6B656E20757020696E746F206120737472696E6720617272617902836868747470733A2F2F78187374617469632E636F6F6C6E667470726F6A6563742E696F702F696D616765732F69636F6E2E706E6703836868747470733A2F2F78187374617469632E636F6F6C6E667470726F6A6563742E696F732F696D616765732F62616E6E6572312E6A7067040005A26774776974746572826868747470733A2F2F7818747769747465722E636F6D2F73706163656275647A4E465467646973636F7264826868747470733A2F2F74646973636F72642E67672F73706163656275647A0670566972747561204D6574617665727365181BA20064302E30350158390003EE65BB9E4EC90B99D6AF6998FE94028A6817B6EA80EBB723BD6EF6F966D78E1BE122A9305FBF963CF796AC4F0E0ADA602335AD1F0C75E0028282581CE97316C52C85EAB276FD40FEACF78BC5EFF74E225E744567140070C38378417769746E657373206D61792062652062726F6B656E20757020696E746F2061207375622D6172726179206F662036342D63686172616374657220737472696E67737842617373756D696E67207468617420746865206F766572616C6C206C656E677468206F6620746865207769746E657373206973206C6F6E676572207468616E20746865783136342D636861726163746572206F6E2D636861696E206D65746164617461206C696D697420666F7220737472696E67732E82581C26BACC7B88E2B40701387C521CD0C50D5C0CFA4C6C6D7F090139575784783D7468697320697320746865207365636F6E64207769746E65737320666F7220612032206F7574206F662033206D756C74692D736967207363726970742E78226974277320616C736F2062726F6B656E20757020696E746F20616E2061727261792E7839737472696E677320696E20746865207769746E6573732061727261792073686F756C6420626520636F6E636174656E617465642061732D69737829776974686F75747320737061636573206F7220616E7920666F726D617474696E67206170706C696564)

CIP-XXXX/README.md
```diff
@@ -0,0 +1,447 @@
...
+
+This makes this CIP backwards compatible with all existing standards (CIP-25, 26, 27, 68, etc) while also providing the
+flexibility for future-proofing and adding additional context and information in the future as additional use cases,
+utility, and standards evolve.
+
+## Path to Active
+
+### Acceptance Criteria
+
+Ideally this CIP would receive feedback, criticism, and refinement from the community of people involved in and with
```
KtorZ, line 434: Let's remove "Ideally" and make it an actual requirement for activation.

The proposal shall also be adopted in major explorers and/or wallets; as well as adopted by various tokens projects.

CIP-XXXX/README.md
```diff
@@ -0,0 +1,447 @@
...
+  }
+}
+```
+
+## Rationale
+
+For this specification, I have drawn inspiration from 
+[CIP-36: Catalyst/Voltaire Registration Metadata Format](https://github.com/cardano-foundation/CIPs/tree/master/CIP-0036)
+which succinctly and canonically publishes data to the main chain (L1) via a metadata transaction and without any
+required modification or customization to the underlying ledger with regard to Native Assets.
```
KtorZ, line 420: CIP-0036 does however tie together some stake credentials with some voting key. The stake credentials act in this scenario as strawman identity. My major question regarding CIP-0088 is how to prevent spoofing and how the witness is actually connected back to the original metadata owner (which is not necessarily the same entity as the one minting the tokens!).

CIP-XXXX/CIP-25/CIP25.schema.json
KtorZ, line 1: Given that this describes an on-chain format, JSON schemas are inappropriate here and we should instead provide CDDL specifications for the corresponding data. It's okay(-ish) to provide some high-level examples in JSON although the CBOR diagnostic notation is arguably as-good-as JSON here given that all object keys are already numbers anyway. 

I can provide some help to translate those schemas to CDDL if needs be.

rphair:
@Crypto2099 time to rename the containing folder to `CIP-0088` as well 🥳

CIP-XXXX/README.md
Crypto2099: That is correct, CIP-88 aims to address the need to explicitly signal support for CIP-25 so that explorers ***may*** refuse to display any CIP-25 metadata found on-chain if a CIP-88 registration is not found.

CIP-XXXX/README.md
Crypto2099: This was added as a preliminary thought of a way to point explorers in the right direction of an on-chain inline datum. Just thinking it through a bit more, it would probably be better to use something like a "beacon token" that could contain inline datum data and so the identifier of that beacon token could be contained within CIP-88 registration and subsequently used by explorers or SCs that need the data housed there.

CIP-XXXX/CIP-25/CIP25.schema.json
Crypto2099: Thank you :)

I'm familiar with writing JSON schemas so that is what I defaulted to while aiming at a CBOR-compatible format. 

The intent would be to publish both as part of the CIP in order to provide support for various toolings for creating the payload (i.e. a JSON schema validator that can then be converted to CBOR for submission)

CIP-XXXX/README.md
Crypto2099: The goal here would be to provide additional context and information (metadata) to the token policy itself. So, I would argue, the holder of the "keys" (or entity responsible for deploying the smart contract) ostensibly does own this particular metadata while things like the metadata attached to a specific token may fall under the "ownership" of an individual holder of that token.

The primary goals being to increase and decentralize discoverability of token projects with centralized ownership and reduce the need for redundant metadata (ledger bloat) information commonly found in currently published metadata standards due to a lack of an over-arching or top-level standard.

In this case, I argue that the signing keys of the Native Script policy constitute sufficient proof of ownership of the project to be "trusted" until such time as a better standard for signaling and/or transferring ownership and control of a native asset policy exists.

CIP-XXXX/README.md
Crypto2099: Thanks for taking the time to discuss and help me see the benefits of describing the schema first in CDDL format to help better inform the JSON data structure so that we have a consistent data transport between both on-chain and JSON representation! Next push should include a first draft of the CDDL spec.

CIP-XXXX/CIP-25/CIP25.schema.json
Crypto2099: Had added CDDL Spec v1.0.0 to the CDDL directory and updated examples in the main README.md to utilize CBOR notation rather than JSON.

CIP-XXXX/README.md
Crypto2099: All examples have been updated to use CBOR-notation format except for the CIP-specific documentation in their respective directories.

CIP-XXXX/README.md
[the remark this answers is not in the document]
Crypto2099: CDDL v1.0.0 spec has been published noting that signature payloads should be arrays of byte strings rather than strings.

CIP-XXXX/README.md
Crypto2099: @fallen-icarus you do need to maintain the inline datum if you want to reference the information within a SC validation logic step which is what @michele-nuzzi was referring to I believe. Having read the Beacon Tokens CIP I don't see how they are an alternative to this approach as merely the presence (or lack thereof) does not provide any further detail or insight.

The intent here is to create a backwards and forwards compatible standard that can be applied even to "locked" native script-minted tokens (and hopefully SC minted tokens in the future) although I believe the "discoverability" steps will be very similar to the methodologies proposed in your Beacon Tokens.

CIP-XXXX/README.md
Crypto2099: I have revised this to reference a token that would be used to "pointer" to an inline datum txn (currently only present in the CIP-26 specific content). This way the inline datum on the txn hosting the reference token could be updated w/o issuing an updated policy registration.

CIP-XXXX/README.md
fallen-icarus: > you do need to maintain the inline datum if you want to reference the information within a SC validation logic step ... I don't see how they are an alternative to this approach as merely the presence (or lack thereof) does not provide any further detail or insight.

If you want to reference the datum in another SC then yes, it does need to be maintained on-chain. I misunderstood that goal. However, that doesn't change my conclusion. Currently, CIP0025 recommends attaching metadata to the same transaction where the token forge occurs. This means that the metadata can only ever be associated with that token. To update the information, you just mint the token again with new metadata attached to the transaction. The downsides to this approach are:
1. The metadata format cannot be enforced.
2. If the information is needed for another SC, it isn't available since it is in the metadata.

Using an inline datum with a beacon token solves both of these issues. IMO The concern of having to maintain a datum for as long as necessary is not an issue:
1. Will it cause blockchain bloat? - No more so than using metadata since the information is identical between what would be in the datum and what would be in the metadata.
3. Will it cause a lot of locked ADA since UTxOs must have a minimum value? - The minimum value is probably only a few ADA (storing the beacon token with the inline datum in the UTxO) and this can be changed with protocol parameter changes. I don't think a few ADA is a big deal for registration. **This deposit is only necessary if the inline datum must be referenced in other SC exectutions.** If the datum does not need to be referenced, it can immediately be consumed (and beacon token burned) and the deposit can be recovered.

The benefits to this approach I think make up for having to use a deposit of a few ADA.  For this example, if the datums must be referenced by other SCs, all registration UTxOs could be locked at a specific SC address dedicated to registrations. The purpose of this SC is only to ensure the beacon tokens are burned whenever the datum is removed from the UTxO set. See [cardano-reference-scripts](https://github.com/fallen-icarus/cardano-reference-scripts) for an example. Since the datum is attached with the beacon, users can easily look up the UTxO TxIx and reference it in their own transactions.

If the datum is never needed during SC executions, then there is no need to keep the inline datum on-chain. The last datum can always be found by looking at the last mint transaction for the native token and finding the output that had the beacon token. This is effectively the same as what CIP0025 currently suggests.

With this datum/beacon approach, the presence/absence of a beacon token absolutely conveys information about the datum. You can make it so that the beacon can only be minted if the beacon token is being stored with the proper datum (and at the proper address if needed). As said before, if a datum is missing a required field, registration would automatically fail since the beacon token would fail to mint unless the datum format is followed. This means that if a UTxO is present with a datum but is missing the corresponding beacon token, then that registration is guaranteed to not be valid. Other smart contracts that use these datums can therefore look for the presence of the beacon token to ensure validity of the registration datum.

In short, using datums and beacon tokens does everything the metadata approach currently does but also guarantees the registration format is followed and allows other smart contracts to use the registration information in their executions. Using metadata labels and metadata cannot give you the same level of guarantees about the format nor can they allow other SCs to use them.

Crypto2099: Note that the pubkey/signature method is precisely what I am proposing specifically for the purpose of backwards compatibility and maintaining the ability to change things (like the project name or social media handles which are not necessarily part of the "what" of an NFT) even after the policy has locked.

I've actually renamed a reference in CIP-26 specific info to "Beacon/Reference Token" as I do think this method has merit but if the Beacon Token standard gains traction and popularity I would view them more as complementary to solve two different tasks rather than contrary or accomplishing the same.

The "Beacon Token" approach could be viewed as analogous to our original "Royalty Token" as proposed and co-authored by me back in 2021. i.e. mint a "nameless" token under the policy (native script) or contract (Plutus script) to provide a verifiable proof of ownership of the policy.

The only modification here is inserting the ability to use just the signing keys attached to the policy in order to make updates even past the timelock period of the script itself for non-mission-critical information (and of course the benefits to FT projects).

fallen-icarus: What about datum vs metadata? I do see these as accomplishing the same thing since the datum and metadata can carry the exact same information. You can convert this entire CIP (including the pubkey signature as ID) from using metadata to using datums and have the exact same functionality + enforcing the format and being able to reference the info in SC executions. Why would you still rather use metadata?

Crypto2099: If the goal is not to maintain this information permanently via an inline datum (and most of the data is unnecessary for SC execution logic) then, as you say, they are basically two different, albeit with the same outcome, approaches to publish the data to the chain.

Monitoring the chain for a specific metadata key index and taking some downstream action is already trivial thanks to tools like [Oura](https://github.com/txpipe/oura). The (to be written) methodology for monitoring and validating these registrations will attempt to make use of as many currently existing, open source solutions as possible to minimize the hassle and burden on the part of third parties.

To your point, the goal of this top-level specification is not to inject data that is usable by smart contracts directly but to provide context to the scope (in this case tokens) within the purview of [CPS-0001](https://github.com/cardano-foundation/CIPs/pull/371).

Since the majority of the data housed in this specification is unnecessary to SC validation logic, it is, in my opinion, preferable (if there is a need to inject data via inline datum to a smart contract) to provide a pointer to a specific "Beacon Token" within the sub-context of this registration.

In this way, this specification lays out the process for discovery, validation, and indexing of information that can then utilize further CIPs (such as beacon token) to inject context-specific information where appropriate.

As mentioned directly in your Beacon Token CIP Pull Request, beacon tokens as described in the CIP seem to provide a framework methodology for utilizing the tokens in dApps rather than a standard for the format of the data. I see these two CIPs as complementary rather than opposed.

CIP-XXXX/README.md
Crypto2099: @Ryun1 @rphair I have added reference to CPS-0001 in the latest push to update the readme document in the "Motivation" section. I have also amended the CDDL slightly after reviewing and absorbing CPS-0001 to provide flexibility in "scoping" CIP-88 in the future to support additional "scopes" beyond tokens should there be community interest in such.

CIP-0088/cddl/v1.0.0.cddl
```diff
@@ -0,0 +1,135 @@
...
+    1 : string,                 ; royalty rate, floating point number (0.000000 - 1.000000)
+    2 : [* string]              ; bech32 address to receive royalties
+}
+
+; CIP-0060: Music NFT Metadata Standard
+; Version: 1.0.0
+cip60_details = {
+    ? 0 : string                ; version
+    1 : music_token_details     ; CIP-0060 Music NFT Details
+}
```
Crypto2099, line 91: @AndrewWestberg would love some feedback from yourself and the co-authors of CIP-60 if you believe there is any information that may stand to benefit from this "higher level" registration to reduce data duplication and save bytes in individual token metadata.

CIP-0088/README.md
```diff
@@ -0,0 +1,845 @@
...
+
+##### CIP-60: Music NFT Standard
+
+`Version: 1.0.0`
+
+The author of this CIP lacks a technical understanding of the many informational fields specific to the music industry
+which are found in CIP-60. The co-authors of CIP-60 are invited and asked to help in defining which fields may be better
+suited to a top-level informational scope.
+
+**TODO: Expand Support for CIP-60**
```
Crypto2099, line 543: @AndrewWestberg also here w.r.t. CIP-0060.

CIP-XXXX/README.md
rphair: Since CPS-0001 is now covered, and the only alternative designs I was aware of (https://github.com/cardano-foundation/CIPs/pull/467#issuecomment-1471481889) appear not to be significant in the community, @Ryun1 I think this question is resolved (you can mark it if so).

CIP-0088/README.md
```diff
@@ -0,0 +1,845 @@
...
+## Abstract
+
+Currently, token projects (NFT or FT) have no mechanism to register the intent, details, or feature set of their minting
+policy on-chain. This CIP will aim to create a method that is backwards compatible and enable projects to declare, and
+update over time, the supported feature set and metadata details pertaining to their tokens.
+
+This CIP will aim to make use of a hybrid (on and off-chain) information schema to enable maximum flexibility and
+adaptability as new and novel use cases for native assets expand and grow over time.
+
+## Motivation
```
rphair, line 26:
```suggestion
## Motivation: why is this CIP necessary?
```

CIP-0088/README.md
```diff
@@ -0,0 +1,845 @@
...
+          "assuming that the overall length of the witness is longer than the",
+          "64-byte on-chain limit."
+        ]
+      ]
+    ]
+  }
+}
+```
+
+## Rationale
```
rphair, line 804:
```suggestion
## Rationale: how does this CIP achieve its goals?
```

rphair (commented):
Everything about the structure & outline looks great to me but it also now seems important that we use the originally conceived [verbose titles](https://github.com/cardano-foundation/CIPs/blob/master/.github/CIP-TEMPLATE.md) for the _Motivation_ and _Rationale_ sections.

I'd hoped we might scrape them in through GitHub CI and web site builders, but as it seems now the most attention to these CIPs is paid is _before_ they are merged, when newer members of our community often need to see exactly what these sections mean.

CIP-XXXX/README.md
fallen-icarus: > I see these two CIPs as complementary rather than opposed.

I think you are misunderstanding my comments. I am in no way saying that the Beacon Token CIP does what I am suggesting above. You are entirely correct in that the Beacon Token CIP is just a framework. What I am saying here is to apply that framework to this CIP. There is currently no CIP doing what I am suggesting here. 

> Monitoring the chain for a specific metadata key index and taking some downstream action is already trivial thanks to tools like [Oura](https://github.com/txpipe/oura). The (to be written) methodology for monitoring and validating these registrations will attempt to make use of as many currently existing, open source solutions as possible to minimize the hassle and burden on the part of third parties.

If you mean to say that the beacon token framework requires more work on the part of third parties, this is not true. All third party tools like Oura can already be used this way. There is no difference between getting the information from the metadata of the transaction vs getting the information from a specific datum in that transaction's output.

The biggest point that I think you are missing is that **metadata should not be used for registration standards because there is no way to force people to ACTUALLY adhere to the format**. How are you stopping people from registering with different/missing fields? Even if you filter out bad formats after the fact with off-chain parsing, that metadata is still taking up space on the chain. Improperly formatted metadata is just useless space.

The way I suggested doesn't just allow for interoperability with smart contracts. That is just a bonus. Using a datum forces all registrations to adhere to the format that you are creating with this CIP. **It wouldn't even be possible to post to the blockchain with improperly formatted information.** Adherence to the format is guaranteed simply by using a datum instead.  And as stated before, the datum does not need to remain on-chain for it to be accessible.  A datum is literally just JSON that gets attached to a UTxO rather than a transaction. Your format in this CIP would remain the same. But since the datum *IS* visible by scripts, you can have a beacon minting policy only mint when the datum is the proper format. The successful mint is effectively the blockchain's stamp of approval that this registration adhered to the format.

Do you not think trustlessly forcing the adherence to a standard is just as important as the standard itself? IMO a standard that naturally enforces itself is superior to one that does not.

> Since the majority of the data housed in this specification is unnecessary to SC validation logic, it is, in my opinion, preferable (if there is a need to inject data via inline datum to a smart contract) to provide a pointer to a specific "Beacon Token" within the sub-context of this registration.

This would make the information on the blockchain completely redundant which is wasted space. Since a SC cannot see metadata, you would need to copy all of the information in the metadata and post it to the chain *again* but this time as a datum just so that a smart contract can see it. Meanwhile, if it was just posted to the chain as a datum in the first place, this redundant information would not be needed.

Jack-0:
I would just like to comment I proposed meta for the metadata before in this [PR](https://github.com/cardano-foundation/CIPs/pull/343). Happy for that to be deprecated by this CIP. My proposal had each asset use an string array to define extensions that define behaviour. But one meta-meta asset to define this makes more sense.

Crypto2099:
> I would just like to comment I proposed meta for the metadata before in this [PR](https://github.com/cardano-foundation/CIPs/pull/343). Happy for that to be deprecated by this CIP. My proposal had each asset use an string array to define extensions that define behaviour. But one meta-meta asset to define this makes more sense.

Thanks Jack, I've been browsing your PR a little bit but haven't had a chance to leave any comments yet. I don't think these are necessarily mutually exclusive but I will try to see where there may be overlap or synergies between the two.

CIP-XXXX/README.md
Crypto2099: I believe you continue to misunderstand the intent here and my differentiation between "SC-scoped" data versus "off chain" information.

1. Publishing this data as transaction metadata versus an inline datum in a regular, simple transaction are equally error prone. The only solution, as far as I'm aware, would mean that a smart contract would need to be developed, deployed, and subsequently maintained (as new versions of this standard were developed over time) in order to **"force adherence to the format"**.
2. Utilizing a smart contract to force adherence to the data standard would break a large majority of the intentional flexibility designed into the standard as written. For example, a project may have CIP-25, CIP-27, CIP-60, and CIP-XX support or any combination of the above. All of which may have additional auxiliary data structures which may have one or more required or optional fields to provide additional information for off-chain purposes. As far as I'm aware and I do stand to be corrected, this would require creating a custom endpoint in the aforementioned smart contract for every possible combination of data that **may** be present in the token or requiring that **all** fields be present (even if blank/empty strings).
3. As mentioned before, 99% of the information here is only relevant to the "off chain" display and formatting of information contained in other/subsequent sources. I will provide some concrete examples of what I mean and envision as a benefit of using the "beacon token" to benefit on-chain, in-contract interactions vs. the data potentially contained within the registration data:

A (Proposed) CIP-25 or CIP-68 "NFT Project" information block is defined as follows:
  ```cbor
  {
    25: {
      0: "1.0.0",
        1: {
          0: "Cool NFT Project",
          1: [
            "This is a description of my project",
            "longer than 64 characters so broken up into a string array"
          ],
          2: [
            "https://",
            "static.coolnftproject.io",
            "/images/icon.png"
          ],
          3: [
            "https://",
            "static.coolnftproject.io",
            "/images/banner1.jpg"
          ],
          4: 0,
          5: [
            [
              "twitter",
              [
                "https://",
                "twitter.com/spacebudzNFT"
              ]
            ],
            [
              "discord",
              [
                "https://",
                "discord.gg/spacebudz"
              ]
            ]
          ],
          6: "Virtua Metaverse"
        }
     }
}
```
Literally no part of this data structure would ever be relevant within the context of a smart contract validation execution and would therefore not benefit from being present on a "beacon token" that pointed to an inline datum.

However, something like CIP-27 (NFT Royalties) could benefit from being housed via an inline datum and referenced to a beacon token with the inline datum being utilized by a smart contract to validate that the correct rate of royalties was being paid to the correct address. The changes to CIP-27 necessary to facilitate this are beyond the scope, but rather than defining the rate and address for CIP-27 during a policy registration, we could point to a "beacon token" that would contain those data via inline datum.

Example (as currently written):

```cbor
{
  27: {
      0: "1.0.0",
      1: "0.05",
      2: ["addr...xyz27"]
    }
}
```

This information is only (as written) useful during off-chain validation and logic whereas this information could stand to benefit from its presence as an inline datum which is discoverable via a beacon token, changing the proposed solution to something similar to:

```cbor
{
  27: {
    0: "1.2.3",
    1: [
      h'myBeaconPolicyIdHere',
      h'myBeaconAssetIdHere'
    ]
  }
}
```

In the context of the challenge proposed by CPS-0001, I am actively working on increasing the future flexibility of this specification to support additional contexts such as stake pools, addresses, and more. Again, in most of these cases the most important part is data being recorded on the chain, not the availability of the data within the context of a smart contract as the data must still be validated to come from an authentic source above and beyond simply being recorded in the correct format or not.

Finally, it is much more readily accessible to search for transaction metadata thanks to the storage formatting of tools like db-sync and built-in triggers in tools like Oura that allow us to filter first by only metadata containing a particular key (867 in this case) and then taking further query action. The (current) state of stored inline datums is not nearly as readily available without having prior knowledge of the precise UTXO or "beacon token" that we are seeking.

Example (DB-Sync Transaction Metadata):

| id | key | json | bytes |
| --- | --- | --- | --- |
| 403867 | 867 | {"0": "1.0.0", "1": {"1": [0, "0x00000002df633853f6a47465c9496721d2d5b1291b8398016c0e87ae"], "2": [26], "3": ["Ed25519"], "4": 12345, "5": ["https://", "oracle.tokenproject.io/"], "6": {"26": {"0": "1.0.0", "1": [{"0": ["d894897411707efa755a76deb66d26dfd50593f2e70863e1661e98a0", "7370616365636f696e73"], "1": "spacecoins", "2": ["the OG Cardano community token", "-", "whatever you do, your did it!", "", "Learn more at https://spacecoins.io!"], "3": "SPACE", "4": 0, "5": ["https://", "spacecoins.io"], "6": ["ipfs://", "bafkreib3e5u4am2btduu5s76rdznmqgmmrd4l6xf2vpi4vzldxe25fqapy"], "7": [["ipfs://", "bafkreibva6x6dwxqksmnozg44vpixja6jlhm2w7ueydkyk4lpxbowdbqly"], "3507afe1daf05498d764dce55e8ba41e4acecd5bf42606ac2b8b7dc2eb0c305e"], "8": ["d894897411707efa755a76deb66d26dfd50593f2e70863e1661e98a0", "7370616365636f696e74"]}]}}}, "2": [["0xe94ecf4caec67c7aa3b4da713c711df1ff0e923c83a34eda4f004721", "0x2df777de7cd2348bd16a5528b070e78341802c0baf0388762bb8b49ca03721c2dd2d76392dcfb52a4f0aff63e440afd883ca43bfc5caa1fd39dc432ae0472a0f"]]} | \<bytestring gibberish here\> |

Example (DB-Sync Inline Datum):

| id | hash | tx_id | value | bytes |
| --- | --- | --- | --- | --- |
| 267738 | \<bytestring gibberish\> | 628552 | {"fields": [{"fields": [{"fields": [{"bytes": "3318ad8f0c743a674831d69e4c61c0d805eda695b0f20a4ff0edebc5"}], "constructor": 0}, {"fields": [{"fields": [{"int": 381754}, {"int": 1679818283579}], "constructor": 0}], "constructor": 0}], "constructor": 0}], "constructor": 1} | \<bytestring gibberish\> |

Subsequently, I can easily query for all Token Policy Registrations with a given policy ID with a simple, fast, and responsive query by utilizing transaction metadata:

```sql
select *
from tx_metadata
where key = 867 
    and json->'1'->'1'->>0 = '0'
    and json->'1'->'1'->>1 = '0x00000002df633853f6a47465c9496721d2d5b1291b8398016c0e87ae'
```

Or perhaps we'd like to find all Token Policy Registrations that claim to support CIP-26 (Fungible Tokens)

```sql
select *
from tx_metadata
where key = 867 
    and json->'1'->'1'->>0 = '0'
    and json->'1'->'2' @> '26'
```

Whereas the query to find the same information when stored as an inline datum would be... Not nearly as user friendly :)

If you can provably show more performant and reliable means of accessing historical data from the ledger, I am happy to reconsider.

Jack-0:
I will just quickly summarise the NFT [metadata extension tag I proposed previously](https://github.com/cardano-foundation/CIPs/pull/343)

### tldr 
Add an optional `ext` property to new nfts if the behaviour of the NFT exceeds the limitations of cip25. (I proposed this to help me with [reference nfts cip48?](https://github.com/cardano-foundation/CIPs/pull/249) to overcome 16kb limits in metadata, and paradoxically/ironically it also helps reduce onchain data)

### Takeaways
- It's backwards compatible
- It assumes that **without** the `ext` property array the nft is a standard cip25 compatible nft
- It uses a `ext` array to define which cips are present in the metadata
- It is embedded in each NFT's meta (but reading this cip a single tx would be more efficient, though it would restrict the policy to conform to the defined standard. I was thinking of the usecase in which the 'client' only makes one request.

### Notes
My CIP for a metadata extension tag assumes the 'client' will request the nft / tx hash from a database and will know from a single request that said nft if a X from the metadata tag if it exists else it's a standard CIP25 NFT. CIP88 would require 2 requests, and all current implementations of fetching NFTS to be changed to look for an 867 transaction. The benefit of CIP88 is less duplicated on change data.

CIP-XXXX/README.md
fallen-icarus: I am going to reply to your comments out of order.

>  the query to find the same information when stored as an inline datum would be... Not nearly as user friendly

I believe this is false as long as the beacon token framework is used. I am not as familiar with db-sync itself to give you a specific query to execute however, I am familiar with both [Koios](https://www.koios.rest/) and [Blockfrost](https://blockfrost.dev/) that use db-sync as their backends. 

As an example, I will use Koios. The example assumes that you are using a registration beacon token that can only ever be minted in a registration transaction and the token name of the beacon token specifies what entity is being registered. You first use this [api](https://api.koios.rest/#get-/asset_history) to get the most recent mint transaction for that specific beacon token. From that response, you take the tx hash and use that in this [api](https://api.koios.rest/#post-/tx_utxos) to get all of the tx inputs and outputs. You only care about the outputs here. You filter the outputs for the UTxO that is holding the freshly minted beacon token in question. The inline datum attached to that UTxO is your registration data (it was returned as part of the previous response). From experience with my PoCs that do very similar things (using Blockfrost), the target inline datum is returned in about 1-2 seconds. I hardly think this counts as "user unfriendly". Based on my understanding of SQL, I see no reason why these two api queries + filtering can't be combined into one step which would make it even faster.

To your point about most of the information not being relevant for SC executions, I disagree again. When you create tools, it is impossible to know in advance how those tools will be used. Just because that information is useless to you does not mean it is useless to someone else. Using your NFT Project example, I think this *would* be useful in smart contract executions that deal with loyalty tokens. For example, the company of Discord can give a loyalty token to anyone who owns an NFT that is registered to use a valid discord server. You can trustlessly use a SC (created by Discord) for this where a user can try minting a loyalty token to themselves by referencing the UTxO with the NFT they own and the UTxO with that NFT's registration data. If the registration has a discord server registered, then the loyalty token can be minted to that user without any input from Discord the company. Discord can also make you provide a third input UTxO (owned by Discord the company) that has a datum of all valid discord servers currently registered to NFT projects to weed out fake servers; to mint a loyalty token, the server registered would also need to be in this third UTxOs datum.

As another example, one of the PoCs I am working on is an NFT marketplace where royalties are enforced by the Cardano settlement layer. There would literally be no way to avoid paying the royalties. Using beacon tokens and registration in datums, it would be possible for users to reference the royalty information for a given token when interacting with the marketplace. My current design is effectively doing this. This would allow the royalty information to be updated without needing to change the marketplace. Also, when a new version of the marketplace is introduced, creators can say in their royalty information which versions of the marketplace they consent to using. Users owning the NFT would only be allowed to use the versions in the royalty information. **This is all only possible if the royalty information is in a datum.**

To your point about flexibility, the design I have in mind is modular. There would be one beacon token for each registration form. If there are multiple tiers to a registration, then each tier would also have their own beacon token. This means users can choose which information to register and can register for multiple CIPs in one tx. Also, CIPs can be updated independently without impacting the other forms. Also the user can choose whether to leave this information on-chain for use in SCs. The user can always immediately consume the UTxO if they wish and the information would still be accessible to everyone else for off-chain use.

To your point about smart contract maintenance and deployment, there really isn't much maintenance since changing any of the code would result in an entirely new beacon token. What I am suggesting here is for CIPs that deal with registrations with datums to also provide a smart contract for everyone to use. If a bug is discovered, the smart contract that is stored with the CIP can be updated very quickly. Changes are made to current CIPs fairly frequently. Whenever a new version of a form is proposed, that proposal must also include the new smart contract for people to use. BTW, these smart contracts are pretty small in size. From my experience with the PoCs, they are only about 100 lines of code.

> I am actively working on increasing the future flexibility of this specification to support additional contexts such as stake pools, addresses, and more.

Since the beacon tokens are modular, new specifications can easily be added with this approach too. The limit for the number of registrations that can be made in a single transaction is likely at least 10. This seems like more than enough for now. This limit will grow as Cardano's tx limits grow.

In the case of optional fields, this is actually easy to do with datums thanks to the `Maybe` Haskell type (or `Optional` type in other languages). If the field isn't required, it is okay to leave it blank. It would basically just appear as null in the JSON.

Finally, in regards to whether or not this is part of the scope of this CIP, again I stress that it is. Until now (with the beacon token framework) it was not simple enough to store registration information in a datum. But now it is which means, going forward, the community needs to actively discuss the pros and cons of using a datum instead of transaction metadata. The community can now decide if the trade-off profile of one is better suited for a given task than the other. However, since you have explicitly mentioned metadata everywhere in your CIP, you have implicitly taken a stand without explaining why in your CIP. If you actually want this discussion to be out-of-scope for your CIP (which is fine if you do) then you need to change it to be agnostic to whether the information is stored in a datum or metadata.

To be fair, as said before, being able to use datums is new. I apologize if you feel like I am singling you out given that all the other registration CIPs solidly deal with metadata. We all want what is best for Cardano and I am trying to spread the word about the possibility of using datums now. AFAICT, this discussion is the very first time anyone in the community has even remotely considered the idea. IMO, all previous CIPs should now be revisited with this datum vs metadata debate as well. Going forward, since there are now two different ways of doing the same thing, it is up to CIP authors to explain why they have chosen the option they chose.

CIP-0088/README.md
```diff
@@ -0,0 +1,867 @@
...
+
+***Reference Fields***
+
+| index | name    | type   | required | notes                                                                                                                 |
+|-------|---------|--------|----------|-----------------------------------------------------------------------------------------------------------------------|
+| 0     | Name    | String | Yes      | A case sensitive path identifier for this reference                                                                   |
+| 1     | Type    | Enum   | Yes      | An enum of accepted types of references which may include direct payloads or pointers to other sources of information |
+| 2     | Payload | Object | Yes      | A "Reference Payload" object containing the information to be substituted in place of the reference                   |
+
+**TODO: Expand Support for CIP-48**
```
Crypto2099, line 578: @Jack-0 see here for addressing the potential for supporting CIP-48 reference data via CIP-88 :)

CIP-XXXX/README.md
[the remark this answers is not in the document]
Crypto2099: Marking this as resolved as examples and CDDL have been updated to use binary strings rather than hex strings.

CIP-0088/README.md
rphair: Also asked here (https://github.com/cardano-foundation/CIPs/pull/249#discussion_r1150384616) if this approach now described in CIP-0088 now leaves this proposal deprecated:

- https://github.com/cardano-foundation/CIPs/pull/343

CIP-0088/cddl/v1.0.0.cddl
Crypto2099: CIP-60 specific CDDL content has been moved to CIP-0088/cddl/CIP-60_v1.0.0.cddl

rphair:
@Crypto2099 I've just done an editorial overhaul of @Jack-0's CIP-0048 to clear the way towards merging according to your suggestion https://github.com/cardano-foundation/CIPs/pull/467#discussion_r1149937589.  I can see you have this item here:
> TODO: Expand Support for CIP-48

so for the time being we'll check that these two CIPs are mutually compatible & please (both authors) let us know if that ever changes.  cc @KtorZ

thaddeusdiamond:
Well reasoned and well-written @Crypto2099.  Thanks for putting this together.  I only have two main thoughts:

1. Should the `nonce` be implied by the block slot height? While I hate "implicit" cache busting, I also worry about an accidental cache mis-deployment that causes a bunch of dependent dApps to break.  Since presumably this entry into the decentralized token registry will take the form of a datum/UTxO combination, that will have an implicit livelihood based on the tx slot height it was locked at.
2. Any way to incorporate [CIP-0071](https://developers.cardano.org/docs/governance/cardano-improvement-proposals/cip-0071/) here? I would love to register my votes as separate from my main collection, but have them linked.  That way, JPG could natively pick up the NFT type of the policy without having to have me explain what each type is and put out a description.

GeorgeFlerovsky:
@Crypto2099 
I noticed that your CIP has implications/interactions for CIP-25.
Our CIP-86 extends CIP-25, so perhaps your CIP might also have corresponding effects on CIP-86.

I'll take a closer look at your CIP to see what those might be.

cc @nicolasayotte @nalane

BrockCruess (changes requested):
Small changes, widening the focus on CIP-25/68 from only tokens with a total mint count of 1 to include FTs/RFTs. Can await CIP editor meeting 74.

rphair:
@BrockCruess @Crypto2099 since our agendas have not been full recently I've put this PR on the agenda for Review again at the next (03 October) meeting (since https://github.com/cardano-foundation/CIPs/pull/467#pullrequestreview-1640013385 overlaps with #593 already on agenda): https://hackmd.io/@cip-editors/74

CIP-0088/CIP-25/README.md
```diff
@@ -0,0 +1,257 @@
...
+}
+```
+
+### Index 6 (Project Name)
+
+##### Type: String | Required: No | Default: Policy Name
+
+For projects that belong under an over-arching umbrella (such as Claymates, Virtua, etc.) where a single "project" may
+have multiple NFT policies or "collections", this field allows a space for the "Project Name". This could also be the
+name of the "artist" if an artist creates NFTs under multiple different policies.
```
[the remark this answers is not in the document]
rphair: I don't know if the term "mint their art" is term most commonly used but I agree with this change in theory.

Crypto2099:
> @Crypto2099 I noticed that your CIP has implications/interactions for CIP-25. Our [CIP-86](https://github.com/cardano-foundation/CIPs/tree/master/CIP-0086) extends CIP-25, so perhaps your CIP might also have corresponding effects on CIP-86.
> 
> I'll take a closer look at your CIP to see what those might be.
> 
> cc @nicolasayotte @nalane

Hey George, thanks for reminding me on this one, I'm getting ready to publish some updates to hopefully get to a "nearly ready" for deployment and integration work standard so I'll make sure this one is also included. :)

CIP-0088/CIP-25/README.md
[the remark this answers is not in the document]
Crypto2099: I have attempted everywhere that CIP-25 or CIP-68 are mentioned to change the language to reflect a "token project" rather than an "NFT" project specifically.

Crypto2099:
I believe, given these latest changes, that the standards contained herein are now at a point that I can begin some publication and verification testing on test net to confirm that everything works as advertised. Unless there is a glaring or obvious error or omission pointed out by community contributors, I will consider this an "official" v1 of the specification. Appreciate any review or feedback from @rphair @Ryun1 @SmaugPool @thaddeusdiamond @yHSJ and any others who may feel the desire. :)

CIP-0088/README.md
```diff
@@ -0,0 +1,556 @@
+---
+CIP: 88
+Title: Token Policy Registration
+Category: Tokens
+Status: Proposed
+Authors:
+- Adam Dean <adam@crypto2099.io>
+Implementors:
+-
```
rphair, line 9:
```suggestion
Implementors: NA
```
... just noticed in the meantime (apologies for quoting set theory) that this is not a null set of implementors: it's a set with a single null implementor 🤓
...

rphair (commented):
thanks @Crypto2099; looking forward to a full re-read...

CIP-0088/README.md
```diff
@@ -0,0 +1,556 @@
+---
+CIP: 88
+Title: Token Policy Registration
```
rphair, line 3: ⬆️

rphair (commented):
> @Crypto2099 https://github.com/cardano-foundation/CIPs/pull/467#issuecomment-1817035594: the standards contained herein are now at a point that I can begin some publication

Before that happens, would you retitle this PR to the the same as the concise (but still complete I think) document title... to avoid difficulty & possible divergence when discussing in the community?

CIP-0088/README.md
Crypto2099: Done

CIP-0088/README.md
Crypto2099: This was a temporary solution in the interim as I intend to list working tools and interfaces here on the "Path to Active" :) Can change to an empty array for the time being if it makes merging easier

rphair: We just can't merge it as-is because it needs to both be valid YAML (which it is already) but also not have a blank implementor... which wouldn't break the site builders but would look like something was off in the metadata. 🤔

CIP-XXXX/README.md
rphair: @fallen-icarus @Crypto2099 @michele-nuzzi this has been a great discussion, maybe to be continued here or elsewhere... I'm just resolving this now because I believe it's been addressed in the CIP according to https://github.com/cardano-foundation/CIPs/pull/467#discussion_r1131923835 above.

CIP-XXXX/README.md
rphair: @Crypto2099 @nullhashpixel looks like you have addressed this question, so resolving.

CIP-XXXX/README.md
rphair: @KtorZ @Crypto2099 I'm resolving this because the purpose of CIP-88 seems explicitly covered in this section; please feel free to re-open if people still think it needs clarification.

CIP-XXXX/README.md
rphair: @Crypto2099 resolving since section heading no longer exists (maybe absorbed into table)

CIP-XXXX/README.md
rphair: resolving since fixed in earlier revision.

CIP-0088/cddl/v1.0.0.cddl
rphair: @Crypto2099 @AndrewWestberg resolving since no objections from CIP-60 subject matter expert.

CIP-0088/README.md
rphair: resolving as per https://github.com/cardano-foundation/CIPs/pull/467#discussion_r1405595192

CIP-0088/README.md
rphair: resolving: fixed in rewrite

rphair:
p.s. have added this to next CIP meeting https://hackmd.io/@cip-editors/77 if only to assign/confirm `Last Check` status (cc @Ryun1 @Crypto2099)

Ryun1 (approved):
Excited to see this on-chain 🤠

cent-development:
> Yes, backward compatibility seems impossible to achieve for Plutus minting policies if we cannot figure out a mechanism that doesn't include the need for identity token minting.
> 
> I agree that the reference UTxOs may be nice feature to make use of in this solution. Throwing out a couple of ideas that also can be investigated (NFT guild is already looking into this and we can help of course as we are eager to participate in introducing token verification to the community:) ):
> 
> 1 - Minting contract owner can mint token called ID (or something that can act as standard) that has the 867 metadata attached either as Datum or tx metadata. (This is perhaps the option you mentioned?) 2 - Mint token called `ID-<ownerpubkeyhash>` or some other prefix if owner wants another name `<random-name>-<ownerpubkeyhash>`.
> 
> Information added to the 867 metadata is the token name as a value in the Witness Method (element 3) and the owner pub key hash as usual in Witness Array (element 7).
> 
> Verification in both options can be done by simply verifying the token is minted using the correct minting policy and that it is named according to the 867 metadata.

Some further thoughts on the smart contract verification: To work around the 32 char limitation for token names, we have  the possibility to mint ID-tokens named as the MD5 hash of each of the owners pub key hashes as MD5 is always 32 char length. Minting owner identity tokens in this way will give us the benefit of proving ownership of the smart contract as we are able to mint tokens from it, but this also has the nice feature that contract owners are able to add/remove valid owners of a policy by simply minting new owner tokens or burning old tokens for member addresses that leave the team(!).
This solution is also backward compatible for existing smart contracts as long as a contract doesn't enforce specific names for tokens minted by it.

cent-development:
Please also consider the problem with scam tokens now seen in the community. I cannot see that this new standard will tackle this issue, or have I missed something? If a scam "project" decides to mint tokens, my hypotheses is that if they follow this standard in the current form, it will not be possible to directly flag a token as a scam as they can publish metadata containing the fake pub keys and fake policy ids on chain in a legal format.

One possible way to fight scam tokens is to add a requirement to the standard that a file containing the pub keys are placed on a web server on the same web domain as the project and that the NFT metadata contains a link to this file.
Scam projects will not be able to place their pub key hashes on legitimate project web sites, and this will then make it impossible (or at least very difficult) for scam tokens to appear legit. They will also not be able to reuse legit pub keys in their metadata as this will cause their metadata signatures to be noticeably fake. 

Suggestion is to add a property for this path, or make use of the existing Oracle property for this (with some additions to the property description), that contains the link to this file with keys. It must reside on the project domain web site. 

You might say that not all projects have a web domain, but most do. It is of course still OK to not have this, but then the project tokens should be verified to status "yellow - not completely verified" as we will not be able to 100% verify the legitimacy. Projects that have a web domain with a valid pub key hash file, will be verified to status "green - verified"

PR #469, opened 2023-03-02
CIP-0092? | First-class errors in Plutus

L-as:
@michaelpj @effectfully

[Rendered in branch](https://github.com/L-as/CIPs/blob/first-class-errors/CIP-XXXX/README.md)

Initial discussion: https://github.com/cardano-foundation/CIPs/pull/459#issuecomment-1443904223

### DIFF
--- a/CIP-XXXX/README.md
+++ b/CIP-XXXX/README.md
@@ -124,6 +124,11 @@ is precisely the example we seek to make work.
 `f True ()` will fail as expected,
 while `f False ()` will return `()` as expected.
 
+### Backward compatibility
+
+This CIP has the issue of old scripts not being portable to the new version trivially.
+You need to make sure your script isn't making use of any impure assumptions.
+
 ## Path to Active
 
 ### Acceptance Criteria
--- a/CIP-XXXX/README.md
+++ b/CIP-XXXX/README.md
@@ -25,7 +25,7 @@ We propose three changes:
 - Erase type abstractions and type applications entirely when going from
   TPLC to UPLC rather than turning them into `delay` and `force`.
 
-## Motivation: why is this CIP necessary?
+## Motivation
 
 Users write code in the style of `if cond then error else False`, which is unidiomatic and also problematic,
 since the seemingly equivalent `cond && error` will **always err**.
--- a/CIP-XXXX/README.md
+++ b/CIP-XXXX/README.md
@@ -66,6 +66,10 @@ As noted in the abstract, we can now also do optimisations on the UPLC that were
 before, as the effects mattered. This notably meant we previously couldn't regard code which result
 wasn't used as dead code. With this change, it's possible to entirely elide such (now) dead code.
 
+One minor effect of this CIP is that evaluation strategy is less important, meaning that you
+could theoretically take a script in this new language and evaluate it lazily, without breaking
+anything.
+
 ## Specification
 
 The changes below would need to take effect in a new ledger era and Plutus version.
@@ -140,6 +144,20 @@ f = λ b -> ifThenElse b ERROR (λ x -> x)
 is precisely the example we seek to make work.
 `f True ()` will fail as expected,
 while `f False ()` will return `()` as expected.
+
+### Soundness of optimisations
+
+We note that you can soundly optimise `(\x -> M) N` into `x` is not mentioned in `M`,
+or similar occurrences where the result of `N` isn't needed.
+Such an optimisation will change the behaviour of the script necessarily,
+as any traces in `N` will no longer happen, yet, this is not important in practice
+as traces are merely a debugging tool. The budget used will decrease,
+but people do not rely on budget exhaustion to signal invalidity.
+
+Transformations like `M || N` into `force (ifThenElse M (delay True) (delay N))` follow a similar principle:
+traces that happened before might no longer happen.
+The used budget will increase if M is `False`, yet this is likely
+still an acceptable transformation for most people.
 
 ### Backward compatibility
 
--- a/CIP-XXXX/README.md
+++ b/CIP-XXXX/README.md
@@ -25,6 +25,19 @@ We propose three changes:
 - Erase type abstractions and type applications entirely when going from
   TPLC to UPLC rather than turning them into `delay` and `force`.
 
+UPLC is now **truly pure** wrt. transaction validation.
+This enables another use-case: We can **optimise** it as if it
+were entirely pure. We can turn `const () x` into `()` for any `x`!
+We can turn `expensive1 || expensive2` into `if expensive1 then True else expensive2`,
+and in general do any transformations that we'd do on a pure language without
+regard for side-effects.
+
+Issues fixed / PRs closed by this CIP:
+- https://github.com/input-output-hk/plutus/issues/4826
+- https://github.com/input-output-hk/plutus/issues/4114
+- https://github.com/input-output-hk/plutus/pull/4118
+- https://github.com/cardano-foundation/CIPs/pull/459
+
 ## Motivation
 
 Users write code in the style of `if cond then error else False`, which is unidiomatic and also problematic,
@@ -48,6 +61,10 @@ toolchains, but also means that all built-ins will no longer need to be forced b
 
 Making Plutus lazy would also solve the problems, yet this has not been done as a strict language
 is easier to reason about wrt. budgetting. Instead, we turn to making errors first-class values.
+
+As noted in the abstract, we can now also do optimisations on the UPLC that were impossible
+before, as the effects mattered. This notably meant we previously couldn't regard code which result
+wasn't used as dead code. With this change, it's possible to entirely elide such (now) dead code.
 
 ## Specification
 
--- a/CIP-XXXX/README.md
+++ b/CIP-XXXX/README.md
@@ -1,0 +1,165 @@
+---
+CIP: ?
+Title: First-class errors in Plutus
+Category: Plutus
+Status: Proposed
+Authors:
+    - Las Safin <me@las.rs>
+Implementors:
+    - Las Safin <me@las.rs>
+Discussions:
+    - https://github.com/cardano-foundation/cips/pulls/?
+Created: 2023-03-01
+License: CC-BY-4.0
+---
+
+## Abstract
+
+Plutus is a strictly evaluated language, where users heavily rely on **impure** errors
+to cause transaction failure to rule out invalid cases.
+
+We propose three changes:
+- Make `error` a first-class value you can pass around. Using it does not inherently
+  cause an error.
+- Change condition for script success from not erring to returning `True`.
+- Erase type abstractions and type applications entirely when going from
+  TPLC to UPLC rather than turning them into `delay` and `force`.
+
+## Motivation: why is this CIP necessary?
+
+Users write code in the style of `if cond then error else False`, which is unidiomatic and also problematic,
+since the seemingly equivalent `cond && error` will **always err**.
+This is not a good situation, and issues have been opened about it before,
+notably https://github.com/input-output-hk/plutus/issues/4114 (over a year ago!).
+The issue was righly not deemed a "vulnerability" or "bug", yet it is quite clearly a misfeature.
+
+The current design in fact **encourages impurity**, as the return value of scripts is ignored entirely;
+only the effects matter. This is entirely antithetical to the design of Plutus.
+With the changes in this CIP, users will naturally return `False` in cases that are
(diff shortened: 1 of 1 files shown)


L-as:
Reply to https://github.com/cardano-foundation/CIPs/pull/459#issuecomment-1450854400 :

I hope the behaviour is more obvious given the explanation in this CIP?
The `factorial` implementation blows up because everything is still evaluated strictly.
The `ifThenElse` example doesn't because passing an error to it is entirely valid.

L-as:
I realised another benefit: This allows optimising `expensive1 || expensive2` into `if expensive1 then True else expensive2`, the reason being that the language is not _truly pure_ (modulo traces, which we don't care about in optimisations).
This is evident by the fact that evaluating something doesn't matter wrt. transaction validity (modulo budgets, but optimisation already messes with that), so we can in fact in general entirely eliminate unused terms with this change.

We also fix the other issue I noted in https://github.com/input-output-hk/plutus/issues/4826, since `const` can now truly ignore the error!

This might be another (quite big?) reason to introduce this change to Plutus.

effectfully:
@L-as 

> I hope the behaviour is more obvious given the explanation in this CIP?
> The `factorial` implementation blows up because everything is still evaluated strictly.
> The `ifThenElse` example doesn't because passing an error to it is entirely valid.

I understood the proposed behavior, but I was failing to understand the motivation for it. Now that I've read this CIP I do indeed understand the motivation, thanks a lot for writing it out.

I think I agree that doing what you suggest would be an improvement over the status quo. I wouldn't like the resulting system, but less so than the current one.

However to be completely frank, I believe chances of anybody being able to convince our academics to give up on canonicity and to convince our managers to overhaul the whole evaluation semantics of the language (i.e. the implementation, the spec and the formalization) are pretty slim. As such, I don't feel like it's a battle worth fighting. Plus, if we were to overhaul the whole evaluation semantics, I would prefer some other solution, perhaps something based on call-by-push-value. Or maybe even just making the language lazy.

Overall, I think that ship has sailed, unfortunately.

L-as:
> I believe chances of anybody being able to convince our academics to give up on canonicity

I don't see why it would be any different from the status quo wrt. canonicity.

> I would prefer some other solution, perhaps something based on call-by-push-value. Or maybe even just making the language lazy.

I think we will likely find a better solution in the future, however, this is a very minor change that can be done without changing most of the ecosystem.

> to overhaul the whole evaluation semantics of the language (i.e. the implementation, the spec and the formalization)

Well that's precisely what it isn't, I'd say! This is a minor change to the implementation, the spec, and the formalisation.

mgajda:
Haskell and GHC had introduced pretty major overhaul to its errors and introduced more complex exceptions. So may there be a chance to do _everything in a monad_?

CIP-XXXX/README.md
```diff
@@ -0,0 +1,205 @@
+---
+CIP: ?
```
```suggestion
CIP: 92
```

L-as:
What do I need to do to get this in now with the new structure?

CIP-XXXX/README.md
```diff
@@ -0,0 +1,205 @@
...
+CIP: 92
+Title: First-class errors in Plutus
+Category: Plutus
+Status: Proposed
+Authors:
+    - Las Safin <me@las.rs>
+Implementors:
+    - Las Safin <me@las.rs>
+Discussions:
+    - https://github.com/cardano-foundation/cips/pulls/?
```
rphair, line 11:
```suggestion
    - https://github.com/cardano-foundation/CIPs/pull/469
```

CIP-XXXX/README.md
```diff
@@ -0,0 +1,205 @@
...
+were entirely pure. We can turn `const () x` into `()` for any `x`!
+We can turn `expensive1 || expensive2` into `if expensive1 then True else expensive2`,
+and in general do any transformations that we'd do on a pure language without
+regard for side-effects.
+
+Issues fixed / PRs closed by this CIP:
+- https://github.com/input-output-hk/plutus/issues/4826
+- https://github.com/input-output-hk/plutus/issues/4114
+- https://github.com/input-output-hk/plutus/pull/4118
+- https://github.com/cardano-foundation/CIPs/pull/459
```
rphair, line 39: (not a format correction, but an editorial question)

1. This is unclear whether these Plutus issues/PRs are closed to to this CIP method being followed (please confirm if so... I can't verify this from the issue/PR threads) or because their solutions share the premise of the CIP.  Can you change "closed by this CIP" to a statement or paragraph that explains the relationship unambiguously?
2. Once we decide where this is going we can also link @nielstron about the stalled https://github.com/cardano-foundation/CIPs/pull/459 to discuss whether candidate CIP-0091 should also be progressed (or deprecated).

CIP-XXXX/README.md
```diff
@@ -0,0 +1,205 @@
...
+  TPLC to UPLC rather than turning them into `delay` and `force`.
+
+UPLC is now **truly pure** wrt. transaction validation.
+This enables another use-case: We can **optimise** it as if it
+were entirely pure. We can turn `const () x` into `()` for any `x`!
+We can turn `expensive1 || expensive2` into `if expensive1 then True else expensive2`,
+and in general do any transformations that we'd do on a pure language without
+regard for side-effects.
+
+Issues fixed / PRs closed by this CIP:
```
rphair, line 35: This list is too much detail for the _Abstract_ and in fact is _Motivation_ so should be moved down to that section.

CIP-XXXX/README.md
```diff
@@ -0,0 +1,205 @@
...
+- https://github.com/input-output-hk/plutus/issues/4826
+- https://github.com/input-output-hk/plutus/issues/4114
+- https://github.com/input-output-hk/plutus/pull/4118
+- https://github.com/cardano-foundation/CIPs/pull/459
+
+## Motivation
```
rphair, line 41:
```suggestion
## Motivation: why is this CIP necessary?
```
(standard as per CIP-0001)

CIP-XXXX/README.md
```diff
@@ -0,0 +1,205 @@
...
+This CIP has the issue of old scripts not being portable to the new version trivially.
+You need to make sure your script isn't making use of any impure assumptions.
+
+## Path to Active
+
+### Acceptance Criteria
+
+It would be needed to implement the necessary changes in `cardano-ledger` and `plutus`.
+
+### Implementation Plan
```
rphair, line 173:
```suggestion
### Acceptance Criteria

It would be needed to implement the necessary changes in `cardano-ledger` and `plutus`.

### Implementation Plan
```
We need more detail about how the both the Ledger & Plutus teams would be expected to verify this.  I was hoping to use the similarly split https://github.com/cardano-foundation/CIPs/pull/758 as an example, but that too has been stalled & is incomplete in that regard.

@zliu41 @lehins can you please provide some help about what exactly would be needed here for verification of this proposal?

CIP-XXXX/README.md
```diff
@@ -0,0 +1,205 @@
...
+
+#### Implementation of built-ins
```
rphair, line 175: This is a proposed specification, so should be moved into that section (probably with a heading reinforcing your first statement that this method is hypothetical and/or illustrative).

CIP-XXXX/README.md
```diff
@@ -0,0 +1,205 @@
...
+toBuiltinMeaning DefaultFunVX IfThenElse =
+  makeLazyBuiltinMeaning
+    (\b x y -> b >>= \case True -> x ; False -> y)
+    (runCostingFunThreeArguments . paramIfThenElse)
+```
+
+## Copyright
+
+[CC-BY-4.0]: https://creativecommons.org/licenses/by/4.0/legalcode
+[Apache-2.0]: http://www.apache.org/licenses/LICENSE-2.0
```
rphair, line 205:
```suggestion
## Copyright

This CIP is licensed under [CC-BY-4.0].

[CC-BY-4.0]: https://creativecommons.org/licenses/by/4.0/legalcode
[Apache-2.0]: http://www.apache.org/licenses/LICENSE-2.0
```
we do need to choose _one_ of these to display in the markup: I'm going to not suggest Apache since the code herein isn't intended for reuse.

rphair (commented):
@L-as thanks for attending to this one.  All changes below are to make the format & content consistent with newer CIPs (with 1 or 2 other suggestions beyond the formatting).

effectfully:
I wrote this last year:

> I understood the proposed behavior, but I was failing to understand the motivation for it. Now that I've read this CIP I do indeed understand the motivation, thanks a lot for writing it out.

I wish I spelled out what convinced me, because rereading the CIP I'm not convinced that it would be worth doing that even given infinite resources for development. I was probably just sloppy last year.

One of the proposed changes is this:

> Erase type abstractions and type applications entirely when going from TPLC to UPLC rather than turning them into `delay` and `force`.

But that just can't work. Take the following TPLC expression:

```
let factorial n = ifThenElse {forall a. integer} (\{_} -> n == 0) 1 (\{_} -> factorial (n - 1)) {()}
in factorial 2
```

If we erase all type abstractions and type applications from it while going from TPLC to UPLC, we'll get

```
let factorial n = ifThenElse (n == 0) 1 (factorial (n - 1))
in factorial 2
```

which is an infinite loop as the proposal acknowledges, unlike the original expression.

The proposal says

> Very few people will make their own `loopforever` function to replicate the old `error` to write impure scripts.

but if you start erasing type abstractions and type applications left and right you'll get those `loopforever` functions all over the place without bothering to put them into the program manually.

@L-as

> I don't see why it would be any different from the status quo wrt. canonicity.

You're proposing to add an axiom that every type is inhabited by the special `ERROR` construct, it's a textbook example of how to break canonicity, see e.g. [this](https://ncatlab.org/nlab/show/canonical+form#canonicity_vs_axioms).

Can we have a formal model where that rule isn't an axiom, but rather an inherent part of the system, like "every type has an `ERROR` constructor, including even `->`" and then elaborate TPLC/UPLC into that system? Maybe, but I doubt anybody is going to do it. It's a major change to the semantics of our languages. And having implemented and recently refactored the way we do error handling in the UPLC evaluator, I can assure you it's plenty of work, so I very much disagree with this:

> Well that's precisely what it isn't, I'd say! This is a minor change to the implementation, the spec, and the formalisation.

If you believe it's a minor change anyway, we'd appreciate a branch implementing the proposed changes to handling of `ERROR`. We'll implement erasure of `force`s and `delay`s ourselves just to meet you halfway.

CIP-XXXX/README.md
effectfully: I've commented down below, I think this proposal can't work as written, so it can't be implemented and verified either. I may be wrong of course, so let's see what further discussion shows.

CIP-XXXX/README.md
effectfully: > Once we decide where this is going we can also link @nielstron about the stalled https://github.com/cardano-foundation/CIPs/pull/459 to discuss whether candidate CIP-0091 should also be progressed (or deprecated).

Michael proposed to merge it as Proposed and I don't see why we should do anything else to it. It seems reasonable, even though I highly doubt we're ever going to implement it, given that a very simple workaround exists.

rphair: @effectfully I think since this is doable, but not likely to be done, it _would_ make sense to put it through as Proposed after some cleanup from @L-as... rather than assign the currently not-yet-used `Rejected` status that e.g. Michael proposed for stalled candidate https://github.com/cardano-foundation/CIPs/pull/336 (though a `Rejected` status could still be considered if everyone is in agreement about that).

effectfully: @rphair sorry, I was commenting about #459 specifically. If my suspicion regarding this CIP not being doable is confirmed (if!), then this CIP is a prime candidate for being rejected.

L-as:
Thanks, that’s a very concrete path forward. Maybe I’ll look at it when I have time.

rphair:
@L-as the last comment (after a pessimistic review + a still-unaddressed request to update to the canonical document structure), now that a couple more months have gone by, suggests that you don't have the time to pursue this.

Ordinarily we would close it when an author doesn't have time to complete a CIP but I don't want to bury the work you & @effectfully have done to document this issue... so in the meantime I'm tagging this `Abandoned` which will close it later... if & when editors don't see any progress as apparently agreed after https://github.com/cardano-foundation/CIPs/pull/469#issuecomment-2158542726.

L-as:
Thank you. I wouldn't necessarily say this will never happen, you should have a long-term view. I'm busy right now with some other open source stuff but when I have that done it's possible I will do this.

L-as:
Rip

PR #471, opened 2023-03-02
CIP-0068 | Clarify label prefixes

mateusap1:
By reading the CIP-0068 and trying to implement it myself I was convinced the asset name should literally be prefixed with the label `(100)` and not with the implementation suggested by CIP-0067, there was no reference whatsoever to the correct label convention as far as I am aware and this PR tries to make things more clear for future developers attempting to implement CIP-0068 on their own.

### DIFF
--- a/CIP-0068/README.md
+++ b/CIP-0068/README.md
@@ -40,7 +40,7 @@ Lastly and most importantly, with this construction, the metadata can be used by
 
 ### Labels
 
-Each asset name must be prefixed by a label. The purpose of this label is to identifty the purpose of the token. For example, a reference NFT is identified by the label 100 and so every token considered a reference NFT should start its asset name with the the hex `000643b0`. This is following [CIP-0067](https://github.com/cardano-foundation/CIPs/tree/master/CIP-0067), which specifies how the label prefix should be formatted.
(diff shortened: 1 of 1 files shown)


alessandrokonrad:
Thanks @mateusap1. This was indeed not very obvious. It wasn't clear back then how the labels will actually look like and then it was forgotten to specify them correctly in CIP-0068. However I think as a general convention and for convenience e.g. `(100)` could be a high level representation of an actual CIP-0067 label.

mateusap1:
@alessandrokonrad Thanks for your quick reply.

> However I think as a general convention and for convenience e.g. (100) could be a high level representation of an actual CIP-0067 label.

I made a comment at the end of the "Labels" section trying to pass that idea. It reads as follows

> For simplicity purposes, the document will use the label `(100)` or `(<label>)` in the following documentation, but understand it should follow the CIP-0067 specification.

Let me know if this satisfies your concern or if you still have any suggestions / issues.

rphair (commented):
Looks to me like the added text clarifies what was posted in comments so far, so I'd approve this as long as it satisfies @alessandrokonrad's latest comment & has no objections from @SmaugPool etc.

alessandrokonrad:
@rphair yes this LGTM

rphair (approved):

KtorZ (approved):

PR #474, opened 2023-03-09
CPS-???? | Reading metadata in Plutus contracts

matiwinnetou:
([rendered proposal in branch](https://github.com/matiwinnetou/CIPs/blob/CPS--metadata-as-reference-inputs/CPS-%3F%3F%3F/README.md))

### DIFF
--- a/CPS-???/README.md
+++ b/CPS-???/README.md
@@ -1,0 +1,31 @@
+---
+CPS: ?
+Title: Reading transaction metadata in Plutus contracts
+Status: Open
+Category: Plutus
+Authors:
+    - Mateusz Czeladka <mateusz.czeladka@cardanofoundation.org>
+Proposed Solutions: 
+    - reading metadata via reference inputs
+
+Created: 2023-03-09
+---
+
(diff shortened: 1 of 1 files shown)


L-as:
You can theoretically already metadata as its part of the tx hash which is part of TxOutRef. You can only read other transactions' metadata though, not your own. If you use a ZKP you can also get around the size issue.

KtorZ:
Closing this pull request for the following reasons:

- The lack of access to transaction metadata within Plutus script is a calculated design choice from the Plutus team; `datum` are meant to replace transaction metadata when it comes to on-chain programmability. 

- The current document is framed as a "problem statement" although it is hinting strongly at only one possible solution: reading metadata in Plutus contracts. So it feels more like a feature request than an actual problem. If it is indeed a problem, then what the problem is and why current alternatives are unsuitable?

- Overall, the proposal is _too succinct_ and lack details. As just stated, if there's indeed a problem, we strongly recommend expanding more on that, describing use-cases with more details etc.

PR #478, opened 2023-03-15
Update top-level README with latest statuses.

KtorZ:


rphair (commented):
I just noticed that today's promoted candidates 88 through 92 didn't have GitHub's label `Candidate CIP` assigned.  Looks like the newest PR that had this label applied was opened in August 2022... has this label become passé?  If not needed or applied anymore, should we remove it? (since it currently applies to only to a subset of candidates)

README.md
```diff
@@ -109,7 +109,6 @@ Below are listed tentative CIPs still under discussion with the community. They
 | 73? | [Oracle Datum Standard](https://github.com/cardano-foundation/CIPs/pull/357) |
 | 76? | [Hash-Checked Data](https://github.com/cardano-foundation/CIPs/pull/363) |
 | 77? | [Verified Stake Pool Identity](https://github.com/cardano-foundation/CIPs/pull/361) |
-| 78? | [Extended Local Chain Sync Protocol](https://github.com/cardano-foundation/CIPs/pull/375) |
 | 79? | [Implement Ouroboros Leios to increase Cardano throughput](https://github.com/cardano-foundation/CIPs/pull/379) |
 | 80? | [Transaction Serialization Deprecation Cycle](https://github.com/cardano-foundation/CIPs/pull/372) |
 | 81? | [Tiered Pricing Protocol](https://github.com/cardano-foundation/CIPs/pull/381)
```
rphair, line 114:
```suggestion
```
(re-submitting from above review since I applied it to the wrong section of the text)

All other stalled proposals have been removed from the _Proposals Under Review_ list, whether they've been assigned a number or not.

Maybe this list should include any tentatively assigned numbers (e.g. 0081 for _Tiered Pricing Protocol_: information which we would lose if the redundant table entry were removed)... so interested visitors don't have to dig it out of the pull request.

KtorZ:
> has this label become passé? If not needed or applied anymore, should we remove it? (since it currently applies to only to a subset of candidates)

I think so indeed; since we started more explicitly assigning numbers or else leaving `????` as numbers to CIP the label has become sort of redundant.

README.md
KtorZ: The follow-up question will be: for how long do we want to keep those around even after marking them as "stall"? In principle, this doesn't change the PR's link, so any past link to the proposal would still work.

KtorZ:
@rphair your last comment reminded me to create labels for categories as we said we would. So this is now done. I've also removed the old labels for Candidate CIP, and CIP's types (removed from CIP-0001 back in November 2022).

rphair (approved):

README.md
rphair: @KtorZ I would posit that the `Waiting for Author` tag on the PR is to get the author's attention, while appearance on the _Stalled_ list is to get the community's attention.  My feeling has always been we should close the PR when it appears (some time after tagging) that the author has lost interest... so likewise we should remove from _Stalled_ when it appears the community has lost interest. 🤔

PR #456, opened 2023-01-30
CIP-0084? | Ledger evolution (meta)

JaredCorduan:
This CIP aims to make the `Cardano ledger` a registered category of the CIP process.

---

([rendered proposal in branch](https://github.com/JaredCorduan/CIPs/blob/ledger-evolution/CIP-%3F%3F%3F%3F/README.md))

PR #482, opened 2023-03-21
Remove RSS category, use Ledger for CIP-0075 & CIP-0082

KtorZ:
I believe that this better reflects the latest discussions on the topic. But that is up-to-debate. We could keep the more specific "RSS" category and simply mark all CIPs directed towards this category as "Inactive"; this would ease their finding later on although there aren't so many "Ledger" CIPs either. 

It just has been bothering me that this category exists, is clearly targeted at some specific areas of the ledger, yet is not embracing the CIP process. Unlike other pre-registered categories that tend to have multi-tenant ownership (e.g. wallets).

### DIFF
--- a/CIP-0001/README.md
+++ b/CIP-0001/README.md
@@ -212,7 +212,6 @@ At present, we consider the following list of initial categories:
 Category               | Description
 ---                    | ---
 Meta                   | Designates meta-CIPs, such as this one, which typically serves another category or group of categories.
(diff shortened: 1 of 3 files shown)


rphair (approved):
sure, my memory wasn't serving me correctly because I'd though we already had `Ledger` as a `Category` for these 2 CIPs.

Also this appears to definitively settle the question raised here: https://github.com/cardano-foundation/CIPs/pull/456#issuecomment-1468065083

Ryun1 (approved):
Makes sense to me.

PR #484, opened 2023-03-21
CIP-0035 | Require proposals for new builtin types to explain how they will be (de)serialized

michaelpj:
As it says. We need to be able to do this in order to (de)serialize
scripts.


rphair (approved):
Approving off the bat because I cannot imagine any way this change could be disputed.

Ryun1 (approved):
LGTM - The Plutus team should be able to adjust their enrolment as they see fit.

michaelpj:
Can we merge this?

PR #486, opened 2023-03-23
CIP-0030 | Better describe API extension compatibility

stevenj:
Adds specification to better describe the "Extensions Enable" API, with an effort to better control backwards and forwards compatibility.
Define a way to specify experimental extensions that do not compete with official extensions, and describe the path forward.
Add language tags to code blocks.
Use `-` consistently for unordered lists.
Properly declare HTTP urls.

------------------------------------

[CIP-0030 with this update, rendered in branch](https://github.com/stevenj/CIPs/tree/cip-30-backward-compat/CIP-0030)

### DIFF
--- a/CIP-0030/README.md
+++ b/CIP-0030/README.md
@@ -169,13 +169,13 @@ APIError {
 
 ```ts
 DataSignErrorCode {
- ProofGeneration: 1,
- AddressNotPK: 2,
- UserDeclined: 3,
+	ProofGeneration: 1,
+	AddressNotPK: 2,
+	UserDeclined: 3,
 }
 type DataSignError = {
- code: DataSignErrorCode,
- info: String
+	code: DataSignErrorCode,
+	info: String
 }
 ```
 
@@ -197,12 +197,12 @@ type PaginateError = {|
 
 ```ts
 TxSendErrorCode = {
- Refused: 1,
- Failure: 2,
+	Refused: 1,
+	Failure: 2,
 }
 type TxSendError = {
- code: TxSendErrorCode,
- info: String
+	code: TxSendErrorCode,
+	info: String
 }
 ```
 
@@ -213,12 +213,12 @@ type TxSendError = {
 
 ```ts
 TxSignErrorCode = {
- ProofGeneration: 1,
- UserDeclined: 2,
+	ProofGeneration: 1,
+	UserDeclined: 2,
 }
 type TxSignError = {
- code: TxSignErrorCode,
- info: String
+	code: TxSignErrorCode,
+	info: String
 }
 ```
 
--- a/CIP-0030/README.md
+++ b/CIP-0030/README.md
@@ -100,7 +100,7 @@ Used to specify optional pagination for some API calls. Limits results to {limit
 
 #### Extension
 
-An extension is an object with a single field `"cip`"` that describes a CIP number extending the API (as a plain integer, without padding). For example:
+An extension is an object with a single field `"cip"` that describes a CIP number extending the API (as a plain integer, without padding). For example:
 
 ```json
 { "cip": 30 }
--- a/CIP-0030/README.md
+++ b/CIP-0030/README.md
@@ -61,7 +61,7 @@ This representation was chosen when possible as it is consistent across the Card
 
 #### DataSignature
 
-```
+```ts
 type DataSignature = {|
   signature: cbor\<COSE_Sign1>,
   key: cbor\<COSE_Key>,
@@ -81,7 +81,7 @@ transaction_unspent_output = [
 
 then we define
 
-```
+```cddl
 type TransactionUnspentOutput = cbor<transaction_unspent_output>
 ```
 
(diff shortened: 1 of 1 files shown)


stevenj:
This arose out of confusion related to CIP-62 and how to handle the API changes for extensions in a way that didn't break existing wallet implementations of the draft to cip-62.

For example, CIP-62 is still draft, so do we refer to it as `{"cip",62}` ?  But its subject to changes, so how does the dApp and wallet know what CIP-62 is being requested.

The previous `enable` method takes no parameters.  The new `enable` method takes 1 parameter which defaults to `{}` but what does `{}` mean?  Does that mean that if a dApp calls `enable()` with the no parameter nothing is enabled?  Not even CIP-30?  If so, then that will break all backwards compatibility with any dApp that implements the previous `enable` method and has not been updated.  So the specification I added here allows for backwards and forwards compatibility.  dApps which implement `enable()` will still keep working.  wallets which do implement the older `enable()` and current extensions like the draft CIP-62 will keep functioning and will still be compliant.

In a nutshell, this mostly consists of better documenting corner cases or previously undefined states.
It does not change the actual API surface.

CIP-0030/README.md
```diff
@@ -81,35 +81,74 @@ transaction_unspent_output = [
...
+```ts
 type Paginate = {|
   page: number,
   limit: number,
 |};
 ```
+
 Used to specify optional pagination for some API calls. Limits results to {limit} each page, and uses a 0-indexing {page} to refer to which of those pages of {limit} items each. dApps should be aware that if a wallet is modified between paginated calls that this will change the pagination, e.g. some results skipped or showing up multiple times but otherwise the wallet must respect the pagination order.
 
 #### Extension
```
Ryun1, line 101: IMO it would make extension discovery easier if we have a list of accepted extensions here in the CIP or an attached file to make it machine readable?

CIP-0030/README.md
```diff
@@ -81,35 +81,74 @@ transaction_unspent_output = [
...
-```
+```json
 { "cip": 30 }
 ```
 
+##### Draft or Experimental Extensions
+
+Extensions that are draft, in development, or prototyped should not use official "cip" numbering.
+
+They should instead be referred to as "x-cip" and the number can be arbitrary and should not be expected to be unique.
```
Ryun1, line 113: I think the term "x-cip" is somewhat ambiguous, I see no reason we could use something like "experimental" or "draft-cip".

CIP-0030/README.md
```diff
@@ -194,15 +234,60 @@ In order to initiate communication from webpages to a user's Cardano wallet, the
...
+This is to prevent a possibility where a dApp asks for basic extensions and then asks for an extension that the wallet would otherwise have prompted the user for permission to enable.
+Otherwise, if the extension list has not changed OR the wallet does not ask for permissions to enable any individual extensions, then this function should not request access a second time, and instead, just return the `API` object.
+
+Upon start, dApp can explicitly request a list of additional functionalities they expect as a list of CIP numbers capturing those extensions.
+This is used as an extensibility mechanism to document what functionalities can be provided by the wallet interface.
+CIP-0030 provides a set of base interfaces that every wallet must support.
+Then, new functionalities are introduced via additional CIPs and may be all or partially supported by wallets.
+
+If the extensions list is not specified, it defaults to an empty list `{}`.
+The wallet should treat an empty list as a request to enable **ALL** extensions which are not mutually exclusive.
```
Ryun1, line 254: By enabling all wallets do lose the ability to permission per extension.

Although nothing was stopping wallets from enabling all supported extensions regardless of what dApps were asking for - this would make wallet implementations simpler. Additionally, I am unsure of the real utility of permissioning access to extensions - I cannot think of a usecase.

CIP-0030/README.md
```diff
@@ -194,15 +234,60 @@ In order to initiate communication from webpages to a user's Cardano wallet, the
...
+*Note:* Extensions can be mutually incompatible.
+Incompatibilities can arise for example:
+
+- When extensions provide conflicting features.
+- If a newer extension wholly replaces the functionality of an older one.
+- A draft `"x-cip"` is replaced by a final `"cip"`.
+
+While conflicting extensions should be avoided where reasonable while designing CIPs, it is also the responsibility of wallet providers to assess whether they can support a given combination of extensions, or not.
+Wallets *MUST NOT* fail to enable the API should they not recognize or not support a particular combination of extensions.
+Instead, they should decide what they enable and reflect their choice in the [`cardano.{walletName}.extensions`](#cardanowalletnamesupportedextensions-extension) property of the Full API.
```
Ryun1, line 274:
```suggestion
Instead, they should decide what they enable and reflect their choice in the [`cardano.{walletName}.supportedExtensions`](#cardanowalletnamesupportedextensions-extension) property of the Full API.
```
typo

CIP-0030/README.md
```diff
@@ -194,15 +234,60 @@ In order to initiate communication from webpages to a user's Cardano wallet, the
...
+DApps will use the extensions reported by that property to decide how to proceed.
+For example, they can fail and inform their users that necessary wallet features are unavailable, or may use a different, less-efficient, strategy to cope with a lack of functionality.
+This may require the wallet to re-request a different set of extensions.
 
 ##### Can extensions depend on other extensions?
 
-Yes. Extensions may have other extensions as pre-requisite. Some newer extensions may also invalidate functionality introduced by earlier extensions. There's no particular rule or constraints in that regards. Extensions are specified as CIP, and will define what it entails to enable them.
+Yes. Extensions may have other extensions as pre-requisite. Some newer extensions may also invalidate functionality introduced by earlier extensions. There are no particular rules or constraints in that regard. Extensions are specified as a CIP, and will define what it entails to enable them.
```
Ryun1, line 281:
```suggestion
Yes. Extensions may have other extensions as pre-requisite, extension dependencies should be unambiguously defined in their respective CIPs. Some newer extensions may also invalidate functionality introduced by earlier extensions. There are no particular rules or constraints in that regard. Extensions are specified as a CIP, and will define what it entails to enable them.
```

CIP-0030/README.md
```diff
@@ -194,15 +234,60 @@ In order to initiate communication from webpages to a user's Cardano wallet, the
...
+
+Wallets should attempt to enable all extensions necessary to enable the extensions requested, even if they were not specifically requested.
+For example, if `CIP-9999` needs `CIP-8123` and a request is made to ONLY enable `[{"cip":9999}]` then the wallet should automatically enable `CIP-8123` as if the request was actually for `[{"cip":8123},{"cip":9999}]`.
+
+##### Can wallets enable more extensions than requested?
+
+Yes.
+It would be a legitimate implementation of this CIP for a wallet to **ALWAYS** enable every extension it supports, regardless of the requested list of extensions.
+The only caveat is that the [`cardano.{walletName}.extensions`](#cardanowalletnamesupportedextensions-extension) property must faithfully reflect the extensions which the wallet API object supports.
```
Ryun1, line 290:
```suggestion
The only caveat is that the [`cardano.{walletName}.supportedExtensions`](#cardanowalletnamesupportedextensions-extension) property must faithfully reflect the extensions which the wallet API object supports.
```
typo

CIP-0030/README.md
```diff
@@ -194,15 +234,60 @@ In order to initiate communication from webpages to a user's Cardano wallet, the
...
 
 ##### Should extensions follow a specific format?
```
Ryun1, line 292: Prospective extension CIPs should clearly outline and log their respective "x-cip" numbering/naming.

Ryun1 (commented):
Overall I quite like the goal and form of these additions.

This layouts a clear path for how extensions can be developed with backwards and forwards compatibility in mind. This aspect were perhaps overlooked when merging the last CIP-30 change in #446.

CIP-0030/README.md
stevenj: There is no security implication documented here in CIP-30 for enabling extensions.  Nor how such a flow would work.
There is also no signalling back to the dApp that it's security restricted vs just not available.
For example,  if a phone app asks to use the camera, it's told either "there is no camera" OR "you don't have permission to use the camera", which are different states to which the app can make different UX decisions.
If this is intended to be a security mechanism, then that should have proper consideration and specification.

CIP-0030/README.md
```diff
@@ -194,15 +234,60 @@ In order to initiate communication from webpages to a user's Cardano wallet, the
...
+The wallet should treat an empty list as a request to enable **ALL** extensions which are not mutually exclusive.
+The wallet may also enable extensions that it *prefers* for any extensions which would conflict.
+Conversely, it would then disable the non-preferred conflicting extensions.
 
-Upon start, dApp can explicitly request a list of additional functionalities they expect as a list of CIP numbers capturing those extensions. This is used as an extensibility mechanism to document what functionalities can be provided by the wallet interface. CIP-0030 provides a set of base interfaces that every wallet must support. Then, new functionalities are introduced via additional CIPs and may be all or partially supported by wallets.
+*Note:* If an extension list is supplied, extension `{"cip":30}` (or the preferred replacement in the case of a conflict) **MUST** always be enabled, even if it's not explicitly listed.
+This is because the [`cardano.{walletName}.extensions`](#cardanowalletnamesupportedextensions-extension) property of the API is necessary for the dApp to query what extensions are available for it to use.
```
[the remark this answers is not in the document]
stevenj: should be `api.getExtensions()`

CIP-0030/README.md
stevenj: shopuld be api.getExtensions()

CIP-0030/README.md
stevenj: Agree, wallets should list all extensions they support,  both experimental/draft and finalized CIPs.

CIP-0030/README.md
rphair: yes, when I saw the prefix `x-` that is exactly what I thought: based on the MIME convention.  So it seems intuitive & well founded to me.

Ryun1: Ahh, I didn't know about that :)

rphair: For a quick view, view full headers in any email message :nerd_face:

Ryun1:
Further community input on this would great if possible, tagging some relevant parties @Scitz0 @refi93 @Quantumplation @rooooooooob 😎

Scitz0:
I dont have much to add to the conversation but looks fine to me.

Ryun1:
I believe I have addressed the concerns raised by this proposal in the changes introduced in https://github.com/cardano-foundation/CIPs/pull/577, do you still want to pursue this update proposal @stevenj?

Ryun1:
Moving to close this, as similar changes were introduced in https://github.com/cardano-foundation/CIPs/pull/577

@stevenj please feel free to reopen if you desire.

PR #494, opened 2023-04-01
CIP-0068 | Add RFT Rich-fungible token support

AndrewWestberg:
Rich-fungible tokens sit somewhere between FTs and NFTs. They need the richness of metadata provided by CIP-25, but require decimals like FTs.

One example is a fractionalized NFT.  It could be represented by 100 tokens and setting decimals to 2 to show that there was only 1 NFT.

Alternatively, you could choose to mint 100m tokens and set decimals to 6. In this scenario, holding 1 token would represent holding 1% as there are 100 tokens in circulation.

A final example would be a card-style game where you have a number of the player cards as NFTs `(222)`, but interchangeable mana/land/utility cards would be semi-fungible tokens `(444)`

---

([updated CIP-0068 rendered in branch](https://github.com/AndrewWestberg/CIPs/blob/amw/cip68_semi_fungible/CIP-0068/README.md))

### DIFF
--- a/CIP-0067/registry.json
+++ b/CIP-0067/registry.json
@@ -16,7 +16,7 @@
   },
   {
     "asset_name_label": 444,
-    "class": "SFT",
+    "class": "RFT",
     "description": "CIP-0068 - Datum Metadata Standard (444 sub standard)"
-  }
+  },
 ]
--- a/CIP-0067/registry.json
+++ b/CIP-0067/registry.json
@@ -13,5 +13,10 @@
     "asset_name_label": 333,
     "class": "FT",
     "description": "CIP-0068 - Datum Metadata Standard (333 sub standard)"
+  },
+  {
+    "asset_name_label": 444,
+    "class": "SFT",
+    "description": "CIP-0068 - Datum Metadata Standard (444 sub standard)"
   }
 ]
--- a/CIP-0068/README.md
+++ b/CIP-0068/README.md
@@ -95,7 +95,7 @@ datum = #6.121([metadata, version, extra])
 
 #### 222 NFT Standard
 
-Besides the necessary standard for the `reference NFT` we're introducing two specific token standards in this CIP. Note that the possibilities are endless here and more standards can be built on top of this CIP for FTs, other NFTs, semi fungible tokens, etc. The first is the `222` NFT standard with the registered `asset_name_label` prefix value
+Besides the necessary standard for the `reference NFT` we're introducing two specific token standards in this CIP. Note that the possibilities are endless here and more standards can be built on top of this CIP for FTs, other NFTs, rich fungible tokens, etc. The first is the `222` NFT standard with the registered `asset_name_label` prefix value
 
 | asset_name_label            | class        | description                                                          |
 | --------------------------- | ------------ | -------------------------------------------------------------------- |
--- a/CIP-0068/README.md
+++ b/CIP-0068/README.md
@@ -338,7 +338,7 @@ In order to reference the correct UTxO containing the metadata, it needs to be a
 The security for the link is derived from the minting policy itself, so it's important to write the validator with the right constraints and rules since this CIP solely defines the interface to keep flexibility as high as possible.
 
 
-## Backward Compatibility
+### Backward Compatibility
 
 To keep metadata compatibility with changes coming in the future, we introduce a `version` field in the datum.
 
--- a/CIP-0068/README.md
+++ b/CIP-0068/README.md
@@ -246,19 +246,19 @@ We want to bring the metadata of the FT `d5e6bf0500378d4f0da4e8dde6becec7621cd8c
 3. Reference the output in the transaction. (off-chain)
 4. Verify validity of datum of the referenced output by checking if policy ID of `reference NFT` and `user token` and their asset names without the `asset_name_label` prefix match. (on-chain)
 
-#### 444 SFT Standard
+#### 444 RFT Standard
 
-The third introduced standard is the `444` Semi-FT standard with the registered `asset_name_label` prefix value
+The third introduced standard is the `444` Rich-FT standard with the registered `asset_name_label` prefix value
 
 | asset_name_label            | class        | description                                                          |
 | --------------------------- | ------------ | -------------------------------------------------------------------- |
-| 444                         | SFT          | SFT hold by the user's wallet making use of the union of CIP-0025 inner structure AND the Cardano foundation off-chain registry inner structure |
+| 444                         | RFT          | RFT hold by the user's wallet making use of the union of CIP-0025 inner structure AND the Cardano foundation off-chain registry inner structure |
 
-Semi-Fungible tokens don't fit cleanly into the other two FT/NFT classes of tokens and thus need their own standard. An example of a SFT would be a fractionalized NFT. The single reference NFT `(100)` represents the NFT itself, and the many `(444)` tokens represent the fractionalized shares. Minting 100 tokens and setting decimals to 2 would represent a single NFT that is split into 100 fractions.
+Rich-Fungible tokens don't fit cleanly into the other two FT/NFT classes of tokens and thus need their own standard. An example of an RFT would be a fractionalized NFT. The single reference NFT `(100)` represents the NFT itself, and the many `(444)` tokens represent the fractionalized shares. Minting 100 tokens and setting decimals to 2 would represent a single NFT that is split into 100 fractions.
 
 ##### Class
 
-The `user token` is an SFT (semi-fungible token).
+The `user token` is an RFT (rich-fungible token).
 
 ##### Pattern
 
@@ -313,7 +313,7 @@ Example datum as JSON:
 
 ##### Retrieve metadata as 3rd party
 
-A third party has the following SFT `d5e6bf0500378d4f0da4e8dde6becec7621cd8cbf5cbb9b87013d4cc.(444)TestToken` they want to lookup. The steps are
+A third party has the following RFT `d5e6bf0500378d4f0da4e8dde6becec7621cd8cbf5cbb9b87013d4cc.(444)TestToken` they want to lookup. The steps are
 
 1. Construct `reference NFT` from `user token`: `d5e6bf0500378d4f0da4e8dde6becec7621cd8cbf5cbb9b87013d4cc.(100)TestToken`
 2. Look up `reference NFT` and find the output it's locked in.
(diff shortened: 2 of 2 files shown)


Crypto2099:
> This is very similar to what I described [here](https://github.com/cardano-foundation/CIPs/pull/467#discussion_r1149267303). What you are calling a "reference NFT" is what I have been calling a "beacon token". #466 is a work-in-process CIP that is looking to explain how beacon tokens can be used for all kinds of applications, including being able to lookup specific datums as you describe. For consistency, I will use beacon tokens for the rest of this comment.
> 
> I do not believe it is necessary, nor a good idea, to have the asset and the beacon token share the same minting policy. What I have in mind is instead to have a separate beacon policy script that everyone uses for a given application. The token name of the beacon token would be the hash of the full asset name for the asset in question. This way there can still be a unique beacon token for every asset. I have not actually tried this yet but I have created a proof-of-concept that does something similar for storing the pre-image of a datum on chain with a beacon (effectively creating an on-chain map from datum hash to pre-image). You can find that [here](https://github.com/fallen-icarus/cardano-datums).
> 
> The main benefit to sharing the beacon policy is that the universal beacon policy can enforce the proper datum format. If we all agree to use beacon policy X, then if you see a beacon token (minted by that policy) at a UTxO, it is guaranteed to have a datum with the proper format for the standard. If minting policies had to both mint the asset and enforce the standard, you have no way of guaranteeing the standard is being followed without reading the code for every minting policy.
> 
> This is more a comment on CIP-0068 as a whole and not on the changes suggested for SFTs.

Just thinking about this in the larger context of beacon tokens in general... If there was some shared "Beacon Policy" how would you verify that I am the entity authorized to mint a "Beacon" for `SpaceBudz1234` (for example)? Utilizing the same policy for both reference and beacon tokens at least provides a cryptographic proof link between the two tokens. Also, how would anyone know that they should look for a beacon token under "Random Other Policy X" in this scenario?

fallen-icarus:
You can require that, in order to mint the beacon for asset A, asset A must be minted in the same transaction as the beacon is minted. The beacon policy would check that asset A is indeed minted in the transaction and would fail to mint the corresponding beacon if asset A was not minted. This is the same cryptographic proof as if they were minted by the same policy. You can make this more flexible by requiring that any asset be minted from the minting policy for asset A. This latter case can be useful if asset A was an NFT that can never be minted again.

As for as looking for the beacon token, that is what the standard is for. People are already looking to CIPs for the information regarding the standard. The CIPs would also mention the corresponding beacon token to check.

alessandrokonrad:
@AndrewWestberg the `444` label needs to be added to the [registry.json](https://github.com/cardano-foundation/CIPs/blob/master/CIP-0067/registry.json) of CIP-0067 as well.

AndrewWestberg:
@alessandrokonrad Updated CIP-67 registry.json

CIP-0068/README.md
```diff
@@ -246,6 +246,92 @@ We want to bring the metadata of the FT `d5e6bf0500378d4f0da4e8dde6becec7621cd8c
...
+    ? name : bounded_bytes, ; UTF-8
+    mediaType : bounded_bytes, ; UTF-8
+    src : bounded_bytes ; UTF-8
+    ; ... Additional properties are allowed
+  }
+
+metadata = 
+  {
+    name : bounded_bytes, ; UTF-8
+    description : bounded_bytes, ; UTF-8
```
alessandrokonrad, line 287: Is there a reason why the description is required for SFTs? I think this can still be optional.

CIP-0068/README.md
```diff
@@ -246,6 +246,92 @@ We want to bring the metadata of the FT `d5e6bf0500378d4f0da4e8dde6becec7621cd8c
...
+    image : bounded_bytes, ; UTF-8
+    ? mediaType : bounded_bytes, ; UTF-8
+    ? ticker: bounded_bytes, ; UTF-8
+    ? url: bounded_bytes, ; UTF-8
+    ? logo: uri,
```
alessandrokonrad, line 292: `logo` and `image` have the same purpose. I see you are merging the fields from the `222` and `333` standard. It's probably enough to just have the fields from the `222` label with the `decimals` field included.

rphair (commented):
This didn't make it into our CIP editors' meeting a couple days ago even thought I think it would have been a good candidate for discussion.  Will recommend it for next meeting's agenda (# 64, not scheduled yet, probably 12 days from now).

Once the stakeholders agree (e.g. https://github.com/cardano-foundation/CIPs/pull/494#discussion_r1159472297 https://github.com/cardano-foundation/CIPs/pull/494#discussion_r1159476645) I think this is well presented & would be ready to merge.

CIP-0068 never got a format review & update as per https://github.com/cardano-foundation/CIPs/issues/389 - although not recommending we update the header for the new CIP-0001 as part of this PR, I wanted to get @AndrewWestberg @alessandrokonrad attention to see if we could fix the _header levels_:

- I can understand why major headings like `333` and `444` were made H2's (`##`): for readability and to make sure their own subheadings aren't inconveniently small.

- Still the standard for a reference manual... and CIP-0068 is becoming a reference manual for these formats... is for strict header levels, like a thesis, rather than chosen cosmetically for type size.

So I am hoping the PR can appropriately demote the `222`, `333`, `444` sections so they're less significant than their container.  Since `Labels` is an H3 (`###`) I guess that would make each of these an H4, with the Class / Pattern / Metadata subsections as H5's.

Apologies for this nitpicky treatment but I do believe this will have to be reconciled someday soon, especially given the importance of this document.  If not handled in this PR I'll submit a separate PR myself for the header level correction allog with the preamble correction as per #389.  cc @KtorZ

CIP-0068/README.md
AndrewWestberg: Modified so it matches 222 with decimals field added.

CIP-0068/README.md
AndrewWestberg: moved to optional.

AndrewWestberg:
@rphair I believe I updated the headers correctly. Please let me know if it's all good.

rphair:
thanks @AndrewWestberg - looks good.  I corrected one header indentation that wasn't part of the new material because there isn't a standard "Backward compatibility" section.

I would vote for adding an H3 ('###`) with an explicit title that introduces all the numbered metadata formats (`222`, `333`, `444`) because otherwise they look in the outline like sub-sections of "Constraints and conditions".  But the added SFT material itself is clear enough now.

CIP-0067/registry.json
```diff
@@ -18,5 +18,5 @@
     "asset_name_label": 444,
     "class": "SFT",
     "description": "CIP-0068 - Datum Metadata Standard (444 sub standard)"
-  }
+  },
```
AndrewWestberg, line 21: Makes sense. thanks!

AndrewWestberg: Hmm. I thought trailing commas are not allowed in json. I know they are allowed in cbor diag, but I thought json was older and more strict.

AndrewWestberg: Note that before I made this change, there was no trailing comma.

rphair: Actually @AndrewWestberg I mucked this up. 😖 The whole year or so I've been looking at CIP-0010 labels & now these, my mind has been turning it into a JS array rather than JSON.  This is probably because of daily work on the Developer Portal's portfolio arrays like `[showcase.js](https://github.com/cardano-foundation/developer-portal/blob/staging/src/data/showcases.js)` which would be a murder of merge conflicts without being able to enforce those trailing commas.  Just committed to reverse the change.

AndrewWestberg:
@rphair Is this one ready for last check?

Ryun1 (approved):
This is not my area of expertise, but I see good discussion and positive reviews from the relevant actors so I will approve on that basis.

papag00se:
I would like to chime and here and say that without being aware of this PR, our team felt the need for this same CIP-68 variant. We even guessed at the (444) label designation.

In our case, ADA Handles will be launching background images with our NFT partners where we and our partners would like to offer the same background artwork to multiple Handle owners in perpetuity. This resonates with the trading card game example of multiple land/mana cards that all share the same properties but need to be in constant print. Not quite an NFT, but also not a monetary coin-like FT needing to be registered in the token registry,

All of that is to say, we endorse this idea and could make immediate use of it.

rphair:
Let's try to merge it at tomorrow's CIP meeting then.  Given the round of approval, promoting to `Last Check` in case the agenda, or items to merge, end up being determined on the fly (or shortly before the meeting).

AndrewWestberg:
@KtorZ @rphair SFT -> RFT update has been completed.

CIP-0068/README.md
```diff
@@ -93,7 +93,7 @@ extra = plutus_data
 datum = #6.121([metadata, version, extra])
 ```
 
-## 222 NFT Standard
+#### 222 NFT Standard
 
 Besides the necessary standard for the `reference NFT` we're introducing two specific token standards in this CIP. Note that the possibilities are endless here and more standards can be built on top of this CIP for FTs, other NFTs, semi fungible tokens, etc. The first is the `222` NFT standard with the registered `asset_name_label` prefix value
```
KtorZ, line 98:
```suggestion
Besides the necessary standard for the `reference NFT` we're introducing two specific token standards in this CIP. Note that the possibilities are endless here and more standards can be built on top of this CIP for FTs, other NFTs, rich fungible tokens, etc. The first is the `222` NFT standard with the registered `asset_name_label` prefix value
```

KtorZ (approved):
Approved. I did catch a tiny oversight but pushed directly on the branch 🙃.

CIP-0068/README.md
```diff
@@ -93,20 +93,20 @@ extra = plutus_data
 datum = #6.121([metadata, version, extra])
 ```
 
-## 222 NFT Standard
+#### 222 NFT Standard
 
-Besides the necessary standard for the `reference NFT` we're introducing two specific token standards in this CIP. Note that the possibilities are endless here and more standards can be built on top of this CIP for FTs, other NFTs, semi fungible tokens, etc. The first is the `222` NFT standard with the registered `asset_name_label` prefix value
+Besides the necessary standard for the `reference NFT` we're introducing two specific token standards in this CIP. Note that the possibilities are endless here and more standards can be built on top of this CIP for FTs, other NFTs, rich fungible tokens, etc. The first is the `222` NFT standard with the registered `asset_name_label` prefix value
```
KtorZ, line 98:
```suggestion
Besides the necessary standard for the `reference NFT` we're introducing three specific token standards in this CIP. Note that the possibilities are endless here and more standards can be built on top of this CIP for FTs, other NFTs, rich fungible tokens, etc. The first is the `222` NFT standard with the registered `asset_name_label` prefix value
```

rphair: good catch... I missed that

rphair:
The last of the items that came up at the last hour's CIP meeting have been addressed and document looks good in final check to me, so merging as planned without further ado 🤠

PR #496, opened 2023-04-04
CIP-0094? | SPO On-chain Polls

KtorZ:
**Abstract**

The Cardano Foundation proposes a mechanism for polling Cardano stake pool operators on specific topics. Polls are done on-chain through transaction metadata and authenticated through stake pool credentials (either VRF public key similar to what's described in [CIP-0022](../CIP-0022) or Ed25519 cold key). The goal is to gather opinions on governance matters such as protocol parameter updates. This standard is an inclusive interim solution while the work on a larger governance framework such as [CIP-1694][] continues.

[CIP-1694]: https://github.com/cardano-foundation/CIPs/pull/380

---

[View rendered version](https://github.com/cardano-foundation/CIPs/tree/cip-spo-polls/CIP-0094)


CIP-????/README.md
```diff
@@ -0,0 +1,250 @@
...
+1. The `$vrf_vkey`, `$vkey` and `$signature` are as defined in the original [Cardano ledger specifications](https://github.com/input-output-hk/cardano-ledger/blob/master/eras/babbage/test-suite/cddl-files/real/crypto.cddl) with the usual rules applying to transaction metadata (i.e. byte string longer than 64 bytes are encoded as lists)
+1. In either case (VRF or Ed25519), the witness' signature is calculated from the blake2b-256 hash digest of the serialised `answer` restricted to fields `2` and `3`.
+
+For example, in [CBOR-diagnostic](https://www.rfc-editor.org/rfc/rfc8949#name-diagnostic-notation) notation:
+
+```cbor
+{
+  94: {
+    2: h'3d4854aed3e31d0f5b6637fcd73e8713d9eee86a81d20598f6cb88d9b2bc0a39',
+    3: 2,
```
Crypto2099, line 140: Is this example based on the `question` example above where two string values are provided as the `choices`? If so, should the entries in `choices` be serialized based on a 1-index of the choices? As defined a "choice" should be a string or an array of strings but here is presented as a uInt value?

CIP-????/README.md
```diff
@@ -0,0 +1,250 @@
...
+
+> **Warning**
+>
+> Only the first answer to a poll for each credential shall be considered. If multiple answers are found, only the first answer submitted (transaction & block ordering tallying) shall be considered **unless** previous answers are from VRF keys and the latter from an associated cold key. In other words, it is possible to override a VRF-authenticated reply with a cold-key-authenticated one.
+
+### Duration
+
+A poll starts when a valid transaction with a question is posted on-chain and ends at the end of the following epoch, so there's always at least one entire epoch for answering the poll.
+
+In particular, the stake distribution of pools will be considered at the moment the poll is _started_.
```
Crypto2099, line 170: Since a question can be arbitrarily started at any point in the chain, should the most recent snapshot value be used for tabulating stake distribution? Or would this require auditors to perform a "live" stake snapshot at the precise block height of the question hitting the chain?

CIP-????/README.md
```diff
@@ -0,0 +1,250 @@
...
+- [ ] The Cardano Foundation has conducted a first trial poll on mainnet
+
+### Implementation Plan
+
+#### Tools Support
+
+- [x] [`cardano-cli`](https://github.com/input-output-hk/cardano-node/tree/master/cardano-cli#readme) will be updated to provide support for constructing and signing relevant transactions
+  - [x] See [input-output-hk/cardano-node#5050](https://github.com/input-output-hk/cardano-node/pull/5050).
+- [ ] [`cncli`](https://github.com/cardano-community/cncli) might be updated with similar support
+
```
Crypto2099, line 237: Add support by cardano-signer open source library by @gitmachtl?

CIP-????/README.md
```diff
@@ -0,0 +1,250 @@
...
+prompt =
+  text .size (0..64) / [ * text .size (0..64) ]
+
+choice =
+  text .size (0..64) / [ * text .size (0..64) ]
+
+nonce =
+  uint
+```
```
Crypto2099, line 78:
```suggestion
```cbor
question =
  { 0: prompt
  , 1: [ * choice ]
  }

prompt =
  text .size (0..64) / [ * text .size (0..64) ]

choice =
  text .size (0..64) / [ * text .size (0..64) ]

nonce =
  uint
  
poll = 
{ 0 : [ + question ]
  , ? "_" : nonce
}
```
```

With a minor change to a master `poll` object that can contain one-or-more questions the flexibility of this system can be greatly enhanced to support multiple-question polling while utilizing all of the same mechanics otherwise described. Here the nonce is added at the top-level to de-duplicate.

File: CIP-????/README.md [hunk 5]
```diff
@@ -0,0 +1,250 @@
...
+
+vrf_witness =
+  [ $vrf_vkey, vrf_proof ]
+
+vrf_proof =
+  [ bytes .size 64, bytes .size 16 ]
+
+ed25519_witness =
+  [ $vkey, $signature ]
+```
```
Crypto2099, line 125:
```suggestion
```cbor
answer =
  { 2: question_hash
  , 3: choice
  , ? 4: vrf_witness
  , ? 5: ed25519_witness
  }

question_hash =
  bytes .size 32

vrf_witness =
  [ $vrf_vkey, vrf_proof ]

vrf_proof =
  [ bytes .size 64, bytes .size 16 ]

ed25519_witness =
  [ $vkey, $signature ]
  
ballot = [ + answer ]
```
```

Here a `ballot` can be viewed as a simple array of question answers. Since each answer contains the hash of the question it answers, users may answer one (or more) of a multi-question poll in the same data structure.

File: CIP-????/README.md [hunk 6]
```diff
@@ -0,0 +1,250 @@
...
+  { 0: prompt
+  , 1: [ * choice ]
+  , ? "_": nonce
+  }
+
+prompt =
+  text .size (0..64) / [ * text .size (0..64) ]
+
+choice =
+  text .size (0..64) / [ * text .size (0..64) ]
```
Crypto2099, line 74:
```suggestion
prompt =
  [ * text .size (0..64) ]

choice =
  [ * text .size (0..64) ]
```

I would advocate in favor of forcing all prompts and choices to be an array-of-strings rather than optionally a single string or array. This simplifies the parsing and analysis process of third party auditors by removing an extra if/else check step while only saving a handful of bytes. Ergo, the benefit of a few bytes of savings in this situation does not justify the uncertainty in format IMO.

gitmachtl:
What i am missing in this specification - and we talked about that in our little dev group - is the method to do it similar like with CIP0036. So only using the VRF secret key once to register a voting key. After that this voting key is referencing the VRF key and so the VRF key does not have to be used anymore until you wanna change the voting key. For all the votings, only the signing via the voting key is used and not the vrf key anymore. The method would be similar to catalyst vote and CIP0036. 

An addition to cardano-cli should only contain the possibility to sign a provided hex/cbor payload via the VRF key, and not producing any other special structure, because it wouldn't be compatible with any other tools also using such a method.

CIP-????/README.md
KtorZ: It's a the index (0-based) of the selected choice from the corresponding question indeed. There's no need to replicate the full answer as this takes a needless amount of space (times 3000 for each pool).

KtorZ: I'll clarify that bit in the proposal.

CIP-????/README.md
KtorZ: That's a good question; and I guess what I meant to say here is that we consider the stake distribution for the _epoch_ where the poll is started. Stake is always only active after 2 epochs; so there's no point really looking at any "live" stake. Rather, stick to the same 'stake distribution' as considered by the ledger for various calculation.

CIP-????/README.md
KtorZ: I am not fond of this added flexibility; because it is also extra complexity for tooling down the line. But more importantly, I don't think we will ever want to ask multiple questions at once. Perhaps the example in the CIP is too silly to make this obvious but, we ain't going to ask about pineapples on pizzas. We need people to take time and do proper research on questions asked; so asking more than one question at a time is a recipe for disaster IMO.

CIP-????/README.md
KtorZ: Yes; I was on the fence on this once but I suppose that for the sake of "keeping-it-simple" it's best to always go with list.

KtorZ:
> _So only using the VRF secret key once to register a voting key. After that this voting key is referencing the VRF key and so the VRF key does not have to be used anymore until you wanna change the voting key._

Indeed, we've purposely kept it simple here. Mostly because this is meant to be an interim solution so we don't want to set precedence for a more complex process and a new set of credentials that'll be dropped once another on-chain governance framework enters the picture. That is debatable, but it would be relatively easy to have a first registration of an Ed25519 credential and use that onwards.  

> _An addition to cardano-cli should only contain the possibility to sign a provided hex/cbor payload via the VRF key, and not producing any other special structure_

Certainly not! That is a recipe for disaster. The next thing you get from that is people signing withdrawal requests or other disruptive actions without them knowing. This is why signing standards (e.g. COSE) always come with well-defined structures; We do not want people to sign arbitrary payloads. They always can through low-level libraries but higher-level interfaces such as the "official" cardano cli shall certainly not promote that as a benign command. 

> I would really avoid using the pool cold key for such actions, its the worse key you wanna want to be exposed. Also, it would automatically leave out all SPOs with there pool cold keys on the hw-wallet.

I am also from that school, but I got convinced by some others who actually prefer using their cold key for such an action as they see it as _the right tool for the job_. Hence the flexibility in the proposal; those who prefer using the VRF can, and those who prefer using their cold key can. I see good arguments for both.

CIP-????/README.md
```diff
@@ -0,0 +1,250 @@
...
+
+Polls will be multiple-choice questions by The Cardano Foundation with pre-defined answers to choose from.
+
+Here's an example of a question and answers:
+
+- _Pineapples on pizza?_
+  - [ ] yes
+  - [ ] no
+
+The serialised question and answers will be posted on-chain and signed by one of the delegate genesis keys owned by The Cardano Foundation. Answers will be provided on-chain by participating SPOs via transaction metadata referring to:
```
rphair, line 50: What about adding a pre-emptive description about how this might be done after CIP-1694, or something like it, removes the Genesis keys?  https://github.com/JaredCorduan/CIPs/blob/voltaire-v1/CIP-1694/README.md#changes-to-the-transaction-body

rphair (commented):
✔️ This looks fine to me as both a reviewer and an SPO but will wait until the technical considerations are settled before submitting an endorsing review.

CIP-????/README.md
Crypto2099: I believe, if I'm not mistaken, that this tool is only intended as a stop-gap until the Voltaire era kicks off and so would become redundant based on having the concept of a "poll" in CIP-1694. Also, interested third parties who wish to conduct polling of SPOs could add some form of key registration (non-Genesis key) in the future as an expansion to this CIP to provide "vote certification provenance".

KtorZ: Indeed. CIP-1694 would technically render this CIP obsolete especially now that we also added a "Poll" governance action to it. I should perhaps make that clearer in the CIP title that this is intended to be an interim solution only.

perturbing:
Great work on the CIP! It's clear that a lot of thought has been put into it.

While this CIP introduces an excellent framework for polling, it would be helpful to include more information on the conditions of the poll itself. Have you considered specifying these details?

For instance, as it stands, anyone can register a stake pool by depositing 500 Ada, which enables them to participate in a vote (potentially multiple times). It would be beneficial to elaborate on the criteria for accepting an answer in the poll and who will be responsible for setting these conditions. Incorporating this information into the CIP could provide a more comprehensive understanding of the polling process.

CIP-????/README.md
```diff
@@ -0,0 +1,250 @@
...
+1. The fields `4` and `5` are not mutually optional; said differently, at least one of the two must be provided.
+1. The `$vrf_vkey`, `$vkey` and `$signature` are as defined in the original [Cardano ledger specifications](https://github.com/input-output-hk/cardano-ledger/blob/master/eras/babbage/test-suite/cddl-files/real/crypto.cddl) with the usual rules applying to transaction metadata (i.e. byte string longer than 64 bytes are encoded as lists)
+1. In either case (VRF or Ed25519), the witness' signature is calculated from the blake2b-256 hash digest of the serialised `answer` restricted to fields `2` and `3`.
+
+For example, in [CBOR-diagnostic](https://www.rfc-editor.org/rfc/rfc8949#name-diagnostic-notation) notation:
+
+```cbor
+{
+  94: {
+    2: h'3d4854aed3e31d0f5b6637fcd73e8713d9eee86a81d20598f6cb88d9b2bc0a39',
```
perturbing, line 139: I just wanted to confirm, is the hash specified here not the "BLAKE2B-256" digest of the serialized pineapple poll example?

If I hash it, I get `c9a077a1098d73498d17e9ea27045af820c311ced91f8c2bb9b5c7f446379063`.  Here's the code snippet that you can run with deno for reference:
```ts
import * as mod from "https://deno.land/std@0.182.0/crypto/mod.ts";
import {decode} from "https://deno.land/std/encoding/hex.ts";

const hexString="A1185EA2007450696E656170706C6573206F6E2070697A7A613F018263796573626E6F";

const hash = await mod.crypto.subtle.digest(
	"BLAKE2B-256",
	decode(new TextEncoder().encode(hexString))
);
console.log(mod.toHashString(hash));
```

KtorZ: The examples are NOT test vectors indeed, mostly gibberish data. Could be arguably more useful if there were test vectors. I can try pulling up the test scenarios implemented for input-output-hk/cardano-node#5050

KtorZ:
> @perturbing: While this CIP introduces an excellent framework for polling, it would be helpful to include more information on the conditions of the poll itself. Have you considered specifying these details?
> 
> For instance, as it stands, anyone can register a stake pool by depositing 500 Ada, which enables them to participate in a vote (potentially multiple times). It would be beneficial to elaborate on the criteria for accepting an answer in the poll and who will be responsible for setting these conditions. Incorporating this information into the CIP could provide a more comprehensive understanding of the polling process.

The CIP is **purposely** vague on that area and we've been quite careful not using the word "vote" exactly because of that too. How will the result be interpreted is still up in the air. Those surveys will be conducted by the genesis delegates in an attempt to bring more data points in the decision making when it comes to changing parameters. They also serve as experiments regarding social behaviors and responsivity of SPOs. 

This idea mainly came after we realized that there's arguably no good way to survey SPOs nowadays. Sending a link to a TypeForm survey in a newsletter for example is highly unsatisfactory and give little guarantees. So, in the spirit of CIP-1694, we are considering a "1 ADA = 1 Voice" approach. Hence, one actor registering a stake pool with minimum pledge / stake at the last minute would have little influence on the survey. 

As @gitmachtl is also pointing out, we may want to give more weight to stake that is pledged, and have a more complex formula to establish legitimacy but these discussions will happen outside of the proposal and we may revisit them based on the community feedback. The primary goal is to produce the data. Then, how we chose to interpret the data is another interesting question.

CIP-????/README.md
```diff
@@ -0,0 +1,250 @@
...
+ed25519_witness =
+  [ $vkey, $signature ]
+```
+
+Some remarks:
+
+1. The field `2` (`question_hash`) is a blake2b-256 hash digest, whose preimage is the entire serialised question metadata payload (with the `94` top-level label).
+1. The fields `4` and `5` are not mutually optional; said differently, at least one of the two must be provided.
+1. The `$vrf_vkey`, `$vkey` and `$signature` are as defined in the original [Cardano ledger specifications](https://github.com/input-output-hk/cardano-ledger/blob/master/eras/babbage/test-suite/cddl-files/real/crypto.cddl) with the usual rules applying to transaction metadata (i.e. byte string longer than 64 bytes are encoded as lists)
+1. In either case (VRF or Ed25519), the witness' signature is calculated from the blake2b-256 hash digest of the serialised `answer` restricted to fields `2` and `3`.
```
dcoutts, line 132: Can we please be _much_ more explicit about what we do with the VRF.

A VRF is not a digital signature scheme, it's a verifiable random function. So please don't refer to VRF signatures. There's no such thing. There is a VRF _proof_. And the VRF needs an input. Lets be explicit about what the input is.

And we should be clear that we are abusing/reusing the VRF for the purpose of proving knowledge of the VRF secret, and that we are ignoring the VRF output: we only care that the proof is valid.

I would separate out the specification of how to construct and verify the witness for the Ed25519 case and the VRF case. In particular it's really important to not confuse people into thinking we use the VRF key as if it were an ED25519 key (['coz that's really bad!](https://github.com/input-output-hk/blogpost_breaking_ed25519_vrf/blob/main/blog.md)).

dcoutts (changes requested):
I have to say I'm not happy with the option to use the SPO VRF, and I'm not happy with the answers including the witness into the metadata rather than just using an ordinary transaction signature (as is proposed for the questions).

I'd recommend no option for the VRF and simply require the answer to have an ordinary transaction signature on the transaction by the SPO cold key (i.e. no signature in the metadata). So the metadata would only contain the question hash and the answer index.

And if you want to use the VRF anyway, please be much more explicit in the description of the use of the VRF. It is not a digital signature algorithm, so don't just elide the details as if it were one. Be explicit that you're abusing/reusing the VRF for the same purpose as a digital signature and be clear about what the VRF inputs are and how the outputs are (un)used.

It needs to be clear you're not using the VRF key as if it were an ED25519 key. Currently that's very unclear.

gitmachtl:
> Or as others have suggested: register another ED25519 key and use that (on the whole transaction in the ordinary way, not a sig within the metadata) so that the cold key doesn't have to be used all the time. That'd also be fine.

Thats what i suggested. Just using the VRF key once to register a "VotingKey" like with CIP36. After that, just use the normal ed25519 key with the usual signing method/tools further on.

CIP-????/README.md
KtorZ: Right, there are a few language abuse in the CIP regarding signature vs proofs; I've been much more careful about these formulation in the cardano-cli changes but it's a bit clumsy in the CIP. I'll edit that. 

> @dcoutts: And we should be clear that we are abusing/reusing the VRF for the purpose of proving knowledge of the VRF secret

I thought this was already clear as we're pointing back to CIP-0022's discussions where it's pretty much stated as-is; but point taken, I'll make that more apparent. 

Note that I haven't a particular attachment or preference to supporting VRF proofs here, but it's something that was explicitly raised by a few SPOs as more practical for them than accessing their cold key. I haven't pushed back on that based on Iñigo's comments on CIP-0022 which didn't really advocate _against_ the (ab)use of VRF for proving knowledge of the secret key, but rather pointed at its inefficiency. Since efficiency isn't a particular concern for this use case here either it seemed like _an acceptable option_.

perturbing: Upon further review and taking into consideration the blog post shared by @iquerejeta, I concur with @dcoutts' perspective.

The primary issue with this CIP is that it combines two cryptographic systems insecurely, even though it might be beneficial to have the option to utilize both. As suggested in the blog post, this could be resolved by segregating the domains and incorporating padding into the hashing process to mitigate potential attacks.

My primary concern is that an implementor other than IO might refer to this CIP and inadvertently apply it incorrectly, thereby exposing private keys. This risk can be averted by either opting for a VRF proof or an Ed25519 signature, or by employing an alternative setup, such as registering another Ed25519 key as outlined below. As it currently stands, this standard introduces the possibility of insecure handling of sensitive information.

iquerejeta:
I think it's ok to use the VRF key to provide a proof of possession of the VRF skey (which is what we are ultimately doing with the VRF proof). However, I agree with Duncan in that it is not a signature, so the wording should be very explicit and should be chosen carefully, as indeed misuse of the keys would have catastrophic consequences. If registering yet another key is a possibility, I'd prefer that. 

I don't like using the cold key for polls because we want to incentivise SPOs to have their cold keys in an air-gapped machine, but we also want to incentives SPOs to participate in polls. If polls happen often, we are not providing the right incentives. 

My order of preference of possible solutions would be (1) registering a new ed25510 key, (2) use KES key (which actually is meant to _sign_ data to identify an SPO, (3) VRF key, and finally (4) the cold key. Using some sort of padding in the message, as @perturbing suggested,  might be a solution against possible misuse, but does not solve the real problem. 

However, I understand that some people might want to have the highest level of identification guarantees for polls, therefore supporting the usage of cold keys.

CIP-????/README.md
KtorZ: They are now.

gitmachtl:
so, if all responses combines. i think its the best to do the method on registering a new ed25519 key once with the vrf key proof. like in cip36. and just use a normal ed25519 key further on? @KtorZ @dcoutts @JaredCorduan @iquerejeta @AndrewWestberg @Crypto2099 @gufmar

KtorZ:
Thanks all for your feedback; I have addressed them directly in the CIP to make sure that the conversations are captured over time. Here's however a summary of the changes / additions:

- [x] (cc @dcoutts, @iquerejeta) I have made clearer the distinction between VRF proofs & digital signatures, see for example:
  > "A digital signature (EdDSA) from the SPO's current cold key, or a VRF proof proving ownership of associated VRF credentials."

- [x] (cc @Crypto2099) I have clarified usage of the field `3`:
  > "The field `3` represents the 0-based index of the chosen answer from the available choices (from field `1` of the target poll)."

- [x] (cc @perturbing) All examples are now proper "test vectors" (i.e. they correspond to what's generated from cardano-cli).

- [x] (cc @dcoutts, @perturbing) I've added a proper disclaimer regarding re-use of Ed25519 secret key as VRF secrets:

  > **Warning**
  >
  > Note that the VRF here is **NOT** used as a digital signature algorithm, but used to produce a proof of possession of the corresponding VRF secret key. This is a _slight abuse_ of the VRF scheme since the verifiable random output produced is completely discarded, though safe provided that the VRF secret key was generated independently of the cold key.
  >
  > Let's stress this out a little: it is extremely important for implementors to keep that difference in mind and to **NOT** attempt to use the cold key as a VRF secret in an attempt to generate a VRF proof! **Ed25519 is only proven secure in the setting where an adversary has access to only signatures**. Attempting to re-use the same secret for different purpose could have dramatic consequences.
  >
  > See [Insecurity On Secret Key Re-Usage](https://github.com/input-output-hk/blogpost_breaking_ed25519_vrf/blob/main/blog.md#insecurity-of-secret-key-re-usage) for further detail.

- [x] (cc @perturbing) I have expanded on the intent of the survey and the interpretation of the data:

  > The outcome of a poll will depend on its level of participation (in **terms of stake**). It is essential to understand that we explicitly call this a _poll_ / _survey_ and not a _vote_ to dispel any possible confusion. So it is akin to `1 Lovelace = 1 Voice` although we may chose to interpret data using different equations (e.g. giving more weight to pledged stake). How the data is interpret is deemed out of the scope of this proposal which aims mainly at producing the data-points. Further conversations and debates will be needed regarding interpretation of the data-points.

- [x] (cc @gitmachtl, @iquerejeta) I have given a justification as for why we do not use the KES scheme here, though left the door open

  > There's a third on-chain element which we could use for identifying SPOs which is a digital signature from their KES credentials. It is however a bit more annoying to leverage mainly because KES are meant to expire and are only loosely tied to pools by operational certificate. Thus, verifying KES signatures on a survey requires a more complex setup and monitoring to keep track of operational certificates and their validity at the time of the survey.
  > 
  > If this CIP was meant to NOT be an interim solution, this is something we would likely consider. However, given the timeframe we're looking at and the overall trade-offs in complexity, we have opted out of using the KES as an authentication mechanism in this iteration. 
  
 - [x] (cc @iquerejeta) I have given a justification as for why we do not use proxy keys here too, though also left the door open
  > Another possible alternative to what's described in the CIP would be to have SPOs register a proxy Ed25519 key and use that proxy key onward. The validity of the proxy key registration would be conditionned to the production of an associated VRF proof or a digital signature from the cold key (very much like it's done for operational certificate).
  > 
  > Yet, like the KES alternative, this option is in conflict with some of the design goals of this CIP: simplicity. All the more so given that we want to maximise participation of SPOs to the various surveys. We aim to make the process of participating to the survey as simple as possible, without compromising on security.
  >
  > > **Note** Both alternative options for KES Signing and Proxy Keys may be re-considered in a future version of the survey. Especially if the solution turns out to be not _as temporary as intended_. Fortunately, the current design decisions do not preclude this from happening as it shall be possible to introduce two new witness types `6` and `7` for those purpose. The KES registration can be handled through a separate on-chain event.

- [x] (cc @dcoutts) I have given a justification as for why we do not leverage the existing witness structure on transactions
  > In the current proposal, we serialize signatures and proofs directly within the transaction metadata. One could make the argument that there's already room for Ed25519 key witness in the transaction which could be re-used here instead of being embedded in metadata. However, this is impractical in two places:
  >
  > - For producers of these metadata, which now have two inconsistent flow for producing witness. Indeed, there's no equivalent witness in the transaction for storing VRF proof, so proving using a VRF secret key will have to go in the metadata, whereas a digital signature could go as normal witness. Note that the ledger doesn't allow extraneous witnesses, which means that a signer would also have to add an "extra required signer" to the transaction; generating even more inconsistency.
  > 
  > - For consumers of these metadata, which instead of relying on one consistent source of data (parsing metadata), now needs to also parse other parts of transactions, making the entire discovery process more involved.

CIP-0094/README.md
```diff
@@ -0,0 +1,350 @@
...
+    1. In either case (VRF or Ed25519), the witness is calculated from the blake2b-256 hash digest of the serialised `answer` restricted to fields `2` and `3`.
+
+> **Warning**
+>
+> Note that the VRF here is **NOT** used as a digital signature algorithm, but used to produce a proof of possession of the corresponding VRF secret key. This is a _slight abuse_ of the VRF scheme since the verifiable random output produced is completely discarded, though safe provided that the VRF secret key was generated independently of the cold key.
+>
+> Let's stress this out a little: it is extremely important for implementors to keep that difference in mind and to **NOT** attempt to use the cold key as a VRF secret in an attempt to generate a VRF proof! **Ed25519 is only proven secure in the setting where an adversary has access to only signatures**. Attempting to re-use the same secret for different purpose could have dramatic consequences.
+>
+> See [Insecurity On Secret Key Re-Usage](https://github.com/input-output-hk/blogpost_breaking_ed25519_vrf/blob/main/blog.md#insecurity-of-secret-key-re-usage) for further detail.
```
iquerejeta, line 156: Maybe we want to link the Essential Cardano article, which contains the final version (I'm happy keeping any of the two)
https://www.essentialcardano.io/article/insecurity-of-secret-key-re-usage

rphair: thanks @iquerejeta for noting the released version; I was wondering where that had ended up.

I agree the _Essential Cardano_ link would be a nice touch, but the GitHub blog page has better code syntax highlighting & has properly typeset maths instead of images.  Then again I've heard the GitHub maths don't work on mobile browsers, which would support using the Essential Cardano page.  Perhaps therefore they both can be linked?

KtorZ: Sweet, thanks for pointing that out. The title suggested this had been posted somewhere but I also couldn't find where.

gitmachtl:
@KtorZ i am a bit confused now about https://github.com/input-output-hk/cardano-node/pull/5132

i thought we shift it to register a normal ed25519 key once with a vrf prove? using the pool cold key disables all SPOs which are having there pool keys on a hw device.

dcoutts:
@gitmachtl That PR from @KtorZ is to do with the timing of the next Cardano node (& API, CLI etc) release. The next release is "Real Soon Now" TM (usual caveats apply of course). More specifically the feature and code freeze is very much RSN.

As implementors, our choices right now are basically:

1. to include as simple as possible a version into the next release (minimising complexity and audit requirements); or
2. to wait until this CIP discussion has come to consensus and the CIP gets to draft phase, and then include an implementation into a later node release.

A lot of people want to get something out as soon as possible, so we're going with 1. By keeping it as simple as possible, it doesn't preclude extending the implementation later once the CIP discussion comes to a conclusion and the CIP reaches draft status.

So for now, by omitting the VRF feature, and using the normal transaction signatures (i.e. not in metadata) we minimise the implementation complexity and the audit requirements.

This is not intended to shut down discussion on this CIP. It's just about getting a minimal implementation out there sooner.

gitmachtl:
@dcoutts thx for your reply. ok i understand. i still thing that using the pool cold key is the worst choice here, because even a vrf key would be compromised, it can be renewed. not so with the "holy" pool cold key. it also leaves out hw pool key users. its just a bit of a surprise for me to choose the pool cold key after all the security discussions over last couple of months. but ok...

KtorZ:
@gitmachtl my apologies for the confusion; we had an off-the-record chat with @dcoutts which I am yet to capture / report on the CIP. So as Duncan said, the immediate goal w.r.t the cardano-cli is to proceed with cold key only for the next cardano-node / cardano-cli release. 

I'd love to still consider the VRF option for this CIP but we might have to go through a non official release track for that. At least for now. One possibility could be to host on https://github.com/cardano-foundation a fork of the cardano-cli which provides an extension of 1.35.7 with VRF support for proving ownership of secret. 

Given the time constrains and heaviness of the release process for the current cardano-cli, we couldn't make that happen for the next release. We are doing our best to explore compromises.

ashisherc:
@KtorZ @gufmar looking at the updated metadata (yet to be updated in this doc), would like to propose a slight change.

The cold key witness is being added to the tx witness set. This is a good choice, and it would be better to add the pool id bytes in the metadata at the key `4`. It would make it easier to verify the witness against the pool id. Currently, it requires us to go through each witness set (there could be more unused witnesses) and match against existing pool ids. This will also make it easier for the tool being used by the CF team to validate/map the vote with the stake pool.

KtorZ:
> It requires us to go through each witness set (there could be more unused witnesses) and match against existing pool ids

That is a non-accidental feature here. If we put he pool id in the metadata themselves, then this extra step is still necessary but won't likely be done at all. Nothing ensures that the pool id actually exist. Plus, the pool ids will be the result of a snapshot; which means that it's even slightly more complicated as the current set may contain non-valid pools ids (or be missing some).

ashisherc:
I do not think it will complicate the process, and there isn't a problem of a non-existent pool id while calculating the final report. What I am suggesting is more like the address(paymentkey) in the utxo which is verified against the set of witnesses. 

Adding an extra poolId field will make it easy derive the poolId (read from the metadata field) of the stake pool which is posting the answer, and then verify if its witness is present in the witness set (just like for a stakeKey withdrawal or a utxo).

While calculating the report if the stake pool does not exist it has to be filtered out even in case the poolId is not part of the metadata, but poolId being part of the metadata makes it easy to map the answer to a stake pool and also validate the mapping easily.

dmitrystas:
imo the extra field looks even better then required-signer, because with extra filed in metadata we can definitely determine which pool we are using (it can be only 1 field with key 4), unlike required-signer, which can be more then one

gitmachtl:
ok, got the "required-signer" working with cardano-cli. but the `--required-signer-hash` parameter must be used, because the normal `--required-signer` parameter expects a payment skey json file and not a pool skey json file. but using the hash works.

KtorZ:
Regarding the use of an extra field, it doesn't actually solve any problem but it introduces new ones. 

The reason why we use the required signers is because this piggybacks on a first class verification of the ledger. Each node will ensure that the transaction is indeed signed by one of the specified credentials. By using an extra field we loose that guarantee and thus, it becomes the responsibility of those who consumes the metadata to do this extra verification step. There's a non zero chance that some (most?) won't and will simply believe whatever is presented in metadata. Plus, the required signer enforces the presence of a corresponding witness whereas an extra field would not.

About the abuse of the field, it's not. And here we have to distinguish two things: the ledger rules and the cardano cli. From the ledger standpoint there's no such thing as "an address key" or "a stake pool key". Those are Ed25519 public key hashes. The distinction only exist at the interface level to prevent silly mistake. Here, there's arguably no reason to only require payment keys as extraneous signer (as if only payment keys were used for signing stuff?)... but that wouldn't be the first time I disagree with some of design choices in the cardano cli ☺️... I am pretty sure the maintainers would be open to a PR to change that.

ashisherc:
if an operator posts an answer with unnecessary signers in the required-signer field along with the pool key, and also adds those witnesses in the tx, would it be considered a valid cast?

dmitrystas:
if an SPO posts an answer that is not in the list of answers (for example the poll contains only 3 options 0,1,2, but SPO posts 5 in the field_3, which logic should be in this case? should we completely ignore this voice, and not take it into account at all (and take into account the SPO voice next time if the SPO sends the correct answer later), or should we take into account this voice as _wrong/broken_, and not take into account any other posts from this SPO?

gitmachtl:
just mark it as "SPO is stupid" 😄

KtorZ:
@ashisherc there can be more than just the cold key hash in the required signers indeed. It doesn't invalidate the answers. There must be *at least* the cold key hash.

@dmitrystas if an SPO submits a well formed but invalid answer then it counts as his/her answer but it doesn't contribute towards one of the existing choice. That means that even if they submit another (valid) answer, it is only the first one that is taken into account.

dmitrystas:
> should the dashboards/tools only take the first valid one into account?

yep, only the 1st one

gitmachtl:
@dmitrystas so valid is if the metadata is ok, the required signer is ok and the transaction is signed with the proper cold key. even if the answer in the metadata has an invalid value. correct?

dmitrystas:
@gitmachtl as far as I understand Matthias' answer - yes, even in case the key 3 has an invalid (not from poll's option list) value. Only the required signers are important, and a well formed metadata with key 2 as poll hash and key 3 with answer. I broke the value for a one pool locally, just for a test to check how it works - it looks something like this https://prnt.sc/0VmWtRJjEFKm

KtorZ:
Indeed @gitmachtl; this is the distinction I make between well-formed and valid. A well formed submission can have an out-of-bound answer. 

I've written [a crawler script](https://github.com/cardano-foundation/CIP-0094-polls/blob/ebd6482a429058d3d1569e72e94e81bdfae773cb/crawler/cmd/crawl.mjs) which captures those validations. I am still on the fence as for whether we should allow multiple submissions in case of a well formed but invalid answer. I don't see any (security) reason to allow multiple submissions in that case and only consider the first well formed but valid answer. But there's also no strong reason to make validations more complicated than necessary... that's more room for mistakes.

CIP-0094/README.md
```diff
@@ -0,0 +1,281 @@
...
+
+Polls will be multiple-choice questions by The Cardano Foundation with pre-defined answers to choose from.
+
+Here's an example of a question and answers:
+
+- _Pineapples on pizza?_
+  - [ ] yes
+  - [ ] no
+
+The serialised question and answers will be posted on-chain and signed by one of the delegate genesis keys owned by The Cardano Foundation. Answers will be provided on-chain by participating SPOs via transaction metadata referring to:
```
Ryun1, line 53: Is it possible for any genesis key holding entity to publish a poll? or is it only possible for the Cardano foundation?

CIP-0094/README.md
```diff
@@ -0,0 +1,281 @@
...
+</tbody>
+</table>
+
+The transaction carrying the answer metadata must then **be signed using a stake pool operator cold key**. Because cold key are not payment keys, this requires specifying an extra required signer on the transaction (transaction's field number 14 as per [Babbage's CDDL](https://github.com/input-output-hk/cardano-ledger/blob/cffa75fdbd800cda60997791e51bf02f2af0c42b/eras/babbage/test-suite/cddl-files/babbage.cddl#L66)). The signature shall be produced in an air-gapped environment only.
+
+> **Warning**
+>
+> Only the first answer to a poll for each credential shall be considered. If multiple answers are found, only the first answer submitted (transaction & block ordering tallying) shall be considered.
+
+### Procedure and Duration
```
Ryun1, line 174: For me, what is missing from this specification is an explanation of the full poll flow, for example: 

- how can SPOs can find new polls to vote on?
- how any ecosystem actor can verify the results of a poll?
- where does the report get posted?

CIP-0094/README.md
```diff
@@ -0,0 +1,281 @@
...
+
+After the epoch in which the Stake Pool Operators have cast their votes, there follows a period of one or more epochs in which the ADA Delegates may respond: If they disagree with the choice of their current Stake pool, they can delegate to another pool. This changes the stake weight and thus influences the result.
+
+![Alt text](CIP-0094_procedure-duration.png "Epoch poll phases")
+
+Indirectly, this results in the possibility of participation for all ada holders. Only after this re-delegation epoch(s) is completed, is the stake snaphot of the pools made to calculate the outcome. 
+
+### Outcome
+
+The outcome of a poll will depend on its level of participation (in **terms of stake**). It is essential to understand that we explicitly call this a _poll_ / _survey_ and not a _vote_ to dispel any possible confusion. So it is akin to `1 Lovelace = 1 Voice` although we may chose to interpret data using different equations (e.g. giving more weight to pledged stake). How the data is interpret is deemed out of the scope of this proposal which aims mainly at producing the data-points. Further conversations and debates will be needed regarding interpretation of the data-points.
```
Ryun1, line 186: > we explicitly call this a _poll_ / _survey_ and not a _vote_

Except in the `Procedure and Duration` section where its referenced as a vote :)

CIP-0094/README.md
KtorZ: Yes, any genesis delegate can. Though we wanted to make it clear that this initiative comes from the CF and will be used by the CF. We want to avoid implying a commitment from the other entity in their stead.

CIP-0094/README.md
```diff
@@ -0,0 +1,281 @@
...
+- The question and answers
+- The index of the chosen answer from the available choices
+- A digital signature (EdDSA) from the SPO's current cold key
+
+> **Note**
+> In this document, every time we refer to a _serialized object_, we refer to its **canonical** CBOR representation. In particular, keys in a map are always ordered alphabetically.
+
+### Question structure
+
+A question is posted in a transaction's metadata using the metadata label `94` and the following metadata structure:
```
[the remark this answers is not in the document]
KtorZ: Not yet! I originally wanted to validate the approach on cip-0094 before doing that but I guess it's now time. Thanks for the reminder 👌

CIP-0094/README.md
KtorZ: We've addressed those points in separate documents and tutorial by now. Thing is, I don't necessarily want to take a strong stance about these in the CIP itself to not force one single approach. 

The data lives on chain as transaction metadata which means that it's just the usual approach for any tool or builder to get access to it. We now even have dashboards in at least two explorers (adastat and cardanoscan) that are collecting and validating these data. And it's relatively straightforward with a chain sync. More tools will likely follow.

CIP-0094/README.md
```diff
@@ -0,0 +1,281 @@
...
+
+The transaction carrying the answer metadata must then **be signed using a stake pool operator cold key**. Because cold key are not payment keys, this requires specifying an extra required signer on the transaction (transaction's field number 14 as per [Babbage's CDDL](https://github.com/input-output-hk/cardano-ledger/blob/cffa75fdbd800cda60997791e51bf02f2af0c42b/eras/babbage/test-suite/cddl-files/babbage.cddl#L66)). The signature shall be produced in an air-gapped environment only.
+
+> **Warning**
+>
+> Only the first answer to a poll for each credential shall be considered. If multiple answers are found, only the first answer submitted (transaction & block ordering tallying) shall be considered.
+
+### Procedure and Duration
+
+A poll starts when a valid transaction with a question is posted on-chain. Ballots can be submitted until the end of the following epoch, so there is always at least one whole epoch to answer the poll.
```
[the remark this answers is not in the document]
KtorZ: The word vote shouldn't have been used in the previous section. I'd like to stick to "answer". Same for ballot.

CIP-0094/README.md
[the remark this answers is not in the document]
gufmar: It should, together with this CIP being set to active state

CIP-0094/README.md
gufmar: At the moment, every question with the label 94 appears on-chain for all to see. Technically, from this moment on, anyone - with a pool s key - can pick up this question and submit an answer. It is up to the publishing entity to set the cast and redelegation duration. But it is not up to this one entity to interpret or publish the result, because it has no right to decide about the interpretation or reaction of the other entities. 

Like the question, all individual ballots are transparent and verifiable on-chain. The transaction format with the `--required-signer-hash` ensures that only ballots validly signed by a pool end up on chain. 

To make the on-chain data easily accessible for those who do not have a full indexing instance (e.g. dbsync) available, the CardanoFoundation has set up a github repository where an index file in json format is maintained. There you can find (for now) the cast/delegation epochs in a standardized machine-readable format. In addition, there is also the possibility to store list of all ballots that were considered valid summarized in one file, in order to give everyone easy access and the possibility of their own analysis and interpretation.

The methodology deliberately chosen in this poll design is a creative+inclusive+motivational iteration. This perhaps somewhat contradicts the common expectation that a CIP must be written (first define, then implement, then execute) 
In this process, we deliberately wanted to present a maximally simple, straight executable thing, with lots of room for ideas. People should have the opportunity to get hands-on experience during the build process and participate in the rest of the design. At the beginning of the project it was still unclear whether this was only a one-time test run with the goal of incorporating the data and findings into the design of CIP-1694, or whether it would make sense to refine this CIP step by step for further test runs.

CIP-0094/README.md
gufmar: fixed this

rphair:
@KtorZ @gufmar This question is long-term for the CIP but also urgent because I'd like to submit the [current mainnet poll](https://forum.cardano.org/t/entering-voltaire-poll-experiment-live-on-mainnet/117879) based on the answer...

The CIP deals with operational details like security, but it's also operationally important for poll responders to add metadata.  What's the proper way of doing this?  And would it therefore make sense for CIP-0094 to explain how to use it with CIP-0020 transaction messaging, assuming that's possible?

Following the mainnet tutorial ([dev portal version](https://developers.cardano.org/docs/operate-a-stake-pool/on-chain-polls/), currently more up-to-date) & adding a `674` metadata section to the generated JSON:

```
{
    "94": {
        "map": [
            {
                "k": {
                    "int": 2
                },
                "v": {
                    "bytes": "96861fe7da8d45ba5db95071ed3889ed1412929f33610636c072a4b5ab550211"
                }
            },
            {
                "k": {
                    "int": 3
                },
                "v": {
                    "int": 5
                }
            }
        ]
    },
    "674":
         {
           "msg":
                  [
"The results of this poll can't be actionable ..."
                  ]
         }
}
```
doesn't seem compatible with `--json-metadata-detailed-schema` which I assume is required to map the poll data... perhaps because the [CIP-0020 message specification](https://github.com/cardano-foundation/CIPs/blob/master/CIP-0020/README.md#specification) doesn't have a detailed schema:
```
cardano-cli transaction build-raw \
...
--metadata-json-file votemeta.json \
--json-metadata-detailed-schema \
...
Command failed: transaction build-raw  Error: Error reading metadata at: "votemeta.json"
JSON schema error within the metadata item 674: {"msg":["The results of this poll can't be actionable ...
JSON object does not match the schema.
Expected a single field named "int", "bytes", "string", "list" or "map".
Unexpected object field(s): {"msg":["The results of this poll can't be actionable ...
```

It would seem an important part of the voting process, especially with the popular "none of the above" category of our current mainnet vote, that explanatory messages with our votes be readable by the growing number of Tx message metadata parsers in pool explorers and wallets.  So how can we combine the CIP-0020 messaging and CIP-0094 voting metadata structures in a single transaction?

dmitrystas:
> [CIP-0020 message specification](https://github.com/cardano-foundation/CIPs/blob/master/CIP-0020/README.md#specification) doesn't have a detailed schema

```
{
    "674": {
        "map": [
            {
                "k": {
                    "string": "msg"
                },
                "v": {
                    "list": [{
                        "string": "The results of this poll can't be actionable ..."
                    }]
                }
            }
        ]
    }
}
```

gitmachtl:
I wanna suggest an interesting idea for future SPO polls or any other things that needs an identification of an SPO. We talked about to sign messages with the VRF key in the past, and there were pros and cons about this. A few of us suggested to use a separate "registered" ed25519 voting key as a "stand in" to represent the SPO. 

The current poll issue with the pool cold key on the hw wallet got me thinking, and i think a wonderful solution would be to simply let the SPOs do a PoolUpdate-Registration with metadata attached. This metadata simply contains the voting key - similar generated than CIP36. That way we have a proof on chain that the SPO is in charge of the right pool secret key, and we also have an association of that SPO/Pool to a normal ed25519 key. No VRF signing required. And it can be done by both, cli pool keys and by hw pool keys. I already tested the hw pool key side.

Thoughts?

gitmachtl:
I could write up a draft for a "Pool Authentication" CIP using such a metadata method. There could also be more information in it. Also would it be possible to combine multiple Pools together under one key if someone likes to do so. Having a standard ed25519 key makes many things easier for signing and without the risk to expose a cold key or the vrf key in the future for every kind of transaction that need such a proof.

PR #505, opened 2023-04-17
CIP-0010 | Add label: CIP-0072 (dApp registration & Discovery)

matiwinnetou:


rphair (approved):
defined in CIP-0072 draft here: https://github.com/ehanoc/CIPs/blob/cip-72_dapp-registration/CIP-0072/README.md#metadata-label

Ryun1:
Is there any reason why this change to CIP-10 cannot be included in the [CIP-72? PR](#355)? do we want to encourage PRs per CIP change? @rphair

rphair:
@Ryun1 I thought that merging it separately, and prior to https://github.com/cardano-foundation/CIPs/pull/355, would allow @matiwinnetou and others to perform testing scenarios using the actual metadata label.  This seemed important given the age of the CIP PR (currently just over 6 months) and I didn't want them to wait forever for a metadata label.

This of course assumes it won't be disputed that CIP-0072 will need a metadata label (currently I think the CIP clearly calls for it), or won't be invalidated or deprecated somehow.

For the general case, what I generalised from prior `CIP-0010` requests was that labels can be added if a project or its representative makes a public case that they are needed, and that this doesn't have to be done with a CIP (many of our labels have no associated CIP).

Ryun1:
@rphair thanks for the great response; as always.

To me, this approach makes perfect sense, that fills in my gap in understanding - thanks.

matiwinnetou:
@Ryun1 @rphair @KtorZ - please merge this as well...

rphair:
yes, @Ryun1 @KtorZ please green-check this as well so we can merge it ASAP... I never would have guessed CIP-0072 was in danger of getting merged first 😅 which somewhat invalidates my argument in https://github.com/cardano-foundation/CIPs/pull/505#issuecomment-1518241010 but I guess you never know what will happen in the real world. 😜

Ryun1 (approved):

PR #455, opened 2023-01-30
CIP-0085? | Sums-of-products in Plutus Core

michaelpj:
This CIP proposes a substantial extension to Plutus Core to support sums-of-products, with the aim of realising some substantial performance improvements.

---

([rendered proposal in branch](https://github.com/michaelpj/CIPs/blob/mpj/sums-of-products/CIP-%3F%3F%3F%3F/README.md))

PR #507, opened 2023-04-20
CIP-0097? | Representing the Script Context as a SOP term

michaelpj:
This is the other part split out from https://github.com/cardano-foundation/CIPs/pull/455 that talks about changing the representation of the script context to be a SOP term rather than `Data`. It depends on CIP-85.

---

([rendered proposal in branch](https://github.com/michaelpj/CIPs/blob/mpj/script-context-sop/CIP-%3F%3F%3F%3F/README.md))

### DIFF
--- a/CIP-????/README.md
+++ b/CIP-????/README.md
@@ -1,5 +1,5 @@
 ---
-CIP: ????
+CIP: ?
 Title: Representing the Script Context as a Sums-of-Products Term
 Authors: 
   - Michael Peyton Jones <michael.peyton-jones@iohk.io>
@@ -21,11 +21,11 @@ This is slow to process, so we propose to use the support for sums-of-products (
 
 ## Motivation: why is this CIP necessary?
 
-## Background
+### Background
 
 This CIP assumes familiarity with the concepts described in CIP-85, including SOPs and the origin of `Data`.
 
-## The Problem
+### The Problem
 
 The problem is:
 1. It is slow to perform any kind of manipulation of the script context (see CIP-85 for benchmarking of operations on `Data` versus SOPs)
--- a/CIP-????/README.md
+++ b/CIP-????/README.md
@@ -1,0 +1,205 @@
+---
+CIP: ????
+Title: Representing the Script Context as a Sums-of-Products Term
+Authors: 
+  - Michael Peyton Jones <michael.peyton-jones@iohk.io>
+Implementors: 
+  - Michael Peyton Jones <michael.peyton-jones@iohk.io>
+Status: Proposed
+Category: Plutus
+Created: 2023-04-20
+Discussions: 
+  - https://github.com/cardano-foundation/CIPs/pull/455
+License: CC-BY-4.0
+---
+
+## Abstract
+
+Plutus scripts receive information about the transaction currently being validated by being passed an argument containing that information, called the "script context".
+The script context is currently represented in Plutus Core using the `Data` builtin type. 
+This is slow to process, so we propose to use the support for sums-of-products (SOPs) in Plutus Core (proposed in CIP-85) to represent the script context directly as a term using sums-of-products.
+
+## Motivation: why is this CIP necessary?
+
+## Background
+
+This CIP assumes familiarity with the concepts described in CIP-85, including SOPs and the origin of `Data`.
+
+## The Problem
+
+The problem is:
+1. It is slow to perform any kind of manipulation of the script context (see CIP-85 for benchmarking of operations on `Data` versus SOPs)
+2. Translating ("decoding") the script context into a native form ([Scott encoding][] or SOPs) is far too slow because of 1 and because script contexts become large
+
+These combine to make decoding real world script contexts prohibitive in many cases:
+budget profiling of examples in `plutus-use-cases` that naively decode the script context revealed that many were spending 30-40% of their budget just on the decoding step, which is unacceptable.
+
+In reality, this is well known among developers, and most have moved to _not_ decoding the script context, and instead working directly with the underlying `Data` object.
+This is slow and error-prone, but at least provides a workaround.
+
+## Specification
+
+A new Plutus ledger language L is added to the ledger.
+The Plutus Core language version 1.1.0 is both allowed and required in L (since the changes to the ledger-script interface will require sums-of-products). 
+
+The ledger-script interface changes in L as follows:
+- Instead of the script context argument being a `Data` object, it is now a Plutus Core term.
+- The translation of the script context into `Data` is replaced by a translation of the script context into a Plutus Core term.
+
+The full specification of the new translation is to be determined in collaboration with the ledger team, but broadly:
+- Integers and bytestrings are translated to Plutus Core integers and bytestrings instead of using the `I` and `B` Data constructors.
+- Constructors are translated to Plutus Core `constr` terms instead of using the Data `Constr` constructor.
+- Uses of the `Map` and `List` Data constructors will need more specific handling but will be translated in line with their underlying dataype representation in Plutus Core.
+
+## Rationale: how does this CIP achieve its goals?
+
+This CIP solves the problem in the most drastic way possible: it removes it entirely.
+With the proposed ledger interface changes the script context is passed across _already decoded_, so there is zero work for the script to do.
+It can immediately operate on it efficiently.
+
+### Discussion
+
+#### Is this still worth it if users aren't decoding the script context?
+
+The issues with decoding the script context only apply to users who actually try and decode it; many people have moved on to just using the underlying `Data` object instead.
+This is a totally reasonable workaround, but it would still benefit those users to have a SOP script context to operate on instead, because they can do so more efficiently that on `Data` (see CIP-85 for benchmarks of operations on `Data` versus SOPs).
+
+#### Why not also make datums and redeemers SOP terms instead of `Data`?
+
+It is always nice to get rid of things, and this CIP does _not_ let us get rid of `Data` as a Plutus Core builtin type, because `Data` is still used for redeemers and datums.
+That begs the question: could we also have redeemers and datums be passed as SOP terms instead of values of type `Data`?
+
+The problem is that, unlike the encoded script context, redeemers and datums are _user facing_.
+They appear in the block format, they can be supplied by users (e.g. directly at the CLI!), and they can be inspected in block explorers.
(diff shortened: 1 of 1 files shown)


L-as:
What is blocking us from simply making handling `Data` faster? Why is it so much slower than SOPs?

MicroProofs:
Now that I understand how SOPs will work in the vm. I'll make some examples comparing and contrasting using data in Aiken vs using sop in Aiken. That way I can show off use case costs. Gonna work on the these examples this week.

CIP-????/README.md
```diff
@@ -0,0 +1,205 @@
+---
+CIP: ????
```
Ryun1, line 2:
```suggestion
CIP: ?
```
Would you mind reordering the header to match the [CIP template](https://github.com/cardano-foundation/CIPs/blob/master/.github/CIP-TEMPLATE.md?plain=1#L1).

CIP-????/README.md
```diff
@@ -0,0 +1,205 @@
...
+
+## Abstract
+
+Plutus scripts receive information about the transaction currently being validated by being passed an argument containing that information, called the "script context".
+The script context is currently represented in Plutus Core using the `Data` builtin type. 
+This is slow to process, so we propose to use the support for sums-of-products (SOPs) in Plutus Core (proposed in CIP-85) to represent the script context directly as a term using sums-of-products.
+
+## Motivation: why is this CIP necessary?
+
+## Background
```
Ryun1, line 24:
```suggestion
### Background
```

CIP-????/README.md
```diff
@@ -0,0 +1,205 @@
...
+
+This CIP assumes familiarity with the concepts described in CIP-85, including SOPs and the origin of `Data`.
+
+## The Problem
```
Ryun1, line 28:
```suggestion
### The Problem
```

CIP-????/README.md
```diff
@@ -0,0 +1,205 @@
...
+
+The problem is:
+1. It is slow to perform any kind of manipulation of the script context (see CIP-85 for benchmarking of operations on `Data` versus SOPs)
+2. Translating ("decoding") the script context into a native form ([Scott encoding][] or SOPs) is far too slow because of 1 and because script contexts become large
+
+These combine to make decoding real world script contexts prohibitive in many cases:
+budget profiling of examples in `plutus-use-cases` that naively decode the script context revealed that many were spending 30-40% of their budget just on the decoding step, which is unacceptable.
```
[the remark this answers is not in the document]
KtorZ: Given that only PlutusTx (as far as I know?) is trying to translate the script context into some intermediate representation, I'd assume that yes. 

So it could be interesting to get some before/after benchmarks for solutions like Plu-ts or Aiken regarding the impact of using SOP vs Data. Though, if the script context still comes in the form of `Data` from the ledger, I believe that the overhead of turning a `Data` into a SOP will be still high.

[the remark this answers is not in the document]
michaelpj: Yes, this is really only for people who actually try to convert the script context into another form. Although I do still think this is _desirable_ - it's just obviously too expensive to be worth it at the moment. 

So the numbers here are more meant to indicate how unacceptably high the cost is, _not_ to argue that these are reductions that real users might see (the test cases in question are _not_ real). I can see that's misleading, I'll rephrase.

MicroProofs:
I like SOP for its fast destructuring, but the problem is the blocker on missing out equals data and serialise_data. As a different compromise how about a new builtin that is similar to case. Say case_constr.

Takes in a constant of type data that is a constr and any term. Then it applies all field arguments from constr plus the constr index to that term. Then that single term can handle constr index and whatever fields it has applies.

Now I know this step is computationally more expensive. But since we have a lot of builtins that operate on data, I wonder if operating on data in a more efficient way might be the solution. Even converting script context to SOP does not change that we still need better ways to operate on data because of datum, redeemer, and script context still has data fields in datums, inputs, outputs, and redeemers fields.

```
[[builtin case_constr (con data #D87A820540)] other_term]

```

this converts to 

```
[[[other_term (con integer 1)] (con data #05]  (con data #40]
```


you could imagine other_term is:

```
(lam constr_index (force [
    [
      [
        (force (builtin if_then_else)) 
        [[(builtin equalsInteger) (con integer 0)] constr_index]
      ]
      (delay (con bool True))
    ]

    (delay
        (force [
          [
            [
              (force (builtin if_then_else)) 
              [[(builtin equalsInteger) (con integer 1)] constr_index]
            ]
            (delay (lam field1
                (lam field2 
                    (con bool True)
                )
            ))
          ]
        ])
    )
  ])
)

```

michele-nuzzi:
not sure `case_constr` would turn more efficient than manual destructuring, especially for structs with many fields, but if it serves to keep only one representation I'm down to it

MicroProofs:
Typically I am seeing dApps that have use cases that check values, addresses, and datums of outputs against input. I'll come back and clarify some more in a bit but here's what I got.

Common Cases where I see SOP as much more efficient:

- Working with value, i.e. List<Pair<ByteArray, List<Pair<ByteArray, Int>>>>
- Pulling out fields out of a constr like pulling out outputs, inputs, etc.

Common Cases where I see SOP as less inefficient:

- Comparing addresses, like comparing an address that comes from a redeemer, datum, or input address, and comparing it to an output address. In this case especially if the address comes from a datum or redeemer. This is where equals data is really powerful
-  Comparing utxo references, i.e. transaction hash plus output index. Again comparing utxo references inputs to the Spend purpose field is much faster thanks to equals data. Also comparing a utxo ref to an output datum or output datum field that holds a utxo ref.


Cases where SOP has no effect:
Comparing redeemer/datum to output datums. This is a pain point where have a builtin like case_constr is nice since datum is still going to be data.

MicroProofs:
> not sure `case_constr` would turn more efficient than manual destructuring, especially for structs with many fields, but if it serves to keep only one representation I'm down to it

It would be more efficient than having to do head and tail list to get each constr field and get the constr index.

michele-nuzzi:
> Cases where SOP has no effect:
> Comparing redeemer/datum to output datums. This is a pain point where have a builtin like case_constr is nice since datum is still going to be data.

I would still argue comparing datums to output datum is more efficient

I usually do stuff like this
```ts
const datumPreserved = output.datum.eq(
  POutputDatum.InlineDatum({
    datum: myInputDatum
  })
)
```
of course that is the case only for datums that are inline

CIP-????/README.md
```diff
@@ -0,0 +1,205 @@
+---
+CIP: ?
```
```suggestion
CIP: 97
```

MicroProofs:
> (just [noted this](https://discord.com/channels/971785110770831360/992011119872970762/1116741037566722099) on Discord but probably better to address here)
> 
> @michaelpj would you come to our [next CIP meeting](https://hackmd.io/@cip-editors/68) - currently [scheduled](https://discord.com/events/971785110770831360/1115962065639714826) 20 June, 9:30 am UTC - to review this & invite any other Plutus representatives you think would have a stake in that issue? (cc @L-as)

Could I get an invite @rphair? The discord links don't work for me.

L-as:
I don't think we should do this. Being able to hash the script context is highly convenient for evaluating snarks on it, since you can use the hash of the script context as a circuit witness trivially this way. If you can't hash it cheaply, it gets an order of magnitude harder, which is not nice.

The real issue is that the Plutus interpreter is unperformant.

rphair:
@michaelpj @L-as @MicroProofs @michele-nuzzi this, with other Plutus items, is on the agenda for review early in the CIP editors' meeting Tuesday 04 July 16:00 UTC: https://hackmd.io/@cip-editors/69

effectfully:
> @michaelpj @L-as @MicroProofs @michele-nuzzi this, with other Plutus items, is on the agenda for review early in the CIP editors' meeting Tuesday 04 July 16:00 UTC: https://hackmd.io/@cip-editors/69

Michael is on leave, so I attempted to join instead of him, but it seems I wasn't let in. If you have any questions, I'd be happy to try to answer them, although I'm not sure what medium we should use.

rphair:
> @effectfully: but it seems I wasn't let in

Are you talking about "let in" to Discord?  If not, here's the invite link: https://discord.gg/kyaTyzkBqd ... if so, I don't see any problems reported in the #verification channel there... let's confirm you're using Discord first, then try to follow the event link (in less than 2 hours): https://discord.com/events/971785110770831360/1120631775404761139

effectfully:
> Are you talking about "let in" to Discord? 

No, I took a meeting from Michael's calendar. Sorry, I should've just asked here. Thank you for the clarification.

> let's confirm you're using Discord first, then try to follow the event link (in less than 2 hours): https://discord.com/events/971785110770831360/1120631775404761139

Unfortunately, in 2 hours from that moment (i.e. now in an hour) I'm going to be traveling likely without access to the Internet. I'll try to make it, but chances of that are not high.

rphair:
Tagging `Waiting for Author` due to @effectfully / @michaelpj report at last hour's CIP Editors' meeting that Plutus team is not currently sure this solution is preferable to alternatives they are also considering.

rphair:
@effectfully at your advice we can try to find some help to confirm, update as necessary & merge this... but without MPJ's original inspiration it seems this idea isn't going to move forward.  Maybe you or @L-as @MicroProofs @michele-nuzzi could give us a tip if anyone in the community has the vision & persistence to document this in a way that will be acceptable to the current Plutus team.  If so we'll remove the `Likely Abandoned` tag & work on moving this forward again.

On the off-chance that the Plutus team is doing this anyway, I'm tagging @zliu41 to indicate if this is so.

MicroProofs:
This was decided against due to the disadvantages that are brought by not having a Plutus Data encoded script context. A couple examples were using equalsData builtin for cheaper equality and using serializeData builtin for cheap access to hashing different parts of the script context extremely efficiently.

zliu41:
This has been superseded by configurable script interfaces. The latter is an early stage idea, not CIP has been written, and we have no plan to pursue it in the near future. Instead we are working on making it easy to work with BuiltinData.

As such this CIP should be considered obsolete.

For anyone interested in learning about configurable script interfaces, look [here](https://docs.google.com/document/d/1Wsbd1EXjD2-JB91pzBc51KDq1Hzzq087ocd1Kfwxkf8).

rphair:
Thanks @MicroProofs @zliu41 for the conclusion & references - therefore closing in the `Deprecated` state.

PR #509, opened 2023-04-27
CIP-0095? | Web-Wallet Bridge | Governance

Ryun1:
These definitions extend [CIP-30 | Cardano dApp-Wallet Web Bridge](https://github.com/cardano-foundation/CIPs/tree/master/CIP-0030) to provide support for [CIP-1694 | A First Step Towards On-Chain Decentralized Governance](https://github.com/cardano-foundation/CIPs/blob/master/CIP-1694/README.md) focussed web-based stacks. Here we aim to support the requirements of the Conway Ledger era, this specification is based on the [Draft Conway Ledger Era Specification](https://github.com/input-output-hk/cardano-ledger/tree/master/eras/conway/test-suite/cddl-files).

For additional discussion see [`wallets-sanchonet`](https://discord.com/channels/826816523368005654/1143258005354328156/1143272934966837309) channel in the [IOG Technical Discord](https://discord.gg/inputoutput) to view you have to opt-in to the SanchoNet group via the `start-here` channel. This channel should compliment the technical discussions on Github rather than replace them - I will ensure any major discussions on Discord are summarized here.

#### Major Changes
- [c481a50](https://github.com/cardano-foundation/CIPs/pull/509/commits/c481a50913ef289f2e9ad201babf581698d3d555) - Moved the burden of transaction/certificate construction from the wallet to client application.
- [6c83052](https://github.com/cardano-foundation/CIPs/pull/509/commits/6c83052722d3ee3bd3c0a1fc63c5fee6e25de209) - See [this comment](https://github.com/cardano-foundation/CIPs/pull/509#issuecomment-1636103821) for description of changes.
- [49c8da7](https://github.com/cardano-foundation/CIPs/pull/509/commits/49c8da7a96f5f9efbf82b60a1234392845499fa0) - Changed DRep key derivation path.
- [1f75f99](https://github.com/cardano-foundation/CIPs/pull/509/commits/1f75f990c4e8fdf308c3ed209bac723a84822931) - Aligned with [this Conway Draft Design commit](https://github.com/input-output-hk/cardano-ledger/blob/1beddd3d9f10d8fcb163b5e83985c4bac6b74be7/eras/conway/test-suite/cddl-files/conway.cddl) and added implementers.
- [d02b02f](https://github.com/cardano-foundation/CIPs/pull/509/commits/d02b02faf9733e3099f71fa2922f8de02fa1e0a3) - Split `.getActivePubStakeKeys()` into `.getRegisteredPubStakeKeys()` and `.getUnregisteredPubStakeKeys()` more info in [this comment](https://github.com/cardano-foundation/CIPs/pull/509#issuecomment-1711852962).

-----
[Rendered proposal in branch](https://github.com/Ryun1/CIPs/blob/governance-wallet-connector/CIP-0095/README.md)


CIP-XXXX/CIP-XXXX.md
```diff
@@ -0,0 +1,916 @@
...
+focussed web-based stacks. Here we aim to support the requirements of both Ada
+holders' and DReps' interactions with such web-based stacks.
+
+> **Note** This proposal assumes knowledge of the ledger governance model as outlined within
+> [CIP-1694?](https://github.com/cardano-foundation/CIPs/pull/380).
+
+## Motivation: why is this CIP necessary?
+
+CIP-1694 introduces many new concepts, entities and actors to Cardano;
+describing their implementation at the ledger level. Yet, for most eco-system
```
rphair, line 35:
```suggestion
describing their implementation at the ledger level. Yet, for most ecosystem
```
I wouldn't disqualify your using the hyphen but this is an irregular spelling of _ecosystem_ & the first time I've seen it in Cardano reference material:
- https://www.merriam-webster.com/dictionary/ecosystem
- https://en.wikipedia.org/wiki/Ecosystem
- https://cardano-foundation.github.io/state-of-the-developer-ecosystem/2022/
- https://landing.essentialcardano.io/guide-to-the-ecosystem
- https://developers.cardano.org/showcase/?tags=ecosystem

CIP-XXXX/CIP-XXXX.md
```diff
@@ -0,0 +1,916 @@
...
+discourage the creation of subsequent CIPs which may support a wider range of
+governance endpoints. Nor does this specification aim to discourage wallet
+providers from fully integrating CIP-1694 governance, side-stepping the
+necessity for clients (and this API).
+
+### Why Web-based Stacks
+
+Web-based stacks, with wallet connectivity, are a familiar place for users to be
+able to interact with Cardano. These tools lower the technical bar to be able to
+engage with the eco-system. Thus we believe encouraging further adoption of this
```
rphair, line 769:
```suggestion
engage with the ecosystem. Thus we believe encouraging further adoption of this
```
same as above

CIP-XXXX/CIP-XXXX.md
```diff
@@ -0,0 +1,916 @@
...
+### DRep Key
+
+We chose to introduce the concept of a DRep Key, building on top of CIP-1694,
+this we see as a necessary step for wallet implementors. By setting a
+(hierarchical) deterministic derivation path it enables restorability from a
+seed phrase.
+
+With this definition we aim to standard for tooling to be able to derive DRep
+credentials from mnemonics, not only those wallets who support this web-bridge.
+This standard brings all the benefits of the application of generic eco-system
```
rphair, line 852:
```suggestion
This standard brings all the benefits of the application of generic ecosystem
```
same as above

CIP-XXXX/CIP-XXXX.md
```diff
@@ -0,0 +1,916 @@
...
+unjustified expense. Future iterations of this specification may expand on this,
+but at present this is seen as unnecessary.
+
+> **Note** `1718` was the year that François-Marie adopted the pseudonym Voltaire.
+
+#### Tooling
+
+Supporting tooling should clearly label these key pairs as "CIP-???? DRep Keys".
+
+Bech32 prefixes of `drep_sk` and `drep_vk` should be used.
```
rphair, line 92: I don't see these in CIP-0005... should a CIP-0005 modification to add these prefixes therefore be included in this PR?

rphair (commented):
To me this looks future-proof enough to handle other governance systems than CIP-1694 (as long as the Ledger components are compatible).  I won't be able to confirm or dispute that _all_ technical details are in order, but this looks to be the result of solid development and cooperation.  

Even at this early review stage, I think it could already serve a guide for governance dApp devs to start development.

Since I can't do much more than confirm its consistency with CIP-0030, I would be happy to sign off on it as soon as the sections marked `// todo` are completed, to leave @KtorZ and/or @SebastienGllmt free to confirm the technical rigour.

I understand balancing this draft with the evolving https://github.com/cardano-foundation/CIPs/pull/380 will probably take as much time as the latter PR, and will look forward to reading updates here in the meantime. ✔️

CIP-XXXX/CIP-XXXX.md
Ryun1: I agree, good spot -- not sure where I got that from 🤠

CIP-XXXX/CIP-XXXX.md
Ryun1: Good spot @rphair, I have added placeholders in.

rphair: resolved in https://github.com/cardano-foundation/CIPs/pull/509/commits/65e34ef65214752adfaf9ac31d67b77297191184

flyerman (changes requested):
💡 Would it be preferable to move the building of the transactions out of the wallet and onto the application?  And provide a support library for building the TX alongside the new data types.

Reducing the burden on the wallet will improve adoption.  As well as keeping the wallet to its current primitives:  inspect, sign and submit.

The CIP extension would then become something like:
```
api.getDRepKey(): Promise<PubDRepKey>
api.getActiveStakeKeys(): Promise<PubStakeKey[]>

api.submitDelegation(tx): Promise<txHash>
api.submitDRepRegistration(tx): Promise<txHash>
api.submitDRepRetirement(tx): Promise<txHash>
api.submitVote(tx): Promise<txHash>
api.submitGovernanceAction(tx): Promise<txHash>
```

each `submitXXX(tx)` would inspect the transaction, present the relevant data to the user, sign and submit.

I expect the key material and way of signing to rarely change.  Unlike the transaction building part, governance flags, fields or parameters.  Those section of the protocol are more likely to evolve.  And that's why it could be more pertinent to place it in an application helper library for faster iterations and upgrades.

Whenever an application submit a transaction that follow a later protocol version than what the was implemented by the wallet, the wallet can still display the backward compatible portions of the transaction.  Indicates that the TX contains extra fields, possibly even display a unserialized representation of those new fields.  And the wallet could still provide the ability to sign and transmit granted proper disclaimer and checkboxes for user consent.

Which bring me to a major concern:
🚧 We are missing an API version.  That's essential for enabling a rolling deployment of newer versions of the API.
And eventually a set of rules for backward compatibility.

Ryun1:
Hey @flyerman thank you for your review/comment - much appreciated!

Following our discussion, I think moving the transaction construction burden from wallets to the application is a good idea. As you mentioned this brings a number of benefits:

- Reduces complexity for wallet implementers.
  - They do not need to know how to construct these new transactions/certs.
  - They do not need to include complex error handling.
- Keeps "the role of the wallet" to it's primitives of inspect, sign and submit.
  - We move application specific complexities to the application.
- Aids iterative updates, all wallet implementers do not need to update if the format of these transactions is adjusted.
  - By utilizing a common transaction building library and maintaining that we avoid the need for wallets to update their implementations.
  - This benefit is immediately relevant as these new certificates/transactions are being developed for the Conway era.

I will work on an update to this specification to match the described changes.

> We are missing an API version.

Yes, this was done intentionally, and perhaps uncommunicated within this specification. My intension for this design was to avoid some of the issues that were present with the versioning of CIP-0030. I aim to avoid the use of numbered versioning because I do not deem it necessary.

Whilst this CIP is in it's unmerged proposed state, it remains very fluid and substantial changes can happen, so I would advise against any implementation. Once more feedback is received, maturing this design I think implementations can emerge alongside this proposal's merger into the CIPs repo. Once merged only small necessary changes should be made, ideally in backwards compatible fashion. This alongside mature implementations should, move this proposal to an active state where changes cannot be made. If any large changes are needed once active then a new proposal should be made to replace this one. This I believe aligns with the (new) extendibility design of CIP-0030.

I would be keen to hear feedback on this "versioning" (or lack there of) approach. :sunglasses:

gitmachtl:
Why is there a need to introduce another derivation path for a voting key? We introduced `m / 1694' / 1815' / account' / chain / address_index` with CIP36. This is already implemented also in the HW-Wallets. You can simply use another index number if you wanna have more Voting-Keys. I don't see the need to introduce another derivation path for that tbh.

Ryun1:
Hey @gitmachtl thanks for great comment (as always :) ). These are good questions and I will add discussion of these to the rationale.

> I don't see the need to introduce another derivation path for that tbh.

This could be an alternative approach. The issue I see with this is that we would be forcing every tooling provider to support CIP-36 and since CIP-36 is not entirely Voltaire/CIP-1694 it could make things more confusing for people (confusing Catalyst and Voltaire).

> Is it just to differentiate from Catalyst?

Yup the primary reason I chose to separate the DRep key here from the CIP-36 Vote key used in Catalyst is so that both keys have clear and explicit usage. In essence, I am trying to avoid potential  mistakes from confusion. As (CIP-95) DRep keys are used for more than just voting. I know originally the intension was for the CIP-36 key to be used in both Voltaire and Catalyst, but due to Voltaire design decisions (CIP-1694) that did not happen unfortunetly.

> It really starts to get confusing i guess.

Yes I totally see the confusion, I have really tried with the naming here to clearly differentiate. But (unfortunately) we have DReps in both Catalyst and Voltaire - perhaps remaining the key here will help avoid confusion?

Furthermore as noted in the open questions, I think it may make sense to lift this DRep key from CIP-95 and place it into CIP-1852, since it is likely to be used in a much wider range of tools than CIP-36 keys. This may help it become a "first class key" similar to Stake and Payment keys. This is probably warranted as it this DRep key is represented in the Conway Ledger here.

> No mather what we come up with, there can't be any exclusion, so it must work for all kinds of wallets.

I completely agree, I would be more than keen to here feedback from CLI-Wallets and HW-Wallets. I have reached out to some for comments. Probably lifting this key from a software-wallet focussed spec may make sense?

> It also blows up the app on the hw-wallet more and more.

Yes, I am aware of this and its an issue/ design decision that needs to be addressed and we have been in contact with the relevant parties to discuss our approach in relation to Cardano's Age of Voltaire.

> Don't wanna blame anyone, but it becomes a bit of a pattern with CIPs imo.

Yes I have noticed this too, sometimes there is a lack of a larger vision (I think I could be guilty for sure)... I think I need to put on my CIP editor hat and pay more attention to this issue. I appreciate you calling this out.

gitmachtl:
> This could be an alternative approach. The issue I see with this is that we would be forcing every tooling provider to support CIP-36 and since CIP-36 is not entirely Voltaire/CIP-1694 it could make things more confusing for people (confusing Catalyst and Voltaire).
Only thing to thing about is to reuse the already implemented derivation path for a voting key. We had this discussion in the past to make it open as a general vote thing if you remember, and to remove all catalyst related references. Like they are also gone in the cardano-hw-cli and refered to just as "vote"
```
cardano-hw-cli vote registration-metadata
usage: cardano-hw-cli vote registration-metadata [-h] [--mainnet [NETWORK]] [--testnet-magic [NETWORK]] [--vote-public-key-jcli VOTEPUBLICKEYS] [--vote-public-key-string VOTEPUBLICKEYS] [--vote-public-key-hwsfile VOTEPUBLICKEYS]
                                                 [--vote-public-key-file VOTEPUBLICKEYS] [--vote-weight VOTEWEIGHTS] --stake-signing-key-hwsfile HWSTAKESIGNINGFILEDATA --payment-address PAYMENTADDRESS --nonce NONCE
                                                 [--voting-purpose VOTINGPURPOSE] [--payment-address-signing-key-hwsfile PAYMENTADDRESSSIGNINGKEYDATA] --metadata-cbor-out-file OUTFILE [--derivation-type DERIVATIONTYPE]
                                                 {} ...
```

But i understand your argument, its just a bit strange too me to reinvent the wheel again and again. The derivation path for CIP36/62 with the `1694` was choosen to do it, thats why we have the 1694 in it 😄 

> Yes I have noticed this too, sometimes there is a lack of a larger vision (I think I could be guilty for sure)... I think I need to put on my CIP editor hat and pay more attention to this issue. I appreciate you calling this out.

No you're not more guilty than me or any other one. But i noticed it. Some developments are slow, some are fast. Some are overtaking others. Its just i thought that especially with voting keys, we were at a point to say: "ok, you derive your voting key via this derivation path". And thats it, how to use that key is another context. From the enduser pov, its confusing. Because why can't i use my Voting Key i use to vote on Catalyst also for Voltaire Votings, etc...   

I am sorry if my text made you feel that i would attack you, thats not the case at all. Just wanna bring up thoughts i have.

Ryun1:
> We had this discussion in the past to make it open as a general vote thing if you remember, and to remove all catalyst related references.

Yes, I do remember, this is a good point. If I remember correctly I think this was partly due so other organisations beyond Catalyst could use the CIP-36 standard. But yup, this is a good point.

> The derivation path for CIP36/62 with the 1694 was choosen to do it, thats why we have the 1694 in it :smile:

Yes I know! seems quite convenient.

> I am sorry if my text made you feel that i would attack you, thats not the case at all. Just wanna bring up thoughts i have.

Not at all @gitmachtl - I'm always keen to hear your feedback 😎.

Ryun1:
As discussed in CIP Editors Meeting 68 a review from yourself @KtorZ would be appreciated.

CIP-0095/README.md
```diff
@@ -0,0 +1,678 @@
...
+These are used by the client to identify the user's on-chain CIP-1694
+interactions. Here we allow for multiple stake credentials to provided at once
+for the case of
+[multi-stake key wallets](https://github.com/cardano-foundation/CIPs/tree/master/CIP-0018).
+
+##### Returns
+
+An array of the connected user's active stake keys.
+
+#### `api.submitVoteDelegation([tx: UnsignedTransaction, stakeKey: PubStakeKey]): Promise<SubmittedTransaction[]>`
```
flyerman, line 287: ❔ Would it be pertinent to allow voting with all your stake at onces?  using all stake keys?
Perhaps leaving `stakeKey` null could signal this intention and the wallet would include them all.

CIP-0095/README.md
```diff
@@ -0,0 +1,678 @@
...
+the certificates.
+
+##### Returns
+
+This returns an array of `SubmittedTransaction` objects which contain the
+details of the submitted delegation certificates, for the client to confirm. The
+returned `txHash`s can be used by the client to track the status of the
+transactions containing the certificates on Cardano.
+
+#### `api.submitDRepRegistration(tx: UnsignedTransaction, dRepKey: PubDRepKey): Promise<SubmittedTransaction>`
```
flyerman, line 322: 🎨 Given that a wallet have only one drep key, could you make `dRepKey` optional ?
When null, the wallet would automatically use the one drep key or return an error if none exists.

The same would apply to `submitDRepRetirement()` and `submitVote()`

flyerman (commented):
NIT: see comments below

CIP-0095/README.md
```diff
@@ -0,0 +1,678 @@
...
+
+#### `api.submitGovernanceAction(tx: UnsignedTransaction): Promise<SubmittedTransaction>`
+
+Errors: [APIError](#extended-apierror), [`TxSignError`](#extended-txsignerror)
+
+This endpoint requests the wallet to inspect, sign and submit a transaction
+containing a governance action. The wallet should articulate this request from
+client application in a explicit and highly informative way. For the governance
+action the user must be shown the amount of ADA to be locked as deposit,
+governance action type and the metadata anchor.
```
Thomas-Upfield, line 423: This is comprehensive for the Info and Motion of No-Confidence governance action types, however all of the other governance action types require some form of input from the  user submitting the governance action (the Proposal Submitter). For example in the case of a Treasury Withdrawal a Proposal Submitter would need to provide an amount (in Lovelace) and an address where that amount would be sent if the governance action were to be enacted. 

I think it would improve the security of the process if these input names and values were also displayed to the Proposal Submitter at this point by the wallet so that the Proposal Submitter could verify that an app creating such a transaction was submitting the correct/expected information to the wallet.

CIP-0095/README.md
Ryun1: Making `dRepkey` optional or even removing it seems sensible, I will amend. The reason I kept it there was just to match the flow of Vote Delegation (pass transaction and tell wallet exactly what key to use).

CIP-0095/README.md
Ryun1: This is an interesting idea, but I don't see the benefit of removing stakeKey from the call as it's inclusion gives the dApp an explicit way to be able to track the governance state of each key.

CIP-0095/README.md
Ryun1: Good suggestion, would add burden to wallet implementers but I think it is likely necessary.

Ryun1: Added in [9af3bf0](https://github.com/cardano-foundation/CIPs/pull/509/commits/9af3bf045daf57bcb46116b69f446db685fb9a79).

mirceahasegan:
What is the expected behavior if web apps use `CIP30 signTx/submitTx` to sign a transaction with governance-related certificates?
This could be a problem since CIP30 implementation might need to be updated to inspect the transaction for these certificates; otherwise, the user would approve it without being aware of the governance contents.

CIP-0095/README.md
mirceahasegan: - `stakeKey` arg might be redundant here. The Vote Delegation Certificate from the unsigned transaction already has a stakeKey, so it can be inferred (and should be inspected anyway to be validated by the user).
- If the above is true, then a single arg `submitVoteDelegation(tx: UnsignedTransaction)` would be sufficient, addressing voting delegation by adding multiple certificates in the same transaction

MichalPetro:
@Ryun1 I would like to second gitmachtl's request not to create a new derivation path mainly for the sake of HW wallets (which are used by the top holders). Adding stuff like this is always causing frowns by Ledger and Trezor and also because we are already pushing too far with the HW wallets code size (on both Ledger and Trezor).

CIP-0095/README.md
```diff
@@ -0,0 +1,771 @@
...
+the certificates.
+
+##### Returns
+
+This returns a `SubmittedTransaction` object which contains all the details of
+the submitted registration certificate, for the client to confirm. The returned
+`txHash` can be used by the client to track the status of the transaction
+containing the certificate on Cardano.
+
+#### `api.submitDRepRetirement(tx: UnsignedTransaction): Promise<SubmittedTransaction>`
```
refi93, line 343: If these are regular cardano transactions, couldn't the DApp use the basic `signTransaction()` call? The logic required to verify that the unsigned transaction has the required structure should be not significantly different from any existing parsing logic of the wallets to show a user-friendly summary of the transaction signed via the CIP-30 call.

The Same applies for `submitDRepRegistration` call.

Ryun1: Also, I am trying to avoid issues with mismatch between CIP30 implementations - its unknown how existing CIP30 endpoints would be able to interact with CIP1694/Conway Ledger era items. I discuss this in [CIP-30 Reuse](https://github.com/Ryun1/CIPs/blob/governance-wallet-connector/CIP-0095/README.md#cip-30-reuse).

CIP-0095/README.md
```diff
@@ -0,0 +1,771 @@
...
+
+```ts
+type UnsignedTransaction = string;
+```
+
+A hex-encoded string representing a CBOR transaction that is completely unsigned
+(has an empty transaction witness set).
+
+This data model is used to represent transactions which contain, for example:
+DRep registration certificates and the client wishes the wallet inspect, add
```
refi93, line 158: are these certificates supposed to be included in the transaction as metadata or will there be a dedicated field in the Cardano transaction body?

If the former, rather than the "UnsignedTransaction" I'd expect there to be an "Unsigned metadata" property representing the unsinged vote delegation, drep registration etc. metadata.

If the latter (i.e. the governance actions are part of the Cardano tx body), I'd expect the CIP-30 `submitTransaction()` call to already cover the use case as it already should be able to work with any raw transaction body (as long as the underlying wallet has the keys to properly sign it)

Are there already any specs for the respective transactions (vote delegation, drep registration, ...)? I couldn't find them in this CIP but maybe I just missed that

CIP-0095/README.md
refi93: I think the answer to my question above: https://github.com/cardano-foundation/CIPs/pull/509/files#r1260839044 would answer also this question

Ryun1:
@mirceahasegan 

> What is the expected behavior if web apps use CIP30 signTx/submitTx to sign a transaction with governance-related certificates
I don't know, I believe it would be dependent on the CIP30 implementation. My assumption is that signTx would be unpredictable for Vote Delgations (signed by stake credential) and likely would not work for signing votes, DRep registration and DRep retirement as these are to be signed using DRep keys (vote credentials).

This is not ideal, do you have any ideas to address this problem?

>  CIP30 implementation might need to be updated to inspect the transaction for these certificates
I dont see this as feasible to do, I think this would only increase existing issues with mismatch between wallet and web app CIP30 capabilities. Requiring CIP30 implementer I think is a mistake.

CIP-0095/README.md
Ryun1: See draft [Conway era ledger spec](https://github.com/input-output-hk/cardano-ledger/blob/master/eras/conway/test-suite/cddl-files/conway.cddl)

Governance actions and votes are added in new transaction fields. Vote delegation, DRep registration and DRep retirement are new certificates.

CIP-0095/README.md
Ryun1: See answer [comment](https://github.com/cardano-foundation/CIPs/pull/509#discussion_r1260873125).

CIP-0095/README.md
Ryun1: Good point @mirceahasegan !

CIP-0095/README.md
refi93: I see - so far when new fields were added to txs, wallets started parsing them as any DApp can start using them and it's in the interest of wallets to work with as many DAps as possible and having a call to sign/submit transactions within CIP-95 that is basically just a limited version of the CIP-30 call seems redundant to me and would actually be additional work for us as wallet developers as apart from the parsing alone to show the tx details, we'd also need to implement additional validation to reject non-compliant txs. So rather than adding a new "sign/submit transaction" method, I'd suggest specifying in this CIP that wallets supporting the CIP-95 extension need to be able to sign/parse conway-era transactions

When it comes to certificates, I'd consider generalizing the individual registration calls to just "signCertificateTx()" which could take one (or more, if it makes sense) certificate(s), parse them (as wallets already parse transactions) and produce a valid transaction containing the signed certificate(s), which should also give room for easier introduction of new kinds of certificates in the future without increasing the API surface, just the extension number would be increased to signal support of certain txs/certificates, or a "specVersion" field introduced to the CIP-95 extension API, to signal to the DApp what certificates/txs the wallet is supposed to support. What do you think?

refi93: the latter ("signCertificateTx") may be a bit tricky as it may not clear how to represent the unsigned certificate in all cases, but extending the signTransaction call in CIP-30 should be IMHO a reasonable expectation from the wallets which should actually be less effort than the proposed dedicated calls

Ryun1: Okay, so instead CIP95 could do its own `signTx` which replaces CIP30's `.signTx` to provide support for Conway Tx/certs? Would you want to support all types of certificate that are being introduced in Conway ledger era OR just the ones needed for Ada Holders/DReps as I have focussed on?

refi93: I just checked and the certificates are indeed part of the transaction body for the conway era: https://github.com/input-output-hk/cardano-ledger/blob/master/eras/conway/test-suite/cddl-files/conway.cddl#L260 and I understand the CIP-30 signTransaction call to be era-agnostic, it just takes the raw tx and does whatever is needed to sign/show it.

In the past, as the Alonzo and Babbage eras were launched, the wallets' CIP-30 signTransaction calls evolved naturally with them (naturally incetivized by the DApps wanting to leverage the new functionalities) and I don't see why wouldn't that be the case with Conway, too, especially for any wallets that would commit to support the CIP-95 extension, so IMHO it's enough to specify what transaction fields/types the wallet must handle in its CIP-30 signTransaction call to be "CIP-95" compliant without resorting to specifying a whole new "cip95" signTx call which would effectively be just an alias for the CIP-30 one, if I'm not missing anything

gitmachtl:
@Ryun1 is currently the "dataSign" part of CIP30 anywhere affected by this CIP?

CIP-0095/README.md
Ryun1: My main concern with “evolved naturally” is the lack of versioning resulting in a mismatch between expected wallet capabilities and actual capabilities. Using CIP-95 as a Conway version could make sense?

Ryun1:
> @Ryun1 is currently the "dataSign" part of CIP30 anywhere affected by this CIP?

Not currently, unless we want to provide an updated version that supports signing with the new voting credentials

gitmachtl:
ok, because there is a CIP30/8 dataSigning catalyst proposal for hw-wallets on the way...

Ryun1:
**2023.07.13 Age of Voltaire Wallets + Tooling Hackathon**
CIP-95 Design Discussion Outcomes:

Move DRep key definitions into a separate CIP
- Change
  - Remove DRep Key definitions from CIP-95 and place into a new CIP.
- Motivation
  - CIP-95 should be a solely focussed on being a wallet web bridge API.
  - It is cleaner for this proposal to move the DRep Key definitions to a new CIP.
  - There are likely other credential definitions for Conway Ledger era that will be required.

Generalise the transaction signing endpoints
- Change
  - Replace the singular prescriptive sign and submit endpoints with an updated CIP-30 `.signTx()`.
  - The CIP-95 `.signTx()` should support all of Conway Ledger era artefacts, not just the ones required for Ada Holders and DReps.
- Motivation
  - This change prevents the need for future proposal to support the Conway artefacts.
  - This allows CIP-95 connection to act as a versioning so client dApps can tell if a connected wallet supports Conway Ledger era.
  - This is a more straight forward design, inheriting flows from CIP-30.

Remove prescriptive transaction data inspection
- Change
  - Remove stipulations on what data must be shown to users at time of transaction signature.
- Motivation
  - By generalizing the sign endpoints we loose the ability to be so prescriptive.
  - It is a wallet decision on what data should be shown to users.

Add an updated version of `.signData()`
- Change
  - Replace CIP30's `.signData()` to support signature using Conway voting credentials.
- Motivation
  - This is a way to prove ownership of voting credentials, which is a useful add for client applications.
  - Matches the flow utilized in CIP-30.

CIP-0095/README.md
refi93: I'd consider a valid way of "versioning" the wallet's API the exposure of the "CIP-95" extension itself which would signal the readiness of the CIP-30 signTransaction call to the respective wallet. Can you clarify what additional benefit would the signTransaction call within the CIP-95 extension have?

My concern is that by haveing a separate "signTx" call in the CIP-95 extension, we'd end up with two different calls which would do the same thing which would only result just in ambiguous implementations of the signing of transactions by DApps (some would go with the CIP-30 call, e.g. due to its use in the past, some would use the "CIP-95" call) and wallet developers may also be confused if the CIP-30 call should be any different from the CIP-95 one, which would be counterproductive overall.

I'd suggest having in this CIP specs of the exact transaction types, ideally even with examples, that wallets need to support to consider themselves "CIP-95" compliant (which the wallet would signal to the DApp with the CIP-95 extension flag being exposed), IMHO that would add most value both to DApp and wallet developers and remove the ambiguity that we might worry about regarding the lack of standardized versioning of the CIP-30 API itself

gitmachtl:
Change:
* Replace CIP30's .signData() to support signature using Conway voting credentials.

I wasn't able to join the live video feed yesterday. What does this exactly mean? CIP30s signData is using payment or stake secret keys for the signing. And a address is given which must be related to the secret key. What are the planed changes using Vonway voting credentials?

I am asking, because there is a Catalyst proposal to implement CIP30 dataSigning on hw-wallets according to the CIP8 COSE-Sign1 standard.

Ryun1:
@gitmachtl 

I apologise we had a few unfortunate permission issues on Discord. 

> CIP30s signData is using payment or stake secret keys for the signing

What was proposed is extending this to also support signing using [voting credentials](https://github.com/input-output-hk/cardano-ledger/blob/master/eras/conway/test-suite/cddl-files/conway.cddl#L321). Specifics of this are TBD and I will be working on defining it more thoroughly this week.

CIP-0095/README.md
Ryun1: > Can you clarify what additional benefit would the signTransaction call within the CIP-95 extension have?

For me, beyond the simplicity, this prevents the need for other proposals to support the Conway artefacts which this proposal lacked, for example: Constitutional Committee member hot/cold key registration certificates.

> I'd suggest having in this CIP specs of the exact transaction types, ideally even with examples, that wallets need to support to consider themselves "CIP-95" compliant (which the wallet would signal to the DApp with the CIP-95 extension flag being exposed), IMHO that would add most value both to DApp and wallet developers and remove the ambiguity that we otherwise might worry about regarding the lack of standardized versioning of the CIP-30 API itself

This is great suggestion, I fully agree that this, I will likely create a test vectors file with such details similar to the [test vectors included in CIP-36](https://github.com/cardano-foundation/CIPs/blob/master/CIP-0036/test-vector.md).

gitmachtl:
thx @Ryun1 , thats fine. but than there just needs to be a clarification about the address part in cip 30 dataSigning. because currently there must be an address in the payload that is related to the signing key. so a bit unclear right now to me what will happend to that field if signing with a voting key.

CIP-0095/README.md
```diff
@@ -0,0 +1,645 @@
...
+
+Here we describe DRep Key derivation as it attains to Cardano wallets who follow
+the
+[CIP-1852 | HD (Hierarchy for Deterministic) Wallets for Cardano](https://github.com/cardano-foundation/CIPs/blob/master/CIP-1852/README.md)
+standard.
+
+To differentiate DRep credentials from other Cardano keys the derivation path
+must follow:
+
+`m / 1718' / 1815' / account' / chain / address_index`
```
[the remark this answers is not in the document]
gitmachtl: I would favour `m / 1694' / 1815' / 0' / 1718 / 0` :smiley:
The hardened path is already voting related and cleared for hw-wallets too.

@Ryun1 its like payment and staking keys. totally different, but they are using the same hardened path, only difference is the chain index 0 vs. 2. So i think its would be ok to also reuse existing hardened paths and alter the chain index for such a purpose.

Ryun1:
@gitmachtl 

> so a bit unclear right now to me what will happend to that field if signing with a voting key.

I’ll probably just do a key hash, IMO no need to be address based.

gitmachtl:
> @gitmachtl
> 
> > so a bit unclear right now to me what will happend to that field if signing with a voting key.
> 
> I’ll probably just do a key hash, IMO no need to be address based.

wallets are using it to do the assignment which wallet to sign with with the given address in the cip30 dataSign call. so that would not work imo.

CIP-0095/README.md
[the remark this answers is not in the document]
disassembler: I'm seconding @gitmachtl suggestion. I'm not sure why we deviated from 1852' in the first place, but if we've already established precedence that voting keys are in 1694', we might as well run with it. account index should be hardened though, e.g. `/ 1694' / 1815' / 0' / 1718 / 0`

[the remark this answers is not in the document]
mmahut: As far as I know, any change to the Ledger code base now require an audit and Trezor does not do any audits on Cardano on any change.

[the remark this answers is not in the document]
Ryun1: @gitmachtl @AndrewWestberg @disassembler 

Thank you for your suggestions, I will make the change to:
`m / 1694' / 1815' / 0' / 1718 / 0`

The `1718'` in the purpose field change was inspired by how [CIP-36 vote keys are derived](https://github.com/cardano-foundation/CIPs/tree/master/CIP-0036#derivation-path) (used in Catalyst), but I agree that reusing the purpose of `1694'` from that makes sense.

Ryun1:
@gitmachtl 

Please see the changes I have included in [121ebf2](https://github.com/cardano-foundation/CIPs/pull/509/commits/121ebf2aa22f0a4ab9c0005409db3f7ead072656), I have just added in an idea for how to do this [here](https://github.com/Ryun1/CIPs/blob/governance-wallet-connector/CIP-0095/README.md#supported-credentials) - let me know your thoughts.

CIP-0095/README.md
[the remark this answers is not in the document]
Ryun1: @gitmachtl @AndrewWestberg @disassembler

What do you think of the derivation path:

`m / 1852' / 1815' / account' / 3 / index`

And move this definition into [CIP-1852 | HD Wallets](https://github.com/cardano-foundation/CIPs/blob/master/CIP-1852/README.md)? This would move this key to sit as a _first class_ credential next to Staking Key, which to me seems appropriate.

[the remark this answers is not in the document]
AndrewWestberg: I'm ok with that too. The only thing I am again is brand new hardened path elements which trigger required audits for hardware wallets.

gitmachtl:
> let me know your thoughts.

> To construct an address for DRep Key, the client application should construct a
> [type 6 address](https://github.com/cardano-foundation/CIPs/blob/master/CIP-0019/CIP-0019-cardano-addresses.abnf#L7C8-L7C93).
> Using an appropriate
> [Network Tag](https://github.com/cardano-foundation/CIPs/blob/master/CIP-0019/CIP-0019-cardano-addresses.abnf#L13)
> and a hash of the DRep Key.

so the idea is to generate kind of a pseudo address for a dRep key, using it like a payment key? if so, should we think about a new bech prefix for such addresses?

> The payload is not hashed and no `external_aad` is used.

this depends on the size of the payload, because with the hw-wallet, this is super limited. in the current catalyst proposal to att cip-8/30 support for the hw-wallets its been left open. so this might end up to use hashed payload for hw-wallets to fit into the memory limitations.

what are the plans/thoughts on the dRep key registration? will this be part of conway itself and not outsourced like now with catalyst and cip15/36? we should think about the whole workflow also from the pure cli perspective. 

where is the payload coming from which should than be signed with the dRep key and cip-95?

Ryun1:
@gitmachtl 

> so the idea is to generate kind of a pseudo address for a dRep key, using it like a payment key?

Yes, but I dont really like this approach feels hacky and Im unsure of potential security issues of making such a "pseudo address".  Perhaps it would just be easier to pass in the key hash, which under CIP-1694 is called a [DRep ID](https://github.com/cardano-foundation/CIPs/blob/master/CIP-1694/README.md?plain=1#L386) and I have added Bech32 for this. I'm not sure why a network tag would be needed for this endpoint?

> what are the plans/thoughts on the dRep key registration? will this be part of conway itself and not outsourced like now with catalyst and cip15/36? we should think about the whole workflow also from the pure cli perspective.

[DRep registration](https://github.com/input-output-hk/cardano-ledger/blob/master/eras/conway/test-suite/cddl-files/conway.cddl#L307C1-L307C14) is a new certificate under Conway (among a few others), similar to [stake key](https://github.com/input-output-hk/cardano-ledger/blob/master/eras/conway/test-suite/cddl-files/conway.cddl#L284) registration. 

> where is the payload coming from which should than be signed with the dRep key and cip-95?

The motivation for this is two fold; one is to match upgrading of CIP-30's `.signTx()`, two is to be able to provide a method for wallets to prove ownership of the public DRep Key presented via `.getPubDRepKey()`.

> > The payload is not hashed and no external_aad is used.

> this depends on the size of the payload, because with the hw-wallet, this is super limited. in the current catalyst proposal to att cip-8/30 support for the hw-wallets its been left open. so this might end up to use hashed payload for hw-wallets to fit into the memory limitations

I just lifted this part straight from CIP-30, I am yet to dig into the Catalyst proposal. The only change to CIP-30 I am trying to support here is signing with one more key type (DRep).

gitmachtl:
@Ryun1 
> > so the idea is to generate kind of a pseudo address for a dRep key, using it like a payment key?
> Yes, but I dont really like this approach feels hacky and Im unsure of potential security issues of making such a "pseudo address". Perhaps it would just be easier to pass in the key hash, which under CIP-1694 is called a [DRep ID](https://github.com/cardano-foundation/CIPs/blob/master/CIP-1694/README.md?plain=1#L386) and I have added Bech32 for this. I'm not sure why a network tag would be needed for this endpoint?

So this is CIP-95 right, and not base CIP-30 anymore. The test/verification of the CIP8 signed data is also done against the included address. That address is included in hex format, and the check is simply done by checking if the blake2b-224 hash of the publickey is included in the hexstring of the given address. Changing the address field to be filled with the DREP ID - which should be the blake2b-224 of the credential publickey - should solve it imo. Using the bech representation starting with `drep_id` can than be used on the tool side for input/output without the need to use the hex drep id directly.

I think the only thing that light wallets would have to do is to store also that drep_id to make the requested signing assignment correctly. It would also make sense to use it,  because hw-wallets could show the drep_id in bech format on the hw-gui. and not a pseudo address that knowbody knows where its coming from and what to do with it.

> > > what are the plans/thoughts on the dRep key registration? will this be part of conway itself and not outsourced like now with catalyst and cip15/36? we should think about the whole workflow also from the pure cli perspective.
> 
> [DRep registration](https://github.com/input-output-hk/cardano-ledger/blob/master/eras/conway/test-suite/cddl-files/conway.cddl#L307C1-L307C14) is a new certificate under Conway (among a few others), similar to [stake key](https://github.com/input-output-hk/cardano-ledger/blob/master/eras/conway/test-suite/cddl-files/conway.cddl#L284) registration.

ok, and the used credentials are basically the same as we use here.

gitmachtl:
about the network-id, are there different bech versions for mainnet and testnet? for the address field we have this differentiation, and the network-id (mainnet or testnet) was just used as a second layer to check that the values are correct and that not a testnet-address is used for mainnet f.e. i don't think we need that check here if there are simply no different versions of the governance keys/bech available.

Ryun1:
@gitmachtl 

> Changing the address field to be filled with the DREP ID - which should be the blake2b-224 hash of the credential publickey - should solve it imo.

I agree, this is a more sensible approach I will make this change, I really appreciate the input  🙏


about the network-id, are there different bech versions for mainnet and testnet? for the address field we have this differentiation, and the network-id (mainnet or testnet) was just used as a second layer to check that the values are correct and that not a testnet-address is used for mainnet f.e. i don't think we need that check here if there are simply no different versions of the governance keys/bech available. the network is encoded into the hex representation of the address.

> also the kind of address payment base, script base, payment script, script script, payment pointer, script pointer, payment enterprise, script, stake, stake script...
> are there thoughts to produce different "tagged" hex representations of the drep id according to the network or if the credentials are not normal keys but scripts or plutus stuff?

I really haven't put much consideration in to this yet (the bech I included in this PR dont care about network), CIP-1694 does not stipulate a network tag for DRep IDs but if the community thinks it would be good, we can ask to add. I think there is a larger conversation on "Conway Credentials" to be had here, I think this should be moved to a different proposal - I will pursue this.

Ryun1:
This proposal is on [the agenda](https://hackmd.io/@cip-editors/70) for CIP Editors Meeting #70 (Aug 1st, 4 PM UTC) held on [CIP Editors Meeting Discord](https://discord.gg/EU9jkX7r). 

If interested please attend, as I will recap the latest discussions and changes to this proposal.

CIP-0095/README.md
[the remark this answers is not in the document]
gitmachtl: @Ryun1 i think staying with the 1852'/1815' path would be ok. people should not mix there normal and there dRep wallet in first place imo. as @disassembler also loves to see it integrated in the 1852'/1815' path i think we should take that route. no new path for the hw wallets, which also means no conflicts maybe with other chains.

gitmachtl:
> I really haven't put much consideration in to this yet (the bech I included in this PR dont care about network), CIP-1694 does not stipulate a network tag for DRep IDs but if the community thinks it would be good, we can ask to add. I think there is a larger conversation on "Conway Credentials" to be had here, I think this should be moved to a different proposal - I will pursue this.

The hash of the vkey for both mainnet and testnet resolves to the same hex data. But there is a leading byte inserted which defines the type and network before the whole hex string is converted into bech format. 

The question is if such a concept will also be used for the dRep-ID. As stated here (https://github.com/cardano-foundation/CIPs/blob/db70d91029a6512934d11c7eb4436acabe5db182/CIP-1694/README.md?plain=1#L381-L386) the dRep ID can come from a verification key or from a native script or plutus script. So it would be possible - if needed - to also add a byte at the front of the hash to specify the type before it gets converted into bech format. 

Maybe i am overthinking it, but there are a few things in the past that could have been solved better if we had included a single byte more in the design phase. :smile:

gitmachtl:
I have opened up a separate issue about extra stuff here https://github.com/cardano-foundation/CIPs/issues/564

gitmachtl:
@Ryun1 didn't we settle now on the dRep key derivation path `1852'/1815'/account'/3/index` ?

gitmachtl:
@Ryun1 is the key derivation for a "Constitutional Committee member" the same as for a dRep?

Ryun1:
Discussed the recent changes and open questions today in CIP Editors Meeting 70. No objections to:
- Adding DRep Key derivation to CIP-1852.
- Proposed design of `.signData` being able to be passed DRepID and Addresses.

Ryun1:
After a review and some thinking, I will include DRep Key in [CIP-1852](https://github.com/cardano-foundation/CIPs/tree/master/CIP-1852), but it will also get it's own CIP matching how stake keys were rolled out, see [CIP-0011](https://github.com/cardano-foundation/CIPs/tree/master/CIP-0011).

CIP-0095/README.md
```diff
@@ -0,0 +1,1005 @@
...
+#### Tooling
+
+Supporting tooling should clearly label these key pairs as "CIP-95 DRep Keys".
+
+Bech32 prefixes of `drep_sk` and `drep_vk` should be used, as described in
+[CIP-0005](https://github.com/cardano-foundation/CIPs/blob/master/CIP-0005/README.md).
+
+Examples of acceptable `keyType`s for supporting tools:
+
+| `keyType`                          | Description           |
```
disassembler, line 138: Lets keep this simple and mirror what we already have for other derived keys, e.g.:
`DRepExtendedSigningKey_ed25519_bip32`

The descriptions can be whatever, but we shouldn't distinguish in the types for HW vs non-hw in the vkeys (and hw signing would never be saived to disk anyways)

Ryun1: Check [fbc5fcb](https://github.com/cardano-foundation/CIPs/pull/509/commits/fbc5fcbb127313ccfd2a30376145f63627f3afd9) -- this should amend.

rphair (approved):
Based on recap of everything so far in the currently ongoing CIP meeting I would be happy to see this go ahead.  There is a plan in progress where a dRep key definition may be moved to a separate CIP but it's marked as "todo" in the document and implementations for SanchoNet can proceed without that separation & I believe we shouldn't wait for that before merging this.

Crypto2099 (approved):
I believe that this CIP is ready and acceptable for merging as written. I have suggested that we again reiterate that the key derivation for dRep keys will be moved to its own separate CIP process and so integrators should be mindful that the key derivation paths are subject to change but otherwise we need to go ahead and get wallets implementing this logic so that we can push SanchoNet and the Voltaire era forward for the everyday users.

rphair (approved):
Also would add that https://github.com/cardano-foundation/CIPs/pull/509#pullrequestreview-1633830056 is a good summary of the consensus at today's CIP Editors' meeting.

PR #510, opened 2023-04-28
CIP-0094 | Procedure and duration

gufmar:
describe the ballot cast + redelegation epoch sequence


rphair:
hey @gufmar I'm not sure what to make of this.  Normally there's a requirement for collaborator reviews, but somehow not in this case.  And it shows as a delta to `./CIP-0094` which hasn't been merged yet (https://github.com/cardano-foundation/CIPs/pull/496).  

This looks like a sensible change (@KtorZ?) but if merged as-is I think we'd have that scrap of the `README.md` and that extra file dangling there in the public repository.  What would you like us to do?

PR #513, opened 2023-05-10
CIP-85: add an alternative about unsaturated constr/case

michaelpj:
We just discussed this in the team and I thought it was worth recording.


rphair (approved):
This is far above my level but I can see it fits in well with the adjacent material.

Ryun1 (approved):
This is also beyond my understanding, but seems sensible and is well written so I will approve on that basis.

PR #523, opened 2023-06-09
CIP-0068 | Bump version for RFT

### LINKED ISSUE #520: Versioning alternatives
So far the CIP editors have not been willing to impose a universal versioning requirement on CIPs mainly because of different treatments of "versioning" according to each application and CIP subject.  We've only rejected the idea that a CIP `Version:` field should be added to the header (suggesting it should be described in the CIP text itself), and have left all definitions and applications of versioning to each CIPs author or maintainer.

Given this _laissez-faire_ approach we have somehow missed the particular importance of versioning for CIP-0068.  My own understanding (forgive me, I'm not a developer in this area) was:
- new asset name labels being added via CIP-0067 would define the set of labels being supported, so the recent addition of the `444` asset name label wouldn't cause any incompatibilities that a version "bump" would fix.
- there is a `version` defined in the `datum` part of the schema which appears to be set always to `1` which (I guess) shows that these datums all together establish a "version 1" of CIP-0068 as it exists now, even with the recent addition.

In the discussion for https://github.com/cardano-foundation/CIPs/pull/494 I didn't see any calls by either @AndrewWestberg, @alessandrokonrad, or implemetor @mmahut to add an explicit versioning scheme, although it's come up after the fact that this would have been a good idea.  (If that were explicitly suggested somewhere then please somebody link to it below.)

Whatever may have been missing in the CIP review & editing process is bracketed in the Discord discussion between [here](https://discord.com/channels/971785110770831360/992011119872970762/1108336998634246185) and [here](https://discord.com/channels/971785110770831360/992011119872970762/1113362054917660735).  There were some longer, prior discussions on Discord about versioning which I would really like to ensure at this point have not been interpreted as an indication that CIP editors are against versioning in general.

So this issue is to get input from all concerned parties to:
- (ASAP) decide the best way to update CIP-0068 to give @mmahut and other implementors a solid base for to support this standard;
- (maybe) decide on a versioning scheme for applicable CIPs, with semantic versioning being generally recommended (most recently by @KtorZ - cc @matiwinnetou);
- more fully consider, between authors and editors, whether this kind of support issue could be avoided in the future by enabling a `Version:` field in CIP headers;
- decide whether CIP versioning should be mandatory in any particular, well-defined cases... e.g. for CIPs which present anything like an "API" as e.g. CIP-0068 presents to Blockfrost;
- (finally) list whatever precise rules can be used to update CIP-0001 accordingly.

Since I've been the main responder to the latest general questions about CIP versioning, I will keep trying to bring in prior discussions on the matter and will keep referring authors & reviewers here to get as broad a consensus as possible.

Please also tag any concerned parties and feel free to re-post any prior discussions or perceived omissions here, to be sure we make a full inventory of everyone's opinions and requirements.

KtorZ:
- [x] Move CIP to `Active` 
- [x] Require version >= 2 for RFT metadata
- [x] Clarify the use of the version number 
- [x] Introduce succinct changelog for major version changes
- [x] Align CIP with new CIP-0001 format

Fixes #520.

cc @rphair @Ryun1  @alessandrokonrad @AndrewWestberg @mmahut @SmaugPool

### DIFF
--- a/CIP-0068/README.md
+++ b/CIP-0068/README.md
@@ -1,12 +1,23 @@
 ---
 CIP: 68
 Title: Datum Metadata Standard
-Authors: Alessandro Konrad <alessandro.konrad@live.de>, Thomas Vellekoop <thomas.vellekoop@iohk.io>
-Comments-URI:
-Status: Proposed
-Type: Informational
+Status: Active
+Category: Tokens
+Authors:
+  - Alessandro Konrad <alessandro.konrad@live.de>
+  - Thomas Vellekoop <thomas.vellekoop@iohk.io>
+Implementors:
+  - Alessandro Konrad (SpaceBudz)
+  - 5Binaries (Blockfrost)
+  - Smaug (Pool.pm)
+Discussions:
+  - https://github.com/cardano-foundation/CIPs/pull/299
+  - https://github.com/cardano-foundation/CIPs/pull/359
+  - https://github.com/cardano-foundation/CIPs/pull/458
+  - https://github.com/cardano-foundation/CIPs/pull/471
+  - https://github.com/cardano-foundation/CIPs/pull/494
+  - https://github.com/cardano-foundation/CIPs/issues/520
 Created: 2022-07-13
-Post-History:
 License: CC-BY-4.0
 ---
 
@@ -14,7 +25,7 @@ License: CC-BY-4.0
 
 This proposal defines a metadata standard for native assets making use of output datums not only for NFTs but any asset class.
 
-## Motivation
+## Motivation: why is this CIP necessary?
 
 This proposal addresses a few shortcomings of [CIP-0025](https://github.com/cardano-foundation/CIPs/blob/master/CIP-0025):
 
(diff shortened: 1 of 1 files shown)


rphair (approved):
thanks, looks complete & effective to me ⭐ https://github.com/cardano-foundation/CIPs/issues/520#issuecomment-1584318483

alessandrokonrad:
I don't quite understand why the version of an independent sub standard needs to be bumped just because a new asset name label is used? They all have nothing to do with each other. `version` here is meant to bumped independently for a specific label in case new properties will be added in the future or a new format is agreed on. 
E.g. let's take the label `222`. It represents NFTs with a focus on media files. Let's say there is a better metadata format for it, then this new format can be adopted by incrementing the version field. You don't introduce a new label in this case, because you want to be in exactly the same category/class.
So labels are for introducing completely new asset classes/categories and versions are for adding/changing the metadata format of that class over time. Maybe the version never changes, but it gives you extra flexibility. 
Why should `444` start with version 2?. Each sub standard can almost be seen as its own little CIP. They are all independent.

KtorZ:
@alessandrokonrad you have to put an implementor hat on, like 5Binaries or Smaug or any wallet provider actually that seek to support the CIP. If the CIP keeps changing and new formats / asset classes are added, we end up in a similar situation to CIP-0030 where everyone is implementing it but no one is actually implementing the same thing. Then, from a user perspective, it's hard to know what version of the CIP is actually supported. 

Thus, we need to have an overarching version number that is incremented when significant changes are made. Introduce a new asset class falls in this category. And indeed, modifying the format of a single one should also lead to a version increase. 

These can't be decoupled in practice because they're part of the same CIP. So they will always be considered as a bundle.

alessandrokonrad:
> Thus, we need to have an overarching version number that is incremented when significant changes are made. Introduce a new asset class falls in this category. And indeed, modifying the format of a single one should also lead to a version increase.

Yes I get this and it makes sense, however this is not the intention of the version field in the datum. CIP-0068 is the base structure and all those sub standards utilize it. They really can be seen as their own independent standards. We could seperate them and have them as their own CIPs, but I feel like this is overkill for some token standards based on CIP-0068. 
I don't see CIP-0068 as a place where all new token standards will be added. It was fine for the first 2-3, but if more sub standards are added we should separate them. That's why I think bumping the version field for certain sub standards doesn't make a lot of sense and imo is confusing.

Crypto2099:
Since this has been on the books a while and the contents of CIP-68 have changed substantially I think we should submit this one for triage, rebasing, and reconsideration during the next CIP Editors Meeting. Thoughts @rphair @KtorZ?

rphair:
yes that's great @Crypto2099 - I knew this material had several upcoming concurrent changes and really wanted to incorporate the above into the baseline back in June, especially since the versioning question already appeared settled by @Ryun1's additions to CIP-0001.  Some of the items in the OP I think are done already & we can confirm whatever remains... I've added it to next week's agenda.

I also think there could be a case for a CIP specification for Beacon Tokens so I would also like to get the CIP-0068 platform fully straightened out before I make that pitch to the developer 🧐

rphair:
@Crypto2099 I'm closing this `Deprecated` because memory (now distant) tells me you were addressing these issues in #586.  I cannot imagine the RFT version continuing for so long without a version bump & without problems otherwise.  If I'm mistaken about these assumptions please reopen.  cc @Ryun1 @KtorZ

PR #525, opened 2023-06-21
CIP-0060 | Fix relative links

rphair:
The Developer Portal build of the CIP directory structure depends on relative links being prefixed by `./` so this is a routine fix that comes up from time to time (we are gradually getting to all of these).

Not sure why these problems aren't all exposed at once, but local builds of the Dev Poral generate warnings like these for relative pathnames and often expand the absolute URL wrong (cc @katomm):
```
[WARNING] Docusaurus found broken links!

Please check the pages of your site in the list below, and make sure you don't reference any path that does not exist.
Note: it's possible to ignore broken links with the 'onBrokenLinks' Docusaurus configuration, and let the build pass.

Exhaustive list of all broken links found:

- On source page path = /docs/governance/cardano-improvement-proposals/CIP-0060:
   -> linking to cddl/version-2.cddl (resolved as: /docs/governance/cardano-improvement-proposals/cddl/version-2.cddl)
```


Ryun1 (approved):
Good spot 🚀

PR #527, opened 2023-06-29
CIP-0025 | Restrict to minting (positive amounts)

nielstron:
As discussed in https://github.com/cardano-foundation/CIPs/pull/392#issuecomment-1612564049 there is a flaw in the design of CIP 25 should new token minting policies make burning permissionless.

In particular, the CIP mentions the latest minting transaction of a token to be the most up-to-date version of the metadata of a token. Burning tokens is not explicitly excluded from being a minting transaction (where technically both minting and burning are minting operations for the ledger, just with positive/negative amounts). If burning would become permisionless, users can arbitrarily change the metadata of their tokens. Hence this change proposes to explicitly forbid burning transactions to be counted for CIP 25 updates.

### DIFF
--- a/CIP-0025/README.md
+++ b/CIP-0025/README.md
@@ -139,7 +139,7 @@ Optional fields allow to save space in the blockchain. Consequently the minimal 
 
 As mentioned above this metadata structure allows to have either one token or multiple tokens with also different policies in a single mint transaction. A third party tool can then fetch the token metadata seamlessly. It doesn't matter if the metadata includes just one token or multiple. The proceedure for the third party is always the same:
 
-1. Find the latest mint transaction with the label 721 in the metadata of the specific token
+1. Find the latest mint transaction with the label 721 in the metadata of the specific token that mints a positive amount of the token
 2. Lookup the 721 key
 3. Lookup the Policy Id of the token
 4. Lookup the Asset name of the token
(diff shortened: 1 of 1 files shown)


Crypto2099:
I believe this is the way most of our community tooling has already operated and this seems a very minor clerical point of order so yeah, I'm totally onboard as we shouldn't consider any update to a token during a full or partial burn (negative mint) operation.

rphair (approved):
Thanks @Crypto2099 for the confirmation - I was suspecting and hoping as such.

Crypto2099 (approved):
This is a relatively minor change and I believe the way that our community tooling already operated under "common sense" principles, but it is good to have it officially canonized.

Ryun1:
@rphair 
> I guess we should also poll the community to find out if this addition to CIP-0025 could possibly break anything. If it could, we should introduce a new version for the CIP as per our discussions & learning experience of https://github.com/cardano-foundation/CIPs/issues/520.

Id much more be in favour of a new version of CIP-25, as I am worried about breaking things in a similar manor to what happened with Issue #520. But then again I'm not familiar enough with such implementations to be able to estimate the impact of this change too accurately, so perhaps Im being too cautious 🤓.

rphair:
@mmahut you made the [report a few weeks ago](https://discord.com/channels/971785110770831360/992011119872970762/1108336998634246185) that adding the `444` implementation as per https://github.com/cardano-foundation/CIPs/pull/494 broke something on Blockfrost.  Is there any chance of something similar happening here if a new version isn't imposed on CIP-25 for this change?

I am also wondering @Ryun1 @KtorZ what token minting / management services we could query to ensure that metadata updates during burn isn't "a thing".

SmaugPool:
For what it's worth, I can confirm that I always considered that "mint transactions" were excluding burn transactions and that pool.pm always ignored the later ones for CIP25 metadata.

So I'm personally in favor of the additional clarification.

rphair:
thanks @SmaugPool - the second confirmation from your perspective is compelling.  If @KtorZ @SebastienGllmt also don't know of an implementation that would require a version bump for CIP25 to make this change, then we might merge this at upon review at the next meeting.

Ryun1:
@rphair 
> @Ryun1 one topic that will come up at the CIP meeting today is whether you are intent on including this in your plan in the works to add versioning treatment explicitly to CIP-0001.

So far I am not being as prescriptive for what to do with older CIPs, I am more focussed on giving direction for new CIPs.

Ryun1 (approved):
Although I am (maybe too)  wary of altering active CIPs, this does seem to be a minor clarification with decent community consensus, on those grounds I approve.

* as long as there are no community objections at the next CIP meeting (1st Aug 4pm UTC), we can place this in `last check` for the following meeting or merge.

rphair:
Promoting to `Last Check` on meeting & agenda since @KtorZ and @SebastienGllmt have not really had a chance to blow the whistle on this yet... we can merge at meeting if no reservations there 😎

PR #539, opened 2023-07-03
CIP-1694 | Adding French translation

Hornan7:
Each line has been meticulously translated into French and we are committed to keeping it updated according to each new commit of the original document.

---

([rendered version in branch](https://github.com/Hornan7/CIPs/blob/master/CIP-1694/README.fr.md))

PR #542, opened 2023-07-04
Extend CIP-0001 to cover internationalized content

KtorZ:
([new addition](https://github.com/cardano-foundation/CIPs/blob/i18n/CIP-0001/README.md#translations) = Translations section)

### DIFF
--- a/CIP-0001/README.md
+++ b/CIP-0001/README.md
@@ -155,6 +155,7 @@ While CIPs are mainly technical documents meant to be read primarily by develope
 
 Field          | Description
 ---            | ---
+`CIP`          | The CIP number (without leading 0)
 `Source`       | The canonical GitHub link to the original CIP document
 `Title`        | A translation of the CIP's title
 `Revision`     | The commit hash (8-10 first hex-digits) of the source (e.g. `12ab34cd`)
--- a/CIP-0001/README.md
+++ b/CIP-0001/README.md
@@ -155,7 +155,7 @@ While CIPs are mainly technical documents meant to be read primarily by develope
 
 Field          | Description
 ---            | ---
-`Source`       | A link to the original CIP document
+`Source`       | The canonical GitHub link to the original CIP document
 `Title`        | A translation of the CIP's title
 `Revision`     | The commit hash (8-10 first hex-digits) of the source (e.g. `12ab34cd`)
 `Translators`  | A list of translators names and email addresses (e.g. `John Doe <john.doe@email.domain>`)
--- a/CIP-0001/README.md
+++ b/CIP-0001/README.md
@@ -41,6 +41,7 @@ The CIP process does not _by itself_ offer any form of governance. For example, 
 - [Document](#document)
   - [Structure](#structure)
     - [Header Preamble](#header-preamble)
+    - [Translations](#translations)
     - [Repository Organization](#repository-organization)
     - [Licensing](#licensing)
   - [Statuses](#statuses)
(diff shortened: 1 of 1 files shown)


CIP-0001/README.md
```diff
@@ -140,6 +141,28 @@ CIP-0010
...
+
+- Translation file basenames must have a suffix using an [ISO 639-1](https://en.wikipedia.org/wiki/List_of_ISO_639-1_codes) language code, separated by a dot `.` character. (e.g. `README.fr.md`).
+
+- When no language code is provided as suffix, the default language for the document is assumed to be English (UK/US).
+
+- Translated CIPs (i.e. `README` files), must not include the original preamble. They must, however, include the following preamble as yaml frontmatter data:
+
+Field          | Description
+---            | ---
+`Source`       | A link to the original CIP document
```
rphair, line 158:
```suggestion
`Source`       | The canonical GitHub link to the original CIP document
```
I believe we need to say "canonical" because links to `README.md` files are often supplied by giving the containing folder (as often done for CIPs themselves).  Only the canonical `README.md` link will have a corresponding list of commit hashes for the `Revision` field.

Likewise (as I mentioned at the meeting), if we combine `Source` and `Revision` into a single link to the commit hash, then the more naive GitHub visitors might not be able to follow the UI to the original document.

rphair (commented):
nicely done; answers all of my questions last phrased in https://github.com/cardano-foundation/CIPs/pull/539#issuecomment-1620280703 except the one that came up in the last hour's CIP meeting; ready to approve otherwise.

CIP-0001/README.md
```diff
@@ -140,6 +141,28 @@ CIP-0010
...
+
+- Translation file basenames must have a suffix using an [ISO 639-1](https://en.wikipedia.org/wiki/List_of_ISO_639-1_codes) language code, separated by a dot `.` character. (e.g. `README.fr.md`).
+
+- When no language code is provided as suffix, the default language for the document is assumed to be English (UK/US).
+
+- Translated CIPs (i.e. `README` files), must not include the original preamble. They must, however, include the following preamble as yaml frontmatter data:
+
+Field          | Description
+---            | ---
+`Source`       | The canonical GitHub link to the original CIP document
```
rphair, line 158:
```suggestion
`CIP`          | The CIP number (without leading 0)
`Source`       | The canonical GitHub link to the original CIP document
```
To match (mostly) https://github.com/cardano-foundation/CIPs/blob/master/CIP-0001/README.md#header-preamble

rphair (commented):
As per https://github.com/cardano-foundation/CIPs/pull/539#issuecomment-1621866159 we're going to get people leaving the `CIP` field out unless it's added explicitly.

If not considered important to the human reader, and even if not used (as I would think it would be) to accumulate translations along with the originals by site building scripts, this number is ultimately like a primary key in a table that, failing everything else, will line up the CIP with its translation (even if URLs go out of date for instance).

Ryun1:
I do wonder if for the `Revision` field we include a URL with the commit hash? something like `18e692b <https://github.com/cardano-foundation/CIPs/pull/539/commits/18e692b5525bb7909c3637e793a449e0c01bd47b>`?  This would make navigation easier, no need to manually search for the commit hash.

rphair:
@Ryun1 how we considered this idea in the rapid-fire discussion at the meeting (and some online afterward) was:
1. The original proposal URL needs to be there in full so that _naive users_ can click into it without having to follow or deduce it from the dressing on the GitHub page for the translated document... since this takes some familiarity with GitHub to know how to do.
2. The commit hash does not require this convenience because anyone interested in a commit hash is a _sophisticated user_ who will know where to find the `History` link on the web page for the canonical GitHub link which they can easily click into.

Ryun1:
@rphair 

I got lost in some of the rapid fire meeting discussions! 🤓

What you described does sound very reasonable to me and I am happy to approve. 

For the sake of it, I guess the other downside of the current approach is that the only way the _naive users_ can tell if their translation is behind the current CIP is by searching and following the commit trail. But this should not really be much of an issue, since merged CIPs do not change too much, thus translations are not too likely to be out of date quickly or frequenctly.

PR #543, opened 2023-07-04
CIP-1694 update | Introduce governance action fee instead of deposit

KtorZ:


CIP-1694/README.md
```diff
@@ -509,20 +508,36 @@ for example, a motion of no confidence is enacted.
...
 > **Note**
 > A **single** governance action might contain **multiple** protocol parameter updates. Many parameters are inter-connected and might require moving in lockstep.
 
+#### Fee
+
+**Any Ada holder** can submit a governance action to the chain.
+Transactions that carry governance actions must satisfy a **minimum fee** to be valid.
+The minimum fee is parameterized by a new protocol parameter `minGovernanceFee`.
+
+When a governance action is ratified, or when it expires,
```
Ryun1, line 526: What happens to fee if a governance action dropped (due to a motion of no confidence for example)?

CIP-1694/README.md
```diff
@@ -509,20 +508,36 @@ for example, a motion of no confidence is enacted.
...
+any governance fee surplus is distributed amongst DReps **that took part in the vote**, regardless of their vote.
```
Ryun1, line 527:
```suggestion
any governance fee surplus is distributed amongst registered DReps **that took part in the vote**, regardless of their vote.
```
To avoid confusion with pre-defined DReps?

CIP-1694/README.md
```diff
@@ -509,20 +508,36 @@ for example, a motion of no confidence is enacted.
...
+The distribution is done proportionnally to the DRep stake distribution, rounded down to the nearest Lovelace.
+Any remainder Lovelace is sent to the treasury.
```
Ryun1, line 529: What happens to the fee if the governance action receives no votes? - I assume it all goes to the treasury?

CIP-1694/README.md
```diff
@@ -731,7 +746,7 @@ The **governance group** consists of all the new protocol parameters that are in
 * governance voting thresholds ($P_1$, $P_{2a}$, $P_{2b}$, $P_3$, $P_4$, $P_{5a}$, $P_{5b}$, $P_{5c}$, $P_6$, $P_7$, $Q_1$, $Q_{2b}$, $Q_4$)
 * constitutional committee term limits
 * governance action expiration
-* governance action deposit (`govDeposit`)
```
Ryun1, line 734: Are governance action deposits being removed? because they are still referenced many times in this proposal

CIP-1694/README.md
```diff
@@ -980,6 +994,26 @@ time, and instead depend on off-chain socialization to guide voters' attention t
 In this scenario, the constitutional committee might choose to only consider proposals which have
 already garnered enough votes from the DReps.
 
+### Governance action fee
+
+By introducing an optional fee on each governance action, we push the question of the DRep incentives down to the users of the system.
```
Ryun1, line 999: Why cannot a deposit and fee both exist at once? why is a fee approach favoured over the deposit?

CIP-1694/README.md
KtorZ: Nice catch, thanks. Yes, in this proposal, fees supersedes the deposit.

WhatisRT:
Here are some concerns I have about this:
- People are now disincentivized to make proposals.
- DReps can hold something hostage by voting `No`, since it collects the fees and anyone making a good change (whatever that means) has to put up more and more money.
- What about `Abstain` voters? Should they be rewarded? I think not, since then there's an incentive to just vote `Abstain` on everything, giving you a second staking income.
- The previous issue arguably also applies to voting `No` every time, only voting `Yes` on something that will give you a direct profit. But only rewarding `Yes` voters has a strong feeling of buying votes.

kevinhammond (commented):
This is an interesting idea (and looks relatively simple, which is always great!).  We should discuss it with you when we are in Edinburgh next week along with other suggestions for DRep incentives.

Note that if an action is dropped, the fee should probably be returned to the submitter, I imagine (so you would still need the rewards address)

thenic95:
Thanks for sharing this proposal @KtorZ.

If fees or any direct, quantitative incentive are introduced, it might prompt a majority of DReps to vote on as many governance actions as possible to maximize rewards. Consequently, the merit-based review of governance actions by DReps could become obsolete. The silver lining is that such behaviors would be visible on-chain, allowing delegators to assess their DReps' actions and determine whether they make informed decisions or merely farm rewards.

I still think it is an interesting approach, and I would strongly prefer a visible on-chain "bribery" system over a hidden black market for buying DRep votes.

It's probably not feasible for MVG, but could a system be introduced that slashes the deposits of DReps that always vote `Yes`, `No` or `Abstain` to farm rewards?

CIP-1694/README.md
Hornan7: Using ''registered'' DRep could possibly conflict with the changes proposed by #545 about DRep registration if merged. 
Suggestion:  Maybe ''active DReps'' instead ?

CIP-1694/README.md
Hornan7: I love this governance fee new protocol parameter idea. May I suggest to keep the `govDeposit` with the `minGovernanceFee` which would be small at first and both ajustable through governance actions (protocol params change). So you get the best of both worlds. 

- A small `minGovernanceFee` shared amongs the DReps, so you keep the incentives of submitting goverance actions.
and
- A bigger `govDeposit` to mitigate the governance action load.

Also a small fix suggestion on Rational links:
image

PR #545, opened 2023-07-05
CIP-1694 | A few smaller changes

WhatisRT:
These are changes that we wanted to make for a while but were too busy and didn't make it in time for the previous merge.

- All governance actions are now enacted one epoch after they have been ratified. This is necessary to comply with requirements coming from moving the stake distribution to disk.
- Remove dropping of governance actions in favour of delaying. The only effect this has is that if a `NoConfidence` GA is enacted and a new committee is elected later, any non-expired GAs can still be voted on and enacted (but the new committee has to approve them explicitly!).
- Make the DRep registration not required. This improves liquidity of democracy.

### DIFF
--- a/CIP-1694/README.md
+++ b/CIP-1694/README.md
@@ -266,8 +266,6 @@ The constitutional committee is considered to be in one of the following two sta
 
 In a _state of no-confidence_, the current committee is no longer able to participate in governance actions
 and must be replaced before any governance actions can be ratified (see below).
-Any outstanding governance actions are dropped immediately after the protocol enters a state of no-confidence,
-and will not be enacted.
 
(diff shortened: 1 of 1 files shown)


WhatisRT:
Let me just clarify a few things. In the currently published design:

- You can already delegate voting power to an unregistered DRep
- You can already vote as an unregistered DRep
- If you do those two things, the vote/stake is **actively** excluded from the results

All this change does is to not actively remove any votes. It simplifies the design (we need to do/check less things) and it means that DReps can't make the right to vote impossible to obtain by setting the deposit to an astronomical value.

tallwalton:
VOTE NO!!! on D repTax

PR #546, opened 2023-07-06
CIP-????: POO Protocol

Crypto2099:
The POO (Proof of Onboarding) Protocol defines a new standard for Cardano CIP-13 URI formats as well as a new standard for a communication protocol and API between wallets and project servers to provide a reliable and standardized method to "airdrop" tokens onto users.

The primary use case scenario for this protocol is for projects attending live events where the users in question may not even have a crypto wallet yet, let alone any crypto coins in their wallet. How do we still onboard (and capture reliable metrics) these users to the ecosystem?

In partnership with $HOSKY Token, VEGAS Stake Pool, and VESPR wallet we have developed this specification to support the following user journey:
1. Project generates a printed flyer or card that contains two (or more) QR codes.
2. The first QR code takes the user to the app store on their device to download and install the wallet software (in this case VESPR wallet).
3. The second QR code can be scanned from within the application (or via deep linking) and makes a call to the project's server w/ the wallet address of the user and a claim code that can be campaign-specific or one-time user.
4. The project server is responsible for validating address and code for eligibility to receive whatever tokens the code may qualify them for.
5. The project server issues one of several standardized responses to inform the wallet of the status of the request.
6. User receives the tokens via a transaction from the project server.


Crypto2099:
Link to rendered version: https://github.com/Crypto2099/CIPs/tree/pooprotocol/CIP-XXXX

rphair (commented):
1 - A note about the title:
1. It's a memorable acronym but people landing on this PR should more importantly know immediately what it's about, with as little distraction as possible.
2. The term "protocol" is optional considering that other CIPs for protocols don't have it in the title... and not necessary here because "onboarding" implies a process (although the extra word does produce another catchy acronym).

In any case, if this new title "sticks" (so to speak) we can update the document to reflect it at some point.

2 - More urgently we should first consider decoupling this from CIP-0013 as much as possible, based on prior agreement.  I would be open to retitling CIP-0013 as "URIs for payment and stake pool links" to restrict it to those two URI protocols alone.  The history:

The decision to produce separate CIPs for further extensions to the Cardano URI scheme has mainly taken place on on the forum: https://forum.cardano.org/t/cip-generalized-cardano-urls/57464

and a little bit on GitHub in response to an aborted request to include some URI protocols for an unpopular app-specific URI scheme: https://github.com/cardano-foundation/CIPs/pull/130 https://github.com/cardano-foundation/CIPs/pull/234

... but we've never formalised the decision to **document additional protocols as new CIPs** because there was nothing to merge yet.  We knew that day was coming and I think your practical requirement will force us to confirm this.

If the consensus is to do that, some time after 18 July (after I have another deadline) I can work on a CIP-0013 update or deprecation depending upon how you, reviewers & editors think the URI Scheme should be maintained going forward... given that this is the first useful extension to it that's been proposed since we merged my own "stake pool link" updates.  Then you'd be free to include the entire "onboarding" spec in a single new CIP document.

I do feel strongly that bulking up the ancient CIP-0013 with the somewhat heavyweight "onboarding" protocol is going to cause some confusion as well as discourage implementors from supporting the 2 URI protocols already documented there.

CIP-XXXX/README.md
```diff
@@ -0,0 +1,315 @@
...
+
+_Examples:_
+```json 
+URI: web+cardano://claim/v1?faucet_url=https%3A%2F%2Fclaim.hosky.io
+URL: https://claim.hosky.io
+POST JSON Data:
+{
+  "address": "addr1abc...xyz"
+}
+
```
rphair, line 136: I tried preparing a commit which goes something like this: but it changes the appearance too radically for me to push it into your branch.  Can you do something like this? ... preserving readability as you have imagined it, and being enough to dismiss the ugly red highlighting of the currently reported JSON syntax errors here.

---

URIs:
```
web+cardano://claim/v1?faucet_url=https%3A%2F%2Fclaim.hosky.io
https://claim.hosky.io
```
POST JSON Data:
```json 
{
  "address": "addr1abc...xyz"
}
```
etc...

rphair (commented):
on another note: somehow we have to remove the illegal comment entries from JSON: otherwise GitHub will highlight them as errors until Kingdom come.

CIP-XXXX/README.md
Crypto2099: Thanks Robert! I will take a look and make the modifications necessary per your feedback :)

Crypto2099:
> 1 - A note about the title:
> 
> 1. It's a memorable acronym but people landing on this PR should more importantly know immediately what it's about, with as little distraction as possible.
> 2. The term "protocol" is optional considering that other CIPs for protocols don't have it in the title... and not necessary here because "onboarding" implies a process (although the extra word does produce another catchy acronym).
> 
> In any case, if this new title "sticks" (so to speak) we can update the document to reflect it at some point.
> 
> 2 - More urgently we should first consider decoupling this from CIP-0013 as much as possible, based on prior agreement. I would be open to retitling CIP-0013 as "URIs for payment and stake pool links" to restrict it to those two URI protocols alone. The history:
> 
> The decision to produce separate CIPs for further extensions to the Cardano URI scheme has mainly taken place on the Forum: https://forum.cardano.org/t/cip-generalized-cardano-urls/57464
> 
> ... and a little bit on GitHub in response to an aborted request to include some URI protocols for an unpopular app-specific URI scheme: #130 #234
> 
> ... but we've never formalised the decision to **document additional URI protocols as new CIPs** because there was nothing to merge yet. We knew that day was coming and I think your practical requirement will force us to confirm this.
> 
> If the consensus is to do that, some time after 18 July (after I have another deadline) I can work on a CIP-0013 update or deprecation depending upon how you, reviewers & editors think the URI Scheme should be maintained going forward... given that this is the first useful extension to it that's been proposed since we merged my own "stake pool link" updates. Then you'd be free to include the entire "onboarding" spec in a single new CIP document.
> 
> I do feel strongly that bulking up the ancient CIP-0013 with the somewhat heavyweight "onboarding" protocol is going to cause some confusion as well as discourage implementors from supporting the 2 URI protocols already documented there.

Do you feel that the best solution in this case then is to bring all of the CIP-13 amendments into the scope of this single README file and so the URI itself is also part of this CIP? I think this makes a lot of sense and allows the versioning syntax of the URI to also be updated alongside this document.

rphair:
@Crypto2099 the way we were leaning from the Forum discussion + the last CIP meeting when this issue came up was to:
1. modify CIP13 so the "scheme" (ABNF) part of it just breaks out the "protocol" string, then declares that each protocol has its own separate definition for query parameters and argument processing.
2. shift the CIP13 content for the two already defined protocols ("payments" and "stake pool links") so they are each fully and separately pre-defined in CIP13 itself... to avoid deprecating CIP13 and then having to define 2 new CIPs for these already accepted protocols.
3. declare in CIP13 that any _new_ protocols besides these two can be defined in a new CIP which handles that newly declared protocol string (`//claim` in this/your case).

Therefore this PR would not modify CIP-13... although someone (me I think) would rewrite CIP-13 accordingly ASAP in a separate PR as we were thinking of doing 2 years ago.

The only difference between this plan and what we talked about back then, with CIP-68 happening in between with its included definitions for certain initially defined asset labels, has made me think that this organisation would also suit CIP-13... and as we've seen in https://github.com/cardano-foundation/CIPs/pull/504 I would like to ensure new protocol definitions, including this PR, don't require modifying CIP-13 each time. 😬

Crypto2099:
@rphair okay yeah, so I think I was a bit vague, that was also my understanding from the links that you mentioned which I did review. So, I will move the modifications that are currently in the CIP-13 README into the POO (CIP-????) README and they will just live there and be modified and iterated along w/ the Proof of Onboarding CIP in the future. This makes sense and actually makes it much easier to follow along with progress and changes IMO.

rphair:
@Crypto2099 sounds fine & I've scheduled to work on a revision of CIP13 accordingly soon after I'm done with that other deadline on the 18th.  There's a CIP editors' meeting scheduled for that day so I'll plan to talk there about the changes proposed to CIP13 itself when we introduce your new proposal from the agenda. 😎

thaddeusdiamond:
@Crypto2099 what do you think about including a few `GET` endpoints to check status? For instance, I could imagine a web architecture (e.g., React or React Native) that submits the initial `POST` to claim the tokens, then continually polls on an interval until the `GET` is met with a 200 and the tx hash of the completed claim.  You could just check the wallet using something like Blockfrost, but it would often be nice to know the exact tx hash linked to the claim, especially if the user is claiming multiple of the same policy ID just with different codes.

Crypto2099:
> @Crypto2099 what do you think about including a few `GET` endpoints to check status? For instance, I could imagine a web architecture (e.g., React or React Native) that submits the initial `POST` to claim the tokens, then continually polls on an interval until the `GET` is met with a 200 and the tx hash of the completed claim. You could just check the wallet using something like Blockfrost, but it would often be nice to know the exact tx hash linked to the claim, especially if the user is claiming multiple of the same policy ID just with different codes.

This should actually already be handled by the various 200 responses already defined within the protocol as it stands. So we move from 200 (initial receipt and acknowledgement) to 201 (pending delivery and queue order assuming same wallet + code) to 202 once completed along with the txn hash. Whether or not the wallet implements a polling or not would be up to their individual implementation.

CIP-XXXX/README.md
Crypto2099: This should be resolved in the latest commit.

Crypto2099:
I've started compiling some companion tools and helper functions which should help others to integrate Proof of Onboarding into their project and/or wallet here: https://github.com/Crypto2099/POO. Of maybe particular interest would be the CIP-13 Parser Lib found here: https://github.com/Crypto2099/POO/tree/main/packages/lib/cardano-poo-ts (@rphair @KtorZ)...

I think we need to work on the spec for staking and clean up the spec for payment addresses r.e. CIP-13 but I realize that's part of a larger conversation specific to CIP-13 itself :)

rphair:
> @Crypto2099 I think we need to work on the spec for staking and clean up the spec for payment addresses r.e. CIP-13 but I realize that's part of a larger conversation specific to CIP-13 itself :)

Thanks for giving me time to finish a Catalyst posting and rest a bit... I'll publish a PR draft for that by the end of the week.

CIP-XXXX/README.md
rphair: @Crypto2099 looks better all around 🤓

CIP-XXXX/README.md
```diff
@@ -0,0 +1,431 @@
...
+1. The Proof of Onboarding protocol was demonstrated via MVP during the Edinburgh, Scotland, UK CIP-1694 workshop hosted
+   by IOG in July 2023. This MVP demonstrated that the protocol works as described in this CIP and the overall feedback
+   from participants has been positive. This MVP was executed in collaboration with VESPR mobile Cardano wallet, HOSKY
+   token project, and Adam Dean as the distributor and architect of the specification.
+   1. [X] VESPR Mobile Wallet as of the time of this writing has support for Version 1 of the Proof of Onboarding Protocol
+      and is freely available for the use of any and all projects.
+   2. [ ] HOSKY Token project is working on finalizing and releasing a working, open source MVP for the software needed
+      to be run by the distributing project that will support fungible and non-fungible tokens or any combination thereof.
+   3. [ ] Adam Dean is in talks with additional wallet providers to bring P.O.O. support to more platforms as well as
+      working on open source tooling to facilitate ease of implementation for other projects.
```
rphair, line 427: The material in this drafted "third section" of _Path to Active_, although it makes sense as narrative, would be confusing if left here because there are items that pertain alternatively to Acceptance and/or Implementation.  This can be cleared up by:

1. Put each of the specific achievements & plans in this section into abbreviated bullet points in the Acceptance or Implementation sections.
2. Since the Edinburgh narrative indicates & confirms your design choice, I think the "story" would work well as an item in the Rationale.

CIP-XXXX/README.md
```diff
@@ -0,0 +1,431 @@
...
+
+# Copyright
```
rphair, line 429:
```suggestion
## Copyright
```

rphair (commented):
Readers who start at the OP above should know this proposal doesn't depend upon QR or any other media to convey the protocol URIs: this protocol doesn't necessarily integrate or depend upon any wallet choice (contrary to how it might sound from the OP).

High level description of my individual edits (most of them about "header level", with apologies for the mess; if I knew how many headers there were going to be, I would have done this in a single initial commit pushed into your branch 🤪):

Identifying a version number explicitly is still awaiting agreement on how to specify this,  as in what we were discussing in https://github.com/cardano-foundation/CIPs/issues/520

Document titles are H1 everywhere so the section headers need to be demoted to H2 and then subheadings further levels down from there.

- Note I was reading the document as I was noting the header level changes... so the changes would be in the same order in the PR thread and the diff... so don't worry, I was reading & happy with the writing in the sections (unless otherwise marked) as I was going through it.
- For clarity & to avoid confusing further review I'm going to commit (after a proofread to make sure I got all of them) the header indentation changes immediately after posting this (apologies to all the email subscribers to this repo).

Following the standards I dug more deeply into when rewriting CIP-0013 () it's apparently deprecated to make a distinction between a URN and a URI: but this shouldn't matter for your CIP because each extension to CIP-0013 only needs to define an `authority` and a `query` rather than quoting the whole URI scheme from the top...  but you can follow whichever you think is more clear for your readers.

The protocol mechanism itself (HTTP response mechanism) seems remarkably well thought out and I can't conceive at this time anything that would make it not work.  Really well done to your team on this concept & documentation. 🤩

rphair:
@Crypto2099 I trust you will be happy with the cognate CIP-0099.  I have scribbled on your PR enough that I will let you do the honours in changing your own number. 😅

CIP-XXXX/README.md
```diff
@@ -0,0 +1,431 @@
+---
+CIP: CIP-????
```
Ryun1, line 2:
```suggestion
CIP: CIP-0099
```

CIP-XXXX/README.md
```diff
@@ -0,0 +1,431 @@
...
+  "status": "ratelimited"
+}
+```
+
+##### Server Error (500)
+
+Implementations should of course be prepared to handle situations where a server is non-responsive for any reason and
+be prepared to handle any other, non-specified error codes including 500 codes.
+
+## Rationale: How does this CIP achieve its goals?
```
Ryun1, line 390: For me; it makes more sense to move the explanation of how the alternative approach of paper wallets is not ideal to the `Rationale`.

CIP-XXXX/README.md
```diff
@@ -0,0 +1,431 @@
...
+CIP: 99
+Title: Proof of Onboarding
+Status: Proposed
+Category: Wallets
+Authors:
+- Adam Dean <adam@crypto2099.io>
+- Carl <vegas@hosky.io>
+- Alex Dochioiu <alex@dochioiu.com>
+Implementors:
+- VESPR Wallet
```
Ryun1, line 11:
```suggestion
- VESPR Wallet <https://www.vespr.xyz/>
```

Ryun1 (commented):
Hey @Crypto2099,

A well defined proposal! see my small comments throughout.

CIP-XXXX/README.md
```diff
@@ -0,0 +1,431 @@
...
+
+**All arguments for Token Claim URIs should be URL-encoded**
+
+***Version 1 URIs***
+
+Version 1 URIs must include a `faucet_url` and a `code` as required parameters. URIs may include additional arguments to
+suit the needs of the project's faucet API.
+
+
+#### Handling Token Claim URI Queries
```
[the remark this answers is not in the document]
rphair: As part of the original OSI layer abstraction, networks must be irrelevant to higher level protocols like HTTP.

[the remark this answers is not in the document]
rphair: @Ryun1 since no dispute I guess this matter is settled then 😎

CIP-XXXX/README.md
```diff
@@ -0,0 +1,431 @@
...
+JSON POST Data:
+```
+```json
+{
+  "address": "addr1abc...xyz",
+  "code": "ABC123"
+}
+```
+
+##### Note on the `address` field
```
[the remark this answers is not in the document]
Crypto2099: This has been clarified to specify that a Shelley-era enterprise address or staking address should be used only. I know Byron addresses can technically receive tokens but think that's a can of worms that probably causes more trouble than it's worth.

CIP-XXXX/README.md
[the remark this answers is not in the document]
Crypto2099: We already end up with quite a bit of data in the URI once we compile it down into a QR code (very busy, hard to print in small form factor for clear legibility) and we explicitly allow for a "incorrect network" response from the monitoring server to handle these cases. I do not believe explicitly stating the network to use in the URI itself is beneficial or critical to successful operation.

CIP-XXXX/README.md
Crypto2099: Done :)

CIP-XXXX/README.md
```diff
@@ -0,0 +1,431 @@
...
+6. The project API returns a documented status code indicating the success or failure of the operation
+7. If a successful status is detected and returned, the project issues tokens to the specified address per their campaign settings
+
+### URI Format
+
+The URI format consists of the CIP-13 `web+cardano://` scheme, followed by the `claim` authority, then a `version` path.
+
+***NOTE: ALL ARGUMENTS SHOULD BE URL-ENCODED***
+
+#### Version 1
```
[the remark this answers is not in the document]
Crypto2099: Touché

CIP-XXXX/README.md
Crypto2099: This has been restructured to be less expositional :)

Crypto2099:
@Ryun1 @rphair I believe all changes (including moving to its new forever home in the CIP-0099 directory) requested or inquired about have been addressed and made in this latest version. Respectfully requesting formal review and potentially submission for Final Review during the next CIP Editors meeting if possible.

CIP-0099/README.md
```diff
@@ -0,0 +1,436 @@
...
+
+From there we will do our best to engage with other teams and projects in the ecosystem to encourage and support
+adoption of this standard on a larger scale.
+
+## Implementation Actions
+
+1. [X] VESPR Mobile Wallet supports the Proof of Onboarding Protocol and is freely available for the use of any and all projects.
+2. [X] HOSKY Project has released an open source server-side implementation software that may be used as a proof of concept for any interested projects.
+3. [X] Multiple projects at multiple, global events have successfully deployed Proof of Onboarding.
+4. [ ] Onboard additional wallet providers, server/service providers, and redemption methods.
```
rphair, line 432: For this to fit in with other CIPs, even those with a similar amount of coordination between separate entities, this section should be stripped of its subjective statements and formatted as a single list of achievements.  I'm not just saying this without consideration of the subject matter... and likewise the community would trust you & your partners not to proceed through the Implementation plan without the "insight" that you would have from the previous stages.

Whatever is in the background paragraphs preceding the Implementation checklist also reduces to checklist items when you remove the subjective stuff.  Once the list of actors & what they are doing is all in one list — ordered as chronologically as you are able, in series-parallel combination if necessary — it will be easier for the community to follow objectively.

rphair (commented):
OK, it seems like everything we have been looking for (all the particulars I was expecting & was double checking from @Ryun1) is there now, with some more corrections besides.

I understand that this CIP won't be "versioned" as such but will require a new CIP to make functional changes.  In general I would hope our CIPs would be more flexible than that, but also see in this case how it would be horrendous to keep multiple behavioural descriptions for all these interdependent pieces in a single document.

The "paper wallet" reference definitely works much better where you have it now in the Rationale.  In summary I think this is ready for a Review at the next meeting in the hope of promoting it soon after.

CIP-0099/README.md
Crypto2099: Is the suggestion to combine both "implementation plan" along with "implementation actions" into a single checkbox list?

rphair: Yes, but what I wrote above wasn't simply saying that.  I think the items that mightn't fit into a checkbox list are implicit and/or can be assumed from good faith in our community over what the implementors will be doing.

Crypto2099: Roger that, I will revise and get a new PR before Tuesday's meeting :)

rphair (approved):
Looks good as well for `Active` status; green lights across the `Path to Active` board.

brouwerQ:
@Crypto2099 Thoughts on my comment above?

PR #547, opened 2023-07-11
CIP-1694 correction | Fix Rationale TOC for undefined Catalyst item

rphair:
Found when reviewing translation in https://github.com/cardano-foundation/CIPs/pull/539.  Removes dead link in document since this subsection no longer exists.

Suggesting this item be simply be removed due to out-of-scoping as per://github.com/cardano-foundation/CIPs/blob/master/CIP-1694/README.md#off-chain-standards-for-governance-actions ... the alternative would be to reintroduce this subsection with a rationale affirming why it's out of scope.

cc @WhatisRT


Ryun1 (approved):
LGTM - good catch!

WhatisRT (approved):
Yeah, I'm all for eliminating distractions

PR #551, opened 2023-07-13
CIP-0102? | Royalty Datum Metadata

SamDelaney:
The inability to create trustless onchain royalty validation with CIP 27 is a major drawback to Cardano NFTs. The pattern defined in CIP 68 represents an opportunity to upgrade the standard to support onchain validation. CIP-102 aims to eliminate that drawback and demonstrate better support for developers, NFT creators, and NFT collectors, ultimately attracting dapps & NFT projects that would otherwise have taken their talents to another blockchain.

This specification is largely based on [the royalty specification in Nebula](https://github.com/spacebudz/nebula/tree/main#royalty-info-specification), with a couple key departures:

1. The royalty token is **recommended** to be locked at a script address, rather than stored in the user's wallet. This encourages projects to guarantee royalties won't change by sending their royalties to an always-fails (or similar) script address, but still allows for creative royalty schemes and minimizes disruption to existing projects.
2. The policyId of the royalty NFT must match that of the reference NFT. This enables lookups based on the user token in the same way as is done for the other tokens specified in CIP 68.

The specification here is made to be as minimal as possible. This is done with expediency in mind and the expectation that additional changes to the specification may be made in the future. The sooner we have a standard established, the sooner we can make use of it. Rather than attempting to anticipate all use cases, we specify with forward-compatibility in mind.

If you're interested in Ikigai's internal discussion that preceded this, check out https://github.com/ikigai-github/CIPs/pull/1.

---

([rendered proposal in branch](https://github.com/ikigai-github/CIPs/blob/cip68-royalties/CIP-0102/README.md))


GeorgeFlerovsky:
Interesting proposal! I'll have to read it more in-depth over the weekend, but your approach is promising.

19714 (approved):

rphair (commented):
I've added this to the agenda for the upcoming scheduled CIP meeting on 01 August (https://hackmd.io/@cip-editors/70).  The most important question to answer ASAP with a quick review & debate there (if not settled here already) is **whether this needs to be a separate CIP**.

_1 - New token types: do they belong in CIP68 at all?_

We have a precedent in CIP13 (pending https://github.com/cardano-foundation/CIPs/pull/559) so that it will only define a general framework for interpretation + the original 2 protocols it was conceived with... with further CIPs submitted for further protocol additions.  For CIP13 that would be newly defined URI authorities like `//claim`, and for CIP68 this would be the new token types `500` and thereafter.

As the author of that proposal and editor I would prefer _new CIPs for the new token types_, with the author of that type clearly defined in the new CIP header as well as the PR... not so that each author can "take credit" but more vitally to assign responsibilities for review, acceptance, and implementation.

Most vitally I foresee that CIP68 is going to be monstrous — both in size and editorial difficulty, given **we are already seeing concurrent work on different token types** (currently `444` and `500`) — if new types keep being added to CIP68 itself.

_2 - Versioning_

Some versioning needs to be done to ensure metadata processors aren't thrown off by the new `500` type (@mmahut claimed this was a problem with Blockfrost and `444`).  The last decision as per https://github.com/cardano-foundation/CIPs/pull/523 is that:
- the version number refers to the whole protocol, and
- the version field specified for each token type indicates in which versions this token type is supported.

So if that explanation & my understanding are correct, if keeping this addition in CIP68 itself you would need to:
- change the overall CIP68 version to 3;
- update the existing token types to add a version `3` as an acceptable value;
- define your `500` token as only being valid for version `3` (currently it just says `1`, as all new entries would in our CIP `main` branch without https://github.com/cardano-foundation/CIPs/pull/523 being merged yet)

I won't suggest code changes here because
- it would have to be rebased from a merge conflict with the still unmerged https://github.com/cardano-foundation/CIPs/pull/523
- this might end up being defined in a separate CIP anyway.

_3 - Sources_

I would assume your "Collaborator Discussion" https://github.com/ikigai-github/CIPs/pull/2 didn't include anyone outside Ikigai.  If so then there's no reason to change anything, but if not then it would be compelling to list those parties here (and maybe in the CIP itself, in the Acceptance section).

---

@alessandrokonrad could you please look at this & my comments above... not just as the CIP68 author, or any expectations of you as a "maintainer", but
- partly because of the work being based on Nebula
- mainly the questions about how the new asset types should be versioned, and more generally whether they should be added to CIP68 in the first place (as per your `444` https://github.com/cardano-foundation/CIPs/pull/523#issuecomment-1585024874).

SamDelaney:
Thanks for the reply! Responding in order here.

1. I'm quite happy to turn this into a new CIP if you all think that would be better. I leave that to your judgement.

1. The changes to versions makes sense. Thank you for clarifying that for me. I had conceptualized it differently but I agree your version is better. I'll follow your lead and wait before making code changes.

1. The collaborator discussion was intended to give developers already working on implementations a chance to give input before moving to the more solidified process here. I reached out to the developers I knew of who were doing so, but didn't get any feedback beyond approving DMs so I decided to go ahead and move it here. Linking that this PR to that one was a misstep, and I undid the tag, but I guess the link persists. Sorry 'bout the distraction.

rphair:
@SamDelaney you would be welcome at the next CIP editors' meeting above (Discord invite https://discord.gg/kyaTyzkBqd).  It's not slated for a detailed review, but the question about whether this should be a separate CIP would likely be covered, and I believe you should be involved in that discussion. 🙂

Crypto2099:
I really like the elegance of this solution for royalties when working w/ CIP-68 tokens.

I believe I would tend to agree that for long-term management and ease these individual/new sub-tokens should be broken out into their own CIPs where they can be individually iterated and ideated w/o potential merge issues arising from concurrent attempts to modify the core document.

I like that the data structure of the royalty token in this case specifies some optional min fee, max fee, and is presumed to be an array of recipient addresses (although I think we'll be a long time before we see widespread support for multi-royalty recipients due to minUTxO).

CIP-0068/README.md
```diff
@@ -329,6 +329,92 @@ We want to bring the metadata of the RFT `d5e6bf0500378d4f0da4e8dde6becec7621cd8
...
+big_nint = #6.3(bounded_bytes)
+
+optional_big_int = #6.121([big_int]) / #6.122([])
+
+royalty_recipient = #6.121([
+              address,                    ; definition can be derived from: 
+                                          ; https://github.com/input-output-hk/plutus/blob/master/plutus-ledger-api/src/PlutusLedgerApi/V1/Address.hs#L31
+              int,                        ; variable fee ( calculation: ⌊1 / (fee / 10)⌋ ); integer division with precision 10
+              optional_big_int,           ; min fee (absolute value in lovelace)
+              optional_big_int,           ; max fee (absolute value in lovelace)
```
Crypto2099, line 370: Are the min and max fee optional or are they required? If so, should there be a note that if you do not wish to set them then these values should be set to 0?

CIP-0068/README.md
```diff
@@ -329,6 +329,92 @@ We want to bring the metadata of the RFT `d5e6bf0500378d4f0da4e8dde6becec7621cd8
...
+
+; To store this in the royalty datum
+1 / (0.016 / 10) => 625
+
+; To read it back
+10 / 625 => 0.016
+```
+Because the computational complexity of Plutus primitives scales with size, this approach significantly minimizes resource consumption.
+
+To prevent abuse, it is **recommended** that the `royalty NFT` is stored at the script address of a validator that ensures the specified fees are not arbitrarily changed, such as an always-fails validator.
```
Crypto2099, line 398: I think we should change the conception that we had during the writing of CIP-27 that royalties should never change. What if I want to put royalties on a collection but only for the first year, or to remove royalties if I no longer intend to continue supporting and working on a project? Having mutable royalties does introduce additional complexity in needing to maintain the token and/or lock it in a script that only you (or a DAO multisig, etc) can control... But probably best for the long term if we do not encourage people to lock up minUTxO worth of Lovelace in a token just for the sake of "not changing the royalties".

CIP-0068/README.md
```diff
@@ -329,6 +329,92 @@ We want to bring the metadata of the RFT `d5e6bf0500378d4f0da4e8dde6becec7621cd8
...
+
+##### Retrieve metadata from a Plutus validator
+
+We want to bring the royalty metadata of the NFT `d5e6bf0500378d4f0da4e8dde6becec7621cd8cbf5cbb9b87013d4cc.(222)TestToken` in the Plutus validator context. To do this we
+
+1. Construct `royalty NFT` from `user token`: `d5e6bf0500378d4f0da4e8dde6becec7621cd8cbf5cbb9b87013d4cc.(500)Royalty` (off-chain)
+2. Look up `royalty NFT` and find the output it's locked in. (off-chain)
+3. Reference the output in the transaction. (off-chain)
+4. Verify validity of datum of the referenced output by checking if policy ID of `royalty NFT` and `user token` and their asset names without the `asset_name_label` prefix match. (on-chain)
+
```
Crypto2099, line 417: Having you tested the realistic feasibility of adding this logic to a standard "marketplace" smart contract? Especially with multiple royalty recipients? While this is certainly a novel concept and a step in the right direction, I wonder on the impact to CPU, memory, and subsequent transaction fees by using this reference token datum versus other, largely off-chain solutions atm.

Crypto2099 (commented):
Ultimately I have to question if this solution provides any additional "security" to a marketplace contract versus the currently off-chain CIP-27 implementations that we've seen. What happens if I submit my buy order w/o the reference UTxO for the royalty token? Can I still subvert the royalties? Is this fundamentally different or better than, say, putting the royalty information in alongisde a token in the listing datum instead?

Crypto2099:
Just brainstorming aloud at this point, but wondering if it would be advisable to provide an example of the logic that would need to be added to a Plutus minting script in order to allow only the creator/deployer of the contract to issue this token once? Bit of a philosophical question but it's all well and good to propose a new standard, but having some example code (most likely hosted in a separate repo of your own) showing how this can be used would help ease the adoption process for new standards.

Also a bit of selfish self-interest as this sort of "informational" token can/could be useful for my own CIP-88 validation for Plutus minting scripts. :)

CIP-0068/README.md
SamDelaney: They are optional. `optional_big_int` is defined in CDDL above this, but perhaps it should be expressed in a less niche format as well.

They are both expected to be a bigint or not be included. Max fee being set to 0 to represent no max fee would be confusing, as it could easily be misread to mean no fees at all, i.e. if the implementer of the protocol missed that wrinkle.

CIP-0068/README.md
SamDelaney: I agree with this notion to some degree. There are lots of perfectly valid use cases for changing royalties. The key word here is **recommended**. With this distinction we allow for such use cases, as well as accommodating existing protocols such as Nebula which store the royalty in wallets rather than script addresses.

However, I believe the recommendation against _arbitrary_ royalty control is important. A collection whose fees are controlled arbitrarily can be easily sabotaged at the whim of the creator by increasing the royalties beyond a reasonable amount. This is bad assurance for buyers, and we should set the precedent for better schemes, the simplest of which is an always-fails validator.

CIP-0068/README.md
SamDelaney: We have implemented this in Grabbit, but haven't benchmarked it very thoroughly yet. I can get back to you on CPU, memory, etc. on the client side. 

As far as TX fees go, this shouldn't have much of an impact. Reading the fee is as simple as parsing the datum, which doesn't require consuming it at all.

One valid concern is the memory cost to the blockchain itself. This is definitely something worth thinking about, and the design has carefully considered memory usage to be as minimal as possible. I personally think the ability for validators to read the royalties directly & trustlessly is worth the cost many times over.

SamDelaney:
> @SamDelaney you would be welcome at the next CIP editors' meeting above (Discord invite https://discord.gg/kyaTyzkBqd)

Thanks! I will definitely be there. 🙂 

> Ultimately I have to question if this solution provides any additional "security" to a marketplace contract versus the currently off-chain CIP-27 implementations that we've seen. What happens if I submit my buy order w/o the reference UTxO for the royalty token? Can I still subvert the royalties? Is this fundamentally different or better than, say, putting the royalty information in alongisde a token in the listing datum instead?

This is a good point. Let me think on this for a bit and get back to you.

> Just brainstorming aloud at this point, but wondering if it would be advisable to provide an example of the logic that would need to be added to a Plutus minting script in order to allow only the creator/deployer of the contract to issue this token once? Bit of a philosophical question but it's all well and good to propose a new standard, but having some example code (most likely hosted in a separate repo of your own) showing how this can be used would help ease the adoption process for new standards.

This is absolutely something I plan on doing. I even have [a Catalyst proposal](https://cardano.ideascale.com/c/idea/106968) out to help fund it's development. 🙂

CIP-0068/README.md
Crypto2099: Derp yeah, I guess I didn't read the names of the variables very well, but I would still add the `?` to indicate in the CDDL that they are optional :)

SamDelaney:
> Ultimately I have to question if this solution provides any additional "security" to a marketplace contract versus the currently off-chain CIP-27 implementations that we've seen. What happens if I submit my buy order w/o the reference UTxO for the royalty token? Can I still subvert the royalties? Is this fundamentally different or better than, say, putting the royalty information in alongisde a token in the listing datum instead?

Thanks again @Crypto2099 for bringing this problem to my attention. Somehow I had missed it until now. I got a chance to think about it over the weekend and I think I've come up with a good solution for it:

I propose a new addition is made to the standard that specifies an optional boolean `hasRoyalty` field in the _Reference NFT_'s datum.
- If the field is present and set to "True" the validators can require a royalty input.
- If the field is present and set to "False" the validators don't need to search for a royalty input.
- If the field is not present, validators should accept a royalty input, but not require one.

In addition to providing a way to create guaranteed royalties, this has several advantages:
1. **Backwards Compatibility** - existing royalty implementations will still work, just not have the same assurances.
2. **Minimal Storage Requirement** - An optional boolean is about as little space as possible. This is especially important because it's attached to the Reference NFT and will be set for each individual NFT.
3. **Intra-Collection Utility** - This would already allow for minting a collection with some NFTs with royalties and some without. It's also easy to imagine a future version which changes this field to a number and is used to allow for multiple versions of royalties for even more granular control.

It isn't functionally important, but this field could make sense at a couple of different locations in the Reference Datum. I don't have strong feelings about these, but if any of you do please let me know.

- As a part of the `extra` field, leaving the Reference NFT specification unchanged
- As a top level field, changing the Reference NFT datum spec' to:
```
...

? hasRoyalty : bool

datum = #6.121([metadata, version, hasRoyalty, extra])
``` 

I'm leaning toward the latter of these two, which I think makes for a clearer developer experience & will be easier to spot if accidentally excluded, but I could certainly be convinced of the alternative.

rphair:
Resolution from CIP editors' meeting in last hour: @SamDelaney has agreed to rework this PR as an independent CIP rather than a modification to CIP68.  We'll leave it in this PR to avoid fragmenting the discussion so far.

That still leaves the need for CIP68 to be re-written a bit: to point to other possible CIPs that define new token types.  We resolved that the most recent inclusion of `444` will be the last addition of a new token type to CIP68, in a pattern more-or-less following the baseline set in CIP-0013 (https://github.com/cardano-foundation/CIPs/pull/559).

- @Crypto2099 has graciously allowed himself to be volunteered to submit a separate PR to open up CIP68 in that way: although not solely responsible & all editors will keep an eye on it to share the responsibility.  cc @alessandrokonrad

We acknowledged that CIP67 was supposed to make CIP68 updates unnecessary for new token types, but in practice the documentation & metadata structures tend to require their own text documentation.

19714 (approved):

19714 (approved):

rphair:
@SamDelaney it looks like your material has been well forked off into a separate CIP.  Is this ready for review?  If so, can you move the CIP `README.md` out of the subdirectory `ref_impl` where it is now?

@Crypto2099 @Ryun1 as I recall, CIP-0068 was going to be modified a bit to specify what new CIPs based on its general metadata structure should look like.  Is anything going to be done with that?  Or will it be implicit that those new CIPs follow the initial ones specified in CIP-0068 (i.e. assuming the reader will deduce that the protocol is extensible by adding new CIP-0067 entries)?

BTW we will have the pending requirement that CIP-0068 be updated somehow so we can explain exactly when CIP-0067 _is_ supposed to be updated, if there are any cases beyond new token types (e.g. https://github.com/cardano-foundation/CIPs/pull/504).

rphair:
It came up in the CIP meeting today that, without being a part of CIP-0068 any longer, this document needs its own _Path to Active_ section, as per https://github.com/cardano-foundation/CIPs/blob/master/CIP-0001/README.md#path-to-active .

@SamDelaney you could start with some of items in CIP-0068's _Path to Active_ (https://github.com/cardano-foundation/CIPs/blob/master/CIP-0068/README.md#path-to-active) and augment/modify it with your own polling of the NFT services that you anticipate will support the new asset class; just please divide it into the 2 sections _Acceptance Criteria_ and _Implementation Plan_ as in our modern CIP format.

(Note we still need a directory name change for the proposal material... the `README.md` in `ref_impl` has a different purpose.)

19714 (approved):

SamDelaney:
@rphair I believe this is ready for review now. README has been moved out of `ref_impl` and Path to Active section has been added.

19714 (approved):

CIP-XXXX/README.md
```diff
@@ -0,0 +1,179 @@
+---
+CIP: ????
```
```suggestion
CIP: ?
```

CIP-XXXX/README.md
```diff
@@ -0,0 +1,179 @@
...
+
+- If the field is present and > 1 the validators must require a royalty input.
+- If the field is present and set to 0 the validators don't need to search for a royalty input.
+- If the field is not present, validators should accept a royalty input, but not require one.
+
+### Examples
+
+#### Retrieve metadata as 3rd party
+
+A third party has the following NFT `d5e6bf0500378d4f0da4e8dde6becec7621cd8cbf5cbb9b87013d4cc.(222)TestToken` and they want to lookup the royalties. The steps are
```
rphair, line 114:
```suggestion
A third party has the following NFT `d5e6bf0500378d4f0da4e8dde6becec7621cd8cbf5cbb9b87013d4cc.(222)TestToken` and they want to look up the royalties. The steps are
```
(this is spelled correctly other places in the document... `lookup` being a noun and `look up` being a verb phrase)

CIP-XXXX/README.md
```diff
@@ -0,0 +1,179 @@
...
+
+1. Construct `royalty NFT` from `user token`: `d5e6bf0500378d4f0da4e8dde6becec7621cd8cbf5cbb9b87013d4cc.(500)Royalty`
+2. Look up `royalty NFT` and find the output it's locked in.
+3. Get the datum from the output and lookup metadata by going into the first field of constructor 0.
```
rphair, line 118:
```suggestion
3. Get the datum from the output and look up metadata by going into the first field of constructor 0.
```

CIP-XXXX/README.md
```diff
@@ -0,0 +1,179 @@
+---
+CIP: ????
+Title: Royalty Datum Metadata Standard
```
```suggestion
Title: Royalty Datum Metadata
```
I wanted to include @KtorZ @Ryun1 @SebastienGllmt in a discussion of whether we should use the term Standard here.  I would anticipate problems if so, because another metadata solution might also address royalties in the future and then we would have uncertainty over which one is "standard" while leaving that word out would force the unambiguous use of a merged CIP number.

(CIP27 wasn't so much a problem in this way because they used the commercial designation "CNFT" in addition to "Standard" in the merged CIP title.)

@SamDelaney since the term "standard" is used so many places in the document I would recommend just changing it in the header metadata & the PR title and leaving everything else as-is: to keep the tone & clarity of what you have already written, therefore leaving more detailed readers to make up their own mind over whether to consider this a "standard".

CIP-XXXX/README.md
```diff
@@ -0,0 +1,179 @@
...
+
+### Implementation Plan
+
+1. Publish open source reference implementation and instructions related to the creation, storage and reading of royalty utxos.
+2. Implement in open source libraries and tooling such as [Lucid](https://github.com/spacebudz/lucid), [Blockfrost](https://github.com/blockfrost/blockfrost-backend-ryo), etc.
+3. Achieve additional "buy in" from existing community actors and implementors such as: blockchain explorers, token marketplaces, minting platforms, wallets.
+
+## References
+
+- CIP-0027: https://github.com/cardano-foundation/CIPs/blob/master/CIP-0027
```
rphair, line 172:
```suggestion
- CIP-0027: https://github.com/cardano-foundation/CIPs/tree/master/CIP-0027
```
The pathname component `/tree` is canonical (as already used for the others).

CIP-XXXX/README.md
```diff
@@ -0,0 +1,179 @@
...
+- [ ] Minimal reference implementation making use of [Lucid](https://github.com/spacebudz/lucid) (off-chain), [PlutusTx](https://github.com/input-output-hk/plutus) (on-chain): [Reference Implementation](./ref_impl/)
+- [ ] Implementation and use demonstrated by the community: NFT Projects, Blockchain Explorers, Wallets, Marketplaces.
+
+### Implementation Plan
+
+1. Publish open source reference implementation and instructions related to the creation, storage and reading of royalty utxos.
+2. Implement in open source libraries and tooling such as [Lucid](https://github.com/spacebudz/lucid), [Blockfrost](https://github.com/blockfrost/blockfrost-backend-ryo), etc.
+3. Achieve additional "buy in" from existing community actors and implementors such as: blockchain explorers, token marketplaces, minting platforms, wallets.
+
+## References
```
rphair, line 170: The CIP-0001 format, and therefore any parsers that may work on CIP material, has no _References_ section.  I would recommend putting these links inline: with the first use of a CIP term linked to its document & further references unlinked (like Wikipedia).

There is a format where anchors can be collected at the bottom of the page but I can't find a reference to it or a name for it... but some CIPs use it (e.g. CIP-0094) and you could model your linking on them.  Defining the links in that way would collect them at the bottom of the page which would help serve the purpose of your list here.

rphair (commented):
Thanks @SamDelaney - now that the `README` pathname is fixed I've added the link to your rendered document in the OP which will make editing & review easier.

Everything here seems to me at least as well documented as any other asset class in CIP-0068.  I don't work with this use case but it fits in with my expection of how a robust system would work.  I trust @Crypto2099 and implementors of explorers, marketplaces, etc. can keep going over it for the technical details. (cc @mmahut @SmaugPool)

CIP-XXXX/README.md
rphair: (as specified in CIP-0001)

SamDelaney: Done. FYI, this syntax appears to cause an error message in the rendered version:
image

It shows up in other PRs as well: https://github.com/serejke/CIPs/blob/CIP-keccak256/CIP-%3F%3F%3F%3F/README.md

CIP-XXXX/README.md
SamDelaney: Good catch thanks! Fixed.

CIP-XXXX/README.md
SamDelaney: References section has been removed per your comment below. The new anchored link is updated to use `/tree`.

CIP-XXXX/README.md
SamDelaney: I wasn't aware of that syntax, very cool. Updated as recommended.

CIP-XXXX/README.md
```diff
@@ -0,0 +1,176 @@
+---
+CIP: ?
+Title: Royalty Datum Metadata
+Authors: Sam Delaney <sam@ikigaitech.org>
```
rphair, line 4:
```suggestion
Authors:
  - Sam Delaney <sam@ikigaitech.org>
```
`Authors` is a list in the new format.

CIP-XXXX/README.md
```diff
@@ -0,0 +1,176 @@
...
+Comments-URI:
```
rphair, line 5:
```suggestion
Discussions:
```
This has changed name (and is also a list, if anything added in the future e.g. Cardano Forum threads).

CIP-XXXX/README.md
```diff
@@ -0,0 +1,176 @@
...
+Status: Proposed
+Type: Informational
```
rphair, line 7:
```suggestion
Category: Tokens
```
There is no longer a `Type` field; with a new `Category` field set as follows: https://github.com/cardano-foundation/CIPs/blob/master/CIP-0001/README.md#categories

CIP-XXXX/README.md
```diff
@@ -0,0 +1,176 @@
...
+Created: 2023-08-08
+Post-History:
```
rphair, line 9:
```suggestion
```
This field no longer exists.

rphair (commented):
re: what causes the YAML errors reported in some CIPs (https://github.com/cardano-foundation/CIPs/pull/551#discussion_r1308074802), I don't have a good feel for that (@KtorZ said something about it not long ago but I haven't detected a pattern).

Here are some other things that need to be fixed in the "new" style CIP-0001 headers as of a month ago (it is generally helpful for these little things that editors can write to CIP PR branches so we can also commit these changes).

(Note CIP-0068 itself still has to be edited according to the new CIP-0001 format.)

SamDelaney:
@rphair I have updated the preamble to match the CIP-0001 standard. You should also have received an invitation with write permissions to the ikigai-github/CIPs repo. I'm happy to add more editors as needed.

rphair:
@SamDelaney thank you, I have accepted that invitation & will try to handle further routine editing on behalf of the editors through that channel 🤓

Crypto2099:
@SamDelaney I'm going to be giving this a more thorough deep dive later today, but please see my recent pull request for CIP-67/CIP-68 modification standards as a reference for the intended (although still debatable) requirements. I think everything looks mostly good here on this front aside from providing a "documentation" link in CIP-67 which will point implementers to this pull request although this may need to wait until we have assigned a number to this one.

https://github.com/cardano-foundation/CIPs/pull/586

rphair:
@SamDelaney following on @Crypto2099's https://github.com/cardano-foundation/CIPs/pull/551#issuecomment-1699391414 I think your own proposal here is already compliant with the proposed changes... perhaps since the same discussion that produced #586 also produced your own PR.

Even if so, since your submission is a good test case for the CIP-0068 update please also review #586 yourself and let the editors know whether you think it's acceptable & clear from a token creator's point of view.

rphair:
@SamDelaney congratulations; we've assigned 102 as a CIP number, so please rename the directory containing your proposal to `CIP-0102` and update the link in your OP 🤓

SamDelaney:
Thanks @rphair! I've updated the CIP to match its new number. 🎉 

Sorry about the delay on reviewing your PR @Crypto2099, I've been pretty busy this weekend. I'll take a look at it in the next couple days.

rphair:
@SamDelaney as the producer of the first CIP-0068 extension you & @Crypto2099 should agree upon the proposed language about versioning the extensions here (https://github.com/cardano-foundation/CIPs/pull/586#discussion_r1312203819).  Then you can update this CIP accordingly... at the meeting today @Crypto2099 suggested "boilerplate" language repeating the originally stated conditions for which the `version` number needs to be updated.

CIP-0102/README.md
```diff
@@ -0,0 +1,177 @@
...
+
+```cddl
+big_int = int / big_uint / big_nint
+big_uint = #6.2(bounded_bytes)
+big_nint = #6.3(bounded_bytes)
+
+optional_big_int = #6.121([big_int]) / #6.122([])
+
+royalty_recipient = #6.121([
+              address,                    ; definition can be derived from:
```
KtorZ, line 60: Is an address truly needed here or could a payment credential (i.e. a vk or script hash) suffices? The address also contains a network discriminant and, a possible delegation payload. If there's a desire to handle delegation payload then, how will pointer addresses be handled?

CIP-0102/README.md
```diff
@@ -0,0 +1,177 @@
...
+
+; version is of type int, we start with version 1
+version = 1  
+
+; Custom user defined plutus data.
+; Setting data is optional, but the field is required
+; and needs to be at least Unit/Void: #6.121([])
+extra = plutus_data
+
+royalty_info = #6.121([royalty_recipients, version, extra])
```
KtorZ, line 77: Suggestion, bundle the version separately, so that the format can be adjusted if needed without prior knowledge of the version. For example:


```
royalty_info = #6.121([1, royalty_info_v1])
royalty_info_v1 = #6.121([royalty_recipients, extra])
```

CIP-0102/README.md
```diff
@@ -0,0 +1,177 @@
...
+### Reference Datum Royalty Flag
+
+If not specified elsewhere in the token's datums, a malicious user could send transactions to a protocol which do not reference the royalty datum. For full assurances, a new optional flag should be added to the reference datum
+
+```
+extra = 
+	{
+		...
+
+		? royalty_included : big_int
```
KtorZ, line 103: What about using `#6.121[] / #6.122[]` (False / True) instead ?

CIP-0102/README.md
```diff
@@ -0,0 +1,177 @@
...
+
+
+## Path to Active
+
+### Acceptance Criteria
+
+- [ ] This CIP should receive feedback, criticism, and refinement from: CIP Editors and the community of people involved with NFT projects to review any weaknesses or areas of improvement.
+- [ ] Guidelines and examples of publication of data as well as discovery and validation should be included as part of of criteria for acceptance.
+- [ ] Minimal reference implementation making use of [Lucid](https://github.com/spacebudz/lucid) (off-chain), [PlutusTx](https://github.com/input-output-hk/plutus) (on-chain): [Reference Implementation](./ref_impl/)
+- [ ] Implementation and use demonstrated by the community: NFT Projects, Blockchain Explorers, Wallets, Marketplaces.
```
KtorZ, line 163: 👍

KtorZ (approved):
I am good with the overall structure proposed in this CIP, though added a few technical comments / suggestions.

rphair (commented):
@SamDelaney I'm standing by to also approve this after you have a chance to respond to https://github.com/cardano-foundation/CIPs/pull/551#pullrequestreview-1616794900 and make any changes you see fit. 🤓

p.s. and also the versioning stub question, whether written or implicit, raised in https://github.com/cardano-foundation/CIPs/pull/551#issuecomment-1707400077 ... p.p.s. @Crypto2099 with a good suggestion already made here: https://github.com/cardano-foundation/CIPs/pull/586#discussion_r1317749255

CIP-0102/README.md
SamDelaney: Yes there's a delegation consideration, as you figured. I'm not sure I understand the concern with pointer addresses, can you elaborate?

Another consideration is consistency with the existing implementations. I want to break things as little as possible here. At the moment no changes are required to Nebula, at least. I'm less sure of closed source implementations, for obvious reasons.

CIP-0102/README.md
SamDelaney: This was my initial idea, and it would certainly work for v1. I have an idea for v2 of this schema to make use of this field to allow for multiple royalty datums to be attached to a policy, with the correct datum to be used for each individual NFT specified in this field.

With that plan in mind, I decided to make the type `big_int` instead, so the type expected by third parties doesn't have to change in future versions. Since computational complexity of Plutus primitives scale with size, performance should be effectively identical.

CIP-0102/README.md
rphair: Are these the same pointer addresses that were determined to be obsolescent here? https://github.com/cardano-foundation/CIPs/pull/374#issuecomment-1646581391

CIP-0102/README.md
SamDelaney: This is a very interesting idea. If I understand you correctly, this would be helpful for parsing the datum as a third party once multiple versions are introduced. One could parse the version number first, then know what type to expect for the royalty info itself. Is that what you're saying?

I like the idea, but I'm not sure how much of a difference it would actually make. Do you have a good idea of how much more efficient a change like this would be?

My only real concern is compatibility with existing implementations. I'm not sure this provides enough value to warrant making Nebula & other nebula-based implementations incompatible.

SamDelaney:
Hey @rphair & @KtorZ this PR has gone quiet for almost a month now. Trying not to let it stagnate.

I don't currently see a need to make any more changes, but am definitely up for further discussion. Not sure if your (@KtorZ) questions are satisfied. If so I think this may be ready for merge.

rphair (approved):
@SamDelaney it sounds like you've been attentive to https://github.com/cardano-foundation/CIPs/pull/551#discussion_r1319427307 and https://github.com/cardano-foundation/CIPs/pull/551#discussion_r1319429607; and so if they are issues for further documentation or improvement they can be done after merging.

@Ryun1 @Crypto2099 do I remember correctly that we'd suggested merging this at either today's meeting or the one 2 weeks ago?  I also think this is ready & so marking `Last Check` for the next CIP meeting (17 October) but perhaps it could be merged before then if you and/or @KtorZ @SebastienGllmt are willing.

CIP-0102/README.md
```diff
@@ -0,0 +1,177 @@
...
+#### Reference Datum Flag
+
+In addition to providing a way to create guaranteed royalties, this has several advantages:
+
+- Backwards Compatibility - Existing royalty implementations will still work, just not have the same assurances.
+- Minimal Storage Requirement - An optional boolean has about the smallest memory impact possible. This is especially important because it's attached to the - Reference NFT and will be set for each individual NFT.
+- Intra-Collection Utility - This already allows for minting a collection with some NFTs with royalties and some without. A future version of this standard will likely make use of this field to allow for multiple versions of royalties for even more granular control.
+
+
+## Backward Compatibility
```
Ryun1, line 152:
```suggestion
### Backward Compatibility
```

CIP-0102/README.md
```diff
@@ -0,0 +1,177 @@
...
+#### Retrieve metadata from a Plutus validator
+
+We want to bring the royalty metadata of the NFT `d5e6bf0500378d4f0da4e8dde6becec7621cd8cbf5cbb9b87013d4cc.(222)TestToken` in the Plutus validator context. To do this we
+
+1. Construct `royalty NFT` from `user token`: `d5e6bf0500378d4f0da4e8dde6becec7621cd8cbf5cbb9b87013d4cc.(500)Royalty` (off-chain)
+2. Look up `royalty NFT` and find the output it's locked in. (off-chain)
+3. Reference the output in the transaction. (off-chain)
+4. Verify validity of datum of the referenced output by checking if policy ID of `royalty NFT` and `user token` and their asset names without the `asset_name_label` prefix match. (on-chain)
+
+### Rationale
```
Ryun1, line 131:
```suggestion
## Rationale
```

CIP-0102/README.md
```diff
@@ -0,0 +1,177 @@
...
+Category: Tokens
+Created: 2023-08-08
+License: CC-BY-4.0
+---
+
+## Abstract
+
+This proposal makes use of the onchain metadata pattern established in [CIP-0068][] to provide a way to store royalties with greater assurance and customizability.
+
+## Motivation
```
Ryun1, line 18:
```suggestion
## Motivation: why is this CIP necessary?
```

CIP-0102/README.md
```diff
@@ -0,0 +1,177 @@
...
+To keep metadata compatibility with changes coming in the future, we introduce a `version` field in the datum.
+
+
+## Path to Active
+
+### Acceptance Criteria
+
+- [ ] This CIP should receive feedback, criticism, and refinement from: CIP Editors and the community of people involved with NFT projects to review any weaknesses or areas of improvement.
+- [ ] Guidelines and examples of publication of data as well as discovery and validation should be included as part of of criteria for acceptance.
+- [ ] Minimal reference implementation making use of [Lucid](https://github.com/spacebudz/lucid) (off-chain), [PlutusTx](https://github.com/input-output-hk/plutus) (on-chain): [Reference Implementation](./ref_impl/)
```
Ryun1, line 162:
```suggestion
- [ ] Minimal reference implementation making use of [Lucid](https://github.com/spacebudz/lucid) (off-chain), [PlutusTx](https://github.com/input-output-hk/plutus) (on-chain): [Reference Implementation](./ref_impl/).
```

CIP-0102/README.md
```diff
@@ -0,0 +1,177 @@
...
+; Custom user defined plutus data.
+; Setting data is optional, but the field is required
+; and needs to be at least Unit/Void: #6.121([])
+extra = plutus_data
+
+royalty_info = #6.121([royalty_recipients, version, extra])
+```
+
+#### Example of onchain variable fee calculation:
+```
```
Ryun1, line 81:
```suggestion
```cddl
```

CIP-0102/README.md
```diff
@@ -0,0 +1,177 @@
...
+Because the computational complexity of Plutus primitives scales with size, this approach significantly minimizes resource consumption.
+
+To prevent abuse, it is **recommended** that the `royalty NFT` is stored at the script address of a validator that ensures the specified fees are not arbitrarily changed, such as an always-fails validator.
+
+### Reference Datum Royalty Flag
+
+If not specified elsewhere in the token's datums, a malicious user could send transactions to a protocol which do not reference the royalty datum. For full assurances, a new optional flag should be added to the reference datum
+
+```
```
Ryun1, line 98:
```suggestion
```cddl
```

CIP-0102/README.md
```diff
@@ -0,0 +1,177 @@
...
+1. Construct `royalty NFT` from `user token`: `d5e6bf0500378d4f0da4e8dde6becec7621cd8cbf5cbb9b87013d4cc.(500)Royalty` (off-chain)
+2. Look up `royalty NFT` and find the output it's locked in. (off-chain)
+3. Reference the output in the transaction. (off-chain)
+4. Verify validity of datum of the referenced output by checking if policy ID of `royalty NFT` and `user token` and their asset names without the `asset_name_label` prefix match. (on-chain)
+
+### Rationale
+
+The specification here is made to be as minimal as possible. This is done with expediency in mind and the expectation that additional changes to the specification may be made in the future. The sooner we have a standard established, the sooner we can make use of it. Rather than attempting to anticipate all use cases, we specify with forward-compatibility in mind.
+
+#### 500 Royalty Token Datum
```
Ryun1, line 135:
```suggestion
### 500 Royalty Token Datum
```

CIP-0102/README.md
```diff
@@ -0,0 +1,177 @@
...
+
+This specification is largely based on [the royalty specification in Nebula](https://github.com/spacebudz/nebula/tree/main#royalty-info-specification), with a couple key departures:
+
+- The royalty token is recommended to be locked at a script address, rather than stored in the user's wallet. This encourages projects to guarantee royalties won't change by sending their royalties to an always-fails (or similar) script address, but still allows for creative royalty schemes and minimizes disruption to existing projects.
+
+- The policyId of the royalty NFT must match that of the reference NFT. This enables lookups based on the user token in the same way as is done for the tokens specified in the original CIP-68 standard.
+
+#### Reference Datum Flag
```
Ryun1, line 143:
```suggestion
### Reference Datum Flag
```
to align with Rationale header change

CIP-0102/README.md
```diff
@@ -0,0 +1,177 @@
...
+### Acceptance Criteria
+
+- [ ] This CIP should receive feedback, criticism, and refinement from: CIP Editors and the community of people involved with NFT projects to review any weaknesses or areas of improvement.
+- [ ] Guidelines and examples of publication of data as well as discovery and validation should be included as part of of criteria for acceptance.
+- [ ] Minimal reference implementation making use of [Lucid](https://github.com/spacebudz/lucid) (off-chain), [PlutusTx](https://github.com/input-output-hk/plutus) (on-chain): [Reference Implementation](./ref_impl/)
+- [ ] Implementation and use demonstrated by the community: NFT Projects, Blockchain Explorers, Wallets, Marketplaces.
+
+### Implementation Plan
+
+1. Publish open source reference implementation and instructions related to the creation, storage and reading of royalty utxos.
```
Ryun1, line 168:
```suggestion
- [ ] Publish open source reference implementation and instructions related to the creation, storage and reading of royalty utxos.
```

CIP-0102/README.md
```diff
@@ -0,0 +1,177 @@
...
+2. Implement in open source libraries and tooling such as [Lucid](https://github.com/spacebudz/lucid), [Blockfrost](https://github.com/blockfrost/blockfrost-backend-ryo), etc.
```
Ryun1, line 169:
```suggestion
- [ ] Implement in open source libraries and tooling such as [Lucid](https://github.com/spacebudz/lucid), [Blockfrost](https://github.com/blockfrost/blockfrost-backend-ryo), etc.
```

CIP-0102/README.md
```diff
@@ -0,0 +1,177 @@
...
+3. Achieve additional "buy in" from existing community actors and implementors such as: blockchain explorers, token marketplaces, minting platforms, wallets.
```
Ryun1, line 170:
```suggestion
- [ ] Achieve additional "buy in" from existing community actors and implementors such as: blockchain explorers, token marketplaces, minting platforms, wallets.
```

CIP-0102/README.md
rphair: https://github.com/cardano-foundation/CIPs/pull/551#discussion_r1362495469

CIP-0102/README.md
Ryun1: Maybe it is not supported, but I think it is good practice to use the label anyway

rphair:
@SamDelaney looks ready to merge now as soon as we answer the question about whether @Ryun1's suggested `cddl` code tag is valid: https://github.com/cardano-foundation/CIPs/pull/551#discussion_r1362464997

rphair:
p.s. since I think we'll therefore try to set a precedent for tagging `cddl` and probable eventual support of this syntax on GitHub, I'll commit that change & push this forward... since we did promise an effective merge today 🤓

SamDelaney:
Ah sorry about the delay here I had several meetings in quick succession. Looks like you got the tickboxes in. I was going to include Nebula & Grabbit as implementors & add the CIP 68 boilerplate, but I don't think any of those are urgent, so I'll add them to the reference implementation PR when I make that.

Thanks for the merge! Really glad to have this ratified! 🎉

rphair:
@SamDelaney since this was agreed at meeting # 75 you can add those implementors here as a change request & I'll include them. ☺️ https://github.com/cardano-foundation/CIPs/pull/606

Likewise with the boilerplate language and I apologise for missing that.  I was working to satisfy 2 goals and my head didn't sort out the schedule & priorities right.  The boilerplate is just a delta to the CIP text if I understood correctly & if you can shoehorn it into #606 then I'd be happy to merge it there if you think it's appropriate. 😖

yHSJ:
First of all, congrats on the merge @SamDelaney! I am really happy to see a push for a standard for CIP68 style royalties.

I'm sorry I'm late to the process, but I do see there are already discussions about submitting change requests so I felt I could still share my feedback:

**1) Rates that cannot be represented**

I know this was discussed in the meeting (@Crypto2099 brought it up), but the method of storing the rate is rather restrictive. There are lots of potential royalty values that cannot be represented in this way. I randomly selected one of these values (7%) and checked how many collections currently have 7% royalties in the JPG Store database. We have 529 collections with a 7% royalty. 

In my opinion, a royalty standard should support as many royalty rates as possible, and this is not achieving that goal. Why not simply store the rate as an integer x, where `x = 10000 * rate` with integer multiplication? Of course this cannot represent *all* possible royalty percentages, but there are only 14 collections with royalty values that would not be supported by this representation currently in the JPG database.

**2) Undefined Behavior and Handling Errors**
There are a couple of places in this CIP that I can identify potential undefined behavior. This is not unique to this CIP at all and lots of the issues are largely inherited from the CIPs this one is based on, but I think that this CIP could do a better job of handling them. They are:

- There is no restriction that states the required quantity of the royalty token.
  - I know the class is `NFT`, but nowhere in CIP-67 nor CIP-68 does it actually enforce that an asset in the `NFT` class have a quantity <= 1. One could argue that the name of the class makes it obvious, but it should be explicit. Furthermore, what happens if that implied restriction is violated? Are both the royalty tokens invalid? Should we accept the older token because it was valid under this specification, or the newer one because it was minted on a policy and thus the owner intended for that to become the new royalty?
- There are no restrictions (beyond the above) on fee value.
  - That means that it is valid to request, for example, a 10,000% royalty. I guess technically that could be requested, but should validators require that? Does it make sense to impose a validity range of [0,100] on the percent fee? If that is imposed, what should happen if a royalty fee is outside of that range? Should it be totally ignored or rounded to the nearest valid value?

**3) Ambiguous Language**
This is an issue with most CIPs, in my opinion. But, the language of "must" and "should" are often slightly ambiguous. Though one can reasonably infer what is absolutely required and what is simply a suggestion, I really recommend following [RFC-2119](https://datatracker.ietf.org/doc/html/rfc2119) to strictly define the language you are using, to avoid ambiguity and make the standard stronger.

SamDelaney:
Hi @yHSJ, thanks for weighing in! Let me answer these in turn:

**1) Rates that can not be represented**

I realize now I didn't include in language specifying that rounding would be allowed for values where the onchain value doesn't result in a whole number. 7% is absolutely doable with this protocol. Here's what that would look like:

```cddl
; Given a royalty fee of 1.6% (0.016)
; To store this in the royalty datum
1 / (0.07 / 10) => 142.857... => 143
; To read it back
10 / 143 => 0.06993... => 0.070
```

I think this example is probably a better illustration, and will likely replace the existing example as part of #606 assuming I get approval. I will also add a bit of wording in the specification that acknowledges rounding.

**2a) Restrictions on NFT quantity**

This is a very good point. For now, there is a valid use case for wanting multiple royalty tokens - having multiple royalty policies within a collection that can be chosen by the frontend. However, there is also potential for abuse, where collections might mint a second, much higher royalty to deceive users.

I think for now multiple royalty tokens shouldn't be explicitly supported or prohibited. If collections mint multiple tokens, which royalty datum is referenced can be left to the user or platform constructing the transaction. This makes the type of abuse I pointed out easy to circumvent, but doesn't prevent people trying to make multiple royalty policies.

I have plans for a better way of doing multiple royalty policies in a collection that I left out of version 1 for the sake of simplicity. Once that change comes along (I plan to introduce V2 shortly after I finish the reference implementation), the valid use case will be supported and it will make sense to explicitly prohibit multiple royalty tokens.

P.S. this should also be raised for clarification on CIP 67/68. Maybe in #586?
 
**2b) Restrictions on fee value**

While there isn't really any reason why NFTs can't have >100% royalties, I don't think it's fair to expect platforms to support that when many may already have checks to ensure royalties don't go above that number.

I think a **recommended** validity window of [1, 100], with the caveat that royalties outside that window MAY be rounded to fit makes the most sense. This way creators are allowed to create extreme royalties if they want, but we don't force platforms to make significant changes to logic to support minor edge cases.

**3) Ambiguous Langauge**

You're absolutely right this should be following RFC-2119. I had implicitly made that assumption, but not formatted correctly or stated that explicitly. Will do a pass to make sure this works as part of #606.

yHSJ:
Thank you for considering my feedback @SamDelaney!

**1) Rates that can not be represented**
> I realize now I didn't include in language specifying that rounding would be allowed for values where the onchain value doesn't result in a whole number.

I see, this makes a lot more sense and allows for many more potential royalty values. That said, I still feel this is more complicated than it needs to be. Why not store the fee as `round(royalty_percentage * 10^x)` where `x` allows you to specify the possible precision of the royalty?  I would suggest `x` to be 3  or 4 to easily allow for decimals.

While your encoding of the value absolutely works, I fear it is unnecessarily overcomplicated. Perhaps there are some reasons your encoding is better that I haven't considered yet, though.

**2a) Restrictions on NFT quantity**
> For now, there is a valid use case for wanting multiple royalty tokens - having multiple royalty policies within a collection that can be chosen by the frontend. However, there is also potential for abuse, where collections might mint a second, much higher royalty to deceive users.

I agree that there are cases where having multiple royalty tokens make sense. However, I think if that is valid, there MUST be a default state specified in the standard since users are not likely to specify a preferred royalty on every platform.

Furthermore, while I appreciate the use of reference datums to enforce royalties, it is important to keep in mind this would require a new contract for marketplaces, such as JPG Store, which would increase friction in adoption. JPG Store could easily much more easily support this CIP if the example section titled  `Retrieve metadata as 3rd party` was clearly stated as a valid way for a marketplace to enforce the royalties. I think that is implied, but it could be stated explicitly that an implementor can follow the standard with either example.

> P.S. this should also be raised for clarification on CIP 67/68. Maybe in https://github.com/cardano-foundation/CIPs/pull/586? 

Yes, absolutely. @Crypto2099 knows that I have a laundry list of "implementor speed bumps" we ran into when implementing the CIP at JPG Store. Hopefully I can find time to get those thoughts organized and communicated in a constructive and meaningful way soon.

**2b) Restrictions on fee value**

> While there isn't really any reason why NFTs can't have >100% royalties, I don't think it's fair to expect platforms to support that when many may already have checks to ensure royalties don't go above that number.

Completely agreed. There is no technical requirement preventing it, but it is a business requirement for many platforms. 

This is a poorly formed and incomplete thought, but would it make sense to write a "sister CIP" that allows platforms to specify the valid royalty range they support so that users can know what to expect? This could be done through an on-chain datum or an off-chain repository.

Thank you for considering my feedback and I'm happy to hear that it has been helpful. Please let me know if it makes more sense to continue this conversation in #606, or to open a new issue and discuss it there. Seems like those would be more appropriate places to have this conversation than an already merged PR.

rphair:
> @SamDelaney: Will do a pass to make sure this works as part of https://github.com/cardano-foundation/CIPs/pull/606.

Thanks for keeping an eye on that & sorry for what may have been a premature merge on this.  We will merge that PR when we get 1 more editorial approval, but that may not happen because of editors involved in the Cardadno Summit.  If it merges before you've made those language / implementation / other updates, we can do another PR on your CIP as soon as you see fit & I'll try to expedite it from this end.

> @yHSJ: Please let me know if it makes more sense to continue this conversation in https://github.com/cardano-foundation/CIPs/pull/606, or to open a new issue and discuss it there. Seems like those would be more appropriate places to have this conversation than an already merged PR.

#606 was open for @SamDelaney to make changes that were discussed at our last CIP meeting, since I inadvertently merged it before all those pending changes were included in this PR branch, but it wouldn't make sense to discuss CIP content issues there because it's only for editorial items & those last-minute additions that were already discussed.  Your contribution would be welcome if & when he submits another PR and/or in a newly created issue 🤓

PR #65, opened 2021-02-09
CIP-0013 | Extend URI scheme with single stake pool links [proposal of update]

nicarq:
Hello CIP editors!

I know that the next CIP meeting is today, so in order to be able to make it in time, I created this PR on top of the work of @rphair. 

Why?

- Fixes the usage of the terminology stake (it should be delegate). Users delegate to stakepools, and stakepools are the ones that stake.
- 
- To make this more friendly to future extension of URI, I updated it the parameter key names are always passed. In the future, we could have more advance features that would require the usage of parameter names and it will look strange that sometimes we enforce that for some actions (delegate) and for others don't.

PR #61, opened 2021-01-17
Extend URI scheme with single stake pool links

rphair:
Including stake pool links with this CIP, rather than submitting a separate CIP, has been agreed upon and discussed in https://github.com/cardano-foundation/CIPs/pull/25 where this new material has already been introduced and held up for comment.  Therefore I've also set this CIP's date (2020-10-20) back to the earlier date of the other CIP's submission (2020-09-22).

PR #559, opened 2023-07-24
CIP-0013 | Generalise for further protocols

rphair:
This change was considered in this forum thread (now added into `Discussions:` history) and deemed necessary for https://github.com/cardano-foundation/CIPs/pull/546.

It's a substantial rewrite but preserves the formatting & language for the Payment links and the additions for Stake Pool links which I added afterwards.

Notably it indicates that further extensions to the protocol (`//keyword` where the keyword is called an "authority" by URI convention) should be defined in further CIPs.  Although that may not be strictly adhered to by developers, I think that is a reasonable thing to require in _this_ CIP and for other CIPs which depend in any way on a URI scheme.

([formatted new version in branch](https://github.com/rphair/CIPs/blob/cip-0013-generalise-protocol/CIP-0013/README.md))

### DIFF
--- a/CIP-0013/README.md
+++ b/CIP-0013/README.md
@@ -4,9 +4,9 @@ Title: Cardano URI Scheme
 Status: Proposed
 Category: Wallets
 Authors:
+    - Robert Phair <rphair@cosd.com>
     - Sebastien Guillemot <sebastien@emurgo.io>
     - Vicente Almonacid <vicente@emurgo.io>
-    - Robert Phair <rphair@cosd.com>
 Implementors: N/A
 Discussions:
     - https://github.com/Emurgo/EmIPs/pull/2
@@ -15,15 +15,23 @@ Discussions:
     - https://github.com/cardano-foundation/CIPs/pull/61
     - https://github.com/cardano-foundation/CIPs/pull/86
     - https://forum.cardano.org/t/cip-stake-uri-scheme-for-pools-delegation-portfolios/40594
+    - https://forum.cardano.org/t/cip-generalized-cardano-urls/57464
+    - https://github.com/cardano-foundation/CIPs/pull/546
 Created: 2020-09-22
 License: CC-BY-4.0
 ---
 
 ## Abstract
 
-This proposal describes a basic URI scheme to handle Ada transfers and links to single stake pools or weighted lists of multiple pools.
+This proposal describes a basic URI scheme with two specific protocols to handle Ada transfers and links to weighted lists of stake pools.
 
 ## Motivation: why is this CIP necessary?
+
+### In general:
+
+Developers of protocols that use URI schemes should be able to choose unique protcol keywords for how these links are handled by applications.
+
+Beyond the two earliest defined protocols below, protocols using distinct keywords (e.g. `//stake`) can be defined in other standards and CIPs and implemented without ambiguity by applications which interpret those particular URI protocols.
 
 ### For payment URIs:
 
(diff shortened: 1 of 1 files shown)


rphair:
@Crypto2099 this should be ready for your own review as well.  Next stop for me is a detailed review of https://github.com/cardano-foundation/CIPs/pull/546 to ensure that these two CIPs will work well together. 😎

CIP-0013/README.md
```diff
@@ -49,74 +57,79 @@ URIs for weighted stake pool lists provide alternatives to using a JSON file to
 
 The core implementation should follow the [BIP-21](https://github.com/bitcoin/bips/blob/master/bip-0021.mediawiki) standard (with `bitcoin:` replaced with `web+cardano:`)
 
-Rationale:
-- Use `cardano:` over `ada:` as other projects that implement this standard tend to take the project name over the currency name (this makes sense if we consider this protocol as a generic way for interacting with the blockchain through wallets - as opposed to a simple payment system)
-- Many wallets support multiple currencies. Following the same standard will ensure higher adoption of our protocol.
-
 Examples:
 ```
 <a href="web+cardano:Ae2tdPwUPEZ76BjmWDTS7poTekAvNqBjgfthF92pSLSDVpRVnLP7meaFhVd">Donate</a>
```
Crypto2099, line 62: > `<a href="web+cardano:Ae2tdPwUPEZ76BjmWDTS7poTekAvNqBjgfthF92pSLSDVpRVnLP7meaFhVd">Donate</a>`

Is there a reason why this URI schema has persisted to this point lacking the two forward slashes? (//)

Since we are looking to standardize this and prepare for future expansion, would it be safe to require that web+cardano:// always be present?

Further, should we consider a "pay to" _authority_ rather than defaulting to hoping that it is a valid address?

CIP-0013/README.md
```diff
@@ -143,13 +162,21 @@ For background, see
...
+
+### Rationale for stake pool links
+
+#### How do URI delegation portfolio links supplement use of JSON files for the same purpose?
 
 URIs facilitate the "social element" of delegated staking and pool promotion through a socially familiar, easily accessible, and less centralised convention for sharing stake pool references and potential delegation portfolios without having to construct or host a JSON file.
 
 The processing of a JSON file delivered by a web server will depend highly on a user's platform and might not even be seen by the wallet application at all.  With a properly associated `web+cardano:` protocol, developers and users have another option available in case JSON files are not delivered properly to the wallet application.
```
Crypto2099, line 177: Could we better and more cleanly handle the varied methods of delegation via a link that would "point" to a well-defined JSON/CBOR schema metadata file that can be fetched? This would greatly simplify the URIs themselves and the associated HTML where they are embedded in addition to improving troubleshooting and diagnostics.

Crypto2099 (commented):
Overall I think this is a step in the right direction for CIP-13. It provides additional clarity and understanding for how to expand the protocol in the future as well as better defining the existing feature set that is part of the core CIP-13 itself.

CIP-0013/README.md
rphair: In the months it took the stake poll links to be put through as an addition to this CIP, the counter of a JSON file was the main objection.  Rather than look at the stake pool approach on its own merit, most CIP editors at the time and IOG developers who jumped in said that JSON would be "better" and for a while the discussion just ended there: let alone being open to a cooperative approach.  `//stake` was eventually merged when I managed to show that a rigorously defined URI syntax had fewer practical limitations than the JSON proponents had all assumed.

That is at least an indicator that a hybrid URI-to-JSON stake portfolio protocol would be unpopular.  In fact, over 2 years after those discussions, CIP-0017 remains unimplemented as far as I know

The motivation for having the one-line, simple syntax available is for the sake of web designers, SPOs, and marketers & is already in the currently defined CIP.  Other URLs (like encapsulated JSON file pathnames) or resource locators might be helpful... though still no commercial support for this kind of robust delegation that I know of.  If & when the time comes, we can suggest to those implementors that they create a new authority (e.g. `//stakelink`) with a CIP for it.

rphair: ... some dev will come along someday wanting to put urlencoded CBOR in a URL: so rather than create a precedent to support this in a constantly modified (or deprecated) CIP-0013 we can simply tell them, "Create a new `authority` keyword for it & define your support for it there" ... whether it relates to payments, delegation and/or anything else 🤓

Crypto2099: I suppose my primary objection is to the use of potentially unicode pool tickers as the "keys" in the URL arguments, I feel like this creates a cumbersome barrier to implementation for the wallets. However, I also understand this was introduced most likely in response to URL max-length limitations that could quickly be exceeded by 3+ pool hex IDs.

The parser that I've written to handle logical parsing of CIP-13 URIs does manage to handle it but has not been rigorously tested against things like URL-encoded UTF-8 tickers so will need some additional work there for matching.

https://github.com/Crypto2099/POO/blob/main/packages/lib/cardano-poo-ts/src/CIP13Parser.ts

rphair: The problem with _not_ using the stake pools as keys is that the syntax blows up, and then fails for multi pool links, when using them as values.

After months of debate, those against using the stake pools as keys couldn't come up with an alternative syntax which was concise _and_ unambiguous:

- https://github.com/cardano-foundation/CIPs/pull/65#issuecomment-783159961
- https://github.com/cardano-foundation/CIPs/pull/61#issuecomment-782141931
- https://github.com/cardano-foundation/CIPs/pull/61#issuecomment-782664031

I don't know if your alternative syntax with key-value pairs would have this same problem, but the main problem with the last suggestion (having a `poolticker` keyword with the hex pool ID as the value`) would prevent multi-pool delegation links, since URI query variables can't have more than one value:

- https://github.com/cardano-foundation/CIPs/pull/61#issuecomment-782732129

So as difficult as it might be for parsers and wallets to implement this (as already pointed out by @SEEEBBBA in the earlier threads, without giving any specifics), I don't see a way around it unless someone wants to spell it out an alternative syntax & attempt (again) another deprecation of the CIP-13 stake pool link protocol as it's been defined for the last 2 years... keeping in mind that the main problem would be the support of _multiple pool values_ rather than the length of the URI string.

rphair:
A suggestion that will help to focus and expedite the review of this particular PR: the changes here are only to refine the language, description, and grammar to support and encourage further protocols in addition to the payment (`authority` == NULL) and weighted stake pool set (`authority` = `stake`) already defined here.

The URI scheme and protocol schemes themselves are not changing even one bit in this PR.  These individual protocols are not subject to re-review, nor do they have to be validated again from first principles, because some other parts of the CIP have been reworded for inclusiveness.

Therefore I would suggest any modifications to those particular protocols be submitted as separate PRs so they can be considered separately.  This PR is only about the expansion into new protocols like `//claim` as submitted in https://github.com/cardano-foundation/CIPs/pull/546.

CIP-0013/README.md
v-almonacid: > We are locked into supporting the payment links without an authority because for a long time that was the only implementation of these URIs that was getting professional consideration (someone suggested stake pool links in 2018 but nobody ever moved on it; timeline & links are in the PRs at the Discussion links).

> As far as I know Yoroi in particular has supported payment links with a bare path-query string that whole time: about 5 years now. If I had to tell them they needed to commit to honouring an authority prefix (let's call it //payment) I would therefore expect about 5 years to get a response.

Personally I think it's not too late to change the URI scheme to add an authority for payments, if that's what you think is best (I do). And while it is difficult to say whether there is a lot of people using payments URIs right now with Yoroi, I don't think it's a big deal to add this breaking change. But it'd be better of course if you try to get feedback from wallet providers first.

> how a wallet or perhaps a dApp would handle a URI with an invalid payment address would be up to the designers

Exactly

Ryun1 (approved):
I like this update, it looks to me like it opens up the standard for further expansion and adoption without compromising backwards compatibility.

SebastienGllmt (approved):

PR #562, opened 2023-07-26
CIP-0068 | Improve image types and URI descriptions

SmaugPool:
1. This patch tries to improve 222, 333 & 444 media types and URIs descriptions.
2. It also proposes to remove the optional top level `mediaType` in 222 & 444 for the following reasons:
   * because it is optional, it cannot be relied on and is therefore basically useless at best
   * it has actually been counterproductive in CIP 25 as some users have minted NFTs with `video/...` or `model/...` media (mime) type for `image` URI, ignoring the `image/*` constraint (I will propose to remove it too in CIP25 if this is accepted here)
3. It adds the Arweave scheme (`ar://`) in the list of supported URI schemes. I wonder though if the CIP should have such a hard list of supported schemes or just suggestions?
4. It improves the References section and uses links instead of text

PS: Not sure how to avoid the duplication of information in each label. It would also be nice to link to [related information in CIP25 about URIs and data URLs](https://github.com/cardano-foundation/CIPs/blob/ff07d2b747b7e9228ec1d9bbd6326ab6663c0303/CIP-0025/README.md?plain=1#L82) but not sure how.

### DIFF
--- a/CIP-0068/README.md
+++ b/CIP-0068/README.md
@@ -124,19 +124,27 @@ files_details =
   {
     ? name : bounded_bytes, ; UTF-8
     mediaType : bounded_bytes, ; UTF-8
-    src : bounded_bytes ; UTF-8
+    src : uri,
     ; ... Additional properties are allowed
   }
 
 metadata = 
   {
     name : bounded_bytes, ; UTF-8
-    image : bounded_bytes, ; UTF-8
-    ? mediaType : bounded_bytes, ; UTF-8
+
+    ; The image URI must point to a resource with media type (mime type) `image/*`
+    ; (for example `image/png`, `image/jpeg`, `image/svg+xml`, etc.)
+    image : uri,
+
     ? description : bounded_bytes, ; UTF-8
     ? files : [* files_details]
     ; ... Additional properties are allowed
   }
+
+; A valid Uniform Resource Identifier (URI) as a UTF-8 encoded bytestring.
+; The URI scheme must be one of `https` (HTTP), `ipfs` (IPFS), `ar` (Arweave) or `data` (on-chain).
+; Data URLs (on-chain data) must comply to RFC2397.
+uri = bounded_bytes ; UTF-8
   
 ; Custom user defined plutus data.
 ; Setting data is optional, but the field is required
@@ -204,17 +212,19 @@ metadata =
     description : bounded_bytes, ; UTF-8
     ? ticker: bounded_bytes, ; UTF-8
     ? url: bounded_bytes, ; UTF-8
-    ? logo: uri,
     ? decimals: int
+
+    ; 'logo' does not follow the explanation of the token-registry, it needs to be a valid URI and not a plain bytestring.
+    ; The logo URI must point to a resource with media type (mime type) `image/png`, `image/jpeg` or `image/svg+xml`.
+    ? logo: uri,
+
     ; ... Additional properties are allowed
   }
 
-; A URI as a UTF-8 encoded bytestring.
-; The URI scheme must be one of `https`, `ipfs` or `data`
-; Do not encode plain file payloads as URI.
-; 'logo' does not follow the explanation of the token-registry, it needs to be a valid URI and not a plain bytestring.
-; Only use the following media types: `image/png`, `image/jpeg`, `image/svg+xml`
-uri = bounded_bytes
+; A valid Uniform Resource Identifier (URI) as a UTF-8 encoded bytestring.
+; The URI scheme must be one of `https` (HTTP), `ipfs` (IPFS), `ar` (Arweave) or `data` (on-chain).
+; Data URLs (on-chain data) must comply to RFC2397.
+uri = bounded_bytes ; UTF-8
 
 ; Custom user defined plutus data.
 ; Setting data is optional, but the field is required
@@ -277,27 +287,28 @@ files_details =
   {
     ? name : bounded_bytes, ; UTF-8
     mediaType : bounded_bytes, ; UTF-8
-    src : bounded_bytes ; UTF-8
+    src : uri,
     ; ... Additional properties are allowed
   }
 
 metadata = 
   {
     name : bounded_bytes, ; UTF-8
-    image : bounded_bytes, ; UTF-8
-    ? mediaType : bounded_bytes, ; UTF-8
+
+    ; The image URI must point to a resource with media type (mime type) `image/*`
+    ; (for example `image/png`, `image/jpeg`, `image/svg+xml`, etc.)
+    image : uri,
+
     ? description : bounded_bytes, ; UTF-8
     ? decimals: int,
     ? files : [* files_details]
     ; ... Additional properties are allowed
   }
 
-; A URI as a UTF-8 encoded bytestring.
-; The URI scheme must be one of `https`, `ipfs` or `data`
-; Do not encode plain file payloads as URI.
-; 'logo' does not follow the explanation of the token-registry, it needs to be a valid URI and not a plain bytestring.
-; Only use the following media types: `image/png`, `image/jpeg`, `image/svg+xml`
-uri = bounded_bytes
+; A valid Uniform Resource Identifier (URI) as a UTF-8 encoded bytestring.
+; The URI scheme must be one of `https` (HTTP), `ipfs` (IPFS), `ar` (Arweave) or `data` (on-chain).
+; Data URLs (on-chain data) must comply to RFC2397.
+uri = bounded_bytes ; UTF-8
 
 ; Custom user defined plutus data.
 ; Setting data is optional, but the field is required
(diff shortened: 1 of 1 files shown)


CIP-0068/README.md
```diff
@@ -124,19 +124,23 @@ files_details =
...
-    src : bounded_bytes ; UTF-8
+    src : uri,
     ; ... Additional properties are allowed
   }
 
 metadata = 
   {
     name : bounded_bytes, ; UTF-8
-    image : bounded_bytes, ; UTF-8
-    ? mediaType : bounded_bytes, ; UTF-8
```
rphair, line 135: (potentially requiring version bump to 2 or 3)

CIP-0068/README.md
```diff
@@ -277,27 +283,25 @@ files_details =
...
+    src : uri,
     ; ... Additional properties are allowed
   }
 
+; Only use `image/*` media types (for example `image/png`, `image/jpeg`, `image/svg+xml`, etc.)
 metadata = 
   {
     name : bounded_bytes, ; UTF-8
-    image : bounded_bytes, ; UTF-8
-    ? mediaType : bounded_bytes, ; UTF-8
```
rphair, line 288: (potentially requiring version bump to 2 or 3)

rphair (commented):
I don't work on these implementations but this looks to me like it clarifies the use of URIs & images and I think it would be an improvement.  As soon as we can agree on whether or not a version bump for `222` and `444` is needed I'm happy to sign off on this.

> PS: Not sure how to avoid the duplication of information in each label.

The only portable way I know of would be to put it in a separate file... but people would expect that for a schema and not for written text.
- Maybe someday it will work better as a schema?
- Could there be any schema format that's even somewhat as human readable as what's there now?

> It would also be nice to link to related information in CIP25 about URIs and data URLs but not sure how.

As it stands now, not even section anchored links (e.g. `#Abstract`) between CIPs are portable.  Since line numbers will also change, I think the only approach now is to refer to statements in other CIPs literally (e.g. CIP > Heading > Subheading > quote).

This deficiency has been around since https://github.com/cardano-foundation/CIPs/issues/109 and in general we are trying fix the raw text of all CIPs via https://github.com/cardano-foundation/CIPs/issues/389 before someone can really work on better scripts for the web sites derived from GitHub source.

@SmaugPool you have my ✔️ on this if we can find developers commenting here who say unanimously that we don't need a version bump because the dropped parameter.  Otherwise I think we could sign off on it right away if you add the version bumps.

The editors have agreed to be cautious here to avoid having to do something like https://github.com/cardano-foundation/CIPs/pull/523 again in a hurry (amazingly this PR is still not merged; so any bumps here would have to be done relative to _those_ bumps 😜). cc @KtorZ @mmahut

Crypto2099:
> 2. It also proposes to remove the optional top level `mediaType` in 222 & 444 for the following reasons:
>    
>    * because it is optional, it cannot be relied on and is therefore basically useless at best
>    * it has actually been counterproductive in CIP 25 as some users have minted NFTs with `video/...` or `model/...` media (mime) type for `image` URI, ignoring the `image/*` constraint (I will propose to remove it too in CIP25 if this is accepted here)

I'm not sure I understand the rationale to remove an optional top-level mediaType. If anything, I believe it would be more helpful to refine and iterate the version of the standard passing in a well-formed object for all files such as:

```cbor
file-ref = [
  media-type,
  uri
]

image-file-ref = [
    image-media-type,
    uri
]

image-media-type = 'image/jpeg' | 'image/png' | 'image/svg+xml'
```

We cannot write standards with the expectation that we will have to make exceptions because people will incorrectly implement the standards. We should write standards in such a way that if it is followed correctly, all the relevant and pertinent information a user might need will be where they expect it to be for ease of integration.

> 3. It adds the Arweave scheme (`ar://`) in the list of supported URI schemes. I wonder though if the CIP should have such a hard list of supported schemes or just suggestions?

I think it makes sense that the standard should define which URI schemes should be supported. There are other schemes (ftp:, network:, etc) that would pass validation as a "valid URI" but would most likely break all integrations due to lack of support.

SmaugPool:
> I'm not sure I understand the rationale to remove an optional top-level mediaType.

The rationale is that I'm aware of 0 use case where it has been useful, and that's because it's already constrained to `image/*`, and is optional so implementations should work without anyway (contrary to the one in `files`). On the other hand I'm aware of several cases where it has been misused (with CIP25 but it's the same).

Therefore I think we should either remove it or make it mandatory. Browsers and most image viewers are able to detect most image types, so I'm not sure it's really useful to make it mandatory as long as only images are supported. 

If we worry about exotic image formats and think it could be useful in this case, an alternative could be to list explicitly supported image formats.

Admittedly I may be biased by my own experience, so if anyone is aware of a single real-world use case where it has proved useful, please share it.

CIP-0068/README.md
Ryun1: Agreed!

CIP-0068/README.md
Ryun1: Agreed!

SmaugPool: ìmage` is not removed, and `mediaType` was optional, so I'm not sure of the benefit of bumping the version?

SmaugPool: It would not be erroneous because additional properties are allowed:

```
; ... Additional properties are allowed
```

rphair: thanks @SmaugPool - I missed that! 🤓 @Ryun1 we can talk about any other reasons why we might need the version bump @ the CIP meeting today, but it seems like we are OK without one (cc @KtorZ).

rphair (approved):
I believe the question of versioning (i.e. not strictly required) has been answered in https://github.com/cardano-foundation/CIPs/pull/562#discussion_r1279969925 ... the guideline `... Additional properties are allowed` effectively makes any property OK to use, so I now fully see @SmaugPool's point that having a property in this list strongly encourages its use and, according to the OP's argument, its misuse.

So the remaining question as per https://github.com/cardano-foundation/CIPs/pull/562#issuecomment-1653815204 is whether that misuse should be addressed here in this PR:

- This could be out of scope for this PR, but it _really_ be confusing to delete this property and then have to reintroduce it later to add a more rigorous definition as @Crypto2099 suggests above.  That would be item # 2 to discuss at the CIP editors' meeting today.
- Item # 1 will still be to call for any other motivation for bumping the version number (in my opinion we are OK without it but we need to be sure).

The definition question is beyond my technical scope as a reviewer so in the meantime I'm approving it here and will leave further reviews to reflect practical experience about current & potential use cases of `mediaType`.

SmaugPool:
Note that my initial motivation to clarify the image `mediaType`situation is that CIP68 currently does not restrict it to `image/*` in 222 and 444 labels. I believe it was @alessandrokonrad goal to mimic CIP25 metadata structure and keep this restriction but I will let him confirm.

Without this important restriction and with the field being optional, any file type could theoretically be linked from `image`, potentially without any mime type indicated, making it very difficult to render them. It could be argued that the `image` field name is clear enough but past experience show it's not.

(Note: I have updated the patch to clarify the image media type restrictions for 222, 333 and 444).

Crypto2099 (approved):
Aside from the couple of points, particularly surrounding cross-compatibility with the Token Registry standard, I believe that this PR is acceptable to be merged as-is. Given that we are dropping an optional field that is also covered by the fact that any and all additional properties are accepted as part of the existing object, I do not believe that a version iteration is needed to implement these changes and these edits provide greater clarity to the existing standard rather than fundamentally modifying it.

CIP-0068/README.md
```diff
@@ -204,17 +212,19 @@ metadata =
...
-; A URI as a UTF-8 encoded bytestring.
-; The URI scheme must be one of `https`, `ipfs` or `data`
-; Do not encode plain file payloads as URI.
-; 'logo' does not follow the explanation of the token-registry, it needs to be a valid URI and not a plain bytestring.
-; Only use the following media types: `image/png`, `image/jpeg`, `image/svg+xml`
-uri = bounded_bytes
+; A valid Uniform Resource Identifier (URI) as a UTF-8 encoded bytestring.
+; The URI scheme must be one of `https` (HTTP), `ipfs` (IPFS), `ar` (Arweave) or `data` (on-chain).
+; Data URLs (on-chain data) must comply to RFC2397.
+uri = bounded_bytes ; UTF-8
```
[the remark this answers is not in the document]
Crypto2099: Point of clarification cause I may have misunderstood here, the token registry stores the image in *ONLY* data: format by default and that is actually an acceptable URI in this case but we just want it to be a "plain text" URI rather than a raw bytestring?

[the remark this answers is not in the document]
SmaugPool: I think this part was written by @AndrewWestberg, maybe he can clarify as I just moved things around.
In the token registry, images are base64 encoded PNGs, they are not data URLs, but I think data urls are valid here.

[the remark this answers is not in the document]
AndrewWestberg: @alessandrokonrad can comment on this. My addition was only to add 444 which is a copy paste of parts of 222 and 333.

KtorZ (approved):
Seems like a case of "reality that has caught up with the specs", and the clarification seems fair 👍

Ryun1 (approved):

PR #563, opened 2023-07-26
CIP-0001 | Add Versioning + other small edits

Ryun1:
### Changes
- Added stipulation that CIPs must include a versioning scheme (Previously no such requirement existed).
- Added that on-chain metadata CIPs must include a CDDL schema (Previously this has been asked for by editors, but was unreflected here).
- Added suggestion that "pre-numbered" CIPs can be named `CIP-XXXX` or `CIP-????` (previously just `CIP-????` suggested).

### Motivation
- Lack of versioning in CIPs has been a long standing frustration in the community. Here I am trying to focus this conversation to reach a conclusion. I do not believe these first additions are sufficient, rather I want to hear communities opinions and this PR can act as a first step.
- We as CIP editors ask authors for this, thus it needs to be mentioned in CIP-0001.
- For Windows users using `?` in filenames is not possible, so offering an standard alternative seems sensible. 

### TODO:
- [x] Reflect historical versioning frustrations in `Rationale` section.
  - Won't do - save this for a larger discussion.


CIP-0001/README.md
```diff
@@ -127,7 +127,7 @@ License: CC-BY-4.0
 
 ##### Repository Organization
 
-A CIP must be stored in a specific folder named after its number (4-digit, left-padded with `0`) and in a file called `README.md`. Before a number is assigned, use `????` as a placeholder number (thus naming new folders as `CIP-????`). After a number has been assigned, rename the folder.
+A CIP must be stored in a specific folder named after its number (4-digit, left-padded with `0`) and in a file called `README.md`. Before a number is assigned, use `????` or `XXXX` as a placeholder number (thus naming new folders as `CIP-????` or `CIP-XXXX`). After a number has been assigned, rename the folder.
```
rphair, line 130:
```suggestion
A CIP must be stored in a specific folder named after its number (4-digit, left-padded with `0`) and in a file called `README.md`. Before a number is assigned, use `????` or `XXXX` as a placeholder number (thus naming new folders as `CIP-????` or `CIP-XXXX`) or a brief, descriptive text string (e.g. `CIP-my-new-feature`). After a number has been assigned, rename the folder.
```
Several CIPs during the draft process have used descriptive names for the content folder instead of the generic 4-character placeholders `????` and `XXXX`.  This also gives authors the opportunity to name the folder identically with their branch.  Most importantly it gives readers & editors a means to find the added content folder in the PR branch according to some meaningful name.

If there is any doubt about this I'll go back in the commit history for CIPs in which this was used during the editing process & post some links here.  For now I can confirm in my own memory that most times I've suggested this as a third option it's almost always been taken over the 4-character placeholders and sometimes imitated spontaneously by other authors.

Ryun1: Yup good spot, I think this is very reasonable.

rphair (approved):
OK, with https://github.com/cardano-foundation/CIPs/pull/563/commits/4877c079a07ec5ceec54b8db30d3691f0768abff I am in favour of merging this for all the reasons already given.

Crypto2099 (approved):
This seems like a simple first step towards a larger discussion on versioning and organization as a whole and the change to pre-numbered organization (thank you) is also a minor change.

Ryun1:
@rphair @Crypto2099 

Thanks for the suggestions, I like the idea of just merging this quickly as a first step and thus I have updated the PR description to better reflect this.

I think we should wait for another author's approval before merging though 🤓

> offer versioning with perhaps a different scheme than the one we've anticipated

Good idea - I have adjusted the wording to address this. As well as made the wording a bit more explicit in general.

rphair:
@Ryun1 @KtorZ if we agree to do this we should also add a "versioning remediation" step to the outline here (at least to consider whether such remediation is possible, especially for the CIPs which have been around for a while): https://github.com/cardano-foundation/CIPs/issues/389#issue-1470899761

CIP-0001/README.md
```diff
@@ -127,7 +127,7 @@ License: CC-BY-4.0
 
 ##### Repository Organization
 
-A CIP must be stored in a specific folder named after its number (4-digit, left-padded with `0`) and in a file called `README.md`. Before a number is assigned, use `????` as a placeholder number (thus naming new folders as `CIP-????`). After a number has been assigned, rename the folder.
+A CIP must be stored in a specific folder named after its number (4-digit, left-padded with `0`) and in a file called `README.md`. Before a number is assigned, use `????` or `XXXX` as a placeholder number (thus naming new folders as `CIP-????` or `CIP-XXXX`) or a brief, descriptive text string (e.g. `CIP-my-new-feature`). After a number has been assigned, rename the folder.
```
KtorZ, line 130:
```suggestion
A CIP must be stored in a specific folder named after its number (4-digit, left-padded with `0`) and in a file called `README.md`. Before a number is assigned, use `????` as a placeholder number (thus naming new folders as `CIP-????`). After a number has been assigned, rename the folder.
```

KtorZ: Rationale: we don't need to give multiple options here or people won't know which one to pick from. There's no benefits in having multiple options so let's keep it simple.

rphair: So our meeting feedback is that anything beginning with `CIP-` should be OK... without mentioning the OS-offending `?`characters... with a suggestion to use a text string abbreviating the CIP title, and a warning about using number because of the possibility of collisions or contradictions with the number chosen by editors.

Ryun1: @KtorZ @rphair 

Check the changes in [05975d1](https://github.com/cardano-foundation/CIPs/pull/563/commits/05975d11318f01bf47208c753e8bc402a43c8e74). I have tried to align with the discussion of the meeting.

rphair: As long as semantic, meaningful folder names can still be used, I'm still happy with it 😎✔️

SebastienGllmt (approved):

PR #574, opened 2023-08-02
CIP-0050 | Fix relative links

rphair:
Usual problem: not having any screening process or script to find the relative links not prefixed by `./`, we will currently run full speed ahead into this when the merged CIP is built on the Developer Portal (see below; cc @katomm @fill-the-fill @rdlrt).

@KtorZ if you have ideas about how to deal with this proactively maybe we can work together to automate the checks somehow.  This is one case where the right approach to scripting could save a lot of manual work by editors. 🤔

---

```
Error: ENOENT: no such file or directory, open './static/img/cip/./CIP-0050/images/k-effective.png'
... path: './static/img/cip/./CIP-0050/images/k-effective.png'
}
```

### DIFF
--- a/CIP-0050/README.md
+++ b/CIP-0050/README.md
@@ -41,13 +41,13 @@ The Cardano network currently produces ~21,600 blocks per epoch with ~2400 group
 $Ref (1)$
 
 <p align="center">
- <img src="images/equation1.png" width="500">
+ <img src="./images/equation1.png" width="500">
 </p>
 
-![Figure 1](images/k-effective.png)
+![Figure 1](./images/k-effective.png)
 Figure 1. Historical k-effective from epoch 245 to present.
 
-![Figure 2](images/k-effective-table.png)
+![Figure 2](./images/k-effective-table.png)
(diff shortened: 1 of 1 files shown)


katomm:
Thanks for flagging @rphair. To me, this is a problem of how the script on the developer portal side handles things, so the issue should be opened on the developer portal. (we cannot rely on 3rd party content to be delivered in a specific syntax)

rphair:
@katomm that's true, but I believe the bare (unprefixed by `./`) relative links also don't build properly on `cips.cardano.org` ... or didn't (maybe @KtorZ can recall).  If true this would be at least a short-term argument for addressing the problem at source: since the more qualified pathnames will never be invalid relative to their own `README` file.

rphair:
thanks to @katomm & downstream https://github.com/cardano-foundation/developer-portal/pull/1122 some of the heat may be off about this & the long standing similar problems... but we should merge this soon just in case 😇

Ryun1 (approved):

KtorZ (approved):

PR #581, opened 2023-08-14
anyajanat-

aleeusgr:
A working title, meaning nothing but a mistranslation of the word Fellowship into Sanscrit.

A pleasant word in an obscure language with the same meaning should be chosen.


aleeusgr:
Thank you for suggesting the right channel

aleeusgr:
https://forum.cardano.org/t/cardano-is-a-sect-we-should-act/120861

rphair:
@aleeusgr I'll keep an eye on the material you have been posting & in case some generalised system for "admonition tokens" or something applicable as a standard comes out, we can help refine it into a CIP if you are still interested.

aleeusgr:
@rphair I am all about discussion and feedback.

The provocative language serves this purpose as well.

I would be honoured. Many thanks.

My next step is to take the rather domain specific language to the certification working group and discuss the process of formalization.

aleeusgr:
https://github.com/input-output-hk/Certification-working-group/pull/28

PR #242, opened 2022-04-05
CIP Liesenfelt Shelleys Voltaire decentralization update

michael-liesenfelt:
Improving decentralization is absolutely necessary for the long term health and growth of the Cardano ecosystem. The current reward formula has resulted in a stable but stagnant level of decentralization. With the benefit of hindsight over the last year the intent of (a0, k) has not resulted in the desired decentralization outcome. This CIP provides the justification, methods, metrics, and implementation schedule for an improvement program to increase decentralization of the Cardano network.

The proposed reward equation retains the function of k for diminishing rewards based on stake but repurposes the a0 parameter for enforcing diminishing rewards based on pledge leverage. The proposed equation enforces a set of principles to ensure stakeholders of dramatically different size can all reach the same maximum yield. The yield ceiling feature prevents the formation of two classes of stakeholders and removes some of the benefits of centralization. The economic motivations of the largest stakeholders will be aligned with decentralization, reward diversification, fault tolerance, and ensuring the sybil protection of the entire community.

PR #582, opened 2023-08-15
CIP-0050 | Translations: fix Spanish metadata, remove Chinese

rphair:
@michael-liesenfelt it appears the Chinese translation file was just a stub, an empty file: so I'm deleting it here.  If there used to be translated content in a file `README.Chinese.md` in your branch (https://github.com/michael-liesenfelt/CIPs/tree/CIP-Liesenfelt-Shelleys_Voltaire_decentralization_update/CIP-0050) please let me know where I can find it... from the version history in that branch it looks like it's always been empty.

For the remaining Spanish version, I've also re-translated the title to update it from the "Shelley" title & can verify it makes sense, but not as a native speaker would.

I've had to put `NA` for a commit hash because the translation was only intermediate before the final revision in the PR which included a lot of reformatting work on the main language version.  I believe we will have to use that syntax because an empty list doesn't make sense in this context.
- If this is acceptable @KtorZ @Ryun1 please confirm & I'll submit a minor update to CIP-0001, or add it to https://github.com/cardano-foundation/CIPs/pull/563, saying what to do in these special cases (without encouraging the submissions of unmaintainable translations, or translations of intermediate drafts).  I don't think it will ever make sense to use a commit hash from the PR branch itself.

I've identified the Spanish translator from this posting (https://github.com/cardano-foundation/CIPs/pull/242#issuecomment-1140374376) and confirmed it in (and moved it from) the document footer... there's no email address available from the web site or any other public feed.  The Twitter account listed on their web site is defunct and not much else is there except Discord: definitely no GitHub account to tag here.

@michael-liesenfelt I do believe it would be worth finding some Spanish maintenance for this now official translation because it would help the consideration of this proposal if & when the RSS proposals are finally opened up to the Governance process.


rphair:
https://github.com/cardano-foundation/CIPs/pull/582/commits/a39d257e2d532b8413b5634b7cf99d0c943a7f78 set the `Translators:` to a list (as it originally should have been, according to what's already written in CIP-0001) and uses the date in the PR translation credit as the `Revision:`.

I'll submit a separate PR to adjust the CIP-0001 language saying, as agreed in the CIP meeting today (with @SebastienGllmt @KtorZ @Ryun1) that `Revision` can be parsed either as a 7-digit repository commit hash (if there is one) or the best known date of the material that was translated: if possible, the date this material was posted in its own PR branch (if not; so it should never be undefined).

Ryun1 (approved):

rphair:
@KtorZ I've formalised the use of ISO date in the `Revision` field in https://github.com/cardano-foundation/CIPs/pull/585 ... can you please review this in accordance with the suggested new language?

SebastienGllmt (approved):

PR #583, opened 2023-08-17
Proposed new CIP-101????

johnnygreeney:
Proposed new CIP number 101 (or 404). A to Z Glossary of Cardano. See README.md, should be self-explanatory

### DIFF
--- a/CIP-0101 proposed new CIP/README.md
+++ b/CIP-0101 proposed new CIP/README.md
@@ -1,0 +1,778 @@
+---
+CIP: ???? 101 or 404  are proposed numbers
+Title: A to Z of Cardano - Glossary of Cardano Terms 
+Authors: John Greene <john.greene@cardanofoundation.org>
+Status: Proposed
+Type: Informational
+Created: 2023-08-16
+License: CC-BY-4.0
+---
+
+## Simple Summary / Abstract
+
+Cardano is laden with jargon and terminology, let this CIP be an official Glossary which everyone can reference.   
+
+## Motivation / History
+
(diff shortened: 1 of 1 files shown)


rphair:
This is perhaps some useful material but it is absolutely not a suitable subject for a CIP.  I would recommend formatting it as a web or GitHub document page and submitting it here: https://github.com/input-output-hk/essential-cardano

rphair:
Good work 🤓

NetWalker108:
@rphair why is this not a suitable subject as a CIP, especially an informational CIP?

rphair:
It does not describe an _improvement_ to Cardano; therefore it is not a Cardano _Improvement_ Proposal.  Improvements are defined as _designs_ (see beginning of [CIP-0001](https://github.com/cardano-foundation/CIPs/blob/master/CIP-0001/README.md)) and there is no design in a glossary or any other mundane set of facts or descriptions.

BTW the term "Informational" CIP is obsolete.  `Informational` used to be a valid `Type` field, but that field and all its contents have not been assigned to new CIPs for quite some time and they'll be removed from old CIPs in the upcoming months.  CIPs classified as "Informational" were also about designs but documented after the designs were produced (hence not subject to so much debate).

NetWalker108:
> BTW the term "Informational" CIP is obsolete.  Informational used to be a valid Type field, but that field and all its contents have not been assigned to new CIPs for quite some time and they'll be removed from old CIPs in the upcoming months.

This is a very odd and rather backwards approach that's being used for the CIP process. 

@KtorZ @Crypto2099 you all approved this?

PR #585, opened 2023-08-25
CIP-0001 | Allow ISO date if translation source uncommitted

rphair:
Motivation and method already discussed at CIP meeting with characteristic example still pending merge in https://github.com/cardano-foundation/CIPs/pull/582.

One side effect of this PR: original language said to use a 8-10 digit commit hash, while the GitHub histories contain only 7-digit commit hashes.  As already identified in https://github.com/cardano-foundation/CIPs/pull/539#pullrequestreview-1514656299, searches for longer hashes on the GitHub history page will fail if longer hashes are used, making the hashes more difficult in general to verify.

### DIFF
--- a/CIP-0001/README.md
+++ b/CIP-0001/README.md
@@ -158,7 +158,7 @@ Field          | Description
 `CIP`          | The CIP number (without leading 0)
 `Source`       | The canonical GitHub link to the original CIP document
 `Title`        | A translation of the CIP's title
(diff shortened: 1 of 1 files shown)


Ryun1 (approved):
LGTM 🦸‍♂️

Crypto2099 (approved):
No complaints, matches what was discussed in previous meeting.

SebastienGllmt (approved):

PR #591, opened 2023-09-10
CIP-0054 | Fix API headers for markdown portability

SmaugPool:
`\<` is supported by GitHub to escape HTML but not part of Markdown spec and not supported by showdown, breaking CIP-0054 rendering on https://cips.cardano.org/:
image

By escaping instead the opening angle bracket with `&lt;`, both GitHub and showdown rendering work correctly:
image

GitHub rendering:
https://github.com/SmaugPool/CIPs/tree/cip54-fix-rendering/CIP-0054

FYI: @kieransimkin

### DIFF
--- a/CIP-0054/README.md
+++ b/CIP-0054/README.md
@@ -160,11 +160,11 @@ It is recommended that the Smart NFT API not be injected for every NFT – only 
 Although an asynchronous API is specified – so the data could be retrieved at the time when the NFT actually requests it – it is expected that in most instances the site which renders the NFT would gather the relevant transaction logs in advance, and inject them into the `<iframe>` sandbox at the point where the sandbox is created, so that the data is immediately available to the NFT without having to
 perform an HTTP request. 
 
-### cardano.nft.fingerprint: String
+### `cardano.nft.fingerprint: String`
 
 The fingerprint of the current token - in the case where we're rendering a child token, this will be the fingerprint of the child token. 
 
-### cardano.nft.metadata : Array
+### `cardano.nft.metadata : Array`
  
 The content of the 721 key from the metadata json from the mint transaction of the current NFT - if we are rendering on behalf of a child NFT, this will be the metadata from the child NFT.
 
--- a/CIP-0054/README.md
+++ b/CIP-0054/README.md
@@ -215,7 +215,7 @@ This function will return a list of the tokens held by the address specified in 
 }
 ```
 
-### cardano.nft.getFileURL( string id = null, string fingerprint = null ) : Promise&lt;String>
+### `cardano.nft.getFileURL( string id = null, string fingerprint = null ) : Promise<String>`
 
 Errors: `APIError`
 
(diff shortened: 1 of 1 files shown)


CIP-0054/README.md
```diff
@@ -168,7 +168,7 @@ The fingerprint of the current token - in the case where we're rendering a child
  
 The content of the 721 key from the metadata json from the mint transaction of the current NFT - if we are rendering on behalf of a child NFT, this will be the metadata from the child NFT.
 
-### cardano.nft.getTransactions( string which,  paginate: Paginate = undefined ) : Promise\<Object>
+### cardano.nft.getTransactions( string which,  paginate: Paginate = undefined ) : Promise&lt;Object>
```
rphair, line 171:
```suggestion
### `cardano.nft.getTransactions( string which,  paginate: Paginate = undefined ) : Promise<Object>`
```

CIP-0054/README.md
```diff
@@ -191,7 +191,7 @@ This function will return a list of transaction hashes and metadata relating to
 ```
 For simplicity, we do not include anything other than the txHash and the metadata – since any other relevant details about the transaction can always be encoded into the metadata, there is no need to over-complicate by including other transaction data like inputs, outputs or the date of the transaction etc. That is left for a potential future extension of the API to include more full GraphQL support.
 
-### cardano.nft.getTokens( string address, paginate: Paginate = undefined ) : Promise\<Object>
+### cardano.nft.getTokens( string address, paginate: Paginate = undefined ) : Promise&lt;Object>
```
rphair, line 194:
```suggestion
### `cardano.nft.getTokens( string address, paginate: Paginate = undefined ) : Promise<Object>`
```

CIP-0054/README.md
```diff
@@ -215,7 +215,7 @@ This function will return a list of the tokens held by the address specified in
 }
 ```
 
-### cardano.nft.getFileURL( string id = null, string fingerprint = null ) : Promise\<String>
+### cardano.nft.getFileURL( string id = null, string fingerprint = null ) : Promise&lt;String>
```
rphair, line 218:
```suggestion
### `cardano.nft.getFileURL( string id = null, string fingerprint = null ) : Promise<String>`
```

rphair (commented):
@SmaugPool can you anticipate any problems with Showdown rendering if these API headers were more properly typeset as code?  https://gist.github.com/rphair/d206e385ef694396841f765a6dd27696

rphair:
p.s. @SmaugPool @Ryun1 this is also an issue for CIP-0030 (I just checked all merged CIP `README`'s only found this additional one with code in the headers) so I'm going to commit https://github.com/cardano-foundation/CIPs/pull/591#pullrequestreview-1618901269 so we can test it a bit more on the smaller & less popular CIP before including those changes here as well.

SmaugPool:
> @SmaugPool can you anticipate any problems with Showdown rendering if these API headers were more properly typeset as code? https://gist.github.com/rphair/d206e385ef694396841f765a6dd27696

Inline code in the headers also solves the issue but it's displayed differently, and I didn't want to change the author formatting intention. I guess it's also an acceptable solution though if preferred.

rphair:
OK @SmaugPool @Ryun1 let's put this through then (unless @kieransimkin has any objections) and after it's agreed upon I'll submit a PR to update CIP-0030 the same way... this may require some care because of potential merge conflicts with pending CIP-0030 updates, so I think we should keep them separate.

rphair (approved):

Crypto2099 (approved):

PR #593, opened 2023-09-15
CIP-0025 | Change "NFT" to "token"

BrockCruess:
As we have standardized the use of fungible tokens for media/collectible tokens (ie: art editions & book copies), we need to avoid using "NFT" as a blanket term in CIP documentation as it can create confusion around whether a CIP includes these other token variants.

### DIFF
--- a/CIP-0010/registry.json
+++ b/CIP-0010/registry.json
@@ -25,7 +25,7 @@
   },
   {
     "transaction_metadatum_label": 721,
-    "description": "CIP-0025 - NFT Metadata Standard"
+    "description": "CIP-0025 - NFT Token Standard"
   },
   {
     "transaction_metadatum_label": 777,
--- a/CIP-0025/README.md
+++ b/CIP-0025/README.md
@@ -1,6 +1,6 @@
 ---
 CIP: 25
-Title: Media NFT Metadata Standard
+Title: Media Token Metadata Standard
 Authors: Alessandro Konrad <alessandro.konrad@live.de>, Smaug <smaug@pool.pm>
 Comments-URI:
 Status: Active
@@ -12,7 +12,7 @@ License: CC-BY-4.0
 
 ## Abstract
 
-This proposal defines an Media NFT Metadata Standard for Native Tokens.
+This proposal defines an Media Token Metadata Standard for Native Tokens.
 
 ## Motivation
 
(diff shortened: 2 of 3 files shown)


rphair (commented):
@BrockCruess I'm trimming the title by removing _avoiding exclusion of FT and RFT media tokens_ because there could well be additional reasons for doing this.  I'm putting it on the agenda for the [next CIP editors' meeting](https://hackmd.io/@cip-editors/73) so we can start thinking about any broader significance of this change beyond simply including semi-fungible tokens.

My "Devil's Advocate" argument for _not_ changing CIP-0025 would be that tokens with some fungibility would be covered more currently by CIP-0068, while on the other hand pure applications of CIP-0025 were more specifically conceived for NFTs and therefore the NFT term serves a historical purpose by justifying the design decisions that were documented at the time.

BrockCruess:
@rphair fair point on the title. However I'd personally prefer we maintain separation of CIP-0068 and CIP-0025, and I know I'm not alone on that. A lot of people already use CIP-0025 for RFTs and a whole art platform is now built and thriving on it (528 artists, 31092 works). There's realistically no downside to the inclusion, and cementing something historically - for the sake of history alone - just leads to bloat in standards. We'd have to copy and paste CIP-0025 and just add RFTs to the new CIP.

Looking forward to the outcome of the meeting's discussion!

Crypto2099:
For what it's worth, I do not believe that this is a "breaking change" and is how many "support" platforms have operated already (i.e. treat every token w/ 721 data the same regardless of the quantity on chain). 

So, I don't feel there are any major hurdles to making this relatively minor change to the wording of CIP-25 as it will not break any existing solutions that I am aware of. However, that said, there are some other facets of support for "Rich Fungible Tokens" (RFTs for the purpose of CIP-25 & CIP-68) that may warrant their own, separate CIP in the future to help explorers and marketplaces navigate and smoothly handle these tokens.

rphair (approved):
OK, sounds good all around.  I do feel better about having the updated term there which correctly indicates the range of contemporary use and don't think it dulls the historical significance to change it. 😇

rphair:
As per our last meeting agreement, not merging yet pending any dispute from implementors (@SmaugPool @alessandrokonrad @mmahut ... feel free to include others who may be itnerested) in case the wording change goes against their ongoing intent in any way.  Otherwise we'll leave it `Last Check` to merge at next CIP meeting: https://hackmd.io/@cip-editors/74

SmaugPool:
> These changes are minor and largely cosmetic but serve to represent and open up this `metadata standard` for more than strictly tokens with a quantity = 1. I see no reasons or means by which this would cause any errors with existing implementations so believe it is ready to be merged in.

I agree with @Crypto2099.

Issue #590, opened 2023-09-09
CIP-30 Extension Naming


Ryun1:
Although we live in a world where CIP-30 is the default Cardano wallet-web connector, this may not always be the case, thus explicitly tying CIP-30 extensions to CIP-30 maybe a mistake.

I propose that we include these proposal titles something like "Web-Wallet Bridge", so something like #509 would be named "Governance Web-Wallet Bridge". I am bad at naming, so just wanted to create an issue to start the conversation.

Thoughts?
@rphair @KtorZ @Crypto2099

rphair:
I seconded this motion at the CIP meeting this week ✅ but would also suggest that _Web-Wallet Bridge_ or equivalent **precede** the extension (e.g. _Web-Wallet Bridge: Governance_) so they all collate together when CIP names are sorted.

Ryun1:
@rphair 

Youre always good at naming! - _Web-Wallet Bridge: xxxx_ sounds good to me 😎

rphair:
@Ryun1 if nobody thinks these new names are out of place (so far the consensus looks in favour), perhaps we can declare this issue closed in meeting # 73 so I'll put it on the agenda.

Ryun1:
@rphair I agree!

Ryun1:
The only is that the colon (:) creates a YAML error when in the proposal header

`Error in user YAML: (<unknown>): mapping values are not allowed in this context at line 2 column 25`

@rphair

rphair:
Aha.  Colon = mapping syntax.  Thanks; I've been trying to narrow that down for the longest time. 🥴

Typographically the best equivalent of a colon would be an em dash (—) but since we should stick to keyboard characters how about a minus sign?  @Ryun1 @Crypto2099 @Scitz0

> Web-Wallet Bridge - xxxx

Ryun1:
In CIP editors call 73, we decided to move to use pipe ( | ) character. PR with information to follow.

rphair:
Suggested update to CIP naming convention here (along with how we might update CIP-0030 in kind): https://github.com/cardano-foundation/CIPs/pull/594

PR #594, opened 2023-09-19
CIP-0001 | Titles for CIPs with common purpose

rphair:
Discussion begun in https://github.com/cardano-foundation/CIPs/issues/590 and currently, as agreed in today's editor meeting, with consensus so far that:
- the "pipe" (`|`) is a better choice than the colon to avoid YAML errors
- this may also be applicable to other CIP names: e.g. may be helpful for new token types as per CIP-0068, so should be generalised into CIP-0001... as opposed to CIP-0030 which would only cover extensions to the web-wallet bridge.

On that last note we also suggested @Ryun1 might add a similar instruction to CIP-0030 that those extensions be named as such (with `Web-Wallet Bridge |`), once it's confirmed here.  In anticipation of this I've added an H6 (so far unused) to the document so it can be referenced easily with an external link.

BTW the merged CIP-0095 itself now follows this convention (`Web-Wallet Bridge | Governance`): https://github.com/cardano-foundation/CIPs/tree/master/CIP-0095

### DIFF
--- a/CIP-0001/README.md
+++ b/CIP-0001/README.md
@@ -93,7 +93,7 @@ Each CIP must begin with a YAML key:value style header preamble (also known as _
 Field          | Description
 ---            | ---
 `CIP`          | The CIP number (without leading 0), or "\?" before being assigned
-`Title`        | A succinct and descriptive title.  If necessary, use a <code>&#124;</code> delimiter to begin with an applicable classification (see [Naming CIPs with similar subjects](#naming-cips-with-similar-subjects)).
+`Title`        | A succinct and descriptive title.  If necessary, use a `-` delimiter to begin with an applicable classification (see [Naming CIPs with similar subjects](#naming-cips-with-similar-subjects)).
 `Status`       | Proposed \| Active \| Inactive (.._reason_..)
 `Category`     | One of the registered [categories](#categories) covering one area of the ecosystem.
 `Authors`      | A list of authors' real names and email addresses (e.g. John Doe <john.doe@email.domain>)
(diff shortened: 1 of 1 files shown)


rphair:
FYI, hold on review... there is a problem using the `|` character in a table, which my happy-go-lucky markdown editor rendered properly... though it's considered a further table column on GitHub (and therefore disappers).  Will push out another commit (or more) based on some testing & syntax changes.

rphair:
(p.s. https://github.com/cardano-foundation/CIPs/pull/594#issuecomment-1726426295) OK, I think it's fixed now, using HTML `<code>` syntax + the DTD for HTML files which should avoid the problems pointed out in https://github.com/cardano-foundation/CIPs/pull/591 (i.e. the "easy fix" is escaping with `\` but that wouldn't be portable).

rphair:
Note: discussion is now open on reverting to the second-most-popular suggestion of using the `-` sign as a delimiter (suggested in https://github.com/cardano-foundation/CIPs/issues/590#issuecomment-1725919385 & https://github.com/cardano-foundation/CIPs/pull/595#discussion_r1330762739) especially because the extra markup required to put `|` in a table is bound to cause even worse problems than the YAML-unfriendly `:`. 😰

rphair:
Reverted back to `-` for a delimiter as per reported problem & consensus in https://github.com/cardano-foundation/CIPs/pull/595#discussion_r1330762739.

Crypto2099 (approved):

Ryun1 (approved):

PR #595, opened 2023-09-19
Update top-level README: meeting #73 promotion

rphair:
apologies for making a separate PR for this: I was going to sneak it in before merging https://github.com/cardano-foundation/CIPs/pull/509 but thought we had other name changes to make as per https://github.com/cardano-foundation/CIPs/pull/594 (these were only PRs & not CIP candidates). 😖

### DIFF
--- a/README.md
+++ b/README.md
@@ -89,7 +89,7 @@ CIP Editors meetings are public, recorded, and [published on Youtube](https://ww
 | 0086 | [NFT Metadata Update Oracles](./CIP-0086) | Proposed |
 | 0093 | [Authenticated Web3 HTTP requests](./CIP-0093) | Proposed |
 | 0094 | [SPO On-chain Polls](./CIP-0094) | Active |
-| 0095 | [Web-Wallet Bridge | Governance](./CIP-0095) | Proposed |
+| 0095 | [Web-Wallet Bridge - Governance](./CIP-0095) | Proposed |
 | 0101 | [Integration of keccak256 into Plutus](./CIP-0101) | Proposed |
 | 0381 | [Plutus Support for Pairings Over BLS12-381](./CIP-0381) | Proposed |
 | 1694 | [A proposal for entering the Voltaire phase](./CIP-1694) | Proposed |
--- a/README.md
+++ b/README.md
@@ -89,6 +89,7 @@ CIP Editors meetings are public, recorded, and [published on Youtube](https://ww
 | 0086 | [NFT Metadata Update Oracles](./CIP-0086) | Proposed |
 | 0093 | [Authenticated Web3 HTTP requests](./CIP-0093) | Proposed |
 | 0094 | [SPO On-chain Polls](./CIP-0094) | Active |
+| 0095 | [Web-Wallet Bridge | Governance](./CIP-0095) | Proposed |
 | 0101 | [Integration of keccak256 into Plutus](./CIP-0101) | Proposed |
 | 0381 | [Plutus Support for Pairings Over BLS12-381](./CIP-0381) | Proposed |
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -98,7 +99,7 @@ CIP Editors meetings are public, recorded, and [published on Youtube](https://ww
 | 1855 | [Forging policy keys for HD Wallets](./CIP-1855/) | Active |
 | 9999 | [Cardano Problem Statements](./CIP-9999/) | Active |
 
-<p align="right"><i>Last updated on 2023-09-08</i></p>
+<p align="right"><i>Last updated on 2023-09-19</i></p>
```
Ryun1, line 102:
```suggestion
```
Do we even need this date? I always thought its not needed considering the lower date

Ryun1 (approved):

README.md
```diff
@@ -89,6 +89,7 @@ CIP Editors meetings are public, recorded, and [published on Youtube](https://ww
 | 0086 | [NFT Metadata Update Oracles](./CIP-0086) | Proposed |
 | 0093 | [Authenticated Web3 HTTP requests](./CIP-0093) | Proposed |
 | 0094 | [SPO On-chain Polls](./CIP-0094) | Active |
+| 0095 | [Web-Wallet Bridge | Governance](./CIP-0095) | Proposed |
```
Ryun1, line 92:
```suggestion
| 0095 | [Web-Wallet Bridge <code>&#124;</code> Governance](./CIP-0095) | Proposed |
```

README.md
rphair: I don't think any of these dates for lists of PRs are needed but only assuming a general solution... within easy reach I think... of generating the lists dynamically through tagging according to their precise editorial status.

I have a planned improvement as part of my Catalyst proposal to introduce such a tagging vocabulary & until such a time comes I would happily see those "last updated" dates dropped since the only one that probably matters to the general public is the list of merged CIPs (which generally aren't obtainable from PRs).

But until then I think we need to continue the tradition of putting 1 "last update" date at the footer of each 1 table.  The alternative to this (I think evidenced before @Ryun1 you arrived) was too difficult to interpret & wasn't modified consistently.

README.md
rphair: Now I am not feeling so good about the "clean" solution of putting `|` characters in our CIP titles, since they so often need to appear in tables.  This makes it look like we should return to my suggestion (https://github.com/cardano-foundation/CIPs/issues/590#issuecomment-1725919385) of using the keyboard-friendly and typographically neutral minus sign (`-`). 😵‍💫

@Ryun1 @Crypto2099 @Scitz0 what do you think of reverting our consensus today to the good old minus sign... keeping in mind the vast number of unpredictable contexts in which the CIP titles will be repurposed?

Scitz0: Agree, let's use `-` to keep it simple.

Ryun1: Im happy to revert to `-`

Crypto2099:
Didn't we settle on the vertical pipe ( | ) character for the delineator rather than the dash? This seems to have reverted to the dash separator format which IMO may be ambiguous and cause confusion if proper spacing is not maintained.

Ryun1:
@Crypto2099 the pipe unfortunately broke how markdown tables are rendered on Github

Ryun1 (approved):

Crypto2099 (approved):

rphair:
> @Crypto2099: Didn't we settle on the vertical pipe ( | ) character for the delineator rather than the dash?

yes, I still feel this would have been functionally & typographically ideal, but we'd all regret it as various artefacts and incompatibilities of `<code>&#124;</code>` started appearing everywhere in repurposed lists of CIPs 😖 https://github.com/cardano-foundation/CIPs/pull/595#discussion_r1330762739

PR #601, opened 2023-10-04
CIP-0030 | Fix API headers for markdown portability

rphair:
Recognised in https://github.com/cardano-foundation/CIPs/pull/591#issuecomment-1712917686 (cc @SmaugPool) and cleaned up here since running this code through Showdown in site builders causes text inside angle brackets beginning `\<` to be dropped unless quoted literally in Markdown (as code would be).

To avoid too many unnecessary changes, I've only applied this change to the [API section](https://github.com/cardano-foundation/CIPs/tree/master/CIP-0030#initial-api) and not the preceding error types or data types.  Perhaps for functional & aesthetic considerations other reviewers might think these literals should also be quoted as code. 🤔 cc @rooooooooob

 - NOTE data type `cbor<T>` suffers the same problem I've also quoted it as code.
 - @Ryun1 @Crypto2099 @SebastienGllmt please look at the code segment in that section where I've also removed the quote back-slashes, which also truncate the code in Showdown and don't look right to me as code in the first place (?).  If the back-slashes _are_ necessary in the code, maybe we can find a portable way of putting them back in. 😵‍💫

What we're mainly trying to achieve is the correct appearance of this page on the Dev Portal (cc @katomm), which currently has sections of code in many of the API section headers truncated (cc @KtorZ): https://developers.cardano.org/docs/governance/cardano-improvement-proposals/CIP-0030

---

([how it renders with the changes in this branch](https://github.com/rphair/CIPs/blob/cip-0030-quote-code-in-headings/CIP-0030/README.md))

### DIFF
--- a/CIP-0030/README.md
+++ b/CIP-0030/README.md
@@ -54,7 +54,7 @@ A string representing an address in either bech32 format, or hex-encoded bytes. 
 
 A hex-encoded string of the corresponding bytes.
 
-#### cbor\<T>
+#### `cbor<T>`
 
 A hex-encoded string representing [CBOR](https://tools.ietf.org/html/rfc7049) corresponding to `T` defined via [CDDL](https://tools.ietf.org/html/rfc8610) either inside of the [Shelley Multi-asset binary spec](https://github.com/input-output-hk/cardano-ledger-specs/blob/0738804155245062f05e2f355fadd1d16f04cd56/shelley-ma/shelley-ma-test/cddl-files/shelley-ma.cddl) or, if not present there, from the [CIP-0008 signing spec](https://github.com/cardano-foundation/CIPs/blob/master/CIP-0008/README.md).
 This representation was chosen when possible as it is consistent across the Cardano ecosystem and widely used by other tools, such as [cardano-serialization-lib](https://github.com/Emurgo/cardano-serialization-lib), which has support to encode every type in the binary spec as CBOR bytes.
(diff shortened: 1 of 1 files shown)


CIP-0030/README.md
```diff
@@ -54,7 +54,7 @@ A string representing an address in either bech32 format, or hex-encoded bytes.
 
 A hex-encoded string of the corresponding bytes.
 
-#### cbor\<T>
+#### `cbor<T>`
```
Ryun1, line 57:
```suggestion
#### cbor<T>

```ts
type cbor<T> = string;
```
What about something like this? more consistent with the other datatype section titles?

rphair: I wouldn't mind adding some code afterward to clarify the data type (though I already think it's clear from the paragraph text that follows)... but the `cbor<T>` needs to be quoted; otherwise it just appears as `cbor`, with Showdown dropping the angle bracketed text, as you can see at this section header: https://developers.cardano.org/docs/governance/cardano-improvement-proposals/cip-0030/#cbor

Ryun1: Ahh I see, lets keep it simple then, what youve suggested with `cbor<T>` makes sense.

Ryun1 (approved):

rooooooooob (approved):
LGTM. It would be nice if we could make them bigger while being code, but judging from CIP-95 it looks good still. And consistent with other CIP(s?) that had this issue. The API is pretty unhelpful as-is on the cardano site so you need to reference the github code, but when searching it's the cardano site that pops up and not the github one so this should help a lot.

Crypto2099 (approved):
These changes look good to me and make sense as a standard we can attempt/encourage to follow going forward for other CIPs requiring code examples.

PR #605, opened 2023-10-14
CIP???? | Domain Address Resolving Standard

SIDANWhatever:
As the ecosystem emerges, more and more domain service projects entering Cardano. Currently noticeably there are 3 domain projects:
1. `ADA Handle`
2. `adadomains`
3. `Cardano Name Service`

Both `adadomains` and `Cardano Name Service` has chosen `.ada` as their domain suffix. When proceeding with wallet integration on address resolving, one common issue faced is the ambiguity in resolving approach, which could lead to potentially serious loss of fund in sending fund to undesired recipients.

The CIP is proposed accordingly to address the community feedback and suggestion.

### DIFF
--- a/CIP-?/README.md
+++ b/CIP-?/README.md
@@ -1,0 +1,99 @@
+---
+CIP: [To be assigned]
+Title: Domain Address Resolving Standard
+Status: Proposed
+Category: Wallets
+Authors:
+  - Hinson Wong <hinson@cns.space>
+Implementors: []
+Discussions: https://github.com/cns-space/CIPs
+Solution-To: N/A
+Created: 2023-10-14
+License: CC-BY-4.0
+---
+
+## Abstract
+
+This proposal suggests a method for Cardano wallets to account for naming conflicts that arise when multiple projects use the same prefix or suffix, such as `$` and `.ada,` for domain names. The proposed solution allows users to choose their desired project when resolving names with colliding prefixes or suffixes.
+
+## Motivation: why is this CIP necessary?
+
+As different domain projects emerge, the same prefix or suffix may be employed by different projects, leading to potential naming conflicts. One of the key features of blockchain domain service is to resolve a holder's address. Conflicting names would lead to the wallet's integration ambiguity in resolving. To address this, a community-aligned mechanism should be in place to enable users to select their preferred project when resolving names. This ensures a seamless user experience.
+
+## Specification
+
+1. **Naming Conflict Detection:**
+
+   - When a user enters a domain name (e.g., `hinson.ada`) with the same prefix or suffix, the wallet should detect the potential conflict. Possibly by attempt to resolve all potential project's corresponding addresses.
+   - When a user enters a domain name (e.g., `hinson.ada`) with the same prefix or suffix, the wallet should detect the potential conflict by attempting to resolve all potential project's corresponding addresses.
+
+2. **User Prompt:**
+
+   - If there is only one corresponding address resolved, the wallet could proceed with the only address for proceeding with the user transaction.
+   - If more than one corresponding address is detected, the wallet should display a prompt to the user, explaining the conflict, and providing options for resolution.
+
+3. **Resolution Options:**
+
+   - The prompt should offer the user the following options:
+
+     - Resolve with Project A
+     - Resolve with Project B
+     - etc...
+     - Cancel
+
+     > Detail UX implementation to be decided by the wallet as long as users are not confused with their choice, so only showing the project name is the minimal information needed, e.g. one for Cardano Name Service and another for adadomain
+
+4. **User Selection:**
+
+   - The user selects one of the options presented in the prompt. If they choose to resolve with Project A, the Project A's resolver is used; if they choose the other project, the respective resolver is used.
+
+5. **Default Project:**
+
+   - Users may have the option to set a default project for names with a colliding suffix in their wallet settings. This default project will be used for resolution unless changed by the user during the resolution prompt.
+   - Without user specification, there should always be an option prompting to user in case of domain name conflict.
+
+## Rationale: how does this CIP achieve its goals?
+
+This proposal embraces the decentralization property of a blockchain where it welcomes multiple domain name service that exists in the market without user experience competition at the infrastructure level. It ensures a user-friendly experience and allows users to make informed decisions in case of naming conflicts. It accommodates multiple projects with the same suffix in wallets while avoiding disputes and confusion, fostering the integration process of wallet and domain service.
+
+### Path to Active
+
+#### Acceptance Criteria
+
+- At least 3 of wallets listed below agree with the approach
+
+  - Begin <https://begin.is/>
+  - Eternl <https://eternl.io/>
+  - Flint <https://flint-wallet.com/>
+  - GeroWallet <https://www.gerowallet.io/>
+  - Lace <https://www.lace.io/>
+  - Nami <https://namiwallet.io/>
+  - NuFi <https://nu.fi/>
+  - RayWallet <https://raywallet.io/>
+  - Yoroi <https://yoroi-wallet.com/>
+
+  > Initial list of wallet adapted from CIP30, other wallet providers could to request to add to the list.
+
+#### Implementation Plan
+
+- Every participating Cardano domain service provider provides an address resolver SDK.
+- Every participating Cardano domain service provider provides either a desired prefix or suffix.
+- Wallet providers to execute and integrate with resolving address, domain service project to provide assistance.
+
+#### Participating Cardano Domain Service
+
+| Domain Service Project     | Prefix | Suffix | Link to Resolver SDK Repo                     |
+| -------------------------- | ------ | ------ | --------------------------------------------- |
+| Cardano Name Service (CNS) | N/A    | `.ada` | https://github.com/cns-space/cns-resolver-sdk |
+|                            |        |        |                                               |
+|                            |        |        |                                               |
+
+## CIP Extendibility: Domain Information Resolving
+
+This approach could be applied in resolving other information as well in case of conflict. When there is no other specific CIP covering the particular domain information resolving mechanism, the similar approach with this CIP would by default covering the particular scope.
+
+## Copyright
+
+This CIP is licensed under [CC-BY-4.0].
+
+[CC-BY-4.0]: https://creativecommons.org/licenses/by/4.0/legalcode


CIP-?/README.md
```diff
@@ -0,0 +1,99 @@
+---
+CIP: [To be assigned]
```
rphair, line 2:
```suggestion
CIP: ?
```
Specific format required by CIP-0001 and the site generators that parse CIP text.

CIP-?/README.md
```diff
@@ -0,0 +1,99 @@
...
+Title: Domain Address Resolving Standard
+Status: Proposed
+Category: Wallets
+Authors:
+  - Hinson Wong <hinson@cns.space>
+Implementors: []
+Discussions: https://github.com/cns-space/CIPs
+Solution-To: N/A
```
rphair, line 10:
```suggestion
```
This field is only required when there's a matching Cardano Problem Statement (CPS).

CIP-?/README.md
```diff
@@ -0,0 +1,99 @@
...
+5. **Default Project:**
+
+   - Users may have the option to set a default project for names with a colliding suffix in their wallet settings. This default project will be used for resolution unless changed by the user during the resolution prompt.
+   - Without user specification, there should always be an option prompting to user in case of domain name conflict.
+
+## Rationale: how does this CIP achieve its goals?
+
+This proposal embraces the decentralization property of a blockchain where it welcomes multiple domain name service that exists in the market without user experience competition at the infrastructure level. It ensures a user-friendly experience and allows users to make informed decisions in case of naming conflicts. It accommodates multiple projects with the same suffix in wallets while avoiding disputes and confusion, fostering the integration process of wallet and domain service.
+
+### Path to Active
```
rphair, line 59:
```suggestion
## Path to Active
```
Indented improperly (as submitted it looks like it's part of the Rationale section, which would be incorrect)

CIP-?/README.md
```diff
@@ -0,0 +1,99 @@
...
+
+#### Acceptance Criteria
```
rphair, line 61:
```suggestion
### Acceptance Criteria
```

CIP-?/README.md
```diff
@@ -0,0 +1,99 @@
...
+  - GeroWallet <https://www.gerowallet.io/>
+  - Lace <https://www.lace.io/>
+  - Nami <https://namiwallet.io/>
+  - NuFi <https://nu.fi/>
+  - RayWallet <https://raywallet.io/>
+  - Yoroi <https://yoroi-wallet.com/>
+
+  > Initial list of wallet adapted from CIP30, other wallet providers could to request to add to the list.
+
+#### Implementation Plan
```
rphair, line 77:
```suggestion
### Implementation Plan
```

CIP-?/README.md
```diff
@@ -0,0 +1,99 @@
...
+
+- Every participating Cardano domain service provider provides an address resolver SDK.
+- Every participating Cardano domain service provider provides either a desired prefix or suffix.
+- Wallet providers to execute and integrate with resolving address, domain service project to provide assistance.
+
+#### Participating Cardano Domain Service
```
rphair, line 83: One way or another this item should be removed, since (bureaucratically) it's not a standard part of Path to Active:

- Based on the branch name I am assuming this is your homegrown solution: if so you can cite it as an implementation & it would address my original reservation (about not being specific enough) if you could talk in your Specification about how your own system works.  If considering it separately from the CIP, you might put it in Rationale to talk about how &^ why the CIP is designed to accommodate implementations like your own.
- Regardless of how much detail you go into, it should instead be an item in your Implementation Plan since you appear to have a working model (or plan to have one).

CIP-?/README.md
```diff
@@ -0,0 +1,99 @@
...
+
+This proposal embraces the decentralization property of a blockchain where it welcomes multiple domain name service that exists in the market without user experience competition at the infrastructure level. It ensures a user-friendly experience and allows users to make informed decisions in case of naming conflicts. It accommodates multiple projects with the same suffix in wallets while avoiding disputes and confusion, fostering the integration process of wallet and domain service.
+
+### Path to Active
+
+#### Acceptance Criteria
+
+- At least 3 of wallets listed below agree with the approach
+
+  - Begin <https://begin.is/>
```
rphair, line 65:
```suggestion
  - [ ] Begin <https://begin.is/>
```
It would help everyone (including your potential implementors) to have these items also as tick-boxes for when you get buy-in from the wallets (with a parent tick-box when you've got >= 3 of them), if you continue to follow this criterion.

CIP-?/README.md
```diff
@@ -0,0 +1,99 @@
...
+| Domain Service Project     | Prefix | Suffix | Link to Resolver SDK Repo                     |
+| -------------------------- | ------ | ------ | --------------------------------------------- |
+| Cardano Name Service (CNS) | N/A    | `.ada` | https://github.com/cns-space/cns-resolver-sdk |
+|                            |        |        |                                               |
+|                            |        |        |                                               |
+
+## CIP Extendibility: Domain Information Resolving
+
+This approach could be applied in resolving other information as well in case of conflict. When there is no other specific CIP covering the particular domain information resolving mechanism, the similar approach with this CIP would by default covering the particular scope.
+
```
rphair, line 94: There is no such standard section here.  I believe your "approach" belongs Rationale because it explains why you've chosen to define the CIP so it can relate well with other CIPs coming along in the future (?), so it could be moved there.

rphair (commented):
I don't know if this CIP is specific enough to be usable by wallets (I'm not saying it's not).  At first reading it seems to set a general behavioural guideline which seems like common sense & therefore any wallet (all those mentioned in _Path to Active_) would probably "agree with the approach" as you say there... but we'd never know for sure, and neither would they, unless your approach were detailed more exactly.

Maybe if a more specific case were illustrated about how exactly a wallet is expected to behave under these circumstances... something that's not being done already... would help understand how necessary this CIP might be.  (Note I don't work directly with these implementations & we have a couple other editors @Ryun1 @Crypto2099 who work extensively with wallets & their standards, and so they might be able to see this from what you have written.)

CIP-?/README.md
SIDANWhatever: So I just need to remove the line `#### Participating Cardano Domain Service` right? I think the content of the session is necessary to stay for community to take reference (and also its part of the implementation).

rphair: Just please make sure that whatever content remains fits into the section where it appears to be.  If it's in either of the two Path to Active sections, it should fit that section as described here: https://github.com/cardano-foundation/CIPs/tree/master/CIP-0001#path-to-active

SIDANWhatever:
> @rphair 
> I don't know if this CIP is specific enough to be usable by wallets (I'm not saying it's not). At first reading it seems to set a general behavioural guideline which seems like common sense & therefore any wallet (all those mentioned in _Path to Active_) would probably "agree with the approach" as you say there... but we'd never know for sure, and neither would they, unless your approach were detailed more exactly.
>
> A more specific example of how exactly a wallet is expected to behave under these circumstances... something that's not being done already... would help show how necessary this CIP might be. (Note I don't work directly with these implementations & we have a couple other editors @Ryun1 @Crypto2099 who work extensively with wallets & their standards, and so they might be able to see this from what you have written.)

Just added some specific examples on the expected behaviour. Does it address your concern, or is it better for me to try to pull wallets include discussion here as well?

Ryun1:
I agree with @rphair, such wallet behaviours are not normally under the remit of CIPs. I believe this proposal to be more _best practices_ rather than a technical problem and a solution. But I would be keen to hear from Wallet providers on this, perhaps this is something that they would be interested in pursuing.

rphair:
OK then @SIDANWhatever @Ryun1 - we can talk about both issues (CIP proper scope & how to get wallet devs into the discussion) on Tuesday; I've added it to the agenda: https://hackmd.io/@cip-editors/75

SIDANWhatever:
As per discussion in bi-weekly call, amended the proposal with format of CPS. Appreciate any suggestion on how could I move contents into suitable session (especially here I have a proposed solution or standard to this particular CPS).

CIP-?/README.md
```diff
@@ -0,0 +1,123 @@
+---
+CPS: ?
+Title: Domain Address Resolving Standard
+Status: Proposed
+Category: Wallets
+Authors:
+  - Hinson Wong <hinson@cns.space>
+Proposed Solution: []
+Discussions: https://github.com/cns-space/CIPs
```
Ryun1, line 9:
```suggestion
Discussions: 
- https://github.com/cns-space/CIPs
- https://github.com/cardano-foundation/CIPs/pull/605
```

rphair:
thanks @SIDANWhatever - this definitely cleans up the difficult criteria of "3 of wallets listed below agree with the approach" since that should occur in discussion of the CPS and any other CIPs submitted for it.

re https://github.com/cardano-foundation/CIPs/pull/605#issuecomment-1766808194 - All the content about your own reference implementation can be posted into a separate CIP document with a separate pull request... since these 2 documents will be considered & reviewed independently.

SIDANWhatever:
@rphair Redrafting it into a CPS, I would create another PR for my suggested solution as CIP particularly to the address resolving (since mis-alignment on it might cause loss of fund, it comes with a higher priority).

One thing I would like to hear editors' advice is that, I amended this CPS to address a higher level problem - information resolving as a whole but not address resolving. In this sense, I am not sure which category should I put on this CPS.

I have invited collaboration on the forked repo. Please feel free to edit and comment. Thanks!

CPS-?/README.md
```diff
@@ -0,0 +1,75 @@
+---
+CPS: ?
+Title: Domain Information Resolving Standard
```
rphair, line 3:
```suggestion
Title: Domain Name Resolution
```
This has become a problem definition rather than a `Standard``... although one or more of the emergent CIPs may eventually come to be seen as standards.

If there is other "Information" besides the domain _name_ that needs to be resolved, then I would encourage further correction from the author & reviewers. 😎

CPS-?/README.md
```diff
@@ -0,0 +1,75 @@
...
+Status: Proposed
+Category: Wallets
```
rphair, line 5:
```suggestion
Category: Tools
```
Based on your observation & a general view I'd re-cast this as `Tools` because it addresses dApp and UX issues as well as `Wallets`.  From [CIP-0001 > Categories](https://github.com/cardano-foundation/CIPs/blob/master/CIP-0001/README.md#categories):
> Tools | A broad category for ecosystem tools not falling into any other category.

rphair (commented):
@SIDANWhatever yes a CIP for the specific (perhaps more urgent) solution sounds like it would be welcome.

CPS-?/README.md
```diff
@@ -0,0 +1,75 @@
+---
+CPS: ?
```
rphair, line 2:
```suggestion
CPS: 8
```

rphair (commented):
Promoted to CPS candidate at today's meeting; thanks @SIDANWhatever for your attendance & progressive work... please change this directory name to `CPS-0008` and update the "rendered proposal" link at the top 🤩

@colll78 in light of https://github.com/cardano-foundation/CIPs/pull/605#issuecomment-1810705340 if you are behind https://github.com/Anastasia-Labs/smart-handles then please grace us this PR with a review and/or feedback whenever you see fit 🤓

rphair (approved):
@HinsonSIDAN I can't see that this problem (as stated in your CPS) has gone away, and in fact I've referred to your exposition of the problem a few times when discussing the new work on Cardano Payment URLs which I believe will require some means of settling domain name conflicts (cc @Crypto2099).

So although it's been somewhat "abandoned" (currently we're finding many such proposals in the CIP repo PR queue) I'm going to vote that this should be merged and therefore tag `Last Check` which which will trigger its appearance on a CIP meeting soon: maybe the one tomorrow.

As far as I'm concerned this passed review a long time ago (and somehow slipped through cracks in editors' attention) & no further action is required on your part, but if there is anything you need to add please update it ASAP.

Ryun1 (approved):
This one slipped through the cracks
Happy to see it merged

PR #607, opened 2023-10-19
Fix link for BIP32-Ed25519 reference paper

danivideda:


rphair (approved):
Looks good & thanks @danivideda.  Other editors: I did check for the "proper" IEEE version of the paper but didn't find anything outside a pay- or account-wall... so it does make sense to still use the Adrestia version.

PR #608, opened 2023-10-20
CIP-???? | Deterministic universal almost-unique Plutus Constructors

nielstron:
OpShin has recently gone the path of redefining the way it uses constructor ids internally to distinguish types. Since the implementation is language agnostic, we provide the specification here for other languages or applications to adopt as they see suitable.

-------------------------------------------------------------------------------------------------------------------------------------------------------------

([rendered proposal in branch](https://github.com/OpShin/CIPs/blob/constr_ids/CIP-%3F%3F%3F%3F/README.md))

### DIFF
--- a/CIP-????/README.md
+++ b/CIP-????/README.md
@@ -28,7 +28,9 @@ License: CC-BY-4.0
 Note: I will use record / Plutus Data exchangibly throughout the document.
 
 ## Abstract
-Plutus Constructor IDs are currently heavily focused around their origin in Haskell. They are usually used to distinguish different constructors of a single declared datatype.
+Plutus Data on Cardano represent product types and are usually identified on chain by a combination of a constructor ID and their fields, 
+i.e. for `type OptionalInt = None | Some (x:Integer)` we would get Plutus Data representing `None` with constructor ID 0 and no fields as well as Plutus Data representing `Some n` with constructor ID 1 and a single field with value `n` (an integer).
+The existence and declaration of these constructor IDs are currently heavily focused around their origin in Haskell. They are usually used to distinguish different constructors of a single declared datatype.
 In contrast, one may introduce universally recognized datatypes that are identified by a unique constructor id and can be expected to behave in a specified way (i.e. contain specific fields with specific types).
 For this purpose, we introduce a generic way to compute an almost unique, deterministic and universal constructor id for objects based on their name and field types.
 Note that it is not expected that every language adopts this standard as a default (i.e. for Haskell-like languages there might not be much use of it).
@@ -86,8 +88,8 @@ It must also explain how the proposal affects the backward compatibility of exis
 We definetly want a few properties on the CONSTR_IDs
 
 - _small_: ideally the constr_id integer should be as small as possible, as smaller integers are encoded more efficiently in CBOR and save the end user minutxo and txfees (constr_ids are encoded as the cbor tag up to 7 bit size, after that encoded as generic integer)
-- _unique_: There should be as little overlap with other values as possible, so that we can group together classes in unions without having to worry about setting/overwriting the constr id. This is reflected by the unique choice of identifiers in `ustr`.
-- _deterministic_: Datatypes that are defined in libraries may be imported in arbitrary contexts. the constr_id must therefore not depend on i.e. what other Unions the datatype is being used in or what other datatypes are declared in its surroundings. This rules out the Haskell approach and any automatically incrementing global counters.
+- _unique_: There should be as little overlap with other values as possible, so that we can group together classes in unions without having to worry about setting/overwriting the constr id. This is reflected by the unique choice of identifiers in `ustr` and rules out the traditional Haskell approach.
+- _deterministic_: Datatypes that are defined in libraries may be imported in arbitrary contexts. the constr_id must therefore not depend on i.e. what other Unions the datatype is being used in or what other datatypes are declared in its surroundings. This rules out any automatically incrementing global counters.
 
 Note that the implementation first computes a `ustr` in human readable form and then transforms it into an integer. This is intentional, since the alternatives (directly computing a large unique number or similar approaches) are much more difficult to debug.
 
--- a/CIP-????/README.md
+++ b/CIP-????/README.md
@@ -7,7 +7,7 @@ Authors:
     - Niels Mündler <niels@opshin.dev>
 Implementors: [Niels Mündler <niels@opshin.dev>]
 Discussions:
-    - https://github.com/cardano-foundation/cips/pulls/?
+    - https://github.com/cardano-foundation/CIPs/pull/608
 Created: 2023-10-20
 License: CC-BY-4.0
 ---
--- a/CIP-????/README.md
+++ b/CIP-????/README.md
@@ -93,7 +93,7 @@ We definetly want a few properties on the CONSTR_IDs
 
 Note that the implementation first computes a `ustr` in human readable form and then transforms it into an integer. This is intentional, since the alternatives (directly computing a large unique number or similar approaches) are much more difficult to debug.
 
-To ensure that this does not only take the structural definition but also the intended usage into account, names of records are taken into account for the computation.
+To ensure that there is no accidental overlap in constructor ids due to having same fields in different applications (i.e. `NegInfPosixTime` and `PosInfPosixTime` without any fields), names of records are taken into account for the computation of the constructor id.
 
 There is no issue with backwards compatability when adopting this implementation as an opt-in choice for users.
 PlutusTx and most other languages allow explicitly setting the constructor id of objects anyways.
--- a/CIP-????/README.md
+++ b/CIP-????/README.md
@@ -57,17 +57,26 @@ The following function describes how to compute `ustr(X)` for a type recursively
 ```
 ustr(bytes) := "bytes"
 ustr(integer) := "int"
-ustr(PlutusData) := "any"  // This covers the case where the structure of the object is now known from the perspective of the class, i.e. when any BuiltinData is allowed
-ustr(list) := "list"  // This covers the case where the type of the elements in the list are not known in advance
+// This covers the case where the structure of the object is now known from the perspective of the class, i.e. when any BuiltinData is allowed
+ustr(PlutusData) := "any"
+  // This covers the case where the type of the elements in the list are not known in advance
+ustr(list) := "list"
+
 ustr(list<X>) := "list<" + ustr(X) + ">
+
 ustr(map<X,Y>) := "map<" + ustr(X) + "," + ustr(Y) + ">"
+
 ustr(union<X,Y,...,Z>) := "union<" + ustr(X) + "," + ustr(Y) + "," + ... + "," + ustr(Y) + ">"
-ustr(constr(name)<id, fields[f1:X,f2:Y,...,fn:Z]>) := "cons[" + name + "](" + str(id) + ";" + f1 + ":" + ustr(X) + "," + f2 + ":" + ustr(Y) + "," + ... + "," + fn + ":" + ustr(Z) + ")"
+
+ustr(constr(name)<id, fields[f1:X,f2:Y,...,fn:Z]>) :=
+    "cons[" + name + "](" + str(id) + ";"
+    + f1 + ":" + ustr(X) + "," + f2 + ":" + ustr(Y) + "," + ... + "," + fn + ":" + ustr(Z) +
+    ")"
 ```
 
 Where `name` and `f1` to `fn` refer to the name of the record and the names of its fields respectively.
 Since the constructor id of a records is not known when computing its constructor id, the constructor id string is set to `_` for this computation.
-As an example, the constructor id of record `A` with fields `b` (record `B`, constructor id 5 with one integer field `i`) and `c` (integer) would result in `ustr(A) = ustr(constr(A)<_,fields[b:B,c:integer]>) = "cons[A](_;b:" + ustr(constr(B)<5,fields[i:integer]>) + ",c:int)" = "cons[A](_;b:cons[B](i:int),c:int)".
+As an example, the constructor id of record `A` with fields `b` (record `B`, constructor id 5 with one integer field `i`) and `c` (integer) would result in `ustr(A) = ustr(constr(A)<_,fields[b:B,c:integer]>) = "cons[A](_;b:" + ustr(constr(B)<5,fields[i:integer]>) + ",c:int)" = "cons[A](_;b:cons[B](i:int),c:int)"`.
 
 ## Rationale: how does this CIP achieve its goals?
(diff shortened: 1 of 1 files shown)


CIP-????/README.md
```diff
@@ -0,0 +1,113 @@
+---
+CIP: ?
+Title: Deterministic universal almost-unique Plutus Constructors
+Category: Plutus
+Status: Proposed
+Authors:
+    - Niels Mündler <niels@opshin.dev>
+Implementors: [Niels Mündler <niels@opshin.dev>]
```
rphair, line 8:
```suggestion
Implementors:
    - Niels Mündler <niels@opshin.dev>
```
What you submitted may be correct YAML but CIP-0001 indicates 1 entry per line (for consistency, readability & parseability).

But @nielstron I wouldn't commit this yet since I don't think our current process can list your as an implementor.  I would happily stand corrected on this (@KtorZ @michaelpj)... especially since I don't work directly with Plutus... but regardless of its implementation in something like OpShin, wouldn't this only be "Implemented" when added to Plutus Core (something you can't do yourself)?

rphair (commented):
Thanks @nielstron - seems detailed & well presented enough to be technically considerable.

~First can you please review these guidelines for `Plutus` proposals - since they have to adhere to [some process considerations](https://github.com/cardano-foundation/CIPs/tree/master/CIP-0035#processes) to facilitate Plutus acceptance & implementation?  To begin with, there are a few items missing from _Path to Active_~ (**edit:** confirmed no core changes)

~Meanwhile~ maybe @michaelpj @L-as @KtorZ can review the idea itself, plus anyone else they can tag from the Plutus teams (I may not be current on this; it's been a while since someone submitted a Plutus CIP).

CIP-????/README.md
nielstron: This proposal does not require any changes or implementation to Plutus Core. It is more of a statement on how to handle situations that arise when dealing with different implementations or usages of the same contract across different languages.

rphair: excellent @nielstron, thanks 😎

rphair:
@nielstron I've put this on the agenda for initial review at next CIP meeting: https://hackmd.io/@cip-editors/76

CIP-????/README.md
```diff
@@ -0,0 +1,113 @@
...
+
+-->
+
+Note: I will use record / Plutus Data exchangibly throughout the document.
+
+## Abstract
+Plutus Constructor IDs are currently heavily focused around their origin in Haskell. They are usually used to distinguish different constructors of a single declared datatype.
+In contrast, one may introduce universally recognized datatypes that are identified by a unique constructor id and can be expected to behave in a specified way (i.e. contain specific fields with specific types).
+For this purpose, we introduce a generic way to compute an almost unique, deterministic and universal constructor id for objects based on their name and field types.
+Note that it is not expected that every language adopts this standard as a default (i.e. for Haskell-like languages there might not be much use of it).
```
effectfully, line 34: How is Haskell different here?

CIP-????/README.md
```diff
@@ -0,0 +1,113 @@
...
+This is currently implemented manually, which is tedious and a potential source of errors.
+
+## Specification
+<!-- The technical specification should describe the proposed improvement in sufficient technical detail. In particular, it should provide enough information that an implementation can be performed solely on the basis of the design in the CIP. This is necessary to facilitate multiple, interoperable implementations. This must include how the CIP should be versioned. If a proposal defines structure of on-chain data it must include a CDDL schema in it's specification.-->
+The deterministic, universal and almost-unique Plutus constructors are computed recursively based on the type definition of a record.
+We first compute a string `ustr(X)` based on the type definition of X. Then we perform a sha256 hash on the UTF8 encoding of this string and interpret the resulting hex digest as a big endian encoded integer.
+The integer is taken modulo 2^32. The resulting integer is the almost-unique, universal, deterministic constructor id of the plutus datum.
```
effectfully, line 53: Any discussion on what happens in case of collision?

CIP-????/README.md
```diff
@@ -0,0 +1,113 @@
...
+- Plutus   | Changes or additions to Plutus
+- Ledger   | For proposals regarding the Cardano ledger (including Reward Sharing Schemes)
+- Catalyst | For proposals affecting Project Catalyst / the Jörmungandr project
+
+-->
+
+Note: I will use record / Plutus Data exchangibly throughout the document.
+
+## Abstract
+Plutus Constructor IDs are currently heavily focused around their origin in Haskell. They are usually used to distinguish different constructors of a single declared datatype.
```
effectfully, line 31: This section lacks an elaborate example, like an actual `Data` object. It took me a while to figure out what you mean by a constructor ID and I'm a Plutus developer.

CIP-????/README.md
```diff
@@ -0,0 +1,113 @@
...
+- _deterministic_: Datatypes that are defined in libraries may be imported in arbitrary contexts. the constr_id must therefore not depend on i.e. what other Unions the datatype is being used in or what other datatypes are declared in its surroundings. This rules out the Haskell approach and any automatically incrementing global counters.
+
+Note that the implementation first computes a `ustr` in human readable form and then transforms it into an integer. This is intentional, since the alternatives (directly computing a large unique number or similar approaches) are much more difficult to debug.
+
+To ensure that this does not only take the structural definition but also the intended usage into account, names of records are taken into account for the computation.
+
+There is no issue with backwards compatability when adopting this implementation as an opt-in choice for users.
+PlutusTx and most other languages allow explicitly setting the constructor id of objects anyways.
```
effectfully, line 97: Yes, but I feel like we've always viewed constructors ids as constructor _indices_. We're discussing the possibility of converting `Data` objects to SOPs via a builtin and this can only work if constructor ids are interpreted as indices. I'll ask the team about the perspective that you bring, it's certainly new to me.

effectfully (commented):
This subverts my entire intuition on what a `Data` object means. Which is fine, evidently my intuition wasn't covering some of the ways people use `Data` objects. I'll bring this CIP to the team and get back to you once we've discussed it nternally. Thank you for submitting it!

CIP-????/README.md
michaelpj: There is nothing that necessitates any particular interpretation of the integers in a `Constr`. We have generally assumed they will be indices: in particular, favouring small numbers is a reflection of that (also see https://www.ietf.org/id/draft-bormann-cbor-notable-tags-09.html#name-enumerated-alternative-data).

The point about conversion to SOPs is a good one. If we are able to offer a fast conversion from `Data` to SOPs, then you really will want to use indices rather than arbitrary ids (since if you want to case analyze the constructor with tag n, you need to provide alternatives for all of the n-1 previous tags too!).

michaelpj (commented):
This seems fine, if it's what you want. It doesn't seem massively appealing to me, but that's okay if other people want to use it. Naming data types and constructors via their transitive structural hash a la [unison](https://www.unison-lang.org/) isn't a crazy idea.

I think the proposal could do with discussing what (to me) is the obvious alternative: wrapping. In the Haskell world, if I want to have either a type T or a type U, I write
```
data TorU = ItsT T | ItsU U
```
This corresponds to adding another layer of constructor tagging to tell me which one I'm looking at, so there's no problem if T and U use the same tags (and indeed, the actual implementation of Haskell very much does use constructor tags in this way). This doesn't seem much worse to me than anonymous unions, and it avoids the problem of tag clashes entirely. 

So the usefulness of this proposal is limited to cases where you really want use values of various overlapping types interchangeably, and wrapping isn't acceptable. The use cases aren't really clear enough for me to say whether or not that's common.

rphair:
Adding to https://github.com/cardano-foundation/CIPs/pull/608#issuecomment-1810601563 from CIP Editors' meeting today, where it was also brought up, that @nielstron we would be interested in seeing responses to the reviews already presented before we can more fully consider this as a CIP.  Until then it does seem like it could instead be a useful "best practice" document with a good idea that might not be compelling for others to adopt.

CIP-????/README.md
```diff
@@ -0,0 +1,113 @@
...
+
+The following function describes how to compute `ustr(X)` for a type recursively.
+
+```
+ustr(bytes) := "bytes"
+ustr(integer) := "int"
+// This covers the case where the structure of the object is now known from the perspective of the class, i.e. when any BuiltinData is allowed
+ustr(PlutusData) := "any"
+  // This covers the case where the type of the elements in the list are not known in advance
+ustr(list) := "list"
```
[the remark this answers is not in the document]
effectfully: > why isn't that `list<data>`?

Hm, I had this comment as well, but either I failed to hit "comment" or GitHub lost it (it's been glitchy lately for me).

CIP-????/README.md
```diff
@@ -0,0 +1,113 @@
...
+
+ustr(union<X,Y,...,Z>) := "union<" + ustr(X) + "," + ustr(Y) + "," + ... + "," + ustr(Y) + ">"
+
+ustr(constr(name)<id, fields[f1:X,f2:Y,...,fn:Z]>) :=
+    "cons[" + name + "](" + str(id) + ";"
+    + f1 + ":" + ustr(X) + "," + f2 + ":" + ustr(Y) + "," + ... + "," + fn + ":" + ustr(Z) +
+    ")"
+```
+
+Where `name` and `f1` to `fn` refer to the name of the record and the names of its fields respectively.
```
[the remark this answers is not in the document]
effectfully: > Commented later, but: you have to have either the name or the id of the constructor, otherwise you can't distinguish two constructors with the same fields. And the whole point of this proposal is to not set the ids, so it has to be names.

I mean, you can get the hash of a constructor from the structure of the data type (`\a -> Either a ()` for both the example types) and the id of the constructor.  You don't need to globally specify the id this way, just for the purpose of computing a hash (and the way hashes are computed can be arbitrary as long as they're almost unique). Or am I misunderstanding it?

CIP-????/README.md
```diff
@@ -0,0 +1,113 @@
...
+Since the constructor id of a records is not known when computing its constructor id, the constructor id string is set to `_` for this computation.
```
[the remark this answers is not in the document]
nielstron: I agree its a bit weird... but I do want to distinguish between classes with same names and fields but different constructor ids to avoid nasty suprises to the user.

class A:
  CONSTR_ID = 0

B = A

class A:
  CONSTR_ID = 1
  
class X:
   x: Union[A, B]
   
If it is pulled out of the ustr for constructors then we loose modularity of the function :thinking:

CIP-????/README.md
```diff
@@ -0,0 +1,113 @@
...
+
+## Rationale: how does this CIP achieve its goals?
+<!-- The rationale fleshes out the specification by describing what motivated the design and what led to particular design decisions. It should describe alternate designs considered and related work. The rationale should provide evidence of consensus within the community and discuss significant objections or concerns raised during the discussion.
+
+It must also explain how the proposal affects the backward compatibility of existing solutions when applicable. If the proposal responds to a CPS, the 'Rationale' section should explain how it addresses the CPS, and answer any questions that the CPS poses for potential solutions.
+-->
+We definetly want a few properties on the CONSTR_IDs
+
+- _small_: ideally the constr_id integer should be as small as possible, as smaller integers are encoded more efficiently in CBOR and save the end user minutxo and txfees (constr_ids are encoded as the cbor tag up to 7 bit size, after that encoded as generic integer)
```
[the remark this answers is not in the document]
nielstron: Types will definitely perform worse on space, I assume most (i.e. roughly 50%) of tags will have size around 2^32. Small refers to these tags being smaller than i.e. 64 bytes (like a script or datum hash). Likely doing modulo 2^64 would not make a big difference on size/cost either but improve uniqueness, so I am looking into adding this as a change.

CIP-????/README.md
nielstron: In Python it is common to declare SoP / Unions after the declaration of the specific types i.e.

```python
class A

class B

class C

AB = Union[A, B]
ABC = Union[A, B, C]
```

In Haskell the definition of the SoP/Union at the same time declares the involved alternatives, hence all involved alternatives are known at the time of declaration

```haskell
data AB = A | B
data ABC = A | B | C -- I guess this would throw an error for redeclaring the constructors A and B?
```

CIP-????/README.md
nielstron: In case of collision during declaration of Sum Type, the compiler has to deny compilation and ask the user to manually declare a constructor id for the involved types. I think this is rare enough to have practically no impact on computation. Regarding attacks I don't think this schema is more vulnerable than any other schema. In the current Plutus schema constructor id overlaps are practically omnipresent (though never in any sum types that occur in the compiled contract)

CIP-????/README.md
nielstron: This sounds like it can become a major issue for the compatability with native SOP.

CIP-????/README.md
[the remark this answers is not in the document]
nielstron: I think this is a valid point. I am currently looking towards making this be compatible with the CIP57 definitions. However CIP57 does not make reproducibility a big thing (i.e. concrete ordering of JSON map elements does usually not matter) however here it is relevant - a re-definition or at least specifictation of a "canonical" blueprint from which to hash is unavoidable.

CIP-????/README.md
```diff
@@ -0,0 +1,113 @@
...
+- _unique_: There should be as little overlap with other values as possible, so that we can group together classes in unions without having to worry about setting/overwriting the constr id. This is reflected by the unique choice of identifiers in `ustr`.
+- _deterministic_: Datatypes that are defined in libraries may be imported in arbitrary contexts. the constr_id must therefore not depend on i.e. what other Unions the datatype is being used in or what other datatypes are declared in its surroundings. This rules out the Haskell approach and any automatically incrementing global counters.
```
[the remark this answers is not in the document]
nielstron: Yes thats correct.

rphair:
@nielstron as far as I can tell, the last 2 commits https://github.com/cardano-foundation/CIPs/pull/608/commits/5d7861c246610535ec0943d42e1b9394a5366d0f & https://github.com/cardano-foundation/CIPs/pull/608/commits/1723c1bb086d069ba1bbc856089c738fd8075528 haven't addressed the feedback that I tried to summarise in https://github.com/cardano-foundation/CIPs/pull/608#issuecomment-1811720014 9 months ago.

We are tagging some proposals `Abandoned` but we never tagged this one `Waiting for Author` first (I'm trying this week to overhaul our editing process with tagging stale PR's) so I'm applying that tag now.  I expect that you can address the review points above (please make some notes in the relevant conversations) or explain why the points have all addressed.

It's just a matter of explaining your points clearly I believe... once we see that this is done we can put it back on the CIP meeting agenda to look at promoting it to a candidate; if no further progress the next few weeks it will probably be moved onto the `Abandoned` list & closed soon afterward.

rphair:
@nielstron let's make one last call for updates before closing this as "abandoned" but please if you are interested in pursuing this then respond to the last comment & we'll plug it back into the review process.

nielstron:
Hi @rphair ,
I don't have time to work on this CIP anymore, so feel free to tag it as abandoned.

PR #611, opened 2023-10-23
CPS-0009? | Coin Selection Including Native Tokens

SIDANWhatever:
This seems a long withstanding issue #232  where most offline library we got now still using sub-optimal coin selection strategy which might lead to inefficiency transaction building process.

Raising this CPS in an attempt to attract community's attraction to solve this issue with combined intelligence.

-------

Recreating the PR #609 

---

([rendered proposal in branch](https://github.com/sidan-lab/CIPs/blob/coin-selection-post-native-tokens-era/CPS-0009/README.md))

PR #617, opened 2023-11-11
CIP-0106? | Web-Wallet Bridge - Multisig wallets

leo42:
A new connector specification that will allow dApps to connect with multisig wallets. And enable defi for DAOs and security conscious users.  

The structure of this CIP follows CIP-30 as close as possible, to make integration by applications already using CIP-30 easy and streamlined.

I am requesting the number 130 for this CIP to make the affiliation with CIP-30 clear and easy for dev's to remember and search for. (planning on making Plutus-powered-wallet connector under CIP-230 in the future)  

[Rendered Version](https://github.com/leo42/CIPs/tree/master/CIP-0106/README.md)


CIP-130/README.md
```diff
@@ -0,0 +1,408 @@
+---
+CIP: ???
+Title: Cardano dApp-MutlisigWallet Web Bridge
+Status: Proposed
+Category: Wallets
+Authors: Leo
+  -
```
Crypto2099, line 7:
```suggestion
Authors: 
  - Leo
```

fix YAML formatting to avoid header errors

CIP-130/README.md
```diff
@@ -0,0 +1,408 @@
...
+## Abstract
+
+This document describes a webpage-based communication bridge allowing webpages (i.e. dApps) to interface with Cardano Multisig-wallets. It follows the specification format of CIP-30 as it attempts to make implementing this connector as easy as possible for applications that have already successfully implemented CIP-30. This document is a work in progress and is not yet finalized. It is expected to be updated as the ecosystem evolves. 
+
+
+## Motivation: why is this CIP necessary?
+
+In order to facilitate future dApp development, we will need a way for dApps to communicate with multisig wallets, given the unique complexities of native script based address special provisions need to be made to make the connector compatible with them.  
+
+Specifically apps building transactions need to be able and get the following information from the wallet:
```
Crypto2099, line 24:
```suggestion
Specifically apps building transactions need to be able to get the following information from the wallet:
```

minor grammatical change

CIP-130/README.md
```diff
@@ -0,0 +1,408 @@
...
+
+Extensions that are draft, in development, or prototyped should not use extension naming nor should they use official namspacing until assigned a CIP number. Draft extension authors are free to test their implementation endpoints by using the [Experimental API](#experimental-api). Once a CIP number is assigned implementors should move functionality out of the experimental API.
+
+##### Can extensions depend on other extensions?
+
+Yes. Extensions may have other extensions as pre-requisite. Some newer extensions may also invalidate functionality introduced by earlier extensions. There's no particular rule or constraints in that regards. Extensions are specified as CIP, and will define what it entails to enable them.
+
+##### Should extensions follow a specific format?
+
+Yes. They all are CIPs.
```
Crypto2099, line 214: Could use a bit more prose here to explain that authors wishing to extend this format should submit a new CIP and what the criteria should be for consideration by wallets to implement said extensions.

Crypto2099 (commented):
Couple of minor housekeeping suggestions here and I will expand on my thoughts in aggregate in a separate comment.

Crypto2099:
1. What is the rationale here to "replicate" the existing CIP-0030 functionality for web wallets under a new global-level namespace rather than simply creating the handful of new API calls as an extension to CIP-0030 itself? i.e.
```javascript
window.cardano.broclan.enable([130])
```
2. The idea of being able to use Script/Multisig addresses to interact with dApps is a good one, I would like to see more rationale and justification surrounding how dApps are expected to behave when encountering a multisig address... Is it up to the dApp to maintain the record of unsigned/unfinished multisig transactions? Should these naturally "decay" over time? Doesn't this present an issue of requiring dApps to be centralized rather than decentralized by placing this onus on the dApp to record state rather than putting the onus on the signers? Maybe a `importTransaction` endpoint makes sense so that members of the multisig can pass it around amongst themselves? Or the wallet provider could have a record of `pending` transactions somehow passed between members of the multisig?
3. We do not allow for "reservation" of CIP numbers, regardless of rationale or justification with only a handful of very notable exceptions. You should move this file to an `CIPXXXX` directory and await assignment of a CIP number.

leo42:
> 1. What is the rationale here to "replicate" the existing CIP-0030 functionality for web wallets under a new global-level namespace rather than simply creating the handful of new API calls as an extension to CIP-0030 itself? i.e.
> 
> ```js
> window.cardano.broclan.enable([130])
> ```

Multisig Wallets cannot implement CIP-30 since it is impossible to implement the api.signTx() and api.signData() endpoints. Implementing it as an extension of CIP-30 will lead to broken applications that try to connect to the multisig wallets and use them as if they are regular wallets.

> 2. The idea of being able to use Script/Multisig addresses to interact with dApps is a good one, I would like to see more rationale and justification surrounding how dApps are expected to behave when encountering a multisig address... Is it up to the dApp to maintain the record of unsigned/unfinished multisig transactions? Should these naturally "decay" over time? Doesn't this present an issue of requiring dApps to be centralized rather than decentralized by placing this onus on the dApp to record state rather than putting the onus on the signers? Maybe a `importTransaction` endpoint makes sense so that members of the multisig can pass it around amongst themselves? Or the wallet provider could have a record of `pending` transactions somehow passed between members of the multisig?

The wallet is responsible for passing the transactions among the participants, gathering the signatures, composing the final transaction and submitting it on chain, dApps need to keep track of transactions only when hidden metadata is included (only the hash is provided) since in that scenario the wallet does not have all the required information to submit the transaction 

The functionality you are asking is already with "importTransaction" included with "submitUnsignedTx"
 

> 3. We do not allow for "reservation" of CIP numbers, regardless of rationale or justification with only a handful of very notable exceptions. You should move this file to an `CIPXXXX` directory and await assignment of a CIP number.

Done, thank you for the clarification and sorry for the confusion.

rphair:
@leo42 I look forward to reading this for content.  In the meantime https://github.com/cardano-foundation/CIPs/pull/617#issuecomment-1806847726 is right that we can't allow choosing CIP numbers in the present, let alone reserving them for the future: it would require a conflict resolution system that would be unmanageable because of several technical & human factors.

I have to rename it in the PR title in the meantime to avoid confusion, and also remove the redundant word "Cardano".  I'm also assuming based on a quick review that, while the API strongly resembles CIP-0030, that this is a completely separate API to CIP-0030, and is therefore not an extension to it... please let me know if I'm wrong about that.  cc @Ryun1

leo42:
@rphair you are right, as I explained in the previous commend :

>Multisig Wallets cannot implement CIP-30 since it is impossible to implement the api.signTx() and api.signData() endpoints. Implementing it as an extension of CIP-30 will lead to broken applications that try to connect to the multisig wallets and use them as if they are regular wallets.

So calling it an extension is confusing.

CIP-XXXX/README.md
```diff
@@ -0,0 +1,408 @@
+---
+CIP: ???
+Title: Cardano dApp-MutlisigWallet Web Bridge
+Status: Proposed
+Category: Wallets
+Authors: 
+  - Leo
+Implementors:
+  - BroClan <https://broclan.io/>
```
rphair, line 9:
```suggestion
Implementors: NA
```
It might not be necessary to remove BroClan from the list of `Implementors` (which should list only the full implementations at the time of CIP submission) if you have a working implementation in your wallet.

If you do have a complete implementation, then it should be clearly marked as implemented down in your _Implementation Plan_ (currently that box isn't checked).

rphair, line 9:
```suggestion
Implementors: NA
```
It might not be necessary to remove BroClan from the list of `Implementors` (which should list only the full implementations at the time of CIP submission) if you have a working implementation in your wallet.

If you do have a complete implementation, then it should be clearly marked as implemented down in your _Implementation Plan_ (currently that box isn't checked).

CIP-XXXX/README.md
```diff
@@ -0,0 +1,408 @@
...
+
+### Acceptance Criteria
+
+- [ ] The interface is implemented and supported by various wallet providers.
+- [ ] The interface is used by dApps to interact with wallet providers. 
+  
+### Implementation Plan
+
+- [ ] Provide some reference implementation of wallet providers
+  - [leo42/BroClanWallet]
```
rphair, line 402: (continuing previous comment) Hard to tell whether "BroClan Wallet" is a working implementation because of the ambiguous language... where do you still have yet to "provide" a reference implementation? in the CIP, or in software?

CIP-XXXX/README.md
```diff
@@ -0,0 +1,408 @@
...
+
+
+
```
rphair, line 405:
```suggestion

```

rphair (commented):
> (https://github.com/cardano-foundation/CIPs/pull/617#issuecomment-1806847726) What is the rationale here to "replicate" the existing CIP-0030 functionality for web wallets under a new global-level namespace rather than simply creating the handful of new API calls as an extension to CIP-0030 itself? i.e.
> `window.cardano.broclan.enable([130])`

I would first want to seriously consider this issue on both sides... on general principle (I'm not a wallet or dApp developer) I can't imagine going forward with this as written due to practically everything from CIP-0030 being duplicated & therefore prone to divergence: not only in the documentation, but even worse in behavioural inconsistencies between conventional & multisig wallets.

I agree with that I think @Crypto2099 is suggesting that the CIP call only for whatever minimum of functional differences is required by the multisigs.

If @leo42 it is a widespread problem that people are connecting with the wrong type of wallets then shouldn't we be able to settle on a definitive way of handling that though an initial query... even if some changes are necessary to CIP-0030 to support this?

I've put this on the Triage list for our upcoming CIP meeting (https://hackmd.io/@cip-editors/76) ... please @leo42 come to the meeting if you can, to help us all agree on the best footing to proceed with this 😎

CIP-XXXX/README.md
leo42: yes I will be creating a reference implementation in BroClanWallet, its the next thing on the work agenda.

I am not sure how to mark that to make it more understandable. 

I expect the implementation to be completed in ~1 month.

leo42:
> @leo42 I look forward to reading this in more detail for content. In the meantime [#617 (comment)](https://github.com/cardano-foundation/CIPs/pull/617#issuecomment-1806847726) is right that we can't allow choosing CIP numbers in the present, let alone reserving them for the future: it would require a conflict resolution system that would be unmanageable because of several technical & human factors.
> 
> I have to rename it in the PR title in the meantime to avoid confusion, and also remove the redundant word "Cardano". I'm also assuming based on a quick review that, while the API strongly resembles CIP-0030, that this is a completely separate API to CIP-0030, and is therefore not an extension to it... please let me know if I'm wrong about that. cc @Ryun1

You are correct, implementing as an extension would would lead to problems and it is 

> > ([#617 (comment)](https://github.com/cardano-foundation/CIPs/pull/617#issuecomment-1806847726)) What is the rationale here to "replicate" the existing CIP-0030 functionality for web wallets under a new global-level namespace rather than simply creating the handful of new API calls as an extension to CIP-0030 itself? i.e.
> > `window.cardano.broclan.enable([130])`
> 
> I would first want to seriously consider this issue on both sides... on general principle (I'm not a wallet or dApp developer) I can't imagine going forward with this as written due to practically everything from CIP-0030 being duplicated & therefore prone to divergence: not only in the documentation, but even worse in behavioural inconsistencies between conventional & multisig wallets.
> 
> I agree with that I think @Crypto2099 is suggesting that the CIP call only for whatever minimum of functional differences is required by the multisigs.
> 
> If @leo42 it is a widespread problem that people are connecting with the wrong type of wallets then shouldn't we be able to settle on a definitive way of handling that though an initial query... even if some changes are necessary to CIP-0030 to support this?
> 
> I've put this on the Triage list for our upcoming CIP meeting (https://hackmd.io/@cip-editors/76) ... please @leo42 come to the meeting if you can, to help us all agree on the best footing to proceed with this 😎

@rphair I will be happy to attend if you provide a meeting link.

CIP-XXXX/README.md
rphair: sure, I've updated the diff above with my suggestion... removing some extra language and unnecessary nesting.

If adoption by other wallet providers is a part of your Implementation Plan, please add some more checkboxes with the adoption plan as you see it. 🤓

leo42: I am in contact with the Gamechanger and ADAO devs trying to get them onboard.

rphair:
Discord server (CIP Editors Meetings) invitation: https://discord.gg/JtWfD4hqgz
Discord server link: https://discord.com/channels/971785110770831360
Meeting link: https://discord.com/events/971785110770831360/1164384910551818330

Crypto2099:
> > 1. What is the rationale here to "replicate" the existing CIP-0030 functionality for web wallets under a new global-level namespace rather than simply creating the handful of new API calls as an extension to CIP-0030 itself? i.e.
> > 
> > ```js
> > window.cardano.broclan.enable([130])
> > ```
> 
> Multisig Wallets cannot implement CIP-30 since it is impossible to implement the api.signTx() and api.signData() endpoints. Implementing it as an extension of CIP-30 will lead to broken applications that try to connect to the multisig wallets and use them as if they are regular wallets.

I disagree with the premise on principle. It would be quite trivial for a multi-sig wallet to only allow `enable` calls if the `cip-130` (I know we've agreed to change the number but we'll use it for example here) extension is enabled and otherwise throw an error similar to not having a dApp account selected.

Either solution would require major overhauls for all existing dApps to account for and handle multisig wallets so it's not really a matter of one being "easier" for the dApp ecosystem to handle or not, but by making this a multisig-specific extension to CIP-30 we have less concern about maintaining feature parity when there are naturally extensions or modifications to CIP-30.

leo42:
I agree that maintainability will be easier if it is implemented as an
extension, and I did not consider simply rejecting connections that do not
contain the extension number.

But it will still require some changes to CIP-30 since its specifications
say that all the base endpoints need to be supported and extensions only
add new endpoints.

Can I add changes to another CIP as part of a single pull request ?

p.s. I know Mlabs is supposed to do a rework of CIP-30 but they have not
come back to me yet about coordinating on that front.

rphair:
> @leo42 https://github.com/cardano-foundation/CIPs/pull/617#issuecomment-1807116642 Can I add changes to another CIP as part of a single pull request ?

Yes, this is common when a CIP requires changes to data structure like the registry tables in CIP-0010 & CIP-0067.  But I don't know of any examples where one CIP PR has made fundamental changes to another prerequisite CIP ... which I think we are considering here.

Generally I think these types of problems are best dealt with through versioning: e.g. "v1 of CIP-0030 doesn't discriminate properly between conventional & multisig wallets, but v2 will according to CIP-0130" (again, naming this informally).

Regardless of exactly how CIP-0030 might be versioned or extended, I believe this PR for this proposal alone should be fully reviewed & merged first.... documenting any dependencies upon further modifications to CIP-0030 in its __Path to Active__ but without modifying CIP-0030 itself... before any future PR to extend or version bump CIP-0030.  Otherwise I can imagine a serious delay in adopting multisig capability as described here because of uncertainty about breaking changes.

Ryun1:
Hey @leo42 

Thanks for this proposal.
I agree that such functionality should be pursued and added to a dApp-wallet connector.

Although, copying CIP30 endpoints under a new namespace I feel is a mistake.
I would be much more in favour to use the CIP-30 extension mechanism for this, this is the sort of thing the mechanism was added for.
Any potential problems of adding this as an extension I think are less consequential than trying to split this from CIP-30 which introduces unneeded fragmentation IMO.
If you strongly believe that this does not belong as a CIP-30 extension, I would implore you to check out [CIP-90](https://github.com/cardano-foundation/CIPs/pull/462/), which was an attempt to allow for extensions without the need to include CIP30 endpoints.

Looking forward to seeing how this proposal progresses.

Crypto2099:
> But it will still require some changes to CIP-30 since its specifications say that all the base endpoints need to be supported and extensions only add new endpoints. 

Having the base endpoints be `supported` doesn't mean that they can't naturally return an error if a specific wallet (i.e. multisig, Plutus script, whatever) cannot support them as-written. I would view this more like function overloading on the part of the wallet implementers rather than complete replication.

CIP-XXXX/README.md
```diff
@@ -0,0 +1,404 @@
...
+## Motivation: why is this CIP necessary?
+
+In order to facilitate future dApp development, we will need a way for dApps to communicate with multisig wallets, given the unique complexities of native script based address special provisions need to be made to make the connector compatible with them.  
+
+Specifically apps building transactions need to be able to get the following information from the wallet:
+- Script descriptor 
+- Script Requirements list 
+- Collateral donator (since native script based addresses cannot provide collateral for transactions)
```
klntsky, line 26: Is this for the case when there are plutus scripts as well as native scripts present? Because native scripts don't require collateral.

leo42:
Yes, in most cases dApps are built using plutus and we need to make special
provisions since Native Scripts cannot provide collateral.

klntsky (commented):
I don't understand why is it needed to return  the native script contents to dApps. As I see it, dApps should be able to work with native script addresses just fine without ever knowing who are the signatories. I think the answer to this should be in the motivation section.

When calling getCollateral, the wallet can return any of the UTxOs controlled by some of the parties as well.

The only real problem that I see in CIP-30 is that gathering of signatures can take a long time and promise-based API does not play well with page reloads.

leo42:
If you create a transaction without knowing the exact signers you will not
be able to calculate the fees correctly.

Also including the correct "required signers" brings UX improvements, and
you need to know if the script has any before-after requirements to set the
correct validBefore and TTL values.

Transactions consuming inputs from a native script address need to attach
the script.

rphair:
thanks @leo42, that definitely reflects the intentions of the CIP Editors' meeting today... @Ryun1 @Crypto2099 if you can confirm this is now properly integrated as an CIP-0030 extension, then I think it would be good to have on the agenda again next CIP meeting (# 77) to confirm the scope & keep it moving forward.

CIP-XXXX/README.md
```diff
@@ -0,0 +1,173 @@
...
+
+Returns the CBOR-encoded native script that controls this wallet. 
+
+#### api.submitUnsignedTx(tx: cbor\<unsignedTransaction>): Promise\<hash32>
+
+Errors: `APIError`, `TxError`
+
+Submits a transaction to the wallet for signing. The wallet should check that the transaction is valid, gather the required signatures, compose the finalized transaction, and submit the transaction to the network. If the transaction is valid and the wallet is able to sign it, the wallet should return the transaction hash. If the transaction is invalid or the wallet is unable to sign it, the wallet should throw a `TxError` with the appropriate error code. The wallet should not submit the transaction to the network if it is invalid or the wallet is unable to sign it.
+
+If the transaction contains hidden metadata, the wallet should not submit the transaction when it is ready, but return it to the dApp when the dApp calls the `getCompletedTx` function.
```
mkazlauskas, line 130: I don't understand this point, could you please elaborate? Is it about witness datum? If so, why can't the dapp provide a partial witness set with it's own witnesses when calling this method?

CIP-XXXX/README.md
```diff
@@ -0,0 +1,173 @@
...
+
+Returns a list of ScriptRequirements that will be used to validate any transaction sent to the wallet.
+
+#### api.getScript(): Promise\<cbor\<nativeScript>>
+
+Errors: `APIError`
+
+Returns the CBOR-encoded native script that controls this wallet. 
+
+#### api.submitUnsignedTx(tx: cbor\<unsignedTransaction>): Promise\<hash32>
```
mkazlauskas, line 124: Maybe we could find a way to utilize existing sign() and submit() instead. For example: to address sign() being a long-running operation, wallets could make `sign()` idempotent and associate subsequent calls with an already ongoing signing operation. Wallets have to maintain state of transaction being signed anyway, so it shouldn't be too difficult to implement. This way, dapps could recover the state (get a callback when signing is complete) when they're reloaded.

CIP-XXXX/README.md
```diff
@@ -0,0 +1,173 @@
...
+
+dApp developers can choose to use this address to add collateral to a transaction, or they can choose to use the `api.getCollateral()` function to get a list of UTXOs that can be used as collateral. If the dApp chooses to use this address, they must ensure that the address is not used for any other purpose, as the wallet may be using it to track collateral, and that the collateral return address is the same one.
+
+#### api.getScriptRequirements: Promise\<ScriptRequirement[]>
+
+Errors: `APIError`
+
+Returns a list of ScriptRequirements that will be used to validate any transaction sent to the wallet.
+
+#### api.getScript(): Promise\<cbor\<nativeScript>>
```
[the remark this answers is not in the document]
leo42: The extension is for Native scripts. 

When it is successful we can design a new extension for Plutus script.

CIP-XXXX/README.md
```diff
@@ -0,0 +1,173 @@
...
+
+### Additional API Endpoints
+
+#### api.getCollateralAddress(): Promise\<Address>
+
+For Plutus V2 and later, partial collateral is supported. This function returns an address that can be used to add collateral to a transaction. The address returned must be owned by one of the signers in the list of signers returned by `api.getScriptRequirements()`. 
+
+dApp developers can choose to use this address to add collateral to a transaction, or they can choose to use the `api.getCollateral()` function to get a list of UTXOs that can be used as collateral. If the dApp chooses to use this address, they must ensure that the address is not used for any other purpose, as the wallet may be using it to track collateral, and that the collateral return address is the same one.
+
+#### api.getScriptRequirements: Promise\<ScriptRequirement[]>
```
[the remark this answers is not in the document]
leo42: As mentioned there are 3 reasons to set this structure and not return a simple int for the number of signatures.
- Script Requirements list:
	-- dApps need to know the number of signers to calculate the fees correctly.
	-- Including the correct "required signers" brings UX improvements.
	-- dApps need to know if the script has any before-after requirements to set the correct validBefore and TTL values.

CIP-XXXX/README.md
leo42: Metadata not datum. 

when using sighTx() you are able to attach the hash of the transaction metadata, sighing the hash makes the transaction valid, but to submit the transaction you need the actual metadata

CIP-XXXX/README.md
leo42: This is a complex problem with many possible solutions.

Long polling, callbacks, web-hooks. I recommend this be addressed in other CIPs each of witch will implement a single solution.  That will save us from getting locked into a specific mode of operation at this stage.

CIP-XXXX/README.md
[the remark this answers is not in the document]
leo42: In your example wallet implementers are responsible to create the UI/UX where the user choses if the signers are (A+B) or C. 

So the output of the connector will be (A+B). 

Or if the user specified C the connector will return only (C)

CIP-XXXX/README.md
```diff
@@ -0,0 +1,173 @@
+---
+CIP: ???
+Title: Cardano dApp-MutlisigWallet Web Bridge (CIP-30 extencion)
```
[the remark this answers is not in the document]
leo42: Multisig wallet has been well established as Native script based wallets in the Cardano ecosystem. 

All Native script wallets implementers have used the Multisig terminology ( ADAO, BroClan, Summon platform and gameChanger). It also taps into the pedagogy from other blockhains that use multisig wallets to make it easier for people to understand the technology as it has been named Multisig in those ecosystems. 

Shared wallets is a pretty new terminology I have seen floating around and has not been adopted by any implementers.  

I am open to changing the terminology here but at the moment is seems like more people will be searching for it under the "multisig" terminology rather.

CIP-XXXX/README.md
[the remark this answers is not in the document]
mkazlauskas: > -- Including the correct "required signers" brings UX improvements.
> -- dApps need to know if the script has any before-after requirements to set the correct validBefore and TTL values.

I'm not sure about the value of those - user could see signing info and status in the wallet. Also if the dapp was multi-sig specific, then it would probably be able to parse the script from `getScript`. I'm probably missing some dapp use case(s) here, do you have something specific in mind?

> In your example wallet implementers are responsible to create the UI/UX where the user choses if the signers are (A+B) or C.

That sounds reasonable, but then this method seems to be misleading - it's not really the script requirements, more like constraints or a plan for signing the next transaction :thinking:

CIP-XXXX/README.md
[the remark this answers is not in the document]
leo42: Plutus connector will have to have much more nuance in the requirements and supported operations. 

A connector that supports both would be overcomplicated  and would make implementing support for only one of the 2 
 very complicated.

CIP-XXXX/README.md
mkazlauskas: Ah I see, thanks. In that case, why can't the dapp provide auxiliary data when signing?

CIP-XXXX/README.md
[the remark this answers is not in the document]
leo42: > I'm not sure about the value of those - user could see signing info and status in the wallet. Also if the dapp was multi-sig specific, then it would probably be able to parse the script from getScript. I'm probably missing some dapp use case(s) here, do you have something specific in mind?

you said it your self, native script are not flat. Before and After conditions could exist but not be used to validate a transaction if alternative exits 

example vesting: (A or (B and AFTER x)), 'AFTER x' is only required if A is not one of the signers. the connector must return the exact list or condition to validate a single Tx.

>That sounds reasonable, but then this method seems to be misleading - it's not really the script requirements, more like constraints or a plan for signing the next transaction 🤔

I am open to modifications that will make the use of this endpoint more intuitive.

CIP-XXXX/README.md
leo42: I could be the intension to hide the metadata from the user.

I created a multisig NFTs minting  service for example, but you don't want the user to know the details of the NFT they are about to mint until the Tx is on-chain.

mkazlauskas: Sounds shady but I'll trust it's a legit use case :joy:

If we can find a way to keep sign and submit separate, then it's a non-issue and would make it simpler.

leo42: >If we can find a way to keep sign and submit separate, then it's a non-issue and would make it simpler. 

This endpoint is where we separate them  🤓

But for regular operation having it all in one go, is much better UX and DevX.

mkazlauskas: > This endpoint is where we separate them 🤓

Hm, I understand `If the transaction contains hidden metadata, the wallet should not submit the transaction when it is ready` that otherwise wallet can/should submit it after it's signed.

mkazlauskas: > But for regular operation having it all in one go, is much better UX and DevX.

I can see the appeal though. Maybe it could be an explicit argument/option set by the dapp instead, like `{ submit: boolean }`. What I don't like in the current design is that it's likely this logic would have to be extended, in the future (like applying similar logic for datums/redeemers if dapp wants to keep them secret).

leo42: Good point, I was not aware this is even possible with datums and redeemers.

CIP-XXXX/README.md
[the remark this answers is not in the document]
mkazlauskas: I have a counter example for UX: consider a wallet with a script requiring signature of `A or B`. Then,
- if this API didn't require providing specific signers, then wallet could just say that it will sign with 1 signature
- if this API requires it, wallet has to add an additional step for the user to select a specific signer for the next transaction. And it can't really select a random one, since the intent specified in this CIP is for 'dapp ux' - so it has to assume that dapp would show the signer's address somewhere.

So if you consider the overall UX (dapp + wallet), not just the dapp, then I'd argue it makes it worse in this case.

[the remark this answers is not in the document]
leo42: No because including the correct signers in the "Required signers" of the TX allows the wallet to guide the user to find the missing signatures much easier. If you provide a simple int you are left with a user that has no idea who has to sigh what. 

And UX design does not require you ask the user every time.  It can even be done automatically for the initial value but give the user the option to change it in the settings. A proper comminication of dApp-wallet should contain the exact precise information.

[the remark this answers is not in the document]
mkazlauskas: > No because including the correct signers in the "Required signers" of the TX allows the wallet to guide the user

Wallet already knows the signing rules based on the script, I don't think the dapp should decide which specific signers to require.

>  And UX design does not require you ask the user every time. It can even be done automatically for the initial value but give the user the option to change it in the settings.

The point I was trying to make is that with current API wallet **shouldn't** decide it automatically due to this point in the cip:
> -- Including the correct "required signers" brings UX improvements.

In the `A or B` example, if the wallet automatically choose `A`, but user signs it with `B`, then there's a mismatch between what the dapp potentially shows to the user and the actual signature. Even worse if the dapp sets `requiredSignatures` to `A` - user then can't even sign it with `B`.

zxpectre:
Hi Leo, Adriano from GameChanger Wallet here. Sorry for the delays on the feedback, i've been offline and working on some deadlines!

> If you create a transaction without knowing the exact signers you will not be able to calculate the fees correctly. Also including the correct "required signers" brings UX improvements, and you need to know if the script has any before-after requirements to set the correct validBefore and TTL values. Transactions consuming inputs from a native script address need to attach the script.

Regarding fees and required signers, I've created the concept of **Optional Signers** on **[GameChanger Wallet multisig dapp connector](https://github.com/GameChangerFinance/gamechanger.wallet/blob/main/DAPP_CONNECTOR.md)** to be able to reserve the greatest (worst) fee (possible or expected) as way to minimize the use of _required signers feature_ because how I see it it harms UX when "signature aggregator entity" cannot not know the final outcome, the final set of witnesses that will be provided. Specially on "N of M signers" use cases. This happends a lot, specially from dapps pov, as far as I can tell from our experience connecting Shared Treasuries with dapps.

It would be nice to talk to coordinate the development of a tiny lib to help us auto-calculate the outcomes more accurately for **BroClan** and **GameChanger Wallet**, currently I have a solution for this but with help we can extend it.

rphair:
FYI @zxpectre haven't noticed you in the latest [Discord discussion of these issues](https://discord.com/channels/971785110770831360/1176612265894613133) so wanted to alert you to Monday's wallet connectors workshop: https://discord.com/events/971785110770831360/1176916387046887535 (invite:https://discord.gg/UGcJSKzDfA )

zxpectre:
Thanks @rphair, i didn't know about these meetings, I will try to assist but not sure if i'm going to make it on time.

Can you please record it?

Btw @leo42 here is one [example use case](https://github.com/GameChangerFinance/gamechanger.wallet/blob/main/examples/Internal%20Multi%20Signing%20Demo.md) of the Optional Signers feature on the transaction builder of our dapp connector. There are other ways you can deal with this.

rphair:
@zxpectre you can join the Discord to find out about future meetings, with all the ideas that came before / during / after this meeting.  Here are the presentation slides with @Ryun1's notes from last hour's meeting taken on top of them: https://docs.google.com/presentation/d/1lXw1yMf-Hyp1eVfaFnbTV9897r6pZEd6zpvLNMqT3cI/

CIP-XXXX/README.md
```suggestion
Title: Web-Wallet Bridge - Mutlisig wallets
```

rphair (commented):
> @leo42 (https://github.com/cardano-foundation/CIPs/pull/617#issuecomment-1811034123): I have reworked and renamed the proposal to be a CIP-30 extension.

So it's clear that this is now an extension for CIP-30 & potentially other wallet interfaces, can you retitle it like our other CIP titles for these extensions? e.g. https://github.com/cardano-foundation/CIPs/pulls?q=is%3Apr+Web-Wallet+Bridge+in%3Atitle

I'm retitling this PR title so we can all see how the suggestion below looks in practice & we can make further adjustments if needed (ultimately keeping the document & PR titles the same):

rphair:
@siegfried I just found from your https://github.com/cardano-foundation/CIPs/pull/611#issuecomment-1828550747 that you've produced a multisig wallet, so your feedback and/or review would be welcome here & and at any of the Discord links above.

Gueedo78 (approved):

CIP-XXXX/README.md
```diff
@@ -0,0 +1,173 @@
+---
+CIP: ???
```
```suggestion
CIP: 106
```

rphair (commented):
Thanks @leo42 for following through at CIP meeting today & happy to see this promoted as Candidate... please rename the directory from to `CIP-0106` and update your original "Rendered version" link accordingly 🎉

rphair:
@leo42 @Crypto2099 @Ryun1 @siegfried @zxpectre, while you are all here in one place: Have any of you produced, or do you know the production of, a multisig wallet which works according to CIP-1854? (re: https://github.com/cardano-foundation/CIPs/pull/727)

leo42:
> @leo42 @Crypto2099 @Ryun1 @siegfried @zxpectre, while you are all here in one place: Have any of you produced, or do you know the production of, a multisig wallet which works according to CIP-1854? (re: #727)

Not me, BroClan is a single address wallet.

siegfried:
> @leo42 @Crypto2099 @Ryun1 @siegfried @zxpectre, while you are all here in one place: Have any of you produced, or do you know the production of, a multisig wallet which works according to CIP-1854? (re: #727)

RoundTable supported CIP-1854 since V2. You can try it for yourself. https://round-table.vercel.app

rphair:
@leo42 an article from @Hornan7 that illustrates multisig features for dReps... perhaps these functions would also influence the design of a multisig wallet API? (cc @Ryun1 @Crypto2099) https://github.com/cardano-foundation/CIPs/pull/617

CIP-0106/README.md
```diff
@@ -0,0 +1,173 @@
...
+CompletedTxErrorCode = {
+	NotFound: 1,
+	NotReady: 2
+}
+```
+
+* NotFound - The transaction with the given id was not found.
+* NotReady - The transaction with the given id is not ready yet. 
+
+## Motivation: why is this CIP necessary?
```
Ryun1, line 83: duplicate section

CIP-0106/README.md
```diff
@@ -0,0 +1,173 @@
...
+
+- Script descriptor
+- Script Requirements list
+- Collateral donator (since native script based addresses cannot provide collateral for transactions)
+
+Additionally, apps need to be able to submit a transaction to the wallet for signing in an asynchronous manner, as gathering of signatures can take a long time and each wallet provider will have its own way of handling this process.
+
+Finally, the signTx() and signData() endpoints will have to be disabled when using this extension since they are not compatible with native script based addresses.
+
+## Rationale for the required data
```
Ryun1, line 96:
```suggestion
### Rationale for the required data
```

CIP-0106/README.md
```diff
@@ -0,0 +1,173 @@
...
+## Rationale: how does this CIP achieve its goals?
+
+See justification and explanations provided with each API endpoint.
+
+
+## Path to Active
+
+### Acceptance Criteria
+
+- [ ] The interface is implemented and supported by various wallet providers.
```
Ryun1, line 163:
```suggestion
- [ ] The interface is implemented and supported by two wallet providers.
```
nice to be specific

CIP-0106/README.md
```diff
@@ -0,0 +1,173 @@
...
+- [ ] The interface is used by dApps to interact with wallet providers.
```
Ryun1, line 164:
```suggestion
- [ ] The interface is used by multiple dApps to interact with wallet providers. 
```

CIP-0106/README.md
rphair: @Ryun1 not really a duplicate section but @leo42 this is the same problem as I had to fix in https://github.com/cardano-foundation/CIPs/pull/798#discussion_r1586188780 where "Rationale for the required data" in the Motivation... really just a definition of specification terms... seemed to be a separate section.

Since your material is in exactly the same form here, it can be fixed exactly the same way as in https://github.com/cardano-foundation/CIPs/pull/798/commits/25b463c99f77dc48d4f8034a3e06e6bfd9fa8323.

CIP-0106/README.md
rphair: A continuation of https://github.com/cardano-foundation/CIPs/pull/617#discussion_r1600416733.

leo42:
> @leo42 an article from @Hornan7 that illustrates multisig configuration for dReps... perhaps these requirements would also influence the design of a multisig wallet API? (cc @Ryun1 @Crypto2099) https://forum.cardano.org/t/how-to-create-a-multi-sig-drep/130159



> @leo42 an article from @Hornan7 that illustrates multisig configuration for dReps... perhaps these requirements would also influence the design of a multisig wallet API? (cc @Ryun1 @Crypto2099) https://forum.cardano.org/t/how-to-create-a-multi-sig-drep/130159

Yes we are fully compatible with this. I will build a dRrep management and registration portal using the connector so all users of BroCllan(and other providers that implement it) the ability to participate in governance .(when I find some time )

CIP-0106/README.md
rphair: @leo42 - https://github.com/cardano-foundation/CIPs/pull/617/commits/9d6a57f339fe1ae3c65ed418a716f1495d8ad38c doesn't fix this problem (an apparently duplicated Rationale section).  Removing that title — and joining the "required data" terms with their definitions, as suggested above — _would_.

CIP-0106/README.md
```diff
@@ -0,0 +1,151 @@
+---
+CIP: 106
+Title: Web-Wallet Bridge - Mutlisig wallets
+Status: Proposed
+Category: Wallets
+Authors: 
+  - Leo
+Implementors: NA
+Discussions:
```
Ryun1, line 9:
```suggestion
Discussions:
  - https://github.com/cardano-foundation/CIPs/pull/617
```

CIP-0106/README.md
```diff
@@ -0,0 +1,152 @@
+---
+CIP: 106
+Title: Web-Wallet Bridge - Mutlisig wallets
+Status: Proposed
+Category: Wallets
+Authors: 
+  - Leo
+Implementors: NA
```
rphair, line 8: Since what you list as a Reference Implementation appears to be a working wallet, you should remove `NA` and list yourself (if alone) or BroClanWallet or its organisation (if this is a group effort) as an Implementor.

CIP-0106/README.md
```diff
@@ -0,0 +1,152 @@
...
+
+### Acceptance Criteria
+
+- [ ] The interface is implemented and supported by multiple wallet providers.
+- [ ] The interface is used by multiple dApps to interact with wallet providers. 
+	
+### Implementation Plan
+
+- [x] Provide some reference implementation of wallet providers
+	- [leo42/BroClanWallet](#completed)
```
rphair, line 148:
```suggestion
	- [leo42/BroClanWallet](https://github.com/leo42/BroClanWallet)
```

rphair (commented):
@leo42 I'm ready to approve this after 1) adjusting the `Implementors:` as below and 2) fixing the section headings & content as requested in https://github.com/cardano-foundation/CIPs/pull/617#discussion_r1600416733 ... as both you & I already have for your earlier 2 CIPs (103, 104) which follow the same form.

CIP-0106/README.md
```diff
@@ -0,0 +1,152 @@
...
+
+If the transaction is not ready, the wallet should throw a `CompletedTxError` with the appropriate error code. If the transaction is ready, the wallet should return the CBOR-encoded transaction and the signatures.
+
+### Altered API endpoints 
+
+#### api.getCollateral(params: { amount: cbor\<Coin> }): Promise\<TransactionUnspentOutput[] | null>
+
+Native script based addresses cannot provide collateral for transactions. Using this function, dApps can request the wallet to provide collateral for a transaction. The collateral must be a pure ADA UTXO, held by one of the signers in the list of signers returned by `api.getScriptRequirements()`.
+
+### Removed API endpoints
```
Ryun1, line 123: Im unsure if CIP-30 extensions are permitted to disable default endpoints from CIP-30.
This is an interesting idea, not sure of potential conflicts with other extensions enabled at the same time.

CIP-0106/README.md
rphair: resolving; fixed by https://github.com/cardano-foundation/CIPs/pull/617/commits/ce033b7d29946956417015b9efe8f3af14d06576

CIP-0106/README.md
rphair: Fixed by https://github.com/cardano-foundation/CIPs/pull/617/commits/ce033b7d29946956417015b9efe8f3af14d06576 ... although FYI @leo42 it was not necessary to remove these "required data" definitions (they could simply be attached to the terms they define, as we did in https://github.com/cardano-foundation/CIPs/commit/25b463c99f77dc48d4f8034a3e06e6bfd9fa8323).

leo42: Good suggestion, sorry I have gotten lost among the several branches of this conversation

rphair: understood @leo42 & fully resolved in https://github.com/cardano-foundation/CIPs/pull/617/files/483b8c87c3872ac44e5f054911dcbea4877969de..1d13e7073301793d2a7e3f54aeea3709f931d27e. 😎

CIP-0106/README.md
mkazlauskas: Please no API endpoint removals :sweat_smile: imho if it throws an informative error then dev experience is fine

CIP-XXXX/README.md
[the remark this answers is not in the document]
mkazlauskas: The CIP now mentions Plutus scripts, did this change?

rphair (approved):
@leo42 after compliance with the last round of recommendations I think this is sound enough to merge, pending @Crypto2099's final review for technical foundation + document clarity for potential implementors.

This has been at `Last Check` for a while, which we will again revisit in 3 days as per meeting agenda https://hackmd.io/@cip-editors/91.

Crypto2099 (approved):
Given that no further issues have been raised and the author also feels that this is in a state that is satisfactory, I see no blockers to prevent this from being merged in and continuing on its path to active.

PR #619, opened 2023-11-16
CPS-0010? | Wallet Connectors

Ryun1:
This problem statement is concerned with the issues surrounding Cardano's current and future wallet connectors.
These interfaces are difficult to define and historically have been even harder to iterate upon.
We wish to provide a comprehensive catalog of the current offerings; and their drawbacks to be able to make suggestions on future standards.

I thought it important that such a problem statement is open for much community input, so have started this as a draft PR.

#### Todo
- [x] Hold first workshop
- [x] Hold second workshop
- [x] Enrich context subsection
- [x] Add to open questions section

#### Open Questions
- <s> Can wallet connector standards define when wallets should ask users for permission?</s>
  - No, this is the responsibility of the wallet, although a separate CIP could define general best practices. 
- <s>Where does the role of transaction building sit?</s>
  - Will push this question to a subsequent CIP.
---

[📰 Rendered Problem Statement (on fork)](https://github.com/Ryun1/CIPs/blob/cps-wallet-connector/CPS-0010/README.md)

PR #620, opened 2023-11-17
CIP-???? | Web-Wallet Bridge - Extensive Light Wallet Standard 

klntsky:
This PR adds a new CIP that describes a webpage-based communication bridge allowing dApps to interface with Cardano wallets. This interface extends CIP-30. Improvements over the original standard are:

- More data queries
- Interfaces that allow transaction building on the wallet side
- Use of JSON instead of CBOR

Context: https://cardano.ideascale.com/c/idea/103857
Discussion: https://discord.gg/XFjAzsQyaM

---

[Rendered Proposal on fork](https://github.com/klntsky/CIPs/blob/klntsky/wallet-standard-cip/CIP-XXXX/README.md)

### DIFF
--- a/CIP-XXXX/README.md
+++ b/CIP-XXXX/README.md
@@ -7,7 +7,7 @@ Authors:
     - Vladimir Kalnitsky <vladimir@mlabs.city>
 Implementors: []
 Discussions:
-    - https://github.com/cardano-foundation/cips/pulls/?
+    - https://github.com/cardano-foundation/cips/pulls/620
     - [Cardano Working Groups Discord server](https://discord.gg/XFjAzsQyaM)
 Created: 2023-11-14
 License: CC-BY-4.0
--- a/CIP-XXXX/README.md
+++ b/CIP-XXXX/README.md
@@ -1,0 +1,230 @@
+---
+CIP: ?
+Title: Extensive Light Wallet Standard
+Category: Wallets
+Status: Proposed
+Authors:
+    - Vladimir Kalnitsky <vladimir@mlabs.city>
+Implementors: []
+Discussions:
+    - https://github.com/cardano-foundation/cips/pulls/?
+    - [Cardano Working Groups Discord server](https://discord.gg/XFjAzsQyaM)
+Created: 2023-11-14
+License: CC-BY-4.0
+---
+
+## Abstract
+<!-- A short (\~200 word) description of the proposed solution and the technical issue being addressed. -->
+
+This document describes a webpage-based communication bridge allowing dApps to interface with Cardano wallets. The interface extends CIP-30. Improvements over the original standard are:
+
+- More data queries
+- Interfaces that allow transaction building on the wallet side
+- Use of JSON instead of CBOR
+
+## Motivation: why is this CIP necessary?
+
+### Existing problems of dApp development
+
+CIP-30 is a universally accepted web-based wallet standard for Cardano. It provides a minimalistic interface that, in principle, can be used to build almost any kind of Cardano dApp. However, the way dApp<->wallet interaction is defined leads to suboptimal dApp architecture due to CIP-30 limits.
+
+Consider the following problems:
+
+#### Manual Transaction Building
+
+On Cardano, the whole process of constructing transactions is delegated to the off-chain code. This process typically includes:
+
+1. Fetching the data from the blockchain, using either CIP-30 interface, custom backend, or, most commonly, both
+2. Determining the parameters (or constraints) the transaction should have using internal dApp logic
+3. Constructing a skeleton of the transaction given data from the previous steps
+4. Making the transaction structure valid for stage1 and stage2 of the validation process - balancing (making sum of outputs plus fees equal to sum of inputs), estimating execution units, and attaching script data (“datums”) and redeemers.
+5. Serializing the transaction to CBOR format (commonly via cardano-serialization-lib)
+6. Passing the serialized transaction for signature and submission
+
+It is clear that steps 3, 4 and 5 are purely technical: most (however, not all) dApps would opt out of having to deal with these if only they could.
+
+#### Use of CBOR representations
+
+CIP-30 standard uses CBOR encoding for all data passed from the wallet, e.g. addresses and UTxOs. Interpreting this data within the app requires CBOR decoding functionality that is tedious to implement manually, and so users resort to using `cardano-serialization-lib` or its close alternative, `cardano-multiplatform-lib`, which both require loading a WebAssembly blob of >1M in size.
+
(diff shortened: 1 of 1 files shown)


CIP-XXXX/README.md
```diff
@@ -0,0 +1,230 @@
...
+CIP: ?
+Title: Extensive Light Wallet Standard
+Category: Wallets
+Status: Proposed
+Authors:
+    - Vladimir Kalnitsky <vladimir@mlabs.city>
+Implementors: []
+Discussions:
+    - https://github.com/cardano-foundation/cips/pulls/620
+    - [Cardano Working Groups Discord server](https://discord.gg/XFjAzsQyaM)
```
rphair, line 11: The YAML section here doesn't understand Markdown.  The invite link should be OK on its own because people following it will have no trouble finding the `#wallet-standard` channel.

rphair (commented):
Looks like a highly useful proposal 👍

CIP-XXXX/README.md
```diff
@@ -0,0 +1,230 @@
...
+
+However, we also don't aim at providing a more *restrictive* interface. Providing a single high-level interface would not be enough for users with extra requirements: such a standard would be a regression from CIP-30.
+
+#### Statelessness
+
+All APIs must be stateless for ease of debugging: assuming on-chain state is constant, all API endpoints must return the same responses regardless of any previous API calls, with the only exception being rate limiting errors.
+
+## Specification
+
+### Query layer
```
Ryun1, line 97: Should wallets be treated as a all-knowing query layer for Cardano?
Do wallet providers want to support this? - could it be expensive for them?

Ryun1 (commented):
This is a very interesting proposal and something I am very keen to follow the development on. 😎
This has spurred me on to continue working on #619, I think that CPS and this proposal will work well being developed side-by-side.

I have added a couple editorial comments. 
A few larger questions I have:
- Does this _need_ to extend CIP-30? 
  - What was the motivation to use the CIP30 extension mechanism?
  - As to me this is close to making CIP30 unneeded, so I think we could consider alternate routes to extending CIP30, such as [CIP90](https://github.com/cardano-foundation/CIPs/pull/462/).
- Is it really the wallet's role to be able to query many aspects of Cardano?
  - To me it makes more sense to split the large `.query` endpoint into separate standards so wallets can choose what things they support querying

CIP-XXXX/README.md
```diff
@@ -0,0 +1,230 @@
+---
+CIP: ?
+Title: Extensive Light Wallet Standard
```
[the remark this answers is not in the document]
mkazlauskas: Also, I think this shouldn't be constrained by *light* wallets - why not just any wallet?

CIP-XXXX/README.md
mkazlauskas: :100:, maybe part of the query layer that allows querying arbitrary (not just user's wallet-related) data such as any utxo or transaction could be a separate extension/cip. Not sure how to define the boundary here.

CIP-XXXX/README.md
```diff
@@ -0,0 +1,230 @@
...
+  * PubKey
+- Register stake
+  * ScriptHash
+  * PubKey
+- Deregister stake scripts / pubkeys
+  * Native Script
+  * Plutus Script
+  * PubKey
+
+### JsonTx API
```
mkazlauskas, line 198: It would be great to specify this as a separate CIP, so that it could be maintained independently of dapp connector and other CIPs could reference it. @KtorZ seems to have done some work on this already https://github.com/CardanoSolutions/cardanonical

CIP-XXXX/README.md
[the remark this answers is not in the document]
rphair: There may be even a better designation than "light" because the incorporation of a query layer would make the wallet quite a bit heavier. 🤔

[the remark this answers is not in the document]
rphair: As per Discord there are better terms evolving for the different classes of wallets: https://discord.com/channels/971785110770831360/992011119872970762/1176547955919044680

(not quoting here yet since these terms are still in flux according to "Workshop" on wallet interfaces scheduled in the coming day... if finalised @klntsky @Ryun1 @mkazlauskas we should copy them here)

CIP-XXXX/README.md
klntsky: > Should wallets be treated as a all-knowing query layer for Cardano?
> Do wallet providers want to support this? - could it be expensive for them?

@Ryun1 I addressed these concerns in my new CPS: https://github.com/cardano-foundation/CIPs/pull/625

klntsky:
@Ryun1 

> Does this *need* to extend CIP-30?

No, there's no technical requirement.

> What was the motivation to use the CIP30 extension mechanism?

I didn't know about CIP-90 at the time. If the APIs are compatible with CIP-30, then it's better to switch to referencing CIP-90 instead, because it would free us from CIP-30 as dependency.

CIP-XXXX/README.md
klntsky: > Should wallets be treated as a all-knowing query layer for Cardano?

Standardizing the query layers would allow wallets to be free from the burden of running their own infrastructure. Thus it would not be a problem to force the wallets to provide query layers - they will be just passing data from JS to query layer provider's HTTP interface, which does not require any infra.

klntsky:
Screenshot from 2023-11-30 15-30-22

Let's put this on hold until we figure out the dependencies - at least as draft CIPs outlining scope.

CIP-XXXX/README.md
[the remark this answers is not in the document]
rphair: ... copying these operating definitions here for public consideration in the meantime, since haven't seen it transcribed anywhere else on GitHub or changed yet (thanks to @yHSJ for the formulation):
- **Zero Data Wallet**: This wallet has no notion of the blockchain, it is simply an interface for cryptographic operations
- **Personal Data Wallet**: This wallet does provide some data from the blockchain, but it is limited to only the data that is relevant to the key pairs controlled by the wallet
- **Full Data Wallet**: This wallet has complete view of the blockchain and can do any querying that a dApp would do

rphair:
@klntsky reflecting your https://github.com/cardano-foundation/CIPs/pull/620#issuecomment-1833693927 I'm finally marking this `Waiting for Author` - the next time we look at PRs that haven't been going anywhere, it would move progressively to (potentially) `Abandoned` and then closed.

But the dependencies you mention above are mostly merged... I would rather continue active review than close it or leave it as a `Draft` forever.  Can you please give us an update ASAP about how this problem looks today & what exactly we are still waiting for?

rphair:
@klntsky this "draft" has been open for nearly a year so I believe it could be effectively abandoned in its current form.  If not, please do something to finalise this draft for review or withdraw it by closing this PR.  As I recall the community was responsive to this idea so personally I hope you will continue work on it.

rphair:
@klntsky closing this due to abandonment as per https://github.com/cardano-foundation/CIPs/pull/620#issuecomment-2369955696.

PR #622, opened 2023-11-20
CIP-1694 | Guard security-relevant protocol parameter changes behind SPO votes

WhatisRT:
These changes came out of recent discussions with security researchers. Guarding certain parameters behind SPO votes results in preserving the current security and liveness properties of the system without having to make further security assumptions.

This is WIP for now, since I'm not sure yet whether the list of parameters guarded behind this extra condition is final.


CIP-1694/README.md
```diff
@@ -1292,6 +1292,7 @@ The following table details the ratification requirements for each governance ac
 | 5<sub>b</sub>. Protocol parameter changes, economic group         | ✓  | $P_{5b}$ | \-       |
 | 5<sub>c</sub>. Protocol parameter changes, technical group        | ✓  | $P_{5c}$ | \-       |
 | 5<sub>d</sub>. Protocol parameter changes, governance group       | ✓  | $P_{5d}$ | \-       |
+| 5<sub>e</sub>. Protocol parameter changes, security group         | \- | \-       | $Q_{5e}$ |
```
Ryun1, line 1295:
```suggestion
| 5<sub>e</sub>. Protocol parameter changes, security group         | ✓ | $P_{5e}$  | $Q_{5e}$ |
```
Does CC and DRep vote not count at all for these parameters?

lehins: This security group is exclusive to SPOs. DReps and CC votes will count, because the same parameters will belong to different DRep's groups and all parameter changes need full CC agreement.

In other words: 
* For DReps: `allPParams == union [network group, economic group, technical group, governance group]`
* For SPOs: `isSubsetOf (security group) allPParams`

Ryun1: Ahh understood, the parameters are in multiple groups, that was not clear to me.
Might want to update the table somehow to reflect this as it is misleading at the moment.

lehins: You don't think it is enough to actually look at the params in the group, there is list below in this PR:

The **security group** consists of:
* `maxBBSize`
* `maxTxSize`
* `maxBHSize`
* `maxValSize`
* `maxBlockExUnits`
* `minFeeA`
* `minFeeB`
* `coinsPerUTxOByte`
* `govActionDeposit`

WhatisRT: There was also a paragraph explaining that membership in the security group does not prevent membership in other groups. But since no one read that & everyone got confused instead, it's now called 'security relevant protocol parameters'. There's no difference and the implementation will be the same, but hopefully it'll be less confusing.

lehins:
We'll need to add this new parameter to the list of security params: https://github.com/IntersectMBO/cardano-ledger/issues/3952

Ryun1:
Whilst we are editing CIP-1694, would it be possible to add a note about script-based constitutional committee credentials?
As these made it into the ledger design post the CIP, and I have found that most people are unaware of these credentials because of this.

WhatisRT:
I've added that CC members can be scripts, that the new PParam `minFeeRefScriptsCoinsPerByte` is security-relevant and fixed the outdated claim that we go into no-confidence when we don't have enough active CC members.

CIP-1694/README.md
```diff
@@ -1295,11 +1303,29 @@ The following table details the ratification requirements for each governance ac
...
+* `maxTxSize`
+* `maxBHSize`
+* `maxValSize`
+* `maxBlockExUnits`
+* `minFeeA`
+* `minFeeB`
+* `coinsPerUTxOByte`
+* `govActionDeposit`
+* `minFeeRefScriptsCoinsPerByte`
+
```
Hornan7, line 1328: If this can be done at the protocol level, you should add `dvtMotionNoConfidence` and `pvtMotionNoConfidence` to this list as well and remove the CC rights to vote on these 2 parameters that were designed originaly to remove them.

They already cannot vote on "Motion of no confidence" and "Update Committee" governance actions, for obvious reasons. Which is why I think, their right to vote on these 2 parameters should be given to SPOs instead in my opinion. 😄👍

Hornan7 (approved):

Ryun1:
I have forwarded this PR to the Intersect governance parameter working group for comment :)

Hornan7:
The only thing I would add is that, having already the CC vote and Dreps votes on `govActionDeposit`. I don't think adding SPOs votes would be nessesairy. Unless we end up with no Committee at all.

But in that case, if you see this as a security issue, why all other deposits wouldn't be equally at risk? If risk there is. (`dRepdeposit`, `poolDeposit`, `keyDeposit`)
Because technicaly speaking, if the risk of having `govActionDeposit` not being protected by SPO would add the possibility of someone raising it to stall governance on Cardano, what would hold em to do the same with DReps deposit, Stake pool deposit, stake key deposit and hold anyone else from registering as a DRep, SPOs or their own stake credentials?

I can understand the importance of adding difficulty layers to all other ones though (Transaction related ones), but i really don't think govActionDeposit is a big of a deal.

CIP-1694/README.md
lehins: Unfortunately that can't happen due to technical reasons, since `dvtMotionNoConfidence` and `dvtMotionNoConfidence` are not individual protocol parameters, but instead they are thresholds that are part of a threshold set, as such they can't be voted on individually. In other words, whenever a proposal is submitted that updates Stake Pool or DRep thresholds, all of the thresholds must be provided.

Moreover, drastic changes to the CIP are not possible at this stage.

The formal specification is complete and undergoing the audit at this point. Any changes right now will cause a significant delay to the release.

That being said, if you believe such functionality is very important, feel free to submit a CIP for the next era.

WhatisRT:
This list of security-relevant parameters comes from multiple sessions with security researchers where we analyzed how the chain might break if certain parameters were set to bad values. There were some strict criteria for a parameter to be included, one of which being unrecoverability. `govActionDeposit` is unrecoverable if set too high: set it to a number over the total amount of lovelace in the system and you can never propose a governance action again. Other deposits don't have this problem, and they don't affect anyone who isn't already registered. There are issues with raising these too high, but the change is reversible.

While this may sound unrealistic, the point of the security-relevant parameters is to cover all possible attacks that satisfy these criteria. You only need to trust the SPOs to trust that the chain will not suffer from these attacks, no matter what happens with DReps & the CC.

CIP-1694/README.md
thenic95: Thank you for this explanation @lehins 
Would you happen to have more information on the mentioned "threshold set" and what they contain?
Are the sets the different governance action thresholds for DReps and SPOs?

lehins: There are two separate protocol parameters one for [DReps](https://github.com/IntersectMBO/cardano-ledger/blob/e8f225c6a82ff02be63d5208fa72d11aee0e5999/eras/conway/impl/src/Cardano/Ledger/Conway/PParams.hs#L264-L274) and another for [Stake Pools](https://github.com/IntersectMBO/cardano-ledger/blob/e8f225c6a82ff02be63d5208fa72d11aee0e5999/eras/conway/impl/src/Cardano/Ledger/Conway/PParams.hs#L198-L204). Both of which contain all the relevant thresholds.

Hornan7: > Thank you for this explanation @lehins 
> Would you happen to have more information on the mentioned "threshold set" and what they contain?
> Are the sets the different governance action thresholds for DReps and SPOs?

Ryun1 (approved):

PR #623, opened 2023-11-21
CIP-???? | Web-Wallet Bridge - Abstraction Layer for DApps

Fell-x27:
Original document: https://github.com/Fell-x27/CIPs/tree/master/CIP-Abstraction-Layer-for-DApps

**Title:** Dapp Bridge Rework

**Goal:** Develop a new standard for the integration of wallets and DApps, addressing a number of existing issues (details inside the document).

**Brief Description:** This CIP proposes a new mechanism for integrating wallets and DApps:
* Clearly separating the roles of DApp and wallet in this interaction.
* Creating an abstraction layer that eases the interaction for both parties.
* Enhancing user security.
* Reducing the impact of human factors by decreasing the number of potential failure points.

This is achieved by transforming the DApp into a client of the wallet, rather than a substitute for it.


Fell-x27:
Now, this mechanism supports not only DataHash for Movement, but also inline Datum.

rphair:
@Fell-x27 I am still trying to understand this material as well as your presentations in the Discord group which I may be helping to put it into context with the other wallet interface improvement initiatives.  I intend to catch up on all this before next week's meetings...

From your activity on Discord I'm guessing you're interested in attending first the [Wallet Connectors Workshop](https://discord.com/events/971785110770831360/1176916387046887535) (27 November 4PM UTC) and then the usual CIP meeting 24 hours after?  I believe the CIP meeting would be a good time for a quick introduction to your proposal and to review the community's initial reaction to these ideas, so it's on the agenda here (https://hackmd.io/@cip-editors/77) cc @Ryun1 @Crypto2099

CIP-dapp-bridge-rework/LICENSE
```diff
@@ -0,0 +1,395 @@
+Attribution 4.0 International
```
rphair, line 1: As also acceptable by commonly used standard for the Creative Commons license etc., and according to CIP practice (https://github.com/cardano-foundation/CIPs/blob/master/CIP-0001/README.md#licensing), this file should be eliminated and referred to by a link at the end of the CIP document.  The way you currently have it in the CIP body is proper & sufficient. 😎

CIP-dapp-bridge-rework/README.md
```diff
@@ -0,0 +1,1169 @@
...
+Category: Wallets
+Authors:
+  - Fell-x27
+Implementors: N/A
+Discussions: N/A
+Created: 2023-11-21
+License: CC-BY-4.0
+---
+
+<!-- TOC -->
```
rphair, line 14: CIP site builders won't be able to parse a Table of Contents here and in fact will create a duplicate.  For this reason it's most portable for CIPs to rely on these site builders to build the table of contents rather than include it in hardcopy... which can get inconsistent with edited headers very easily, especially when other authors submit PRs based on your material.

We have a hardcoded TOC in CIP-0001 (yes it's gotten out of sync several times & I've had to correct it myself a couple times) and you can see how kind of silly it looks here in the generated site: https://cips.cardano.org/cips/cip1/

If you continue to mimic CIP-0030 you can see that the generated TOCs on whatever site builder will contain certain heading levels already; currently:
- [inline: all API headings] https://cips.cardano.org/cips/cip30/
- [sidebar: section headings only] https://developers.cardano.org/docs/governance/cardano-improvement-proposals/CIP-0030

I am hoping @Ryun1 @Crypto2099 @KtorZ will corroborate this but I think the most portable way therefore is to consider the TOC a matter for the presentation layer, not the CIP content itself.

CIP-dapp-bridge-rework/README.md
```diff
@@ -0,0 +1,1169 @@
...
+{
+    certificates: [{
+        type: "stakeDelegation",
+        stakeKeyId: 0,
+        poolHash: "660f2b816e90e4e826daebf33357d76ab89414f88d57dd7d9f7898fe"
+    }]
+}
+```
+
+
```
rphair, line 138:
```suggestion

```
Parts of the document have a loose spacing which doesn't work well between the number of headings & length of the document.  (Some of the sections with shorter content might even work better as bulleted lists or tables.)

This is completely optional & bottom priority for review, especially considering the weighty matters of scope & purpose currently under debate.

CIP-dapp-bridge-rework/README.md
```diff
@@ -0,0 +1,1169 @@
...
+    * For example, `inputs: "auto"` indicates that the wallet should take care of having transaction inputs that cover its tasks;
+* Typically, the "auto" value can be found in optional fields, and this is not accidental:
+    * "auto" is the default value;
+    * If any value was omitted, the wallet should treat it as if it were equal to "auto";
+
+```ts
+    type Auto = "auto";
+```
+
+#### `cbor<T>`
```
rphair, line 159: note I think it's very helpful here that you've preemptively code-quoted these headings to make sure the angle brackets don't cause formatting problems, as we originally saw with CIP-0030.

CIP-dapp-bridge-rework/README.md
```diff
@@ -0,0 +1,1169 @@
...
+
+  Contains the [hexadecimal](#hexstring) representation of the asset's `policyID`.<br><br>
+
+* **`name`**
+
+  Contains the [hexadecimal](#hexstring) representation of the asset's `name`. If absent, the value should be set to `null`.<br><br>
+
+* **`amount`**
+
+  Contains the quantity of the asset [without a decimal part](#bigintstring). That is, if we want to describe 1 token with a decimal part of 6 digits, the amount should be set to 1,000,000, i.e., 1 * 10^6. Setting the value to 1 would be equivalent to 0.000001!<br><br>
```
rphair, line 255: This is an example of what I was referring to above: loose content which would be better formatted as bullet lists or tables... especially because you're not using headings that would create deep links to these entries.

This pattern is repeated several times throughout the document, and unless addressed it will affect readability by making the CIP look longer & more complicated than it really is.

rphair (commented):
@Fell-x27 thanks for elaborating your ideas in the Discord and I hope you will consider re-posting some of your relevant [writing](https://discord.com/channels/971785110770831360/1176612265894613133/1177154891295240283) and [diagrams](https://discord.com/channels/971785110770831360/1176612265894613133/1177153290555895858) into this PR, if not the CIP itself, if & when you are ready.  Whenever discussions on Discord clarify an issue raised here or present an item for review, I'll also do what I can to keep the two media & sets of ideas connected.

If this CIP stays in its current form... rather than changing the scope or document type (e.g. a new CPS, or additions to @Ryun1's https://github.com/cardano-foundation/CIPs/pull/619) based on the current discussions... there are some formatting changes to please make while you're editing this material:

CIP-dapp-bridge-rework/README.md
```diff
@@ -0,0 +1,1169 @@
...
+
+### Acceptance Criteria
+
+- [] The interface is implemented and supported by various wallet providers.
+- [] The interface is used by DApps to interact with wallet providers.
+
+### Implementation Plan
+
+- [] Provide some reference implementation of wallet providers
+- [] Provide some reference implementation of the dapp connector
```
rphair, line 1164:
```suggestion
- [ ] The interface is implemented and supported by various wallet providers.
- [ ] The interface is used by DApps to interact with wallet providers.

### Implementation Plan

- [ ] Provide some reference implementation of wallet providers
- [ ] Provide some reference implementation of the dapp connector
```
addendum to last format-nitpick review: these only format properly as tick-boxes on GitHub & other Markdown previewers when either a space or `x` (maybe other characters?) is between the brackets.

Fell-x27:
@rphair , thanks for the review! I'll check this again later and implement everything!

CIP-dapp-bridge-rework/README.md
Fell-x27: That's why it has a  wrapper, which allows third parties to ignore it if needed. My IDE added it, so I thought it was a standard to resolve issues like you described. It was added to improve navigation in the large document, at least during the 'github only' stage. I also use it in my md-editor. If it's not an option to cut it off by the TOC tag, I think it would be OK if it remains here during the 'github only' stage, and then, if it is included in the CIPs list, I'll remove the TOC to avoid any conflicts with document builders. What do you think?

CIP-dapp-bridge-rework/README.md
Fell-x27: The document is not final and could be improved several times based on community feedback so I think, we can delay it until the finalization phase.

CIP-dapp-bridge-rework/README.md
Fell-x27: "Should I remove the `brs` only, or attach the content of each item to its header in one row?

CIP-dapp-bridge-rework/README.md
```diff
@@ -0,0 +1,1169 @@
...
+
+## Motivation: why is this CIP necessary?
+
+This CIP addresses several issues that complicate the development of the Cardano ecosystem with the current integration standard:
+
+1. **Complexity in Writing Client-Side Logic for DApps**: As they essentially need to implement their own wallet each time, including algorithms for coin selection, UTxO management, UTxO validation, TX building, etc. They shouldn't be doing this. There is already a separately written wallet that connects to their application for this purpose. <br><br>
+
+2. **Integration Difficulties**: This implies complicating the internal logic of wallets, adding endpoints for importing transactions, writing logic for checks/validation of foreign transactions, etc. This could be avoided if the transaction was created and processed natively by the wallet.<br><br>
+
+3. **Privacy Issues**: Currently, the wallet must pass all information to the DApp, including the UTxO set and full addresses on one hand, and trust the user's funds to the transaction building code inside the DApp on the other. This is not obvious to the user, and in case of problems, they will blame the wallet developers.<br><br>
```
klntsky, line 89: This is not a privacy problem, because there is no way to issue a transaction without revealing your address anyway

klntsky: Wallets do not reveal addresses from other accounts

CIP-dapp-bridge-rework/README.md
```diff
@@ -0,0 +1,1169 @@
...
+
+4. **Current Role of Wallets**: At present, the wallet is essentially just a synchronization point for the client side of the DApp with the blockchain. Yes, it generates user signatures, but, for example, if we use a HW-wallet, then the task of the SW-wallet is reduced to 'submit tx'. This is an incorrect design of communication.<br><br>
+
+5. **Scalable Complexity**: Currently, the integration standard implies that the complexity of the solution on the DApp side **must be** unjustifiably high even for elementary things that do not require knowledge of CBOR, CDDL, or an understanding of the difference between bech32 and base58.<br><br>
+
+6. **Reducing Dependencies**: Currently, all DApps developers **must** operate only with CBOR-CDDL entities, making them dependent on third-party libraries such as `cardano_serialization_lib`, and forcing them to integrate these libraries into their projects. This complicates the logic, increases the size of the loaded page, necessitates keeping up with updates, etc., thereby increasing the number of failure factors in the application's logic.<br><br>
```
klntsky, line 95: This CIP still uses CBOR and does not provide an abstraction from data layout because of that

CIP-dapp-bridge-rework/README.md
```diff
@@ -0,0 +1,1169 @@
...
+
+7. **Lack of abstraction**: The Cardano ecosystem is actively developing, especially in everything related to Plutus. DApps struggle to keep up with the innovations. Because of this, we encounter a heavy inconsistency in approaches. For example, this concerns everything related to collaterals at the moment. Cardano has long allowed making their use transparent and convenient for the user, but not all DApps support these approaches, forcing wallets to "reserve collateral UTxO" and somehow integrate this into UX. Moreover, the very fact that the DApp operates the collateral, and not the wallet (which only offers collateral but without guarantees that it will be used), makes the user vulnerable.<br><br>
```
klntsky, line 97: > makes the user vulnerable

No it does not, phase-2 validity is still checked by the wallet.

CIP-dapp-bridge-rework/README.md
```diff
@@ -0,0 +1,1169 @@
...
+
+8. **Security**: Currently, DApps are given too much responsibility in the formation of transactions. This could potentially become an attack vector on a user who is confident in their security. Besides the risk of leaking wallet data, including addresses, there is a possibility that a DApp could form a malware transaction. This transaction might include fields not checked by the wallet but still obtain all necessary signatures. For example, the user could unknowingly become part of a criminal scheme or lose funds.<br><br>
```
klntsky, line 99: > the user could unknowingly become part of a criminal scheme

Not our concern

> or lose funds

Balance change is always shown upon confirmation

CIP-dapp-bridge-rework/README.md
```diff
@@ -0,0 +1,1169 @@
...
+
+9. **Trustful system**: While blockchain is inherently a trustless system, it seems very odd when a wallet unconditionally trusts all user data to a third party at the first request.<br><br>
```
klntsky, line 101: The user trusts their data, not the wallet. That's what confirmation popups on first connection are for

CIP-dapp-bridge-rework/README.md
```diff
@@ -0,0 +1,1169 @@
...
+
+By solving these problems, we will lower the entry barrier for DApp developers into the ecosystem, while simultaneously enhancing the security of the end-user who has entrusted their funds to a wallet, but should not trust their management to just any application they encounter. The wallet will act as a black box, and the DApp as an extension of its UI, not a replacement of its engine.
+
+Additionally, this approach will simplify the integration of wallets and DApps from the wallet developers' perspective, as it will provide a more natural interface for interaction, easily translatable into native scenarios of interaction with the ordinary user through the wallet's own UI.
+
+## Specification
+
+### The main idea
+The main idea is to delegate as many operations as possible to the wallet, which are its core functions. This will greatly simplify the lives of DApps developers. It will also normalize the transaction pipeline, which is currently highly intermingled between the DApp and the wallet.
```
klntsky, line 110: > This will greatly simplify the lives of DApps developers

It will only simplify the lives of developers who want to operate on a high level of abstraction

CIP-dapp-bridge-rework/README.md
```diff
@@ -0,0 +1,1169 @@
...
+We are moving away from imperative interaction with the wallet towards a declarative approach, where the DApp does not have to assemble the transaction itself; it should be sufficient to **describe** the desired outcome. Just as a regular wallet user is not required to delve into all the details - they simply specify the address, amount, and press a button. This is literally why software wallets exist.
```
klntsky, line 111: > it should be sufficient to **describe** the desired outcome

No it is not for many developers

CIP-dapp-bridge-rework/README.md
```diff
@@ -0,0 +1,1169 @@
...
+#### `Auto`
+
+Technically, it's not a type, but just a string constant "auto" passed instead of any specific value.
+
+* "auto" can only be used where it is explicitly stated, for example, if a value can be `Movement[] | Auto`;
+* When we pass "auto" instead of something, we imply that the wallet should take care of substituting the correct value on its own:
+    * For example, `inputs: "auto"` indicates that the wallet should take care of having transaction inputs that cover its tasks;
+* Typically, the "auto" value can be found in optional fields, and this is not accidental:
+    * "auto" is the default value;
+    * If any value was omitted, the wallet should treat it as if it were equal to "auto";
```
klntsky, line 153: There is no need for introducing this value, it's better to design the constraints interface properly to cover all cases.

CIP-dapp-bridge-rework/README.md
```diff
@@ -0,0 +1,1169 @@
...
+
+A string indicating whether the account belongs to the `Byron` or `Shelley` eras of Cardano.
+
+```ts
+type AccountEra = "Byron" | "Shelley";
+```
+
+#### `AccountId`
+
+An arbitrary string that serves as the identifier of the wallet account connected to the DApp, formed by the wallet according to its own rules. The main point is that the wallet understands what is required of it when it receives this identifier in a request.
```
klntsky, line 226: There is no notion of account on Cardano. If you mean account id in the derivation path, this is unnecessarily restrictive. There isn't even a guarantee that an address does not have scripthash components.

Also, consider that address indices also exist.

https://cips.cardano.org/cips/cip1852/

CIP-dapp-bridge-rework/README.md
rphair: I think that's OK as long as we remember to delete the TOC before any merging, or at least if/when this proposal is in Last Check and after you're sure it's no longer going through your IDE.  still I'll leave this "unresolved" in the meantime so other editors & reviewers can be aware of the issue & help remember this if/when merge time comes.

SIDANWhatever:
First of all, I really like the idea of streamlining transaction building process specifically. Some brief comments:

1. On "building" transaction from an object, there is sample implementation on Mesh's latest lower level APIs (https://meshjs.dev/apis/transaction/builderExample). Maybe I would invite discussion based on it, whether we could rely on middle layer libraries like Mesh to handle this logic or putting this logic on wallet side (a concern is that as Adam mentioned it might add extra burden to wallet side).

2. On the object itself, if we go for standardization, I think it is better to following the native object format (like using validity range as one item rather than splitting as `from` and `to` separately) to make the full compatibility, an example object for tx body from aiken's doc:

```rs
Transaction {
  inputs: List<Input>,
  reference_inputs: List<Input>,
  outputs: List<Output>,
  fee: Value,
  mint: MintedValue,
  certificates: List<Certificate>,
  withdrawals: Dict<StakeCredential, Int>,
  validity_range: ValidityRange,
  extra_signatories: List<Hash<Blake2b_224, VerificationKey>>,
  redeemers: Dict<ScriptPurpose, Redeemer>,
  datums: Dict<Hash<Blake2b_256, Data>, Data>,
  id: TransactionId,
}
```

CIP-dapp-bridge-rework/README.md
Fell-x27: Yes. [This CIP is not about "getting rid of CBOR".](https://github.com/Fell-x27/CIPs/blob/c3f2dfd84447176403458ac0cbd810ed6f44f8d7/CIP-dapp-bridge-rework/README.md?plain=1#L77-L79). Actually we have to use CBOR for some cases and it's okay. 

This CIP is about **complexity reducing and automation**. So, if you don't need a CBOR, don't use it. If you want to produce some signatures, or finalize a transaction by yourself, then you are already working with CBOR data and it's not a problem for you.  But you still don't have to write your own "semi-wallet" to get your DApp working.

If you're already working with CBOR on the DApp side, trying to convert it to JSON in some way will only add new layers of complexity, instead of solving any problems.

CIP-dapp-bridge-rework/README.md
Fell-x27: No, it's a privacy problem, because the 3rd party software gets all my data. They have my addresses (even for an enterprise account with "not bunchable" addresses because of lack of a stake part). They have my UTxO set. Doesn't matter if I use a new address for every new transaction, they have it all. They can track everything they want then. And I can't control it with the CIP30. I can't prevent it. And I can't hide it then, but they can share it to everyone. So it's almost the same as if I had posted my public keys on Twitter.

CIP-dapp-bridge-rework/README.md
Fell-x27: In a perfect world everything is perfect. But as a software engineer I can say that in a real world we have to deal with imperfect software which can work in imperfect way. With each new link in the chain of responsibility, new potential points of failure arise. Each new link increases the number of potential attack vectors if there is lack of checking. The goal of this CIP is to make all operations for the wallet as **explicit as possible**. The wallet won't have to recheck every field of the transaction passed to it from the outside, if it assembled the transaction itself and processed every aspect of it 100%.

In this case, even in an imperfect world, we will reduce the likelihood of problems.

CIP-dapp-bridge-rework/README.md
Fell-x27: >Not our concern

No, it is our concern. At least mine, if you prefer. If we are talking about user security, and if we aim for "mass adoption" of Cardano, we must acknowledge that the convenience of software is **_inversely proportional_** to the entry barrier into the ecosystem. And we **are obliged** to protect the average user from things that could happen to them if the wallet can be attacked or compromised in the manner described in the CIP. As a wallet developer, I can say that user safety and protecting their interests is **the main concern**. And if we can **_passively_** solve this problem simply at the API level, we should do so.


>Balance change is always shown upon confirmation

See [link](https://github.com/cardano-foundation/CIPs/pull/623#discussion_r1408157914)

CIP-dapp-bridge-rework/README.md
Fell-x27: See the [link](https://github.com/cardano-foundation/CIPs/pull/623#discussion_r1408146806)

CIP-dapp-bridge-rework/README.md
Fell-x27: Not only that. Even if you want to work with a lower level of abstraction, this CIP brings automation to your life. But you can still operate at a low level. You can request a draft of the transaction from the wallet for your `TransactionScript`, and then use its inputs, outputs, whatever, for your purpose to build another `TransactionScript`. You can even merge many transactions into one. You used the automation at the first stage, but then you can build a low-level `TransactionScript` if you wish. But you still don't have to write your own "semi-wallet". So even if you are a low-level developer, your life becomes easier.

That's what I was talking about - the scalability of complexity. Or "adaptive complexity". It's as high/low level as you need.

CIP-dapp-bridge-rework/README.md
Fell-x27: That is why there is only "should". Anyway, the standard allows you to describe anything you want.

CIP-dapp-bridge-rework/README.md
Fell-x27: I wouldn't do this if it weren't necessary and didn't form part of the concept of **this CIP**, which reflects my ideas and considerations as a developer. Moreover, almost everywhere a field that can take the value Auto can simply be omitted. Explicitly stating Auto is only needed if it's more convenient for the DApp developer.

CIP-dapp-bridge-rework/README.md
rphair: :+1: I think without the `<br>` elements Markup wiil create a white spacing consistent with other documents.... in fact in hindsight that's what was creating the extra line spacing rather than the carriage returns as I suggested above (so resolving that one).

Yes I'd also look to see how it looks as a table with a row for each parameter, if you're also interested, which I think would make it a better match for other CIPs & computer documentation in general.

CIP-dapp-bridge-rework/README.md
rphair: resolving as per https://github.com/cardano-foundation/CIPs/pull/623#discussion_r1408201380 in favour of a denser markup format (e.g. lists or tables), so we'll keep that one open instead.

CIP-dapp-bridge-rework/README.md
Fell-x27: No I mean:

>An arbitrary string that serves as the identifier of the wallet account connected to the DApp, formed by the wallet according to its own rules. The main point is that the wallet understands what is required of it when it receives this identifier in a request.

It's just an internal identifier. It can be anything that a specific wallet can recognize.

Fell-x27:
@SIDANWhatever , hi!

>a concern is that as Adam mentioned it might add extra burden to wallet side

Actually, it's not a problem. The thing is, the wallet has likely already implemented a similar mechanism. When we interact with it through the GUI, we don't directly form the transaction. We create some sort of transaction object/description based on input fields, using which we assemble the real one. Essentially, my CIP proposes using the same logic for interaction. So, the only thing needed is to write a "translator" from `TransactionScript` to a format the wallet understands. And simply extend the logic that is already working.

>On the object itself, if we go for standardization, I think it is better to following the native object format 

Yes, initially I thought about that, but then decided to move away from it. My idea was to lower the entry barrier and add a layer of abstraction over the existing terms so that the result would be as intuitive as possible for someone "outside the ecosystem". And, if possible, self-documenting. That's why the structure of `TransactionScript` and its subtypes deliberately do not replicate CDDL. Trying to bring in a more declarative approach. But yes, it can be changed later if we decide that it was the wrong way. It's not a big deal :)

CIP-dapp-bridge-rework/README.md
Fell-x27: >Yes I'd also look to see how it looks as a table with a row for each parameter, if you're also interested, which I think would make it a better match for other CIPs & computer documentation in general.

I'll check this approach. But there are many variations of docs styles. My reference in most cases was the [MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object).

CIP-dapp-bridge-rework/README.md
```diff
@@ -0,0 +1,1169 @@
+---
+CIP: ????
+Title: Cardano dApp-Wallet Web Bridge
```
rphair, line 3: Therefore we would not really be "reworking" anything (hence we need a new title for this PR) and the CIP would be named to reflect this new serialization method: particularly, naming by the _characteristics_ of your proposed Tx interface rather than any name you would give to e.g. a support library that @Crypto2099 was suggesting at today's meeting.

rphair (commented):
@Fell-x27 the sentiment, as best I understand it, from the CIP meeting today is that this method & CIP would be well suited to a serialization library that will give wallets a well-defined standard interface to offer these services to dApps.

The impact to existing trade practice of changing CIP-30 itself, and obligating all wallet providers to support this interface, would be too much of a burden on Cardano's ecosystem at this developmental stage.

But I think we all agree that there's potential here, enough to promote implementation as an additional library with a CIP here in support of its API which could acheive common if not universal use.

I believe this PR, and the CIP document with some additional rewriting, would be usable for that idea & resolve the editorial and community difficulty we would have with imposing these changes on CIP-30 and any future wallet interface.  cc @Ryun1 @Crypto2099

Fell-x27:
@rphair 

> this method & CIP would be well suited to a serialization library that will give wallets a well-defined standard interface to offer these services to dApps. The impact to existing trade practice of changing CIP-30 itself, and obligating all wallet providers to support this interface, would be too much of a burden on Cardano's ecosystem at this developmental stage.

No, this is not a serialization library interface. This is an interface that allows a DApp to interact with a wallet in almost the same way as a user does. In fact, this standard was designed with the idea that the implementation of the API should be as painless as possible for wallets, which bear the brunt of the standard. Because it should be similar to how they already work internally. As I already replied to @SIDANWhatever, this standard **_should not add new entities but expand existing ones_**.

As for DApps, there are not so many of them at the moment that it would be a problem. But with this standard, writing DApps will become easier and there will be more of them. This is the main goal of this CIP. Focused on future projects in the ecosystem.

Regarding the need for changes - they are inevitable in any case. If there is a need in the ecosystem to replace CIP30, whatever replaces it will require changes from the ecosystem. Whatever it is. It's already painful now. But still not as painful as, say, in a year.

The ecosystem will need a transition period anyway, during which both the new standard and the old, but already deprecated, will have to be supported.

Moreover, this is not the final implementation. I've already received feedback that it might be more appropriate to implement this standard as a CIP30-like interface, to simplify integration. Or even as a high-level extension of CIP30. I'm still thinking about it.

CIP-dapp-bridge-rework/README.md
Fell-x27: But it's not a serialization method. It's a "dapp bridge with adaptive level of abstraction", which is quite long for a name, I think. Any suggestions?

rphair: my apologies for not focusing on the part of your proposal that is truly distinctive... what about:
- _dApp-to-Wallet Abstraction Layer_ - if it ends up as a stand-alone method
- _Web-Wallet Bridge - Abstraction Layer for dApps_ - if it becomes an extention to CIP-30 as possible per https://github.com/cardano-foundation/CIPs/pull/623#issuecomment-1830623796

rphair:
@Fell-x27 since editors never marked this as `Waiting for Author` in response to your https://github.com/cardano-foundation/CIPs/pull/623#issuecomment-1830623796, specifically:
> this is not the final implementation. I've already received feedback that it might be more appropriate to implement this standard as a CIP30-like interface, to simplify integration. Or even as a high-level extension of CIP30. I'm still thinking about it.

... I'm applying the (potentially) `Abandoned` tag that we are putting on PRs that have not progressed in a long time (8 months in this case).  If you are still "thinking about" the problem then please indicate here how you want to progress and what exactly we are waiting for.  Without a clear idea of your deliberations, this does appear "abandoned" by default & would eventually be closed.

Fell-x27:
> @Fell-x27 since editors never marked this as `Waiting for Author` in response to your [#623 (comment)](https://github.com/cardano-foundation/CIPs/pull/623#issuecomment-1830623796), specifically:
> 
> > this is not the final implementation. I've already received feedback that it might be more appropriate to implement this standard as a CIP30-like interface, to simplify integration. Or even as a high-level extension of CIP30. I'm still thinking about it.
> 
> ... I'm applying the (potentially) `Abandoned` tag that we are putting on PRs that have not progressed in a long time (8 months in this case). If you are still "thinking about" the problem then please indicate here how you want to progress and what exactly we are waiting for. Without a clear idea of your deliberations, this does appear "abandoned" by default & would eventually be closed.

Hi, I've become a father, and it's harder to find time for everything. I'll be back as soon as possible.

Fell-x27:
Overall, the standard is quite ready, but I need time to implement the reference solution.

PR #628, opened 2023-11-30
Initial draft

AngelCastilloB:
CIP for multi-stake delegation from a single account.

[Human-readable version](https://github.com/AngelCastilloB/CIPs/blob/multistake-delegation-from-single-account/CIP-XXXX/README.MD).


yHSJ:
Thanks for submitting this! I left the following feedback in the [Lace Discord](https://discord.com/channels/1073034511664824360/1174398420006477864/1179795276303712377), but felt I should share it here too:

> My initial feedback would be that you should use [RFC-2119](https://datatracker.ietf.org/doc/html/rfc2119) to strictly define the language use, and then modify the specification to use that language. This is feedback I've given to almost every CIP. Cardano's standards right now use a lot of ambiguous language that one could argue is implicitly defined, but as an implementer I've seen several different interpretations of standards. If RFC-2119 was used, that ambiguity is removed.

CIP-XXXX/README.MD
```diff
@@ -0,0 +1,146 @@
...
+
+### Multi-Delegating
+- **Stake Key Derivation:** Begin by deriving the required number of stake keys needed for multi-delegation. While this derivation does not have to be sequential, it is recommended for clarity and orderly management. Ensure that the gap between any two consecutive active stake addresses does not exceed 20, in line with the [BIP-44](https://github.com/bitcoin/bips/blob/master/bip-0044.mediawiki) address gap limit. 
+
+- **Key Registration:** Once derived, register these stake keys. This registration process remains consistent with current procedures, with the added context that all these new stake keys will be associated with a common payment key.
+
+- **Fund Distribution:** Distribute the funds across addresses formed by pairing the first  payment key with each of the newly created stake keys. This creates an association between all stake delegations. The details of how funds are distributed are specified by the user when creating the multi delegation transaction.
+
+Wallet software can then easily discover all stake keys and associated delegations either by sequential discovery or identifying addresses  containing the first payment credential.
```
yHSJ, line 64: > identifying addresses containing the first payment credential

This method of identification could incorrectly identify some stake credentials as belonging to the account of the first payment credential. Bob could create a mangled address with his own stake credentials and Alice's first payment credential. Bob could use this to make it appear that Alice has chosen to delegate some stake to a pool. 

Of course Alice, controlling the payment credential, could move control of the delegation to a different stake credential, but perhaps Alice isn't aware that this has happened or is inactive but staking her held ADA. Wallet software should only discover derived stake keys through discovery.

I didn't call it sequential discovery because according to this standard, the derivation does not need to be sequential, so sequential discovery could result in a subset of all derived keys.

CIP-XXXX/README.MD
```diff
@@ -0,0 +1,146 @@
...
+      "weight": 37
+    },
+    {
+      "id": "9f9bdee3e053e3102815b778db5ef8d55393f7ae83b36f906f4c3a47",
+      "weight": 25
+    }
+  ]
+}
+```
+
```
Crypto2099, line 111: I would strongly recommend that we do not allow for the use of "ticker" or any other non-cryptographically derived identifier for any on-chain data that may be consumed by third parties. This can create an ambiguity if multiple pools are registered with the same ticker.

Furthermore, for ease of implementation and consistent interoperability between various implementations I would say that we should definitively settle on either the hex pool ID representation or the bech32 representation.

CIP-XXXX/README.MD
```diff
@@ -0,0 +1,146 @@
...
+Therefore, the following key principle will guide the balancing change process for wallets that are multi-delegating:
+
+Every multi delegation transaction (creation or update) must include the delegation portfolio metadata specifying the percentage of funds they wish to delegate to each pool. This portfolio serves as a guideline for how funds should be distributed across different stake keys.
+When following transactions are built, input selection is performed as usual. The resulting change from the transaction is then evaluated for rebalancing. The structure and size of change outputs determined by the input selection algorithm are preserved.
+The change outputs will be distributed as-is across multiple addresses associated with different stake keys, without altering the outputs values. Priority is given to stake keys that are most divergent from the set preferences, thereby gradually realigning the actual distribution with the user's delegation strategy.
```
HeptaSean, line 127: Why not changing the output values? If it is a transaction created in the soverignty of the wallet app, it *can* choose the output sizes freely and can probably better redistribute if it does.

Nit-pick: If the stake is divergent in the direction of too much stake, priority should specifically *not* be given to this stake. Could maybe be worded a bit clearer.

CIP-XXXX/README.MD
```diff
@@ -0,0 +1,146 @@
...
+      "id": "9f9bdee3e053e3102815b778db5ef8d55393f7ae83b36f906f4c3a47",
+      "weight": 25
+    }
+  ]
+}
+```
+
+The inclusion of the delegation portfolio object as transaction metadata serves to reflect any updates to the user's delegation preferences on-chain. This ensures that external applications retrieving this information are always presented with the most current data, without the need to parse through multiple transactions.
+
+External applications can retrieve the most up-to-date delegation portfolio for a given wallet by scanning the blockchain for the latest transaction with the metadata label 6862 originating from the user's wallet. This transaction will contain the metadata with the current delegation portfolio, thereby affirming the wallet owner's latest delegation preferences.
```
HeptaSean, line 114: Scanning historical transactions for a specific metadatum label is rather inefficient. For wallet apps, that's a minor problem since they will typically want to present the whole transaction history to the user anyway and they only deal with one or a small set of wallets. But for dApps that are confronted with ever new wallets every moment, it might well be prohibitive.

CIP-XXXX/README.MD
```diff
@@ -0,0 +1,146 @@
...
+. Deregister stake keys
+
+**Note that all rewards calculated but not distributed will be lost.**
+
+#### Interoperability
+Live delegation portfolio might drift overtime due to:
+- Receiving funds to a single address 
+- Building transactions with software that doesn’t support multi-stake delegation (e.g. via dapp connector)
+
+In order to maintain a record of user’s delegation portfolio preferences , all actions pertaining to stake delegation, including funds transfers, stake key registration, and deregistration, must be executed within a single atomic transaction when performing an update. This transaction should also include the delegation portfolio metadata, using the reserved metadata label 6862 (hexadecimal 0x1ace), as outlined in [CIP-0017](https://github.com/cardano-foundation/CIPs/tree/master/CIP-0017) preferably without any of the optional fields.
```
HeptaSean, line 85: Maybe make even more explicit that wallet apps must ensure that the chosen pools match between the CIP-17 JSON specification and the pools actually delegated to. … and that other wallet apps managing the same account are free to correct it if one misbehaves.

Should a new delegation certificate be included for *all* stakes and pools even if there is no change in that part of the multi-delegation? Advantage would be that this ensures that all stake keys have to sign the transaction partly attenuating @yHSJ's point on “Franken” addresses above. Disadvantage would be that it bloats the transaction and makes it more expensive unnecessarily.

CIP-XXXX/README.MD
```diff
@@ -0,0 +1,146 @@
...
+
+
+### Change Balancing Algorithm 
+
+Balancing change is a crucial feature designed to maintain a user’s stake distribution, using the portfolio preferences attached to the delegation transaction. As transactions occur over time, the distribution of funds across different pools will deviate from the initial portfolio percentages as the change would otherwise be concentrated onto a single address, and therefore be delegated to a single pool.  This drift can be corrected while constructing transactions for a user who has defined a multi-delegation intent by reading the preferences from the delegation transaction metadata and distributing the change to best adhere
+
+When implementing this feature, it is essential to maintain the integrity of the change outputs as produced by the input selection algorithm. The design of these algorithms, such as the Random Improve input selection, is aimed at optimizing the overall health of the wallet's UTXO (Unspent Transaction Output) set. This optimization includes avoiding the creation of dust and ensuring that the size of outputs aligns with the user's spending habits over time.
```
HeptaSean, line 121: Is Random Improve really a good coin selection algorithm for this? Is it – in an unmodified form – even a good coin selection algorithm for wallets with native tokens?

Shouldn't it at least be modified to select UTxOs from all stakes as proportional as possible and not leave that to the randomness of the algorithm?

CIP-XXXX/README.MD
HeptaSean, line 127: The problem of NFTs and other native tokens should at leas be mentioned.

Is it maybe feasible to demand that native tokens are *not* moved between stakes in the process that they stay on the same stake address unless the user *explicitly* moves them to another stake?

CIP-XXXX/README.MD
```diff
@@ -0,0 +1,146 @@
...
+
+Therefore, the following key principle will guide the balancing change process for wallets that are multi-delegating:
+
+Every multi delegation transaction (creation or update) must include the delegation portfolio metadata specifying the percentage of funds they wish to delegate to each pool. This portfolio serves as a guideline for how funds should be distributed across different stake keys.
+When following transactions are built, input selection is performed as usual. The resulting change from the transaction is then evaluated for rebalancing. The structure and size of change outputs determined by the input selection algorithm are preserved.
```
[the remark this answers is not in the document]
yHSJ: I think this is an implementation detail. Wallets may choose to leave the distribution of UTxOs up to the user, for example.

CIP-XXXX/README.MD
rphair: yes, I also think it's important that we avoid putting pool "ticker" references on chain.  When I advocated for pool tickers in stake pool URIs, we hotly debated using tickers at all in the "delegation portfolio" links in both CIP-0013 and CIP-0017 and could only collectively justify using tickers when both
1. off-chain (leaving wallets to settle ambiguities); and
2. in user-facing cases: where brevity, human readability, and/or convenience are important: i.e not in the off-chain JSON spec of CIP-0017 which is written for precision and where brevity isn't important.

CIP-XXXX/README.MD
AngelCastilloB: This needs to be reworded. The two methods of discovery are suggested here, sequential discovery is only possible to do for wallets. But external entities such as blockchain explorers or some dApps do not have access to the keys to do address derivation, but they can do the search by key association.

So wallets could either do sequential discovery (or association by key but then will need to take the extra precaution of pruning addresses which contains credentials it doesn't control).

External entities have to rely on key association (same as they do now for multi address wallets).

So this could probably be reworded as:

```
Wallet software can then easily discover all stake keys and associated delegations either by sequential discovery.
External entities such as blockchain explorers or dApps can do the discovery by identifying addresses containing the first payment credential.
```

thoughts?

CIP-XXXX/README.MD
AngelCastilloB: Agree, it is an error, all entries should be id + weight, although the CIP just makes the suggestion to not use optional fields (such as ticker), we could make it mandatory to have this shape, id (hex pool ID) + weight

CIP-XXXX/README.MD
AngelCastilloB: Random Improve was just provided as an example. The point here would be that implementing multi-delegation shouldn't require changing the current input selection algorithm to accommodate it, the wallet could be using an input selection strategy that optimize for a given use case, ideally we dont want to disrupt that, as forcing the wallet to come up with a new input selection algorithm that satisfies his use case and multi delegation could be too problematic. Also during our internal tests, just directing the change output was sufficient to advert drift in the long run

CIP-XXXX/README.MD
AngelCastilloB: could you elaborate more on is the problem of NFTs and other native tokens?

Adding this extra requirement of allowing users to fix tokens to specific stake address would make the input selection harder

CIP-XXXX/README.MD
rhyslbw: > Maybe make even more explicit that wallet apps must ensure that the chosen pools match between the CIP-17 JSON specification and the pools actually delegated to. … and that other wallet apps managing the same account are free to correct it if one misbehaves.

I like this idea

> … partly attenuating …

Since it doesn’t fully attenuate I’d vote in favour of not enforcing unnecessary certificates.

https://github.com/cardano-foundation/CIPs/pull/628#discussion_r1411743581

AndrewWestberg:
This is basically the approach Lace wallet takes. I disagree entirely with this approach. It has the following negative consequences:

1.) Requires registering of additional stake addresses.
2.) The rewards claiming transaction becomes more complex.
3.) dApps that also build transactions outside the wallet know nothing about this approach and won't care about re-balancing.

A better approach is to simply change the delegation certificate transaction to allow for multi-pool delegation. This way, you only need a single stake key.  Wallets do not need to change much. dApps do not need to change at all.

The necessary changes are then constrained to updates to the ledger to track multi-delegation rewards properly.  No re-balancing of utxos is necessary. Everything moves to just being math inside the ledger staking/rewards system.

AndrewWestberg:
image
Just do it here.  Change it to an array where you can specify poolid/percentage.  Then make the necessary changes on the ledger team behind a hardfork flag. All math and updates are nicely self-contained in the node instead of bleeding out to community wallets, dapps, etc...

CIP-XXXX/README.MD
```diff
@@ -0,0 +1,146 @@
...
+
+In order to maintain a record of user’s delegation portfolio preferences , all actions pertaining to stake delegation, including funds transfers, stake key registration, and deregistration, must be executed within a single atomic transaction when performing an update. This transaction should also include the delegation portfolio metadata, using the reserved metadata label 6862 (hexadecimal 0x1ace), as outlined in [CIP-0017](https://github.com/cardano-foundation/CIPs/tree/master/CIP-0017) preferably without any of the optional fields.
+
+Example:
+
+``` Json
+{
+  "pools": [
+    {
+      "ticker": "DARK",
```
klntsky, line 93: I believe that putting stake keys here instead of pool ids/tickers would be better, because:

- there wouldn't be a need to update the metadata if delegation changes, but not the weight
- wallet implementations will have to perform one series of lookups less: instead of caring about pools, they will be only concerned with stake keys
- removing a delegation in favor of another would be as easy as re-delegating stake for one stake key - again, no metadata update

klntsky: As a separate note, there is probably no need in supporting tickers at all: reverse lookups only introduce complexity for the implementors to deal with

JaredCorduan:
@AndrewWestberg 's suggestion is indeed much cleaner, and indeed this has been suggested many times over the years.

TLDR - I think many people that think it is a good idea, it will just take time. So it's a matter of how soon do people want progress on multi-delegation.

The reasons the ledger change was never implemented were:
* the reward calculation is already a beast, and much effort was put into minimizing the consequences of this massive calculation. nobody wanted to make the situation worse until after the "big data structures" were moved out of memory. once this is done, the calculation can be done in a separate process and the worry is gone. (yes, it could be moved to a separate processes before it was moved out of memory, but that then just exacerbates the already big task of moving stuff out of memory...).
* the reward calculation is not currently easily swappable. What looks like a very easy change to the CDDL, and an easy change to the reward calculation, requires more refactoring than is immediately obvious.

Scitz0:
I recognize the issues raised by Jared but I fully agree with Andrews suggestion to try and raise the priority to get this included in the ledger instead of making various patch-work on top that will create a ton of bad UX, constant rebalancing of wallets, and a bloated utxo set. We have lived without "proper" multi-delegation for over three years, and I don't see a huge need for it. Account delegation has served us sufficiently so far I would say. So if to be implemented, let's do it properly.

gitmachtl:
> image Just do it here. Change it to an array where you can specify poolid/percentage. Then make the necessary changes on the ledger team behind a hardfork flag. All math and updates are nicely self-contained in the node instead of bleeding out to community wallets, dapps, etc...

I can double down on that. Since the beginning with Shelley we had an array output for a `query stake-address-info` on the CLI. And thats for a reason, because the idea was to list multiple reward-sources there. So yes, the best solution would be to talk to the ledger team and figure out a way to have a delegation/weight array instead of a single pool_keyhash in there.

gitmachtl:
> TLDR - I think many people that think it is a good idea, it will just take time. So it's a matter of how soon do people want progress on multi-delegation.

I think people are now familiar with "delays" on cardano. Creating a difficult workaround where you can ruin everything with a single wrong transaction is a bad replacement imo. Wallets can already automatically split an account into multiple accounts if people want to. I would like to see a ledger based solution here - even if it takes longer - than a complicated workaround. Oh by the way @JaredCorduan - Its lovely to hear from you. :-)

yHSJ:
> This is basically the approach Lace wallet takes. I disagree entirely with this approach.

I would agree this feels like an incomplete solution to a problem that can, and should, be solved at the ledger level for a number of reasons. That being said, I also think it's worth considering how a wallet could choose to implement multi-stake key accounts and I applaud the Lace team for working within the bounds that the ledger has defined currently.

It is also worth mentioning due to the reaction that I saw on twitter that a CIP is just that -- a proposal. Even if it were to get merged it does not become network "law". It would sit as `Proposed` until it's completed it's "Path to Active". In general, it's really great to see so many people engaging with the process and sharing their thoughts on multi-delegation!

QuasarChains:
I would like to ask that someone here look into the process [Gamechanger Wallet](https://gamechanger.finance/) uses for multi-delegation both for single account and also additional accounts and key derivations. CIP-30 does not support key derivations. Similar to other wallets and the ability to have different accounts, GameChanger calls these workspaces. Here's a [link to the repo](https://github.com/GameChangerFinance/gamechanger.wallet/blob/main/examples/Multi%20Stake%20Delegation.md) with an outline of the code and process for multi-delegation.

I think this is worth reviewing because I think an answer that is best for Cardano is closer than it may appear and one that allows anonymity a chance to remain prevalent.

Gamechanger is a web wallet so there is no download and the testnet version comes with an airdrop into a burner wallet so there is no password and no mnemonics and there's a built-in IDE to write/review/edit the code. Link to [testnet wallet](https://beta-preprod-wallet.gamechanger.finance/start) for easy access. Gamechanger is also a DApp so you can connect the Lace wallet along with any other browser extension to get a better view on how to build things so that they are composable and interoperable within the ecosystem.

Gamechanger and Lace handle multi-delegation differently and that is why it deserves this group's attention. 
Thank you.

rphair:
@zxpectre would you be willing to bring your perspective into this discussion as suggested by @QuasarChains in https://github.com/cardano-foundation/CIPs/pull/628#issuecomment-1837023314?

ccgarant:
> A better approach is to simply change the delegation certificate transaction to allow for multi-pool delegation. This way, you only need a single stake key.  Wallets do not need to change much. dApps do not need to change at all.

CIP-XXXX/README.MD
klntsky: dApps who will actually make use of this standard may not care about delegations at all. So why force them to perform additional lookups to figure out which of the stake addresses correspond to which pools?

CIP-XXXX/README.MD
klntsky, line 114: > for the latest transaction with the metadata label 6862 originating from the user's wallet

There's no such thing as user's wallet. There are utxos at addresses that are being consumed. Consider this: I construct an atomic swap transaction and put metadata that redefines the portfolio. I send it to the swap counterparty and they sign it. Who of us now has the delegation preferences updated?

Or even simpler, I am a malicious NFT minting dApp that wants to steal user's delegation. I put a metadata hash but do not disclose it to the user (which is normally expected)...

It is important to note that currently metadata posting never imposes any financial consequences for the users. But with this proposal adopted, it will.

This should not be accepted without a clear algorithm how to go from a chain of transactions to a mapping of `payment_address -> delegation_map`

CIP-XXXX/README.MD
rphair: thanks @klntsky I was only referring to your 2nd https://github.com/cardano-foundation/CIPs/pull/628#discussion_r1412220561 in my own https://github.com/cardano-foundation/CIPs/pull/628#discussion_r1412254242; apologies for the ambiguity & I've tried to correct that now.  BTW I would support the suggestion to use stake addresses vs. pool IDs.

CIP-XXXX/README.MD
AngelCastilloB: Thanks for the feedback @klntsky, only updating the metadata doesn't affect the users delegation, the delegation can only be altered by submitting certificate and signatures, so a malicious dApp cant steal user delegation by only updating the metadata of the transaction. The portfolio metadata is only intended to be consumed by off chain clients for context, I.E to determined the users preferences vs theirs actual stake distribution

CIP-XXXX/README.MD
```diff
@@ -0,0 +1,146 @@
...
+
+### Efficient Fund Tracking with Optional Privacy
+By linking all stake keys to the first payment key, this proposal introduces a transparent and simple mechanism for multi-delegation, greatly simplifying the fund tracking process. This association not only simplifies the delegation and deregistration processes but also facilitates easier tracking of funds and delegations. Besides, this method does not preclude the use of the multi-account approach for users desiring enhanced privacy. Stakeholders can still derive multiple accounts, each potentially capable of multi-delegation, granting them granular control while retaining the option for increased privacy.
+
+### Foundation for Advanced Wallet Features and Enhanced User Experience
+One of the underlying rationales of this proposal is to provide a robust foundation for more advanced wallet features. The dynamic nature of this approach offers wallets the opportunity to innovate. For example allowing users to support multiple dReps from a single account, or implementing sophisticated heuristics to maintain stake distribution to different pools. .
```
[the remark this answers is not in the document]
AngelCastilloB: thanks for the feedback @Ryun1, as you mentioned in your comment, DRep vote delegation and Stake delegation use the same key, so the two systems are intertwined at the protocol level, the only important consideration this CIP adds regarding DReps is that users will use more than one stake key and can have funds distributes among those stake keys, this will mean that users must also delegate their vote to the DReps for each key, for example:

Alice have a portfolio delegating to three pools, A, B and C with the following distribution 10%, 50%, 40%. She can chose to either delegate all three stake keys to the same DRep (effectively delegating 100% of her voting power to him) or delegate her different stake keys to different DReps, for example if she want to distribute her voting power between two DReps, she could delegate the same stake key used for pool A and B to DRep1 and the remaining stake key to DRep2 delegating to each each (approximately) 50% of her voting power, alternatively she could delegate to three different DReps. This is how the system works at the protocol level, the difference here is that wallets that implement this CIP will have the additionally flexibility to allow Alice to delegate to more than one DRep if she has more than one stake key.

As for how this would work with future types of delegations is hard to say since it will depend on the mechanics of those, but this CIP is not proposing/altering any feature at the protocol level, having multiple stake keys is already supported by the protocol, so I would assume there will be no conflict with any future use of stake keys unless there is a breaking protocol change in that regard.

CIP-XXXX/README.MD
klntsky: @AngelCastilloB 

> delegation can only be altered by submitting certificates and the right signatures in the transaction

They do not specify the amount, so that any ADA that is on a stake pubkey is automatically delegated.

Which means a wallet or a dApp that consumes this info will find the wrongly updated metadata and prompt the user to either perform a rebalance, or will silently generate a change distribution such that ADA is moved to a SP the user didn't want to prefer more.

Right now as I understand there is no well-defined connection between user's "identity" and metadata, and what is this "identity" is also undecided.

AngelCastilloB: Ah I see your point now @klntsky, so yes it would be possible for a malicious dApp to move the percentages, but it wouldn't be possible to make the user delegate to a different pool. 

So I guess the attack vector you are suggesting here would be for an adversarial pool to create a dApp that when it detects an user that is multi-delegating and one of the pools is multi-delegating to is in his control, it would then change the percentages to make it more favorable for him.

I think we have a few options here:

- Add a signature to the metadata
- Place the metadata in an script output the user controls

CIP-XXXX/README.MD
yHSJ: This is much better, just a small typo in the text itself.

> Wallet software can then easily discover all stake keys and associated delegations either by
sequential discovery.

Just remove the `either` and the sentence makes sense.

CIP-XXXX/README.MD
klntsky: > The users also interact with the delegation from the point of view of pools not stake keys, as in, users select pools and percentages and this information would need to be converted to key hashes in order to create the metadata as you are suggesting

You are only considering wallet's and user's point of view. But what about dApps that just want to generate change such that the distribution is preserved. These dApps may as well be other wallets that do not fully support this feature (i.e. in the UI), but try to respect the distribution.

I think it's obvious that one can always go from poolIDs to stake addresses and in the opposite direction. Stake addresses just have strictly more use cases without an extra lookup

CIP-XXXX/README.MD
klntsky: > Add a signature to the metadata

Yep. That would also make it unambiguous who is the issuer of the metadata in case of multiple user inputs. Moreover, to support this edge case, user address or pubkeyhash can be used as a key in the map.

> Place the metadata in an script output the user controls

There is no way to prevent consuming an UTxO from another wallet

Ryun1:
Hey @AngelCastilloB!

I have put this proposal on the Triage agenda for CIP Editors Call # 78, held in [The CIP Editors Meetings Discord](https://discord.gg/jvgrKbER), see [agenda](https://hackmd.io/@cip-editors/78) and [discord event](https://discord.com/events/971785110770831360/1179107190796210316).

Would be great to see you there to discuss this proposal.

kevinhammond:
> This is basically the approach Lace wallet takes. I disagree entirely with this approach. It has the following negative consequences:
> 
> ...  2.) The rewards claiming transaction becomes more complex; ...

I'm not sure I understand this, don't you simply register the same rewards address for multiple staked addresses?  Then the rewards withdrawal transaction is just the same.  Lace may not do that, but that's really a UI/UX issue isn't it?

leo42:
One solution that could stop applications from braking when attempting to connect to a wallet implementing this standard could be returning the cip Number in the supportedExtencions() api call if the selected wallet uses multi-delegation. 


This way it becomes easy for dApp developers to simply reject multi-delegation wallets if the application does not support them.

yHSJ:
> This way it becomes easy for dApp developers to simply reject multi-delegation wallets if the application does not support them.

Generally, most dApps won't "break" due to this change. For example, at JPG Store, we use the first index of the `getRewardAddresses` response to identify an "account". They have to sign with that stake key to authorize with our dApp and our transactions require that a signature from that key provides a witness. This is how many other dApps operate too. 

But, the dApp would behave in a "strange" way for users with multiple stake keys. There would be assets that the wallet displays that dApps may not show, for example. Or change management could get funky, etc.

leo42:
> Generally, most dApps won't "break" due to this change. For example, at JPG Store, we use the first index of the `getRewardAddresses` response to identify an "account". They have to sign with that stake key to authorize with our dApp and our transactions require that a signature from that key provides a witness. This is how many other dApps operate too.

Yes that is why is made the suggestion, an easy way for dApps to check if they are connecting to such a wallet will make all the3 controversy go away since dApp developers can now  reject such wallets if the dApp does not support it.

QuasarChains:
If it's left up to the wallets and the dapp builders I think we will see more problems in the long term. Composability should in my opinion always carry priority when designing and building.

I believe multi-delegation is important and is still a very unrealized opportunity to improve the stake pool network.

And how should DAOs or other groups or organizations that are using a multi-sig address that also have multi-delegation for that wallet going to be able to interact with dapps?

CIP-XXXX/README.MD
AngelCastilloB: done here https://github.com/cardano-foundation/CIPs/pull/628/commits/cdafdd40bf23e4eb200bff240c1af2d6c9974eb2

CIP-XXXX/README.MD
```diff
@@ -0,0 +1,146 @@
...
+Created: 2023-11-30
+License: CC-BY-4.0
+---
+
+## Abstract
+
+This document presents an alternative approach for Cardano sequential wallets to support multiple stake keys by deriving them from a single account. Building
+upon the foundations of [CIP-1852](https://github.com/cardano-foundation/CIPs/tree/master/CIP-1852) and [CIP-0011](https://github.com/cardano-foundation/CIPs/tree/master/CIP-0011), this proposal introduces a simple method for managing multiple delegations within a single account.
+
+## Terminology
```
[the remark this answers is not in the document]
AngelCastilloB: done here https://github.com/cardano-foundation/CIPs/pull/628/commits/cdafdd40bf23e4eb200bff240c1af2d6c9974eb2

CIP-XXXX/README.MD
```diff
@@ -0,0 +1,146 @@
...
+## Reference Implementation
+
+An already working implementation can be found in Lace wallet and the Cardano-js-sdk:
+
+- Lace: https://github.com/input-output-hk/lace
+- Cardano JS SDK: https://github.com/input-output-hk/cardano-js-sdk/blob/master/packages/e2e/src/tools/multi-delegation-data-gen/index.ts
+
+## Copyright
+
+CC-BY-4.0
```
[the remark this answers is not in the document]
AngelCastilloB: done here https://github.com/cardano-foundation/CIPs/pull/628/commits/cdafdd40bf23e4eb200bff240c1af2d6c9974eb2

CIP-XXXX/README.MD
```diff
@@ -0,0 +1,146 @@
...
+
+#### Benefits and Implementation Considerations
+
+Consistent Delegation Strategy: Ensures that the user's delegation strategy is consistently maintained, even as funds are spent and rewards are received.
+
+UTXO Set Optimization: Maintains the health and efficiency of the UTXO set as intended by the input selection algorithm.
+
+Reduced Manual Adjustments: Minimizes the need for users to manually rebalance their stake across different pools, saving time and effort.
+
+## Reference Implementation
```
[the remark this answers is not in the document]
AngelCastilloB: done here https://github.com/cardano-foundation/CIPs/pull/628/commits/cdafdd40bf23e4eb200bff240c1af2d6c9974eb2

CIP-XXXX/README.MD
```diff
@@ -0,0 +1,146 @@
...
+
+An already working implementation can be found in Lace wallet and the Cardano-js-sdk:
+
+- Lace: https://github.com/input-output-hk/lace
+- Cardano JS SDK: https://github.com/input-output-hk/cardano-js-sdk/blob/master/packages/e2e/src/tools/multi-delegation-data-gen/index.ts
+
```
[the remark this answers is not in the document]
AngelCastilloB: done here https://github.com/cardano-foundation/CIPs/pull/628/commits/cdafdd40bf23e4eb200bff240c1af2d6c9974eb2

AngelCastilloB:
> Hey @AngelCastilloB!
> 
> I have put this proposal on the Triage agenda for CIP Editors Call # 78 (December 12th 4pm UTC), held in [The CIP Editors Meetings Discord](https://discord.gg/jvgrKbER), see [agenda](https://hackmd.io/@cip-editors/78) and [discord event](https://discord.com/events/971785110770831360/1179107190796210316).
> 
> Would be great to see you there to discuss this proposal.

As we discussed during this meeting, I added a disclaimer to the document in the abstract:


>While this proposal outlines a method for multi-delegation within the current framework of Cardano's features, it is acknowledged that more efficient strategies may exist, particularly if multi-delegation capabilities were to be integrated at the protocol level. This CIP offers a solution that operates effectively under the existing system. However, it is important to note that should the Cardano protocol evolve to natively support multi-delegation in the future, this CIP would be considered deprecated in favor of the more direct, protocol-level approach. Such an advancement would likely offer a more streamlined and potentially more robust method for managing multi-delegation, aligning with the evolving needs and capabilities of the Cardano ecosystem.

Let me know if we are happy with this @yHSJ @Ryun1 @rphair @rhyslbw et al

AngelCastilloB:
> Thanks for submitting this! I left the following feedback in the [Lace Discord](https://discord.com/channels/1073034511664824360/1174398420006477864/1179795276303712377), but felt I should share it here too:
> 
> > My initial feedback would be that you should use [RFC-2119](https://datatracker.ietf.org/doc/html/rfc2119) to strictly define the language use, and then modify the specification to use that language. This is feedback I've given to almost every CIP. Cardano's standards right now use a lot of ambiguous language that one could argue is implicitly defined, but as an implementer I've seen several different interpretations of standards. If RFC-2119 was used, that ambiguity is removed.

I tried my best to add this into the CIP, but I found it moistly made sense in the specification section, I will further improve it if needed tho

AngelCastilloB:
> > Generally, most dApps won't "break" due to this change. For example, at JPG Store, we use the first index of the `getRewardAddresses` response to identify an "account". They have to sign with that stake key to authorize with our dApp and our transactions require that a signature from that key provides a witness. This is how many other dApps operate too.
> 
> Yes that is why is made the suggestion, an easy way for dApps to check if they are connecting to such a wallet will make all the3 controversy go away since dApp developers can now reject such wallets if the dApp does not support it.

dApps can already do this by querying CIP-30  `getRewardAddresses`  and checking if there is more than one. Some dApps have incompatibility issues with wallets having more than one stake key particularly, because by convention they are assuming wallets will only have one. This CIP is not introducing the ability for wallets to have more than one stake key, this is already a feature of the protocol, this CIP or basically any use of multiple stake keys will break these dApps.

In my personal opinion dApps shouldn't build around conventions but rather the protocol, this will ensure the dApps are robust to all use cases

rphair:
> (https://github.com/cardano-foundation/CIPs/pull/628#issuecomment-1875784358) Let me know if we are happy with this

Ordinarily a statement like this would go in the _Rationale_ but I can admit it would be worth the couple inches of text to highlight it in the _Abstract_ because it is an important qualification to the proposal's scope and target audience.  If "@yHSJ @Ryun1 @rphair @rhyslbw et al" are offended by its presence there, it should be moved to the top of the _Rationale_.

rhyslbw:
> Let me know if we are happy with this @yHSJ @Ryun1 @rphair @rhyslbw et al

LGTM @AngelCastilloB

CIP-XXXX/README.MD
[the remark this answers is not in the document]
rphair: double-checked; resolving.

CIP-XXXX/README.MD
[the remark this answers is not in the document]
rphair: double-checked; resolving.

CIP-XXXX/README.MD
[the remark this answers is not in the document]
rphair: double-checked; resolving.

CIP-XXXX/README.MD
```diff
@@ -0,0 +1,164 @@
...
+## Rationale: how does this CIP achieve its goals?
+
+This CIP details a method for delegating to multiple pools from a single account, effectively meeting the Cardano community's demand for more versatile delegation strategies. By permitting users to manage multi-pool delegations within one account, this proposal significantly enhances the user experience, offering benefits in terms of investment diversification and risk management. Furthermore, it lays the groundwork for advanced wallet functionalities, like supporting multi-DRep delegation and introducing more complex algorithms for stake distribution. It is important to note, however, that this approach is designed within the constraints of the current Cardano features. Should future protocol developments introduce native multi-delegation capabilities, this CIP would be superseded.
+
+## Path to Active
+
+### Acceptance Criteria
+
+- [ ] The proposal is adopted by all major wallets and dApps in the ecosystem.
```
rphair, line 156: I think keeping this item as written would ensure that is stays in `Proposed` status forever.  Depending on the outcome of the discussion planned in my review above, this could be a sensible approach... although if other implementations are on the horizon then a more flexible acceptance threshold should be defined here.

rphair (commented):
The bump from @rhyslbw's https://github.com/cardano-foundation/CIPs/pull/628#issuecomment-2211692621 makes me think we should move this forward if there are no grounds for disqualifying it: so I've put it up for Review again at the next CIP meeting (https://hackmd.io/@cip-editors/92).

Most are in favour of doing this in the Ledger as per @AndrewWestberg https://github.com/cardano-foundation/CIPs/pull/628#issuecomment-1836103372, although the CIP editors haven't been made aware (as a group) in the meantime of any upcoming changes in the Ledger that would facilitate multi-delegation in the protocol.

* @WhatisRT @lehins can we record here any relevant changes corresponding to Andrew's comment in the upcoming hard fork... or confirm that they're being considered for a future hard fork... or note here if they're not being planned at all?

I would be inclined to merge this as per what @yHSJ says about also having merged CIP-0018 (https://github.com/cardano-foundation/CIPs/pull/628#issuecomment-1852443987) but I would definitely want to update it first about any Ledger changes or changing community expectations in the meantime.

Before merging I think we also need to see a review of any wallets besides Lace supporting this method as outlined (cc @Crypto2099 @Ryun1).  I think it's OK to merge as Proposed given @AngelCastilloB's comment on the qualification added here (https://github.com/cardano-foundation/CIPs/pull/628#issuecomment-1875784358) but the community & wallet / dApp developers might want to see a stronger statement (if true) saying (more or less): "This approach is taken in the Lace wallet, but nowhere else, and ideally it should be done in the Ledger."

WhatisRT:
I'm not aware of any plans to implement multi-delegation. Intersect would have to put that on the roadmap. There are a bunch of technical hurdles to this feature, since it touches some of the most complicated pieces of Ledger code. There are also no new features in Conway that should help implementing multi-delegation without ledger support.

CIP-XXXX/README.MD
```diff
@@ -0,0 +1,164 @@
+---
+CIP: ????
```
rphair, line 2:
```suggestion
CIP: 126
```

rphair (commented):
Thanks again for a constructive submission; please also retitle the containing directory as `CIP-0126`. 🎉

@AngelCastilloB @rhyslbw the CIP editors and community at our last meeting confirmed the need to qualify this proposal as "unique to Lace" — somehow, according to the recommendations above — before it is merged: to avoid institutionalising the potential conflicts with dApps that this alternative delegation approach might lead to.

AngelCastilloB:
> Thanks again for a constructive submission; please also retitle the containing directory as `CIP-0126`. 🎉
> 
> @AngelCastilloB @rhyslbw the CIP editors and community at our last meeting confirmed the need to qualify this proposal as "unique to Lace" — somehow, according to the recommendations above — before it is merged: to avoid institutionalising the potential conflicts with dApps that this alternative delegation approach might lead to.

Hi @rphair how should this be represented in the CIP?, not sure where to add this in the proposal, any preference/recomendation?, maybe we extend the disclaimer?

> This proposal outlines a method for multi-delegation within the current framework of Cardano's features, and is currently unique to Lace wallet. It is acknowledged that more efficient strategies may exist, particularly if multi-delegation capabilities were to be integrated at the protocol level. This CIP offers a solution that operates effectively under the existing system. However, it is important to note that should the Cardano protocol evolve to natively support multi-delegation in the future, this CIP would be considered deprecated in favor of the more direct, protocol-level approach. Such an advancement would likely offer a more streamlined and potentially more robust method for managing multi-delegation, aligning with the evolving needs and capabilities of the Cardano ecosystem.

What do you think? ☝🏼

CIP-XXXX/README.MD
AngelCastilloB: done here https://github.com/cardano-foundation/CIPs/pull/628/commits/b38ef167f24ce18ce3069ccb3c98a46545ac0c4c

rphair:
re: https://github.com/cardano-foundation/CIPs/pull/628#issuecomment-2242826531

I think adding the term "currently unique to Lace wallet" - as you have done to the "disclaimer" paragraph in the Abstract - is both the best way of doing it (because it appears early in the document in a place where target audience is established) and a sufficiently response to the community feedback from last meeting.

You could merge this paragraph into your document in this branch, so we can read how it appears in the regular flow... if not done before tomorrow's meeting (in 12 hours) we'll review it as above.  If editors / devs think more is required than "unique to Lace" then we'll list the suggestions here after the meeting.  cc @Crypto2099 @Ryun1

AngelCastilloB:
> re: [#628 (comment)](https://github.com/cardano-foundation/CIPs/pull/628#issuecomment-2242826531)
> 
> I think adding the term "currently unique to Lace wallet" - as you have done to the "disclaimer" paragraph in the Abstract - is both the best way of doing it (because it appears early in the document in a place where target audience is established) and a sufficient response to the community feedback from last meeting.
> 
> You could merge this paragraph into your document in this branch, so we can read how it appears in the regular flow... if not done before tomorrow's meeting (in 12 hours) we'll review it as above. If editors / devs think more is required than "unique to Lace" then we'll list the suggestions here after the meeting. cc @Crypto2099 @Ryun1

done in https://github.com/cardano-foundation/CIPs/pull/628/commits/89bff9a612acbc1a4ecad9cc67bee1d283d1c736

rphair (approved):
@Ryun1 @Crypto2099 I think we agreed at the last relevant CIP meeting that we would merge this as soon as officially documented as an approach unique to the Lace wallet.  Since this did not trigger an "automatic" review I suppose therefore this just "fell through the cracks" so I'm tagging it `Last Check` for confirmation at (the next / a near-term) CIP meeting.  cc @AngelCastilloB @rhyslbw

rphair:
Note @Ryun1 @Crypto2099 it may in fact be better to leave the objectionable statement https://github.com/cardano-foundation/CIPs/pull/628#discussion_r1667340747 in the CIP:
> The proposal is adopted by all major wallets and dApps in the ecosystem.

... since that would in fact leave the CIP at `Proposed` state indefinitely... unless somehow the unconventional Lace standard were to become universal, and then could be made `Active`.  Until such a time, this CIP still serves a purpose to document their own approach & potentially make itself available to other implementations.  (cc @AngelCastilloB @rhyslbw)

CIP-XXXX/README.MD
rphair: @AngelCastilloB - continued in https://github.com/cardano-foundation/CIPs/pull/628#issuecomment-2297894492 with a reason why it may actually make sense to leave this apparently unsatisfiable criterion on the list.

CIP-0126/README.MD
```diff
@@ -0,0 +1,164 @@
...
+
+- **First Payment Key:** This refers to the external (role 0) payment key derived at index 0.
+
+- **First Stake Key:** Similarly, this is the stake key (role 2) derived at index 0.
+
+Throughout this proposal, the terms "first payment key" and "first stake key" will be used to refer to these specific keys.
+
+### Overview
+
+Under this proposal, the constraints of [CIP-0011 ](https://github.com/cardano-foundation/CIPs/tree/master/CIP-0011)regarding the sole derivation index for the stake key are removed. While we introduce the ability to derive stake keys beyond the first , there is no strict requirement to do so in a sequential manner, although it is RECOMMENDED for logical organization. A pivotal change in this approach is that all derived stake keys MUST be associated with the first payment key. This consistent association to the first payment key serves as the foundation for the efficient discovery of these stake keys by wallet software and services.
```
Ryun1, line 59:
```suggestion
Under this proposal, the constraints of [CIP-0011](https://github.com/cardano-foundation/CIPs/tree/master/CIP-0011)regarding the sole derivation index for the stake key are removed. While we introduce the ability to derive stake keys beyond the first , there is no strict requirement to do so in a sequential manner, although it is RECOMMENDED for logical organization. A pivotal change in this approach is that all derived stake keys MUST be associated with the first payment key. This consistent association to the first payment key serves as the foundation for the efficient discovery of these stake keys by wallet software and services.
```

CIP-0126/README.MD
```diff
@@ -0,0 +1,164 @@
...
+### Reference Implementation
+
+An already working implementation can be found in Lace wallet and the Cardano-js-sdk:
+
+- Lace: https://github.com/input-output-hk/lace
+- Cardano JS SDK: https://github.com/input-output-hk/cardano-js-sdk/blob/master/packages/e2e/src/tools/multi-delegation-data-gen/index.ts
+
+## Rationale: how does this CIP achieve its goals?
+
+This CIP details a method for delegating to multiple pools from a single account, effectively meeting the Cardano community's demand for more versatile delegation strategies. By permitting users to manage multi-pool delegations within one account, this proposal significantly enhances the user experience, offering benefits in terms of investment diversification and risk management. Furthermore, it lays the groundwork for advanced wallet functionalities, like supporting multi-DRep delegation and introducing more complex algorithms for stake distribution. It is important to note, however, that this approach is designed within the constraints of the current Cardano features. Should future protocol developments introduce native multi-delegation capabilities, this CIP would be superseded.
```
Ryun1, line 150: From the CIP template

"The rationale fleshes out the specification by describing what motivated the design and what led to particular design decisions. It should describe alternate designs considered and related work. The rationale should provide evidence of consensus within the community and discuss significant objections or concerns raised during the discussion."

Before I press the approve, I really think this proposal requires "describe alternate designs considered and related work."

rphair:
@AngelCastilloB @rhyslbw marking `Waiting for Author` until reservation @Ryun1 already posted in https://github.com/cardano-foundation/CIPs/pull/628#discussion_r1723629043 is met.  Today's CIP meeting confirmed with community today that this is not admissible as a CIP unless it includes a forthright presentation about how & why this approach diverges from the prevailing approaches of other contemporary Cardano wallets.

rphair:
@AngelCastilloB @rhyslbw I had always thought this document was ready to merge with a single disclaimer as suggested in the editorial reviews above (about this approach not being universal).

Normally after so long `Waiting for Author` we might naturally close a PR as likely abandoned... but could you consider taking one of these actions instead:
1) close it voluntarily, if this is no longer the approach taken by the Lace wallet
2) I think everyone would prefer, if this CIP is in fact implemented by Lace, that you permit the editors to provide such a disclaimer (generally written above already), approve, and merge this into the repository.

Please let us know which course we should take.  If not hearing anything after this point we would after some time close it as "abandoned"... which I think would be a shame after you & your team have already gone to such well-appreciated lengths to document this approach.

rhyslbw:
Hey @rphair. Sub-account multi-delegation will not be implemented in Lace v2, since we have a more natural multi-account foundation to offer a way for distributing stake across multiple pools. We still feel there are benefits to the documented approach but lack of support from other ecosystem players creates poor UX that we want to avoid. We will close this PR to signal abandonment of the concept. Thanks for the prompt here

PR #631, opened 2023-12-04
 CIP-0072 | Off-chain schema fixes

danielmain:
By creating a product on top of this specification we came up with two small fixes on the schnma definiton of the off-chain JSON

Fixes:
- We use `anyOf` instead of `oneOf` for the **logo**, by accepting any image format like png/jpeg/svg
- We use `anyOf` instead of `oneOf` for the **screenshots**, by accepting any image format like png/jpeg/svg
- The subject accpets any string and dots, adhering to the reverse domain name pattern commonly employed as the package or bundle identifier in Android and iOS applications. This widely accepted convention serves to guarantee uniqueness and mitigate naming conflicts, having established itself as a standard within the development community. For example: `com.company.application`
- Small text in long-descriptoion changed. (Removed the word optional) 
- Limitations in size of metadata and having each picture in base64 makes it difficult to handle big pictures. Setting maximum of logo and screenshots.
- Added a requirement of minimum one social link and at least one screenshot to add in the offchain json.
- Added regex pattern to validate that the string matches base64 images

Since this a fix we are upgrading the version of the schema file to 2.0.1


rphair (approved):
To me this seems like both a justification of the semantic versioning [previously debated](https://github.com/cardano-foundation/CIPs/pull/612#discussion_r1394835574) and an example of a change editors can approve as a matter of routine.

Ryun1 (approved):
LGTM

Quantumplation:
@danielmain re your comments in the CIP editors meeting on "subject" vs "domain" validation;

I really like the idea, my only suggestions would be:
- make it an explicit `domain` field, rather than something like a `subject`
- don't use the "reverse-domain" convention from Java; it's not widely adopted in the industry, and IMO is confusing as hell haha

rphair:
👍 @Quantumplation https://github.com/cardano-foundation/CIPs/pull/631#issuecomment-1852352314 above... in this particular (crypto) industry the "app stores" should be able to support plenty of "headless" dApps without web sites, and the reverse-domain convention relies apparently on DNS and nothing else to avoid collisions:

https://docs.oracle.com/javase/specs/jls/se7/html/jls-6.html#jls-6.1
> The name of a package is not meant to imply where the package is stored on the Internet. The suggested convention for generating unique package names is merely a way to piggyback a package naming convention on top of an existing, widely known unique name registry instead of having to create a separate registry for package names.

danielmain:
> make it an explicit domain field, rather than something like a subject

You are totally right! We will make this change :) 

> don't use the "reverse-domain" convention from Java; it's not widely adopted in the industry, and IMO is confusing as hell haha

I only know the two giants in the industry, Google Play Store and IOS App Store 🤷‍

Google Play Store does it (becaus it started with java)

And the second largest App Store do it in that fasion too, look at the apple apps  https://support.apple.com/en-gb/guide/deployment/depece748c41/web.

If the industry or at least the vast mayority follows this pattern, @Quantumplation what do you think? would not make sense to make it as familiar as possible?

danielmain:
> 👍 @Quantumplation [#631 (comment)](https://github.com/cardano-foundation/CIPs/pull/631#issuecomment-1852352314) above... in this particular (crypto) industry the "app stores" should be able to support plenty of "headless" dApps without web sites, and [reverse domain name notation](https://en.wikipedia.org/wiki/Reverse_domain_name_notation) relies apparently on DNS and nothing else to avoid collisions:
> 
> https://docs.oracle.com/javase/specs/jls/se7/html/jls-6.html#jls-6.1
> 
> > The name of a package is not meant to imply where the package is stored on the Internet. The suggested convention for generating unique package names is merely a way to piggyback a package naming convention on top of an existing, widely known unique name registry instead of having to create a separate registry for package names.

Good point, @rphair, but isn't the domain a shared element across all dapps 🤔? Are there any Cardano dapps that operate without a domain?

However, it's important to note that the subject or bundleID isn't the domain itself, but rather an identifier. Often it's in the reverse-domain format, which is common for many apps on iOS/Android.

While we can recommend using a specific 'reverse-domain', the choice is ultimately yours to make 🙂

Also, it's not just Java that follows this practice; Apple does too: https://developer.apple.com/documentation/bundleresources/information_property_list/cfbundleidentifier
`A bundle ID uniquely identifies a single app throughout the system. The bundle ID string must contain only alphanumeric characters (A–Z, a–z, and 0–9), hyphens (-), and periods (.). Typically, you use a reverse-DNS format for bundle ID strings. Bundle IDs are case-insensitive.`

CIP-0072/version_2.0.1_offchain.json
```diff
@@ -0,0 +1,270 @@
...
+            "contentMediaType": "image/jpeg"
+          },
+          {
+            "contentMediaType": "image/svg+xml"
+          }
+        ],
+        "maxLength": 2722000
+      },
+      "pattern": "^(data:image\\/png;base64,|data:image\\/jpeg;base64,|data:image\\/svg\\+xml;base64,)[a-zA-Z0-9\\+/]*={0,2}$",
+      "description": "Screenshots of the DApp encoded in base64. We recommend to share screenshots from the dApp usage itself."
```
greatertomi, line 93:
```suggestion
        ],
        "pattern": "^(data:image\\/png;base64,|data:image\\/jpeg;base64,|data:image\\/svg\\+xml;base64,)[a-zA-Z0-9\\+/]*={0,2}$",
        "maxLength": 2722000
      },
      "description": "Screenshots of the DApp encoded in base64. We recommend to share screenshots from the dApp usage itself."
```

rphair:
@danielmain we are not in a hurry to close this but we do need to know where this submission is going.  Do you still want/need to make these changes?  It's been 7 months... do your plans still look this same?  When will this be ready for review?  What issues are you looking at now, as updated from last year when you submitted this?  Please let us know what we are waiting for (tagging `Waiting for Author` in the meantime). (cc @greatertomi)

rphair:
@danielmain if this is going to move forward as a document (as we would definitely hope, if you have been progressing this effort in the development world!) then please let us know soon (marking potentially "abandoned").  cc @greatertomi @Quantumplation @Ryun1

PR #634, opened 2023-12-06
CIP-0010 | Add proof of origin metadata label

iFergal:
We would like to register a new metadata label `1904` reserved for proof of origin and quality control in the supply chain space. The first use case is our [pilot project](https://cardanofoundation.org/en/news/cardano-foundation-partners-with-georgian-national-wine-agency/) at the Cardano Foundation with the National Wine Agency of Georgia.

We have prepared a specification which should hopefully become its own CIP but is still internally under final review. We would like to for now reserve this label and later update the description to point to the CIP.

The standard aims to cover proof of origin (see some initial documentation on the metadata [here](https://cardano-foundation.github.io/cf-georgian-wine-resolver/)) and quality control through Certificates of Conformity (documentation pending). The upcoming CIP intends to cover a broader set of use cases than just Georgian wine.

### DIFF
--- a/CIP-0010/registry.json
+++ b/CIP-0010/registry.json
@@ -56,6 +56,10 @@
     "description": "amphitheatre by the ape society"
   },
   {
+    "transaction_metadatum_label": 1904,
+    "description": "Proof of origin and quality related verification data in supply chain"
+  },
+  {
     "transaction_metadatum_label": 1967,
     "description": "nut.link metadata oracles registry"
   },


CIP-0010/registry.json
```diff
@@ -55,6 +55,10 @@
     "transaction_metadatum_label": 1888,
     "description": "amphitheatre by the ape society"
   },
+  {
+    "transaction_metadatum_label": 1904,
+    "description": "Proof of origin and quality related verification data in supply chain"
```
rphair, line 60:
```suggestion
    "description": "Proof of origin and quality verification data in supply chain"
```
Considering space is at a premium here, and that this is our longest `description`, wouldn't it mean the same thing without `related`?

When you have your new CIP with the full spec, we would be adding that here & therefore wouldn't need so many words beyond the CIP number: so this isn't vital & there'd be no point shortening it further.

rphair (approved):
looks well anticipated & documented; I've been following & bookmarking this project for some time.  Metadata label `1904` confirmed in developer documentation above.

CIP-0010/registry.json
rphair: On second thought, that introduces an ambiguity implying "proof of quality" (generally impossible) so better to leave it the way it is & trust it will be elaborated properly in the forthcoming CIP spec.

Ryun1 (approved):

PR #556, opened 2023-07-20
CIP-0100? | Governance Metadata

Quantumplation:
This CIP is the second draft for a standard surrounding any "metadata" attached to the activities that users engage in during governance.

The first draft, along with comments left by some members of the Cardano community, can be found here:
https://gist.github.com/Quantumplation/56ef2ffccb05f5d2974fd17240dc406c

A recording of the workshop held to solicit feedback that went into this second draft can be found here:
https://www.youtube.com/watch?v=TnPJ11Po04s

There are some `<!-- inline comments -->` with notes from various workshops, if you're curious, but that aren't relevant to the main CIP.

There are some open questions in the form of task lists that I could use some help answering.

[Rendered](https://github.com/Quantumplation/CIPs/blob/master/CIP-0100/README.md)

PR #625, opened 2023-11-27
CPS-0012? | Query Layer Standardization

klntsky:
Cardano lacks a standardized query layer. This leads to suboptimal tooling, dApp and wallet architecture.

---

([latest document](https://github.com/klntsky/CIPs/blob/klntsky/query-layer-cps/CPS-0012/README.md))

PR #635, opened 2023-12-11
CIP-0107? | URI Scheme - Block and transaction objects

Quantumplation:
This CIP extends CIP-0013 with two new authorities, and enabling a simple and canonical reference to 5 new objects or locations on the Cardano blockchain.

In particular, this will be useful in [CIP-100](https://github.com/cardano-foundation/CIPs/pull/556), where we want to allow users to make references to on-chain transactions, blocks, and metadata, or store the transaction metadata within the same transaction.

[Rendered](https://github.com/Quantumplation/CIPs/blob/cip-13-extension/CIP-0107/README.md)


CIP-?/README.md
```diff
@@ -0,0 +1,109 @@
...
+```
+
+### Grammar & interpretation
+
+We extend the grammar from CIP-0013 with two new authorities:
+
+```
+authorityref = (stakepoolref | blockref | transactionref | otherref)
+```
+
```
rphair, line 50:
```suggestion
```
This isn't needed because you are only defining grammars for the cases you are covering here: it must be one or the other, and so the `stakepoolref` and `otherref` cases are irrelevant and the definitions for `blockref` and `transactionref` below are sufficient.

Also: for all CIPs extending the CIP-0013 URI Scheme to be equally meaningful without ordering, I wouldn't want further CIPs besides this one to try to list all the previously defined authorities (note we also already have https://github.com/cardano-foundation/CIPs/pull/546 in the works, which defines `claimtokenref` and should likewise be considered irrelevant).

CIP-?/README.md
```diff
@@ -0,0 +1,109 @@
+---
+CIP: ?
+Title: Cardano URI Scheme Extensions
```
rphair, line 3:
```suggestion
Title: URI Scheme - Block and transaction objects
```
1 - Since https://github.com/cardano-foundation/CIPs/pull/559 we have expected more CIPs to define new CIP-0013 `authority` keywords, and since https://github.com/cardano-foundation/CIPs/pull/594 we've been grouping e.g. the wallet API extensions with the CIP `Title:` prefix `Web-Wallet Bridge -`.

Therefore I would suggest we begin CIPs for new CIP-0013 extensions with this title prefix: `URI Scheme -`

2 - I also think "common objects" is not specific enough to indicate what these links are referring to (the term "object" is too broad) and whenever possible they should be mentioned by function (or in some cases the `authority` itself).

@Quantumplation I'm setting the PR title as such but we can follow whatever is agreed upon here.  @Crypto2099 @Ryun1

rphair (commented):
@Quantumplation I am really happy to see this presented, especially with concurrent discussion of https://github.com/cardano-foundation/CIPs/pull/625 highlighting that objects should be possible to reference independently of query providers (cc @klntsky).

rphair:
@Quantumplation on agenda for quick Triage assessment in next CIP meeting on 12 December, if you can make it (this is our last CIP meeting for 4 weeks): https://hackmd.io/@cip-editors/78

CIP-?/README.md
Quantumplation: I pictured this more as a delta, to show where it fits within the existing BNF; BNF doesn't really have a good mechanism for this, afaik.

Technically speaking, CIP-0013 is written in such a way that this doesn't need a grammar at all; `otherref` is defined super generally, and so every one of the url patterns I defined would be covered by "otherref", I think.  So, to be clear and precise, `authorityref` needs to be updated, otherwise my BNF would just be dangling and unreferenced.

That being said, it's more for human consumption, and not being consumed programatically, so I'm happy to conform to whatever convention you see fit. To prevent a bunch of back-and-forth, feel free to just update it to the style you prefer.  Perhaps something like `authorityref = (blockref | transactionref | ...)` to indicate "and all the rest from other CIPs"?

rphair: yes, I know what you mean; it's hard to put into words how ABNF isn't written with "hooks" or "plugins" in mind.  I'd tried to rewrite CIP-0013 that way and failed.  It was my hope that no provider of further CIPs would have to include any BNF line that was in CIP-0013 and I do think that it all fits together with the redundant `authorityref` definition above removed.

If that seems incomplete or too abrupt to you, even if `authorityref = (blockref | transactionref | ...)` is not strictly valid ABNF then I think it would work well here... as would a verbal description like "This CIP adds two new possibilities for the `authorityref` term." 🤔

Quantumplation: Given that we don't really expect to consume the ABNF programatically, I'm personally in favor of that syntax, as I think it's immediately clear what it means.

Crypto2099: I think I would be in favor of:
`authorityref = (... | blockref | transactionref)`

From a purely programmatic standpoint to me this means "everything that came before plus this"

I would agree with Pi's assessment that since this is merely syntactic sugar for human readers it's not crucially important either way and more important that we get the meaning across more than anything.

CIP-?/README.md
```diff
@@ -0,0 +1,109 @@
...
+
+> WARNING: Referencing blocks by height is provided in the case of extremely tight space requirements, but referencing a recent block (within the rollback horizon of the chain) is unstable.  Due to chain re-orgs, it may refer to different content for different users, or at different times. This should only be used when this is not critical, or for blocks outside of the rollback horizon.
+
+```
+blockref = "//block" query
+query = ( "?block=" block_hash | "?height=" block_height)
+
+block_hash = 64HEXDIG
+block_height = *digit
+```
```
Crypto2099, line 63: I don't think we would ever run into "space" constraints with this URI schema so I would prefer to see that all blocks be referenced by their cryptographic hash rather than a specific numeric index that is fragile for the aforementioned reasons.

CIP-?/README.md
```diff
@@ -0,0 +1,109 @@
...
+For grammar reference, see:
+
+  - [Wikipedia > Augmented Backus–Naur form](https://en.wikipedia.org/wiki/Augmented_Backus%E2%80%93Naur_form)
+  - [RFC 2234: Augmented BNF for Syntax Specifications: ABNF](https://datatracker.ietf.org/doc/html/rfc2234)
+  - [Unicode in ABNF](https://tools.ietf.org/html/draft-seantek-unicode-in-abnf-00)
+
+## Rationale: how does this CIP achieve its goals?
+
+This CIP defines a canonical format for URIs referencing four new Cardano objects: blocks, transactions, transaction metadata, and specific tags within the transaction metadata. It utilizes existing cardano standards (CIP-0013) and industry standards (URIs), minimizing the number of new concepts that a developer nees learn. By utilizing URIs, it creates a natural path to integration with existing tools, such as browsers. And finally, it allows a canonical URI for these objects, such as storing CIP-100 metadata on-chain, and referring to it in the anchor field.
+
```
Crypto2099, line 89: #
```suggestion
This CIP defines a canonical format for URIs referencing four new Cardano objects: blocks, transactions, transaction metadata, and specific tags within the transaction metadata. It utilizes existing cardano standards (CIP-0013) and industry standards (URIs), minimizing the number of new concepts that a developer needs to learn. By utilizing URIs, it creates a natural path to integration with existing tools, such as browsers. And finally, it allows a canonical URI for these objects, such as storing CIP-100 metadata on-chain, and referring to it in the anchor field.
```

minor grammatical fix

Crypto2099 (commented):
This looks great, easy, and clear to me at first reading. See no reason why it should not be adopted. :)

CIP-?/README.md
Quantumplation: You'd be surprised; Currently the `anchor` field in CIP-1694 is limited to 64 characters, for example.

Quantumplation: I would say leave it off, because I don't currently have a strong motivation for it; but if we do, the next step of simplification makes it hard to add later; `web+cardano://block/{block_hash}` doesn't have a clear way to represent height.

rphair: Since this is rolling out in a governance context, wouldn't a human readable `height` be important to compare dates of governance propositions or decisions (even with lack of finality sometimes applicable)?  I can see how could be a "strong motivation" although @Quantumplation you would have a better idea of that from the workshops.

CIP-?/README.md
```diff
@@ -0,0 +1,108 @@
...
+Examples:
+```
+web+cardano://block/c6a8976125193dfae11551c5e80a217403d953c08ebbd69bba904d990854011f
+web+cardano://transaction/7704a68404facf7126fa356f1b09f0e4c552aeef454cd0daba4208f3a64372e9
+web+cardano://transaction/7704a68404facf7126fa356f1b09f0e4c552aeef454cd0daba4208f3a64372e9#1
+web+cardano://transaction/7704a68404facf7126fa356f1b09f0e4c552aeef454cd0daba4208f3a64372e9/metadata
+web+cardano://transaction/d3616b772c91f118346e74863d1722810a4858e4d7cc7663dc2eed345d7bca72/metadata/674
+web+cardano://transaction/self/metadata/1694
+```
```
Crypto2099, line 41: I believe if we're going to keep the block hash or block height language in the lower section we need to update the block example to account for the `?hash=<blah>` or `?height=<big_int>` format as the first example here does not reflect that potential.

CIP-?/README.md
Quantumplation: Not really; you could still grab that data on-chain, or store a timestamp in the governance metadata, etc.

This is more so about generating a URL *only* based on the height, like being able to link to `web+cardano://block?height=9664256` vs `web+cardano://block?hash=4bfe21ff68e22266c14bcbdce6fc76d31c9813516c4bcdfa697069bb5e9e1f41` or `web+cardano://block/4bfe21ff68e22266c14bcbdce6fc76d31c9813516c4bcdfa697069bb5e9e1f41`

From a data availability standpoint they're the same, it's just a matter of what URLs people produce, read, and click on.

CIP-?/README.md
Quantumplation: Oh yea, good point. I'll wait until a decision is made on that, and then revise this if needed

CIP-?/README.md
Quantumplation: @Crypto2099 what do you think, what's the verdict here?

CIP-?/README.md
```diff
@@ -0,0 +1,108 @@
+---
+CIP: ?
```
rphair, line 2:
```suggestion
CIP: 107
```

rphair (commented):
@Quantumplation you are all set to rename the directory to `CIP-0107` and update your "Rendered" link accordingly 🎉 ...

Quantumplation:
Done and Done.

CIP-0107/README.md
```diff
@@ -0,0 +1,108 @@
...
+
+Without a canonical standard for how to define these URIs, these objects are either unaddressable, not machine-interpretable, or will suffer from a divergence of convention. Similarly, we seek to fit an existing structure, such as CIP-0013, to reduce the number of different conventions that need be supported by the ecosystem.
+
+## Specification
+
+We extend CIP-13 with 2 new authorities for referencing blocks and transactions.
+
+Examples:
+```
+web+cardano://block/c6a8976125193dfae11551c5e80a217403d953c08ebbd69bba904d990854011f
```
rphair, line 35: (1 of 2) If `height=` is preserved in the query part of the URI syntax documented below, an example should be added here.

CIP-0107/README.md
```diff
@@ -0,0 +1,108 @@
...
+
+#### Block queries
+
+A link referencing a block by either block hash or height.
+
+> WARNING: Referencing blocks by height is provided in the case of extremely tight space requirements, but referencing a recent block (within the rollback horizon of the chain) is unstable.  Due to chain re-orgs, it may refer to different content for different users, or at different times. This should only be used when this is not critical, or for blocks outside of the rollback horizon.
+
+```
+blockref = "//block" query
+query = ( "?block=" block_hash | "?height=" block_height)
```
rphair, line 59: (2 of 2) If `height=` is eliminated due to reservations about immutability (as per the previous & current discussion), it should be removed here.

rphair (commented):
@Quantumplation given that a new span of URI protocols + a containing CPS (https://github.com/cardano-foundation/CIPs/pull/841) are now moving ahead together, I would happily approve & merge this as soon as we settle the question about whether & how the `?height=` query is used as per currently unresolved discussions:
https://github.com/cardano-foundation/CIPs/pull/635#discussion_r1423463179
https://github.com/cardano-foundation/CIPs/pull/635#discussion_r1424224364

I would prefer a consensus to keep `height=` in the syntax, despite its lack of cryptographic certainty, since I believe it would likely be a popular choice when this URI syntax becomes more generally adopted.

Specifically, I would not want to be the one to be explain to members of the public why they couldn't use it (in historical and/or governance contexts as I argued for in https://github.com/cardano-foundation/CIPs/pull/635#discussion_r1424199929) when it is meaningful for a far greater time span (in hindsight, forever) than its period of uncertainty (contemporary with its creation).

In any case as soon as there is a consensus about `height=`, and the CIP is updated according to that consensus, I'm ready to approve this.  Adding to CIP meeting 9 days from now (https://hackmd.io/@cip-editors/91) to make sure this is progressed & maybe even merge if the way is cleared in the ensuring discussion here.  cc @Crypto2099

CIP-0107/README.md
rphair: p.s. should be considered along with @Crypto2099's https://github.com/cardano-foundation/CIPs/pull/635#discussion_r1424224364.

rphair (commented):
@Quantumplation revisiting this CIP was favourable at the editors' meeting tonight & we're looking forward to merging as soon as @Crypto2099 commits final adjustments as per the dialogue above (I'm ready to approve it after those update(s)).

rphair:
Per last meeting (# 91), expecting some final updates to this PR & likely merge by next meeting, so marking `Last Check` & putting on agenda: https://hackmd.io/@cip-editors/92 (cc @Crypto2099)

CIP-0107/README.md
```diff
@@ -0,0 +1,108 @@
...
+
+This extends CIP-13, which describes a Cardano URI scheme, with several more objects that would be useful to be able to assign addresses to, in particular blocks, transactions, transaction metadata, and transaction outputs.
+
+## Motivation: why is this CIP necessary?
+
+CIP-13 defined two initial URL authorities, for payment links and delegating to a stakepool.
+
+However, in a number of contexts, it would be useful to canonically link to other Cardano objects, such as:
+- Providing links to a transaction, to be opened in a wallet or chain explorer of the users choice
+- To provide richly interconnected metadata, such as in [CIP-100](https://github.com/cardano-foundation/CIPs/pull/556)
```
Ryun1, line 25:
```suggestion
- To provide richly interconnected metadata, such as in [CIP-100](https://github.com/cardano-foundation/CIPs/tree/master/CIP-0100)
```

Ryun1:
In the pull request, it is probably a good idea to update the reference to this proposal within CIP-100 - [here](https://github.com/cardano-foundation/CIPs/blob/master/CIP-0100/README.md?plain=1#L88C18-L88C78).

CIP-0107/README.md
Crypto2099, line 35:
```suggestion
web+cardano://block?hash=c6a8976125193dfae11551c5e80a217403d953c08ebbd69bba904d990854011f
web+cardano://block?height=12345678890
```

CIP-?/README.md
Crypto2099: I am content to leave the `height=` syntax in given that we have provided a disclaimer that this cannot/should not be trusted within the rollback horizon of the chain.

Crypto2099 (approved):
I am happy to see this one FINALLY get merged if @Quantumplation can take a quick glance at the last outstanding/remaining issues (examples and minor cross-linking suggestions).

Quantumplation:
@Crypto2099 I just applied the two suggestions; were there others you were thinking of? I haven't been able to keep up with this super well.

Crypto2099:
> @Crypto2099 I just applied the two suggestions; were there others you were thinking of? I haven't been able to keep up with this super well.

Looks like everything outstanding/pending has been resolved so I think this one is ready to merge IMO @Ryun1 @rphair

rphair (approved):
Thanks @Crypto2099 @Quantumplation for pushing this one through. 👏

PR #638, opened 2023-12-14
CPS-0013? | Better builtin data structures in Plutus

michaelpj:
Based on some discussions with Pi and others.

Some tasks today are difficult because we don't have the data structures to do them efficiently. This CPS outlines a couple of examples and gestures towards some possible solutions (which will require a proper CIP).

-----------------------------------------
[Rendered Version](https://github.com/michaelpj/CIPs/blob/master/CIP-%3F%3F%3F%3F/README.md)

### DIFF
--- a/CIP-????/README.md
+++ b/CIP-????/README.md
@@ -1,0 +1,97 @@
+---
+CPS: ????
+Title: Lack of builtin data structures with good asymptotic performance in Plutus Core
+Status: Proposed
+Category: Plutus
+Authors:
+    - Michael Peyton Jones <michael.peyton-jones@iohk.io>
+    - Philip DiSarro <philip.disarro@iohk.io>
+    - Pi Lanningham <pi@sundaeswap.finance>
+Discussions:
+Created: 2023-12-14
+License: CC-BY-4.0
+---
+
+## Abstract
+
+Plutus Core lacks builtin data structures with good asymptotic performance for some use cases.
+
+## Problem
+
+Plutus Core has a few builtin data structures, but these are mostly used to make a minimally adequate representation of the `Data` type. 
+It does not have builtin data structures optimized for performance.
+
+Users can implement their own data structures (since Plutus Core is an expressive programming language), but in practice this has not happened much. 
+In particular, we will focus on two examples here:
+
+1. Arrays with constant-time lookup
+2. Maps with logarithmic-time lookup (also Sets, but we can treat them as a special case of Maps)
+
+Both of these are difficult to implement in Plutus Core:
+
+1. Arrays are (we believe) impossible without some kind of primitive with constant-time lookup
+2. Maps are possible but are typically moderately complex data structures which require a lot of code, and this has not been done in practice
+
+## Use cases
+
+### Arrays
+
+#### Order matching
+
+A common pattern in DEXs is to have a list of inputs/outputs to match up in a datum. 
+For example, we might have:
+```
+inputIdxs :: BuiltinList Integer
+outputIdxs :: BuiltinList Integer
+```
+We then want to go through these lists, looking up the corresponding inputs and outputs and check some property (e.g. that the value is directly transferred from one to the other).
+
+This requires a quadratic amount of work, which puts a low ceiling on how many orders can be processed at once. 
+Empirically, many are capped at about 30, whereas if they were limited only by the amount of space in the transaction for inputs and outputs the limit would be hundreds.
+
+If we had arrays with constant time indexing, we could make this linear instead (we would still need to do a linear amount of work to create arrays for the transaction inputs and outputs from the lists in the script context).
+
+#### `Constr` arguments
+
+The `Data` type has a `Constr` alternative which is used for encoding datatype constructors. 
+This is used for encoding the script context, and is used by languages such as Aiken extensively for representing user-defined datatypes also.
+
+The fields of the constructor are encoded in a list; hence to access a particular field the compiled code needs to do a linear amount of work. 
+If the arguments to a `Constr` were an array, we could access the fields in constant time.
+
+### Maps
+
+#### Operations on `Value`
+
+The `Value` type is a nested map: it is a map from bytestrings (representing policy IDs) to maps from bytestrings (representing token names) to integers (representing quantities).
+Since map operations are currently linear, this means that even simple operations like checking whether one value is less than another can have quadratic cost.
+
+This would be much better if map operations were logarithmic cost.
+
+#### Indexing by party
+
+Many applications have a known set of participants identified by some bytestring, typically a public key. 
+It is therefore natural to store per-party state in a map indexed by the party identifier.
+
+Since map operations are currently linear, this is needlessly expensive and imposes a limit on the number of parties.
+
+## Goals
+
+1. Reduce the cost of operations on `Value` by a factor of 2-10
+2. Reduce the cost of a matching algorithm such that we can handle hundreds of matches for the same cost it currently takes to do 30.
+
+## Open questions
+
+- Can we implement a set/map data structure in Plutus Core code that has acceptable performance and doesn’t require too much size overhead?
+- Do we need generic maps or is a map-from-bytestring sufficient? What about map-from-integer?
+    - Generic maps are harder since we typically need to know how to order the key type
+- Is an array type useful even if it is immutable?
+    - We are unlikely to be able to offer mutable arrays
+- Are builtin data structures useful enough even if they can only contain builtin types?
+    - This would mean that complex data structures would have to be stored inside arrays as `Data`, rather than using Scott encoding or sum-of-products representation
+- Can we feasibly change the structure of the builtin `Data` type so that `Constr` arguments are in an array?
+    - We would need to retain both versions for backwards compatibility
+    
+## References
+
+- https://x.com/Quantumplation/status/1733298551571038338?s=20
--- a/CIP-????/README.md
+++ b/CIP-????/README.md
@@ -1,6 +1,6 @@
 ---
 CPS: ????
-Title: Lack of builtin data structures with good asymptotic performance in Plutus Core
+Better builtin data structures in Plutus
 Status: Proposed
 Category: Plutus
 Authors:
--- a/CIP-????/README.md
+++ b/CIP-????/README.md
@@ -1,6 +1,6 @@
 ---
 CPS: ????
-Better builtin data structures in Plutus
+Title: Better builtin data structures in Plutus
 Status: Proposed
 Category: Plutus
 Authors:
--- a/CIP-????/README.md
+++ b/CIP-????/README.md
@@ -1,5 +1,5 @@
 ---
-CPS: ????
+CPS: 13
 Title: Better builtin data structures in Plutus
 Status: Proposed
 Category: Plutus
--- a/CIP-????/README.md
+++ b/CIP-????/README.md
@@ -39,6 +39,8 @@ Both of these are difficult to implement in Plutus Core:
 #### Order matching
 
 A common pattern in DEXs is to have a list of inputs/outputs to match up in a datum. 
+In some cases the order is highly significant, e.g. earlier orders should be processed first, and the outcome of processing an earlier order may affect later ones.
+
 For example, we might have:
 ```
 inputIdxs :: BuiltinList Integer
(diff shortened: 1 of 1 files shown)


CIP-????/README.md
```diff
@@ -0,0 +1,97 @@
+---
+CPS: ????
```
rphair, line 2: The containing folder is currently `CIP-????` so can you please rename it to `CPS-`something... preferably something unique e.g. `data-structures` since this is from your `master` branch & you might want to submit another one?

CIP-????/README.md
```diff
@@ -0,0 +1,97 @@
...
+Title: Lack of builtin data structures with good asymptotic performance in Plutus Core
```
```suggestion
Better builtin datastructures in Plutus
```
The current PR title sounds both broad and specific enough to cover this: if not this one, we would need a final title of around this length.  Your stating the "problem" in terms of what we are hoping to achieve ("Better XXX") I think works well here.

CIP-????/README.md
```diff
@@ -0,0 +1,97 @@
...
+- Is an array type useful even if it is immutable?
+    - We are unlikely to be able to offer mutable arrays
+- Are builtin data structures useful enough even if they can only contain builtin types?
+    - This would mean that complex data structures would have to be stored inside arrays as `Data`, rather than using Scott encoding or sum-of-products representation
+- Can we feasibly change the structure of the builtin `Data` type so that `Constr` arguments are in an array?
+    - We would need to retain both versions for backwards compatibility
+    
+## References
+
+- https://x.com/Quantumplation/status/1733298551571038338?s=20
```
rphair, line 97:
```suggestion
```
We don't have a section for this in the CPS template... to me the discussion at this link seems split between _Problem_ and _Use cases_, so could it be linked in one or both those places as appropriate?

rphair (commented):
Very nice to see such a proposal sourced in cooperation with the community 🤓

Quantumplation:
Thanks for taking the time to compile this @michaelpj!

Not sure if this is appropriate for a CPS, but in the spirit of inspiring some CIPs to solve this CPS, let me document a few of the rough ideas we kicked around on that twitter thread:

# Length-prefixed arrays with constant time lookup

New builtin type `array`, with a `length` and `type`, such as:
```
(con (array 1 integer) [1])
(con (array 3 string) ["1", "2", "3"])
(con (array 7 data) [I 1, B aabb, I 2, I 3, B ddee, B ffgg, I 7])
```

New builtin function `arrayLength` that returns the length of the array

New builtin function `newArray` that constructs a new array of a given length
```
(builtin.newArray 5 data)

evalutes to `(con (array 5 data) [bottom, bottom, bottom, bottom, bottom])
```
(or perhaps you provide the initial value?)

New builtin function `index`, which retreives the `nth` item, or faults if its out of bounds
```
(builtin.index 3 (con (array 7 data) [I 1, B aabb, I 2, I 3, B ddee, B ffaa, I 7]))

evalutes to `I 3`
```

New builtin function `set` that returns a new array, with the `nth` item replaced by a new value, or faults if its out of bounds
```
(builtin.set 3 (B beef) (con (array 7 data) [I 1, B aabb, I 2, I 3, B ddee, B ffgg, I 7]))

evalutes to (con (array 7 data) [I 1, B aabb, I 2, B beef, B ddee, B ffgg, I 7])
```

## Why is this useful? 
There are several good candidates in the script context that are "index heavy" (most notably the inputs and outputs, but also withdrawals) that would benefit from being made into arrays. Even if not, a single linear pass to construct the array, and then constant time lookups after that, this would save quadratic work in many contracts.

GADTs from languages that compile to UPLC could compile structs to "Constructors with one field, which is an array of the logical fields of the struct", and then any field access is a `builtin.head` followed by a constant time access for the field we need.

# Associative Arrays

I have less fidelity here, so either a new builtin or a new Plutus Data type that got accelerated with a hashmap implementation; ideally, one that mapped directly onto CBOR maps for efficient encoding.

Would likely need builtins for looking up by key, and inserting (producing a new map with the entry added).

## Why is this useful?
Structs could compile down to maps, for log (with very small constants) time lookup of fields.
Datums / Redeemers could contain maps from address to value, or maps from txInput to action to take on that input, etc.

# Preserve the order of inputs from the transaction

This one is kind of an orthogonal / out of the box solution, but is highlighted by the struggles in this CPS; in particular, the best optimization is just to not do the work at all anymore, and a huge source of indexing slowdown is having to re-sort the transaction inputs in the first place.

CIP-????/README.md
michaelpj: um, I very much picked the wrong branch here, oops :facepalm: I guess I'll leave it for now :(

CIP-????/README.md
michaelpj: I'm not sure. I think the problem _is_ broader, I think it's jumping to conclusions a bit to structure it directly around the solution we're currently thinking of!

CIP-????/README.md
michaelpj: Maybe we should add a `References` section. It's pretty useful.

Ryun1: Good suggestion, see [Issue #731](https://github.com/cardano-foundation/CIPs/issues/731).

CIP-????/README.md
rphair: OK sounds like we're good to go with the broad, brief title (our CIP titles read more like newspaper headlines than academic papers)

CIP-????/README.md
rphair: @michaelpj no problem here for now... but if another CIP comes in on the `master` branch it will have to use a different subdirectory name 😅

CIP-????/README.md
rphair: ... specifically @michaelpj we are talking about restoring the removed `## References` section in https://github.com/cardano-foundation/CIPs/issues/731#issuecomment-1874421788 and with consensus will add it back in via https://github.com/cardano-foundation/CIPs/pull/730.

CIP-????/README.md
```diff
@@ -0,0 +1,97 @@
...
+
+The fields of the constructor are encoded in a list; hence to access a particular field the compiled code needs to do a linear amount of work. 
+If the arguments to a `Constr` were an array, we could access the fields in constant time.
+
+### Maps
+
+#### Operations on `Value`
+
+The `Value` type is a nested map: it is a map from bytestrings (representing policy IDs) to maps from bytestrings (representing token names) to integers (representing quantities).
+Since map operations are currently linear, this means that even simple operations like checking whether one value is less than another can have quadratic cost.
```
zliu41, line 67: > Since map operations are currently linear

To be precise, operations that should be logarithmic are currently linear (e.g., insertion, deletion, member), whereas operations should be linear (or nlogn) are currently quadratic (e.g., union)

CIP-????/README.md
```diff
@@ -0,0 +1,97 @@
...
+2. Reduce the cost of a matching algorithm such that we can handle hundreds of matches for the same cost it currently takes to do 30.
+
+## Open questions
+
+- Can we implement a set/map data structure in Plutus Core code that has acceptable performance and doesn’t require too much size overhead?
+- Do we need generic maps or is a map-from-bytestring sufficient? What about map-from-integer?
+    - Generic maps are harder since we typically need to know how to order the key type
+- Is an array type useful even if it is immutable?
+    - We are unlikely to be able to offer mutable arrays
+- Are builtin data structures useful enough even if they can only contain builtin types?
```
zliu41, line 90: `Value` alone should make this useful. Also Aiken users would certainly answer "yes".

CIP-????/README.md
```diff
@@ -0,0 +1,97 @@
...
+
+A common pattern in DEXs is to have a list of inputs/outputs to match up in a datum. 
+For example, we might have:
+```
+inputIdxs :: BuiltinList Integer
+outputIdxs :: BuiltinList Integer
+```
+We then want to go through these lists, looking up the corresponding inputs and outputs and check some property (e.g. that the value is directly transferred from one to the other).
+
+This requires a quadratic amount of work, which puts a low ceiling on how many orders can be processed at once.
```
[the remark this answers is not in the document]
Quantumplation: For example, the ledger reorders transaction inputs and reference inputs from what is on-chain, sorting them by transaction ID and then by index.

And for many cases (such as processing DEX orders), the order you process the list in is important, as each one impacts how the next one is evaluated. So we provide a list in the redeemer specifying which order to process the inputs in, and thus iterating over the list in that order is quadratic (or near quadratic) time.

CIP-????/README.md
```diff
@@ -0,0 +1,97 @@
...
+```
+inputIdxs :: BuiltinList Integer
+outputIdxs :: BuiltinList Integer
+```
+We then want to go through these lists, looking up the corresponding inputs and outputs and check some property (e.g. that the value is directly transferred from one to the other).
+
+This requires a quadratic amount of work, which puts a low ceiling on how many orders can be processed at once. 
+Empirically, many are capped at about 30, whereas if they were limited only by the amount of space in the transaction for inputs and outputs the limit would be hundreds.
+
+If we had arrays with constant time indexing, we could make this linear instead (we would still need to do a linear amount of work to create arrays for the transaction inputs and outputs from the lists in the script context).
```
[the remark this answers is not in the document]
michaelpj: This is assuming we didn't change Data.

CIP-????/README.md
michaelpj: Yes, I think those two are compelling

CIP-????/README.md
colll78: If we had a list with O(1) index lookup then `valueOf` would be constant time in the majority of DApps in practice. This is because every robust DApp already explicitly prevents non-protocol related tokens from being stored in protocol script UTxOs, and in the vast majority of cases the validator will have access to the currency symbol of all the protocol tokens that it must check for. In all such cases you can just predetermine the indices for the relevant protocol tokens and then use `elemAt index valuePairList` instead of `valueOf`

michaelpj:
@rphair I think this is ready for review

CIP-????/README.md
rphair: I'm preparing a PR that will document optional sections like this in the CPS template the way we added them for CIPs in #730.  We should have added it there, but by the time that discussion concluded I couldn't recall any CPS examples.  I'll have that PR for CIP-9999 submitted (and maybe merged) before this in final review, and in any case I believe a `## References` section by community consensus by editor unanimity is OK.

BUT @michaelpj @Quantumplation I don't think this link belongs just hanging here.  People not present in that Twitter discussion will have difficulty pulling the relevant elements out of it... so can you
- add some introduction about what is at this link?
- then consider moving it to _Problem_ or _Goals_ according to what this discussion most appropriately defines?

rphair (commented):
@michaelpj  all looks generally in order to me.  Have added it to CIP agenda # 81 for Review this time, since at meeting # 79 it only got Triage.  I expect we'll be able to assign it a number at the meeting & I'm ready to green-check this soon after that's done... unless any major outcry or new unresolved issues.  cc @Quantumplation

To avoid confusion for reviewers, can you rename the containing directory to something beginning with `CPS-` and re-link to the latest file in your branch in your original comment up top?

Also we agreed `## References` is OK but I made a https://github.com/cardano-foundation/CIPs/pull/638#discussion_r1471403744 that this item currently looks out of place there & needs to be defined a bit more & maybe moved into a relevant main section.

**Note** this was supposed to be combined with https://github.com/cardano-foundation/CIPs/pull/638#discussion_r1471356114 but my Internet connection went boogey.

CIP-????/README.md
rphair: p.s. the official inclusion of _References_ in the CPS has been proposed in https://github.com/cardano-foundation/CIPs/pull/756.

Crypto2099 (approved):
This looks good and in order for a CPS to me. Thanks @rphair @michaelpj @Quantumplation for getting this together. Definitely worthy of a number at the next meeting at the very least!

Ryun1 (approved):
This appears to be a well written problem statement, with good discussions, happy to see a number applied then merged.

rphair (commented):
I"m also very happy with it but will wait until it gets a number (very likely in couple days' time after meeting) in the usual cycle before I ✅ it for merge.

CIP-????/README.md
```diff
@@ -0,0 +1,102 @@
+---
+CPS: ????
```
```suggestion
CPS: 13
```

rphair (commented):
@michaelpj this has passed editorial review but zooming ahead too quickly to merge now.  Please update the number here & in your branch (and correcting the document link in your OP please)... it's on `Last Check` now for next meeting agenda, if not merged by popular request & consensus sooner 🚀

I think https://github.com/cardano-foundation/CIPs/pull/638#discussion_r1471426353 still needs to be resolved although I believe that's a document quality & not a "standards" issue & therefore up to @michaelpj @Quantumplation about whether it needs to be fixed; so I won't let the merge be contingent upon it.
- TL;DR (already documented above) I think our body of CIPs shouldn't have `References` which are links of unknown significance to social discussions included without context: our CIPs are generally written at a higher calibre than that.

colll78:
https://github.com/cardano-foundation/CIPs/pull/767#issuecomment-1977363621

Here are benchmarks that are relevant to the discussion of the benefits of indexed data-structures. The illustrate how incredibly expensive recursive lookups are currently.

PR #640, opened 2023-12-16
CPS-???? | CIP67 prefixes for validity/auth twinned tokens

waalge:
[cps](https://github.com/waalge/CIPs/tree/waagle/cps-x/CPS-xxxx)

Discussions welcome. 

---

(my terminal is unhappy with directory names including `?`)


rphair:
thanks @waalge ... `?` characters definitely not required in directory name, `x` or `X` is fine & so would be a short descriptive name for your proposal.  But in any case please rename your document to `README.md` and also please link this file into your OP above so we can click into a readable version easily 😎

rphair:
Closing since draft still not marked "Ready for review" in 8 months.  This would generally be a sign that a CIP/CPS is not going to be followed through the review process... but if @waalge you still intend to pursue this, please feel free to update the document so it's ready for peer / editor review & then reopen this PR.

waalge:
Sure. I don't think this deserves a full independent cip. Is there a way to make suggested additions to existing CIPs?

rphair:
My guess @waalge is that what you've written could be something like a "best practice" document.  Maybe @Crypto2099 would please have a quick look to see if this might be an addendum to CIP-0067... if so then we can reopen it and mount the usual review process.

PR #645, opened 2023-12-20
CIP-0006 | Adjust preamble and structure w.r.t CIP-0001

### LINKED ISSUE #644: CIP-0006 | Adjust preamble and structure w.r.t CIP-0001
Fixed by #645.

rphair:
Fixes #644.

([edited version](https://github.com/rphair/CIPs/blob/remediate-0006/CIP-0006/README.md))

### DIFF
--- a/CIP-0006/README.md
+++ b/CIP-0006/README.md
@@ -21,7 +21,7 @@ This CIP defines the concept of extended metadata for pools that is referenced f
 
 ## Motivation: why is this CIP necessary?
 
-As the ecosystem around Cardano stake pools proliferate so will the desire to slice, organize and search pool information dynamically.  Currently the metadata referenced on chain provides 512 bytes that can be allocated across the four information categories ([delegation-design-specification Section 4.2)](https://github.com/input-output-hk/cardano-ledger/releases/latest/download/shelley-delegation.pdf):
+As the ecosystem around Cardano stake pools proliferate so will the desire to slice, organize and search pool information dynamically.  Currently the metadata referenced on chain provides 512 bytes that can be allocated across the four information categories ([delegation-design-specification Section 4.2)](https://github.com/IntersectMBO/cardano-ledger/releases/latest/download/shelley-delegation.pdf):
 
 | key           | Value                                |  Rules  |
 | ---           | ---                                  |  ---  |
--- a/CIP-0006/README.md
+++ b/CIP-0006/README.md
@@ -89,7 +89,7 @@ This re-registration of the main metadata file with the `extData.vkey` and the t
 
 ## Rationale: how does this CIP achieve its goals?
 
-In the following we describe a first minimal version of the extended JSON file format
+In the following we describe a first minimal version of the extended JSON file format.
 
 Since this extended metadata file can be updated at any time by the pool operator, a **serial number** is useful for consuming applications and services to identify updates.
 
--- a/CIP-0006/README.md
+++ b/CIP-0006/README.md
@@ -1,12 +1,17 @@
 ---
 CIP: 6
 Title: Stake Pool Extended Metadata
-Authors: Markus Gufler <gufmar@gmail.com>, Mike Fullman <mike@fullman.net>
-Comments-URI: https://github.com/cardano-foundation/CIPs/pull/15
-Status: Draft
-Type: Standards
+Status: Active
+Category: Tools
+Authors:
+  - Markus Gufler <gufmar@gmail.com>
+  - Mike Fullman <mike@fullman.net>
+Implementors:
+  - pooltool.io
+  - cexplorer.io
+Discussions:
+  - https://github.com/cardano-foundation/CIPs/pull/15
 Created: 2020-07-20
-Updated: 2020-11-24 2nd key-pair for validation 2021-02-08 json schema
 License: CC-BY-4.0
 ---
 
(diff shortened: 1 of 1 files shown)


CIP-0006/README.md
```diff
@@ -79,7 +87,7 @@ The operator now:
 
 This re-registration of the main metadata file with the `extData.vkey` and the two URLs is only necessary once. Afterwards, the operator can update his extended metadata at any time, compute the new signature with the `cardano-cli stake-pool rawdata-sign` command, and publish both files at the existing `extDataUrl` and `extSigUrl`.
 
-### Extended Metadata structure
+## Rationale: how does this CIP achieve its goals?
 
 In the following we describe a first minimal version of the extended JSON file format
```
Ryun1, line 92:
```suggestion
In the following we describe a first minimal version of the extended JSON file format.
```
grammar :)

CIP-0006/README.md
```diff
@@ -1,20 +1,25 @@
 ---
 CIP: 6
 Title: Stake Pool Extended Metadata
-Authors: Markus Gufler <gufmar@gmail.com>, Mike Fullman <mike@fullman.net>
-Comments-URI: https://github.com/cardano-foundation/CIPs/pull/15
-Status: Draft
-Type: Standards
+Status: Active
+Category: Tools
```
Ryun1, line 5:
```suggestion
Category: Metadata
```
Happy for it to be either 🤓

CIP-0006/README.md
```diff
@@ -29,7 +34,10 @@ Many additional attributes can be envisioned for future wallets, pool explorers,
 
 ## Specification
 
+> **Note** Updated: 2020-11-24 2nd key-pair for validation 2021-02-08 json schema
+
 ### On Chain referenced (main) metadata file
```
Ryun1, line 39: Are we going to enforce the any CIP which has on-chain data must provide a CDDL file (from #563) - retroactively?
Probably not practical.

Ryun1 (approved):
LGTM 

See few small comments.

CIP-0006/README.md
rphair: I'm glad you're happy with `Tools` then 🤓 since I already thought about which of these choices to make.  For other other CIPs tagged `Metadata` that metadata, even when off-chain, is associated with the chain somehow & in this case it isn't.

Also the target is a broad spectrum of operational procedures and tools as summarised best in this sentence (which itself could be a good description of the `Tools` category):
> Many additional attributes can be envisioned for future wallets, pool explorers, and information aggregators.

CIP-0006/README.md
rphair: I think it would make sense to demand CDDL here if there had been a CIP for the original on-chain pool metadata specification, which there wasn't: as far as I know that was only ever documented in apocrypha like SPO setup procedures.

So providing CDDL for the extended structure, but not the original structure, doesn't generally solve that problem.

The alternative would be to broaden the scope of this CIP so it includes _both_ the original metadata spec _and_ the extended spec.  That would avoid having to write an "informational" CIP over 4 years after this method was put into practice: looking a bit silly perhaps.  I would still be in favour of it though: but only in a separate PR because it's beyond the scope of the CIP1 remediation effort.

CIP-0006/README.md
Ryun1: > For other CIPs tagged Metadata, their metadata - even when off-chain - is associated with the chain somehow, and in this case it isn't.

Makes sense to me! agreed.

CIP-0006/README.md
```diff
@@ -1,20 +1,25 @@
...
 ---
 
 ## Abstract
 
 This CIP defines the concept of extended metadata for pools that is referenced from the pool registration data stored on chain.
 
-## Motivation
+## Motivation: why is this CIP necessary?
 
 As the ecosystem around Cardano stake pools proliferate so will the desire to slice, organize and search pool information dynamically.  Currently the metadata referenced on chain provides 512 bytes that can be allocated across the four information categories ([delegation-design-specification Section 4.2)](https://github.com/input-output-hk/cardano-ledger/releases/latest/download/shelley-delegation.pdf):
```
Ryun1, line 24:
```suggestion
As the ecosystem around Cardano stake pools proliferate so will the desire to slice, organize and search pool information dynamically.  Currently the metadata referenced on chain provides 512 bytes that can be allocated across the four information categories ([delegation-design-specification Section 4.2)](https://github.com/IntersectMBO/cardano-ledger/releases/latest/download/shelley-delegation.pdf):
```
do we want to do this?

CIP-0006/README.md
rphair: @Ryun1 the period (rather than a colon) isn't a grammar mistake.  Mainly, it's a complete and grammatically correct sentence.  Also, the colon would normally be used if the "following" were following immediately, which it doesn't... instead it's distributed through this whole section & that's why I used the `.` rather than the `:`. 😛

CIP-0006/README.md
rphair: @Ryun1 @Crypto2099 @KtorZ two possible ways of answering your question (although definitely **yes** we want to correct the out of date links):

- We want to change these links to point to IntersectMBO because the IOHK links have become non-canonical and eventually the redirects will stop working (or get lost behind further redirects someday).
- We want to use the versions in our own repos since the academic releases of these technical papers are generally paywalled or at least require academic accounts to view them.

CIP-0006/README.md
Ryun1: youre right!!

PR #650, opened 2023-12-20
CIP-0009 | Adjust preamble and structure w.r.t CIP-0001

rphair:
Fixes #649.

([edited version](https://github.com/rphair/CIPs/blob/remediate-0009/CIP-0009/README.md))

PR #370, opened 2022-11-08
Editors meeting #57 housekeeping

KtorZ:
- :round_pushpin: **Turn CIP-0060 as active in the top-level README.**
  
- :round_pushpin: **Turn CIP-0055 as 'active', rename CIP-0009, CIP-00028 & CIP-0055 for clarity, and link them together.**
  
- :round_pushpin: **Add new proposals from triage to the top-level README table**

PR #676, opened 2023-12-21
CIP-0028 | Adjust preamble and structure w.r.t CIP-0001

### LINKED ISSUE #675: CIP-0028 | Adjust preamble and structure w.r.t CIP-0001
Fixed by #676.

rphair:
Fixes #675.

@kevinhammond as with https://github.com/cardano-foundation/CIPs/pull/650 I've resolved the discrepancy in licensing agreements in favour of Apache (CC was indicated in the YAML header).

Editors: we need to keep an eye on how these Protocol Parameter CIPs are stacking on top of each other.  The last solid resolution I remember (before @Ryun1 @Crypto2099 arrived) was that protocol parameter CIPs would describe differences with the previous ledger era, rather than a complete re-enumeration of parameters... but a lot of this material feels already stated in CIP-0009.

So although solving any issues of reference & redundancy is beyond the scope of the #389 "Remediation" (and therefore doesn't have to be attended to in this PR) I still think we should keep an eye on this issue (with help from @kevinhammond and @WhatisRT I hope) because otherwise this mangled history of protocol parameters is going to end up fossilised.

In this light I've **removed** a forward reference to CIP-0055 which was shoe-horned in here with #370 apparently after debate at at CIP meeting, since this would therefore be in direct conflict with the decision that parameter updates should be cumulative (and therefore containing only backward references):
```
### Change Log

See [CIP-0055: Protocol Parameters (Babbage Era)](../CIP-0055).
```
([updated proposal](https://github.com/rphair/CIPs/blob/remediate-0028/CIP-0028/README.md))

### DIFF
--- a/CIP-0028/README.md
+++ b/CIP-0028/README.md
@@ -292,4 +292,4 @@ script failure if the collateral was not available).
 
 ## Copyright
 
-This CIP is licensed under [Apache-2.0][https://www.apache.org/licenses/LICENSE-2.0].
+This CIP is licensed under [Apache-2.0](https://www.apache.org/licenses/LICENSE-2.0).
--- a/CIP-0028/README.md
+++ b/CIP-0028/README.md
@@ -1,24 +1,27 @@
 ---
 CIP: 28
 Title: Protocol Parameters (Alonzo Era)
-Authors: Kevin Hammond <kevin.hammond@iohk.io>
 Status: Active
-Type: Informational
+Category: Ledger
+Authors:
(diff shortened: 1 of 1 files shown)


Ryun1 (approved):
LGTM

Crypto2099 (approved):
LGTM

rphair:
Merging this one with respect to the CIP-0001 "remediation" without an answer to my last question because, although consistency (redundancy vs. reference) between parameter update CIPs _is_ important, that question isn't part of this remediation effort.

PR #666, opened 2023-12-20
CIP-0021 | Adjust preamble and structure w.r.t CIP-0001

rphair:
Fixes #665.

([updated proposal](https://github.com/rphair/CIPs/blob/remediate-0021/CIP-0021/README.md))

PR #687, opened 2023-12-21
CIP-0021 | Conway update

janmazak:


rphair (approved):
@Ryun1 @Crypto2099 I know I said I wouldn't tag you over the holidays, but also I thought it would be better in the long run to avoid a merge conflict with https://github.com/cardano-foundation/CIPs/pull/666 by promptly reviewing & merging either this one or that one.

Crypto2099 (approved):
LGTM

rphair:
@janmazak we've updated the CIP document this is based on recently (according to new specs in #389)... can you rebase your changes on the newly merged document?

janmazak:
I've rebased it and made a couple of additional small changes, so please review it again to make sure I've not missed anything.

rphair (approved):
@janmazak still looks good to me & matches the changes I recall from before I edited the code base.

I would be OK with merging this ~~now~~ whenever (https://github.com/cardano-foundation/CIPs/pull/687#issuecomment-1883644119) but would also want @KtorZ to look it over first... maybe also with @Ryun1 double checking that key (or cert?) names and/or official descriptions aren't also about to change when #737 is merged (which should ~~also~~ be soon).

Ryun1:
Id vote to keep this PR open for a little longer, whilst the HWW implementations are being finished off.
Maybe wait until SanchoNet is feature complete?

I dont think there is a rush to merge it?

Ryun1:
On the Conway parts, I think this addition is wholly complete.

Past that, I do have two suggestions;
- With the app split on Ledger devices, it probably makes sense to list out the device specific restrictions here.
- It might make sense to add a section for a light summary Changelog, using the new optional sections. This would just make it much easier to transverse the history of this document.

rphair:
@janmazak I would also be ready to merge this, between the CIP meetings or at the next meeting, once you can follow or refute @Ryun1's suggestions in https://github.com/cardano-foundation/CIPs/pull/687#issuecomment-1985538516.

CIP-0021/README.md
```diff
@@ -94,34 +104,36 @@ Since multiassets (`policy_id` and `asset_name`) are represented as maps, both n
...
-Certificates of type `genesis_key_delegation` and `move_instantaneous_rewards_cert` are not supported and must not be included.
+Certificates of the following types are not supported and must not be included:
+* `genesis_key_delegation`
+* `move_instantaneous_rewards_cert`
+* `stake_vote_deleg_cert`
+* `stake_reg_deleg_cert`
+* `vote_reg_deleg_cert`
+* `stake_vote_reg_deleg_cert`
 
 If a transaction contains a pool registration certificate, then it must not contain:
```
Ryun1, line 115: why do we not add any protections like this around constitutional committee keys or DRep keys

janmazak: The main reason for pool registration certificate restrictions is that pool owners are signing them with stake keys, which also sign other elements of the transaction (e.g. withdrawals and other certificates), and that's a security risk: the transactions are not created by them and so they might contain, from their viewpoint, third-party elements --- they can't easily distinguish whether a key hash used in a withdrawal belongs to them or not, and a HW wallet does not see the whole tx at once, so cannot warn them that the same stake key hash is used in several places. 

With DRep or constitutional committee keys, only votes are signed, nothing else AFAIK.

gitmachtl: > With DRep or constitutional committee keys, only votes are signed, nothing else AFAIK.

I have not tested this, but when you do a DRep deregistration, you could also sneak in a vote too. Not sure what the protocol is doing with that.

Ryun1: DRep keys are used for votes, registration certs, update certs and resignation certs.

CC Hot keys are only used for voting.
CC Cold keys are used for authorising hot cert and resign cert.

gitmachtl: @Ryun1 do you think its worth to check if there is a vote together with a cert in the transaction?

Ryun1: @gitmachtl 

to follow the precedent of how stake key signing is handled, yes it makes sense for the check.

But practically speaking, I don't think this is a significant issue, so IMO it is probably fine.
It is quite easy for a DRep to re-register.

gitmachtl: @Ryun1 i was more thinking in a way that if online services provide the feature to easy register as a DRep, if they could sneak in a vote for there proposal f.e. and the user would not see that directly.

Ryun1: yes this would be easy to do over a dApp.
but of course on the hardware wallet device it still shows the vote to the user, at signing time.

rphair:
@janmazak just following up on this for my last month's reports.  @Ryun1 are you OK with this now & does https://github.com/cardano-foundation/CIPs/pull/687/commits/a52b1ead9a626d4209745c8c8a4cde70ec05781c cover your most recent recommendations?

Ryun1 (approved):
Very detailed update, looking great.

PR #694, opened 2023-12-22
CIP-0037 | Adjust preamble and structure w.r.t CIP-0001

### LINKED ISSUE #693: CIP-0037 | Adjust preamble and structure w.r.t CIP-0001
Fixed by #694.

rphair:
Fixes #693.

I don't think this proposal really got much editorial attention prior to merging, and the attitude about RSS proposals (prior to the ruling by default that they would be `Inactive`) was "Why not merge it without a rigorous review... since declaring it `Active` would depend on a bureaucratic process which still remains undefined."

So somehow it missed our across-the-board standardisation of RSS proposals, which I've applied here including a `Ledger` category and the same `Path to Active` boilerplate I wrote & applied to the other RSS CIPs.

@CyberCyclone I had to haul what you'd written as _Path to Active_ and some other non-canonical sections into the _Motivation_ section since that is really the only place for inspirational writing about why we might need the CIP.

... This also means I had to write your _Abstract_ from scratch, so please double-check to ensure it's accurate with respect to your specification (which was not precisely formulated as such).

([edited proposal](https://github.com/rphair/CIPs/blob/remediate-0037/CIP-0037/README.md))

### DIFF
--- a/CIP-0037/README.md
+++ b/CIP-0037/README.md
@@ -14,7 +14,7 @@ License: CC-BY-4.0
 
 ## Abstract
 
-The pledge should be used to calculate the saturation point of a pool: setting a maximum delegation proportional to their total pledge.
+The pledge should be used to calculate the saturation point of a pool: setting a maximum delegation proportional to pledge.
 
 ## Motivation: why is this CIP necessary?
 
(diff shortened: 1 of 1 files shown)


Ryun1 (approved):
LGTM

nice abstract Robert :)

Crypto2099 (approved):
LGTM

CyberCyclone:
Sorry @rphair for the late response. The PR looked good to me. Thanks for fixing that up.

PR #696, opened 2023-12-22
CIP-0040 | Adjust preamble and structure w.r.t CIP-0001

### LINKED ISSUE #695: CIP-0040 | Adjust preamble and structure w.r.t CIP-0001
Fixed by #696.

rphair:
Fixes #695.

@SebastienGllmt @WhatisRT there was a title clash between a longer title in the CIP vs. a shorter title in the top level `README`.  I've ruled in favour of the longer title since I think it captures the point of the CIP better... but if that would cause any confusion we'll set them both to the shorter one.

([updated proposal](https://github.com/rphair/CIPs/blob/remediate-0040/CIP-0040/README.md))

### DIFF
--- a/CIP-0040/README.md
+++ b/CIP-0040/README.md
@@ -1,18 +1,24 @@
 ---
 CIP: 40
 Title: Explicit Collateral Output
-Authors: Sebastien Guillemot <seba@dcspark.io>, Jared Corduan <jared.corduan@iohk.io>, Andre Knispel <andre.knispel@iohk.io>
 Status: Active
-Type: Standards
+Category: Ledger
+Authors:
+  - Sebastien Guillemot <seba@dcspark.io>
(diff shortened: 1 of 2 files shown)


WhatisRT:
Actually, I think `Collateral Output` would be a better title. `Explicit Collateral Output` somehow suggests that there is some 'implicit' collateral output, but that's not the case. This CIP proposed a collateral output, which didn't exist at all before.

rphair:
thanks @WhatisRT - this change just removes _Explicit_ from the CIP title now & keeps the shorter title in the CIP directory as before.

Crypto2099 (approved):
LGTM

Ryun1 (approved):
LGTM

KtorZ (approved):

PR #702, opened 2023-12-22
CIP-0052 | Adjust preamble and structure w.r.t CIP-0001

### LINKED ISSUE #701: CIP-0052 | Adjust preamble and structure w.r.t CIP-0001
Fixed by #702.

rphair:
Fixes #701.

@simonjohnthompson I've had to make some adjustments due to more rigorous formatting requirements and stricter requirements in anticipation of becoming `Active` since this was originally merged.

The main text shuffle was to move your _Appendix_ into the "Prerequisites" for your _Specification_.  This assumes these terms are essential to implementing the audit process exactly as envisioned in your Specification.
* If that assumption isn't true, or if the background material is cumbersome in that position, especially if common knowledge to auditors, this might be offloaded to a document like "Handbook for audits" that exists outside the CIP.

I had to delete the "example audit" since its web page is defunct and the Wayback Machine version doesn't run the right code to download the PDF.  If you have that document in the raw perhaps it can be linked into the CIP somehow.

Finally I do believe this should have a proper _Path to Active_ since neither auditors nor those who use audit services have provided the CIP repository with examples of CIP-0052 in audits which would qualify this CIP as `Active`.
* Hopefully with the qualification of usage that I have added, any auditors following CIP-0052 would be encouraged to submit their audits as evidence of `Active` status for this CIP.

([updated proposal](https://github.com/rphair/CIPs/blob/remediate-0052/CIP-0052/README.md))

### DIFF
--- a/CIP-0052/README.md
+++ b/CIP-0052/README.md
@@ -1,11 +1,16 @@
 ---
 CIP: 52
 Title: Cardano audit best practice guidelines
-Authors: Simon Thompson <simon.thompson@iohk.io>
-Comments-Summary: No comments yet.
-Comments-URI: https://github.com/cardano-foundation/CIPs/wiki/Comments:CIP-0052
 Status: Proposed
-Type: Process
+Category: Meta
+Authors:
+  - Simon Thompson <simon.thompson@iohk.io>
+Implementors: []
+Discussions: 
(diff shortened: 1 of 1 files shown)


Ryun1 (approved):
LGTM

Crypto2099 (approved):
LGTM

PR #586, opened 2023-08-29
CIP-0067, CIP-0068 | Add Clarification for Edits and Modification

Crypto2099:
- Some minor grammatical/typographical errors were fixed
- Reformatting some table blocks w/o any modification of content
- Update existing language in a few places to follow RFC 2119 format
- Add `documentation` property (URL) to the CIP-67 registry.json file. If we are going to require that future extensions of the CIP-68 standard be submitted via a separate CIP, then when documenting them within the context of the CIP-67 registry it's important for implementers to be able to find documentation on these asset label names and how to support them (if necessary).
- Add blocks clarifying the methods and mechanisms to submitting updates or extensions to CIP-68 standard following RFC 2119 format.

Happy to discuss and refine this as needed but consider this simply a first draft.

PR #715, opened 2023-12-23
CIP-0068 | Adjust preamble and structure w.r.t CIP-0001

### LINKED ISSUE #714: CIP-0068 | Adjust preamble and structure w.r.t CIP-0001
Fixed by #715.

rphair:
Fixes  #714.

@Crypto2099 had already done some good work to remediate this with respect to CIP-0001 in #586 but it was still left with non-standard major headings undefined in CIP-0001 (like `## References`) and the _Path to Active_ needed some items shuttled between Acceptance and Implementation.

In cases like this I would want to put References in the section where it is most applicable.  Usually that's the Specification which draws upon standards documented elsewhere, but sometimes that's Rationale because these references can be either reference implementations or points of contrast with approaches others have taken.  

In any case I would be open to discussion about where things get moved but for sake of parseability I think we need to stick to the half dozen major sections defined in CIP-0001 as the only major headings.

([updated proposal](https://github.com/rphair/CIPs/blob/remediate-0068/CIP-0068/README.md))

### DIFF
--- a/CIP-0068/README.md
+++ b/CIP-0068/README.md
@@ -17,6 +17,7 @@ Discussions:
   - https://github.com/cardano-foundation/CIPs/pull/471
   - https://github.com/cardano-foundation/CIPs/pull/494
   - https://github.com/cardano-foundation/CIPs/issues/520
+  - https://github.com/cardano-foundation/CIPs/pull/586
 Created: 2022-07-13
 License: CC-BY-4.0
 ---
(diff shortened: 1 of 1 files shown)


Ryun1 (approved):
LGTM

Crypto2099 (approved):
LGTM

rphair:
Since as of https://github.com/cardano-foundation/CIPs/pull/730#pullrequestreview-1807263973 I think we now have consensus that References are valid as an "optional section", I've unstuck it from where it had been demoted & put it back where it used to be.

PR #733, opened 2024-01-04
CIP-0111? | Web-Wallet Bridge - Wallet Transaction Caching

yHSJ:
dApps on Cardano have been taking advantage of "transaction chaining," where a transaction spends outputs of a previous transaction that has not yet settled on chain. While this leads to a very smooth user experience, it also means users are getting used to signing transactions which have confusing or incorrect transaction summaries in the wallet UI.

It also allows scammers to take advantage of that confusing or incorrect transaction summary to build malicious chained transactions and trick users into signing transactions they did not want to approve.

This proposal introduces transaction caching into the wallet and requires that wallets display a clear warning when unable to provide a complete transaction summary.

[Rendered](https://github.com/yHSJ/CIPs/tree/jshy/cache-txs/cip-smart-tx-signing)

### DIFF
--- a/cip-smart-tx-signing/README.md
+++ b/cip-smart-tx-signing/README.md
@@ -77,4 +77,4 @@ N/A
 
 ## Copyright
 
-[CC-BY-4.0]: https://creativecommons.org/licenses/by/4.0/legalcode
+This CIP is licensed under [CC-BY-4.0](https://creativecommons.org/licenses/by/4.0/legalcode).
--- a/cip-smart-tx-signing/README.md
+++ b/cip-smart-tx-signing/README.md
@@ -5,7 +5,7 @@ Status: Proposed
 Category: Wallets
 Authors:
     - Josh Marchand <josh@securitybot.info>
-Implementors: N/A
+Implementors: []
 Discussions:
  - https://github.com/cardano-foundation/CIPs/pull/733
 Created: 2023-12-22
--- a/cip-smart-tx-signing/README.md
+++ b/cip-smart-tx-signing/README.md
@@ -12,8 +12,6 @@ Created: 2023-12-22
 License: CC-BY-4.0
 ---
 
-# CIP-? Wallet Transaction Caching
-
 ## Abstract
 
 This CIP extends [CIP-0030 Cardano dApp-Wallet Web Bridge](https://github.com/cardano-foundation/CIPs/tree/master/CIP-0030/) and [CIP-0103? | Web-Wallet Bridge - Bulk Transaction Signing](https://github.com/cardano-foundation/CIPs/pull/587) to allow wallets to provide a complete and accurate summary of transactions that spend outputs that don't exist on chain yet when possible, and clearly communicate to the user risks when they cannot.
--- a/cip-smart-tx-signing/README.md
+++ b/cip-smart-tx-signing/README.md
@@ -18,7 +18,7 @@ This CIP extends [CIP-0030 Cardano dApp-Wallet Web Bridge](https://github.com/ca
 
 ## Motivation: why is this CIP necessary?
 
-Transaction chaining, building transactions that spend outputs of transactions that aren't settled on chain, allows dApps to quickly do several on-chain transactions instead of waiting for each indvidual transaction to settle before building and submitting the next one. This is a powerful feature that has allowed dApps to provide seamless experiences to users. However, due to the tranasaction data structure, wallets are unable to provide a complete and accurate summary of these chained transactions when prompting users for a signature.
+Transaction chaining, building transactions that spend outputs of transactions that aren't settled on chain, allows dApps to quickly do several on-chain transactions instead of waiting for each individual transaction to settle before building and submitting the next one. This is a powerful feature that has allowed dApps to provide seamless experiences to users. However, due to the transactions data structure, wallets are unable to provide a complete and accurate summary of these chained transactions when prompting users for a signature.
 
 A transaction input, which is a reference to a previous transaction output, is defined in the [babbage CDDL](https://github.com/IntersectMBO/cardano-ledger/blob/master/eras/babbage/impl/cddl-files/babbage.cddl) as follows:
 
--- a/cip-smart-tx-signing/README.md
+++ b/cip-smart-tx-signing/README.md
@@ -1,7 +1,7 @@
 ---
 CIP: ?
 Title: Wallet Transaction Caching
-Status: Active
+Status: Proposed
 Category: Wallets
 Authors:
     - Josh Marchand <josh@securitybot.info>
--- a/cip-smart-tx-signing/README.md
+++ b/cip-smart-tx-signing/README.md
@@ -1,5 +1,5 @@
 ---
-CIP: ?
+CIP: 111
 Title: Wallet Transaction Caching
 Status: Proposed
 Category: Wallets
--- a/cip-smart-tx-signing/README.md
+++ b/cip-smart-tx-signing/README.md
@@ -1,0 +1,82 @@
+---
+CIP: ?
+Title: Wallet Transaction Caching
+Status: Active
+Category: Wallets
+Authors:
+    - Josh Marchand <josh@securitybot.info>
+Implementors: N/A
+Discussions:
+ - N/A
+Created: 2023-12-22
+License: CC-BY-4.0
+---
+
+# CIP-? Wallet Transaction Caching
+
+## Abstract
+
+This CIP extends [CIP-0030 Cardano dApp-Wallet Web Bridge](https://github.com/cardano-foundation/CIPs/tree/master/CIP-0030/) and [CIP-0103? | Web-Wallet Bridge - Bulk Transaction Signing](https://github.com/cardano-foundation/CIPs/pull/587) to allow wallets to provide a complete and accurate summary of transactions that spend outputs that don't exist on chain yet when possible, and clearly communicate to the user risks when they cannot.
+
+## Motivation: why is this CIP necessary?
+
+Transaction chaining, building transactions that spend outputs of transactions that aren't settled on chain, allows dApps to quickly do several on-chain transactions instead of waiting for each indvidual transaction to settle before building and submitting the next one. This is a powerful feature that has allowed dApps to provide seamless experiences to users. However, due to the tranasaction data structure, wallets are unable to provide a complete and accurate summary of these chained transactions when prompting users for a signature.
+
+A transaction input, which is a reference to a previous transaction output, is defined in the [babbage CDDL](https://github.com/IntersectMBO/cardano-ledger/blob/master/eras/babbage/impl/cddl-files/babbage.cddl) as follows:
+
+```
+transaction_input = [ transaction_id : $hash32
+                    , index : uint
+                    ]
+```
+
+Therefore, in order for a wallet to know what assets are being spent by a transaction, it must have access to the transaction that the input is referencing. This is not a problem when the transaction inputs reference transactions that have settled on chain. However, when a transaction is being signed that spends outputs that don't exist on chain yet, the wallet cannot know what the spent UTxOs actually contain.
+
+A malicious actor could take advantage of this by creating two transactions that are chained together. The first transaction would be a transaction that spends a user's UTxOs and creates new UTxOs that are locked by the user's address. The second transaction would then spend these new UTxOs and send the funds to the attacker's address. The malicious actor would then prompt the user to sign both transactions, using [CIP 30's signTx](https://github.com/cardano-foundation/CIPs/blob/master/CIP-0030/README.md#apisigntxtx-cbortransaction-partialsign-bool--false-promisecbortransaction_witness_set) endpoint. The first transaction summary would show nothing malicious, and the user would thus sign it. The second transaction would show an incorrect or incomplete transaction summary, and because user's are used to seeing those summaries from other dApps, they may sign the second transaction, effectively sending all their assets to the attacker.
+
+Simply caching signed tranasctions before they appear on chain and refusing to sign transactions that spend outputs that don't exist on chain yet would solve this problem. However, this would break some transaction chaining done by dApps, where they reference UTxOs that were created in a transaction that didn't involve the user but hasn't settled yet. Therefore, the best solution would be to cache signed transactions and display a clear warning to the user when a transaction includes inputs that reference outputs that can't be found.
+
+## Specification
+
+The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "SHOULD NOT", "RECOMMENDED", "MAY", and "OPTIONAL" in this document are to be interpreted as described in [RFC 2119](https://datatracker.ietf.org/doc/html/rfc2119).
+
+### Defintions
+
+**Wallet**: A wallet is any software that manages a user's private keys and provides basic cryptographic operations such as signing data, for the purpose of interacting with the Cardano blockchain.
+
+---
+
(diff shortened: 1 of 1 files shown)


cip-smart-tx-signing/README.md
```diff
@@ -0,0 +1,82 @@
...
+    - Josh Marchand <josh@securitybot.info>
+Implementors: N/A
+Discussions:
+ - N/A
+Created: 2023-12-22
+License: CC-BY-4.0
+---
+
+# CIP-? Wallet Transaction Caching
+
```
rphair, line 16:
```suggestion
```
CIP-0001 says "no H1 title` generally since this is constructed from the header in other contexts.

cip-smart-tx-signing/README.md
```diff
@@ -0,0 +1,82 @@
...
+### Acceptance Criteria
+
+In order for this standard to be considered 'Active', the following criteria must be met:
+
+- [ ] A thorough review by all relevant stakeholders
+- [ ] Implementation in at least two Cardano wallets
+
+### Implementation Plan
+
+N/A
```
rphair, line 78: What about demonstrating this in the extension namespace as a reference implementation?

cip-smart-tx-signing/README.md
```diff
@@ -0,0 +1,82 @@
...
+
+## Copyright
+
+[CC-BY-4.0]: https://creativecommons.org/licenses/by/4.0/legalcode
```
rphair, line 82:
```suggestion
## Copyright

This CIP is licensed under [CC-BY-4.0](https://creativecommons.org/licenses/by/4.0/legalcode).
```

rphair (commented):
Nice addition to the `Web-Wallet Bridge` family. 👶

klntsky:
I propose a simpler solution: extend `signTx` with an optional argument of type `[CborHex<Transaction>]` which would include all the relevant not-yet-onchain context.

This approach would also cover

> some transaction chaining done by dApps, where they reference UTxOs that were created in a transaction that didn't involve the user but hasn't settled yet

The wallets will always be able to reliably tell what is happening, including showing the entire chain (or, rather, DAG) of  txs to the user.

yHSJ:
> I propose a simpler solution: extend `signTx` with an optional argument of type `[CborHex<Transaction>]` which would include all the relevant not-yet-onchain context.


Yes, this was another idea that I had that is a totally valid solution, and may be better. The key difference would be that it is then the responsibility of the dApp to cache transactions instead of the responsibility of the wallet. I felt that it made more sense to have transactions cached at the wallet level.

Consider the case where a user is using multiple dApps at one time. The caching of transactions at the wallet level also allows wallets to 1) remove UTxOs that are consumed in cached transactions but not yet on chain from the getUtxos response and 2) not let a user sign a transaction if a UTxO in a proposed transaction has been consumed in a cached transaction, preventing some of the "cardano-y" errors that users run into and become frustrated by.

I'm not sure that those are necessarily good ideas, hence why they weren't included in this CIP, but the caching of transactions at the wallet level has multiple beneifts.

NicholasMaselli:
Adding comment of support because I saw this on twitter.

Yes we need some way for wallets to handle chained transactions. Now that there have been a few malicious wallet drainers in the ecosystem its very important for wallets to display exactly what is happening or put a warning (perhaps referring to chaining resources) if it cannot

cip-smart-tx-signing/README.md
yHSJ: The plan was to make this a fundamental change to the web-wallet bridge standards themselves. Leaving CIP-30's `signTx` and CIP-103's `signTxs` endpoint as is would leave the attack vector available malicious users and render this CIP largely unhelpful.

I am happy to rewrite the CIP to clarify that. I tried to make it clear that, if a wallet implements this CIP, the wallet should implement transaction caching for _any_ current or future transaction signing functionality.

rphair: I wouldn't suggest rewriting the CIP... this question has just been for my own understanding 🙏

klntsky:
@yHSJ 

> I felt that it made more sense to have transactions cached at the wallet level.

They will still be cached on the wallet side. dApps will be able to write to the cache by explicitly providing a set of utxos that should be considered spent after the user signs a given transaction. This is not, in fact, an approach different from caching, just a path to make caching work nicely with transaction chaining.

> The caching of transactions at the wallet level also allows wallets to 1) remove UTxOs that are consumed in cached transactions but not yet on chain from the getUtxos response and 2) not let a user sign a transaction if a UTxO in a proposed transaction has been consumed in a cached transaction, preventing some of the "cardano-y" errors that users run into and become frustrated by.

Both cases could also be implemented with the extra argument approach.

`1)` only concerns wallet's address utxos, and the wallet can infer which of the utxos of the transactions provided via the extra argument are owned by it, and remove them from the set of available utxos

cip-smart-tx-signing/README.md
```diff
@@ -0,0 +1,80 @@
...
+
+If a user signs a transaction through the [CIP-30 signTx](https://github.com/cardano-foundation/CIPs/blob/master/CIP-0030/README.md#apisigntxtx-cbortransaction-partialsign-bool--false-promisecbortransaction_witness_set) endpoint, the [CIP-103? signTxs](https://github.com/cardano-foundation/CIPs/blob/508ea0557bcd17d793da90312789165dcef8a4db/CIP-0103/README.md#apicip103signtxstxs-transactionsignaturerequest-promisecbortransaction_witness_set) endpoint, or any other CIP-30 extension that allows signing transactions, the wallet MUST cache the signed transaction with a TTL (time-to-live) greater than or equal to the [TTL of the transaction](https://github.com/IntersectMBO/cardano-ledger/blob/master/eras/babbage/impl/cddl-files/babbage.cddl#L57C8-L57C9). The wallet MUST NOT remove the transaction from the cache until the TTL has expired, the transaction has been seen on chain, or the cache is full and the wallet needs to remove transactions to make room for newer transactions.
```
klntsky, line 49: I don't think cache size is a concern here. The cache will only contain utxos owned by the wallet, so even if we consider ttl, it is safe to assume that the size will not grow beyond manageable values, as each utxo creation will incur costs. We already implicitly allow the wallet size to grow indefinitely (because the number of utxos in a multi-address wallet is unlimited).

If we allow valid implementations to drop cache data due to size limits, then a dApp would not be able to rely on presence of any data in the cache. And an implementation with SIZE=0 will be a valid one.

yHSJ:
> @yHSJ
> 
> > I felt that it made more sense to have transactions cached at the wallet level.
> 
> They will still be cached on the wallet side. dApps will be able to write to the cache by explicitly providing a set of utxos that should be considered spent after the user signs a given transaction. This is not, in fact, an approach different from caching, just a path to make caching work nicely with transaction chaining.
> 
> > The caching of transactions at the wallet level also allows wallets to 1) remove UTxOs that are consumed in cached transactions but not yet on chain from the getUtxos response and 2) not let a user sign a transaction if a UTxO in a proposed transaction has been consumed in a cached transaction, preventing some of the "cardano-y" errors that users run into and become frustrated by.
> 
> Both cases could also be implemented with the extra argument approach.
> 
> `1)` only concerns wallet's address utxos, and the wallet can infer which of the utxos of the transactions provided via the extra argument are owned by it, and remove them from the set of available utxos

I agree in general it makes sense to also support providing "context" at the sign endpoint level. I think the two ideas work well together, but it is important that a wallet does the caching as well. Otherwise, the attack vector would always exist (just don't pass in the optional additional transactions)

cip-smart-tx-signing/README.md
yHSJ: Sorry I missed this somehow. I think that it's really important to define this behavior in a standard. Lots of the current CIPs have undefined behaviors in edge cases and it can cause real issues across the ecosystem.

I want this standard to be comprehensive, regardless of device or wallet software. In the (rare, as you correctly point out) case that a cache has to be purged, it is important the wallet purges transactions in an appropriate manner (ascending chronologically). 

It's an excellent point that a wallet could correctly implement this CIP with a cache size of zero bytes. I think it makes sense to add a "minimum cache size" that is `n * maxTxSize`.  I would love to hear others' thoughts on what the value of `n` should be, since it will necessarily add (small) overhead to the wallet.

Ryun1:
Should this be a CPS rather than a CIP?

yHSJ:
> Should this be a CPS rather than a CIP?

I think this particular proposal provides one solution to an (currently) unwritten CPS. I thought about writing a CPS first, which @Crypto2099 also suggested, but I felt that I'd rather not release the blueprint for a scammer without also providing a potential solution for wallets.

cip-smart-tx-signing/README.md
klntsky: > n * maxTxSize

Why do we need to cache whole transactions? I think it could work with caching just UTxOs, and only those that belong to one of the user's addresses.

cip-smart-tx-signing/README.md
```diff
@@ -0,0 +1,80 @@
...
+When displaying a transaction summary to the user, the wallet MUST search for the outputs referenced by inputs in the following order:
+
+1. The latest known on-chain UTxO set
+2. The cache of signed transactions
+3. The set of transactions being signed
+
+If a transaction contains an input that references an output that cannot be found, the wallet MUST display a warning to the user that the transaction includes unidentifiable inputs. The style and wording of this warning is left up to the wallet. The wallet MAY prevent the user from signing a transaction with unidentified inputs, but this may cause issues when interacting with some dApps.
+
+## Rationale: how does this CIP achieve its goals?
+
```
Ryun1, line 60: I think it is worth discussing here the alternatives to this approach, such as stronger warnings within wallets OR not letting wallets sign unknown transactions.

cip-smart-tx-signing/README.md
```diff
@@ -0,0 +1,80 @@
...
+License: CC-BY-4.0
+---
+
+## Abstract
+
+This CIP extends [CIP-0030 Cardano dApp-Wallet Web Bridge](https://github.com/cardano-foundation/CIPs/tree/master/CIP-0030/) and [CIP-0103? | Web-Wallet Bridge - Bulk Transaction Signing](https://github.com/cardano-foundation/CIPs/pull/587) to allow wallets to provide a complete and accurate summary of transactions that spend outputs that don't exist on chain yet when possible, and clearly communicate to the user risks when they cannot.
+
+## Motivation: why is this CIP necessary?
+
+Transaction chaining, building transactions that spend outputs of transactions that aren't settled on chain, allows dApps to quickly do several on-chain transactions instead of waiting for each indvidual transaction to settle before building and submitting the next one. This is a powerful feature that has allowed dApps to provide seamless experiences to users. However, due to the tranasaction data structure, wallets are unable to provide a complete and accurate summary of these chained transactions when prompting users for a signature.
```
Ryun1, line 21:
```suggestion
Transaction chaining, building transactions that spend outputs of transactions that aren't settled on chain, allows dApps to quickly do several on-chain transactions instead of waiting for each individual transaction to settle before building and submitting the next one. This is a powerful feature that has allowed dApps to provide seamless experiences to users. However, due to the transactions data structure, wallets are unable to provide a complete and accurate summary of these chained transactions when prompting users for a signature.
```

spelling

cip-smart-tx-signing/README.md
```diff
@@ -0,0 +1,80 @@
+---
+CIP: ?
```
rphair, line 2:
```suggestion
CIP: 111
```

rphair (commented):
Agreed to promote at today's CIP meeting 🚀

NicholasMaselli:
+1 from me. Caching is a good solution and this is definitely critical to Cardano UX.

Also it might be wise to add some method of adding transaction to the wallet cache. This can help sites that submit transactions through the their own backend

cip-smart-tx-signing/README.md
yHSJ: While for the specific problem in the CIP is solved with just the UTxOs, I felt it may be useful to have the complete transaction. For example, caching the inputs as well would allow a wallet to refuse a signature on a transaction which consumes inputs which are already consumed in a cached transaction. Perhaps that is out of the scope of this CIP in particular though.

cip-smart-tx-signing/README.md
```diff
@@ -0,0 +1,80 @@
...
+Implementors: []
+Discussions:
+ - https://github.com/cardano-foundation/CIPs/pull/733
+Created: 2023-12-22
+License: CC-BY-4.0
+---
+
+## Abstract
+
+This CIP extends [CIP-0030 Cardano dApp-Wallet Web Bridge](https://github.com/cardano-foundation/CIPs/tree/master/CIP-0030/) and [CIP-0103? | Web-Wallet Bridge - Bulk Transaction Signing](https://github.com/cardano-foundation/CIPs/pull/587) to allow wallets to provide a complete and accurate summary of transactions that spend outputs that don't exist on chain yet when possible, and clearly communicate to the user risks when they cannot.
```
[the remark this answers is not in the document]
yHSJ: I think I agree in general, I was just trying to follow the standards for "adding on" to the existing CIPs. That said, I think that it makes more sense to not explicitly extend any one CIP, especially as this CIP should be used along side ANY CIP that involves a wallet prompting users for a signature.

cip-smart-tx-signing/README.md
```diff
@@ -0,0 +1,80 @@
...
+
+When displaying a transaction summary to the user, the wallet MUST search for the outputs referenced by inputs in the following order:
+
+1. The latest known on-chain UTxO set
+2. The cache of signed transactions
+3. The set of transactions being signed
+
+If a transaction contains an input that references an output that cannot be found, the wallet MUST display a warning to the user that the transaction includes unidentifiable inputs. The style and wording of this warning is left up to the wallet. The wallet MAY prevent the user from signing a transaction with unidentified inputs, but this may cause issues when interacting with some dApps.
```
rhyslbw, line 57: > The wallet MAY prevent the user from signing a transaction with unidentified inputs, but this may cause issues when interacting with some dApps

What potential issues are there?

yHSJ: Many dApps use chaining with UTxOs that are owned by a smart contract or another address that the user's wallet is not aware of. If a wallet were to prevent signing anytime there was a UTxO it wasn't aware of, it would prevent interactions with many dApps.

rhyslbw: Understood. How about being more assertive, given you're already saying "some dApps":
```suggestion
If a transaction contains an input that references an output that cannot be found, the wallet MUST display a warning to the user that the transaction includes unidentifiable inputs. The style and wording of this warning is left up to the wallet. The wallet MAY prevent the user from signing a transaction with unidentified inputs, but this will cause issues when interacting with some dApps.
```

rphair:
@yHSJ I've been assuming that this will be moving ahead with some concurrent developments from the Wallets Working Group... but we do need to update the status to `Waiting for Author` because of the indefinite delay: which means you should please also update this PR to detail whatever we have been waiting for these past several months.

If you can please keep us updated, the editors would want to ensure that all these `Web-Wallet Bridge` proposals can move together as a pack & get merged when the dependencies are met.  If you have anything that can help that happen, please let's get it on record here.  cc @Ryun1 @Crypto2099

yHSJ:
@rphair Thanks for the update! I apologize, I've been very distracted with lots of other work.

We have been waiting on some implementations of this CIP to go live along with feedback from the wallets. Other than that, I think we're good to go...?

 I will follow up with some wallets again to see if there is a status update on these things. Thanks for the reminder!

yHSJ:
Very belated update, but it seems like this is not something the wallets are planning on implementing directly. I guess we can go ahead and close this for now?

rphair:
thanks for attending to this @yHSJ & yes it does seem that way.  Any methods or proposals that appear to relate to those discussed above, editors & other interest parties can link them back here: so you can close this for now.

PR #736, opened 2024-01-08
CIP-0100 | Small clean up on example files

Ryun1:
### Motivation
Whilst making some[ more example files](https://github.com/Ryun1/metadata), I have found a couple inconsistencies with the provided examples.

### Changes
- Removed proposal property which does not exist in the standard
- Aligned capitalisation
- Changed hashing algorithm from blake2b-224 to blake2b-256 as only 256 [is the only supported](https://github.com/cardano-foundation/CIPs/blob/master/CIP-0100/README.md?plain=1#L108)
- Changed links to reference github rather than CIPs website

cc @Quantumplation

### DIFF
--- a/CIP-0100/README.md
+++ b/CIP-0100/README.md
@@ -273,7 +273,7 @@ An example document with an inline context is provided in [this](./example.json)
 
 Canonicalization should produce the following blak2b-256 hash:
 
-34cbb91e044cbd1dbda9e5aa542a74e712d23d0bf151c436c52920dcac03629c
+`6ea05c0e97b1b7316d48b5587a9e60d47c0c44993c8609c9544d88c111814819`
 
 ## Path to Active
 
--- a/CIP-0100/cip-0100.common.jsonld
+++ b/CIP-0100/cip-0100.common.jsonld
@@ -1,6 +1,5 @@
 {
     "@context": {
-        "Proposal": "https://cips.cardano.org/cip/CIP-0100#Proposal",
         "hashAlgorithm": "https://cips.cardano.org/cip/CIP-0100#hashAlgorithm",
         "body": {
             "@id": "https://cips.cardano.org/cip/CIP-0100#body",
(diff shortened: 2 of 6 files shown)


Ryun1:
One question;

The [cips.cardano.org](https://cips.cardano.org/) site seems to be broken for CIP-100 see [here](https://cips.cardano.org/cip/CIP-0100), so I am tempted to change all references to [cips.cardano.org](https://cips.cardano.org/) to direct to Github instead.

rphair:
@Ryun1 for these links I think consistency is vital, and GitHub has always been a constant (since converting from `CIP-XXXX.md` to `README.md` files) and the source of truth with respect to `cips.cardano.org` and the CIPs on the Developer Portal.  @KtorZ 's https://github.com/cardano-foundation/CIPs/issues/109#issuecomment-1193075214 suggests there may be another overhaul on `cips.cardano.org` still in the works, which by https://github.com/cardano-foundation/CIPs/issues/436#issuecomment-1385476991 might happen now that #389 is finally done.

Whatever cleanup is done there might not invalidate old pathnames but it would likely change canonical pathnames which are the only ones which make sense to use in a schema.  I believe this confirms the importance of using GitHub for canonical links & relying upon user interpretation and parsing for human readers to visit the derived web sites instead.

rphair (approved):
So assuming https://github.com/cardano-foundation/CIPs/pull/736#issuecomment-1881436503 is satisfactory, everything looks good to me.

Ryun1:
> GitHub has always been a constant

@rphair 
Agreed, lets just switch to github links

Quantumplation:
It's worth noting that the urls in the context don't actually need to resolve to anything; the computer doesn't recursively fetch those documents. They are effectively just constants that everyone uses for disambiguation.

So, I'm fine with updating them now, but even if GitHub changes, it's important to not update them once there is data in the wild.

KtorZ (approved):

KtorZ:
(side-note, I am checking why that CIP -- and possibly a few others --  is missing from the new website).

Ryun1:
Also; in the examples I just changed hashing algorithm from blake2b-224 to blake2b-256 as only 256 [is the only supported](https://github.com/cardano-foundation/CIPs/blob/master/CIP-0100/README.md?plain=1#L108).

Quantumplation:
@Ryun1 did you update the resulting hash in the test vector in the readme.md?

Ryun1:
> @Ryun1 did you update the resulting hash in the test vector in the readme.md?

Good spot!
addressed via [41b9ec6](https://github.com/cardano-foundation/CIPs/pull/736/commits/41b9ec6ee7119228a86e2e10bb7b5575a451deaf)

PR #739, opened 2024-01-09
CIP-0088 | Introducing Policy Intent Classification

BrockCruess:
Frontends can use intent classifications to filter which assets they allow on their platform. For example, "NFT" marketplaces can choose to allow only "collectible" classification. DEXs can allow "trading" and "governance" classifications. Bond marketplaces can focus on a "Bond" classification. These are just examples of course, and the classification strings themselves can be community-created (options mainly set by minting platforms who build the initial CIP-88 registration for new minting policies) so that a list of allowed classification strings doesn't have to be maintained by CIP editors. DEXs will require legacy support for the Cardano Token Registry, however for future token listings they can rely on the policy intent classification in the policy's CIP-88 registration.

Proposed for both CIP-25 and CIP-68 as referenced under CIP-88. Can be added to other CIPs that have different genres of assets.

### DIFF
--- a/CIP-0088/CIPs/0025/CIP25_v1.json
+++ b/CIP-0088/CIPs/0025/CIP25_v1.json
@@ -34,7 +34,8 @@
           ]
         ]
       ],
-      "6": "SpaceBudz"
+      "6": "SpaceBudz",
+      "7": "Collectible"
     }
   }
 }
--- a/CIP-0088/CIPs/0025/CIP25_v1.json
+++ b/CIP-0088/CIPs/0025/CIP25_v1.json
@@ -35,7 +35,7 @@
         ]
       ],
       "6": "SpaceBudz",
-      "7": "Collectible"
+      "7": 2
     }
   }
 }
--- a/CIP-0088/CIPs/0025/README.md
+++ b/CIP-0088/CIPs/0025/README.md
@@ -31,7 +31,7 @@ relevant information under index #6, prefixed by the number of the CIP (25 or 68
 | 4     | NSFW Flag           | 0 or 1   | No       |
 | 5     | Social Media        | Array    | No       |
 | 6     | Project/Artist Name | String   | No       |
-| 7     | Policy Intent Class | String   | No       |
+| 7     | Policy Intent Class | Integer  | No       |
 
 For details on what these fields represent and how they should be structured in the metadata, please refer to
 [Token Project Details](../common/Token-Project-Details_v1.md)
--- a/CIP-0088/CIPs/0025/README.md
+++ b/CIP-0088/CIPs/0025/README.md
@@ -31,6 +31,7 @@ relevant information under index #6, prefixed by the number of the CIP (25 or 68
 | 4     | NSFW Flag           | 0 or 1   | No       |
 | 5     | Social Media        | Array    | No       |
 | 6     | Project/Artist Name | String   | No       |
+| 7     | Policy Intent Class | String   | No       |
 
 For details on what these fields represent and how they should be structured in the metadata, please refer to
 [Token Project Details](../common/Token-Project-Details_v1.md)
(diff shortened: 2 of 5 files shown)


Crypto2099:
Sorry for the delay on leaving a comment here @BrockCruess. Is the thought here to have clearly defined classifications?

Generally speaking, in order to reduce ambiguity in CIP-88 I took the approach that "well defined" values should use integer-based indexes to prevent common typos and spelling errors.

Some of the language based on your comment above seems to suggest that this is a more broad "categorization" of the types of token in which case I could see justification that it should instead be an array of values but then this raises the issue of inconsistent "categorization" across marketplaces and how to define and maintain a community-curated list of categories and their associated spellings/indexes.

I fail to see much value-add here if this is just yet another free-form label that may actually introduce more complications and exceptions to rules if people publish data that does not follow a "marketplace-accepted" classification.

BrockCruess:
@Crypto2099 pre-defined and expandable categories with assigned integers was my original intention for this idea, but in our private discussion about it you had mentioned that you were opposed to a "whitelist" of classification options.

if you've changed your mind on that we could work out the best way to maintain a list of valid classifications and if someone comes up with a new category a simple pull request could have it merged and active. we'd just need a way to notify front ends to allow the new option in their registration tools. or find a way to have the registration tools detect new category options in main branch (my personal fav idea).

or we keep it this way and do as you originally recommended and have explorers/front ends build synonym lists to translate synonyms and spelling errors into the valid option. but that seems like a lot of work for front ends.

i'm open to explore any option that makes this idea work, as i really think it'll be valuable for asset filtration in the long run as hundreds of thousands to millions (to billions?) of new assets are minted to chain in the future and front ends either have to see all of them and build/store data for all of them, or filter them manually.

rphair:
@BrockCruess we have to move this on the path toward closure if we don't have a firm justification for making changes to CIP-0088.  It's been 6 months since your stating (https://github.com/cardano-foundation/CIPs/pull/739#issuecomment-1925301746) you are "open to ideas" about the exact implementation & we do need more than that to proceed through the CIP process.

From that point of view this looks (potentially) `Abandoned` so I'm tagging it as such.  We can throttle it back to `Waiting for Author` if @BrockCruess you can identify & post any specific issues you are trying to solve or demonstrate before progressing the CIP.  In any case please try develop your case more if you would still like to keep this PR open.

BrockCruess:
@rphair the issues I'm trying to solve seem clear to me and the commits demonstrate the simple addition to the metadata. I haven't heard back from @Crypto2099 regarding the preference of an integer-based index (a whitelist of acceptable inputs) over a user-defined string that relies on the platforms to determine which strings they accept. Further progression of this PR does rely on that decision, and an editor should probably make that call as you guys would know better than I would. I do personally agree that an integer-based index would work best and that was my initial idea for this addition. If an index format is should be added to this PR please let me know or please recommend a simple format as my first attempt may not be ideal and will require revisions anyways.
I'm happy to edit my existing commits with example integers in the meantime.

rphair:
Thanks for the update @BrockCruess ... changing tag accordingly back to `Waiting for Author`.  Ideally we should never need a `Waiting for Editor` tag but we need something more specific than "Waiting for Someone other than Author".  It's one of the process issues that we're trying to settle in tagging these proposals & thanks for your status update in any case.

@Crypto2099 I trust that you will respond as the author has requested & keep interacting until that point is settled? 🙏

rphair:
@Crypto2099 it seems we are waiting for you to respond here.  Since we have no "Waiting for Editor" tag we may have to leave it here in half-life unless @BrockCruess can make a case for these changes that everyone can understand.  In the meantime we don't have enough of a consensus to modify CIP-0088 as suggested here.

I hope the 2 of you can work it out & if that isn't done in a reasonable amount of time then we should probably close this.  (If any third parties are following this proposed change & think it would be beneficial, please post here in the meantime.)

PR #745, opened 2024-01-19
CIP-1694 | Correctly characterise delegated active stake

CarlosLopezDeLara:
The stake distribution used for talllying spo votes is not the _registered voting stake_ (delegated to DReps), but instead the _active stake_: the stake delegated and active on the epoch for block production.


rphair (approved):
Looks both more accurate and more precise.

Ryun1 (approved):
LGTM

PR #746, opened 2024-01-19
 CIP-1694 | French - Correctly characterise delegated active stake

Hornan7:
- Correction to the french version related to PR #745 
- From commit "fix errata about stake distribution."

### DIFF
--- a/CIP-1694/README.fr.md
+++ b/CIP-1694/README.fr.md
@@ -2,7 +2,7 @@
 CIP: 1694
 Source: https://github.com/cardano-foundation/CIPs/blob/master/CIP-1694/README.md
 Title: Un premier pas vers une gouvernance décentralisée on-chain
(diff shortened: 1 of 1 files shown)


rphair (approved):
_sans reproche_ as usual 😎

KtorZ (approved):
Pas convaincu par l'emploi de "mise" pour décrire la stake; _sounds a bit like gambling_. Mais bon, c'est pas le sujet ici. 

LGTM.

rphair:
@Hornan7 I see the last point & maybe _[le principal](https://www.collinsdictionary.com/dictionary/english-french/principal)_ would be better, which has the advantage of meaning & spelling the same in English.  Yet that word is used all through this document and it would be a substantial change for another PR & also perhaps should be coordinated with other French translated Cardano docs.  cc @KtorZ

PR #747, opened 2024-01-22
CIP-0117? | Explicit script return values

michaelpj:
There is a papercut in how we check whether scripts accept transactions. There are a few ways that people can make mistakes, not notice, and have their scripts fail open. This proposal suggests that we instead be prescriptive about what scripts must return in order to succeed, which would make such mistakes quickly obvious.

This is not a particularly important change, but it would be relatively cheap to do and would fix a minor problem.

---

[Rendered Proposal on Fork](https://github.com/michaelpj/CIPs/blob/mpj/explicit-fail/CIP-0117/README.md) < edit: changed to latest version in branch

### DIFF
--- a/CIP-0117/README.md
+++ b/CIP-0117/README.md
@@ -1,5 +1,5 @@
 ---
-CIP: 0117
+CIP: 117
 Title: Explicit script return values
 Category: Plutus
 Status: Proposed
--- a/CIP-0117/README.md
+++ b/CIP-0117/README.md
@@ -76,7 +76,7 @@ This is very unlikely to happen by accident: mistakes in the number of arguments
 
 ### Implementation Plan
 
-- [ ] The Plutus team will implement the changes to the Ledger.
+- [x] The Plutus team will implement the changes to the Ledger.
 
 ## Copyright
 
--- a/CIP-0117/README.md
+++ b/CIP-0117/README.md
@@ -71,7 +71,7 @@ This is very unlikely to happen by accident: mistakes in the number of arguments
 
 ### Acceptance Criteria
 
-- [ ] The proposal is implemented in the Ledger.
+- [x] The proposal is implemented in the Ledger.
 - [ ] The ledger changes are released on Mainnnet.
 
 ### Implementation Plan
--- a/CIP-0117/README.md
+++ b/CIP-0117/README.md
@@ -1,5 +1,5 @@
 ---
-CIP: ?
+CIP: 0117
 Title: Explicit script return values
 Category: Plutus
 Status: Proposed
--- a/CIP-????/README.md
+++ b/CIP-????/README.md
@@ -1,4 +1,4 @@
---
+---
 CIP: ?
 Title: Explicit script return values
 Category: Plutus
--- a/CIP-????/README.md
+++ b/CIP-????/README.md
@@ -8,7 +8,7 @@ Authors:
 Implementors: 
     - Michael Peyton Jones <michael.peyton-jones@iohk.io>
 Discussions:
-    - https://github.com/cardano-foundation/cips/pulls/?
+    - https://github.com/cardano-foundation/CIPs/pull/747
 Created: 2024-01-22
 License: CC-BY-4.0
 ---
--- a/CIP-????/README.md
+++ b/CIP-????/README.md
@@ -1,4 +1,4 @@
----
+--
 CIP: ?
 Title: Explicit script return values
 Category: Plutus
@@ -50,18 +50,20 @@ While these mistakes are relatively easily avoidable (any good smart contract to
 
 The specification for checking whether a Plutus Core script accepts a transaction changes as follows (the new part is in brackets):
 
-> A Plutus Core script S with arguments A1...An accepts a transaction if 'eval(S A1 ... An)' succeeds [and evaluates to the builtin boolean constant 'true'].
+> A Plutus Core script S with arguments A1...An accepts a transaction if 'eval(S A1 ... An)' succeeds [and evaluates to the builtin constant 'unit'].
 
 This change is not backwards-compatible and will need to go into a new Plutus ledger language.
 
 ## Rationale: how does this CIP achieve its goals?
 
-Since the return value of a script will now be significant, a script will only succeed if the whole thing evaluates to 'true'.
+Since the return value of a script will now be significant, a script will only succeed if the whole thing evaluates to 'unit'.
 This is very unlikely to happen by accident: mistakes in the number of arguments or in what to return will result in failure.
 
 ### Alternatives 
 
 - The status quo is not terrible, and we could simply accept it.
+- The return value could be a boolean constant, with 'true' indicating success and 'false' indicating failure.
+    - This is slightly more complicated, and technically we only need a designated success value, since "anything else" indicates failure. We don't need to distinguish between "normal exit indicating rejection of the transaction" and "abnormal exit".
 - We could specifically detect when a script returns a lambda, and say that that is a failure.
     - This is patching up one particular hole, whereas the proposal here has much more coverage by failing everything that doesn't quite specifically return 'true'.
 
--- a/CIP-????/README.md
+++ b/CIP-????/README.md
@@ -71,12 +71,12 @@ This is very unlikely to happen by accident: mistakes in the number of arguments
 
 ### Acceptance Criteria
 
-- The proposal is implemented in the ledger
-- The ledger changes are released on mainnnet
+- [ ] The proposal is implemented in the Ledger.
+- [ ] The ledger changes are released on Mainnnet.
 
 ### Implementation Plan
 
-- The Plutus team will implement the changes to the ledger
+- [ ] The Plutus team will implement the changes to the Ledger.
 
 ## Copyright
 
--- a/CIP-????/README.md
+++ b/CIP-????/README.md
@@ -1,0 +1,83 @@
+---
+CIP: ?
+Title: Explicit script return values
+Category: Plutus
+Status: Proposed
+Authors:
+    - Michael Peyton Jones <michael.peyton-jones@iohk.io>
+Implementors: 
+    - Michael Peyton Jones <michael.peyton-jones@iohk.io>
+Discussions:
+    - https://github.com/cardano-foundation/cips/pulls/?
+Created: 2024-01-22
+License: CC-BY-4.0
+---
+
+## Abstract
+
+Today, Plutus Core scripts signal success or failure exclusively by whether the script terminates normally or abnormally (with an error).
+This leads to some false positives, where a script terminates normally, but this was not intended by the author.
+We propose to additionally look at what the script evaluates to when checking for success or failure.
+
+## Motivation: why is this CIP necessary?
+
+Consider the following Plutus scripts, intended to be used as minting policy scripts:
+
+```
+\redeemer -> \datum -> \script-context -> (builtin error)
+```
+
+```
+\redeemer -> \script-context -> (con false)
+```
+
+Today, both of these will unconditionally succeed! 
+
+1. Minting policies only receive two arguments, but this script expects three before it does any work. It therefore evaluates successfully to a lambda.
+2. This script evaluates _successfully_ to `(con false)`, but the return value is irrelevant since it terminates successfully.
+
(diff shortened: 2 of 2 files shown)


CIP-????/README.md
```diff
@@ -0,0 +1,83 @@
...
+> A Plutus Core script S with arguments A1...An accepts a transaction if 'eval(S A1 ... An)' succeeds [and evaluates to the builtin boolean constant 'true'].
+
+This change is not backwards-compatible and will need to go into a new Plutus ledger language.
+
+## Rationale: how does this CIP achieve its goals?
+
+Since the return value of a script will now be significant, a script will only succeed if the whole thing evaluates to 'true'.
+This is very unlikely to happen by accident: mistakes in the number of arguments or in what to return will result in failure.
+
+### Alternatives
```
zliu41, line 62: How about unit (or either unit or true)? Most scripts written in Plutus Tx return unit, so they won't need to change anything.

effectfully: Yeah, breaking backwards compatibility in such a horrible implicit way sounds like an even worse footgun than the status quo. I do understand that people will need to explicitly pick a new language version, but we can't be sure that they will adapt their code properly.

effectfully (commented):
I don't like it. I don't like the status quo either, though. But less so.

colll78:
Please add `unit` as a supported acceptance return value.

CIP-????/README.md
michaelpj: Most scripts written in Plutus Tx return the _Plutus Tx_ datatype unit, i.e. some scott-encoded or SOP type. What that looks like is up to the compiler, so I don't think we can easily state a good recognition procedure for it here. We could recognize builtin unit, though.

CIP-????/README.md
```diff
@@ -0,0 +1,83 @@
...
+
+## Rationale: how does this CIP achieve its goals?
+
+Since the return value of a script will now be significant, a script will only succeed if the whole thing evaluates to 'true'.
+This is very unlikely to happen by accident: mistakes in the number of arguments or in what to return will result in failure.
+
+### Alternatives 
+
+- The status quo is not terrible, and we could simply accept it.
+- We could specifically detect when a script returns a lambda, and say that that is a failure.
```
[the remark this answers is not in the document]
michaelpj: Yeah, I just put it in here for completeness.

CIP-????/README.md
```diff
@@ -0,0 +1,83 @@
...
+Historically, Plutus Core was going to be a typed language, and at least the first kind of error would have been caught by the typechecker. 
+However, today there is little stopping people from making such mistakes.
+
+While these mistakes are relatively easily avoidable (any good smart contract toolkit should prevent them), it is nonetheless a potential landmine for users.
+
+## Specification
+
+The specification for checking whether a Plutus Core script accepts a transaction changes as follows (the new part is in brackets):
+
+> A Plutus Core script S with arguments A1...An accepts a transaction if 'eval(S A1 ... An)' succeeds [and evaluates to the builtin boolean constant 'true'].
```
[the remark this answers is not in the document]
michaelpj: No, it won't. But you can return (and construct) a `BuiltinBool` easily enough, or we can provide an adapter.

michaelpj:
I don't think we should have multiple different kinds of accepted return value. `unit` would be okay, so long as then it _has_ to be `unit`. The whole point of this CIP is to narrow down what scripts return as much as possible, so its very clear if you're doing it correctly or not. I don't really see the benefit to having a larger space of possibilities.

I will add an alternative for the return type being `unit`, although I do kind of like the fact that returning `bool` lets us actually interpret the result of the script as "should this transaction be accepted, y/n?".

colll78:
> I don't think we should have multiple different kinds of accepted return value. `unit` would be okay, so long as then it _has_ to be `unit`. The whole point of this CIP is to narrow down what scripts return as much as possible, so its very clear if you're doing it correctly or not. I don't really see the benefit to having a larger space of possibilities.
> 
> I will add an alternative for the return type being `unit`, although I do kind of like the fact that returning `bool` lets us actually interpret the result of the script as "should this transaction be accepted, y/n?".

 That makes sense, I see the value in having a single accepted return value. 
 
 I see a case for both `unit` and `bool`. Like you suggest, `bool` makes interpreting the result of the script more direct; however, I would slightly prefer `unit` be the accepted return value because `bool` might confuse people into believing that the only possible return values are `true / false`. `unit` is more transparent in signaling that anything other than `()` should be a script failure. 
 
Either is fine though. I have seen a lot of contract developers especially early on struggle with debugging because they didn't apply enough arguments to their validator, and they couldn't figure out why it was succeeding no matter what they changed.

michaelpj:
I updated the PR to suggest using unit. That does seem more parsimonious since we only need a single distinguished value for success.

michaelpj:
@rphair I think this is ready for review

CIP-????/README.md
```diff
@@ -0,0 +1,85 @@
+--
+CIP: ?
+Title: Explicit script return values
```
rphair, line 3: PR retitled same as CIP (and on meeting agenda # 81) in keeping with most of them being noun phrases rather than verbs made to look like nouns.  If you think `Checking` is more clear please reset both titles that way or let me know & I'll do it.

CIP-????/README.md
```diff
@@ -0,0 +1,85 @@
...
+Category: Plutus
+Status: Proposed
+Authors:
+    - Michael Peyton Jones <michael.peyton-jones@iohk.io>
+Implementors: 
+    - Michael Peyton Jones <michael.peyton-jones@iohk.io>
+Discussions:
+    - https://github.com/cardano-foundation/cips/pulls/?
```
rphair, line 11:
```suggestion
    - https://github.com/cardano-foundation/CIPs/pull/747
```

CIP-????/README.md
```diff
@@ -0,0 +1,85 @@
...
+
+Consider the following Plutus scripts, intended to be used as minting policy scripts:
+
+```
+\redeemer -> \datum -> \script-context -> (builtin error)
+```
+
+```
+\redeemer -> \script-context -> (con false)
+```
```
rphair, line 32: Is there a GitHub language type that causes this Plutus syntax not to be marked as errors?

CIP-????/README.md
```diff
@@ -0,0 +1,85 @@
...
+## Path to Active
+
+### Acceptance Criteria
+
+- The proposal is implemented in the ledger
+- The ledger changes are released on mainnnet
+
+### Implementation Plan
+
+- The Plutus team will implement the changes to the ledger
```
rphair, line 79:
```suggestion
- [ ] The proposal is implemented in the Ledger.
- [ ] The ledger changes are released on Mainnnet.

### Implementation Plan

- [ ] The Plutus team will implement the changes to the Ledger.
```
These will likely be done at different times so let's use checkboxes. 🤓  Also capitalising some proper nouns (there's only one Ledger and one Mainnet) and adding full stops at the end of these complete sentences. 😇

rphair (commented):
@michaelpj it's already on Review for meeting # 81 in 1 week's time as agreed in last meeting.  I'll plan to give this a number after that meeting (you are invited of course) .  If no dissent we might keep pushing it forward to `Last Check`. https://hackmd.io/@cip-editors/81

It seems like the `unit` question is settled as of https://github.com/cardano-foundation/CIPs/pull/747#issuecomment-1908256914 but of course that can also be triple checked with the other devs at the meeting.

Document itself looks good to me & I would approve it after we assign a number: still keeping an eye on the reservations with @effectfully https://github.com/cardano-foundation/CIPs/pull/747#pullrequestreview-1837029375.  I would rather see harmony about this before assigning `Last Check` if possible: it would be nice if both if you could be at the meeting & we could hear some debate about it.

michaelpj:
To be clear, we might not do this, but I think it's helpful to have it merged as Proposed.

CIP-????/README.md
rphair: Seems to be airing well with new name, so resolving

rvcas:
Just seeing this, I think it's a good idea. The `unit` stuff makes sense, anything else fails including (obviously) explicit `error`. I think many of the compile-to-uplc languages already have a wrapper that gets injected which uses an `ifThenElse` to check the Boolean value returned from the inner function. The wrapper redirects `true` to `()` and `false` to `error`.

effectfully:
Somehow `unit` doesn't feel bad to me either, unlike `true`. I think it's good design, but I believe we should assess the impact before rolling out such a breakage change by studying what kind of validators we already have on the chain. If those routinely return non-`unit` values, then we still might consider making `unit` mandatory, but we'd need to go on a campaign explaining to everybody how the new version of Plutus expects things to be done very differently.

CIP-????/README.md
```diff
@@ -0,0 +1,85 @@
+---
+CIP: ?
```
```suggestion
CIP: 0117
```

rphair (commented):
thanks @michaelpj - confirmed as candidate at CIP editors' meeting today, though still not certain to be implemented: clearly beneficial but uncertain how much.  In any case we agree it should stand as a reference among other valid ideas.

Please update directory name to `CIP-0117` and change the link to the rendered proposal in the OP accordingly 🎉

zliu41:
CIP-69, which makes all Plutus script take a single argument, makes the problem described in this CIP more likely to happen, and thus more desirable to requiring scripts to return builtin unit to succeed.

I don't know how this impacts other languages and compilers. @rvcas @MicroProofs @KtorZ @nielstron @michele-nuzzi @nau any opinions? Do you think this is worth doing? Any concerns?

nielstron:
OpShin already only allows smart contracts to return builtin unit.

MicroProofs:
Aiken also enforces unit is returned on a successful execution.

zliu41:
Ok, this has been implemented in plutus-ledger-api: `evaluateScriptCounting` and `evaluateScriptRestricting` now require returning `BuiltinUnit` for the evaluation to succeed, except if the script is PlutusV1 or PlutusV2.

@rphair I think we can merge this CIP.

MicroProofs:
Yo that's pretty awesome didn't expect this to make into PlutusV3

CIP-????/README.md
rphair: Somehow GitHub changed its parsing in the last 4 or 5 months; not marked as errors today. 🤔 (resolving)

rphair (approved):
thanks @zliu41 - marked `Last Check` so we should merge this no later than the coming Tuesday's meeting (https://hackmd.io/@cip-editors/90) cc @Ryun1 @Crypto2099

CIP-0117/README.md
```diff
@@ -0,0 +1,85 @@
+---
+CIP: 0117
```
Ryun1, line 2:
```suggestion
CIP: 117
```

CIP-0117/README.md
```diff
@@ -0,0 +1,85 @@
...
+## Path to Active
+
+### Acceptance Criteria
+
+- [ ] The proposal is implemented in the Ledger.
+- [ ] The ledger changes are released on Mainnnet.
+
+### Implementation Plan
+
+- [ ] The Plutus team will implement the changes to the Ledger.
```
Ryun1, line 79:
```suggestion
- [x] The Plutus team will implement the changes to the Ledger.
```
i believe this has been implemented; https://github.com/cardano-foundation/CIPs/pull/747#issuecomment-2150723408

CIP-0117/README.md
```diff
@@ -0,0 +1,85 @@
...
+- The return value could be a boolean constant, with 'true' indicating success and 'false' indicating failure.
+    - This is slightly more complicated, and technically we only need a designated success value, since "anything else" indicates failure. We don't need to distinguish between "normal exit indicating rejection of the transaction" and "abnormal exit".
+- We could specifically detect when a script returns a lambda, and say that that is a failure.
+    - This is patching up one particular hole, whereas the proposal here has much more coverage by failing everything that doesn't quite specifically return 'true'.
+
+## Path to Active
+
+### Acceptance Criteria
+
+- [ ] The proposal is implemented in the Ledger.
```
Ryun1, line 74:
```suggestion
- [x] The proposal is implemented in the Ledger.
```
i believe this has been implemented; https://github.com/cardano-foundation/CIPs/pull/747#issuecomment-2150723408

Ryun1 (approved):
Good to go ✅

PR #748, opened 2024-01-22
CIP-0100 | Add index and hash to `references` and allow for basic HTML tags

Ryun1:
### Changes
- Added an index field to the `references` field.
  - Allowing each reference to be easily be identified.
- Added optional hash to `references` field.
  - So that references can be optionally cryptographically verified.
- Added the ability for authors to include a HTML tags in their freeform text fields.
  - To allow for authors to highlight bold text, italicised text or code.

cc: @Quantumplation

### DIFF
--- a/CIP-0100/README.md
+++ b/CIP-0100/README.md
@@ -162,6 +162,16 @@ We expect a rich ecosystem of CIPs to emerge defining different extensions, so t
 
 The rest of this document will provide a high level description of how these fields should be interpreted
 
+### HTML Tags
+
+For downstream tooling we define a reduced supported set of HTML tags which can be included in freefrom text fields.
+
+| Tag      | Description |
+| ---      | ----------- |
+| `<b>`    | bold        |
+| `<i>`    | italicized  |
+| `<code>` | code        |
+
 ### High level description
 
 The following properties are considered common to all types of transactions, and the minimal set needed for "minimum viable governance":
@@ -180,8 +190,12 @@ The following properties are considered common to all types of transactions, and
   - `references`
     - An array of objects
     - Each object specifies a `@type`, which is either "GovernanceMetadata" or "Other"
+    - Each object specifies a unique integer `index` field value which increases without gaps from `0`. This is to act as a unique referable ID which can be mentioned in via other fields
     - Each object specifies a `label`, which serves as a human readable display name
     - Each object specifies a `uri`; when the type is set to "GovernanceMetadata", the URI should point to another CIP-0100 compliant document
+    - Each object MAY have a OPTIONAL `uriHash` object.
+      - Each object MUST have a `hashAlgo` field. To indicate what hashing algorithm is being used.
+      - Each object MUST have a `hash` field. To show the hash digest of the content at the URI.
   - `comment`
     - Freeform text attached to the metadata; richer structures for justifying the transaction this is attached to is left to future CIPs
     - Tooling authors SHOULD emphasize that these comments represent the *authors* views, and may contain bias.


CIP-0100/README.md
```diff
@@ -162,6 +162,16 @@ We expect a rich ecosystem of CIPs to emerge defining different extensions, so t
 
 The rest of this document will provide a high level description of how these fields should be interpreted
 
+### HTML Tags
+
+For downstream tooling we define a reduced supported set of HTML tags which can be included in freefrom text fields.
+
+| Tag      | Description |
```
Ryun1, line 169: Please people feel free to suggest more :)

Also is HTML the method we want to use, could use the markdown ones like `**bold text**`.

CIP-0100/README.md
```diff
@@ -180,8 +190,12 @@ The following properties are considered common to all types of transactions, and
   - `references`
     - An array of objects
     - Each object specifies a `@type`, which is either "GovernanceMetadata" or "Other"
+    - Each object specifies a unique integer `index` field value which increases without gaps from `0`. This is to act as a unique referable ID which can be mentioned in via other fields
     - Each object specifies a `label`, which serves as a human readable display name
     - Each object specifies a `uri`; when the type is set to "GovernanceMetadata", the URI should point to another CIP-0100 compliant document
+    - Each object MAY have a OPTIONAL `uriHash` object.
```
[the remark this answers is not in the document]
Ryun1: > Make sure to add these to the context and schema.
Will do, but wanted to get confirmation on people's opinion on these changes first.

> We should also define a time where we're cutting off all further changes (barring mistakes / errors) to CIP-100; This could be a separate CIP which defines the context for this addition, for example, that was the whole point of using JSON-LD.
Yes, I did only make a PR because I don't believe there are any implementations of CIP-100 yet, so thought id try to sneak in these changes.

CIP-0100/README.md
rphair: > Also, why HTML over the much more industry standard markdown?

one justification: because there is one convergent definition of HTML while the "industry standard" of Markdown has produced a bunch of flavours with irreconcilable special features.  Those special features in Markdown come from special symbol behaviour, while in HTML it's delineated with tags which can be specifically enumerated herein.

Generally (and maybe even more relevant for blockchain histories), if I had to produce a richly formatted document to go into a time capsule to be readable and/or parseable in 100 years I would rather it be in HTML than any other format.

Quantumplation: I don't actually feel that that's the case, except in some extreme niche features; for the most part, people have standardized on github flavored markdown.

The reason I'd argue for GHMD over html, is that it's a lot easier for someone (especially non-technical users) to *write* in a textbox, and it's also a lot less confusing to read if you're looking at the raw data without the formatting applied.

Even if you know nothing about markdown, asterisks around a word are pretty intuitively indicative of emphasis, while `<b>Some word</b>` could be confusing to someone very non-technical.

Ryun1: > The reason I'd argue for GHMD over html, is that it's a lot easier for someone (especially non-technical users) to write in a textbox, and it's also a lot less confusing to read if you're looking at the raw data without the formatting applied.

This is a good point, id be in favour of adopting this.

CIP-0100/README.md
```diff
@@ -180,8 +190,12 @@ The following properties are considered common to all types of transactions, and
   - `references`
     - An array of objects
     - Each object specifies a `@type`, which is either "GovernanceMetadata" or "Other"
+    - Each object specifies a unique integer `index` field value which increases without gaps from `0`. This is to act as a unique referable ID which can be mentioned in via other fields
```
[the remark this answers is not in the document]
Ryun1: Great idea!

Ryun1:
Thinking about it more, with @Quantumplation's insights 🙏

I will move the HTML/Markdown out of this CIP, as they really only apply to longer form text, which is what CIP-108 adds.
This maintains CIP-100 as a barebones standard.

I will keep the suggestion of adding the `@list` keyword on the `references`.
But move the additional hashing to CIP-108 and just extend CIP-100 on it.

Quantumplation:
@Ryun1 I'd also suggest making sure to specify html (or markdown) on a field by field basis; otherwise you end up with like, the authors name having html in it, as that's "free form text" heh

Ryun1:
> @Ryun1 I'd also suggest making sure to specify html (or markdown) on a field by field basis; otherwise you end up with like, the authors name having html in it, as that's "free form text" heh

Good call!

CIP-0100/README.md
rphair: @Ryun1 @Quantumplation yes, Markdown in hindsight I do think would be better.  I was thinking last night in terms of generality but today in terms of failure modes... mainly what happens if the markup _isn't rendered_.  As also suggested in https://github.com/cardano-foundation/CIPs/pull/748#discussion_r1462312205 the non-rendered Markdown is more readable than non-rendered HTML.

CIP-0100/README.md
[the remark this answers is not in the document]
Ryun1: might even specify that we use `@set` keyword, then use something else to reference

Ryun1:
For ease, I am just going to add these additions via CIP-108
So I will close this for now

PR #735, opened 2024-01-08
CIP-0148? | Tag / Redeemer field in TxOut

colll78:
We propose to allow the attachment of arbitrary, temporary data to transaction outputs within the script context. This data, akin to redeemers in their operational context, is intended to be used exclusively during the execution of Plutus scripts and thus are not recorded by the ledger. This will facilitate a wide variety of smart contract design patterns, one of which can be used as a general solution for double satisfaction without sacrificing script composability.

---

([proposal rendered from branch](https://github.com/colll78/CIPs/blob/patch-1/CIP-0148/README.md))

PR #749, opened 2024-01-23
CIP-0112? | Observe script type

colll78:
We propose to introduce a new Plutus scripts type Observe in addition to those currently available (spending, certifying, rewarding, minting, drep). The purpose of this script type is to allow arbitrary validation logic to be decoupled from any ledger action. Since observe validators are decoupled from actions, you can run them in a transaction without needing to perform any associated action (ie you don't need to consume a script input, or mint a token, or withdraw from a staking script just to execute this validator).

---

[Rendered Proposal on Fork](https://github.com/colll78/CIPs/blob/Observe-Script-Type/cip-observe-script-type/README.md)

### DIFF
--- a/cip-observe-script-type/README.md
+++ b/cip-observe-script-type/README.md
@@ -1,0 +1,125 @@
+---
+CIP: ?
+Title: Observe Script Type
+Status: Proposed
+Category: Plutus
+Authors:
+    - Philip DiSarro <info@anastasialabs.com>
+Implementors: []
+Discussions: []
+Created: 2024-1-8
+License: CC-BY-4.0
+---
+
+
+<!-- Existing categories:
+
+- Meta     | For meta-CIPs which typically serves another category or group of categories.
+- Wallets  | For standardisation across wallets (hardware, full-node or light).
+- Tokens   | About tokens (fungible or non-fungible) and minting policies in general.
+- Metadata | For proposals around metadata (on-chain or off-chain).
+- Tools    | A broad category for ecosystem tools not falling into any other category.
+- Plutus   | Changes or additions to Plutus
+- Ledger   | For proposals regarding the Cardano ledger (including Reward Sharing Schemes)
+- Catalyst | For proposals affecting Project Catalyst / the Jörmungandr project
+
+-->
+
+## Abstract
+<!-- A short (\~200 word) description of the proposed solution and the technical issue being addressed. -->
+We propose to introduce a new Plutus scripts type `Observe` in addition to those currently available e (spending, certifying, rewarding, minting, drep). The purpose of this script type is to allow arbitrary validation logic to be decoupled from any ledger action. 
+Since observe validators are decoupled from actions, you can run them in a transaction without needing to perform any associated action (ie you don't need to consume a script input, or mint a token, or withdraw from a staking script just to execute this validator). 
+
+## Motivation: why is this CIP necessary?
+<!-- A clear explanation that introduces the reason for a proposal, its use cases and stakeholders. If the CIP changes an established design then it must outline design issues that motivate a rework. For complex proposals, authors must write a Cardano Problem Statement (CPS) as defined in CIP-9999 and link to it as the `Motivation`. -->
+Often in a plutus validator you want to check "a particular Plutus script checked this transaction", but it's annoying (and wasteful) to have to have to lock an output in a script and then check if that output is consumed, or mint a token, or whatever else just to trigger script validation. 
+
+Currently the main design pattern used to achieve this is a very obscure trick involving staking validators and the fact that you can withdraw 0 from a staking validator to trigger the script validation. A summary of the trick is:
+Implement all the intended validation logic in a Plutus staking validator, we will call this validator `s_v`. To check that this validator was executed in the transaction you check if the credential of `s_v` (`StakingCredential`) is present in `txInfoWdrl`, this guarantees that `s_v` was checked in validation. 
+This relies on the fact that unlike in `txInfoMint` the ledger does not filter out 0 amount entries in `txInfoWdrl`. This means that you are allowed to build transactions that withdraw zero from a staking credential which in-turn triggers the staking script associated with that credential to execute in the transaction,
+which makes it available in `txInfoWdrl`. This is a enables a very efficient design pattern for checking logic that is shared across multiple scripts.
+
+For instance, a common design pattern is a token based forwarding validator in which the validator defers its logic to another validator by checking that a state token is present in one of the transaction inputs:
+```haskell
+forwardNFTValidator :: AssetClass -> BuiltinData -> BuiltinData -> ScriptContext -> () 
+forwardNFTValidator stateToken _ _ ctx = assetClassValueOf stateToken (valueSpent (txInfo ctx)) == 1
+```
+This pattern is common in protocols that use the batcher architecture. Some protocols improve on the pattern by including the index of the input with the state token in the redeemer:
+```haskell
+forwardNFTValidator :: AssetClass -> BuiltinData -> Integer -> ScriptContext -> () 
+forwardNFTValidator stateToken _ tkIdx ctx =  assetClassValueOf stateToken (txInInfoResolved (elemAt tkIdx (txInfoInputs (txInfo ctx)))) == 1 
+
+forwardMintPolicy:: AssetClass -> Integer -> ScriptContext -> () 
+forwardMintPolicy stateToken tkIdx ctx =  assetClassValueOf stateToken (txInInfoResolved (elemAt tkIdx (txInfoInputs (txInfo ctx)))) == 1 
+```
+With this pattern DApps are able to process roughly 12-30 forwardNFTValidator UTxO's  per transaction without exceeding script budget limitations.
+The time complexity of this validator is **O(n)** where n is the number of tx inputs. This logic is executed once per input being unlocked  / currency symbol being minted. 
+The redundant execution of searching the inputs for a token is the largest throughput bottleneck for these DApps; it is **O(n*m)** where n is the number of inputs and m is the number of `forwardValidator` inputs + `forwardValidator` minting policies.
+Using the stake validator trick, the time complexity of the forwarding logic is improved to **O(1)**. The forwardValidator logic becomes:
+```haskell
+forwardWithStakeTrick:: StakingCredential -> BuiltinData -> BuiltinData -> ScriptContext -> ()
+forwardWithStakeTrick obsScriptCred tkIdx ctx = fst (head stakeCertPairs) == obsScriptCred 
+  where 
+    info = txInfo ctx 
+    stakeCertPairs = AssocMap.toList (txInfoWdrl info)
+```
+IE check that the StakingCredential is in the first pair in the `txInfoWdrl`.  This script is **O(1)** in the case where you limit it to one Observe script, or if you don't want to break composability with other Observe scripts, then it becomes** O(obs_N)** where `obs_N` is the number of Observe validators that are executed in the transaction.
+
+This proposal just makes this design pattern indepedent from implementation details of stake validators and withdrawals, and improves efficiency and readability for validators that implement it. 
+
+## Specification
+<!-- The technical specification should describe the proposed improvement in sufficient technical detail. In particular, it should provide enough information that an implementation can be performed solely on the basis of the design in the CIP. This is necessary to facilitate multiple, interoperable implementations. This must include how the CIP should be versioned. If a proposal defines structure of on-chain data it must include a CDDL schema in it's specification.-->
+The type signature of this script type will be consistent with the type signature of minting and staking validators, namely:
+```haskell
+Redeemer -> ScriptContext -> () 
+```
+
+The type signature of the newly introduced `Purpose` will be:
+```haskell
+Observe Integer -- ^ where integer is the index into the observations 
+```
+
+A new field will be introduced into the script context:
+
+```haskell
+-- | TxInfo for PlutusV3
+data TxInfo = TxInfo
+  { txInfoInputs                :: [V2.TxInInfo]
+  , txInfoReferenceInputs       :: [V2.TxInInfo]
+  , txInfoOutputs               :: [V2.TxOut]
+  , txInfoFee                   :: V2.Value
+  , txInfoMint                  :: V2.Value
+  , txInfoTxCerts               :: [TxCert]
+  , txInfoWdrl                  :: Map V2.Credential Haskell.Integer
+  , txInfoValidRange            :: V2.POSIXTimeRange
+  , txInfoSignatories           :: [V2.PubKeyHash]
+  , txInfoRedeemers             :: Map ScriptPurpose V2.Redeemer
+  , txInfoData                  :: Map V2.DatumHash V2.Datum
+  , txInfoId                    :: V2.TxId
+  , txInfoVotingProcedures      :: Map Voter (Map GovernanceActionId VotingProcedure)
+  , txInfoProposalProcedures    :: [ProposalProcedure]
+  , txInfoCurrentTreasuryAmount :: Haskell.Maybe V2.Value
+  , txInfoTreasuryDonation      :: Haskell.Maybe V2.Value
+  , txInfoObservations          :: [ScriptHash] -- ^ newly introduced list of observation scripts that executed in this tx. 
+  }
+```
+
+## Rationale: how does this CIP achieve its goals?
+<!-- The rationale fleshes out the specification by describing what motivated the design and what led to particular design decisions. It should describe alternate designs considered and related work. The rationale should provide evidence of consensus within the community and discuss significant objections or concerns raised during the discussion.
+
+It must also explain how the proposal affects the backward compatibility of existing solutions when applicable. If the proposal responds to a CPS, the 'Rationale' section should explain how it addresses the CPS, and answer any questions that the CPS poses for potential solutions.
+-->
+
+## Path to Active
+
+### Acceptance Criteria
+<!-- Describes what are the acceptance criteria whereby a proposal becomes 'Active' -->
+
+### Implementation Plan
+<!-- A plan to meet those criteria. Or `N/A` if not applicable. -->
+
+## Copyright
+<!-- The CIP must be explicitly licensed under acceptable copyright terms. -->
+
+[CC-BY-4.0]: https://creativecommons.org/licenses/by/4.0/legalcode
+[Apache-2.0]: http://www.apache.org/licenses/LICENSE-2.0
--- a/cip-observe-script-type/README.md
+++ b/cip-observe-script-type/README.md
@@ -148,7 +148,7 @@ transaction_body =
 required_observations = set<scripthash>
 ```
 
-The required observations (field 23) is a set of script hashes that can be used to require the associated Plutus script to present in the witness set and is executed in the transaction. If a script hash is present but the corresponding Plutus script is not in the witness set, the transaction will fail in phase 1 validation. This way, Plutus validation scripts can know which observation scripts were executed in the transaction.
+The required observations (field 23) is a set of script hashes that can be used to require the associated Plutus script to present in the witness set and is executed in the transaction. If a script hash is present but the corresponding Plutus script is not in the witness set, the transaction will fail in phase 1 validation. This way plutus scripts can check the script context to know which observation scripts were executed in the transaction.
 
 ## Rationale: how does this CIP achieve its goals?
 <!-- The rationale fleshes out the specification by describing what motivated the design and what led to particular design decisions. It should describe alternate designs considered and related work. The rationale should provide evidence of consensus within the community and discuss significant objections or concerns raised during the discussion.
--- a/cip-observe-script-type/README.md
+++ b/cip-observe-script-type/README.md
@@ -33,7 +33,7 @@ Since observe validators are decoupled from actions, you can run them in a trans
 
 ## Motivation: why is this CIP necessary?
 <!-- A clear explanation that introduces the reason for a proposal, its use cases and stakeholders. If the CIP changes an established design then it must outline design issues that motivate a rework. For complex proposals, authors must write a Cardano Problem Statement (CPS) as defined in CIP-9999 and link to it as the `Motivation`. -->
-Often in a plutus validator you want to check "a particular Plutus script checked this transaction", but it's annoying (and wasteful) to have to have to lock an output in a script and then check if that output is consumed, or mint a token, or whatever else just to trigger script validation. 
+Often in a plutus validator you want to check "a particular (different) Plutus script checked this transaction", but it's annoying (and wasteful) to have to have to lock an output in a script and then check if that output is consumed, or mint a token, or whatever else just to trigger script validation. 
 
 Currently the main design pattern used to achieve this is a very obscure trick involving staking validators and the fact that you can withdraw 0 from a staking validator to trigger the script validation. A summary of the trick is:
 Implement all the intended validation logic in a Plutus staking validator, we will call this validator `s_v`. To check that this validator was executed in the transaction you check if the credential of `s_v` (`StakingCredential`) is present in `txInfoWdrl`, this guarantees that `s_v` was checked in validation. 
--- a/cip-observe-script-type/README.md
+++ b/cip-observe-script-type/README.md
@@ -189,7 +189,7 @@ This change is not backwards-compatible and will need to go into a new Plutus la
 ## Path to Active
 
 ### Acceptance Criteria
-- [] Fully implemented in Cardano.
+- [ ] These rules included within a official Plutus version, and released via a major hard fork.
       
 ### Implementation Plan
 - [ ] Passes all requirements of both Plutus and Ledger teams as agreed to improve Plutus script efficiency and usability.
--- a/cip-observe-script-type/README.md
+++ b/cip-observe-script-type/README.md
@@ -8,7 +8,7 @@ Authors:
 Implementors: []
 Discussions: 
     - https://github.com/cardano-foundation/CIPs/pull/418/
-Created: 2024-1-8
+Created: 2024-01-08
 License: CC-BY-4.0
 ---
 
--- a/cip-observe-script-type/README.md
+++ b/cip-observe-script-type/README.md
@@ -30,6 +30,7 @@ License: CC-BY-4.0
 <!-- A short (\~200 word) description of the proposed solution and the technical issue being addressed. -->
 We propose to introduce a new Plutus scripts type `Observe` in addition to those currently available (spending, certifying, rewarding, minting, drep). The purpose of this script type is to allow arbitrary validation logic to be decoupled from any ledger action. 
 Since observe validators are decoupled from actions, you can run them in a transaction without needing to perform any associated action (ie you don't need to consume a script input, or mint a token, or withdraw from a staking script just to execute this validator). 
+Additionally, we propose to introduce a new assertion to native scripts that they can use to check that a particular script hash is in `required_observers` (which in turn enforces that the script must be executed successfully in the transaction). This addresses a number of technical issues discussed in other CIPs and CPS such as the redundant execution of spending scripts, and the inability to effectively use native scripts in conjunction with Plutus scripts. 
 
 ## Motivation: why is this CIP necessary?
 <!-- A clear explanation that introduces the reason for a proposal, its use cases and stakeholders. If the CIP changes an established design then it must outline design issues that motivate a rework. For complex proposals, authors must write a Cardano Problem Statement (CPS) as defined in CIP-9999 and link to it as the `Motivation`. -->
@@ -53,7 +54,6 @@ forwardNFTValidator stateToken _ tkIdx ctx =  assetClassValueOf stateToken (txIn
 forwardMintPolicy:: AssetClass -> Integer -> ScriptContext -> () 
 forwardMintPolicy stateToken tkIdx ctx =  assetClassValueOf stateToken (txInInfoResolved (elemAt tkIdx (txInfoInputs (txInfo ctx)))) == 1 
 ```
-With this pattern DApps are able to process roughly 12-30 forwardNFTValidator UTxO's  per transaction without exceeding script budget limitations.
 The time complexity of this validator is **O(n)** where n is the number of tx inputs. This logic is executed once per input being unlocked  / currency symbol being minted. 
 The redundant execution of searching the inputs for a token is the largest throughput bottleneck for these DApps; it is **O(n*m)** where n is the number of inputs and m is the number of `forwardValidator` inputs + `forwardValidator` minting policies.
 Using the stake validator trick, the time complexity of the forwarding logic is improved to **O(1)**. The forwardValidator logic becomes:
@@ -70,8 +70,20 @@ stakeValidatorWithSharedLogic stateToken _rdmr ctx = assetClassValueOf stateToke
 For the staking validator trick (demonstrated above), we are simply checking that the StakingCredential of the the staking validator containing the shared validation logic is in the first pair in `txInfoWdrl`. If the StakingCredential is present in `txInfoWdrl`, that means the staking validator (with our shared validation logic) successfully executed in the transaction. This script is **O(1)** in the case where you limit it to one shared logic validator (staking validator), or if you don't want to break composability with other staking validator, 
 then it becomes **O(obs_N)** where `obs_N` is the number of Observe validators that are executed in the transaction as you have to verify that the StakingCredential is present in `txInfoWdrl`.
 
-
-This proposal makes this design pattern indepedent from implementation details of stake validators and withdrawals, and improves efficiency and readability for validators that implement it. 
+The proposed changes in this CIP enable this design pattern to exist indepedently from implementation details of stake validators and withdrawals, and improve efficiency and readability for validators that implement it. Furthermore, with the proposed extension to native scripts, we are able to completely get rid of the redundant spending script executions like so:
+```haskell
+observationValidator ::  AssetClass -> BuiltinData -> ScriptContext -> ()
+observationValidator stateToken _redeemer ctx = assetClassValueOf stateToken (valueSpent (txInfo ctx)) == 1
+```
+We simply include the script hash of the above `observationValidator` in the `required_observers` field in the transaction body and we lock
+all the UTxOs that we would like to share the same spending condition into the following native script:
+```json
+{
+  "type": "observer",
+  "keyHash": "OUR_OBSERVATION_SCRIPT_HASH"
+}
+```
+The above solution (enabled by this CIP) is more clear, concise, flexible and efficient than the alternatives discussed above.
 
 ## Specification
 <!-- The technical specification should describe the proposed improvement in sufficient technical detail. In particular, it should provide enough information that an implementation can be performed solely on the basis of the design in the CIP. This is necessary to facilitate multiple, interoperable implementations. This must include how the CIP should be versioned. If a proposal defines structure of on-chain data it must include a CDDL schema in it's specification.-->
@@ -154,6 +166,31 @@ required_observers = set<scripthash>
 ```
 
 The `required_observers` (field 23) is a set of script hashes that can be used to require the associated Plutus script to be present in the witness set or as a reference script and executed in the transaction. If a script hash is present but the corresponding Plutus script is not in the witness set or present in a reference script, the transaction will fail in phase 1 validation. This way plutus scripts can check the script context to know which observation scripts were executed in the transaction.
+
+### Native Script Extension
+
+The BNF notation for the abstract syntax of native scripts change as follows to reflect the new field.
+
+```BNF
+<native_script> ::=
+             <RequireSignature>  <vkeyhash>
+           | <RequireObserver>   <scripthash>
+           | <RequireTimeBefore> <slotno>
+           | <RequireTimeAfter>  <slotno>
+
+           | <RequireAllOf>      <native_script>*
+           | <RequireAnyOf>      <native_script>*
+           | <RequireMOf>        <num> <native_script>*
+```
+
+Native scripts are typically represented in JSON syntax. We propose the following JSON representation for the `RequireObserver` constructor:
+```JSON
+{
+  "type": "observer",
+  "keyHash": "OBSERVATION_SCRIPT_HASH"
+}
+```
+
 
 ## Rationale: how does this CIP achieve its goals?
 <!-- The rationale fleshes out the specification by describing what motivated the design and what led to particular design decisions. It should describe alternate designs considered and related work. The rationale should provide evidence of consensus within the community and discuss significant objections or concerns raised during the discussion.
--- a/cip-observe-script-type/README.md
+++ b/cip-observe-script-type/README.md
@@ -12,20 +12,6 @@ Created: 2024-01-08
 License: CC-BY-4.0
 ---
 
-
-<!-- Existing categories:
-
-- Meta     | For meta-CIPs which typically serves another category or group of categories.
-- Wallets  | For standardisation across wallets (hardware, full-node or light).
-- Tokens   | About tokens (fungible or non-fungible) and minting policies in general.
-- Metadata | For proposals around metadata (on-chain or off-chain).
-- Tools    | A broad category for ecosystem tools not falling into any other category.
-- Plutus   | Changes or additions to Plutus
-- Ledger   | For proposals regarding the Cardano ledger (including Reward Sharing Schemes)
-- Catalyst | For proposals affecting Project Catalyst / the Jörmungandr project
-
--->
-
 ## Abstract
 <!-- A short (\~200 word) description of the proposed solution and the technical issue being addressed. -->
 We propose to introduce a new Plutus scripts type `Observe` in addition to those currently available (spending, certifying, rewarding, minting, drep). The purpose of this script type is to allow arbitrary validation logic to be decoupled from any ledger action. 
--- a/cip-observe-script-type/README.md
+++ b/cip-observe-script-type/README.md
@@ -65,7 +65,7 @@ forwardWithStakeTrick obsScriptCred tkIdx ctx = fst (head stakeCertPairs) == obs
 ```
 IE check that the StakingCredential is in the first pair in the `txInfoWdrl`.  This script is **O(1)** in the case where you limit it to one Observe script, or if you don't want to break composability with other Observe scripts, then it becomes** O(obs_N)** where `obs_N` is the number of Observe validators that are executed in the transaction.
 
-This proposal just makes this design pattern indepedent from implementation details of stake validators and withdrawals, and improves efficiency and readability for validators that implement it. 
+This proposal makes this design pattern indepedent from implementation details of stake validators and withdrawals, and improves efficiency and readability for validators that implement it. 
 
 ## Specification
 <!-- The technical specification should describe the proposed improvement in sufficient technical detail. In particular, it should provide enough information that an implementation can be performed solely on the basis of the design in the CIP. This is necessary to facilitate multiple, interoperable implementations. This must include how the CIP should be versioned. If a proposal defines structure of on-chain data it must include a CDDL schema in it's specification.-->
@@ -76,8 +76,20 @@ Redeemer -> ScriptContext -> ()
 
 The type signature of the newly introduced `Purpose` will be:
 ```haskell
-Observe Integer -- ^ where integer is the index into the observations 
+Observe Integer -- ^ where integer is the index into the observations list.
 ```
+
+### Script context
+
+Scripts are passed information about transactions via the script context.
+We propose to augment the script context to include information about the observation scripts that are executed in the transaction.
+
+Changing the script context will require a new Plutus language version in the ledger to support the new interface.
+The change is: a new field is added to the script context which represents the list of observation scripts that must validate the transaction.
+The observation scripts in the list are represented by their hash. 
+
+The interface for old versions of the language will not be changed.
+Scripts with old versions cannot be spent in transactions that include observation scripts, attempting to do so will be a phase 1 transaction validation failure.
 
 A new field will be introduced into the script context:
 
@@ -104,6 +116,39 @@ data TxInfo = TxInfo
   }
 ```
 
+### CDDL
+
+The CDDL for transaction body will change as follows to reflect the new field.
+```
+transaction_body =
+  { 0 : set<transaction_input>             ; inputs
+  , 1 : [* transaction_output]
+  , 2 : coin                               ; fee
+  , ? 3 : uint                             ; time to live
+  , ? 4 : certificates
+  , ? 5 : withdrawals
+  , ? 7 : auxiliary_data_hash
+  , ? 8 : uint                             ; validity interval start
+  , ? 9 : mint
+  , ? 11 : script_data_hash
+  , ? 13 : nonempty_set<transaction_input> ; collateral inputs
+  , ? 14 : required_signers
+  , ? 15 : network_id
+  , ? 16 : transaction_output              ; collateral return
+  , ? 17 : coin                            ; total collateral
+  , ? 18 : nonempty_set<transaction_input> ; reference inputs
+  , ? 19 : voting_procedures               ; Voting procedures
+  , ? 20 : proposal_procedures             ; Proposal procedures
+  , ? 21 : coin                            ; current treasury value
+  , ? 22 : positive_coin                   ; donation
+  , ? 23 : required_observations           ; New; observation scripts that must execute in phase 2 validation
+  }
+
+required_observations = set<scripthash>
+```
+
+The required observations (field 23) is a set of script hashes that can be used to require the associated Plutus script to present in the witness set and is executed in the transaction. If a script hash is present but the corresponding Plutus script is not in the witness set, the transaction will fail in phase 1 validation. This way, Plutus validation scripts can know which observation scripts were executed in the transaction.
+
 ## Rationale: how does this CIP achieve its goals?
 <!-- The rationale fleshes out the specification by describing what motivated the design and what led to particular design decisions. It should describe alternate designs considered and related work. The rationale should provide evidence of consensus within the community and discuss significant objections or concerns raised during the discussion.
 
@@ -114,10 +159,12 @@ It must also explain how the proposal affects the backward compatibility of exis
 
 ### Acceptance Criteria
 <!-- Describes what are the acceptance criteria whereby a proposal becomes 'Active' -->
-
+- [] Fully implemented in Cardano.
+      
 ### Implementation Plan
 <!-- A plan to meet those criteria. Or `N/A` if not applicable. -->
-
+- [] Passes all requirements of both Plutus and Ledger teams as agreed to improve Plutus script efficiency and usability.
+      
 ## Copyright
 <!-- The CIP must be explicitly licensed under acceptable copyright terms. -->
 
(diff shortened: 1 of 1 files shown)


cip-observe-script-type/README.md
```diff
@@ -0,0 +1,178 @@
...
+CIP: ?
+Title: Observe Script Type
+Status: Proposed
+Category: Plutus
+Authors:
+    - Philip DiSarro <info@anastasialabs.com>
+Implementors: []
+Discussions: 
+    - https://github.com/cardano-foundation/CIPs/pull/418/
+Created: 2024-1-8
```
Ryun1, line 11:
```suggestion
Created: 2024-01-08
```
fix date format :)

cip-observe-script-type/README.md
```diff
@@ -0,0 +1,178 @@
...
+Scripts are passed information about transactions via the script context.
+We propose to augment the script context to include information about the observation scripts that are executed in the transaction.
+
+Changing the script context will require a new Plutus language version in the ledger to support the new interface.
+The change is: a new field is added to the script context which represents the list of observation scripts that must validate the transaction.
+The observation scripts in the list are represented by their hash. 
+
+The interface for old versions of the language will not be changed.
+Scripts with old versions cannot be spent in transactions that include observation scripts, attempting to do so will be a phase 1 transaction validation failure.
+
```
michaelpj, line 99:
```suggestion
```

michaelpj (approved):
I'm in favour of this. There is clearly a need; the existing workaround is gross; this is a clean and orthogonal addition that also mirrors other functionality (e.g. `requiredSigners`).

cc @Quantumplation , I think you have also been thinking about this

rphair:
@colll78 we're not adding this for discussion in today's CIP meeting, where new proposals are normally introduced in Triage, because of the `Draft` status (as also with #735).  Please feel free to leave it in this state as long as you need... then we will proceed to meeting discussion as well as more rigorous community review (though leaving it here for public comment now is also welcome).

@michaelpj's positive feedback is also noted but generally we wouldn't try to solicit expert opinions about this while it's still in `Draft` state.

Quantumplation:
Yea, I've been thinking about this for a while, and @colll78 and I have discussed it before. It's clearly the "right" way to express this concept, if this was greenfield work.

My only reservations are that it adds more complexity to the ledger, for largely aesthetic / *very* small micro-optimization benefits, without any big material advantage over the stake script approach.

I'd much rather have @colll78's output redeemers proposal, for example.

colll78:
> Yea, I've been thinking about this for a while, and @colll78 and I have discussed it before. It's clearly the "right" way to express this concept, if this was greenfield work.
> 
> My only reservations are that it adds more complexity to the ledger, for largely aesthetic / _very_ small micro-optimization benefits, without any big material advantage over the stake script approach.
> 
> I'd much rather have @colll78's output redeemers proposal, for example.

Generally, from developers in the community that I have discussed these with, most share @Quantumplation's opinion that output redeemers is a much more pressing and valuable proposal given that for the problems it addresses every solution currently employed has serious drawbacks (breaks composability, heavily bloats the ledger, bloats the UTxO set, not generalized, etc).

Where-as for this CIP, the withdraw-zero design pattern, although awkward, obscure and not optimal does provide equivalent functionality. Still given that nearly every major DApp in the ecosystem is relying on this withdraw-zero trick, clearly this is a desired functionality, and having it rely on some obscure mechanic about zero amount not being filtered from the `txInfoWdrl` arbitrarily enforcing staking script execution is clearly not healthy for the ecosystem (and it is inconsistent with how zero amount values are filtered from the txInfoMint so the same cannot be used to execute minting validators without any ledger action).

micahkendall:
This would greatly simplify explaining the pattern to new developers because we wouldn't need to explain it in terms of a hack or an unexpected feature of the ledger, as you have to with the withdraw-0 trick. There are many scenarios even with prominent developers where we have this long back and forth on why we're using a stake validator to export logic.

Additionally, where we use withdraw-0, we aren't sure of the future of this in V3 and whether there will be adjustments to prevent us using this. This makes depending on the pattern long-term (despite how necessary it is) troublesome.

cip-observe-script-type/README.md
```diff
@@ -0,0 +1,178 @@
...
+<!-- Describes what are the acceptance criteria whereby a proposal becomes 'Active' -->
+- [] Fully implemented in Cardano.
+      
+### Implementation Plan
+<!-- A plan to meet those criteria. Or `N/A` if not applicable. -->
+- [] Passes all requirements of both Plutus and Ledger teams as agreed to improve Plutus script efficiency and usability.
+      
+## Copyright
+<!-- The CIP must be explicitly licensed under acceptable copyright terms. -->
+
```
Ryun1, line 176:
```suggestion
This CIP is licensed under [CC-BY-4.0](https://creativecommons.org/licenses/by/4.0/legalcode).

```
to match license in the header

Quantumplation:
@michaelpj well, just to be clear, we and many other dApps now *are* depending on that 0 withdrawal trick, so I think we're stuck with that now. Removing that capability would constitute a really rough breaking change, unless you had some mechanism to "fake" the withdrawals for old plutus versions or something.

cip-observe-script-type/README.md
```diff
@@ -0,0 +1,178 @@
+---
+CIP: ?
```
rphair, line 2:
```suggestion
CIP: 112
```

rphair (commented):
Agreed to promote at today's CIP meeting 🚀

michaelpj:
> Removing that capability would constitute a really rough breaking change, unless you had some mechanism to "fake" the withdrawals for old plutus versions or something.

It would have to be done in a non-breaking way obviously, so in a new era.

Quantumplation:
Also worth noting that if this is added, native scripts could be extended to have a `script` type, similar to `sig`; This is similar to what we're doing in [Aicone](https://github.com/SundaeSwap-finance/aicone).

michaelpj:
To be clear, you mean we could add an assertion to native scripts that says that a particular script hash is in `observers`?

michaelpj:
That might in fact be a simpler and less controversial alternative to https://github.com/cardano-foundation/CIPs/pull/309

Quantumplation:
@michaelpj yep, exactly; Then, if you *did* want the semantics of "this script must be on the inputs", you could write an observer to check that. I'm not sure why Seba's proposal was controversial (I haven't read it in a long time), so I don't know if this would be less controversial but it certainly achieves similar goals.

colll78:
Yes, I agree that with the `required_observers` field that is introduced into the in the `transaction_body` in this proposal, extending native scripts to be able to check for observation scripts should be trivial, and would address many of the same problems that other CIPs were designed for:

https://github.com/cardano-foundation/CIPs/pull/309
https://github.com/cardano-foundation/CIPs/pull/418
https://github.com/cardano-foundation/CIPs/pull/440

rphair:
@colll78 (just bulleted your last CIP PR list so we can see the titles easily)

fallen-icarus:
Would native scripts also be executable as Observer scripts? I have use cases where a spending plutus script delegates spending authorization to the script address' staking credential. If the staking credential is a script, then I use the 0 ADA withdrawal trick to ensure the staking script was executed. Since native scripts can be staking credentials, the script address can be protected by a simple multisig.

I think it would be cleaner if the staking credential was executed as an Observer script - it would still determine which script to look for from the staking credential but would check the `txInfoObservations` for proper execution instead of the `txInfoWdrl`. But for a one-to-one replacement, native scripts would also need to be executable as observer scripts.

cip-observe-script-type/README.md
```diff
@@ -0,0 +1,188 @@
...
+<!-- The technical specification should describe the proposed improvement in sufficient technical detail. In particular, it should provide enough information that an implementation can be performed solely on the basis of the design in the CIP. This is necessary to facilitate multiple, interoperable implementations. This must include how the CIP should be versioned. If a proposal defines structure of on-chain data it must include a CDDL schema in it's specification.-->
+The type signature of this script type will be consistent with the type signature of minting and staking validators, namely:
+```haskell
+Redeemer -> ScriptContext -> () 
+```
+
+The type signature of the newly introduced `Purpose` will be:
+```haskell
+Observe Integer -- ^ where integer is the index into the observations list.
+```
```
fallen-icarus, line 86: Is there a reason why the purpose should include the index instead of the script hash like with minting policies and staking scripts? The spending purpose is the only one that does not provide the hash directly and I only ever use it to lookup the script's hash from the input being spent. It seems other developers do the same since there is now even this [CIP](https://github.com/cardano-foundation/CIPs/pull/754) which suggests changing the spending purpose to include the script's hash. I would much rather have the `Purpose` be:
``` Haskell
Observe ScriptHash -- ^ the hash of the script currently observing.
```

colll78: I would argue it should be the other way around, the spending purpose should include the index of the UTxO being spent. The standard usage for spending scripts is to find the input that is currently being validated against.

fallen-icarus: > the spending purpose should include the index of the UTxO being spent.

I think this makes sense for spending scripts but not for observation scripts. A `TxInInfo` contains a lot more information than just a `TxOutRef` or a `ScriptHash`. It probably doesn't make sense to put *all* of that information into the `Purpose` so an index place-holder makes sense. But for observation scripts, the only thing you can get from the `TxInfoObservations` is a `ScriptHash`, and a `ScriptHash` fits no problem in the `Purpose`. If the observation index serves no purpose other than to get the script hash for the current observation script, then doesn't it makes more sense to inline the script hash instead of using an index place-holder? Otherwise, the index is forcing the unnecessary extra step of looking up the script hash in the observation list. Do you personally have another use for the observation index in mind?

colll78: Yes, I agree. 

Spending Script purpose should be `Spending Integer ScriptHash` and observe purpose should be `Observe ScriptHash`

michaelpj: This is a weak spot in the current design of things to be sure. The thing that would be closest to the ledger would be to always provide a "pointer" to the thing that caused the run, in which case we should indeed probably change the `Spending` purpose to have an index instead. The current situation is a bit inconsistent; in some cases we resolve some of the information for you, in others not.

colll78: If we do decide to resolve information for ScriptPurpose, then we should make the information we resolve consistent across all the different constructors. I think `ScriptHash` is a good option in that case, since it is universally useful across the common script types (Minting, Spending, Staking). 

But I really like the idea of integer indices for everything because then we can move towards a future where we can work towards accessing the unique efficiency benefits from transaction determinism that other smart contract platforms (ie EVM) cannot have access to (especially if in the far future we have support for constant access indexed data structures directly in the script context). I feel like currently, we haven't even begun to scratch the surface of utilizing the advantages afforded to us via Cardano's transaction determinism. 

For instance, since we know at the time of transaction construction what the majority of the script context will look like, we can theoretically bring most lookups from O(n) to O(1). Especially with [TxOut Redeemers](https://github.com/cardano-foundation/CIPs/pull/735) you can imagine a world where common lookup operations like `findOwnInput`, `valueOf`, `findStateThreadOutput`, are all O(1). Instead of looking up elements on-chain, we look them up off-chain and provide the index to the on-chain computation which then only has to check that the element is indeed where we claim it is.

Quantumplation:
I was thinking about this a bit more; since one of the most common use cases for this is to allow a large batch of inputs to defer their logic to an observer script... is there a variant of this proposal that lets you just lock the output by the observe script?

i.e. write a script that checks the script purpose for `Observe` and checks the global script logic; then compute the hash, and spend funds to that script hash. Then in order to spend the input, (perhaps indicated with `Observe(input_idx)` in the redeemers? or something?) the transaction has to run the script with the observe idx;

remove the need to write the indirection script entirely. I'm not sure exactly how to make the puzzle pieces fit, but would be pretty cool to cut a bit more out.

michaelpj:
> is there a variant of this proposal that lets you just lock the output by the observe script?

I think the proposal to add an "observer script S" is present condition to the native script language is pretty much that, no?

colll78:
> > is there a variant of this proposal that lets you just lock the output by the observe script?
> 
> I think the proposal to add an "observer script S" is present condition to the native script language is pretty much that, no?

Yes. I am extremely excited about this. I think it would address so many other CPS/ CIPs with very little complexity. Should I add the native script extension to this CIP? Alternatively, I could contract seba to see if I can work it into that CIP?

solidsnakedev:
@michaelpj Can we revive this CIP https://github.com/cardano-foundation/CIPs/pull/231 ?, I think the implementation of `inputs as list` + `observable scripts` are imperative, this will allow us to fold a validation through the input list without having to use yet another trick called redeemer index https://github.com/Anastasia-Labs/design-patterns/blob/main/UTXO-INDEXERS.md

Quantumplation:
> > is there a variant of this proposal that lets you just lock the output by the observe script?
> 
> I think the proposal to add an "observer script S" is present condition to the native script language is pretty much that, no?

Oh, yea, you're right :sweat_smile:

michaelpj:
> Should I add the native script extension to this CIP?

I think you could include it here. It would make sense to do it in the same piece of work.

@solidsnakedev I don't think that's particularly relevant to _this_ CIP, let's keep it on-topic.

cip-observe-script-type/README.md
```diff
@@ -30,6 +30,7 @@ License: CC-BY-4.0
 <!-- A short (\~200 word) description of the proposed solution and the technical issue being addressed. -->
 We propose to introduce a new Plutus scripts type `Observe` in addition to those currently available (spending, certifying, rewarding, minting, drep). The purpose of this script type is to allow arbitrary validation logic to be decoupled from any ledger action. 
 Since observe validators are decoupled from actions, you can run them in a transaction without needing to perform any associated action (ie you don't need to consume a script input, or mint a token, or withdraw from a staking script just to execute this validator). 
+Additionally, we propose to introduce a new assertion to native scripts that they can use to check that a particular script hash is in `required_observers` (which in turn enforces that the script must be executed successfully in the transaction). This addresses a number of technical issues discussed in other CIPs and CPS such as the redundant execution of spending scripts, and the inability to effectively use native scripts in conjunction with Plutus scripts.
```
Quantumplation, line 33: If we're making a change to the native script version, we should probably also get things on POSIX time rather than slot times.

Quantumplation: Whether that's another CIP or part of this one, i'd probably lean to a separate CIP, but it'd be a shame to see one without the other.

colll78:
Updated the CIP to add the discussed extension to native scripts.

cip-observe-script-type/README.md
```diff
@@ -0,0 +1,225 @@
...
+  , ? 20 : proposal_procedures             ; Proposal procedures
+  , ? 21 : coin                            ; current treasury value
+  , ? 22 : positive_coin                   ; donation
+  , ? 23 : required_observers              ; New; observation scripts that must execute in phase 2 validation
+  }
+
+required_observers = set<scripthash>
+```
+
+The `required_observers` (field 23) is a set of script hashes that can be used to require the associated Plutus script to be present in the witness set or as a reference script and executed in the transaction. If a script hash is present but the corresponding Plutus script is not in the witness set or present in a reference script, the transaction will fail in phase 1 validation. This way plutus scripts can check the script context to know which observation scripts were executed in the transaction.
```
fallen-icarus, line 168: > The `required_observers` (field 23) is a set of script hashes that can be used to require the associated Plutus script ...

Native scripts should also be in the `required_observers` list. It should be:
> ... require the associated script ...

cip-observe-script-type/README.md
```diff
@@ -0,0 +1,225 @@
...
+
+It must also explain how the proposal affects the backward compatibility of existing solutions when applicable. If the proposal responds to a CPS, the 'Rationale' section should explain how it addresses the CPS, and answer any questions that the CPS poses for potential solutions.
+-->
+Currently Plutus scripts in a transaction will only execute when the transaction performs the associated ledger action (ie. a Plutus minting policy will only execute if the transaction mints or burns tokens with matching currency symbol). The only exception is the withdraw zero trick which relies on an obscure mechanic where zero amount withdrawals are not filtered by the ledger. Now using `required_observers` we can specify a list of Plutus scripts to be executed in the transaction independent of any ledger actions. The newly introduced `txInfoObservations` field in the script context provides a straightforward way for Plutus scripts to check that "a particular script validated this transaction".
```
fallen-icarus, line 200: Here, too, it focuses exclusively on plutus scripts.
> Currently Plutus scripts in a transaction will only execute when the transaction performs the associated ledger action ...

The same is true for native scripts.

cip-observe-script-type/README.md
```diff
@@ -0,0 +1,225 @@
...
+
+This change is not backwards-compatible and will need to go into a new Plutus language version.
```
fallen-icarus, line 202: I'm not sure what the equivalent of a `plutus language version` is for native scripts but I am guessing they will also need a new version.

fallen-icarus (commented):
The CIP reads as if only plutus scripts can be executed as observer scripts. ~~Native scripts should **also** be executable as observer scripts.~~ There are use cases where a spending plutus script delegates spending authorization to another credential. If that credential is a native multisig, then it can only signal approval ~~(without side-effects)~~ through the withdraw 0 ada trick. This cip (as it is currently written) does nothing to change this. ~~If native scripts can spend, mint, and stake, why can't they also observe?~~

Edit: It has been a while since I used native scripts and after re-reading some of the [documentation](https://github.com/IntersectMBO/cardano-node/blob/1.26.1-with-cardano-cli/doc/reference/simple-scripts.md) again, I think I am wrong about native scripts requiring ledger events. The docs only use them as witnesses which I believe act more like keys. But this does not change my view that native scripts should also be in the observe list since plutus scripts should be able to verify that a native script was actually executed without needing to look for side-effects like spending, minting, or withdrawals.

cip-observe-script-type/README.md
fallen-icarus: Unless I am wrong (if I am, please correct me) and there is way for plutus scripts to verify that a native script was executed **without** having to look for side-effects like spending/minting/withdrawing, then native scripts should also be in the observe list. This way plutus scripts can verify a native script was executed *and* native scripts can verify a plutus script was executed; all without having to rely on ledger side-effects. Multisig is a really good use-case for this.

colll78:
> e use cases where a spending plutus script delegates spending authorization to another credential. If that credential is a native multisig, then it can only signal approval through the withdraw 0 ada trick. There is no other way for plutus scripts to see that a native script was executed without visible side-effects like spending and minting. This cip (as it is currently written) does nothing to change this. If native scripts can spend, mint, and stake, why can't they also observe?

I want to minimize the implementation complexity of this CIP. I think a possible extension to allow native scripts to observe is better suited for a separate CIP.

michaelpj:
I actually think that it reduces the implementation complexity to allow native scripts there, and indeed I simply assumed that was the case. At the moment, anywhere you put a script hash you can use any script of any kind; it would be inconsistent and confusing for this to be the only place where that was not true.

lehins:
We already have a similar field for `reqSignerHashes`. What if we just promoted it to contain credentials, i.e. either a KeyHash or ScriptHash. This would provide a consistent story with other parts of the transaction that are unlocked by a script or a key witness.

What I am trying to say is, we don't need two separate field in the transaction that mean a similar thing, we could gracefully deprecate `reqSignerHashes` in favor of required observers that could contain either a KeyHash or ScriptHash. CC @WhatisRT 

Tangentially related note. I am not sure if it was discussed here or in other CIPs, since I don't follow plutus related CIPs too closely, but there is a fun fact: it is already possible to witness native scripts without too much overhead. Whenever a native script is in the TxOut that is listed in the `referenceInputs`, it will be enforced by the ledger, regardless if anything in the transaction is locked by that script or not.

cip-observe-script-type/README.md
```diff
@@ -0,0 +1,225 @@
...
+  , ? 17 : coin                            ; total collateral
+  , ? 18 : nonempty_set<transaction_input> ; reference inputs
+  , ? 19 : voting_procedures               ; Voting procedures
+  , ? 20 : proposal_procedures             ; Proposal procedures
+  , ? 21 : coin                            ; current treasury value
+  , ? 22 : positive_coin                   ; donation
+  , ? 23 : required_observers              ; New; observation scripts that must execute in phase 2 validation
+  }
+
+required_observers = set<scripthash>
```
lehins, line 165: I would suggest to rather have `required_signers` field be promoted to the more powerful `required_observers`. The nice thing about it we can do this in a backwards compatible manner in order to allow tooling for graceful transition:
```suggestion
  , ? 14 : required_observers              ; Upgraded `required_signers`
  , ? 15 : network_id
  , ? 16 : transaction_output              ; collateral return
  , ? 17 : coin                            ; total collateral
  , ? 18 : nonempty_set<transaction_input> ; reference inputs
  , ? 19 : voting_procedures               ; Voting procedures
  , ? 20 : proposal_procedures             ; Proposal procedures
  , ? 21 : coin                            ; current treasury value
  , ? 22 : positive_coin                   ; donation
  }

; addr_keyhash variant is included for backwards compatibility and will be 
; deprecated in the future era, because `credential` already contains `addr_keyhash`.
required_observers = nonempty_set<credential / addr_keyhash>
```

michaelpj:
> We already have a similar field for reqSignerHashes. What if we just promoted it to contain credentials, i.e. either a KeyHash or ScriptHash. This would provide a consistent story with other parts of the transaction that are unlocked by a script or a key witness.

A few thoughts:
1. The name would be confusing if we kept it.
2. Maybe it is confusing to conflate scripts-accepting-transactions with signatures-for-transactions-being-present? 
3. This implicitly commits us to supporting other kinds of credential if we added them.

But overall I think this is a good idea.

WhatisRT:
I like @lehins idea. Why don't we name it something like `extra_credentials`? To @michaelpj's point 3: I think that would actually be a positive.

lehins:
> 1. The name would be confusing if we kept it.

I wasn't suggesting to keep the `reqSignerHashes` name, see: "we could gracefully deprecate reqSignerHashes in favor of required observers"

> 2. Maybe it is confusing to conflate scripts-accepting-transactions with signatures-for-transactions-being-present?

I have no idea what you mean. In my understanding that field is a list of things (key hashes, plutus scripts, native scripts) that must be satisfied. It is no different from how, for example, we require certificates that are locked by a staking credential being satisfied. With this new fields, it is just a list of things that the transaction itself requires to be satisfied.

> 3. This implicitly commits us to supporting other kinds of credential if we added them.

First of all: what credentials?

Second of all: we can't just stick anything else into a `Credential` in some new era and expect it to work. That would be a tremendous amount of work to satisfy a change like that, because it would affect all things that are locked by a credential, which we can't do without turning a `Credential` into a type family parameterized on `era`. That would be a major refactoring effort! So, in an very unlikely situation that this would ever happen, this field would be the least of our worries.

cip-observe-script-type/README.md
lehins: > I don't think there's a need to change the field name 

We'd have to rename the field, since it's type in ledger would change.


That will not work, since it is not backwards compatible with the previous format
```
required_signers = [ * credential ]
```

Quantumplation:
This is the way I've always thought of this feature. In fact, in the CIP in my head I never got around to writing, the script purpose was going to be "sign", and conceptually a script was just providing its signature.

michaelpj:
> I have no idea what you mean. In my understanding that field is a list of things (key hashes, plutus scripts, native scripts) that must be satisfied.

Credentials mean different things in different places, they don't have a universal meaning. Public key credentials do: they always mean "the transaction must be signed by this key", but script credentials vary in terms of what it is they are "inspecting" and how that is arranged. That said: we certainly do mix pubkey-credentials and script-credentials in other places, and nobody is confused that the pubkey credentials mean "the transaction must be signed" and the script credentials mean "the script must accept this (in some context-specific way)". So I retract my worry.

> First of all: what credentials?

I don't have any concrete example, just a generic worry. But I agree that it's pretty unlikely that we ever change this and if we do, as Andre says maybe it's even a bonus.

colll78:
Okay, given the discussion above I will change the CIP to:

1. Allow native scripts in `required_observers` (potentially renamed to `required_credentials` or `extra_credentials`)

Re
> I would suggest to rather have required_signers field be promoted to the more powerful required_observers. The nice thing about it we can do this in a backwards compatible manner in order to allow tooling for graceful transition:

I am fine with this for the actual CDDL representation, but I would prefer if they were separated in the script context since from the perspective of smart contract development this will introduce another layer of inefficiency and complication. 

IE I want to check that the transaction is signed by a pkh , now instead of `elem pkh txInfoSignatories`, we need to pattern match on each element, or we need to promote the pkh to a `PubKeyCredential` and check if that is an element, both of which add unnecessary computation. The two lists `txInfoObservations` and `txInfoSignatories` would be used for very separate purposes and there isn't really ever a case where you would want to search both lists in one iteration so storing them together just will make search complexity higher (now you need to search through more elements).

solidsnakedev:
> Okay, given the discussion above I will change the CIP to:
> 
> 1. Allow native scripts in `required_observers` (potentially renamed to `required_credentials` or `extra_credentials`)
> 
> Re
> 
> > I would suggest to rather have required_signers field be promoted to the more powerful required_observers. The nice thing about it we can do this in a backwards compatible manner in order to allow tooling for graceful transition:
> 
> I am fine with this for the actual CDDL representation, but I would prefer if they were separated in the script context since from the perspective of smart contract development this will introduce another layer of inefficiency and complication.
> 
> IE I want to check that the transaction is signed by a pkh , now instead of `elem pkh txInfoSignatories`, we need to pattern match on each element, or we need to promote the pkh to a `PubKeyCredential` and check if that is an element, both of which add unnecessary computation. The two lists `txInfoObservations` and `txInfoSignatories` would be used for very separate purposes and there isn't really ever a case where you would want to search both lists in one iteration so storing them together just will make search complexity higher (now you need to search through more elements).

Agreed, this make sense, pattern matching will bring unnecessary computation to the script

Quantumplation:
Us plutus devs are extremely protective of each execution unit :sweat_smile:

lehins:
> I would prefer if they were separated in the script context

I think that probably would be alright, as long as we decide that the redeemer pointer index can never be translated to plutus context for that field.

In other words ScriptPurpose for that field will only ever contain a ScriptHash, but not the index (like it is the case for minting or spending purpose for example). Which I suspect is ok, since that will be just a set of hashes without any other useful information.

fallen-icarus:
Perhaps it is okay to just have scripts first in the `txInfoObservers` list for `ScriptContext`?

For starters, how often do you have a very large list of `txInfoSignatories`? Signing transactions with a bunch of keys is *very* difficult to do in a decentralized fashion. How many DApps actually deal with this? In my experience, most DApps would likely have only one or two keys in `txInfoSignatories` (if any). With that in mind, I did some quick benchmarking using aiken tests. I tested three scenarios where 6 scripts observed and 2 keys signed; the last script in the observers is the target script to make it as costly as possible. TBH, I think it is rare to even see 6 scripts in a single transaction so I think these results are over-estimates of what one would see.

Here is what aiken spit out:
```
    │ PASS [mem: 29710, cpu: 10719872] separate_list
    │ PASS [mem: 35482, cpu: 21560938] combined_list_keys_first
    │ PASS [mem: 28086, cpu: 16876974] combined_list_scripts_first
    ┕━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 3 tests | 3 passed | 0 failed
```
Even if the keys are first in the list, since it is already rare to have keys in the list when a DApp is executed, the difference is only about 6000 mem units per execution. After factoring in redundant executions where this check is done 30 times (to spend 30 UTxOs that delegate to the target script), that is only 180000 mem units (1.3% of total budget). When scripts are first in the list, the memory usage is basically identical to the `separate_list`.

I would guess just putting scripts before keys would be easier than trying to create two "virtual" lists in the script context. Even if the cddl puts keys first, the ledger can just reverse the list before including it in the script context; the scripts themselves don't need to be in ascending order. Furthermore, even if there are more keys in the list, if the scripts are first, the memory usage doesn't really change from `combined_list_scripts_first`.

<details><summary>Test source code:</summary>

      test separate_list() {
        let x = [
          "000000000000000000000000000",
          "000000000000000000000000001",
          "000000000000000000000000002",
          "000000000000000000000000003",
          "000000000000000000000000004",
          "000000000000000000000000005",
          "000000000000000000000000006"
        ]
        list.has(x,"000000000000000000000000006")
      }

      test combined_list_keys_first() {
        let x = [
          VerificationKeyCredential("000000000000000000000000000"),
          VerificationKeyCredential("000000000000000000000000001"),
          ScriptCredential("000000000000000000000000000"),
          ScriptCredential("000000000000000000000000001"),
          ScriptCredential("000000000000000000000000002"),
          ScriptCredential("000000000000000000000000003"),
          ScriptCredential("000000000000000000000000004"),
          ScriptCredential("000000000000000000000000005"),
          ScriptCredential("000000000000000000000000006"),
        ]
        list.has(x,ScriptCredential("000000000000000000000000006"))
      }

      test combined_list_scripts_first() {
        let x = [
          ScriptCredential("000000000000000000000000000"),
          ScriptCredential("000000000000000000000000001"),
          ScriptCredential("000000000000000000000000002"),
          ScriptCredential("000000000000000000000000003"),
          ScriptCredential("000000000000000000000000004"),
          ScriptCredential("000000000000000000000000005"),
          ScriptCredential("000000000000000000000000006"),
          VerificationKeyCredential("000000000000000000000000000"),
          VerificationKeyCredential("000000000000000000000000001"),
        ]
        list.has(x,ScriptCredential("000000000000000000000000006"))
      }

</details>

lehins:
> I would guess just putting scripts before keys would be easier than trying to create two "virtual" lists in the script context

It would certainly be easier for ledger to implement, since they are guaranteed to be script hashes first. This is a Set of credentials, so elements are always in a sorted order and script hashes come before the KeyHashes. I am not sure if Plutus scripts can rely on that fact, but that is what ledger is guaranteed to provide, regardless of the order they were encoded on the wire.

FYI, this already applies to Withdrawals as well, since that is a Map of Credentials

michaelpj:
I'm :-1: on doing something other than having a list of `Credential`s. It's inconsistent with how we do it elsewhere. If we want it to be a list of credentials then it should be a list of credentials.

lehins:
>  It's inconsistent with how we do it elsewhere. If we want it to be a list of credentials then it should be a list of credentials.

I am all for consistency as well. I am supporting @michaelpj on this one!

colll78:
`txInfoSignatories` and `txInfoObservations` both contain very different information. Combining them into a single field in the script context seems like it would only serve to complicate things, and is inconsistent with the current `txInfoSignatories` (which currently contains `PubKeyHash`).

It makes sense to have `Credential` be the type for withdrawal, because it accurately represents the information (ie the rewards account from which the withdrawal is occuring). When we consider `txInfoSignatories` and `txInfoObservations`, it doesn't make much sense at all. They aren't representing the same thing as scripts cannot actually sign anything. One list contains the set of script hashes corresponding to the scripts that were executed in the transaction, the other list contains the set of pubkey hashes corresponding to the signatures of the transaction. 

Re the above benchmarks, they don't encapsulate the actual issue at hand. Whether PubKeyCredentials or ScriptCredentials are first, searching for elements in whichever is not first becomes higher complexity. 

Here is benchmark that should help illustrate the issue:
```
separated list check for both:
  ExBudget {exBudgetCPU = ExCPU 11955188, exBudgetMemory = ExMemory 33421}

 combined list check for both:
   ExBudget {exBudgetCPU = ExCPU 36637392, exBudgetMemory = ExMemory 59607}
```
Nearly double the memory and over three times the CPU. 

Often, these checks will be executed across different validators where one validator might only care about `txInfoSignatories` and the other only would care about `txInfoObservations`. When a validator cares to check if a user signed the transaction there is no reason to assume it would also care about the scripts that executed in the transaction (and vise-versa). 

```haskell
pcredentials = 
  pconstant 
    [ ScriptCredential "6c7bfa6b888fb3e600d4d9505b4fbca905df8ac58ed623b7170ab12a" 
    , ScriptCredential "6c7bfa6b888fb3e600d4d9505b4fbca905df8ac58ed623b7170ab12b"
    , ScriptCredential "6c7bfa6b888fb3e600d4d9505b4fbca905df8ac58ed623b7170ab12c"
    , ScriptCredential "6c7bfa6b888fb3e600d4d9505b4fbca905df8ac58ed623b7170ab12d"
    , ScriptCredential "6c7bfa6b888fb3e600d4d9505b4fbca905df8ac58ed623b7170ab13a"
    , ScriptCredential "6c7bfa6b888fb3e600d4d9505b4fbca905df8ac58ed623b7170ab14a"
    , ScriptCredential "6c7bfa6b888fb3e600d4d9505b4fbca905df8ac58ed623b7170ab15a"
    , PubKeyCredential "7c7bfa6b888fb3e600d4d9505b4fbca905df8ac58ed623b7170ab12a" 
    , PubKeyCredential "7c7bfa6b888fb3e600d4d9505b4fbca905df8ac58ed623b7170ab12a"]
  
ourCred = pconstantData $ ScriptCredential "6c7bfa6b888fb3e600d4d9505b4fbca905df8ac58ed623b7170ab15a" 
ourSigCred = pconstantData $ PubKeyCredential "7c7bfa6b888fb3e600d4d9505b4fbca905df8ac58ed623b7170ab12a"
pscriptHashes = pconstant 
    [ "6c7bfa6b888fb3e600d4d9505b4fbca905df8ac58ed623b7170ab12a" 
    , "6c7bfa6b888fb3e600d4d9505b4fbca905df8ac58ed623b7170ab12b"
    , "6c7bfa6b888fb3e600d4d9505b4fbca905df8ac58ed623b7170ab12c"
    , "6c7bfa6b888fb3e600d4d9505b4fbca905df8ac58ed623b7170ab12d"
    , "6c7bfa6b888fb3e600d4d9505b4fbca905df8ac58ed623b7170ab13a"
    , "6c7bfa6b888fb3e600d4d9505b4fbca905df8ac58ed623b7170ab14a"
    , "6c7bfa6b888fb3e600d4d9505b4fbca905df8ac58ed623b7170ab15a"
    ]
ppubkeyHashes = pconstant ["7c7bfa6b888fb3e600d4d9505b4fbca905df8ac58ed623b7170ab12a", "7c7bfa6b888fb3e600d4d9505b4fbca905df8ac58ed623b7170ab12a"]

combinedListCheckForBoth = 
  pand'List 
    [ (pelem # ourCred # pcredentials)
    , (pelem # ourSigCred # pcredentials)
    ]
    
separatedListCheckForBoth = 
  pand'List 
    [ (pelem # pconstant "6c7bfa6b888fb3e600d4d9505b4fbca905df8ac58ed623b7170ab15a" # pscriptHashes)
    , (pelem # pconstant  "7c7bfa6b888fb3e600d4d9505b4fbca905df8ac58ed623b7170ab12a" # ppubkeyHashes)
    ]
```

> TBH, I think it is rare to even see 6 scripts in a single transaction so I think these results are over-estimates of what one would see.

This was the case in the previous round of DApps because most contracts were designed to sacrifice composability / DApp interoperability in favor of efficiency. Also, smart contract based account abstraction was not a popular topic and nothing serious was being developed in that domain. Now, the landscape is drastically different. If you look at composable systems like Lending Pond's Exo, or Spectrum Finance Bloom / Splash Protocol, or even Sundae's extendable order types you'll see that these models desperately want to maximize the number of scripts they can check for (because it directly translates to the number of composable conditions that they can support). 

> How many DApps actually deal with this? In my experience, most DApps would likely have only one or two keys in txInfoSignatories (if any).

Plutus Stateful Dynamic MultiSigs are extremely common in practice (often with many required signatures). Off the top of my head, the DApps that implement it are:

Genuis Yield
Liqwid Finance 
Wanchain
Newm
Splash 

But in practice, it doesn't matter whether it is one or two signatures, if `ScriptCredentials` are first, because to check for the presence of those one or two signatures, they would need to go through the whole list.

Quantumplation:
Note that much of Phil's performance concern disappears if we get arrays and use them for many of the script context fields 😅

michaelpj:
I'm not sure if that's true? If you're searching through a list you potentially have to look at all the elements, which is O(n) anyway, and I thought his worry was the extra overhead on looking at each element.

colll78:
>  If you're searching through a list you potentially have to look at all the elements, which is O(n) anyway, and I thought his worry was the extra overhead on looking at each element.

In practice, you don't need to look through each element, because we know what the list looks like at the time of transaction construction we can just pass in the index where the item we are looking for should be and fail if it is not there. This way, only assuming transactions are built correctly, it will succeed. 

https://github.com/Anastasia-Labs/design-patterns/blob/main/UTXO-INDEXERS.md

colll78:
Updated the CIP to remove the new field from the CDDL instead opting to promote the existing `required_signatures` as per the feedback above. Also updated the CIP to reflect that native scripts are allowed to be used as observers; although given that based on the feedback this is desired, the current method of achieving this (include native script inputs in reference inputs) should probably be removed from future versions in the same way that withdraw zero will be.

peter-mlabs:
Would love to see this come to fruition!

cip-observe-script-type/README.md
```diff
@@ -0,0 +1,224 @@
...
+- Wallets  | For standardisation across wallets (hardware, full-node or light).
+- Tokens   | About tokens (fungible or non-fungible) and minting policies in general.
+- Metadata | For proposals around metadata (on-chain or off-chain).
+- Tools    | A broad category for ecosystem tools not falling into any other category.
+- Plutus   | Changes or additions to Plutus
+- Ledger   | For proposals regarding the Cardano ledger (including Reward Sharing Schemes)
+- Catalyst | For proposals affecting Project Catalyst / the Jörmungandr project
+
+-->
+
```
rphair, line 28:
```suggestion
```
... etc [in all sections] ...

rphair (commented):
The feedback from Plutus reps at the CIP editor's meeting today was positive: particularly that the IOG Plutus Team would prefer this approach to what @SebastienGllmt had proposed in candidate CIP-0038 (https://github.com/cardano-foundation/CIPs/pull/309).

@colll78 I'm ready to ✅ this after you can please remove the template comments from the document:

colll78:
> The feedback from Plutus reps at the CIP editor's meeting today was positive: particularly that the IOG Plutus Team would prefer this approach to what @SebastienGllmt had proposed in candidate CIP-0038 (#309).
> 
> @colll78 I'm ready to ✅ this after you can please remove the template comments from the document:

done

rphair (approved):
as per https://github.com/cardano-foundation/CIPs/pull/749#pullrequestreview-1946937999... thanks @colll78.  We can mark this `Last Check` for next meeting if reviewed favourably by other editors.

@Ryun1 @Crypto2099 I was thinking the _Path to Active_ looks emaciated compared to other Plutus proposals but since the Plutus Team (cc @michaelpj) is looking kindly on its implementation then that may be OK.

rphair:
OK no objection from @michaelpj in last month regarding CIP-0035 compliance... so marking this `Last Check` as proposed above & @Ryun1 @Crypto2099 let's try to maintain some Plutus category visibility for this one.

cip-observe-script-type/README.md
```diff
@@ -0,0 +1,201 @@
...
+### Alternatives 
+
+- We could decide to accept the withdraw-zero staking script trick as an adequate solution, and just preserve the nonsensical withdraw zero case in future language versions. 
+- The staking script trick could be abstracted away from the developer by smart contract languages that compile to UPLC. 
+    - This can be dangerous since by distancing the developer from what is actually happening you open up the door for the developer to act on misguided assumptions. 
+
+## Path to Active
+
+### Acceptance Criteria
+- [] Fully implemented in Cardano.
```
Ryun1, line 192:
```suggestion
- [ ] These rules included within a official Plutus version, and released via a major hard fork.
```

cip-observe-script-type/README.md
```diff
@@ -0,0 +1,201 @@
...
+      
+### Implementation Plan
+- [] Passes all requirements of both Plutus and Ledger teams as agreed to improve Plutus script efficiency and usability.
```
rphair, line 195: p.s. to https://github.com/cardano-foundation/CIPs/pull/749#discussion_r1585200445 ... changing the other tickbox to proper GitHub format. 🤓
```suggestion
- [ ] Passes all requirements of both Plutus and Ledger teams as agreed to improve Plutus script efficiency and usability.
```

Ryun1 (approved):

GeorgeFlerovsky (approved):
I wholeheartedly endorse this CIP in its current form.

I also support the changes to the initial draft that have emerged from the PR discussion thread, so far.

My vote is for separate fields in TxInfo for signatures and observer scripts. Whereas, for CDDL, I'm fine with either a combined field or separate fields — which ever Ledger/Plutus teams prefer.

GeorgeFlerovsky:
One interesting corollary of this CIP (if I understand it correctly) is that we will now be able to attach conditional consent to a partially drafted tx, instead of waiting for the completed tx to absolutely consent via tx signature.

This would be quite useful in multi-party interactive tx drafting because the party finalizing the transaction body doesn't necessarily have to go back to all the counter-parties for signatures before submitting.

Concretely, Alice can conditionally consent by attaching her personalized script to the required observers, with the redeemer set to a `(Signature, Instruction)` that the script interprets as her authorized intent.

One of the spending/minting/staking scripts would require Alice's personalized script to be in the tx observers, so that other drafters can't remove her conditions.

(Alice's instructions should probably include some nonce tied to the transaction to prevent replay, but I'm sure that can be figured out)

We can already do something similar with signed tuples in spending/minting/staking scripts, but I think observer scripts will give us more general/novel versions of this pattern.

lehins:
We had a long discussion internally at the ledger team meeting and nobody really liked the name of "observer", because it doesn't portray the intent very well.
The suggested name that we came up with that everyone was in favor of was "guard"

So, this would mean that we would be introducing a new "Guard script type", instead of "Observe script type".

The reason why we felt for "observe" being a wrong name, was because the term assumes it cannot change the course of validation and is there purely for observation reasons, which is clearly not the case. While a "guard" is there to protect against undesirable content in a transaction. 
This name would also be inline with the naming of "Guardrails script" that is part of the constitution.

FTR, this comment is not suggesting any semantic changes, just the name for the concept.

Thoughts? Suggestions? Recommendations? Protests?

Quantumplation:
I'm fine with that.

Personally I just wanted to call it "Sign", and add the scripts to the required signers field, but if they have to be separate then guard is good.

GeorgeFlerovsky:
Yes, I also think "observer script" is not a great name. "Guard script" is better.

Furthermore, I like "guard" more than "sign".

rphair:
I'd be happy to change it to the new name straight away... but editors would need help since the term `observe` is in so many literals: notably the proposed JSON representation.

@colll78 @GeorgeFlerovsky @Quantumplation @fallen-icarus can someone give us a PR that updates to the new term?  I'll make sure it's pushed through review before too many people continue using the old term.  cc @lehins

colll78:
I'm okay with changing the name, but I don't think `Guard` is a great name. The reason that I gave the title Observe is because this is the only script type which does not directly correspond to "guarding" a given ledger action. All existing script types correspond to and "guard" a given ledger action. Observation scripts is in-fact the only script type which doesn't "guard" anything. 

Spending -> guards the spending of a UTxO at a script address
Minting -> guards the minting/burning of tokens of a currency symbol 
Rewarding -> guards the withdrawal of rewards from a script stake account
Certifying -> guards the registration / deregistration / delegation of a script stake credential
Voting -> guards the voting of a script DRep / CCM
Proposing -> guards parameter change and treasury withdrawal governance actions.
Observing -> guards no action

Every other script type is REQUIRED for a transaction with a corresponding ledger action to be valid, they all guard those actions. Observer scripts on their own do not actually guard anything; they don't correspond to any ledger action at all. The exact same transaction (in terms of impact on ledger state) can be valid with or without an observation script. It is only through other scripts that they "guard" things. 

An example is spending scripts forwarding their validation logic to an observation script. In this case, a spending script that is guarding the spending of a UTxO at a script address, has logic to check that a given observation script is present. The spending script is what is actually guarding, the observation script is just making an observation, in this case, it just so happens that the guarding logic of the spending script enforces that a given observation must be made.

micahkendall:
How is “access”?

GeorgeFlerovsky:
@colll78 It doesn't directly guard any action, but it's not a passive observer. The only reason to include such a script in a transaction is for other scripts (which are triggered by transaction components) to delegate their validation logic to it. Moreover, the transaction can't succeed validation unless this script succesfully executes.

I think "Guard script" make perfect sense for this script type: it is a generic guard of the transaction as a whole, presumably invoked because some specifically-triggered script in the transaction needs it.

Guarding -> guards the transaction because some other script requested it

colll78:
> The only reason to include such a script in a transaction is for other scripts (which are triggered by transaction components) to delegate their validation logic to it. Moreover, the transaction can't succeed validation unless this script successfully executes.

This is not true. There are valid use-cases where an observation script is included even when it's not required by other scripts, to simply make a verified observation about the transaction. This is especially relevant for cross-chain interoperability protocols, L2 interop, bridging, and BTC-DeFi. 

> Moreover, the transaction can't succeed validation unless this script successfully executes.

Strictly speaking yes, that exact serialized transaction cannot succeed unless the script successfully validates, however, that is a property that is not unique to observation scripts, that is true of every script type supported by Cardano. Importantly, while the exact serialized tx cannot succeed, a transaction which has an identical impact on the ledger state (same UTxOs spent, same outputs produced, same rewards withdrawn, same credentials registered, etc) can indeed succeed without the inclusion of an observation script, unless that observation script is required by some other script which actually guards ledger state transitions.

lehins:
> Every other script type is REQUIRED for a transaction with a corresponding ledger action to be valid

If a script hash is supplied in the transaction, then it is also REQUIRED, so I don't see you point.
Observation implies it cannot affect the outcome of transaction validation, which is not true.

> Observer scripts on their own do not actually guard anything; 

I respectfully disagree with this. They guard a transaction as a whole. If a script fails, then the whole transaction will be considered as phase2 invalid, just like any other plutus script in a transaction.

I could agree that addition to the native script `RequireObserver` can be viewed as an observer, since all it does is observing that the required script has been added to required scripts necessary for validation. However, this feature is not limited to native scripts and any arbitrary plutus can be included, which does not necessarily depend on any other scripts in a transaction, thus acting as transaction guarding script.

@colll78 One way or another we will not be naming this feature "observer scripts" in the ledger implementation, since we do not believe this naming accurately represents its semantics. I would prefer to avoid introducing new terminology, and instead it would be better for the same naming to be used everywhere. But, I will avoid wasting my time on name bikeshedding and if we have to use different naming internally will to do so.

lehins:
Just out of curiosity a simple poll:
* Thumbs up for naming this CIP: "Guard script type"
* Thumbs down for keeping the name for this CIP: "Observe script type"

colll78:
I don’t disagree that observer isn’t a great name, but neither is guard because every script type is indeed a guard, it doesn’t give any information about what this script type does.

Quantumplation:
What about Criteria scripts? as in, additional criteria that must be true for the transaction to pass? It is somewhat vulnerable to the same criticisms as guard (since all scripts are criteria that must pass), but seems to be to be slightly more distinct.

lehins:
> every script type is indeed a guard

Scripts are indeed guards, that is why I like the term. All scripts are guards because the protect individual aspects of a transaction, while the script type in this CIP would guard the transaction as a whole.

> it doesn’t give any information about what this script type does.

That is because it can do anything, but in the end it results in pass or fail, which will dictate whether the whole transaction is valid or not.

> What about Criteria scripts?

A set of criteria usually implies a non-binary decision, (eg. judge completeness of a project by a set of criteria), while "guard" term implies pass or fail, which is why I personally like it better.

lehins:
"Guard" term is so good that even Haskell has it :wink: 
```haskell
changeName participants = do
  guard (all agreeOnNewName participants)
  Just "New name is Guard"
```

dragospe:
Perhaps "Assertion" script and "Asserting" script purpose? 

In addition to the existing usage of "guard" within plutus, an additional issue with `guard` is that the `guard :: Alternative f => Bool -> f ()` function in Haskell doesn't (always) throw, while `assert :: Bool -> a -> a` does.

solidsnakedev:
`Predicate scripts` is a naming convention used in other blockchains. They're always called to check some preconditions, have no side effects, and simply return either true or false.

lehins:
> while assert :: Bool -> a -> a does.

It doesn't throw anything unless `ignore-asserts` flag is turned off, which is on by default. :smile: In any case, Haskell example was brought in by me as a joke.
Joking aside, assertion assumes complete termination due to assertion failure, instead of a decision on validity. The fact that Plutus throws an error, is not visible to ledger, since for ledger the outcome of the script is returning False instead of True, which still leads to a transaction being placed on chain, albeit as an invalid one.

> Predicate scripts is a naming convention used in other blockchains. 

I would prefer to not use term predicate, since ledger is already full of different predicate checks. This would lead to confusion.

I honestly was trying to avoid name bikeshedding. The whole reason why I brought up this discussion is because we already found a good name that we are planning to use in the implementation. My suggestion was to also change it in the CIP, so it would be consistent, but this is not strictly necessary. I personally would like to use consistent naming, since it would avoid confusion when communicating the concept outside of ledger, but this wouldn't be the first time when a concept is named differently in implementation when comparing to how it is called by the community. So, I won't be crying if this CIP is not gonna adopt the term "guard". That being said, since many agree that Observer is a terrible name, including @colll78 the author, I'd highly recommend renaming it.

lehins:
Could I ask adding [Category: Ledger](https://github.com/cardano-foundation/CIPs/issues?q=state%3Aopen%20label%3A%22Category%3A%20Ledger%22) label to this PR, since it is very much ledger related

rphair:
@lehins periodically we've revisited the suggestion for CIPs to have multiple categories, and each time have turned back from it.  It may be time to revisit that again now that category support is detaching from the original organisations and stakeholders / key developers are distributing into multiple working groups on different node projects.

For some of these reasons we're also trying to define an extended vocabulary for CIPs and CPSs.  We've already underlined the reasons for not including more than 1 or 2 new keywords, and for those a conservative limit on their possible values.

For the new tags we have already specified that more than one choice (or no choice) could be applied for each keyword.  I hadn't yet thought of making `Category:` also a multiple-select: which is what we would have to do to honour your request.

If you want to pursue this particular update — and we will need advocates across the community to pursue the extended vocabulary at all, and so would welcome your help & input — would you please submit an argument here for a multiple-select of `Category:` and so try to build up some momentum in this otherwise stalled discussion?

https://github.com/cardano-foundation/cf-cip-frontend/discussions/55

lehins:
> would welcome your help & input — would you please submit an argument here for a multiple-select of Category: and so try to build up some momentum in this otherwise stalled discussion?

Arguably, this CIP affects Ledger more than it does Plutus codebase, hence is the suggestion. There are no changes to the plutus code needed, aside from changes to the plutus context definition which is usually specified and constructed by the Ledger component.

I believe it would make sense to tag CIPs with multiple categories, when they affect multiple components.

I believe this CIP was tagged as Plutus, because it affects DApp developers. But from implementation perspective, this has little to no impact on the Plutus component.

rphair:
@lehins in the meantime — while deciding whether to admit multiple categories (**edit, p.s.** I've posted your request here: https://github.com/cardano-foundation/cf-cip-frontend/discussions/55#discussioncomment-14769398) — would you like the editors to prepare a pull request to change the category from `Plutus` to `Ledger`?  cc @colll78 @Quantumplation @fallen-icarus @WhatisRT @GeorgeFlerovsky @michaelpj

PR #754, opened 2024-01-27
CIP-???? | Augment Spend Purpose

micahkendall:
We should change the purpose [Spend(OutputReference)](https://github.com/aiken-lang/stdlib/blob/97cd61345bcc8925c521b30d0f354859eb0148cd/lib/aiken/transaction.ak#L36C2-L36C25) to instead be Spend(x, ScriptHash)

EDIT: I have changed the proposal from Spend(OutputReference, ScriptHash) to Spend(x, ScriptHash), to indicate that this proposal is compatible with any changes to the first argument.

This would introduce asymptotic time savings for smart contracts using the pattern found in the [Fortuna hard fork](https://github.com/aiken-lang/fortuna/blob/d2e5831f92b1b596bb2cc1ab801509da28650fcc/validators/hard_fork.ak#L522), or described in[ my blogpost](https://forum.butane.dev/t/optimising-defi-stake-instead-of-spend/21?u=micah). The cost to search the inputs list for an output reference, and then to find its address is fairly high in my personal benchmarking. Every other purpose exposes the script hash in some form, this unifies the ability of a script to know its own address (cheaply). 

As an unverified, allegorical benchmark, we cut down from about 40k memunits per UTxO spent down to 14k memunits per UTxO spent by optimising around this limitation. We expect we could lift this limitation more with the acceptance of this CIP, and accomplish these numbers in an ergonomic, developer-friendly way.

Some different ways we currently optimise around this:
- parameterise a script with an NFT, putting the script hash into a utxo that owns that NFT, which we use to know the hash of the spending script, from reference inputs.
- put the script hash into the datum, in the case of validators locked by tokens

---

[Rendered proposal on a author's fork](https://github.com/butaneprotocol/CIPs/tree/master/cip-augment-spend-purpose)

### DIFF
--- a/cip-augment-spend-purpose/README.md
+++ b/cip-augment-spend-purpose/README.md
@@ -1,7 +1,7 @@
 ---
-CIP: ?
-Title: ?
-Category: ?
+CIP: ????
+Title: Augment Spend Purpose
+Category: Plutus
 Status: Proposed
 Authors:
     - Micah Kendall <micah@mcomp.tech>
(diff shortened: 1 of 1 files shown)


colll78:
This CIP synergizes well with the multi-validator trick and [CIP-112](https://github.com/cardano-foundation/CIPs/pull/749). It would provide a large benefit to DApp throughput using current design patterns (withdraw-zero trick), that would transfer over directly to the equivalent pattern with observation scripts. 

Personally, I would like to see this extended to:
`Spending Integer ScriptHash`

Where integer is the index of the input that is being spent in the `txInfoInputs` list. I haven't really ever seen `TxOutRef` be used in a way other than to search for the input being unlocked.

micahkendall:
> Personally, I would like to see this extended to: `Spending Integer ScriptHash`

I'm very happy with any use case which supersets the functionality I propose superseding this CIP! If we get the full input (as @microproofs wants), or we have minor adjustments on the first argument as you say, I'm happy.

Frankly I'm moving away from any pattern which uses the spend validation directly in all my code so anything goes as long as the ScriptHash is somewhere in the constructor.

longngn:
+1
It's very useful for a validator to know its own scriptHash and reduce the cost of un-packing and searching the inputs. This is especially prominent in the batching pattern where in 1 tx, we have dozens of script inputs that are doing this exact same thing. By reducing the cost of each script, we can marginally increase the total number of script inputs in the transaction (hence the throughput of Dapp).

Ryun1:
Hey @micahkendall,

Since this proposal is currently _very_ brief on all sections, I have converted it to a draft.

micahkendall:
> Hey @micahkendall,
> 
> 
> 
> Since this proposal is currently _very_ brief on all sections, I have converted it to a draft.

What needs elaboration?

rphair:
> @micahkendall: What needs elaboration?
- The single-sentence _Motivation_
  - e.g. what problems have you run into without this?  How often do these problems occur in general practice?
- The single-sentence _Specification_
  - what does the proposed code look like?  what do examples of its usage look like?
- The single-sentence _Rationale_: maybe OK; depends on how much is covered in _Motivation_ about why we are doing this & the alternatives.
- The cursory _Path to Active_: please see CIP-0035 to see what needs to be added for Plutus changes.
  - The requirements are strict for non-Core changes and even stricter for Core changes.
  - Please make them into checkboxes so implementors can mark these as resolved if & when it moves from `Proposed` to `Active`.
- _Copyright_ requires open-source licensing for a reason you may not have considered: a disclaimer of liability.  The community needs this as well as you.  Please choose CC or Apache as _required_ here for all CIPs: https://github.com/cardano-foundation/CIPs/blob/master/CIP-0001/README.md#licensing

cc @Ryun1

rphair:
p.s. @micahkendall a good start would be to put all the things from this PR's OP into the proposal itself.  My last answer was just to address the editing question literally.  But you have some good detail in your original posting that answers a lot of these questions & would at least provide a starting point for proper review _in the document itself_.

michaelpj:
I think we should instead go for consistent integer indexes and fix https://github.com/cardano-foundation/CIPs/pull/638 to get O(1) lookup in the inputs list.

colll78:
> I think we should instead go for consistent integer indexes and fix #638 to get O(1) lookup in the inputs list.

100% onboard with this. Would be an incredible improvement.

micahkendall:
@michaelpj 

If we're going to be getting O(1) lookup on inputs in time for Plutus V3 then I'm happy to withdraw this CIP. I think from a developer experience standpoint it's still sane to bring the information together, as it can hardly be considered random additional information when it is the exact information that every other script purpose variant contains. However to me that doesn't make it a meaningful improvement because I primarily care about performance.

michaelpj:
I think it's pretty unlikely that any further significant changes to the script context are going to hit PlutusV3, sorry.

Issue #731, opened 2024-01-02
Add References section to CIPs/CPSs


Ryun1:
Should we add a References section to CIPs or CPSs?

Inspired by [#638 comment](https://github.com/cardano-foundation/CIPs/pull/638#discussion_r1439333464)

rphair:
I'd be in favour of `## References`, but it should be an optional section and if so it would be our first optional section in the template.  That usually leads also to ambiguities about where it's placed, but most authors tend to put it before `## Copyright` and if they don't then we can just move it there.

I had actually been feeling bad about having to inline these Referenes which doesn't always suit the flow of what the author is saying.  I would been in favour of keeping it in the template after the last big remodel of CIP-0001 but none of the suggestions I made at the time were incorporated (though over time we've had to add all of them since then).

Unless objection from @Crypto2099 or possibly @KtorZ, and with a heads-up to @fill-the-fill @katomm who are parsing these for the Dev Portal, I can restore this in https://github.com/cardano-foundation/CIPs/pull/730.

michaelpj:
I am probably a great sinner but I often add additional sections to my CIPs. For example, the SOPs CIP had an `Appendices` section. 

I wonder if we should just say "you can add extra sections if that is useful, e.g. References, Appendices"? It's not clear that we benefit much from asserting that there are no additional sections, or that it makes rendering any harder.

rphair:
Parsability on derived sites like the Dev Portal CIP folder and `cips.cardano.org` was a main issue in the last major revision where this became a requirement.  @KtorZ was adamant about it & I've been trying to get him to chime in about implementing some of these rules universally in the older CIPs but he seems to have moved on.

So I think instead we have to consider the intent in why that was set up, and since really pushing #389 I had to think "Should we still do this?" about every one of these standardisations before implementing them over 40 times.  Mainly: most of these rules that were added after 2 years were added because of editing & indexing difficulties that we'd have again if those rules weren't there.

We don't see the consequences of extra sections much with a small archive, but we will when there are 10x as many CIPs: often with more than one covering the same issue & so will more frequently have to compare them in parallel.

Brevity and concise reference also become greater issues then: is a Reference useful without an inline link to the material which it refers to?  And is an Appendix, which some readers but not all will read... and which by nature is outside the flow of a CIP's essentials... not better off in a linked Markdown file: as are schemas, tables, and other material that might also be useful as externally linked technical documents in their own right?

[...]

PR #730, opened 2023-12-27
CIP-0001 | Structure and template revisions

rphair:
A few months since @Ryun1 made the general improvement to versioning expectations in #563, and now in the course of #389, we are seeing Versioning treatments stand out as a common feature & therefore looking again at adding a Versioning subsection (of Specification) as an obligatory heading.

This was considered to add to the remediation effort since https://github.com/cardano-foundation/CIPs/issues/389#issuecomment-1660405709 and now @Ryun1 and I think it would be easier for people to check in general if there were a common subheading to review.

This addition & how it is worded should of course be unanimously agreed upon by all active CIP editors.  CIP authors & reviewers who are interested in the versioning issue should also please voice any opinions & suggestions here.

It also fixes some spelling and other chronic annoyances like the `/pulls` URI syntax error in the 2 templates.

---

([updated CIP-0001](https://github.com/rphair/CIPs/blob/add-versioning-section/CIP-0001/README.md))
([updated CIP template](https://github.com/rphair/CIPs/blob/add-versioning-section/.github/CIP-TEMPLATE.md))
([updated CPS template](https://github.com/rphair/CIPs/blob/add-versioning-section/.github/CPS-TEMPLATE.md))

PR #756, opened 2024-01-30
CIP-9999 | Structure and template revisions

rphair:
To fix:
- the CPS half of https://github.com/cardano-foundation/CIPs/issues/731 (already fixed for CIPs)
- to legitimise the top level `References` section - and maybe other optional sections - as per discussion beginning https://github.com/cardano-foundation/CIPs/pull/638#discussion_r1439333464
- to complete what was started in #730 which added optional sections for the CIP but not for the CPS.

Whatever _optional sections_ we agree on for CPSs should be more reserved than for CIPs which have more of a precent for formality... otherwise the less formal problem statements will have arbitrary top-level sections everywhere for things that would be better focused into the 5 major headings so that CIP authors can more readily use them.
- This is why I left out the "custom" sections which we have recently admitted for the CIP format.

**BONUS QUESTION**: Should we add a _Copyright_ section to CPSs?
- Related question: why was _Copyright_ not added to the CPS in the first place, while it's always been a _de rigueur_ part of the CIP?
- asked already here: https://github.com/cardano-foundation/CIPs/pull/730#discussion_r1439757810

This is a short change initially but could end up longer if the format "rules" end up being more flexible or more rigorous.  cc @KtorZ @michaelpj

### DIFF
--- a/.github/CPS-TEMPLATE.md
+++ b/.github/CPS-TEMPLATE.md
@@ -42,3 +42,5 @@ Finally, goals may also serve as evaluation metrics to assess how good a propose
 
 ## Open Questions
 <!-- A set of questions to which any proposed solution should find an answer. Questions should help guide solutions design by highlighting some foreseen vulnerabilities or design flaws. Solutions in the form of CIP should thereby include these questions as part of their 'Rationale' section and provide an argued answer to each. -->
+
+<!-- OPTIONAL SECTIONS: see CIP-0001 > Document > Structure table -->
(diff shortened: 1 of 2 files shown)


Crypto2099 (approved):
Seems pretty straight forward. Looks good to me.

Ryun1:
> Should we add a Copyright section to CPSs?

It would add extra clarity and would not be too hard to retroactively enforce, due to the low number of CPSs.

Ryun1 (approved):
Simple, yet effective 😎

rphair:
> > Should we add a Copyright section to CPSs?

> @Ryun1: It would add extra clarity and would not be too hard to retroactively enforce, due to the low number of CPSs.

OK, let's merge this and keep building up the baseline slowly: especially because of potential liability & any legal consideration that might need to, or have been, done by the CF regarding copyright issues & warranty disclaimer.

Unless there's a posted reason not to, I'll submit a change for the Copyright heading in another PR: leading in with the primary consideration of whether to make Copyright an "optional section" as introduced here _or_ mandatory.  I don't personally see the need to make a license selection under a Copyright heading mandatory, but:

@KtorZ please this would be my 3rd attempt to try to find out why it was considered mandatory for CIPs but not CPSs... so I'd like to establish some background for that issue before introducing it in that new PR.  I still have no idea whether this decision was an oversight or deliberate.  It could be that it doesn't matter very much & the legal considerations aren't a big deal.  In any case please let us know whatever background there is, either by reply here or when the new PR comes along 🙏

PR #758, opened 2024-02-02
CIP-0128? | Preserving Order of Transaction Inputs

solidsnakedev:
This CIP is desired by Plutus developers to improve the validation efficiency of transaction inputs.

We propose the introduction of a new structure for transaction inputs aimed at significantly enhancing the execution efficiency of Plutus contracts.

This CIP facilitates explicit ordering of transaction inputs, diverging from the current state. This explicit ordering enables seamless arrangement of input scripts intended for utilization within the application's business logic.

This implementation was somehow discussed in an old CIP:
https://github.com/cardano-foundation/CIPs/pull/231

[Rendered Version](https://github.com/solidsnakedev/CIPs/tree/master/CIP-0128)


CIP-transaction-inputs-as-list/README.md
```diff
@@ -0,0 +1,155 @@
...
+- Meta     | For meta-CIPs which typically serves another category or group of categories.
+- Wallets  | For standardisation across wallets (hardware, full-node or light).
+- Tokens   | About tokens (fungible or non-fungible) and minting policies in general.
+- Metadata | For proposals around metadata (on-chain or off-chain).
+- Tools    | A broad category for ecosystem tools not falling into any other category.
+- Plutus   | Changes or additions to Plutus
+- Ledger   | For proposals regarding the Cardano ledger (including Reward Sharing Schemes)
+- Catalyst | For proposals affecting Project Catalyst / the Jörmungandr project
+
+-->
```
Crypto2099, line 26:
```suggestion
```

CIP-transaction-inputs-as-list/README.md
```diff
@@ -0,0 +1,155 @@
...
+
+## Abstract
+<!-- A short (\~200 word) description of the proposed solution and the technical issue being addressed. -->
```
Crypto2099, line 29:
```suggestion
```

CIP-transaction-inputs-as-list/README.md
```diff
@@ -0,0 +1,155 @@
...
+We propose the introduction of a new structure for transaction inputs aimed at significantly enhancing the execution efficiency of Plutus contracts. 
+
+This CIP facilitates explicit ordering of transaction inputs, diverging from the current state. This explicit ordering enables seamless arrangement of input scripts intended for utilization within the application's business logic. 
+
+Consequently, this implementation alleviates the necessity to set an index list through the redeemer to gather the required inputs from the transaction. Additionally, it mitigates the computational overhead associated with ensuring the uniqueness of the index list.
+
+## Motivation: why is this CIP necessary?
+<!-- A clear explanation that introduces the reason for a proposal, its use cases and stakeholders. If the CIP changes an established design then it must outline design issues that motivate a rework. For complex proposals, authors must write a Cardano Problem Statement (CPS) as defined in CIP-9999 and link to it as the `Motivation`. -->
```
Crypto2099, line 37:
```suggestion
```

CIP-transaction-inputs-as-list/README.md
```diff
@@ -0,0 +1,155 @@
...
+0 : [* transaction_input]    ; inputs
+```
+
+
+
+## Rationale: how does this CIP achieve its goals?
+<!-- The rationale fleshes out the specification by describing what motivated the design and what led to particular design decisions. It should describe alternate designs considered and related work. The rationale should provide evidence of consensus within the community and discuss significant objections or concerns raised during the discussion.
+
+It must also explain how the proposal affects the backward compatibility of existing solutions when applicable. If the proposal responds to a CPS, the 'Rationale' section should explain how it addresses the CPS, and answer any questions that the CPS poses for potential solutions.
+-->
```
Crypto2099, line 119:
```suggestion
```

CIP-transaction-inputs-as-list/README.md
```diff
@@ -0,0 +1,155 @@
...
+
+To mitigate such attacks, one approach is to compute the uniqueness of this index list, but this can result in costly computations if the business logic requires validation of a large number of inputs. A simpler approach to mitigate these hacks is to offer flexibility by explicitly allowing the order of inputs to be determined by a transaction builder instead of the ledger. This eliminates the need to pass indexes through the redeemer.
+
+## Specification
+<!-- The technical specification should describe the proposed improvement in sufficient technical detail. In particular, it should provide enough information that an implementation can be performed solely on the basis of the design in the CIP. This is necessary to facilitate multiple, interoperable implementations. This must include how the CIP should be versioned, if not covered under an optional Versioning main heading. If a proposal defines structure of on-chain data it must include a CDDL schema in its specification.-->
```
Crypto2099, line 82:
```suggestion
```

CIP-transaction-inputs-as-list/README.md
```diff
@@ -0,0 +1,155 @@
...
+
+<!-- OPTIONAL SECTIONS: see CIP-0001 > Document > Structure table -->
+[] Passes all requirements of both Plutus and Ledger teams as agreed to improve Plutus script efficiency and usability.
+
+
+## Copyright
+<!-- The CIP must be explicitly licensed under acceptable copyright terms.  Uncomment the one you wish to use (delete the other one) and ensure it matches the License field in the header:
+
+<!-- This CIP is licensed under [CC-BY-4.0](https://creativecommons.org/licenses/by/4.0/legalcode). -->
+<!-- This CIP is licensed under [Apache-2.0](http://www.apache.org/licenses/LICENSE-2.0). -->
```
Crypto2099, line 153:
```suggestion
```

CIP-transaction-inputs-as-list/README.md
```diff
@@ -0,0 +1,155 @@
...
+## Path to Active
+
+### Acceptance Criteria
+<!-- Describes what are the acceptance criteria whereby a proposal becomes 'Active' -->
+[] Fully implemented in Cardano.
+
+### Implementation Plan
+<!-- A plan to meet those criteria or `N/A` if an implementation plan is not applicable. -->
+
+<!-- OPTIONAL SECTIONS: see CIP-0001 > Document > Structure table -->
```
Crypto2099, line 145:
```suggestion
```

CIP-transaction-inputs-as-list/README.md
```diff
@@ -0,0 +1,155 @@
...
+However, it was determined that such approaches could introduce complexity and computational overhead
+
+#### 2. Hash the spending outref and use it for datum tagging: 
+  
+This alternative suggests hashing the spending outref of the script input and employing it for datum tagging to enforce the uniqueness of outputs. However, this method introduces extra overhead to the datum structure.
+
+## Path to Active
+
+### Acceptance Criteria
+<!-- Describes what are the acceptance criteria whereby a proposal becomes 'Active' -->
```
Crypto2099, line 139:
```suggestion
```

CIP-transaction-inputs-as-list/README.md
```diff
@@ -0,0 +1,155 @@
...
+
+It must also explain how the proposal affects the backward compatibility of existing solutions when applicable. If the proposal responds to a CPS, the 'Rationale' section should explain how it addresses the CPS, and answer any questions that the CPS poses for potential solutions.
+-->
+The motivation behind this CIP stems from the observed limitations and inefficiencies associated with the current lexicographical ordering of transaction inputs. Currently, the strict lexicographical ordering necessitates the passing of index numbers through the redeemer to select the appropriate input, leading to potential inefficiencies and vulnerabilities to vector attacks.
+
+To mitigate these issues, the proposed solution suggests transitioning from a set-based representation of transaction inputs to a list-based representation.
+
+This CIP tries to revive the original draft [CIP-0051](https://github.com/cardano-foundation/CIPs/pull/231)
```
Crypto2099, line 124: Do we need a new CIP to attempt to revive an old/previously not implemented CIP? What were the rationale or reasons that the earlier CIP were not adopted?

CIP-transaction-inputs-as-list/README.md
```diff
@@ -0,0 +1,155 @@
...
+
+Specifically, the inputs are currently represented as a set:
+```
+0 : set<transaction_input>    ; inputs
+```
+
+The proposed solution suggests modifying the inputs to a list format:
+```
+0 : [* transaction_input]    ; inputs
+```
```
Crypto2099, line 111: Does this modification mean that this is actually a proposal to change the **Ledger** rather than the **Plutus** scope and it just so happens to benefit contract developers working in Plutus?

CIP-transaction-inputs-as-list/README.md
Crypto2099: After reviewing the referenced PR it seems like there was strong pushback against this change back in 2022 and very strong and valid arguments against this change from members of both the Plutus and Ledger teams.

solidsnakedev: Back in 2022, hardly any projects were diving into new design patterns in order to optimize their Plutus contracts and allow for concurrent executions
But fast forward to now, most projects are adopting new design patterns that heavily rely on passing lists of indices through the redeemer, in order to identify the right inputs for their validation logic.
During 2022 the context was different, because the previous CIP didn't even have any specific use cases laid out.

CIP-transaction-inputs-as-list/README.md
solidsnakedev: Indeed it is not only about Plutus developers. Removing computation overhead leads to lower transaction fees for end-users. Moreover, it lowers the likelihood of potential exploits and opens up the possibility of including more inputs in a single transaction.

rphair: @Crypto2099 @solidsnakedev whether the current category `Plutus` also might be changed to `Ledger` I think would be a matter of "project enlisting" as per which of those teams would first have to accept the responsibility for implementation: https://github.com/cardano-foundation/CIPs/blob/master/CIP-0001/README.md#project-enlisting)

... so you can also consider it in that regard.  As I see it, generally the Plutus requirements for efficiency and desirability of CIPs have been more strict: so it makes sense that `Plutus` would "win" as `Category:` in an apparent "tie".  

To think about "enslistment" I'd suggest comparing these (if you haven't already) to see which set of requirements is more appropriate to this change, and then you & editors can hopefully agree with the implementation team heads:
- [CIP35: Changes to Plutus Core](https://github.com/cardano-foundation/CIPs/blob/master/CIP-0035/README.md) cc @michaelpj 
- [CIP84: Cardano Ledger Evolution](https://github.com/cardano-foundation/CIPs/blob/master/CIP-0084/README.md) cc @WhatisRT

Crypto2099: This is (via implementation) a modification that would need to be made to the ledger rules in order to go into effect as it would affect all transactions and is not specifically a change to Plutus Core. Thus my suggestion that it should be better categorized as a Ledger modification with a rationale that happens to be supported by benefits that exist to Plutus developers. There is really no functional change for "standard" transactions.

CIP-transaction-inputs-as-list/README.md
Crypto2099: This seems like a valid rationale to at least reopen the conversation here, although I believe a lot of the criticisms about design fragility (and subsequent lack of interoperability) when building systems that are dependent on both/either input or output ordering are still valid criticisms of this approach and would need to be overcome with some overwhelming amount of evidence that this would drastically improve QoL (Quality of Life) for developers and/or users before having a hope of seeing the work necessary to accommodate this change be placed onto the roadmap.

Ryun1:
This proposal would benefit with comments from the IOG Plutus and Ledger teams 🙏
@michaelpj @lehins

CIP-transaction-inputs-as-list/README.md
```diff
@@ -0,0 +1,128 @@
...
+
+## Rationale: how does this CIP achieve its goals?
+The motivation behind this CIP stems from the observed limitations and inefficiencies associated with the current lexicographical ordering of transaction inputs. Currently, the strict lexicographical ordering necessitates the passing of index numbers through the redeemer to select the appropriate input, leading to potential inefficiencies and vulnerabilities to vector attacks.
+
+To mitigate these issues, the proposed solution suggests transitioning from a set-based representation of transaction inputs to a list-based representation.
+
+This CIP tries to revive the original draft [CIP-0051](https://github.com/cardano-foundation/CIPs/pull/231)
+
+### Alternatives
+#### 1. Retain the existing set-based representation with additional validation mechanisms:
```
michaelpj, line 108: I can't work out what this is trying to say

CIP-transaction-inputs-as-list/README.md
```diff
@@ -0,0 +1,128 @@
...
+
+This CIP tries to revive the original draft [CIP-0051](https://github.com/cardano-foundation/CIPs/pull/231)
+
+### Alternatives
+#### 1. Retain the existing set-based representation with additional validation mechanisms: 
+
+This approach involves maintaining the current set-based representation of transaction inputs while implementing additional validation mechanisms to guarantee the uniqueness of index lists.
+However, it was determined that such approaches could introduce complexity and computational overhead
+
+#### 2. Hash the spending outref and use it for datum tagging:
```
michaelpj, line 113: also unsure what this means. examples would help

WhatisRT:
I don't fully understand the use-case. Why is the list of indices necessary? Can you not filter the set of inputs by their redeemer? Is that too slow? Also, how does having a list of inputs fix this issue? If your script expects all relevant inputs in the beginning of the list, it still needs to check that this is in fact the case, which should be almost as expensive as filtering them.

So if you need a list of indices without duplicates, the easiest way to check this is by also requiring the list to be sorted (failing if it isn't). Then it can be done easily in linear time. Is that maybe the problem? Do you want a list of indices that selects a sublist in an arbitrary ordering?

Like @michaelpj, I also don't understand the two points in the Alternatives section. The first looks like it may relate to some of the things I mentioned, but I'm not sure about that.

CIP-transaction-inputs-as-list/README.md
solidsnakedev: Hope this can clarify the current design most Dapps are applying, I can update the CIP to clarify this if needed.

https://github.com/Anastasia-Labs/design-patterns/blob/main/UTXO-INDEXERS.md

CIP-transaction-inputs-as-list/README.md
solidsnakedev: The other workaround is tagging the output with the hash of the outref, I can update the CIP to clarify with a PlutusTX example, let me know.

https://plutus.readthedocs.io/en/latest/reference/writing-scripts/common-weaknesses/double-satisfaction.html?highlight=double#unique-outputs

CIP-transaction-inputs-as-list/README.md
rphair: Yes @solidsnakedev please put it in the CIP itself 🤓

CIP-transaction-inputs-as-list/README.md
rphair: Yes @solidsnakedev I think example(s) will help here (and generally everywhere). 🤓

lehins:
In general I am in favor of this change for reasons unrelated to how Plutus see those inputs. I haven't gone through the CIP in detail, since that is unfortunately not going to be on our priority list until the next era. However, here is just a quick feedback.

First of all, it should not be a list of inputs, but an ordered set of inputs, because we can't allow duplicates in that field.

Furthermore this is a fairly complicated topic in general and we'd have to be very careful in order to get it right, because:
* we can't change the behavior of the old scripts, so PlutusV1 through PlutusV3 will probably continue getting those in sorted order
* redeemers currently rely on the sorted order, which is not good, so we'll need to address that somehow in a backwards compatible manner, which is gonna be pretty tricky.

So, as the CIP currently stands it is definitely far away from how it should be addressed.
But, I promise you that I will revisit this topic once Conway is released.

colll78:
> I don't fully understand the use-case. Why is the list of indices necessary? Can you not filter the set of inputs by their redeemer? Is that too slow? Also, how does having a list of inputs fix this issue? If your script expects all relevant inputs in the beginning of the list, it still needs to check that this is in fact the case, which should be almost as expensive as filtering them.

The benefit of being able to control the ordering of the inputs is not to move everything to the beginning of the inputs list. It is to be able to order the elements with respect to how they will be processed so that it becomes unnecessary for a validator to traverse the tx inputs multiple times to find all the inputs they need (and sort them into the order that the validator needs them in). Additionally, the benefit is that having full control over the order of the inputs vastly simplifies the following design pattern and makes it accessible to DApps that do not have extremely specialized (proprietary) offchain tooling:

```
validatorB :: AssetClass -> BuiltinData -> BuiltinData -> ScriptContext -> () 
validatorB stateToken _ _ ctx =
  let ownInput   = findOwnInput ctx
      authInput  = findAuth ctx  
      goodOutput = findOutputWithCriteria ctx
   in validate ownInput authInput goodOutput
    where
    findAuth :: ScriptContext -> Maybe TxInInfo
    findAuth ScriptContext{scriptContextTxInfo=TxInfo{txInfoInputs},                   
                           scriptContextPurpose=Spending txOutRef} =
        find (\TxInInfo{txInOutput} -> assetClassValueOf stateToken (txOutValue txInOutput) == 1) txInfoInputs
    findAuth _ = Nothing

    findOutputWithCriteria :: ScriptContext -> Maybe TxInInfo
    findOutputWithCriteria ScriptContext{scriptContextTxInfo=TxInfo{txInfoOutputs}} = find (\txOut -> criteria txOut) txInfoOutputs 
  ```
The above script is extremely inefficient. For every input and output we are searching for, we apply expensive checks to each element in the inputs / outputs until we find the elements that pass the checks. 

It can be vastly improved to:
```haskell
validatorB :: AssetClass -> BuiltinData -> (Integer, Integer, Integer) -> ScriptContext -> () 
validatorB stateToken (inputIdx, outputIdx, authIdx) _ ctx =
  let ownInput   = elemAt inputIdx
       authInput  = elemAt authIdx  
       ownOutput = elemAt outputIdx
   in 
      (assetClassValueOf stateToken (txOutValue (txInInfoOutput authInput)) == 1) -- check that element at authIndex does indeed have the auth token
      && (ownOutRef == txOutRef ownInput) -- check that element at input index is indeed the input that is being unlocked
      && (criteria ownOutput) -- check that the output at outputIdx does indeed satisfy the criteria required to unlock ownInput
    where
      txInfo  = scriptContextTxInfo context
      inputs  = txInfoInputs  txInfo
      outputs = txInfoOutputs txInfo
      Spending ownOutRef = scriptContextPurpose ctx
      findOutputWithCriteria :: ScriptContext -> Maybe TxInInfo
      findOutputWithCriteria ScriptContext{scriptContextTxInfo=TxInfo{txInfoOutputs}} = find (\txOut -> criteria txOut) txInfoOutputs 
  ```
  You can read more about the above:
  https://github.com/Anastasia-Labs/design-patterns/blob/main/UTXO-INDEXERS.md
 
This whole design pattern is designed to take advantage of the `Deterministic script evaluation` [property](https://github.com/IntersectMBO/cardano-ledger/blob/6781c2711c9ac3875f36f0da4e22a07724ff25f8/eras/alonzo/formal-spec/properties.tex#L348). Because inputs to Plutus scripts are fixed, we don't need to ever apply "search" for anything; because we know what the script context looks like at the time of transaction construction we can search for the element off-chain, and provide the location index of the element to the onchain code (via redeemer) and then the onchain code only needs to verify that the element at the provided index does indeed satisfy the criteria of what we are expecting. Even without O(1) index lookup this is extremely powerful (from O(n) conditional executions to O(1) conditional executions). When we get O(1) index access data structures this becomes even more powerful as it will allow validators to lookup anything in O(1) with this pattern. 

The issue is that without this CIP, this design pattern is extremely difficult to implement because of the complexity this introduces to offchain code. There is currently no open-source off-chain transaction framework that is capable of taking advantage of this design pattern. The issue that you run into when you attempt to use this pattern with existing offchain frameworks is as follows: When building the tx offchain, you search through the inputs, add the indices of the inputs you are looking for to the redeemer in the above example we pass redeemer as `new Constr(0, [INDEX_OF_OWN_INPUT, INDEX_OF_INPUT_WITH_BATCHERTOKEN])` But after, you balance the transaction new inputs are added to the transaction so the indices you put in the redeemer are no longer correct, and then if you adjust the redeemer again after the balancing to reflect the new locations of the inputs you are interested in, when you rebalance it can change the input set again. So currently you need to first put a dummy redeemer in with expected size as the actual redeemer, then perform balancing, and then swap out the dummy redeemer for the redeemer with the correct indices provided. 

Filtering by redeemer is far too expensive. Right now, nearly every smart contract protocol on Cardano has a `RequestValidator`

RequestValidator: The validator that users interact with. User sends a UTxO to the request validator and the datum of that UTxO descibes the action that the user wants to perform (and thus the action that the protocol is allowed to perform with that UTxO). Technical users run bots to continuously process these UTxOs in bulk (in some protocols this action can only be performed by a permissioned actor, where-as in others anyone can perform this action). 

For these protocols, the number of `requests` they can process in a transaction is extremely important because it represents the protocol throughput. The more `requests` they can process in a transaction, the higher throughput the protocol can achieve. So for instance a DEX that can process 50 requests in a transaction will have 5 times the throughput of a DEX that can only process 10 requests in a transaction. 

The criteria that is required to process a request depends on the content of the datum of the request. For instance, for a request UTxO for any of the major DEXs, the most common request type is swap. The criteria to process a swap request is:

For each request UTxO in the inputs (ie being processed by the transaction) there must be a corresponding output that pays out the requested amount of tokens (or an amount that is within some acceptable slippage percentage of that amount in which case the slippage percent allowed is also present in the request UTxO) to the address provided in the request UTxO's datum. A corresponding output that fulfills the request is not in and by itself sufficient. To process a request UTxO, you must also modify some global state (global in the sense it is shared across all request UTxOs in the transaction, ie a pool UTxO) correctly based on the content of the request. 

In general, the criteria for processing any request UTxO (of any type on any DApp in Cardano) typically requires that there must be one or more outputs (often just one) that correspond the to request UTxO, these outputs "fulfill" the request, and are commonly referred to as "destination outputs" or "payouts".

When processing requests in bulk, this means each request UTxO in the transaction inputs needs to be matched with one or more "destination outputs" / "payouts" in the transaction outputs. Because each request has some impact on a shared state (ie pool UTxO), the order in which requests are processed is important and DApps need to be able to efficiently control this.

The common way to "match" these request UTxOs with the outputs that fulfill them is to create two lists:
1. List of request UTxOs from transaction inputs `[i0, i1, i2, ..]`
2. List of payouts (from the transaction outputs) `[p0, p1, p2]` such that `p0` fulfills request `i0`, `p1` fulfills request `i1` and so forth. 

Then we traverse the lists together and check that the conditions required to fulfill each request are indeed fulfilled by the corresponding UTxO in the payouts list and that the pool state is adjusted properly with respect to each request. Currently locating this list of payouts is easy, we provide an index in the redeemer to the location of the first payout and then grab n elements starting at that index in the tx outputs list where n is the number of requests, this is linear time. The problem of creating the request UTxOs list is much harder. We cannot control the order of the tx inputs, so we cannot enforce that the first request utxos to be processed must be before the other request UTxOs in the list. This means we have to traverse the tx inputs list once for each request UTxO, thus the time complexity here is quadratic, or we have to get all the request inputs and then sort them which is also quadratic. 

This problem (of matching inputs to corresponding outputs) is broadly described in the following CPS:
https://github.com/cardano-foundation/CIPs/blob/5ec27da12ce01197fc9202f890e4914231ee9d4a/CIP-%3F%3F%3F%3F/README.md

> Do you want a list of indices that selects a sublist in an arbitrary ordering?
Yes.

**Comment by michele-nuzzi** [comment]
I'm sorry I struggle to see what problem does this CIP solve.

The inputs are ordered on the ledger, but don't have to be in the transaction.

The plutus script context is constructed based on the order of the inputs in the CBOR, not on the ledger.

I a transaction is built with a certain order, the order will be preserved in the script context, and will be ordered on the ledger (but that doesn't matter for plutus)

so the property you are looking for is already there.

Perhaps the tools you are using to build transactions will order the inputs, but that is not necessary; plu-ts-offchain preserves the inputs order as they are specified and I have never had problems with the transaction ordering.

P.S. redeemer indexing is handled by the ledger, so the redeemer index will need to be the one of the sorted set, not the one of the order of the transactio CBOR

**Comment by lehins** [comment]
> I a transaction is built with a certain order, the order will be preserved in the script context, and will be ordered on the ledger (but that doesn't matter for plutus)

That is actually not true.

The order in which inputs or redeemers are placed into the transaction is not preserver. It will always be sorted by ledger. Which is a problem that we would like to fix in ledger and a way to solve this would be to preserve the order of inputs in which they were placed on the wire.

**Comment by michele-nuzzi** [comment]
> That is actually not true.
> 
> The order in which inputs or redeemers are placed into the transaction is not preserver. It will always be sorted by ledger. Which is a problem that we would like to fix in ledger and a way to solve this would be to preserve the order of inputs in which they were placed on the wire.

I'll run some test transaction to make sure my understanding is correct

**Comment by WhatisRT** [comment]
Ok, so from the above discussion it seems to me like this would really solve the following two problems:
- Coin selection for the currently viable solutions to select (reordered) sublists of inputs is hard, and
- it is evidently unintuitive behaviour to not preserve the order.

I think these are somewhat good points, but for the sake of completeness, here are some counterarguments to those points:
- Coin selection was always going to be hard in the presence of scripts (we've had discussions about this before Alonzo was launched). A general solution to this problem probably has to accept slightly overpaying the fees sometimes. Fixing this problem here doesn't address the general issue.
- While unintuitive, it's consistent with how we present other information to scripts, first and foremost multi-asset bundles. It is possible to put them in an arbitrary order inside a transaction as well, and they'll just get sorted. And that's a good thing in this case, since people really don't think that two token bundles are different just because you swap out the order. And arguably that's the same with inputs: it doesn't matter in what order I'm taking my cash out of the drawer, just that it's sufficient to pay for whatever I want.

Now maybe this is one of these cases where a specific use-case trumps somewhat abstract arguments. I still consider the case for this somewhat weak, but I think I'm now at the point where I'd regard it as a nuisance rather than an anti-feature.

For completeness, let me repeat my main argument against preserving the order: It will likely make future scripts completely uninteroperable. It is likely that every script will simply assume that the inputs it cares about will be at the beginning, and two such scripts will be unable to run at the same time, even if it would have been perfectly fine to do it otherwise. For example, I know that (at least some versions of) the MuesliSwap script don't care at all about other inputs. So any user can do swaps more efficiently by combining them with some other transaction if they want to. I think that implementing this CIP closes the door on that opportunity for future scripts.

**Comment by michele-nuzzi** [comment]
I would add that it could turn incredibly useful to extend the CIP to reference inputs, also a set in the [babbage.cddl](https://github.com/IntersectMBO/cardano-ledger/blob/master/eras/babbage/impl/cddl-files/babbage.cddl#L70) hence ordered in the script context

**Comment by michaelpj** [comment]
> For completeness, let me repeat my main argument against preserving the order: It will likely make future scripts completely uninteroperable. It is likely that every script will simply assume that the inputs it cares about will be at the beginning, and two such scripts will be unable to run at the same time, even if it would have been perfectly fine to do it otherwise.

I think this argument proves too much, since the same argument applies today to _outputs_. Very many scripts care about outputs, if your argument holds then today they would all just assume that their outputs came first, rendering them non-interoperable. I think this means that making inputs a list can't make things much worse.

**Comment by solidsnakedev** [comment]
> I would add that it could turn incredibly useful to extend the CIP to reference inputs, also a set in the [babbage.cddl](https://github.com/IntersectMBO/cardano-ledger/blob/master/eras/babbage/impl/cddl-files/babbage.cddl#L70) hence ordered in the script context

I think it makes sense, I'll update the cddl accordingly

**Comment by ch1bo** [comment]
I just wanted to chime in on this proposal and mention that the [hydra](https://hydra.family/) protocol would benefit from this too as we have a quite constrained and not expected to compose validator. Furthermore, I remember we too were quite puzzled and annoyed of inputs being "re-ordered".

@lehins I see the `cardano-ledger` has a `OSet` type now which could be used on the `spendingInputs` of a tx body now? Besides that and updating the `plutus-ledger-api` to resemble this, e.g. in a `PlutusV4`, do you see any other steps this is needing?

**Comment by solidsnakedev** [comment]
I've update the CIP as we are getting close to Conway transition, addressing the concerns from @lehins @michaelpj @michele-nuzzi @rphair

**Comment by rphair** [comment]
thanks @solidsnakedev ... have put on CIP meeting agenda for Review next time; hope you can make it: https://hackmd.io/@cip-editors/93

**Comment by fallen-icarus** [comment]
> For completeness, let me repeat my main argument against preserving the order: It will likely make future scripts completely uninteroperable. It is likely that every script will simply assume that the inputs it cares about will be at the beginning, and two such scripts will be unable to run at the same time, even if it would have been perfectly fine to do it otherwise.

@WhatisRT I don't think this is true. I think this misses the economic utility that is possible with DApp composability. Imagine if there was an options contract you wanted to buy, but the sale price was in WMT while you only had DJED. You have your assets in DJED to protect yourself from market volatility. If you had to first convert your DJED to WMT in one transaction and then buy the options contract in another, you are exposing yourself to the market volatility of WMT while you wait to see if you can actually get the options contract (before someone else buys it). What if someone beats you to the options contract *after* you converted your DJED to WMT? You now need to convert it back to DJED, and you likely lost money (due to the tx fees + DApp fees + market volatility).

The above scenario is entirely avoidable if you just compose converting DJED to WMT with buying the options contract. If the options contract is bought before your transaction is processed, your composed transaction will fail due to the options contract UTxO being missing (no collateral is lost either since no scripts need to be run). This also means your DJED wasn't unnecessarily converted to WMT. Composing the actions **guarantees** that your DJED will only be converted to WMT if, and only if, you successfully buy the options contract. The risk of loss in the case where you don't get the options contract is entirely eliminated due to DApp composability. Risk management plays a huge role in economics (and regulations) and I think DeFi DApps that compose will out-compete those that do not. If corporations are also going to eventually use DeFi, they need to be able to manage their economic risk as much as possible. For the big players (governments, corporations, etc), throughput is no where near as important as risk management.

Another example that doesn't deal very much with risk management is the ability to unify the liquidity across all stablecoins. Currently, if you have DJED but need USDC to buy something, you need to take the same two step approach as in the previous example. The reason is, despite both DJED and USDC effectively being USD, smart contracts cannot securely know this and therefore, must treat them differently. As a consequence, the USD liquidity in DeFi is fractured across all stablecoins. But what if you could convert DJED to USDC (for a slight conversion fee) in the same transaction where you buy the item? If you could, the liquidity would no longer be fractured across all stablecoins; the ability to compose converting them with other DApps means DeFi would effectively have a single "meta" stablecoin. And again, the conversion would only happen if you successfully buy the item at the end of the composed chain of actions. DApps that sacrifice composability in the name of throughput will be cutting themselves off from this "meta" stablecoin liquidity. 

IMO, DApp composability is the killer feature for eUTxO. You can compose 5-10 separate DApps on Cardano right now; AFAIU this would be prohibitively expensive on an account style blockchain. I think this composability will make things possible in DeFI that aren't even possible in TradFi. I'm sure some DApps will sacrifice composability for throughput in the short-term (like they are currently doing), but I think the economic utility strongly favors composable DApps to the point where future DApps will prioritize composability. 

I am personally in favor of this CIP because I think it will actually help composability in certain scenarios. For example, for some of my protocols, the order of the protocol's required outputs depends on the order of the protocol's inputs. The *only* thing that matters is the order of the protocol's inputs/outputs: other inputs/outputs can be interspersed and the inputs/outputs can appear at any point in their respective lists. (This is for throughput reasons since it allows me to only traverse the inputs and outputs list once each.) Right now, I can't control the order of the inputs which means I can't control the order of the outputs. This doesn't stop me from being able to compose my protocols, but if there was another DApp where the input/output requirements were more strict (for whatever reason), that DApp would be more likely to compose with my protocols if the order of the inputs could be controlled (and therefore the order of the outputs could also be controlled). Even if the customized ordering is less efficient overall, composing the actions in a single transaction can actually save the end-user money on net, and decrease their overall economic risk.

File: CIP-transaction-inputs-as-unordered-set/README.md [hunk 13]
```diff
@@ -0,0 +1,140 @@
...
+
+Furthermore, this implementation offers the potential to enhance existing design patterns already used in production, specifically by improving the ability to group and match inputs and outputs located at the specific index positions.
+
+## Motivation: why is this CIP necessary?
+
+According to the Babbage CDDL [transaction_body](https://github.com/IntersectMBO/cardano-ledger/blob/master/eras/babbage/impl/cddl-files/babbage.cddl) , the inputs and reference inputs of a transaction body are represented as a set, which indicates the non-duplicate inputs. However, the ledger not only process the [transaction_inputs](https://github.com/IntersectMBO/cardano-ledger/blob/0274cf65dbb79773122b69dfd36a8299eec2783f/eras/babbage/impl/cddl-files/babbage.cddl#L75-L77) as a set , but also orders them lexicographically, first by `transaction_id` and then by `index`.
```
Ryun1, line 26:
```suggestion
According to the Babbage CDDL [transaction_body](https://github.com/IntersectMBO/cardano-ledger/blob/master/eras/babbage/impl/cddl-files/babbage.cddl) , the inputs and reference inputs of a transaction body are represented as a set, which indicates the non-duplicate inputs. However, the ledger not only process the [transaction_inputs](https://github.com/IntersectMBO/cardano-ledger/blob/0274cf65dbb79773122b69dfd36a8299eec2783f/eras/babbage/impl/cddl-files/babbage.cddl#L75-L77) as a set , but also orders them lexicographically, first by `transaction_id` and then by `index`.
```
maybe the CDDL should also reflect that is is a non_empty_set 🤔

lehins: CDDL does not prevent for inputs from being empty. It is the ledger rules that enforce this

WhatisRT:
@fallen-icarus I fully agree that composability is a fantastic feature, but that's the point I was making: the average script would be less composable with this feature. The cheapest way to get the inputs for your script is just to require that they are at the beginning of the list. So if you want your script to be composable, you now have to argue that it's better to linearly search through the list of inputs (or maybe include that information in the redeemer), both of which increases execution cost. So now the cheapest way to implement your script is non-composable.

fallen-icarus:
> So now the cheapest way to implement your script is non-composable.

This is what I was trying to argue is false. I think you are thinking about a specific script in isolation which I do not think is realistic. If Alice is considering buying an options contract with an asset she doesn't have, what matters to her is the *total* cost of the high-level action (DEX conversion + options purchase). Whether this requires one transaction or two transactions is (mostly) irrelevant to her.

In an extreme case, if Alice chooses to sacrifice composability, she could possibly save 1 ADA in execution costs for this high-level action. But since she couldn't compose, she is now subject to market volatility which could easily be 5%. If 1000 ADA was involved, the total extra cost to Alice is the 50 ADA volatility loss. 

However, if she chooses to compose, she pays the extra 1 ADA in execution costs, but doesn't have to deal with the market volatility so the total extra cost to her is just the 1 ADA in extra tx fees. Alice *saves* 49 ADA by composing! To break even would require market volatility of 0.1% which is extremely likely to be exceeded by pretty much all trading pairs...

The math seems to clearly favor composable DApps. *The cheapest way for end-users to use DApps is to compose them, even if individually the DApps are slightly more expensive.* 

> So if you want your script to be composable, you now have to argue that it's better ...

I don't think you need to argue anything. The market will naturally punish (through increased costs) those who choose to not compose DApps. The same goes for DApp developers who sacrifice composability. Users will gravitate towards DApps that allow them to accomplish their high-level actions more cheaply, which likely means composable DApps.

WhatisRT:
Well, you're making the argument that users want composable scripts, but the question is whether authors will make them. Writing and auditing a script is very expensive, and script authors might not care much about composability themselves. Also, users might simply not have the choice here: if somebody wants to execute a particular script, and it happens to be non-composable, then they only have the choice of not using it. Maybe they'll complain to the script author, but what's the realistic chance that the author will spend a bunch of extra money and effort on making a composable version? And will it be adopted properly? There are still massive amounts of Plutus V1 transactions being made.

So I think it's unlikely that market forces are strong enough to ensure composability. Companies really like to make walled gardens for all sorts of things, and lots of people dislike it but the dislike it clearly not strong enough to put pressure on them. I don't see why it would be different here.

rphair:
@WhatisRT I hope we are not saying here that core development should align itself with the needs of "walled garden" systems at the expense of support for progressive models that aren't in robust demand _yet_.  I could apply the same argument as above and say "Why would any blockchain application bean-count all these execution costs on Cardano when they can do the same thing with a cheap flood of transactions on Solana?"

Advocates for Cardano are expecting markets to be developed here that might not be able to exist anywhere else... I understand this is part of the reason why the eUTxO model was chosen in the first place despite the fact that "walled gardens" tend to find it easier to follow the accounts model.

I worry that if core development is diverted away from composability and transaction parallelism then Cardano is likely to be stranded with a set of capabilities that are implemented better elsewhere.  Where would we be then?  I'd rather see the groundwork developed now to support interoperable, progressive markets that competitors would _never_ be able to support.

In terms of "market forces" there is nothing more values than a "unique selling proposition" and all the CIPs and smart contract libraries in favour of composability, parallelism, and peer-to-peer markets have to be assessed with Cardano's uniqueness in mind.  Nobody will care about that uniqueness if our developer community only supports what our entrenched competitors are doing.

CIP-transaction-inputs-as-unordered-set/README.md
```diff
@@ -0,0 +1,140 @@
...
+```
+0 : set<transaction_input>    ; inputs
+```
+```
+18 : set<transaction_input> ; reference inputs; New
+```
+
+The proposed solution suggests modifying the inputs to an unordered set format:
+```
+0 : unordered_set<transaction_input>    ; inputs
```
lehins, line 108: See my earlier comment
```suggestion
0 : oset<transaction_input>    ; inputs
```

CIP-transaction-inputs-as-unordered-set/README.md
```diff
@@ -0,0 +1,140 @@
...
+```
+```
+18 : unordered_set<transaction_input> ; reference inputs; New
```
lehins, line 111: Is there a need to preserve the order of reference inputs? Reference inputs aren't identified by an index anywhere, at least from what I know. For example there is no script purpose for reference inputs, like it is for regular inputs. So, I personally don't see a need to switch them to an ordered set.

That being said, I might be missing some use case in Plutus scripts. There is not too much down side in ledger to changing them to ordered set, therefore I would not oppose changing it either.
```suggestion
18 : set<transaction_input> ; reference inputs; New
```

CIP-transaction-inputs-as-unordered-set/README.md
```diff
@@ -0,0 +1,140 @@
+---
+CIP: xxx
+Title: Transaction Inputs as Unordered Set
```
lehins, line 3: There is a bit of terminology inconsistency IMHO.

Unordered set is usually a set that not only does not preserve the order, but it also doesn't have any well defined order. In the matter of fact CDDL's [`set`](https://github.com/input-output-hk/cbor-sets-spec/blob/master/CBOR_SETS.md#introduction) is such a type: 

> order has no significance

Other examples are a [HashSet](https://hackage.haskell.org/package/unordered-containers-0.2.20/docs/Data-HashSet.html#t:HashSet) is an unordered set in Haskell or [`unordered_set`](https://en.cppreference.com/w/cpp/container/unordered_set) in C++.

What we have already implemented in ledger is ordered sets. Commonly an ordered set in math refers to a set with a specified ordering on elements. However, CDDL does not really have a notion of ordering outside of canonical CBOR, which isn't very useful for Cardano anyways. So, I believe it is safe to refer to an `oset` in CDDL as a Set that preserves the order of elements that was supplied in the serialized form of a transaction, which seems like exactly what you are looking for in this CIP

```suggestion
Title: Transaction Inputs as an Ordered Set
```

CIP-transaction-inputs-as-unordered-set/README.md
```diff
@@ -0,0 +1,140 @@
...
+Category: Plutus
+Status: Proposed
+Authors:
+  - Jonathan Rodriguez <info@anastasialabs.com>
+Implementors: []
+Discussions:
+  - https://github.com/cardano-foundation/CIPs/pull/?
```
lehins, line 10:
```suggestion
  - https://github.com/cardano-foundation/CIPs/pull/758
```

CIP-transaction-inputs-as-unordered-set/README.md
```diff
@@ -0,0 +1,140 @@
...
+
+According to the Babbage CDDL [transaction_body](https://github.com/IntersectMBO/cardano-ledger/blob/master/eras/babbage/impl/cddl-files/babbage.cddl) , the inputs and reference inputs of a transaction body are represented as a set, which indicates the non-duplicate inputs. However, the ledger not only process the [transaction_inputs](https://github.com/IntersectMBO/cardano-ledger/blob/0274cf65dbb79773122b69dfd36a8299eec2783f/eras/babbage/impl/cddl-files/babbage.cddl#L75-L77) as a set , but also orders them lexicographically, first by `transaction_id` and then by `index`.
+
+The primary issue with the ledger ordering inputs lexicographically is that validators must traverse all the inputs or sort them within the validator to find the required inputs for the validation logic. This process can lead to inefficiencies and increases the risk of exceeding the execution budget specially when processing a large set of inputs.
+
+For instance, consider a spending validator that needs to access its own input and output. Each function described below necessitates traversing all inputs and outputs to fulfill such validation:
+
+ ```haskell
+ findOwnInput :: ScriptContext -> Maybe TxInInfo
```
lehins, line 33:
```suggestion
findOwnInput :: ScriptContext -> Maybe TxInInfo
```

CIP-transaction-inputs-as-unordered-set/README.md
```diff
@@ -0,0 +1,140 @@
...
+findOwnInput ScriptContext{scriptContextTxInfo=TxInfo{txInfoInputs}, scriptContextPurpose=Spending txOutRef} =
+    find (\TxInInfo{txInInfoOutRef} -> txInInfoOutRef == txOutRef) txInfoInputs
+findOwnInput _ = Nothing
+ ```
+
+ ```haskell
+getContinuingOutputs :: ScriptContext -> [TxOut]
+getContinuingOutputs ctx | Just TxInInfo{txInInfoResolved=TxOut{txOutAddress}} <- findOwnInput ctx = filter (f txOutAddress) (txInfoOutputs $ scriptContextTxInfo ctx)
```
lehins, line 41: Code snippet line is a bit too long
```suggestion
getContinuingOutputs ctx
    | Just TxInInfo{txInInfoResolved=TxOut{txOutAddress}} <- findOwnInput ctx =
        filter (f txOutAddress) (txInfoOutputs $ scriptContextTxInfo ctx)
```

lehins (commented):
As I mentioned before, I am totally in favor of making this change in the next era. I always believed that current approach of having serialization rely on Haskell ordering of types is error prone. Not even mention all the performance drawbacks described in this CIP.

That being said, there are some outstanding issues with this proposal that we will need to figure out. In particular: how do we deal preserving the old behavior for PlutusV1..PlutusV3, which expect the Haskell ordering for inputs. One way to solve this problem I could foresee is with a new type of script purpose for PlutusV4, that would specify an index, rather than TxIn. We could even combine this feature with the one proposed in [CIP-???? | SpendMany script purpose](https://github.com/cardano-foundation/CIPs/pull/858)

CIP-transaction-inputs-as-unordered-set/README.md
rphair: thanks @lehins for giving voice to my own feelings - @solidsnakedev I was also wondering why the term "unordered" was being used to impose constraints on ordering.

We've decided at the last CIP meeting to assign this a number & traditionally this is when we also confirm the title (since it appears in the repository `README`) so please post ASAP with reasons for any objection you may have, and otherwise I'll change it as recommended above in both the PR & document title.

ch1bo: Maybe seeing it from a programmers perspective rather than from math the term "Unique list" would be less ambiguous? i.e. a list with only unique elements.

fallen-icarus:
> Well, you're making the argument that users want composable scripts, but the question is whether authors will make them.

@WhatisRT I'm making them, and I think this CIP will help. We seem to have a similar philosophy on how to design DApps so there is a good chance mine are close to what you want. I think you should at least take a look at them before you say "it can't realistically be done so we shouldn't even try."

- [DEX](https://github.com/fallen-icarus/cardano-swaps) - a distributed order-book DEX with composable atomic swaps, endogenous liquidity, and a built-in fee market to help prioritize swap executions. It is a fully open ecosystem where anyone can permissionlessly build market makers and aggregator businesses on top.
- [Lending/Borrowing](https://github.com/fallen-icarus/cardano-loans) - a fully p2p lending/borrowing protocol where borrowers and lenders can trustlessly find each other and negotiate all loan terms. It has a built-in trustless credit history, tradable bonds, and naturally incentives periodic loan payments. Lenders even get paid directly.
- [Options Trading](https://github.com/fallen-icarus/cardano-options/tree/v1.0.0.0rc) (the v1.0.0.0rc branch) - a fully p2p options trading protocol where all contract terms are negotiable; you can even decide which asset the premium must be paid in. The contracts are tradable on a secondary market.
- [Aftermarket](https://github.com/fallen-icarus/cardano-aftermarket) - a fully p2p secondary market protocol for the previous two DApps, but is also generalizable enough to be the secondary market for any asset type, including NFTs.

These protocols are fully composable with themselves and all other DApps. They are also a fraction of the cost of contemporary DApps (a single action only costs 0.2 ADA, composing 3-5 actions is still usually less than 1 ADA), and users maintain full custody and delegation control of their assets while using them. I am also building a desktop wallet with a transaction builder that enables users to easily build their own composed transactions. You can think of it like a cart: users shop around for actions they want to take and add them to the builder. When they are done, they go to the builder and just click "Build" and the builder handles everything.

I think this CIP would help all of my projects except for the DEX. The builder in my desktop wallet can decide on the best order of inputs based on what actions the user is trying to compose.

(Apologies to everyone for plugging my own projects, but it seemed entirely relevant.)

CIP-transaction-inputs-as-unordered-set/README.md
rphair: I wish we could have had engagement from @solidsnakedev about the title issue before promoting this to a candidate, but in the meantime I have to do this with a best-guess title that I think is accurate & least likely to be modified later.

Therefore am changing to the title that @lehins recommended, which would be especially important given the corresponding code change recommendation (https://github.com/cardano-foundation/CIPs/pull/758#discussion_r1690173711).  The life cycle of this CIP depends upon Ledger  support so I'd give first priority to whatever title the Ledger team would prefer to assign to the data structures _they_ ultimately maintain.

Obviously some rewrite of the CIP text would then also be required since the term "unordered" is consistently used to mean "not lexicographically reordered" and so we'll leave this thread unresolved until the author & readers completely agree on a way of addressing that ambiguity.

CIP-transaction-inputs-as-unordered-set/README.md
```diff
@@ -0,0 +1,140 @@
+---
+CIP: xxx
```
rphair, line 2:
```suggestion
CIP: 128
```

rphair (commented):
The last CIP meeting was in favour of giving this a CIP number sooner rather than later... as I recall, changes (like this one) requiring a hard fork are generally considered around the time of the previous hard fork so it seems a good time to robustly discuss this idea.

@solidsnakedev one of the things that should be discussed & resolved early on is the question not only of the now oppositely-sensed title but the changes in the text that would be required to resolve this crucial ambiguity (https://github.com/cardano-foundation/CIPs/pull/758#discussion_r1694004635).

Please change the directory name to `CIP-0128` and update the link that points to the rendered of your proposal with the new pathname. 🎉

colll78:
> Well, you're making the argument that users want composable scripts, but the question is whether authors will make them. Writing and auditing a script is very expensive, and script authors might not care much about composability themselves. Also, users might simply not have the choice here: if somebody wants to execute a particular script, and it happens to be non-composable, then they only have the choice of not using it. Maybe they'll complain to the script author, but what's the realistic chance that the author will spend a bunch of extra money and effort on making a composable version? And will it be adopted properly? There are still massive amounts of Plutus V1 transactions being made.
> 
> So I think it's unlikely that market forces are strong enough to ensure composability. Companies really like to make walled gardens for all sorts of things, and lots of people dislike it but the dislike it clearly not strong enough to put pressure on them. I don't see why it would be different here.

Yes the author of this CIP wants composability. This CIP in general actually helps facilitate composability and I don't think that any developers in the ecosystem willing to invest money and development hours to get an application to mainnet and pay for an audit would sacrifice composability (a powerful feature that attracts liquidity and users) to save an **extremely** small amount of ex-units by enforcing that the expected inputs must be at the beginning of the list. If there was any demand to make that trade-off then the DApps that are currently on mainnet today would be doing this except for the outputs. 

The amount of ex-units you save by enforcing all inputs relevant to validation must appear at the beginning of the inputs list is completely negligible. What developers actually care about is that ordering of inputs can be preserved. In practice what you will see is that developers will require that relevant inputs are a continuous sub-list within the tx inputs list, and it will not matter where this sub-list starts because the start of the sub-list will be indexed via the redeemer (this is a design pattern that is already in practice today in nearly every major DApp protocol except it is a thousand times more error prone because the redeemer must contain a list of indices to the inputs which represents each index of each relevant input in the canonical ordering instead of just the start of the contiguous already ordered sub-list). The cost of indexing the start of the relevant inputs is extremely trivial `(dropList indexOfSubListFromRedeemer txInputs)`. If you want composability, you do not have to argue that it is better to linearly search through the list of inputs, linear search is extremely inefficient and unnecessary. Linear search should never be performed in onchain code under any circumstance ever because it completely throws away what is perhaps the biggest advantage that Cardano's smart contract platform has over those in other ecosystems, namely the deterministic script evaluation property. We made huge design sacrifices to obtain this property so to not take advantage of it frankly would be like running a marathon with ankle weights. By taking advantage of this deterministic script evaluation property we never have to perform linear search onchain because we know what the list looks like at the time of transaction construction we can just pass in the index where the item we are looking for should be and fail if it is not there. This way, only assuming transactions are built correctly, it will succeed. Because we can lookup anything in O(1) checks/boolean-conditions by providing the onchain code with the index (via redeemer) to where the element you want to find is supposed to be and just erroring if the indexed element is not indeed what you are looking for. The fact any element can be found onchain without linear-search is an extremely powerful property of our smart contract platform that simply doesn't exist outside our ecosystem.

CIP-transaction-inputs-as-unordered-set/README.md
colll78: Currently transaction outputs are an unordered set in the CDDL, and semantically this means that you can put them in whatever order you want while constructing the transaction and that order will be preserved in the script context. That is why the name unordered set was chosen for this, to be consistent with the naming of the type in the CDDL for tx outputs.

rphair: @colll78 as I was updating the title I also considered suggesting:
> Preserving Order of Transaction Inputs

Would that be accurate & less ambiguous?  If not, could it be modified a bit to explain completely... _in terms of what the CIP provides_, rather than how it relates to internal data structures?  cc @solidsnakedev @lehins

lehins: I totally support this title: "Preserving Order of Transaction Inputs"

It removes the concern and decouples the CIP title from terminology used in the implementation

lehins: > That is why the name unordered set was chosen for this

@colll78 I understand the motivation and reasoning for choosing this name, but the name of unordered set means something totally different in programming. It means a Set with these properties:
* amortized constant time for lookups
* inserts and deletes also have amortized constant time
* order of inserts is not preserved!

In other words Hash Set is a true unordered set, but it is absolutely opposite from what is desired in this CIP.

rphair: OK, I'm changing it to `Preserving Order of Transaction Inputs` since it either addresses or evades all the points made in the conversation above (so resolving this thread).

CIP-transaction-inputs-as-unordered-set/README.md
solidsnakedev: I see this makes sense, I didn't understand the meaning of `oset`which in haskell terms means , `An OSet behaves much like a Set but remembers the order in which the elements were originally inserted.` taken from this hackage link [Oset](https://hackage.haskell.org/package/oset-0.4.0.1/docs/Data-Set-Ordered-OSet.html)

lehins: Yep, except we use our own implementation of `OSet`, since we need to ensure strictness of elements

CIP-transaction-inputs-as-unordered-set/README.md
solidsnakedev: We need reference inputs to be `oset` for the same reason as described in this CIP. 
Also Preserving the order of both inputs and reference inputs allows us to do other things like folding UTXOs that are constructed as a linked list. 
For instance In the case of linked list UTXOs, the folding validation must ensure that the current UTXO correctly points to its successor UTXO. This validation can be easily achieved if we allow both the reference inputs and the inputs to be of `oset` type.

There are protocols already using linked list utxos and applying folding validation, but again it is very expensive just because the order of inputs and reference inputs are not preserved.

solidsnakedev:
@rphair I’ve made the updates based on the feedback. Please let me know if there’s anything else that needs attention.

rphair:
thanks @solidsnakedev - I think all the points have been addressed; we don't have much CIP bandwidth now so I've put it on the agenda again with updated title & assigned CIP number, so hopefully can move this to `Last Check` at that meeting (https://hackmd.io/@cip-editors/94) or in the meantime if editors & Ledger reviewers are in agreement.

cc @Ryun1 @Crypto2099
cc @lehins @WhatisRT
cc @colll78 @fallen-icarus @michele-nuzzi

CIP-0128/README.md
```diff
@@ -0,0 +1,143 @@
...
+
+This approach involves maintaining the current set-based representation of transaction inputs and reference inputs.
+
+## Path to Active
+
+### Acceptance Criteria
+[] Fully implemented in Cardano.
+
+### Implementation Plan
+[] Passes all requirements of both Plutus and Ledger teams as agreed to improve Plutus script efficiency.
```
Ryun1, line 138:
```suggestion
- [ ] Passes all requirements of both Plutus and Ledger teams as agreed to improve Plutus script efficiency.
```

CIP-0128/README.md
```diff
@@ -0,0 +1,143 @@
...
+
+### Alternatives
+#### 1. Retain the existing set-based representation:
+
+This approach involves maintaining the current set-based representation of transaction inputs and reference inputs.
+
+## Path to Active
+
+### Acceptance Criteria
+[] Fully implemented in Cardano.
```
Ryun1, line 135:
```suggestion
- [ ] Included within a Plutus version within a Cardano mainnet hardfork.
```

CIP-0128/README.md
```diff
@@ -0,0 +1,143 @@
+---
+CIP: 128
+Title: Preserving Order of Transaction Inputs
+Category: Plutus
```
rphair, line 4:
```suggestion
Category: Ledger
```
@solidsnakedev as you recall from last hour's meeting that this is similar to other CIPs on the `Ledger` to `Plutus` spectrum which support Plutus implementations but are really implemented in the Ledger.  This category will be especially useful since Ledger has already given positive feedback here and (as I understand it) will be responsible for the implementation.

This goes somewhat against what's currently published in CIP-0084 regarding the [ledger-script interface](https://github.com/cardano-foundation/CIPs/tree/master/CIP-0084#the-ledger-script-interface) so I would welcome @WhatisRT @lehins to post what they feel about both this suggestion & the quotation above (maybe the latter needs to be updated based on 2 years of Plutus, Ledger & CIP process evolution).

rphair (commented):
Pending categorisation as per https://github.com/cardano-foundation/CIPs/pull/758#discussion_r1705913240 the CIP meeting has decided this should be `Last Check` for the next meeting (https://hackmd.io/@cip-editors/95): since the construction & validity of the proposal itself were considered satisfactory.  I'll ✅ this as soon as the pending conversations are resolved.

CIP-0128/README.md
lehins: This is definitely more relevant to Ledger, since almost all of the changes will be on the Ledger side and affect Plutus only slightly.

CIP-transaction-inputs-as-list/README.md
rphair: @solidsnakedev @Crypto2099 I'm resolving this since we have an entirely new chain of growth now... thanks to review & response... to produce a more solid candidate distinct from the original candidate https://github.com/cardano-foundation/CIPs/pull/231.

CIP-transaction-inputs-as-list/README.md
rphair: Resolved in favour of `Ledger` definitively as of https://github.com/cardano-foundation/CIPs/pull/758#discussion_r1710577368.

CIP-transaction-inputs-as-list/README.md
rphair: Resolving, since original unclear statements is gone & further clarification provided in general.

CIP-transaction-inputs-as-list/README.md
rphair: Resolving; there's now a bit of text that appears to explain this expression.

CIP-transaction-inputs-as-unordered-set/README.md
rphair: Original statement is gone so this appears to be resolved in the latest round of commits.

rphair (approved):
Just went back and checked that all previously raised issues have been resolved... mainly the confirmation of the `Ledger` category.  Looking forward to seeing this merged at next CIP meeting unless there are any further reservations.

Ryun1 (approved):

PR #742, opened 2024-01-10
CPS-0011? | Universal JSON Encoding for Domain Types

klntsky:
Cardano domain types have canonical CDDL definitions (for every era), but when it comes to use in web apps, where JSON is the universally accepted format, there is no definite standard.

As a result, software solutions are incompatible with each other, and dApp developers are forced to write code for conversions that could in princible be unnecessary, because the semantics of different JSON layouts are often the same. In particular, this problem is very real when offchain libraries need to provide support for different query layer providers (examples: [Lucid](https://lucid.spacebudz.io/docs/getting-started/choose-provider/), [Mesh.js](https://meshjs.dev/providers), [cardano-transaction-lib](https://github.com/Plutonomicon/cardano-transaction-lib/blob/develop/doc/runtime.md), [Atlas](https://haddock.atlas-app.io/GeniusYield-Providers.html)).

The [initiative](https://github.com/cardano-foundation/CIPs/pull/625) to standardize query layers on Cardano is currently blocked due to absence of a standardized JSON data schema. However, such a schema would be useful in contexts other than query layers, which is the reason why this CPS is separate.

([proposal rendered from branch](https://github.com/klntsky/CIPs/blob/klntsky/json-spec/CPS-0011/README.md))

PR #766, opened 2024-02-21
CIP-0116? | Universal JSON Encoding for Domain Types

klntsky:
This is a CIP corresponding to [CPS-11](https://github.com/cardano-foundation/CIPs/pull/742).

It includes a [json-schema file](https://github.com/klntsky/CIPs/blob/klntsky/json-spec-cip/CIP-XXXX/cardano-babbage.json) specifying JSON layout for Babbage-era Cardano domain types.

The markdown file contains motivation for encoding choices and proposed conventions for building schemas for future eras.

([Rendered version](https://github.com/klntsky/CIPs/blob/klntsky/json-spec-cip/CIP-0116/README.md))


CIP-XXXX/README.md
```diff
@@ -0,0 +1,181 @@
+---
+CIP: ?
+Title: Standard JSON encoding for Domain Types
+Category: Tools
+Status: Proposed
+Authors:
+    - Vladimir Kalnitsky <klntsky@gmail.com>
+Implementors: []
+Discussions:
+    - https://github.com/cardano-foundation/cips/pulls/?
```
Ryun1, line 10:
```suggestion
    - https://github.com/cardano-foundation/cips/pulls/766
```
small house keeping

CIP-XXXX/README.md
```diff
@@ -0,0 +1,181 @@
...
+
+## Abstract
+
+Canonical JSON encoding for Cardano domain types lets the ecosystem converge on a single way of serializing data to JSON, thus freeing the developers from repeating roughly the same, but slightly different encoding/decoding logic over and over.
+
+## Motivation: why is this CIP necessary?
+
+Cardano domain types have canonical CDDL definitions (for every era), but when it comes to use in web apps, where JSON is the universally accepted format, there is no definite standard. This CIP aims to change that.
+
+The full motivation text is provided in [CPS-11](https://github.com/cardano-foundation/CIPs/pull/742).
```
Ryun1, line 23:
```suggestion
The full motivation text is provided in [CPS-11 | Universal JSON Encoding for Domain Types](https://github.com/cardano-foundation/CIPs/pull/742).
```
maybe a little nicer for readability

CIP-XXXX/README.md
```diff
@@ -0,0 +1,181 @@
...
+
+## Specification
+
+This CIP is expected to contain multiple schema definitions for Cardano Eras starting from Babbage.
```
Ryun1, line 27: Why dont we allow new schemas to be added to this set via new CIPs? 
in a similar to the approach to CIP30 extensions and the CDDL proposals

CIP-XXXX/README.md
```diff
@@ -0,0 +1,181 @@
...
+
+Some non-standard `format` types are used:
+
+- `hex`
+- `bech32`
+- `base58`
+
+TODO: describe the formats
+
+## Limitations
```
Ryun1, line 158:
```suggestion
### Limitations
```
section heading fix

CIP-XXXX/README.md
```diff
@@ -0,0 +1,181 @@
...
+
+### Byte length limits for strings
```
Ryun1, line 160:
```suggestion
#### Byte length limits for strings
```

CIP-XXXX/README.md
```diff
@@ -0,0 +1,181 @@
...
+- [ ] Provide an implementation of validating functions that uses this json-schema
+- [ ] Collect a list of cardano domain types implementations and negotiate transition to the specified formats with maintainers (if it makes sense and is possible)
+
+### Acceptance Criteria
+
+### Implementation Plan
+<!-- A plan to meet those criteria. Or `N/A` if not applicable. -->
+
+## Copyright
+<!-- The CIP must be explicitly licensed under acceptable copyright terms. -->
```
Ryun1, line 178:
```suggestion

This CIP is licensed under [CC-BY-4.0](https://creativecommons.org/licenses/by/4.0/legalcode).
```
add in copyright

CIP-XXXX/README.md
```diff
@@ -0,0 +1,181 @@
...
+## Path to Active
+
+- [ ] Complete the specification
+- [ ] Provide an implementation of validating functions that uses this json-schema
+- [ ] Collect a list of cardano domain types implementations and negotiate transition to the specified formats with maintainers (if it makes sense and is possible)
+
+### Acceptance Criteria
+
+### Implementation Plan
+<!-- A plan to meet those criteria. Or `N/A` if not applicable. -->
```
Ryun1, line 175:
```suggestion

- [ ] Complete the specification
- [ ] Provide an implementation of validating functions that uses this json-schema
- [ ] Collect a list of cardano domain types implementations and negotiate transition to the specified formats with maintainers (if it makes sense and is possible)
```
reorganise

CIP-XXXX/README.md
```diff
@@ -0,0 +1,181 @@
...
+## Rationale: how does this CIP achieve its goals?
+
+## Path to Active
+
+- [ ] Complete the specification
+- [ ] Provide an implementation of validating functions that uses this json-schema
+- [ ] Collect a list of cardano domain types implementations and negotiate transition to the specified formats with maintainers (if it makes sense and is possible)
+
+### Acceptance Criteria
+
```
Ryun1, line 173:
```suggestion

- [ ] Completed specifications for: Byron, Shelly, Allegra, Mary, Alonzo, Babbage and Conway.
- [ ] Have one future ledger era supported via subsequent CIP.
```

Ryun1 (commented):
A really great proposal 💪

See, my mainly housekeeping comments.

CIP-XXXX/README.md
klntsky: Because there is no point in keeping multiple CIPs for the same thing for different eras. From the data layout perspective, eras are mostly the same, so the jsons will be almost the same.

Once we "set in stone" a schema for a given era, we are not going to update it, because of course the data layout will stay the same during an era.

In other words, making a CIP extension is pointless because there is nothing to extend, only additions are possible.

CIP-XXXX/README.md
klntsky: Specifications for past eras are not needed: no one is going to deploy new software for these eras. There is nowhere to deploy, even.

I decided to start with babbage, because the cddl spec for Conway does not look finalized (yet).

https://github.com/cardano-foundation/CIPs/pull/742/files#diff-d1d319150c1890f55b035a33282c3b890d814360fff9778f6dd5208b7eb9ecb0R45

micahkendall:
Awesome

CIP-XXXX/cardano-babbage.json
```diff
@@ -0,0 +1,1499 @@
...
+      "title": "32-bit unsigned integer",
+      "type": "integer",
+      "minimum": 0,
+      "maximum": 4294967295
+    },
+    "RewardAddress": {
+      "title": "RewardAddress",
+      "type": "string",
+      "format": "bech32",
+      "pattern": "^(addr1|addr_test1)[02-9ac-hj-np-z]*$"
```
Ryun1, line 46:
```suggestion
      "pattern": "^(stake1|stake_test1)[02-9ac-hj-np-z]*$"
```

CIP-XXXX/README.md
Ryun1: Agreed!

CIP-XXXX/cardano-babbage.json
```diff
@@ -0,0 +1,1826 @@
...
+              "enum": [
+                "map"
+              ]
+            },
+            "contents": {
+              "type": "array",
+              "items": {
+                "type": "object",
+                "properties": {
+                  "key": {
```
rooooooooob, line 464: Is this JSON format taken from somewhere or just here? It would be good for things that have JSON formats used by other tools like tx metadatums, plutus datums etc to stick to existing formats if possible.

klntsky: There is a plan to synchronize CSL with the schema, https://github.com/Emurgo/cardano-serialization-lib/issues/658

rooooooooob: cardano-node has a schema for metadatums and plutus datums. CSL and CML both support it. Note that for CSL it also has a slightly different auto-generated format if you use the auto-generated to_json() instead of the dedicated ones that let you use the cardano-node schemas.

overall I would recommend sticking as close as possible to be 1-to-1 with the .cddl binary spec to make it easier for tools in general to support it

micahkendall: CIP57 too

rooooooooob: not sure if anyone uses them much but there are also cardano-node and cardano-cli formats for native scripts.

it would be good to use whatever is the most common formats for anything that has some kind of format already established, and then for the remainder to make it as close to 1-to-1 to the binary .CDDL format since thatś as close as we have to a standard format in general for things and CBOR and JSON can generally be converted directly as long as you're not using the features of CBOR not supported by JSON (complex keys e.g. in plutus/metadatums) or that aren't necessary like cbor tags/etc. Although the group choices e.g. certificates will need more thought on how to discriminate those

rooooooooob: with obvious considerations for making things more human readable e.g. hex string or bech32 for addresses makes much more sense than the raw bytes of the CBOR format

klntsky: @rooooooooob see this issue in cardano-cli for more context: https://github.com/IntersectMBO/cardano-cli/issues/570

rphair:
@klntsky it would be good if you can come to the next CIP meeting (https://hackmd.io/@cip-editors/83) where I've added this for Triage on the agenda & we can digest the current feedback there & probably assign a number.

CIP-XXXX/cardano-babbage.json
```diff
@@ -0,0 +1,1826 @@
...
+          "title": "TimelockStart",
+          "type": "object",
+          "properties": {
+            "tag": {
+              "enum": [
+                "timelock_start"
+              ]
+            },
+            "slot": {
+              "$ref": "cardano-babbage.json#/definitions/UInt64"
```
lisicky, line 307: Even in conway cddl a slot is uint, and uint type has no upper bound limitation. Probably you need to define unit type for this case

lisicky:
Might it better to prepare schemas for Conway cddl ? Because it might that the proposed CIP will be approved when Babbage is will be already outdated

CIP-XXXX/README.md
```diff
@@ -0,0 +1,250 @@
+---
+CIP: ?
```
```suggestion
CIP: 116
```

rphair (commented):
@klntsky well done on a popular proposal with great anticipated utility.  I'm ready to approve this soon in confirmation of further general technical review, though I'm not qualified to technically review it myself.

Please also rename the directory to `CIP-0116` and update the link to the current proposal in your original comment for this PR thread. 🎉

CIP-XXXX/README.md
```diff
@@ -0,0 +1,250 @@
...
+            "$ref": "cardano-babbage.json#/definitions/NativeScript"
+          }
+        },
+        "plutus_scripts": {
+          "type": "array",
+          "items": {
+            "$ref": "cardano-babbage.json#/definitions/PlutusScript"
+          }
+        }
+      },
```
rphair, line 228:
```suggestion
      },
    },
```
BTW I think you are missing a closing curly brace here.

Ryun1: this is probably what is breaking how github renders it, in the files view

CIP-XXXX/cardano-babbage.json
```diff
@@ -0,0 +1,1826 @@
...
+      "minLength": 175
+    },
+    "CostModels": {
+      "title": "CostModels",
+      "type": "object",
+      "properties": {
+        "plutus_v1": {
+          "$ref": "cardano-babbage.json#/definitions/PlutusV1CostModel"
+        },
+        "plutus_v1": {
```
lisicky, line 575: I guess it should be `plutus_v2`

CIP-XXXX/README.md
rphair: let's test it to be sure (committing) 🤓

CIP-XXXX/cardano-babbage.json
stevenj: cddl uint is 64 bits.  
https://datatracker.ietf.org/doc/html/rfc8610#section-3.3
`"uint"  An unsigned integer (major type 0).`
https://www.rfc-editor.org/rfc/rfc8949.html#section-3.1
`Major type 0:  An unsigned integer in the range 0..2^64-1 inclusive.`

lisicky: Oh sorry totally mixed up it, sorry for that. Thanks @stevenj !

CIP-XXXX/cardano-babbage.json
```diff
@@ -0,0 +1,1826 @@
...
+          "type": "object",
+          "properties": {
+            "tag": {
+              "type": "string",
+              "enum": [
+                "integer"
+              ]
+            },
+            "value": {
+              "$ref": "cardano-babbage.json#/definitions/Int128"
```
lisicky, line 992: My previous comment about slot and unit was wrong. But anyway plutus_data has datatype differ than int128. In the [cddl](https://github.com/IntersectMBO/cardano-ledger/blob/6d860294304749a767d99db92fab42f7c18d9af4/eras/babbage/impl/cddl-files/babbage.cddl#L308-L317) it can be a  bignum that is defined here https://datatracker.ietf.org/doc/html/rfc7049#section-2.4.2

klntsky: > In the [cddl](https://github.com/IntersectMBO/cardano-ledger/blob/6d860294304749a767d99db92fab42f7c18d9af4/eras/babbage/impl/cddl-files/babbage.cddl#L308-L317) it can be a bignum

Hmmm, 

```
big_int = int / big_uint / big_nint
big_uint = #6.2(bounded_bytes)
big_nint = #6.3(bounded_bytes)
```

```
bounded_bytes = bytes .size (0..64)
  ; the real bounded_bytes does not have this limit. it instead has a different
  ; limit which cannot be expressed in CDDL.
  ; The limit is as follows:
  ;  - bytes with a definite-length encoding are limited to size 0..64
  ;  - for bytes with an indefinite-length CBOR encoding, each chunk is
  ;    limited to size 0..64
  ;  ( reminder: in CBOR, the indefinite-length encoding of bytestrings
  ;    consists of a token #2.31 followed by a sequence of definite-length
  ;    encoded bytestrings and a stop code )
```

So, effectively we have a `BigInt` here (signed, long integer, arbitrary length). Thanks for catching this!

CIP-XXXX/cardano-babbage.json
klntsky: Thanks!

CIP-XXXX/cardano-babbage.json
```diff
@@ -0,0 +1,1826 @@
+{
+  "$schema": "https://json-schema.org/draft/2020-12/schema",
+  "$id": "cardano-babbage.json",
```
mkazlauskas, line 3: It would be nice to have a property for version/hardfork of the schema as a property in the object

CIP-XXXX/cardano-babbage.json
```diff
@@ -0,0 +1,1826 @@
...
+            "tag"
+          ]
+        }
+      ]
+    },
+    "VRFKeyHash": {
+      "title": "VRFKeyHash",
+      "type": "string",
+      "description": "blake2b_256 digest of a VRF verification key, encoded as bech32",
+      "type": "string",
```
lisicky, line 1121: Here two fields with the same name

CIP-XXXX/cardano-babbage.json
```diff
@@ -0,0 +1,1826 @@
...
+      "enum": [
+        "plutus_v1",
+        "plutus_v2"
+      ]
+    },
+    "Mint": {
+      "title": "Mint",
+      "description": "Minting or burning of assets. A mapping from policy IDs (script hashes) to mappings from asset names to amounts (that can be negative)",
+      "type": "object",
+      "patternProperties": {
```
lisicky, line 1432: CBOR map allows duplicate keys and in case of mint it is also valid logic. But I'm not sure that this json schema definition allows it for mint

klntsky: @lisicky 

It looks like there are multiple map-like formats:

https://www.ietf.org/archive/id/draft-bormann-cbor-cddl-map-like-data-01.html#section-2

what format in particular is used for cddl's `{ * key => value }` ?

CIP-XXXX/README.md
```diff
@@ -0,0 +1,250 @@
...
+
+## Motivation: why is this CIP necessary?
+
+Cardano domain types have canonical CDDL definitions (for every era), but when it comes to use in web apps, where JSON is the universally accepted format, there is no definite standard. This CIP aims to change that.
+
+The full motivation text is provided in [CPS-11 | Universal JSON Encoding for Domain Types](https://github.com/cardano-foundation/CIPs/pull/742).
+
+## Specification
+
+This CIP is expected to contain multiple schema definitions for Cardano Eras starting from Babbage.
```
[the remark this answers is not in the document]
klntsky: Updated the wording

CIP-XXXX/cardano-babbage.json
mkazlauskas: I believe Haskell ledger/node implementation wouldn't coalesce the values when there are duplicate keys. If that's true, then there's probably no need for json schema to support it either, even if it's possible in CBOR?

lisicky: Example of tx with duplicated keys for mint https://preprod.cardanoscan.io/transaction/3f79ac84ce62fb0223176905c757594130177c21963ab1964aa5d4def04a861e

lisicky: Okay, let me explain smth about this tx. Seems cardanoscan doesn't show it explicitly. The tx body has a mint field that have map with duplicated keys in the cbor it should be smith like that `{ mint : policy_id_1 ->  { asset_name_1 -> amount } , policy_id_1 -> { asset_name_2 -> amount } } ` instead  `{ mint : policy_id_1 ->  { asset_name_1 -> amount, asset_name_2 -> amount  } }`. I just wanted to demonstrate that it is possible from protocol and node side to have duplicated keys in a map. You can also observe that the tx has 2 mint redeemers with different indexes but they are execute the same script.
Also cardano node should not consolidate maps onchain because it can break signatures and also indexes for redeemers ( if we are talking about minting)

lisicky: Anyway you can get cbor of the tx from chain and see it by yourself

lisicky: @klntsky `{ * key => value }` - means map with zero or more elements

lisicky: block hex, it should be the second tx ( you can look inside it by this [tool](https://cardananium.github.io/cquisitor/) ):
`85828a1a0018ba521a02ae24895820413ca2266bac4f438bb66be88bc805ca5d1f4d2a8c3bdeb583c42df7474e14b958200339649fecc4224156494435a46d783d72c8b47fe54e411e6b8fdadda3a7585058203997f8cde4ebadc8993fb0e73c88a7058ec089a6311cc1f930388b9114fc3c99825840180f8b92442933ddb1d6e7a51c2e446fec230c72ff2924df1211cd698212534b488c2b1af6d12109fd2880aa43bbdd92cd0c12f08e50826abb80de13d44b075d5850da506b142cc8d33ae6d2aecfdb327622e9721e49417679f1f05f5229a0924577b80bb0b457fd99fbabd117e758bb49edf17a9a0a52b3a40b923e0faa369e687125787b70361442be3d7b91cebb53da0c19093c58200e3d4834a4dd7a556937b6c1b36a9d766cfff56b9ca28431fe4c7271008fa473845820d35f6405c9d73869b4a9ee42ac578af5d32faa1d79b4013bc620b97a61a8b7da0419014358405c6face300da458ca9453363480a961522daa402e852ba808f2ca9360ffe81e16bc4a8619e0872aeb2b6108b0809c6a0c7b3333162a0c33ffcbde8db6dad440f8208005901c0db84609ce526a38235eea695e2ed42aa2dafc739b23d395edc8e65c33553f987386182dc2a02d8a098e0e688ed9361c036704d16781489e6adca39989b436a035f17db62b490762330be236cf92e5a1bf20a9b671d062b4b38f2581498cd75fbfc4f24e21239468edbea69397a03ce91c1682c08975fc92a4873370722d6fe50876ffb1bf5c2b541cee23ac6e716788630c00d94335589dfd6161efaaa67e56dbe773acbe8c8c8e63f3ad4490342670aac252efed07b51fd082d80c78c4acf07bcc60377ae9ec07b705131841c99e5eb297355b9196d9b63aa1737dd9fdbf7a9282251a6f2e8e157aa141640fd347e18194788e13496abe4b7b2339bdad4ce5c12c25effc103e9ea2b91adf81ef9b6efd23dfe1a86f9b8d594ec536c277876365cd43984ac6ebb89df693036fc01b70cec8f3d64ca49577dc924ab08c71ef3869916e8e934b81b95c427a9d04733707887064620f05869a556d043658324654684d6e8a3a439b5e50a9cac1684ca857fe73cc12fb9ab5ac91a4b12345152344cb5edf6d226a581bb29fea7947ee7b5f1639b2a3bd5032ba2a15308c10d3d3fd75516f7b3f3ed4e24a1c9985c934984cbf867f1c3787accd5034ffec5b26f3abf83aa008282582055ba023aaccefe6888628a8513805ddad2a65f8278ca5c3397a80da6c7890b66018258208d9181649b3671475686851153eadbb8c0b90103d0c10c2ac6a9b1973d4abf49000d818258207aa9cd4999a21c3951de9e330e0ce1fba5943d79ef5bfd90628c2d41e982becd001281825820c59678b6892ba0fbeeaaec22d4cbde17026ff614ed47cea02c47752e5853ebc8010182a200581d600a11b0c7e25dc5d9c63171bdf39d9741b901dc903e12b4e162348e07011a018d55b0a2005839007ea04a6b4942fc0db5f81a86847d2545c3810717ca95f546901cd08cfffa1d082a6132fd1789d48fd348fae9e88d5521c20fb8e1045b0048011a001e848010a200581d600a11b0c7e25dc5d9c63171bdf39d9741b901dc903e12b4e162348e07011a0054cc87111a000a6361021a0006eceb031a02b574ff081a02ae246b0b58204e15e4b830f4f5f029a4779f51735b5dd8462c23a85319166ada33c56ea2ab5aa600818258201d3ae3ffec2d9228c7d5bb9870f5cff032d0f31c83c22c4bd7c5e3ffead4470c0001818258390079467c69a9ac66280174d09d62575ba955748b21dec3b483a9469a65cc339a35f9e0fe039cf510c761d4dd29040c48e9657fdac7e9c01d94821b0000000253e00ccca1581cf923673e7bc027eba929a9086b7b04a126a18ef524b41db6fb014a26a24a414c414c415f54455354024c4741474147415f544553543202021a0002c20e09a2581cf923673e7bc027eba929a9086b7b04a126a18ef524b41db6fb014a26a14a414c414c415f5445535401581cf923673e7bc027eba929a9086b7b04a126a18ef524b41db6fb014a26a14c4741474147415f5445535432010b5820e4af6dc836dada8e84d698b26252f8586f4261ddc6090fc22eac4a059bfe6b040d8382582078869651d619a1832d90b7c843f6d674707fc4c1c2a24e205145656aba7533f600825820c111c9d57fb9ee06f54f0f8bc035f9762db24eada643509afbdbbae98b3ae8f800825820ceff84520384d2f4cdf2017bab6b86a7b46b81ef8e40c3d4f08b699a8a27bade00ab00838258201388fc6611c1aea7ddb4071ce03ee041c7ae2a3b809f730c99ea5d737cb9836b03825820a4d5f23d70af6e494237da29ebd3d0c13f8fde2c3c3b169980a4b63417753a7700825820fd12b648ce816a3aee8f0f0235303958b57bccf3929c88b1f99d6605f1c648ca020183a300581d70f5542fa9d9c61b4dbccd71654cc5a72a9fddc9d563f503682e772ec901821a001e8480a1581c5e4a2431a465a00dc5d8181aaff63959bb235d97013e7acb50b55bc4a1484e6f64654665656401028201d818583cd87a9fd8799f581c4ad1571e7df63d4d6c49240c8372eb639f57c0ef669338c0d752f29bd8799fd8799f1a0025ebd11b0000018bf6a73cbdffffffff82581d604ad1571e7df63d4d6c49240c8372eb639f57c0ef669338c0d752f29b1a0089544082581d604ad1571e7df63d4d6c49240c8372eb639f57c0ef669338c0d752f29b1a005ed7c4021a00044e78031a02ae24e3081a02ae246b0b58205f658fa6994cd9157a48835760a4421da6fec59400f666b343e114283f454eea0d81825820a4d5f23d70af6e494237da29ebd3d0c13f8fde2c3c3b169980a4b63417753a77020e81581c4ad1571e7df63d4d6c49240c8372eb639f57c0ef669338c0d752f29b1082581d604ad1571e7df63d4d6c49240c8372eb639f57c0ef669338c0d752f29b1a004df682111a00370c3f12818258204b61acedca6426ec989ef78a85ed97a8b4ffd47d9c40d4155e453d7e6b25ae580083a30081825820669ed15b1bc5e97ec45af8951e9cbcbd33a3b5878943704d054a1a3ec46be28258407861d1a46731116eb7b140c1a9efbd19e2727f472d4e3503ff96d3125fcc3e11982320322e99fafb61d7016443ac815d909e573bd721efbbeba0cd7a78a11f020481d8799fd8799f40ffd8799fa1d8799fd8799fd87980d8799fd8799f581c7ea04a6b4942fc0db5f81a86847d2545c3810717ca95f546901cd08cffd8799fd8799fd8799f581cfffa1d082a6132fd1789d48fd348fae9e88d5521c20fb8e1045b0048ffffffffffd8799f4040ffff1a001e8480a0a000ffd87c9f9fd8799fd87a9fd8799f5545766572797468696e6720697320616c7269676874d8799fd87980d8799fd8799f581c7ea04a6b4942fc0db5f81a86847d2545c3810717ca95f546901cd08cffd8799fd8799fd8799f581cfffa1d082a6132fd1789d48fd348fae9e88d5521c20fb8e1045b0048ffffffffffff9fd8799f0000ffffffd87980ffd8799fd87a9fd8799f4e5265706f72742070726f626c656dd8799fd87980d8799fd8799f581c7ea04a6b4942fc0db5f81a86847d2545c3810717ca95f546901cd08cffd8799fd8799fd8799f581cfffa1d082a6132fd1789d48fd348fae9e88d5521c20fb8e1045b0048ffffffffffff9fd8799f0101ffffffd87980ffd8799fd87a9fd8799f5243686f696365206265747765656e20312d33d8799fd87980d8799fd8799f581c7ea04a6b4942fc0db5f81a86847d2545c3810717ca95f546901cd08cffd8799fd8799fd8799f581cfffa1d082a6132fd1789d48fd348fae9e88d5521c20fb8e1045b0048ffffffffffff9fd8799f0103ffffffd87980ffd8799fd87a9fd8799f5243686f696365206265747765656e20312d34d8799fd87980d8799fd8799f581c7ea04a6b4942fc0db5f81a86847d2545c3810717ca95f546901cd08cffd8799fd8799fd8799f581cfffa1d082a6132fd1789d48fd348fae9e88d5521c20fb8e1045b0048ffffffffffff9fd8799f0104ffffffd87980ffff1b0000018bf65a9496d87980ffff058184000080821a002f80a61a339742d1a30081825820489ef28ea97f719ee7768645fc74b811c271e5d7ef06c2310854db30158e945d584023dcbaf18dff2ed33a436cbc9036c70deacb573d810df73c41c72c0ebc6183f3d5b44d24d9689db9d3254f3a6720f71175bc1ac32607758276c2cf64ca7dbb0403815251010000322253330034a229309b2b2b9a0105828401000182190bb91a000c79f88401000282190bb91a000c79f8a200818258204b5d869e21e1d1d6f84fdc1e5811eda837ca9fdf2068c69255cd43f62e2711f25840888846f463da49f6f4fbbe4fbfc056901521852d203f1b8bf1a54af2c4efd1cf292ef983a8dbfd3b9266194025028ea59ce34453ddc6a4e0861ddc5db6ee01050581840001d87980821a0012c2f41a15f4bff0a080`

lisicky: Also small clarification, from protocol side it is possible and we can see it onchain but a cardano node just will use the last one key if duplicates are presented.

mkazlauskas: Thanks for clarification @lisicky. I still don't think JSON schema needs to support this, because as CIP already says
```
Every transaction (i.e. CBOR-encoded binary) must have exactly one valid JSON encoding, up to entry ordering in mappings (that are represented as key-value pairs).

For a single JSON fixture, however, there are multiple variants of encoding it as CBOR.
```
If we wanted a JSON schema that has only 1 CBOR representation, then it would be a lot more complicated and not as convenient to use.

As long as it consistently maps CBOR->JSON, I think it's fine

lisicky: @mkazlauskas let me add some details. This duplicated mint fields can affect tx logic, it is not just a way to represent a tx. And  {` mint : policy_id_1 ->  { asset_name_1 -> amount } , policy_id_1 -> { asset_name_2 -> amount } }`  it's not the same as  `{ mint : policy_id_1 ->  { asset_name_1 -> amount, asset_name_2 -> amount  } }`, because original number and order of mint fields is very important for minting itself and also for redeemers indexes. And if you merge two field with the same key you will loose some information that is important for tx logic.

And it's okay to merge duplicated keys or just left only non ignored by node fields if you need to have some tx summary. But if you need to inspect a tx as developer, I guess it's better to see a tx "as is" without any modification or merge.

mkazlauskas: I see, that makes sense. I think it boils down to the intended use cases. Is this loss of information relevant to the use cases of the [CPS](https://github.com/cardano-foundation/CIPs/tree/master/CIP-0011) (I genuinely don't know)?

There are more things in CBOR that are not easily/cleanly representable in JSON (for example, list encoding). So you wouldn't be able to use this JSON object for an api like `(JsonTx) => TransactionWitnessSet`, for example. But in this case it's probably ok.

mkazlauskas: Afterthought: does it matter which way it's encoded as long as you know how? If it doesn't, then as a workaround 'solution', we could specify that json->cbor implementations have to follow some specific rules. This way we could keep the schema clean and push the complexity from dapps to serialization libraries.

klntsky: > Is this loss of information relevant to the use cases of the [CPS](https://github.com/cardano-foundation/CIPs/tree/master/CPS-0011) (I genuinely don't know)?

It could have been. I updated the spec to take duplicate keys in the `Mint` datatype into account. The new naming convention is: if the keys are not supposed to be duplicated, a mapping-like struct should be encoded with `key` and `value` properties for the entries, otherwise the names should be made different. For `Mint`, the naming is ` [ script_hash /\ [ asset_name /\ amount ] ]`

CIP-0116/cardano-babbage.json
```diff
@@ -0,0 +1,1952 @@
...
+      ],
+      "additionalProperties": false
+    },
+    "TransactionHash": {
+      "title": "TransactionHash",
+      "type": "string",
+      "format": "hex",
+      "pattern": "^[0-9a-f]{64}$",
+      "maxLength": 64,
+      "minLength": 64
```
Ryun1, line 210: Why do we specify minLength for transaction hashes but not other things, such as keyHashes or addresses?

CIP-0116/cardano-babbage.json
```diff
@@ -0,0 +1,1952 @@
...
+          "type": "array",
+          "items": {
+            "$ref": "cardano-babbage.json#/definitions/TransactionInput"
+          }
+        },
+        "required_signers": {
+          "title": "Required signers",
+          "type": "array",
+          "items": {
+            "$ref": "cardano-babbage.json#/definitions/Ed25519KeyHash"
```
Ryun1, line 942: shouldnt this be credential rather than keyHash?

CIP-0116/cardano-babbage.json
```diff
@@ -0,0 +1,1952 @@
...
+            },
+            "rewards": {
+              "title": "MIRToStakeCredentials",
+              "description": "Distribution of rewards",
+              "type": "array",
+              "items": {
+                "type": "object",
+                "properties": {
+                  "key": {
+                    "$ref": "cardano-babbage.json#/definitions/Credential"
```
Ryun1, line 1341: shouldnt this be stake account/ reward address?

CIP-0116/cardano-babbage.json
```diff
@@ -0,0 +1,1952 @@
...
+            "tag": {
+              "enum": [
+                "stake_delegation"
+              ]
+            },
+            "credential": {
+              "$ref": "cardano-babbage.json#/definitions/Credential"
+            },
+            "pool_keyhash": {
+              "$ref": "cardano-babbage.json#/definitions/Ed25519KeyHash"
```
Ryun1, line 1443:
```suggestion
              "$ref": "cardano-babbage.json#/definitions/PoolPubKeyHash"
```
maybe?

CIP-0116/cardano-babbage.json
klntsky: It should be a keyhash. It is impossible to sign something with a script.

CIP-0116/README.md
```diff
@@ -0,0 +1,271 @@
+---
+CIP: 116
+Title: Standard JSON encoding for Domain Types
+Category: Tools
+Status: Proposed
+Authors:
+    - Vladimir Kalnitsky <klntsky@gmail.com>
+Implementors: []
```
Ryun1, line 8:
```suggestion
Implementors:
    - Vladimir Kalnitsky <klntsky@gmail.com>
```

Considering https://github.com/Emurgo/cardano-serialization-lib/issues/672

CIP-0116/cardano-babbage.json
klntsky: A reward address is more than a credential because it includes a network tag.

```cddl
move_instantaneous_reward = [ 0 / 1, { * stake_credential => delta_coin } / coin ]

stake_credential =
  [  0, addr_keyhash
  // 1, scripthash
  ]
```

Ryun1: Good catch

Ryun1 (approved):
This is is now a  mature proposal, with lots of community technical review.
The author is now working on the Implementation Plan.
Great work 💪

Pre-meeting approval, ready for CIP editors meeting #88.

rphair (approved):
Likewise I am more than satisfied by the rounds of review & response: marked `Last Check` to keep attention on it until merging when fully ready (re: Implementation Plan).

PR #768, opened 2024-02-27
CIP-0057 | Include Plutus V3

Unisay:
While the next Plutus Ledger Language version is not yet available on mainnet, its conceivable that one might want to produce a Contract  Blueprint for it. 

This PR adds `v3` to the list of allowed `plutusVersion` values.

### DIFF
--- a/CIP-0057/schemas/plutus-blueprint.json
+++ b/CIP-0057/schemas/plutus-blueprint.json
@@ -48,7 +48,7 @@
                 },
                 "plutusVersion": {
                     "type": "string",
-                    "enum": [ "v1", "v2" ]
+                    "enum": [ "v1", "v2", "v3" ]
                 },
                 "license": {
                     "type": "string"


rphair:
@KtorZ @scarmuega @michaelpj how do you feel about the timing of this change?  Would be {useful, misleading) to add this now?  If not now, what should this change coincide with exactly?

Ryun1:
Do the supporting schema's provided with CIP57 also need to also be updated with the new stuff being added via v3? (apologises I am not a plutus expert, so maybe a silly question)

michaelpj:
I think this seems pretty reasonable. I can see the argument for it being premature, but at this point it seems pretty inevitable that there will be a `PlutusV3` ledger language. I guess conceivably we could decide to rename it at the last minute out of spite :joy:

rphair (approved):
OK, in light of https://github.com/cardano-foundation/CIPs/pull/768#issuecomment-1977105002 I'll get behind it now.  If any reservations come up at the meeting then we can apply these to further reviews.  Otherwise if no reservations we could push it ahead at the meeting.

Ryun1 (approved):
Very small change, no danger of breaking backwards compatibility

Issue #769, opened 2024-02-29
CIP-0069: alternative approach


zliu41:
I had a conversation about this CIP several weeks ago with a few others, including @lehins from the ledger team. This seems desirable to do, easy to implement (on both the Plutus and the ledger side), and if we want it in Plutus V3 then it must be done soon.

I'm opening this issue to discuss an alternative approach: the problem with the proposed approach is that it adds a `ScriptArgs` argument, which leads to the "ugly unchecked dependence" problem as pointed out in https://github.com/cardano-foundation/CIPs/pull/321#issuecomment-1250053504. But this doesn't seem necessary: we can simply use two different types for `scriptContextPurpose`, and the key of `txInfoRedeemers`. E.g., Alexey suggested that we can create a `ScriptPurposeWithDatum` type for `scriptContextPurpose`, while keeping the `txInfoRedeemers` key type as it is, which would avoid the problem.

If we can agree and settle on the design soon, it should have a good chance to be done in time for Plutus V3. @KtorZ @michaelpj @zygomeb

michaelpj:
I don't find the unchecked dependence problem to be that bad, but I also don't desperately hate the combined thing. Perhaps we could give it a different name. `ScriptInfo`? IDK.

I was going to say that the unchecked approach would make it easier if we decided to allow the spending of script outputs without datums. But if we did that we would probably need to tie it to a new ledger language anyway, in which case we could change the `ScriptInfo` type just as easily.

MicroProofs:
I made some updates to CIP 69 to mention the no datum case, but looks like the typing for ScriptInfo is better since you don't end up with an impossible type combination of ScriptArgs DatumAndRedeemer with a ScriptPurpose of Mint, Withdraw, etc.

zliu41:
@rphair has this been discussed and what was the outcome?

MicroProofs:
@zliu41 The outdone was to discuss with the Plutus Core team next CIP meeting to get the path forward that everyone is most comfortable implementing. 

It sounds like the approach talked about in this issue is a more wholistic approach for typing validators. 
We can also continue the discussion here in the comments. I can update the pr to reflect any changes decided on and finalized here.

leo42:
I would love to see this implemented. 

It is a prerequisite for smart wallets if we want them to have a smooth and intuitive UX .

I will also be implementing the Plutus-wallet connector soon and hopefully we can get this section of the market going, so having it in Plutus V3 will help everything accelerate . 

[...]

PR #774, opened 2024-03-04
CIP-0069 | single argument Scripts

MicroProofs:
Not sure if we should discuss here or in #769

The goal is to add additional clarity to this CIP or determine if another CIP is better suited to 2 major use cases

1. Having a spend script share the same script hash as a mint, withdrawing, observing, or certifying script without using a argument check workaround currently used in Opshin and Aiken
2. Giving smart contract creators the ability to handle the no datum case with spend contracts.

### DIFF
--- a/CIP-0069/README.md
+++ b/CIP-0069/README.md
@@ -65,11 +65,17 @@ data ScriptArgs =
   | RedeemerAndDatum Redeemer Datum
 ```
 
+In the case of a spend purpose, the ledger would execute the script with RedeemerAndDatum Redeemer Datum if a datum is present in the spending input. Otherwise the ledger would execute the script with RedeemerOnly Redeemer. In all other purposes the ledger would enforce only RedeemerOnly Redeemer is used with the script execution.
+
 ## Rationale: how does this CIP achieve its goals?
 
 Unifying of the script signature is a very elegant solution to the problem, streamlining the experience of developing on cardano.
 Given that accessing the datum is almost always made by a spending script, it makes sense to introduce that argument back to the `ScriptPurpose` that now plays a more important role.
 It begs the question if it should be added as an argument to all validators, to further emphasize that fact.
+
+I'm not sure what the above section is talking about?
+
+This CIP turns all scripts into 2 arg scripts with a variant for the first argument that can be matched on to determine if a datum and redeemer or only redeemer is present.
 
 ## Backwards compatibility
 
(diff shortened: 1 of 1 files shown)


CIP-0069/README.md
```diff
@@ -30,10 +32,24 @@ As it stands the scripts being made on cardano often suffer this problem, and th
...
+
+I see three major use cases when it comes to running spend scripts without datums:
+
+- Use Case 1: A script that acts as a wallet for users. By having no datum spending the user can send directly from exchanges or have friends send to their smart contract wallet with no datum needed.
+
+- Use Case 2: As a DAO treasury. The funds in this script would be controlled by a DAO and anyone can donate/contribute to the DAO without a datum.
+
+- Use Case 3: Allow dApp protocols to have a claim or withdraw mechanism (similar to Ethereum for tokens sent to a contract without call) for claiming tokens sent without a datum.
+
+I'd be remiss if I didn't mention CIP-0112 which has been expanded to improve native script capabilities to provide an alternative solution for use case 1 and 2. But for use case 3, CIP-0112 does not enable a "claim or withdraw mechanism" for contracts.
```
colll78, line 51: +1

CIP-0069/README.md
```diff
@@ -30,10 +32,24 @@ As it stands the scripts being made on cardano often suffer this problem, and th
 
 - Use Case 3 (taken from Minswap's Dex V1): NFT is minted for the same reason as above. It allows a minting policy to later mint LP tokens with that unique id token name.
 
-We see a similar pattern repeating over and over again as witnessed by dapp developers and auditors alike. By allowing the multi-purpose policies (spending and any other) we increase the security of Cardano by giving us more confidence and allowing to design protocols that have their architecture driven by Cardano's features, not limited by Cardano's language.
+We see a similar pattern repeating over and over again as witnessed by dapp developers and auditors alike. By allowing the multi-purpose scripts (spending and any other) we increase the security of Cardano by giving us more confidence and allowing to design protocols that have their architecture driven by Cardano's features, not limited by Cardano's language.
```
colll78, line 35: Soley this reason is enough justification for this CIP. As in the case of the withdraw-zero trick and CIP 112, it is simply not healthy for the most critical and widely used design patterns throughout the ecosystem to rely on what is essentially an obscure implementation detail / black magic trick. Given the importance of this design pattern, and its demonstrable utility (on mainnet across a huge number of protocols in the ecosystem in production) it is of great importance that this is granted more official support and formalization.

colll78: I would, however, remove the part about the "recent Nami bug" since AFAIK CIP are supposed to be timeless and this might lose context in the future. Maybe replace with a brief line about how a bug in nami caused UTxOs without datums to be sent to a script and thus mistakenly locked forever.

PR #779, opened 2024-03-11
CPS-0015? | Intents for Cardano

michaelpj:
We describe a family of issues around “intents” on Cardano. We do some basic analysis on the problem space, and suggest some high-level areas for investigation.

---
[Rendered on author's fork](https://github.com/michaelpj/CIPs/tree/mpj/cps-intents/CPS-0015)


CPS-????/README.md
```diff
@@ -0,0 +1,173 @@
...
+
+## Problem
+
+In many cases, a user’s desired interaction with a system is not fully specified. 
+Classic examples include things like “market orders” where a user wants to buy an asset at the “market price”, and does not specify exactly what that is, leaving it up to another system to resolve. 
+A common name for these kinds of “incomplete” or “partial” actions is “intents”. 
+We refer to the process of turning an intent into something complete as “fulfilment”.
+
+Intents are [quite prevalent](https://members.delphidigital.io/reports/wtf-is-anoma-part-2-wtf-are-intent-based-apps) in blockchain systems (see “Use cases” for more discussion of specific examples), but they are also problematic since blockchain systems usually require a fairly high degree of specificity in their settlement layer. 
+For example, if you want to perform a trade, you usually need to specify exactly who with and exactly how much for.
```
polinavino, line 26: i feel like this isnt really the case for account-based systems, at least

CPS-????/README.md
```diff
@@ -0,0 +1,173 @@
...
+In contrast, on Cardano smart contracts only validate the behaviour that the user has specified in their transaction. 
+This gives us very nice determinism properties... but when resolving intents we actually want non-determinism!
+
+We can see the “concurrency problem” through this lens. 
+Cardano insists that state updates (via datums in UTXOs) are completely explicit and deterministically sequenced, in contrast to systems like Ethereum, where state access is implicit and non-deterministically sequenced. 
+This is quite inconvenient, since what users often want is to interact with “the current state” of an application, and they are happy for this to be determined non-deterministically.[^concurrency]
+
+[^concurrency]: See https://github.com/input-output-hk/Developer-Experience-working-group/issues/47 for discussion of a very intent-like approach to solving the concurrency problem.
+
+To support intent-based interactions, we need two things:
```
polinavino, line 39: we (obviously) also need a language for expressing intents and their fulfillments

CPS-????/README.md
```diff
@@ -0,0 +1,173 @@
...
+[^spam]:  For example, since intents are fundamentally incomplete and cannot be settled by themselves, it is easy to spam a network with unsatisfiable intents. 
+
+A crucial feature that makes intent processing much easier is if intent settlement can be unidirectional: that is, it is not necessary to go back to the author of an intent in order to fulfil it. 
+This makes it much easier to distribute intents in a decentralised manner.
+
+### UX considerations
+
+Intent processing is the part that has the most effect on UX. 
+Users will be interacting directly with an intent-processing system to submit their intents, and they may need various kinds of feedback or to make choices. 
+This is complicated by the fact that some information may instead come from the underlying blockchain, e.g. about settlement.
```
polinavino, line 76: I don't really understand what you are saying in these two sentences above

CPS-????/README.md
```diff
@@ -0,0 +1,173 @@
...
+### Security considerations
+
+Intent processing has additional security issues to deal with.
+
+The first is spam/DoS prevention. 
+Since intents are incomplete (and maybe incompletable), an intent-processing system needs to be robust to being spammed with a large number of incompletable intents.
+
+The second is front-running or more generally the various kinds of activity often discussed as MEV (“miner-extractable value”) in other blockchains. 
+This is an inevitable consequence of non-determinism: since the user has not specified exactly how things happen, the system or the actors within it are free to arrange the eventual fulfilment in a way that is maximally profitable for them. 
+This is a major topic, but one that needs to be addressed (or explicitly allowed) by an intent-processing system.
```
polinavino, line 98: I would mention that because of the determinism inherent to Cardano system design, the opportunity/risk of MEV can be adjusted by the user (by changing the intent expression), and can range from "i wanna buy something for market price at whatever time" to the quite strict current babel fees proposal. Choose your own adventure here!

polinavino: I guess you sketch out this range in usecases below

CPS-????/README.md
```diff
@@ -0,0 +1,173 @@
...
+
+A transaction requiring fee sponsorship does not specify where the money for the fees is coming from. 
+The sponsor can fulfil it by picking it up and providing the money.
+
+In principle intents that need sponsorship can be sent directly to the application operator, which means these kinds of intents may need less sophisticated networking, at the cost of centralisation.
+
+### Non-deterministic access to state
+
+A user who interacts with an application’s state stored in the datum of a UTXO may well be happy to interact with the “current state” whatever that is. 
+An intent of this kind must specify the address of the UTXO it is spending, but not its content, and similarly not specify the state of the output it creates.
```
polinavino, line 140: Maybe better to describe this not as "the address of the UTxO" but more like, transactions with outputs that have a fill-in-able datum

CPS-????/README.md
michaelpj: You still need to say who you're sending money to and how much money you're spending! 

I do think there's an argument that smart contracts on Ethereum effectively support a limited kind of intent-based interaction where you tell the smart contract the action you want to take and then it implements it, potentially non-deterministically.

CPS-????/README.md
michaelpj: Yeah, the more non-deterministic you make it, the more risk you have of the bits you don't specify being set to your disadvantage. But indeed if you have something like an open swap with fixed prices there is really no way to exploit you that isn't fair (the processing system might choose to fulfil or not fulfil your order to its advantage, but "this may or may not happen" is surely a fair expectation if anything is).

CPS-????/README.md
michaelpj: Sometimes the spender provides the datum for the output they are spending, if the output is a hash. So if you want to be insensitive to that, then someone else needs to provide that. So it can apply on both end.s

polinavino: `(a, v, ???) \in outputs tx1` has index 1, lets say
And inputs now can be "resolved" inside the transaction by specifying only the missing data
`(txId tx1, 1) |-> (_, _, myDatum) \in inputs tx2`

Do you mean something like this?

CPS-????/README.md
polinavino: yeah, this is what i mean - systems where you don't know what state your tx will be applied to are already kind of intent-processing systems. The only argument for less on-chain intent processing here is that it will be cheaper/faster.

CPS-????/README.md
polinavino: Mostly I just meant here that the last sentence of this section should connect the theme here to the risk levels showcased by the usecases that follow.

CPS-????/README.md
```diff
@@ -0,0 +1,173 @@
+---
+CPS: ?
+Title: Intents for Cardano
+Status: Open
+Category: ?
```
rphair, line 5:
```suggestion
Category: Tools
```
Since this includes interpretations of new things in the Ledger (not just Ledger changes themselves) and maybe even a language (and therefore tooling) to express these structures (https://github.com/cardano-foundation/CIPs/pull/779#discussion_r1520748340), I would recommend in the next CIP meeting that we apply and/or confirm the `Tools` category.

rphair (commented):
@michaelpj @polinavino I've added this to introduce in the agenda of our CIP meeting in a week's time (https://hackmd.io/@cip-editors/84).  Nearly our entire agenda is Plutus at this point so your attendance would be welcome for a number of reasons. 🙏

CPS-????/README.md
michaelpj: No, the other way around. Normally when you spend an output, you have to either:
1. If the output has a datum hash, specify the pre-image
2. Otherwise do nothing

You might also have to specify a datum or a datum hash for the output that you are creating.

To do "nondeterministic access to state", you need to specify _neither_ datum. So that means either being in case 2, or in case 1 and not providing the pre-image; and also creating an output which doesn't specify the datum.

To make this complete, someone would need to fill in those details. I don't really know how.

CPS-????/README.md
michaelpj: Yeah, and also that your intents don't get "matched", there is a single provider which is the smart contract. I think it's maybe less confusing to not call those "intents" and just say it's non-deterministic transaction processing, but IDK.

polinavino: Saying that, in general, we want intent processing on systems because we want that functionality **added**, rather than that "we want to do it cheaper and faster and easier" is a bit misleading, so i would reword.

vlasin:
Let me point out that implementating intents is already technically possible on Cardano without changes to the Ledger. It can be done with the help of a hypothetical smart contract wallet where all user UTxOs are locked by Plutus scripts.
1. A user can submit an intent to the intent-processing server. The message would also contain a signature on the hash of the user's intent.
2. The server can then construct, sign, and submit the transaction that satisfies the user's intent.
3. The wallet's script is used to validate that the transaction satisfies the user's intent and that the user's intent is signed by the user's private key.

The problem with the above approach, in my opinion, lies mostly in its practicality and efficiency. In particular, the wallet's script needs to run for each of the affected UTxOs even though it verifies the same exact statement. There are a number of other proposals that could deal with the efficiency problem if implemented:
https://github.com/cardano-foundation/CIPs/tree/master/CIP-0069
https://github.com/cardano-foundation/CIPs/blob/arbitrary-script/CIP-XXXX/README.md
https://github.com/fallen-icarus/CIPs/tree/maybe-datum/CIP-%3F%3F%3F%3F

michaelpj:
@vlasin I tried to gesture in that direction in the CPS here: https://github.com/cardano-foundation/CIPs/pull/779/files#diff-5a0dba075d998658d72169818d839f4c63cf105e4d6c3fe81e46b20d5fd3dc8fR45

I should elaborate!

I agree that practical considerations dominate here. Fundamentally, any on-chain solution requires paying the costs necessary for the network to reach consensus on that state, which may be both unnecessary and too costly. We need consensus on the final change, it's not clear that we need consensus on the proposal. Transaction fees and overhead is what made the Stellar DEX not function at large trading volumes - I think this is ultimately unsurprising.

rphair (commented):
@michaelpj thanks for introducing this & related CIP https://github.com/cardano-foundation/CIPs/pull/780 at the meeting today; both agreed all around as candidates (though expecting a long gestation period) so assigning numbers.

Since the scope might be clear at this point I hope we can confirm a category of `Tools` or indicate that it might be narrower (to be determined) and therefore leave the category open... though it will be helpful for review it we can assign an initial category as usual.

Please update the directory name & the link to your proposal in the OP 🙏

AndrewWestberg:
Linking to a previous discussion in a working group on this topic.

https://github.com/input-output-hk/Developer-Experience-working-group/issues/47

polinavino:
Here is an idea about how to introduce a new kind of intent that is fulfilled within a valid validation zone : _the intent to spend a certain output_. This would allow us to implement the following kind of protocol for **Light Clients** that has the service provider (SP) build the transaction for the (LC) rather than supply data for the LC to do it themselves. This comes with a way to ensure that the SP is compensated, but only if they build a valid transaction. 

This requires us to build on top of the validation zones infrastructure (in-progress implementation specifically for the babel fees/swaps usecase [here](https://github.com/IntersectMBO/formal-ledger-specifications/compare/polina-babel?diff=unified&w=#diff-5f67369ed62c0dab01e13a73f072b664ada237d094bbea4582365264dd163bf9) )

This also requires an intents-constructing API (an "intents language"), which can be as simple or fancy as we want. E.g. it can support requesting transactions that
- pay a value from key to key
- mint an NFT
- make an offer/trade on a DEX

**Ledger changes** : 

- the transaction body has two new fields ( i)  `noRefInputs : Map Ix TxOut` (which are the outputs corresponding to the inputs the transaction would be spending if it could look them up in the actual UTxO state) and ( ii) `matchingRefs : Map TxIn TxIn` , which specify the output references in the UTxO that the outputs have to match to, and are indexed by the transaction that is requesting to spend the no-input references 

- the UTxO state is a pair of a UTxO and `Intents : Map TxIn TxOut`

- during the validation of a zone, the `Intents` map is populated with the `noRefInputs` of transactions in the zone (the `TxIn` index is constructed from the `TxId` of the transactions that is adding the entry, and the `Ix` of that entry)

- an entry in `Intents` is removed when another transaction provides the TxIn corresponding to the `TxOut` for the transaction ID and index specified in the entry key, and the full UTxO entry is verified to exist in the UTxO set. That entry is also then removed from the UTxO set.

- once the entire zone is applied, the `Intents` structure must be empty for a zone to be valid

**Off-chain infrastructure** : 

- Intents language 

- support for transaction dissemination

**Protocol** : 

1. LC constructs an intent and sends it to SP

2. SP constructs a transaction tx that satisfies the intent sent to them, and includes in that transaction a payment made from LCs key to the SP

- must include at least 1 input from their own wallet (to satisfy the ledger requirement without revealing too much to the LC)

3. SP removes from tx any output references whose associated outputs contain scripts, or ones that belong to the LCs keys, and instead adds the output data associated to those references to noRefInputs 

4. SP sends tx to LC to sign 

- it must also send the resolved output for any “regular” inputs, and pre-image for script integrity hash

- this is needed for the LC to be able to check the scripts

5. LC checks the transaction to make sure it satisfies its intent, runs the scripts, signs it, and sends it back (if OK)

- note that LC cannot check validity, but in the UTxO model, an invalid transaction cannot be applied, so there is no danger of something bad happening on-chain, and the SP is incentivized to construct only valid transactions in order to get paid

- LC does need to check the scripts, and is able to do that just from the received data

7. SP constructs a transaction that resolves the “missing inputs” by constructing another transaction tx' which provides the required output references in the matchingRefs field

8. SP submits a validation zone with both transactions to the Cardano network


**Possible issues**
- possibility of a spam attack (but not on the Cardano network), i.e. LCs sending multiple unfillable requests to SPs off-chain

this is a problem for any type of request system

could have some sort of key registration system, and require LCs to sign their requests

- not compatible with current versions of Plutus, since TxInfo would be constructed differently (however, probably OK for versions written specifically for this)

- if an intent is submitted to multiple SPs, only the one whose response to the LC gets submitted first (alongside their own transaction fulfilling it) gets paid

we might even have to worry about another SP intercepting, dismantling, and re-constructing the transaction built by an SP, and changing the payment made to one that is to themselves

hopefully, that it’s about as expensive as constructing the transaction from scratch

rphair:
@polinavino - would you the best person to advocate for & update this PR and https://github.com/cardano-foundation/CIPs/pull/780?

1 - Whatever is in the "babel fees/swaps usecase [here](https://github.com/IntersectMBO/formal-ledger-specifications/compare/polina-babel?diff=unified&w=#diff-5f67369ed62c0dab01e13a73f072b664ada237d094bbea4582365264dd163bf9)" is already in the [Use Cases](https://github.com/michaelpj/CIPs/tree/mpj/cps-intents/CPS-0015#use-cases), but this mechanism for Light Clients is new... could you add a section for it at the appropriate level of detail for the CPS?  Also if the Light Clients idea matches some other project under development, I imagine it would be filled out into a CIP at some point.

2 - Also: your earlier suggestion of an Intents language (https://github.com/cardano-foundation/CIPs/pull/779#discussion_r1520748340) was why this CPS was classified as `Tools` (rather than Ledger) (https://github.com/cardano-foundation/CIPs/pull/779#discussion_r1521864652) but there's still no detail in the CPS about what would be expected from this language... would you be willing to contribute this?

Editorially I think this document and proposal https://github.com/cardano-foundation/CIPs/pull/780 need an "owner" and I would be happy to hear that we can rely on you for current & future updates, until these are merged in whatever state is confirmed to be useful for further development.

polinavino:
I will likely be the one to advocate for this, but I will have to work on the CPS and think about your other comments this more next week!

polinavino:
After some discussion with other folks working on light clients, we decided that we will instead attempt to use fancy cryptography to implement a similar protocol (co-constructing a transaction using 2PC) fully off-chain. So , for the time being I will not pursue this.

rphair:
@Ryun1 @Crypto2099 - we have lost the push behind this with MPJ leaving, and I believe with @polinavino committing to a different approach this is effectively `Deprecated` & so I'm marking as such.

@lehins @fallen-icarus - as advocates of varying methods in this space, if you would have any other suggestion than deprecating please let us know.  I would prefer to see this merged as-is... after giving it an explicit `Inactive` status (perhaps `Inactive (lack of advocacy)`) to avoid too much duplication of effort the next time the community manifests some other intent-like transaction paradigm.

If this would be OK all around I'll make the necessary cleanup to this PR and move it from `Deprecated` to `Last check` in time for some future meeting... an unusual workflow but more realistic than the current tag 'Waiting for Author`.

rphair:
Effectively deprecated in favour of currently pursued alternatives already documented above.

PR #784, opened 2024-03-14
CIP-0069 | add Datum to ScriptContext

### LINKED ISSUE #769: CIP-0069: alternative approach

zliu41:
Closes #769 

cc @michaelpj @lehins @MicroProofs @rphair @KtorZ @zygomeb @L-as

### DIFF
--- a/CIP-0069/README.md
+++ b/CIP-0069/README.md
@@ -69,9 +69,13 @@ data ScriptContext
   | Voting TxInfo Redeemer Voter
   | Proposing TxInfo Redeemer Integer ProposalProcedure
 ```
-The Datum is optional, which will enable us to allow the execution of spending scripts without a datum.
+The Datum is optional, which will enable us to allow the execution of spending scripts without a datum. 
+One more change will be needed on the ledger side in order to make the Datum optional for spending scripts.
+The ledger UTXOW rule needs to be relaxed, this ledger rule checks if a utxo has an existing datum if the address's payment credential is a phase 2 validation script.
 
 The ScriptPurpose type is left the same used in the Redeemers Map is left the same.
+
+
 
 ## Rationale: how does this CIP achieve its goals?
 
--- a/CIP-0069/README.md
+++ b/CIP-0069/README.md
@@ -44,14 +44,16 @@ It has the following form: `Datum -> Redeemer -> ScriptContext -> a`. This is en
 We make the following modification:
 
 - The signature of all scripts will be `Redeemer -> ScriptContext -> a`.
-- The `ScriptPurpose` type is modified such that the `Spending` variant contains a `Datum`:
+- The `ScriptPurpose` type is modified such that the `Spending` variant contains an optional `Datum`:
 
 ```hs
 data ScriptPurpose
   = ...
-  | Spending TxOutRef Datum
+  | Spending TxOutRef (Maybe Datum)
     ...
 ```
+
+The `Datum` is optional, which will enable us to allow executing spending scripts without a datum.
 
 - A new data type `ScriptIdentifier` is created.
   It is identical to the original `ScriptPurpose` and is used in the `txInfoRedeemers` map to uniquely identify a script within a transaction.
--- a/CIP-0069/README.md
+++ b/CIP-0069/README.md
@@ -80,7 +80,6 @@ The ScriptPurpose type used in the Redeemers Map is left the same.
 ## Rationale: how does this CIP achieve its goals?
 
 Unifying of the script signature is a very elegant solution to the problem, streamlining the experience of developing on cardano.
-Given that accessing the datum is almost always made by a spending script, it makes sense to introduce that argument back to the `ScriptPurpose` that now plays a more important role.
 It begs the question if it should be added as an argument to all validators, to further emphasize that fact.
 
 
--- a/CIP-0069/README.md
+++ b/CIP-0069/README.md
@@ -41,12 +41,29 @@ This primarily manifests in the ability to use a single validator for both minti
 All the script purposes have a form of `Redeemer -> ScriptContext -> a` except the `Spending` one.
 It has the following form: `Datum -> Redeemer -> ScriptContext -> a`. This is enforced by the Cardano ledger.
 
-We modify the general signature to be `ScriptArgs -> ScriptContext -> a`.
+We make the following modification:
+
+- The signature of all scripts will be `Redeemer -> ScriptContext -> a`.
+- The `ScriptPurpose` type is modified such that the `Spending` variant contains a `Datum`:
 
 ```hs
-data ScriptArgs =
-    RedeemerOnly Redeemer
-  | RedeemerAndDatum Redeemer Datum
+data ScriptPurpose
+  = ...
+  | Spending TxOutRef Datum
+    ...
+```
+
+- A new data type `ScriptIdentifier` is created.
+  It is identical to the original `ScriptPurpose` and is used in the `txInfoRedeemers` map to uniquely identify a script within a transaction.
+
+```hs
+data ScriptIdentifier
+  = MintingScript CurrencySymbol
+  | SpendingScript TxOutRef
+  | RewardingScript Credential
+  | CertifyingScript Integer TxCert
+  | VotingScript Voter
+  | ProposingScript Integer ProposalProcedure
 ```
 
 ## Rationale: how does this CIP achieve its goals?
--- a/CIP-0069/README.md
+++ b/CIP-0069/README.md
@@ -59,15 +59,28 @@ All the script purposes have a form of ```Redeemer -> ScriptContext -> a``` exce
 We propose to make the following modification:
 
 The signature of all scripts will be ```ScriptContext -> a```.
-The ScriptContext type is now a union type with a variant for each script purpose.
+The ScriptInfo type is a union type with a variant for each script purpose.
 ```
-data ScriptContext
-  = SpendingContext TxInfo Redeemer (Maybe Datum) TxOutRef
-  | MintingContext TxInfo Redeemer CurrencySymbol
-  | CertifyingContext TxInfo Redeemer Integer TxCert 
-  | RewardingContext TxInfo Redeemer Credential
-  | VotingContext TxInfo Redeemer Voter
-  | ProposingContext TxInfo Redeemer Integer ProposalProcedure
+data ScriptContext = ScriptContext
+  { scriptContextTxInfo  :: TxInfo
+  , scriptContextScriptInfo :: ScriptInfo
+  }
+
+data ScriptInfo
+  = MintingScript V2.CurrencySymbol Redeemer
+  | SpendingScript V3.TxOutRef (Maybe Datum) Redeemer
+  | RewardingScript V2.Credential Redeemer
+  | CertifyingScript
+      Haskell.Integer
+      -- ^ 0-based index of the given `TxCert` in `txInfoTxCerts`
+      TxCert
+      Redeemer
+  | VotingScript Voter Redeemer
+  | ProposingScript
+      Haskell.Integer
+      -- ^ 0-based index of the given `ProposalProcedure` in `txInfoProposalProcedures`
+      ProposalProcedure
+      Redeemer
 ```
 The Datum is optional, which will enable us to allow the execution of spending scripts without a datum. 
 One more change will be needed on the ledger side in order to make the Datum optional for spending scripts.
--- a/CIP-0069/README.md
+++ b/CIP-0069/README.md
@@ -62,12 +62,12 @@ The signature of all scripts will be ```ScriptContext -> a```.
 The ScriptContext type is now a union type with a variant for each script purpose.
 ```
 data ScriptContext
-  = Minting TxInfo Redeemer CurrencySymbol
-  | Spending TxInfo Redeemer (Maybe Datum) TxOutRef
-  | Rewarding TxInfo Redeemer Credential
-  | Certifying TxInfo Redeemer Integer TxCert 
-  | Voting TxInfo Redeemer Voter
-  | Proposing TxInfo Redeemer Integer ProposalProcedure
+  = SpendingContext TxInfo Redeemer (Maybe Datum) TxOutRef
+  | MintingContext TxInfo Redeemer CurrencySymbol
+  | CertifyingContext TxInfo Redeemer Integer TxCert 
+  | RewardingContext TxInfo Redeemer Credential
+  | VotingContext TxInfo Redeemer Voter
+  | ProposingContext TxInfo Redeemer Integer ProposalProcedure
 ```
 The Datum is optional, which will enable us to allow the execution of spending scripts without a datum. 
 One more change will be needed on the ledger side in order to make the Datum optional for spending scripts.
--- a/CIP-0069/README.md
+++ b/CIP-0069/README.md
@@ -73,7 +73,7 @@ The Datum is optional, which will enable us to allow the execution of spending s
 One more change will be needed on the ledger side in order to make the Datum optional for spending scripts.
 The ledger UTXOW rule needs to be relaxed, this ledger rule checks if a utxo has an existing datum if the address's payment credential is a phase 2 validation script.
 
-The ScriptPurpose type is left the same used in the Redeemers Map is left the same.
+The ScriptPurpose type used in the Redeemers Map is left the same.
 
 
 
--- a/CIP-0069/README.md
+++ b/CIP-0069/README.md
@@ -38,7 +38,7 @@ This primarily manifests in the ability to use a single validator for both minti
 
 ### No Datum Spend Purpose
 
-One of the major footguns of Plutus scripts is if a user sends to the script with a wrong or missing datum. This has happened in the case of the Nami wallet having a bug that caused the wrong address to be chosen. There are other instances of user error where they send from a CEX to a script address. A wrong datum can be handled by the Plutus scripts themselves by having an alternative execution branch if type does not match the expected datum type. But in the case of no datum the script is not run and fails in phase 1 validation. The other motivation of this CIP is to be able to create spend scripts that can handle the no datum case. 
+One of the major footguns of Plutus scripts is if a user sends to the script with a wrong or missing datum. This has happened in the case of the Nami wallet having a bug that caused the wrong address to be chosen. There are other instances of user error where they send from a CEX to a script address. A wrong datum can be handled by the Plutus scripts themselves by having an alternative execution branch if type does not match the expected datum type. But in the case of no datum the script is not run and fails in phase 1 validation. The other motivation of this CIP is to be able to create spend scripts that can handle the no datum case.
 
 I see three major use cases when it comes to running spend scripts without datums:
 
@@ -59,35 +59,43 @@ All the script purposes have a form of ```Redeemer -> ScriptContext -> a``` exce
 We propose to make the following modification:
 
 The signature of all scripts will be ```ScriptContext -> a```.
-The ScriptInfo type is a union type with a variant for each script purpose.
-```
+The `ScriptInfo` type is a union type with a variant for each script purpose.
+It is the same as `ScriptPurpose`, except for the additional optional datum in spending scripts.
+
+```haskell
+-- | The context that the currently-executing script can access.
 data ScriptContext = ScriptContext
-  { scriptContextTxInfo  :: TxInfo
+  { scriptContextTxInfo     :: TxInfo
+  -- ^ information about the transaction the currently-executing script is included in
+  , scriptContextRedeemer   :: V2.Redeemer
+  -- ^ Redeemer for the currently-executing script
   , scriptContextScriptInfo :: ScriptInfo
+  -- ^ the purpose of the currently-executing script, along with information associated
+  -- with the purpose
   }
 
+-- | Like `ScriptPurpose` but with an optional datum for spending scripts.
 data ScriptInfo
-  = MintingScript V2.CurrencySymbol Redeemer
-  | SpendingScript V3.TxOutRef (Maybe Datum) Redeemer
-  | RewardingScript V2.Credential Redeemer
+  = MintingScript V2.CurrencySymbol
+  | SpendingScript V3.TxOutRef (Haskell.Maybe V2.Datum)
+  | RewardingScript V2.Credential
   | CertifyingScript
       Haskell.Integer
       -- ^ 0-based index of the given `TxCert` in `txInfoTxCerts`
       TxCert
-      Redeemer
-  | VotingScript Voter Redeemer
+  | VotingScript Voter
   | ProposingScript
       Haskell.Integer
       -- ^ 0-based index of the given `ProposalProcedure` in `txInfoProposalProcedures`
       ProposalProcedure
-      Redeemer
 ```
-The Datum is optional, which will enable us to allow the execution of spending scripts without a datum. 
+
+The datum in `SpendingScript` is optional, which will allow the execution of spending scripts without a datum.
 One more change will be needed on the ledger side in order to make the Datum optional for spending scripts.
 The ledger UTXOW rule needs to be relaxed, this ledger rule checks if a utxo has an existing datum if the address's payment credential is a phase 2 validation script.
 
 The ScriptPurpose type used in the Redeemers Map is left the same.
-
+It is used to uniquely identify a Plutus script within a transaction.
 
 
 ## Rationale: how does this CIP achieve its goals?
@@ -115,7 +123,7 @@ The Cardano Ledger and Cardano Plutus teams would need to implement this in foll
   IntersectMBO/plutus
   IntersectMBO/cardano-ledger
 
-The following languages that compile to uplc would need to update to support the new ScriptContext argument that 
+The following languages that compile to uplc would need to update to support the new ScriptContext argument that
 is passed in for the next Plutus Version:
 Aiken
 Helios
--- a/CIP-0069/README.md
+++ b/CIP-0069/README.md
@@ -22,6 +22,8 @@ The exact change would be to have every validator take as argument the redeemer 
 
 ## Motivation: why is this CIP necessary?
 
+### Multi-purpose Scripts
+
 As it stands the scripts being made on cardano often suffer this problem, and the tokens usually are made to be able to be minted at any time. This leads to further checks being made on the frontend and further fragilitiy of the systems we create. When a mutual dependency arises we are forced to choose which script gets to statically know what's the hash of the other, and which has to be provided 'during runtime'.
 
 - Use Case 1: Minting validator checks datum given to spending validator. The spending validator requires the token be present as witness of the datum's correctness.
@@ -30,49 +32,55 @@ As it stands the scripts being made on cardano often suffer this problem, and th
 
 - Use Case 3 (taken from Minswap's Dex V1): NFT is minted for the same reason as above. It allows a minting policy to later mint LP tokens with that unique id token name.
 
-We see a similar pattern repeating over and over again as witnessed by dapp developers and auditors alike. By allowing the multi-purpose policies (spending and any other) we increase the security of Cardano by giving us more confidence and allowing to design protocols that have their architecture driven by Cardano's features, not limited by Cardano's language.
+We see a similar pattern repeating over and over again as witnessed by dapp developers and auditors alike. By allowing the multi-purpose scripts (spending and any other) we increase the security of Cardano by giving us more confidence and allowing to design protocols that have their architecture driven by Cardano's features, not limited by Cardano's language.
 
 This primarily manifests in the ability to use a single validator for both minting and spending but the proposed solution makes it possible to use one validator for any and all purposes at once.
+
+### No Datum Spend Purpose
+
+One of the major footguns of Plutus scripts is if a user sends to the script with a wrong or missing datum. This has happened in the case of the Nami wallet having a bug that caused the wrong address to be chosen. There are other instances of user error where they send from a CEX to a script address. A wrong datum can be handled by the Plutus scripts themselves by having an alternative execution branch if type does not match the expected datum type. But in the case of no datum the script is not run and fails in phase 1 validation. The other motivation of this CIP is to be able to create spend scripts that can handle the no datum case. 
+
+I see three major use cases when it comes to running spend scripts without datums:
+
+- Use Case 1: A script that acts as a wallet for users. By having no datum spending the user can send directly from exchanges or have friends send to their smart contract wallet with no datum needed.
+
+- Use Case 2: As a DAO treasury. The funds in this script would be controlled by a DAO and anyone can donate/contribute to the DAO without a datum.
+
+- Use Case 3: Allow dApp protocols to have a claim or withdraw mechanism (similar to Ethereum for tokens sent to a contract without call) for claiming tokens sent without a datum.
+
+I'd be remiss if I didn't mention CIP-0112 which has been expanded to improve native script capabilities to provide an alternative solution for use case 1 and 2. But for use case 3, CIP-0112 does not enable a "claim or withdraw mechanism" for contracts.
 
 ## Specification
 
 ### Removing the datum argument
 
-All the script purposes have a form of `Redeemer -> ScriptContext -> a` except the `Spending` one.
-It has the following form: `Datum -> Redeemer -> ScriptContext -> a`. This is enforced by the Cardano ledger.
+All the script purposes have a form of ```Redeemer -> ScriptContext -> a``` except the Spending one. It has the following form: ```Datum -> Redeemer -> ScriptContext -> a```. This is enforced by the Cardano ledger.
 
-We make the following modification:
+We propose to make the following modification:
 
-- The signature of all scripts will be `Redeemer -> ScriptContext -> a`.
-- The `ScriptPurpose` type is modified such that the `Spending` variant contains an optional `Datum`:
-
-```hs
-data ScriptPurpose
-  = ...
-  | Spending TxOutRef (Maybe Datum)
-    ...
+The signature of all scripts will be ```ScriptContext -> a```.
+The ScriptContext type is now a union type with a variant for each script purpose.
 ```
-
-The `Datum` is optional, which will enable us to allow executing spending scripts without a datum.
-
-- A new data type `ScriptIdentifier` is created.
-  It is identical to the original `ScriptPurpose` and is used in the `txInfoRedeemers` map to uniquely identify a script within a transaction.
-
-```hs
-data ScriptIdentifier
-  = MintingScript CurrencySymbol
-  | SpendingScript TxOutRef
-  | RewardingScript Credential
-  | CertifyingScript Integer TxCert
-  | VotingScript Voter
-  | ProposingScript Integer ProposalProcedure
+data ScriptContext
+  = Minting TxInfo Redeemer CurrencySymbol
+  | Spending TxInfo Redeemer (Maybe Datum) TxOutRef
+  | Rewarding TxInfo Redeemer Credential
+  | Certifying TxInfo Redeemer Integer TxCert 
+  | Voting TxInfo Redeemer Voter
+  | Proposing TxInfo Redeemer Integer ProposalProcedure
 ```
+The Datum is optional, which will enable us to allow the execution of spending scripts without a datum.
+
+The ScriptPurpose type is left the same used in the Redeemers Map is left the same.
 
 ## Rationale: how does this CIP achieve its goals?
 
 Unifying of the script signature is a very elegant solution to the problem, streamlining the experience of developing on cardano.
 Given that accessing the datum is almost always made by a spending script, it makes sense to introduce that argument back to the `ScriptPurpose` that now plays a more important role.
 It begs the question if it should be added as an argument to all validators, to further emphasize that fact.
+
+
+This CIP turns all scripts into 1 arg scripts with a Script Context union type for each purpose.
 
 ## Backwards compatibility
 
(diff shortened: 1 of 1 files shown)


rphair:
@zliu41 we're reviewing some related & alternative proposals at the next CIP meeting (https://hackmd.io/@cip-editors/84) and expect good Plutus representation there, so I've added this to the agenda.  Would be nice if you could make it to that meeting. 🤓

MicroProofs:
I am in support of this alternative over my own PR.

CIP-0069/README.md
```diff
@@ -41,12 +41,29 @@ This primarily manifests in the ability to use a single validator for both minti
...
-We modify the general signature to be `ScriptArgs -> ScriptContext -> a`.
+We make the following modification:
+
+- The signature of all scripts will be `Redeemer -> ScriptContext -> a`.
+- The `ScriptPurpose` type is modified such that the `Spending` variant contains a `Datum`:
+
+```hs
+data ScriptPurpose
+  = ...
+  | Spending TxOutRef Datum
```
MicroProofs, line 52: Should this be Maybe Datum instead of Datum? Or is this a Datum field containing
Datum Hash, InlineDatum, No Datum?

michaelpj:
I was thinking of a different variant, something like this:

- All scripts have the signature `EvaluationInfo -> TxInfo -> ()`
- `ScriptPurpose` as used elsewhere is left untouched (the ledger uses the "purpose" terminology so I think we should keep it)
- The type `EvaluationInfo` is defined similarly to `ScriptPurpose`, as follows:

```
data EvaluationInfo
  = Minting CurrencySymbol Redeemer
  | Spending TxOutRef Redeemer Datum
  | Rewarding Credential Redeemer
  | Certifying Integer TxCert Redeemer
  | Voting Voter Redeemer
  | Proposing Integer ProposalProcedure Redeemer
```

("Evaluation" info since it contains information about the specific evaluation of a script that is occurring, since a given script can be evaluated multiple times in multiple ways. Could have a better name)

zliu41:
> All scripts have the signature EvaluationInfo -> TxInfo -> ()

The problem this naming is that `TxInfo` is also the info you need for evaluating a script, i.e., "evaluation info". I think I like "redeemer" and "script context" better.

> ScriptPurpose as used elsewhere is left untouched (the ledger uses the "purpose" terminology so I think we should keep it)

Could be true, but I thought it'd be just as easy for the ledger if we add Datum to ScriptPurpose. I believe when constructing ScriptPurpose, the ledger does have access to the Datum, so it can simply stuff it in. But I could be wrong. @lehins @WhatisRT let me know whether you prefer ScriptContext/ScriptIdentifier, or ScriptInfo/ScriptContext (in both cases, the former contains the additional datum).

MicroProofs:
> > All scripts have the signature EvaluationInfo -> TxInfo -> ()
> 
> The problem with this naming is that `TxInfo` is also the info you need for evaluating a script, i.e., "evaluation info". I think I like "redeemer" and "script context" better.
> 
> > ScriptPurpose as used elsewhere is left untouched (the ledger uses the "purpose" terminology so I think we should keep it)
> 
> Could be true, but I thought it'd be just as easy for the ledger if we add Datum to ScriptPurpose. I believe when constructing ScriptPurpose, the ledger does have access to the Datum, so it can simply stuff it in. But I could be wrong. @lehins @WhatisRT let me know whether you prefer ScriptContext/ScriptIdentifier, or ScriptInfo/ScriptContext (in both cases, the former contains the additional datum).

What would the case be for no datum btw?

zliu41:
> What would the case be for no datum btw?

I thought allowing datum to be optional is the subject of another CIP, but `Maybe Datum` seems sensible to me.

zliu41:
Updated; datum is now optinal

rphair:
@zygomeb this was reviewed at the CIP editor's meeting today & the response was favourable, indicating that the IOG Plutus team would prefer to do it this way (cc @michaelpj) ... i.e. they would prefer to implement your original proposal with this change.

I'm marking it `Last Check` with the hope we can merge this in 2 weeks' time so please post here if you have any objections before then... also it would help to have your affirmation if you can provide it, or to settle any reservations in discussion here in the meantime.

WhatisRT:
I think I don't mind adding the `Datum` to `ScriptPurpose`, it seems like it might actually simplify some code a little bit. It also makes sense philosophically since we sometimes treat the `Datum` as part of the `Credential`.

MicroProofs:
@lehins 

Is it possible to make this or a simplified version of just Maybe Datum instead of Datum make it into the Chang hardfork?
I think the importance of having no datum spend execution will greatly enhance the Dev experience and UX.

I believe the feature is worth delaying the Chang hard fork if we can get even a more simplified version of this CIP.

rphair:
re: https://github.com/cardano-foundation/CIPs/pull/784#issuecomment-2059166336 @MicroProofs I've added this question to tonight's CIP meeting agenda in the hope you can make it.  If not done already in the PR thread in the meantime, we can clarify your request and maybe get feedback from other developers about how much their experience might be affected with & without it.

MicroProofs:
Ok I'll make it to the meeting today.

leo42:
I have said it many time at every opportunity I get basically, but this functionality is game-changing for smart-wallets and will enable an entire new class of users for the ecosystem.

for my work the lack of this functionality has been a blocking issue for a very long time.

MicroProofs:
> @lehins
> 
> Is it possible to make this or a simplified version of just Maybe Datum instead of Datum make it into the Chang hardfork? I think the importance of having no datum spend execution will greatly enhance the Dev experience and UX.
> 
> I believe the feature is worth delaying the Chang hard fork if we can get even a more simplified version of this CIP.

If the ledger team is not feature frozen, I am wondering if it's possible to change the ledger to pass in a Maybe Datum instead of Datum.

Here is the code that would change

```
 Right scriptContext ->
          let spendingDatum = getSpendingDatum utxo tx $ hoistPlutusPurpose toAsItem plutusPurpose
              datums = maybe id (:) spendingDatum [d, scriptContext]
           in Right $
                withPlutusScript plutusScript $ \plutus ->
                  PlutusWithContext
                    { pwcProtocolVersion = protVerMajor
                    , pwcScript = Left plutus
                    , pwcScriptHash = plutusScriptHash
                    , pwcDatums = PlutusDatums (getPlutusData <$> datums)
                    , pwcExUnits = exUnits
                    , pwcCostModel = costModel
                    }
```

I believe here you would change how spendingDatum returns to differentiate between NoDatum vs other script purposes.

This would enable a wide variety of features here

Use cases include

Smart wallets, wallets that can do 2fa, vault features
(prevent wallet drainers), and intention based transactions (single transactions swaps with dexes and free limit order placements).

Datumless utxos accidentally sent to a Dapp could be recovered for the user.

Easier subscription services that don't require you to send funds to a contract

lehins:
Sorry guys, I was on PTO for a while and was catching up with all of the other important Conway related issues.

Unfortunately proposed solution requires quite a bit of work on the ledger side and we just don't have enough time to get it done, considering how close we are to the Chang hardfork deadline.

> I believe the feature is worth delaying the Chang hard fork ...

Such decision is definitely above my pay grade. :smiling_face_with_tear: So if we were to postpone the deadline this would have to come from the upper management, not from the developers.

With respect to the CIP. To be honest with you I am not very fond of the proposed solution and I think we can do better if we are not gonna rush it. That being said we might be able to do an interim solution for PlutusV3 that resembles the original suggestion quite easily, if there is interest in it. Solution I am talking about is what @MicroProofs suggested:

> If the ledger team is not feature frozen, I am wondering if it's possible to change the ledger to pass in a Maybe Datum instead of Datum.

In my opinion this is a safer solution, albeit less type safe, than the one proposed in this CIP. In particular my worry is that plutus evaluator is essentially a variadic function that will always succeed if it is not fully saturated. Which means that if we reduce number of arguments to 2 for spending scripts then older scripts blindly ported to PlutusV3 would always succeed, because they would not have enough arguments.

What I personally would suggest instead, since everyone is in favor of having the same number of arguments for all type of scripts, is to stop accepting a list of `Data` arguments to plutus evaluator altogether and completely switch to accepting a single argument. That argument could be even called `ScriptContext` and it would contain all of the required information:

```haskell
data ScriptContext
  = Minting TxInfo Redeemer CurrencySymbol
  | Spending TxInfo Redeemer Datum TxOutRef
  | Rewarding TxInfo Redeemer Credential
  | Certifying TxInfo Redeemer Integer TxCert 
  | Voting TxInfo Redeemer Voter
  | Proposing TxInfo Redeemer Integer ProposalProcedure
```
In other words changing the type signature of this function to:

https://github.com/IntersectMBO/plutus/blob/16a986fd800e044773e27718fc78c0c2885854e6/plutus-ledger-api/src/PlutusLedgerApi/V3.hs#L207-L221

```haskell
evaluateScriptRestricting ::
  -- | Which protocol version to run the operation in
  MajorProtocolVersion ->
  -- | Whether to produce log output
  VerboseMode ->
  -- | Includes the cost model to use for tallying up the execution costs
  EvaluationContext ->
  -- | The resource budget which must not be exceeded during evaluation
  ExBudget ->
  -- | The script to evaluate
  ScriptForEvaluation ->
  -- | The arguments to the script
  PLC.Data -> -- <-------- This is the change we should aim for.
  (LogOutput, Either EvaluationError ExBudget)
```

I honestly don't understand what was the reason for evaluator to accept a list of `Data` arguments if `Data` can already represent a list `Data`.

Of course, this is just my opinion and maybe it either needs to be polished a bit more or maybe it has some serious downsides that I am not aware. In either case neither this solution nor the one proposed in the CIP can make it into Conway due to the time constraints.

Back to the potential interim solution for PlutusV3 proposed in https://github.com/cardano-foundation/CIPs/pull/784#issuecomment-2059523313. Please, let me know if it would be a sufficient interim solution for Plutus script writers and if it would at least unblock all those use cases mentioned in this discussion. If there would be no objections from all the parties interested in this we could potentially squeeze it in for Conway, because it is simple and requires no changes on the Plutus side and very minimal changes on the Ledger side. Here is a proof of concept implemented in ledger: https://github.com/IntersectMBO/cardano-ledger/pull/4271

Couple of notes about the interim solution:

* There is no `Maybe` type in plutus `Data`, so `Maybe Data` has to be represented as a singleton list with an optional element `List [Data]`. Feel free to suggest a different representation.
* This extra new argument will not be optional for Spending scripts, that is because Spending datum is enforced by the ledger rules, which we are not going to change on such a short notice. It is a lot more complicated than it sounds. So this https://github.com/cardano-foundation/CIPs/pull/784#issuecomment-2005440532 is not part of this interim solution.

MicroProofs:
> Sorry guys, I was on PTO for a while and was catching up with all of the other important Conway related issues.
> 
> Unfortunately proposed solution requires quite a bit of work on the ledger side and we just don't have enough time to get it done, considering how close we are to the Chang hardfork deadline.
> 
> > I believe the feature is worth delaying the Chang hard fork ...
> 
> Such decision is definitely above my pay grade. 🥲 So if we were to postpone the deadline this would have to come from the upper management, not from the developers.
> 
> With respect to the CIP. To be honest with you I am not very fond of the proposed solution and I think we can do better if we are not gonna rush it. That being said we might be able to do an interim solution for PlutusV3 that resembles the original suggestion quite easily, if there is interest in it. Solution I am talking about is what @MicroProofs suggested:
> 
> > If the ledger team is not feature frozen, I am wondering if it's possible to change the ledger to pass in a Maybe Datum instead of Datum.
> 
> In my opinion this is a safer solution, albeit less type safe, than the one proposed in this CIP. In particular my worry is that plutus evaluator is essentially a variadic function that will always succeed if it is not fully saturated. Which means that if we reduce number of arguments to 2 for spending scripts then older scripts blindly ported to PlutusV3 would always succeed, because they would not have enough arguments.
> 
> What I personally would suggest instead, since everyone is in favor of having the same number of arguments for all type of scripts, is to stop accepting a list of `Data` arguments to plutus evaluator altogether and completely switch to accepting a single argument. That argument could be even called `ScriptContext` and it would contain all of the required information:
> 
> ```haskell
> data ScriptContext
>   = Minting TxInfo Redeemer CurrencySymbol
>   | Spending TxInfo Redeemer Datum TxOutRef
>   | Rewarding TxInfo Redeemer Credential
>   | Certifying TxInfo Redeemer Integer TxCert 
>   | Voting TxInfo Redeemer Voter
>   | Proposing TxInfo Redeemer Integer ProposalProcedure
> ```
> 
> In other words changing the type signature of this function to:
> 
> https://github.com/IntersectMBO/plutus/blob/16a986fd800e044773e27718fc78c0c2885854e6/plutus-ledger-api/src/PlutusLedgerApi/V3.hs#L207-L221
> 
> ```haskell
> evaluateScriptRestricting ::
>   -- | Which protocol version to run the operation in
>   MajorProtocolVersion ->
>   -- | Whether to produce log output
>   VerboseMode ->
>   -- | Includes the cost model to use for tallying up the execution costs
>   EvaluationContext ->
>   -- | The resource budget which must not be exceeded during evaluation
>   ExBudget ->
>   -- | The script to evaluate
>   ScriptForEvaluation ->
>   -- | The arguments to the script
>   PLC.Data -> -- <-------- This is the change we should aim for.
>   (LogOutput, Either EvaluationError ExBudget)
> ```
> 
> I honestly don't understand what was the reason for evaluator to accept a list of `Data` arguments if `Data` can already represent a list `Data`.
> 
> Of course, this is just my opinion and maybe it either needs to be polished a bit more or maybe it has some serious downsides that I am not aware. In either case neither this solution nor the one proposed in the CIP can make it into Conway due to the time constraints.
> 
> Back to the potential interim solution for PlutusV3 proposed in [#784 (comment)](https://github.com/cardano-foundation/CIPs/pull/784#issuecomment-2059523313). Please, let me know if it would be a sufficient interim solution for Plutus script writers and if it would at least unblock all those use cases mentioned in this discussion. If there would be no objections from all the parties interested in this we could potentially squeeze it in for Conway, because it is simple and requires no changes on the Plutus side and very minimal changes on the Ledger side. Here is a proof of concept implemented in ledger: [IntersectMBO/cardano-ledger#4271](https://github.com/IntersectMBO/cardano-ledger/pull/4271)
> 
> Couple of notes about the interim solution:
> 
> * There is no `Maybe` type in plutus `Data`, so `Maybe Data` has to be represented as a singleton list with an optional element `List [Data]`. Feel free to suggest a different representation.
> * This extra new argument will not be optional for Spending scripts, that is because Spending datum is enforced by the ledger rules, which we are not going to change on such a short notice. It is a lot more complicated than it sounds. So this [CIP-0069 | add Datum to ScriptContext #784 (comment)](https://github.com/cardano-foundation/CIPs/pull/784#issuecomment-2005440532) is not part of this interim solution.

My apologies on suggesting the delayed hard fork. The better option is to rally community support to put it at the forefront of the next HF or possibly an early mini one (depending on community feedback and testing and scope of changes of course)

MicroProofs:
> Back to the potential interim solution for PlutusV3 proposed in [#784 (comment)](https://github.com/cardano-foundation/CIPs/pull/784#issuecomment-2059523313). Please, let me know if it would be a sufficient interim solution for Plutus script writers and if it would at least unblock all those use cases mentioned in this discussion. If there would be no objections from all the parties interested in this we could potentially squeeze it in for Conway, because it is simple and requires no changes on the Plutus side and very minimal changes on the Ledger side. Here is a proof of concept implemented in ledger: [IntersectMBO/cardano-ledger#4271](https://github.com/IntersectMBO/cardano-ledger/pull/4271)
> 
> Couple of notes about the interim solution:
> 
> * There is no `Maybe` type in plutus `Data`, so `Maybe Data` has to be represented as a singleton list with an optional element `List [Data]`. Feel free to suggest a different representation.
> * This extra new argument will not be optional for Spending scripts, that is because Spending datum is enforced by the ledger rules, which we are not going to change on such a short notice. It is a lot more complicated than it sounds. So this [CIP-0069 | add Datum to ScriptContext #784 (comment)](https://github.com/cardano-foundation/CIPs/pull/784#issuecomment-2005440532) is not part of this interim solution.

The interim solution with both enforcing that Spend still takes in 3 arguments as well as it being a List [Data] is a fine solution for PlutusV3. This would work for the use cases mentioned and unblocks a lot for wallet maker to move forward with smart contract wallets. Since this change requires buy in from the active Language developers, I will tag them here. If you need a new CIP as well I can create one. This is in my opinion a good improvement over status quo. I'll tag the other language creators.

@nielstron @nau @michele-nuzzi @christianschmitz @michaelpj 
I have not tagged Plutarch creator because I believe that is not in development for PlutusV3 support.

MicroProofs:
> The interim solution with both enforcing that Spend still takes in 3 arguments as well as it being a List [Data] is a fine solution for PlutusV3. 

An alternative solution is constr 0 {[Datum]} to represent Some(Datum) and constr 1 {[]} to represent None in Plutus Data
I'll let others way in to which decision is preferable.

nielstron:
I would prefer the constr0/constr1 distinction (which is essentially PlutusData maybe).

A list gives the illusion of the possibility of multiple parameters to the script.

That said my opinion on this is not very strong except for "it would be really nice to have any solution ASAP" :)

lehins:
Please thumbs up who is in favor of and thumbs down who is against this representation for `Maybe Datum`:

Going for the best correspondence for the Haskell's `data Maybe a = Just a | Nothing` definition:
* `Constr 0 [Data]` for `Just Data`
* `Constr 1 []` for `Nothing`

lehins:
Same as above, but constructors swapped as in [proposed solution](https://github.com/cardano-foundation/CIPs/pull/784#issuecomment-2060316435) by @MicroProofs:
>  alternative solution is constr 0 {[Datum]} to represent Some(Datum) and constr 1 {[]} to represent None in Plutus Data

* `Constr 0 []` for `Nothing`
* `Constr 1 [Data]` for `Just Data`

lehins:
Just for completeness sake. Same as above :+1: and :-1:, for the less popular solution of using `List`:

* `List []` for `Nothing`
* `List [Data]` for `Just Data`

KtorZ:
> Please thumbs up 👍 who is in favor of or thumbs down 👎 who is against this representation for `Maybe Datum`:
> 
> Going for the best correspondence for the Haskell's `data Maybe a = Just a | Nothing` definition:
> 
>     * `Constr 0 [Data]` for `Just Data`
> 
>     * `Constr 1 []` for `Nothing`

It seems to me that this is the most (and only?) viable choice given that it's already the encoding used for `Maybe a` in Haskell and `Option<a>` in Aiken & Helios. I assume it's the same for OpShin, Plu-ts and Scalus. Going for the opposite (`Constr 0 []` for `Nothing`) would be utterly confusing without also changing this in the serialization of all ABIs.

I believe @MicroProofs just mixed up the constructor indices in his comment 😅, but the core of the suggestion was to say "use Constr instead of a List".  

@nau @nielstron any reason for preferring `Constr 0 []` for `Nothing` 🤔?

michele-nuzzi:
the "Haskell Maybe" (Just  contsr 0 and Nothing constr 1) representation is already used anyway in V1 TxOuts for datum hashes and V2 TxOuts for ref scripts; I don't see why we should go for something else

nielstron:
> @nau @nielstron any reason for preferring Constr 0 [] for Nothing 🤔?

for my part no, just confusion. sticking to the existing way for maybe is better.

MicroProofs:
> > Please thumbs up 👍 who is in favor of or thumbs down 👎 who is against this representation for `Maybe Datum`:
> 
> > 
> 
> > Going for the best correspondence for the Haskell's `data Maybe a = Just a | Nothing` definition:
> 
> > 
> 
> >     * `Constr 0 [Data]` for `Just Data`
> 
> > 
> 
> >     * `Constr 1 []` for `Nothing`
> 
> 
> 
> It seems to me that this is the most (and only?) viable choice given that it's already the encoding used for `Maybe a` in Haskell and `Option<a>` in Aiken & Helios. I assume it's the same for OpShin, Plu-ts and Scalus. Going for the opposite (`Constr 0 []` for `Nothing`) would be utterly confusing without also changing this in the serialization of all ABIs.
> 
> 
> 
> I believe @MicroProofs just mixed up the constructor indices in his comment 😅, but the essence of the suggestion was about suggesting to "use Constr instead of a List".  
> 
> 
> 
> @nau @nielstron any reason for preferring `Constr 0 []` for `Nothing` 🤔?  
> 
> 
> 
> 
> 
> 

I don't understand my comment matches Some with tag 0 and None with tag 1? I am the crazy one lol.

KtorZ:
Indeed. So I think we're all saying the same thing :D

nau:
> @nau @nielstron any reason for preferring `Constr 0 []` for `Nothing` 🤔?

No, let's do it same as everywhere. Just confused which one was the OG :)

lehins:
Looks like we are all on the same page, this is the preferred representation
* `Constr 0 [Data]` for `Just Data`
* `Constr 1 []` for `Nothing`

> I don't understand my comment matches Some with tag 0 and None with tag 1? Am I the crazy one lol.

@MicroProofs Sorry, it was me who mixed it up :see_no_evil: It was getting pretty late for me when I was writing those messages.

I'll ask around and see if there are any objects from the Plutus team and if there are none, then we'll go with this interim solution for Conway and we'll plan to fix it properly for the next era.

christianschmitz:
I agree with the consensus:

`Constr(0, [data])` for Some, `Constr(1, [])` for None

For the Helios language, I have a slight preference for the Empty/SIngleton List approach because it is cheaper. However, the off- to on-chain interface benefits from having the most consistent representation possible, so best keep it unchanged wrt. the current representation.

colll78:
> > Back to the potential interim solution for PlutusV3 proposed in [#784 (comment)](https://github.com/cardano-foundation/CIPs/pull/784#issuecomment-2059523313). Please, let me know if it would be a sufficient interim solution for Plutus script writers and if it would at least unblock all those use cases mentioned in this discussion. If there would be no objections from all the parties interested in this we could potentially squeeze it in for Conway, because it is simple and requires no changes on the Plutus side and very minimal changes on the Ledger side. Here is a proof of concept implemented in ledger: [IntersectMBO/cardano-ledger#4271](https://github.com/IntersectMBO/cardano-ledger/pull/4271)
> > Couple of notes about the interim solution:
> > 
> > * There is no `Maybe` type in plutus `Data`, so `Maybe Data` has to be represented as a singleton list with an optional element `List [Data]`. Feel free to suggest a different representation.
> > * This extra new argument will not be optional for Spending scripts, that is because Spending datum is enforced by the ledger rules, which we are not going to change on such a short notice. It is a lot more complicated than it sounds. So this [CIP-0069 | add Datum to ScriptContext #784 (comment)](https://github.com/cardano-foundation/CIPs/pull/784#issuecomment-2005440532) is not part of this interim solution.
> 
> The interim solution with both enforcing that Spend still takes in 3 arguments as well as it being a List [Data] is a fine solution for PlutusV3. This would work for the use cases mentioned and unblocks a lot for wallet maker to move forward with smart contract wallets. Since this change requires buy in from the active Language developers, I will tag them here. If you need a new CIP as well I can create one. This is in my opinion a good improvement over status quo. I'll tag the other language creators.
> 
> @nielstron @nau @michele-nuzzi @christianschmitz @michaelpj I have not tagged Plutarch creator because I believe that is not in development for PlutusV3 support.

Plutarch is in development for PlutusV3 support.

leo42:
> So to be clear, the idea is to retroactively accept #440? We should revise it and resubmit it if that is the case.
> 
> > This extra new argument will not be optional for Spending scripts
> 
> I just want to be sure that everyone is crystal clear on what this means: the change being proposed here will _not_ allow you to make script outputs without datums. It will make it possible to _represent_ that case, but it will not be allowed by the ledger. So if that is the thing you want, it will need at least another HF. And if that's the primary motivator, then I think there is no reason to rush this change in anyway.

This will not fulfill my needs. 

Just to clarify when you say we will not be able to make script outputs without datums, you mean consume utxos inside script 
 controlled address without a datum?

michele-nuzzi:
> So to be clear, the idea is to retroactively accept #440? We should revise it and resubmit it if that is the case.

@michaelpj No

The idea is to [move the datum inside the ScriptPurpose](https://github.com/cardano-foundation/CIPs/pull/784/commits/91c0493ebf0bc92089504e071dcbc234aac16bc2#diff-ea330e3feffc9cf436408b712ffacaac438815fed35af265c2a7bae6abfd7ef4R52)

so that the Spending script purpose ( the only one that needs the datum ) will have an additional field with "arbitrary data" which is the datum

MicroProofs:
> So to be clear, the idea is to retroactively accept #440? We should revise it and resubmit it if that is the case.
> 
> > This extra new argument will not be optional for Spending scripts
> 
> I just want to be sure that everyone is crystal clear on what this means: the change being proposed here will _not_ allow you to make script outputs without datums. It will make it possible to _represent_ that case, but it will not be allowed by the ledger. So if that is the thing you want, it will need at least another HF. And if that's the primary motivator, then I think there is no reason to rush this change in anyway.

I thought the ledger change being made here by @lehins was allowing utxos without datums at a script payment credentials to be spent by that script if made after PlutusV3? I don't think the ledger right now prevents you from making a utxo at an address that is a plutus script without a datum. 

Can you go in more detail what you mean?

MicroProofs:
> > So to be clear, the idea is to retroactively accept #440? We should revise it and resubmit it if that is the case.
> 
> @michaelpj No
> 
> The idea is to [move the datum inside the ScriptPurpose](https://github.com/cardano-foundation/CIPs/pull/784/commits/91c0493ebf0bc92089504e071dcbc234aac16bc2#diff-ea330e3feffc9cf436408b712ffacaac438815fed35af265c2a7bae6abfd7ef4R52)
> 
> so that the Spending script purpose ( the only one that needs the datum ) will have an additional field with "arbitrary data" which is the datum

The interim solution recently proposed does not attempt to move the Datum inside the script context due to the rippling changes this causes throughout the ledger.

michaelpj:
@michele-nuzzi That's not the thing that @lehins wrote. We're not discussing the change in the PR diff any more, it seems.

> I thought the ledger change being made here by @lehins was allowing utxos without datums at a script payment credentials to be spent by that script if made after PlutusV3?

AIUI he is proposing _only_ to change the signature of the arguments, _not_ to make the ledger allow spending outputs without datums and setting it to Nothing. So `Spending` scripts will always get `Just` the datum. The only change will be that all scripts will have the same number of arguments.

This is important - I think previous discussions of this topic have assumed that these changes go together but they do not necessarily do so, and I believe @lehins is not proposing to make the ledger rule changes that would be necessary there.

MicroProofs:
> Sorry guys, I was on PTO for a while and was catching up with all of the other important Conway related issues.
> 
> Unfortunately proposed solution requires quite a bit of work on the ledger side and we just don't have enough time to get it done, considering how close we are to the Chang hardfork deadline.
> 
> > I believe the feature is worth delaying the Chang hard fork ...
> 
> Such decision is definitely above my pay grade. 🥲 So if we were to postpone the deadline this would have to come from the upper management, not from the developers.
> 
> With respect to the CIP. To be honest with you I am not very fond of the proposed solution and I think we can do better if we are not gonna rush it. That being said we might be able to do an interim solution for PlutusV3 that resembles the original suggestion quite easily, if there is interest in it. Solution I am talking about is what @MicroProofs suggested:
> 
> > If the ledger team is not feature frozen, I am wondering if it's possible to change the ledger to pass in a Maybe Datum instead of Datum.
> 
> In my opinion this is a safer solution, albeit less type safe, than the one proposed in this CIP. In particular my worry is that plutus evaluator is essentially a variadic function that will always succeed if it is not fully saturated. Which means that if we reduce number of arguments to 2 for spending scripts then older scripts blindly ported to PlutusV3 would always succeed, because they would not have enough arguments.
> 
> What I personally would suggest instead, since everyone is in favor of having the same number of arguments for all type of scripts, is to stop accepting a list of `Data` arguments to plutus evaluator altogether and completely switch to accepting a single argument. That argument could be even called `ScriptContext` and it would contain all of the required information:
> 
> ```haskell
> data ScriptContext
>   = Minting TxInfo Redeemer CurrencySymbol
>   | Spending TxInfo Redeemer Datum TxOutRef
>   | Rewarding TxInfo Redeemer Credential
>   | Certifying TxInfo Redeemer Integer TxCert 
>   | Voting TxInfo Redeemer Voter
>   | Proposing TxInfo Redeemer Integer ProposalProcedure
> ```
> 
> In other words changing the type signature of this function to:
> 
> https://github.com/IntersectMBO/plutus/blob/16a986fd800e044773e27718fc78c0c2885854e6/plutus-ledger-api/src/PlutusLedgerApi/V3.hs#L207-L221
> 
> ```haskell
> evaluateScriptRestricting ::
>   -- | Which protocol version to run the operation in
>   MajorProtocolVersion ->
>   -- | Whether to produce log output
>   VerboseMode ->
>   -- | Includes the cost model to use for tallying up the execution costs
>   EvaluationContext ->
>   -- | The resource budget which must not be exceeded during evaluation
>   ExBudget ->
>   -- | The script to evaluate
>   ScriptForEvaluation ->
>   -- | The arguments to the script
>   PLC.Data -> -- <-------- This is the change we should aim for.
>   (LogOutput, Either EvaluationError ExBudget)
> ```
> 
> I honestly don't understand what was the reason for evaluator to accept a list of `Data` arguments if `Data` can already represent a list `Data`.
> 
> Of course, this is just my opinion and maybe it either needs to be polished a bit more or maybe it has some serious downsides that I am not aware. In either case neither this solution nor the one proposed in the CIP can make it into Conway due to the time constraints.
> 
> Back to the potential interim solution for PlutusV3 proposed in [#784 (comment)](https://github.com/cardano-foundation/CIPs/pull/784#issuecomment-2059523313). Please, let me know if it would be a sufficient interim solution for Plutus script writers and if it would at least unblock all those use cases mentioned in this discussion. If there would be no objections from all the parties interested in this we could potentially squeeze it in for Conway, because it is simple and requires no changes on the Plutus side and very minimal changes on the Ledger side. Here is a proof of concept implemented in ledger: [IntersectMBO/cardano-ledger#4271](https://github.com/IntersectMBO/cardano-ledger/pull/4271)
> 
> Couple of notes about the interim solution:
> 
> * There is no `Maybe` type in plutus `Data`, so `Maybe Data` has to be represented as a singleton list with an optional element `List [Data]`. Feel free to suggest a different representation.
> * This extra new argument will not be optional for Spending scripts, that is because Spending datum is enforced by the ledger rules, which we are not going to change on such a short notice. It is a lot more complicated than it sounds. So this [CIP-0069 | add Datum to ScriptContext #784 (comment)](https://github.com/cardano-foundation/CIPs/pull/784#issuecomment-2005440532) is not part of this interim solution.

@michaelpj 

To my knowledge the PR linked should allow utxos without datums to be spent by scripts since there will always be 3 arguments. Could you let me know what else is blocking the spending of datumless utxos?

lehins:
> I don't think the ledger right now prevents you from making a utxo at an address that is a plutus script without a datum.

Yes it does so inderectly. It prevents you from spending it, if it is locked by a Plutus script and datum is not supplied

lehins:
@MicroProofs See the very last thing in my comment:

> This extra new argument will not be optional for Spending scripts, that is because Spending datum is enforced by the ledger rules, which we are not going to change on such a short notice. It is a lot more complicated than it sounds. So this https://github.com/cardano-foundation/CIPs/pull/784#issuecomment-2005440532 is not part of this interim solution.

MicroProofs:
> > I don't think the ledger right now prevents you from making a utxo at an address that is a plutus script without a datum.
> 
> Yes it does so inderectly. It prevents you from spending it, if it is locked by a Plutus script and datum is not supplied

Does your proposed pr allow no datum sepnds

> @MicroProofs See the very last thing in my comment:
> 
> > This extra new argument will not be optional for Spending scripts, that is because Spending datum is enforced by the ledger rules, which we are not going to change on such a short notice. It is a lot more complicated than it sounds. So this [#784 (comment)](https://github.com/cardano-foundation/CIPs/pull/784#issuecomment-2005440532) is not part of this interim solution.

Doesn't your proposed solution allow no datums to resolve to Nothing based on the lookup function I copied over. Is there some other error in the ledger that prevents no datum utxo spends?

MicroProofs:
The interim solution was an attempt to bypass where the check for the Datum happens by replacing it with a Data instead. But I see it is in a separate file where the utxo itself is checked. Is there a solution we can do for this before Conway. This is the main blocker in this thread for various solutions.

lehins:
> The interim solution was an attempt to bypass where the check for the Datum happens by replacing it with a Data instead. ... Is there a solution we can do for this before Conway.

Sorry to be a bearer of bad news, but not without moving the Conway deadline.

Note that the goal of this CIP was to change the number of arguments that all scripts get. At least that was my understanding of the CIP. Making Datum optional was just a minor addition to the CIP. That is why I suggested the interim solution that would solve the main problem.

Looks like we all got too happy too soon. :disappointed:

MicroProofs:
```
datums <-
        case toSpendingPurpose plutusPurpose of
          Just (AsIxItem _ txIn) -> do
            txOut <- note (UnknownTxIn txIn) $ Map.lookup txIn (unUTxO utxo)
            datum <-
              case txOut ^. datumTxOutF of
                Datum binaryData -> pure $ binaryDataToData binaryData
                DatumHash dh -> note (MissingDatum dh) $ Map.lookup dh dats
                NoDatum -> Left (InvalidTxIn txIn)
            pure [datum, rdmr, scriptContext]
          Nothing -> pure [rdmr, scriptContext]
```

Is this code in  libs/cardano-ledger-api/src/Cardano/Ledger/Api/Scripts/ExUnits.hs
also an issue too?

lehins:
> Is this code in ... also an issue too?

Yes, but it is a minor issue. This is just a helper function for the cardano-api/cli It is not used by the ledger

CIP-0069/README.md
```diff
@@ -30,31 +32,60 @@ As it stands the scripts being made on cardano often suffer this problem, and th
...
+  | Voting TxInfo Redeemer Voter
+  | Proposing TxInfo Redeemer Integer ProposalProcedure
 ```
+The Datum is optional, which will enable us to allow the execution of spending scripts without a datum. 
+One more change will be needed on the ledger side in order to make the Datum optional for spending scripts.
+The ledger UTXOW rule needs to be relaxed, this ledger rule checks if a utxo has an existing datum if the address's payment credential is a phase 2 validation script.
+
+The ScriptPurpose type used in the Redeemers Map is left the same.
+
+
```
lehins, line 78: I suggest we also specify somewhere here maybe that we need enforce a single argumet to plutus evaluating function, which at the very least means changing type signatures for these two functions for `PlutusV3` in a way that they will accept a single `PLC.Data` agument, instead of `[PLC.Data]`:

* [`evaluateScriptCounting`](https://github.com/IntersectMBO/plutus/blob/7707aab438f7266bd99c0a72e78da7b6f708603c/plutus-ledger-api/src/PlutusLedgerApi/V3.hs#L186-L198)
* [`evaluateScriptRestricting`](https://github.com/IntersectMBO/plutus/blob/7707aab438f7266bd99c0a72e78da7b6f708603c/plutus-ledger-api/src/PlutusLedgerApi/V3.hs#L207-L221)

CIP-0069/README.md
```diff
@@ -30,31 +32,60 @@ As it stands the scripts being made on cardano often suffer this problem, and th
...
 
 ## Rationale: how does this CIP achieve its goals?
 
 Unifying of the script signature is a very elegant solution to the problem, streamlining the experience of developing on cardano.
 Given that accessing the datum is almost always made by a spending script, it makes sense to introduce that argument back to the `ScriptPurpose` that now plays a more important role.
```
lehins, line 83: `ScriptPurpose` will stay unchanged with these changes in the PR
```suggestion
```

lehins (commented):
Looks great. Couple suggestions from me

rphair:
@lehins @zliu41 @MicroProofs @michaelpj @KtorZ @colll78 should https://github.com/cardano-foundation/CIPs/issues/769 be updated with any of these deliberations?  Does what's being agreed upon here settle that issue?

rphair:
Because this has been technically at `Last Check` status here for 5+ weeks, I've put it for `Last Check` status on the next CIP Editors Meeting agenda (https://hackmd.io/@cip-editors/87) where either we can merge it... if @MicroProofs @lehins agree it now reflects currently agreed upon changes for the immediate next stage of this CIP's implementation... or remove the `Last Check` flag if in abeyance for whatever reason.

zliu41:
After some discussion here's the current proposal: All scripts have the signature of `ScriptContext -> a`. A new `ScriptInfo` type is to be created to replace the `ScriptPurpose` field in `ScriptContext`. `ScriptInfo` is similar to `ScriptPurpose`, except that each constructor contains a `Redeemer` argument, and the `Spending` constructor contains a `Maybe Datum` argument. The `ScriptPurpose` type is unchanged and continues to be used as the key of the `txInfoRedeemers` map.

```haskell
data ScriptContext = ScriptContext
  { scriptContextTxInfo  :: TxInfo
  , scriptContextScriptInfo :: ScriptInfo
  -- ^ Previously this was `ScriptPurpose`
  }

data ScrintInfo
  = MintingScript V2.CurrencySymbol Redeemer
  | SpendingScript V3.TxOutRef (Maybe Datum) Redeemer
  | RewardingScript V2.Credential Redeemer
  | CertifyingScript
      Haskell.Integer
      -- ^ 0-based index of the given `TxCert` in `txInfoTxCerts`
      TxCert
      Redeemer
  | VotingScript Voter Redeemer
  | ProposingScript
      Haskell.Integer
      -- ^ 0-based index of the given `ProposalProcedure` in `txInfoProposalProcedures`
      ProposalProcedure
      Redeemer
```

zliu41:
I need to make another update before merging

zliu41:
I made an update to the `ScriptContext` and `ScriptInfo` types, and this is now ready to merge - cc @rphair @MicroProofs

PR #786, opened 2024-03-18
CIP-0010 | Register metadata key number: 620

rmcilraith:
At Seedtrace (https://seedtrace.org/), we have been working with the Cardano Foundation to set up storage of farmer fair payments data in the metadata of Cardano blockchain transactions. This is already live on the mainnet, and in the future we would like to extend the usage for other traceable data in our client's supply chains.

As suggested by our advisor from the Cardano Foundation, we have chosen a metadata key number which our transactions metadata will use and this PR is for registering this number to Seedtrace.

### DIFF
--- a/CIP-0010/registry.json
+++ b/CIP-0010/registry.json
@@ -20,6 +20,10 @@
     "description": "Proof of Existence record"
   },
   {
+    "transaction_metadatum_label": 620,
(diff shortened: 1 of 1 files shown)


Ryun1 (approved):
Looks good to me

rphair (approved):

PR #789, opened 2024-03-19
Update top-level README: post meeting #84

rphair:
Merge (before meeting, based on prior consensus):
- https://github.com/cardano-foundation/CIPs/pull/611

Promotions:
- https://github.com/cardano-foundation/CIPs/pull/747
- https://github.com/cardano-foundation/CIPs/pull/779
- https://github.com/cardano-foundation/CIPs/pull/780

### DIFF
--- a/README.md
+++ b/README.md
@@ -143,8 +143,10 @@ Below are listed tentative CIPs still under discussion with the community. They 
 | 0112? | [Observe script type](https://github.com/cardano-foundation/CIPs/pull/749) |
 | 0113? | [Programmable token-like assets](https://github.com/cardano-foundation/CIPs/pull/444) |
 | 0116? | [Universal JSON Encoding for Domain Types](https://github.com/cardano-foundation/CIPs/pull/766) |
+| 0117? | [Explicit script return values](https://github.com/cardano-foundation/CIPs/pull/747) |
+| 0118? | [Validation zones](https://github.com/cardano-foundation/CIPs/pull/780) |
 
-<p align="right"><i>Last updated on 2024-03-06</i></p>
+<p align="right"><i>Last updated on 2024-03-19</i></p>
 
 ## Cardano Problem Statements (CPS)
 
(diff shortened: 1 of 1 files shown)


Ryun1:
Since we are doing house keeping, do you mind fixing the Copywright section in the template;
we need to add a end comment [here](https://github.com/cardano-foundation/CIPs/blob/master/.github/CIP-TEMPLATE.md?plain=1#L54C187-L54C188), otherwise the uncommenting of different licenses doesnt work properly.
🤓

Ryun1 (approved):

rphair:
@Ryun1 I see what you mean... to someone who does a lot of HTML editing the whole of the uncommenting is intuitive, but the lack of symmetry in the comment closures requires more editing attention & makes it error-prone.  Have fixed it in both the CIP and CPS templates.

Crypto2099 (approved):
Looks good

polinavino:
Here is an idea of a usecase for intents : Light Clients that never query the ledger state, and instead use an intents API to describe what kind of transaction they want **constructed for them**. The API would have things like
- pay value from key to key
- mint an nft, etc.
- make a DEX trade/offer

1. An LC constructs an intent, sends it out to a service provider (SP) to fill.
2. The SP responds with a transaction that can be demonstrated to satisfy the intent, but is neither valid nor contains enough data for the LC to construct a valid transaction from it (in particular, it is **missing all output references** except one, which is from the SPs wallet). This transaction includes a payment to the SP for their service. 
3. The LC signs this transaction, and sends it back to the SP. 
4. The SP then constructs another transaction such that the two transactions together in a validation zone constitute a complete valid transaction (missing references are supplied), and sends this across the Cardano network. 

If this sounds interesting, I can elabourate.

rphair:
@polinavino yes, it sounds interesting!  Can you please re-post & elaborate in:
- https://github.com/cardano-foundation/CIPs/pull/779

and maybe also in:
- https://github.com/cardano-foundation/CIPs/pull/780

PR #799, opened 2024-04-20
CIP-0010 | Add Clarity (DAO creation) metadata label

t4ccer:
Added metadata label for DAO bootstrap metadata performed via [clarity.vote](https://www.clarity.vote/) platform

### DIFF
--- a/CIP-0010/registry.json
+++ b/CIP-0010/registry.json
@@ -1,5 +1,9 @@
 [
   {
+    "transaction_metadatum_label": 22,
+    "description": "Clarity - DAO creation metadata"
+  },
+  {
     "transaction_metadatum_label": 87,
     "description": "milkomeda.com - The protocol magic for the milkomeda protocol"
   },


rphair:
@t4ccer how would you justify claiming the low number `22` given that
- this would be by far the lowest number we have on the metadata label index (by a factor of 4)
- a general dispersion of these metadata labels through 16-bit / 4- or 5-digit numbers

?

t4ccer:
What part of CIP10 says anything about "low" numbers? 22 is not in 0 - 15 or 65536 - 131071 ranges that are marked as reserved in CIP10

rphair:
@t4ccer that's not an answer to my last question.  I could elaborate on it, but to put my own question differently: What is it about your own project that should make you "first" on the list (forever, even if your project discontinues) until someone chooses one of the 6 available integers before yours?

This could be your chance to talk about how relevant your project is to the community & especially if others will be following your metadata standard e.g. for other DAOs.  If this is already part your framework or its future plans then personally I would love to hear about it and it would be an excellent justification for such a memorable tag assignment.

None of this is required by the CIP and of course I'll pass it if you insist, and there are 2 other editors that could pass it without me even if I did object (which I haven't).  It IS within my purview, through, to ask _why_ you chose the number `22`.  In the meantime, instead of maintaining a defensive response, you could write a little about how such a significant number would be significant to the Cardano community.  Otherwise, it's a logical assumption by default that low number choices are self-promotional.

Ryun1 (approved):
Although I agree that it would be nice to know why `22`?
I don't think authors _have to_ justify why they have chosen x number

Crypto2099 (approved):
I will also say that I would like to know `"Why 22?"`

That said, the number is not within the reserved ranges and someone or something was bound to use low numbers like these eventually so... looks good to me.

PR #810, opened 2024-05-08
CIP-???? | X509 RBAC Registration (1/5) - Metadata Envelope

stevenj:
This is a draft PR for a standard to create an Envelope to secure Auxiliary Data and a Transaction to prevent replayability of a transaction's Auxiliary data and for an attempt to replay the data to be detectable off chain.

It is the first in a series of PR's detailing a x509 based Role-Based Access Control (RBAC) registration scheme for a dApp on Cardano.

---

([latest version in branch](https://github.com/input-output-hk/catalyst-CIPs/tree/x509-envelope-metadata/CIP-XXXX))


rphair (commented):
@stevenj impressive material as always - I'll make some general comments in this PR thread about the 5 simultaneous CIP submissions here:
- https://github.com/cardano-foundation/CIPs/pull/810
- https://github.com/cardano-foundation/CIPs/pull/811
- https://github.com/cardano-foundation/CIPs/pull/812
- https://github.com/cardano-foundation/CIPs/pull/813
- https://github.com/cardano-foundation/CIPs/pull/814

👉 Comparing them side by side is difficult while digging the `README`s out of your branches & we shouldn't have to work off of personal bookmarks.  Can you please add a link in your first comment above to each of these which points to either the current `README.md` file or directory (as we suggested in CIP-0001 that authors do for any proposal)?

👉 These titles are uncommonly long, so if keeping these CIPs distinct I'd suggest choosing shorter names by _both_:
- assigning a common name (like `Web-Wallet Bridge`) shorter than the current `X509 RBAC Registration (M/N)`... I would vote for removing `Registration` as arguably redundant (from the CIP title point of view).
- assigning shorter individual names for some of these proposals... e.g. `Signing X509 RBAC Transactions with CIP-30 Wallet Connect` puts it over twice the longest CIP title length.  

I appreciate the specificity but I believe we can preserve this while keeping more concise names not likely to develop duplicates over time.

👉 Generally we cannot be sure there will be only ever 5 of these proposals: however much it would look like this today.  Unless you've defined in the spec that these will be the only 5 CIPs on the subject... with future versions requiring new CIPs... we have to assume someone else might have a different or broader view of the scope & partitioning and therefore the qualifiers `(M/N)` should be dropped from the titles.

👉 I'm assuming you could well have practical reasons to split these into 5 CIPs, but if the only consideration is document length then I would also want to debate putting them all into a single CIP no matter what its length.

---

Discussion of the above points & triage of all these proposals at our CIP meeting next week (added: https://hackmd.io/@cip-editors/88) will be much easier if we can first get those `README` links into the original postings: a good habit for all CIP authors so the editors don't have to do it themselves. 🙏

CIP-XXXX/README.md
```diff
@@ -0,0 +1,235 @@
...
+Any dApp is free to have as many "purposes" defined as suits their goals.
+
+#### Key 1: txn-inputs-hash
+
+This is a hash of the transaction inputs field from the transaction this auxiliary data was originally attached to.
+This hash anchors the auxiliary data to a particular transaction, if the auxiliary data is moved to a new transaction
+it is impossible for the txn-inputs-hash to remain the same, as UTXO's can only be spent once.
+
+This is a key to preventing replayability of the metadata, but is not enough by itself as it needs to be able to be
+made immutable, such that any change to this field can be detected.
```
Crypto2099, line 85: Need to provide more explicit context for exactly how the hash of the inputs will be determined and then matched against the transaction in question to validate and verify.

* What type of hash is used (not otherwise specified)?
* How must one assemble the txn-inputs in order to create a canonical package suitable for hashing and validating?

Crypto2099:
Per @rphair point about the multiple submitted documents...

It would seem that proposals #1-4 all define (together) the standard and several, if not all of them, do not work when isolated out of scope and context of the larger payload. These should all be rolled together into a singular submission and proposal.

Proposal #5 seems to be specific to Project Catalyst implementation of this standard so may make sense to be separate.

However, in that case, this will create a precedent similar to what we've seen with CIP-67/68 where we end up needing a registry/repository entry as well as the standard definition and this should be considered and defined within the scope of the standard before progressing these.

rphair:
yes @Crypto2099 that's why I didn't speed ahead with the format reviews & at tonight's meeting we might confirm the need to combine these into 1 or 2 proposals; although @stevenj in the meantime I have had to take care of this issue, to speed up the introductory review at the CIP meeting:
> 👉 Comparing them side by side is difficult while digging the `README`s out of your branches & we shouldn't have to work off of personal bookmarks.  Can you please add a link in your first comment above to each of these which points to either the current `README.md` file or directory (as we suggested in CIP-0001 that authors do for any proposal)?

... it's always a nice touch with authors take care of this 🙏 https://github.com/cardano-foundation/CIPs/blob/master/CIP-0001/README.md#1a-authors-open-a-pull-request
> *Note* In the original comment for your pull request, please include a link to the `README.md` for the CIP in your working branch, so readers and reviewers can easily follow your work. If this link changes (e.g. from the CIP directory being renamed), please keep this link updated.

BTW @Ryun1 @Crypto2099 I should modify this language I added to CIP-0001 which isn't always correct... in some cases, as in these where there are asset files, it makes more sense to link to the containing directory than to the `README.md`.

Crypto2099:
@rphair I believe that even linking directly to the README.md file is sufficient as assets should usually use relative links to the .md file in most cases and should correctly populate in the "rendered" version so that is probably a minor point that's not worth introducing the additional complexity to concern ourselves with and we can handle on a case-by-case basis of PR's, particularly those that may be longer lived due to a lot of back/forth?

rphair:
@stevenj I'll follow the pattern of using this PR as a container for the others that were submitted concurrently.

At the CIP Editors' Meeting tonight we confirmed the feeling that combining the ones currently numbered 1 through 4 would produce a document without unnecessary cross references to explain different parts of the same system.  It might be different if any of these interdependent pieces could have applications in future CIPs, but it doesn't seem that way after a quick review by @Crypto2099 & me because the components all seem uniquely mated to each other.

There could conversely be a strong argument for keeping number 5 separate if it would be used more frequently by Catalyst designers to contain a "registry" that needed to be updated often... and therefore should be in a separate CIP from the description of the main X509 RBAC scheme.

In any case please consider rewriting as such as this material comes out of "draft" and @stevenj please post if you have any strong feelings to the contrary.  Considering the work is all currently on separate branches it would be best to resolve this question early in the editorial process... and to keep this review history I would support merging the content for at least proposals 2/4 through 4/4 into this branch.  (cc @Ryun1)

rphair (commented):
FYI @stevenj the application of `State:` label `Unconfirmed` to each of these 5 proposals is just to reflect that they haven't been given a candidate (potential) CIP number (yet).

There's been no call to tag these `Waiting for Author` yet, so in lieu of that:
* Could you update us here about what we are waiting for in these submissions?
* If still considered "Drafts", should they be marked as such on GitHub?
* Has your team considered the editor + community consensus suggestion that the first 4 symmetrical proposals be combined into a single proposal?

**p.s.** On an unrelated but perhaps encompassing note: Has there (ever) been any progress on the official enlistment (through a flagship CIP) of a `Catalyst` category as described (and currently void in the enlistment table) here? https://github.com/cardano-foundation/CIPs/blob/master/CIP-0001/README.md#categories

stevenj:
Hi @rphair We are in the process of implementing initial versions of theses CIPs to validate the approach.
After that process has completed, we will edit the Cips to incorporate any changes found to be necessary and also to take into account the feedback given so far.
Currently, they should be considered draft, and I can make the PR's draft if thats preferable.

rphair:
thanks for the update  @stevenj - yes it's preferable from the repository view, and sounds appropriate based both on your progress, to mark the whole set as `Draft` for now.

rphair:
@stevenj the CIP editors have been watching this series of 5 proposals (long suggested to combine into 1 or 2: https://github.com/cardano-foundation/CIPs/pull/810#pullrequestreview-2045269950) from afar for two years now & it definitely seems that neither the group nor any of its members will be progressing.

Collectively the ideas seem promising: perhaps in whatever form future Catalyst operations will take & maybe for electronic signatures on Cardano in general.  I believe you have a clear idea what kind of content would stand up to the CIP review process and so I am confident you can re-open any of these CIP PRs if & when its branch content contains something ready for peer review.

stevenj:
We were waiting on live validation and were actively refining it in the background.  But as the catalyst project as it was, is sadly no more, then yes closing these was the correct move.

PR #815, opened 2024-05-15
CIP-???? | Wallet best practices

Ryun1:
This proposal aims to describe a common set of best practices for Cardano wallet implementors, to improve security and user experience.

[📄 Rendered proposal on fork](https://github.com/Ryun1/CIPs/blob/cip-wallet-best-practices/cip-wallet-best-practices/README.md)

### DIFF
--- a/cip-wallet-best-practices/README.md
+++ b/cip-wallet-best-practices/README.md
@@ -31,6 +31,8 @@ The stakeholders of this proposal are wallet users and wallet implementors.
 
 ### Security
 
+By documenting 
+
 - documenting best practices ensures that all wallet implementors can at least be aware, first step to conforming
 - improve security baselines, no excuses
 - may help users choose *safer* wallets
@@ -43,15 +45,27 @@ The stakeholders of this proposal are wallet users and wallet implementors.
 
 ## Specification
 
+This specification aims to be a superset of best practices for all types of wallet.
+
 ### Meta
 
-- open source
-- should attempt to conform to standards/ formally describe
-- passcode protected
-- should allow users to do what they want, but can warn of dangerous situations
+Wallets SHOULD;
+- Open source
+- conform to accepted standards
+- use standard libraries
+- allow users to do what they want, but can warn of dangerous situations
 
-### Derivation/ Addresses
+### Censorship
 
+Wallets SHOULD;
+- allow users to do what they want, but can warn of dangerous situations
+- follow the permissionless philosophy of Cardano, and not censor.
+
+### Keys
+
+Wallets SHOULD;
+- encrypt secret keys behind a passcode, thus only being able to use the secret key when user provides passcode to decrypt.
+- never store the passcode in plain text, rather store a hash
 - single address wallets should be aware that other wallets are multi-address
 - be aware of relevant standards
 - show bech32 encoded at all times
@@ -59,6 +73,7 @@ The stakeholders of this proposal are wallet users and wallet implementors.
 
 ### Secrets
 
+Wallets SHOULD;
 - encrypted using user's passphrase
 - decrypted secrets are in-memory for the least possible amount of time
 - any references to secret objects should be overwritten with byte arrays with 0s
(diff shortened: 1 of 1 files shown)


cip-wallet-best-practices/README.md
```diff
@@ -0,0 +1,97 @@
+---
+CIP: ?
+Title: Wallet best practices
+Category: Wallets
+Status: Proposed
+Authors:
+    - Ryan Williams <ryan.williams@intersectmbo.org>
+Implementors: []
+Discussions:
+    - https://github.com/cardano-foundation/CIPs/pull/?
```
Ryun1, line 10:
```suggestion
    - https://github.com/cardano-foundation/CIPs/pull/815
```
add discussion link

Ryun1:
> Is this CIP attempting to identify specific CIPs which all wallets should follow for best practices or a parent CIP

The aim is more general, to provide a set of best practices for wallet developers.
Some of the best practices for navigating pitfalls/problems are already described by other CIPs, such as your CIP-111.

> attempting to identify common potential issues (if that's the best word) and request that the wallet implement some solution to that issue?

This is more my idea.
The bulk will be some low hanging fruit best practices that the vast majority of wallets are already doing.

> Would it make sense to basically say "here is a potential issue (malicious transaction chaining) and here is a list of solutions, at least one of which a wallet should implement"?

yeah this seems reasonable, as you can see its still quite a rough draft so some shaping like this might become clear as I (we? 🥺) work on it.

yHSJ:
> The aim is more general, to provide a set of best practices for wallet developers.
Some of the best practices for navigating pitfalls/problems are already described by other CIPs, such as your CIP-111.

OK, that makes a lot of sense. Then we should probably include some CIPs as potential solutions, rather than just state some basic behavior that should be included. That would make it easier to actually implement this CIP, no? It provides some framework for how to fix the issues, instead of just that the issues exist. Of course, a wallet could still provide their own solutions, I just think it's best to provide some reference solutions.

> yeah this seems reasonable, as you can see its still quite a rough draft so some shaping like this might become clear as I (we? 🥺) work on it.

I'd be happy to help out where I can. What is the best way for me to do that? Write up a small blurb about the malicious transaction chaining? Suggest changes? Something else? Let me know! :)

yHSJ:
> @Ryun1 @yHSJ I've been wondering from the beginning whether this would make more sense as a CPS, though happy to accept Ryan's presentation as a CIP especially if the recommendations end up being very specific...

My mental model of this proposal definitely doesn't perfectly fit in either bucket (CIP vs CPS). It feels like this document is really a collection of several CPSs, each with their own proposed solutions. Best practices, in general, are not "problems", they are a set of solutions to problems. And, generally a wallet could say "we follow the best practices", which feels a lot like implementing a CIP.

There doesn't seem to be a framework for this kind of grouping of problems by some logical groups. I think this is something that can be discussed by the working group and the CIP editors while this is being written, as you suggested. In the meantime, I think it makes sense for us to move forward and focus on writing each section in such a way it could also exist as a standalone document, to allow for future decisions to be made without causing a lot of extra work.

rphair:
@Ryun1 without movement for almost 2 years I've gone over this content again.  I'm sympathetic to the original goal of setting best practices for wallets (though it seems many of these points are common sense and/or common knowledge) but also agree with @yHSJ's https://github.com/cardano-foundation/CIPs/pull/815#issuecomment-2133957076 that these postings don't fit well with the CIP/CPS process.

Coincidentally & as a case in point, these difficulties were just recently recognised in a similar issue of Node design practice:
* https://github.com/cardano-foundation/CIPs/pull/1031#issuecomment-4381865627
* https://github.com/cardano-foundation/CIPs/issues/1145#issuecomment-4381778010

After 2 more years of trying to engage developer communities in the CIP process — especially with the great difficulty we've seen involving Wallet developers in standardisation issues over the course of this still-inconclusive effort:
* https://github.com/cardano-foundation/CIPs/issues/836

— I can't imagine a CIP expressed like this, even if perfected, becoming a rallying point for developers.  However I think a good home for this material might be somewhere in the design hierarchy of the Developer Portal, e.g. alongside this menu entry:
* https://developers.cardano.org/docs/build/integrate/wallet-authentication/overview/

... which takes content submissions through their own repository (cc @katomm @0xbora):
* https://github.com/cardano-foundation/developer-portal/tree/staging/docs/build/integrate

Anyway this is not an "executive" decision since you are equally well informed & permissioned to continue this PR as you see fit: so I feel comfortable leaving this in your hands if you would want to reopen & progress this somehow.

PR #827, opened 2024-05-22
CIP-0001 | Emphasise request to link PR OP to readable CIP

rphair:
Following up from https://github.com/cardano-foundation/CIPs/pull/810#issuecomment-2110422819 and based on the historical majority of CIPs not being submitted with authors linking a "Human readable version" in their original posting (OP) for the pull request.

We'd already added language for this (https://github.com/cardano-foundation/CIPs/blob/master/CIP-0001/README.md#1a-authors-open-a-pull-request) but it's been in a "Note:" which is apparently generally overlooked.  I've moved the text into a proper heading where it should stand out more, and emphasised it a bit.

Also I've added words that should leave authors free to link either to the CIP directory or its top level `README.md` as they see fit; following up on editors' comments made afterward.  In any case as long as a link is added then we save most of the trouble of editors having to maintain the readable proposal link.


Ryun1 (approved):
Nice addition!

PR #828, opened 2024-05-23
CIP-0010 | Add oracle tx metadata tag to registry

ross-spencer:
Provides a general purpose tag specifically for indexing and other uses off-chain by oracles and oracle consumers.

### DIFF
--- a/CIP-0010/registry.json
+++ b/CIP-0010/registry.json
@@ -52,6 +52,10 @@
     "description": "finitum.io token bridge transactions metadata"
   },
   {
+    "transaction_metadatum_label": 1226,
+    "description": "Oracle related transaction metadata"
+  },
+  {
     "transaction_metadatum_label": 1564,
     "description": "Marlowe - a DSL for writing and executing financial contracts"
   },


rphair (commented):
@ross-spencer this looks in order so far.  Do you have any documentation about your intended use of the tag?  (Mainly this is to help determine if the tag is just for a single project or if other oracle projects might share it.)

ross-spencer:
thanks @rphair! 

> this looks in order so far. Do you have any documentation about your intended use of the tag?

Nothing concrete to point to, but the idea behind this tag is for Cardano Oracles to make use of it for:

* ease of indexing oracle data
* contextualizing oracle datum
* providing additional user-facing context and information about how information was derived and put on-chain if desired.

It is largely in support of the audit functions surrounding oracles and oracle specific indexes and user-facing display mechanisms, e.g. oracle explorers.

It would be something all oracle services could make use of. Further filtering and disambiguation of those services would obviously be done via different policy ids and smart contract addresses.

rphair (approved):

Ryun1 (approved):

PR #444, opened 2023-01-20
CIP-0113? | Programmable token-like assets

michele-nuzzi:
**Developers wishing to participate**: please read & follow the **definitive recommendations at https://github.com/cardano-foundation/CIPs/pull/444#issuecomment-3638443804** which include:
* documentation = https://github.com/cardano-foundation/cip113-programmable-tokens
* reference implementation = https://github.com/input-output-hk/wsc-poc

For lightweight discussions prior to posting, feel free to use the [Discord CIP server channel `#cip-113`](https://discord.com/channels/971785110770831360/1448538251316035666) (server invite on [this page](https://github.com/cardano-foundation/CIPs/wiki#key-links--resources))

---

This CIP proposes a solution to [CPS-0003 (Smart Tokens)](https://github.com/cardano-foundation/CIPs/blob/master/CPS-0003/README.md) implementable already with V2 Plutus contracts.

If adopted as a standard it would allow to emulate the behavior of account-based ledgers tokens on Cardano.

---

([rendered proposal in branch](https://github.com/HarmonicLabs/CIPs/blob/master/CIP-0113/README.md))

PR #832, opened 2024-05-30
CIP-0125? | Arrestable native assets

micahkendall:
This proposal has overlap with the problem statement [CPS-3? Smart Tokens](https://github.com/cardano-foundation/CIPs/pull/382)

([rendered in branch](https://github.com/butaneprotocol/subledger/tree/master/CIP-SUBLEDGER))


micahkendall:
This is open to comment pre-implementation to get feedback on the high-level design.

micahkendall:
For the record, there are significant but maybe subtle differentiators from [CIP-0113](https://github.com/cardano-foundation/CIPs/pull/444) (including, differentiators from the most recent 'V2' which @matteocoppola clarified to me)
- this proposal has a single spending script hash, rather than many (however the implementation may use an additional withdrawal script for optimisation purposes)
- the owner is a field in the datum
- the implementation is key to the proposal, this is due to network effects as ideally many different seizable assets may be used together, this is best accomplished if they may be in the same utxo
- we don't try to do anything related to 'accounts' or ERC-20
- we specify the CIP more succinctly, because it doesn't try to generalise

michele-nuzzi:
This feels like a specific subset of [CIP-0113](https://github.com/cardano-foundation/CIPs/pull/444) 

I'm not saying it shouldn't be implemented, after all if we take Ethereum as an example they too have multiple standards for multiple specific token kinds (ERC20, ERC721, and so on...)

Before proceeding with making it a standard when should also consider the benefits of having a specialized one for seizure-specific assets.

I'll cover a few points mentioned about CIP-0113

> * this proposal has a single spending script hash, rather than many (however the implementation may use an additional withdrawal script for optimisation purposes)
> * the owner is a field in the datum

the "V2" of CIP113 is not final yet, in the sense that we are implementing it to see how it works.

if multiple scripts are a problem we can always move the parameters in a datum and we have a single script.

Multiple scripts were introduced to facilitate the offchain, as indexing based on the datum is still pretty limited.

> * the implementation is key to the proposal, this is due to network effects as ideally many different seizable assets may be used together, this is best accomplished if they may be in the same utxo
> * we don't try to do anything related to 'accounts' or ERC-20
> * we specify the CIP more succinctly, because it doesn't try to generalise

As you mentioned there are few benefits to having an ad-hoc standard, but we should also consider if it is worth the differentiation of assets

CIP-SUBLEDGER/README.md
```diff
@@ -0,0 +1,144 @@
+---
+CIP: ?
+Title: A Subledger in Plutus Enabling Native-Asset Seizure for Regulatory Compliance
```
rphair, line 3:
```suggestion
Title: A Plutus Subledger Enabling Native Asset Seizure
```
CIP editors & community reviewers had a meeting discussion this week about the suitability of "design ideas" for CIPs, since these tend to represent individualised methods rather than standards.  Therefore I think the otherwise wordy `A` beginning this title would help the community understand that this is only one of potentially many implementations of this general idea.

CIP-SUBLEDGER/README.md
```diff
@@ -0,0 +1,144 @@
...
+---
+
+# CIP \*\*\*: A Subledger in Plutus Enabling Native-Asset Seizure for Regulatory Compliance
+
+## Abstract
+We aim to establish a simple standard to ensure native assets can exist in a locked smart contract system, referred to as the "subledger". This subledger will enable functionality defined in the minting policy to seize assets.
+
+This design preserves the benefits of Cardano's native assets by enabling the reuse of most EUTxO patterns. It is significant that DEXs or other Plutus applications can be easily adapted to function within the subledger environment.
+
+# Motivation: why is this CIP necessary?
```
rphair, line 22: _This_ is the section there "Regulatory Compliance" should be mentioned as a principal motivation, rather than in the CIP title.

CIP-SUBLEDGER/README.md
```diff
@@ -0,0 +1,144 @@
...
+    - Micah Kendall <micah@butane.dev>
+Implementors: [Micah Kendall, <you?>]
+Discussions:
+    - https://github.com/cardano-foundation/CIPs/pull/?
+Created: 2024-05-30
+License: CC-BY-4.0
+---
+
+# CIP \*\*\*: A Subledger in Plutus Enabling Native-Asset Seizure for Regulatory Compliance
+
```
rphair, line 16:
```suggestion
```
There is no H1 element in the CIP outline since this is provided by the title in the YAML header.

CIP-SUBLEDGER/README.md
rphair, line 22:
```suggestion
## Motivation: why is this CIP necessary?
```
(correcting header level for standard CIP outline)

CIP-SUBLEDGER/README.md
```diff
@@ -0,0 +1,144 @@
...
+## Abstract
+We aim to establish a simple standard to ensure native assets can exist in a locked smart contract system, referred to as the "subledger". This subledger will enable functionality defined in the minting policy to seize assets.
+
+This design preserves the benefits of Cardano's native assets by enabling the reuse of most EUTxO patterns. It is significant that DEXs or other Plutus applications can be easily adapted to function within the subledger environment.
+
+# Motivation: why is this CIP necessary?
+
+A solution is required to enable asset seizure such as revoking USDC. We attempt to provide one in an EUTXO-harmonious style, for the sake of developer & user experience. 
+
+# Specification
```
rphair, line 26:
```suggestion
## Specification
```
(correcting header level for standard CIP outline)

CIP-SUBLEDGER/README.md
```diff
@@ -0,0 +1,144 @@
...
+## Objectives
```
rphair, line 27:
```suggestion
### Objectives
```
(correcting header level for standard CIP outline)

Likewise, please indent all further headers in your specification as necessary so that it's organised properly within the H2 _Specification_ until the next H2 section _Rationale_.

CIP-SUBLEDGER/README.md
```diff
@@ -0,0 +1,144 @@
...
+    input_oref: OutputReference,
+    assets: List<ByteArray>
+}
+```
+- Where assets represent the list of asset names being withdrawn.
+- The extra field is for arbitrary extra redeemer data
+- Extra field example: could be used for a signed payload if it's an admin endpoint using a key.
+
+## Development
+The primary improvement for developers is the ability to specify all these "capabilities to freeze/unfreeze" externally to the user-level experience of transferring the UTxOs.
```
rphair, line 94: This first of 3 items is apparently Motivation for the CIP and should be moved to that section...

CIP-SUBLEDGER/README.md
```diff
@@ -0,0 +1,144 @@
...
+
+This script can be implemented without any parameterization of UTxOs or network-specific values, allowing us to compile, deploy, and spend to the same script hash across mainnet, testnets, and emulators.
+
+For an enhanced developer experience, we should subject the implementation (in Aiken) to external analysis from the developer community before finalizing the exact instance to be enshrined in this CIP.
```
rphair, line 98: ... and these 2nd and 3rd items are apparently Rationale for these designs and would work well in that section.  In any case the non-standard `## Development` section should be flattened into these other standard sections.

CIP-SUBLEDGER/README.md
```diff
@@ -0,0 +1,144 @@
...
+
+## Rationale: how does this CIP achieve its goals?
+
+We accomplish the seizure capabilities on a technical basis, and the implementation style follows Keep-It-Simple-Stupid principles to minimise developer headache.
```
rphair, line 102: This is where I'm hoping you can compare/contrast with other potential "programmable asset" reference implementations and design choices like the ones @michele-nuzzi and predecessors have documented in the work-in-progress https://github.com/cardano-foundation/CIPs/pull/444.

rphair (approved):
@micahkendall we have to shrink the title for consistency with other CIPs, especially to remove words that will be implied to the reader and/or present as Motivation.  We can see how this works in the PR title & if you have an alternative recommendation let's try to keep the CIP + PR titles consistent.  (Note there is no hyphen in the phrase _native asset_.)

**Note** re: the compilation stub here: even though only currently redirecting to another repository, the CIP format might contain code samples and schemas but we have decided by consensus in the past _not_ to include source trees.  It would be sufficient in this case to include a link to the `butaneprotocol/subledger` repository (with some introduction if appropriate), and unless disputed your `/subledger` directory should be removed.

@michele-nuzzi's https://github.com/cardano-foundation/CIPs/pull/832#issuecomment-2139159659 suggesting this CIP is a subset of candidate CIP-0113's functionality is compelling, and might be a benefit to avoid fragmentation of these implementations as the community is trying to launch them in Cardano.  I would love to see some cooperation between you two & to see these proposals combined unless there are irreconcilable specifications that you want to pursue independently.

@micahkendall if you do choose to submit this as an independent proposal please
1. consider marking it as a solution to CPS-0003 as @michele-nuzzi has for his parallel proposal in https://github.com/cardano-foundation/CIPs/pull/444#issue-1550171726
2. please adopt the standard format corrections & editorial guidelines below.

micahkendall:
> @michele-nuzzi's [#832 (comment)](https://github.com/cardano-foundation/CIPs/pull/832#issuecomment-2139159659) suggesting this CIP is a subset of candidate CIP-0113's functionality is compelling, and might be a benefit to avoid fragmentation of these implementations as the community is trying to launch them in Cardano. I would love to see some cooperation between you two & to see these proposals combined unless there are irreconcilable specifications that you want to pursue independently.

I think there are significant differences in the CIPs, my goals around simplicity and compatibility with existing EUTXO patterns are not present in [CIP-0113](https://github.com/cardano-foundation/CIPs/pull/444). Furthermore on the topic of fragmentation, I think the CIP-0113 proposal is wholly incompatible with the goals of this CIP, so I think fragmentation is inevitable, whether both are accepted or not. I would also argue that my CIP is more favourable because it addresses the core regulatory issues for RWA, however does not enable any additional functionality outside of that scope, which would be appealing to the RWA ecosystem which may not want to break too far from the native asset standard.

One non-negotiable in my proposal is script hash per protocol: in my design, every asset class and user would share the same script hash, forming a closed but wholly interoperable system. An example of how this is functionally different, is that you can automate the transformation as described here:
- 1. take an existing smart contract system such as an AMM DEX
- 2. wrap the validators such:
    - the top-level validation is always a 2-arg validator
    - when there is a mint or certificate, normal logic except for the script context mapping
    - wrap the redeemer so that the difference between Spending, Spending&Withdrawing, Withdrawing can be branched
    - for each spend from the sub-ledger, invoke the validator in the wrapper
    - the context arg has it's properties mapped such that the input elements in the subledger and the output elements are mapped to be regular
- 3. send utxos such that the credential field in the datum is set to the new validator hash, instead of the address field of the output, locking them in the 'mapped validator'

This construction is meant to indicate more the power of maintaining EUTXO ideas, rather than as an actual purpose or goal of the CIP. It is merely emergent advantage of the CIP, not to be included in the CIP description.

> @micahkendall if you do choose to submit this as an independent proposal please
> 
> 1. consider marking it as a solution to CPS-0003 as @michele-nuzzi has for his parallel proposal in [CIP-0113? | Programmable token-like assets #444 (comment)](https://github.com/cardano-foundation/CIPs/pull/444#issue-1550171726)

I have included a link due to the overlap, however my proposal actually doesn't aim to address the same issues as CPS-0003. The express goal of that CPS is to control the native token lifecycle, which my proposal explicitly avoids, within the subledger you have full regular native asset functionality, such as sending to others without restriction and no ability to tax or validate any transfer or receipt of assets.

My proposal comes from the angle of approaching a very particular aspect of regulation, rather than any goal for more expressiveness in token programmability. The scope is very narrow, and the proposal is very simple, hence the **Path To Active** is much more achievable. Especially so when we consider the wide ecosystem changes that would be necessary.

> 2. please adopt the standard format corrections & editorial guidelines below.

Yes sir.

rphair:
@micahkendall thank you for clearing up my understanding on these points.  It sounds like it would suit the Cardano ecosystem well for these two CIP's to develop independently and I can understand also why we would want to keep it separate from CPS-0003.

I'd already added it to the agenda to introduce (not necessarily review) at the next CIP Editors' Meeting (https://hackmd.io/@cip-editors/90) in 1½ weeks and you are welcome to come along to help with that introduction.

@michele-nuzzi if you would like to take the opportunity to attend for a good comparison & contrast of your two proposals, please let me know and I'll also put https://github.com/cardano-foundation/CIPs/pull/444 on the agenda for a review (since it's been a while and we could use an update if you are ready).

micahkendall:
@rphair Here is the new [Compare and contrast](https://github.com/cardano-foundation/CIPs/blob/19b5180683ee5c1dacfc2a680f719ed7e8483dbe/CIP-SUBLEDGER/README.md#compare-and-contrast) section, for your enjoyment

rphair:
thanks @micahkendall ... yes, I think that's a good addition.

CIP-SUBLEDGER/README.md
```diff
@@ -0,0 +1,179 @@
+---
+CIP: ?
+Title: A Plutus Subledger Enabling Native Asset Seizure
+Status: Proposed
+Category: Ledger
+Authors:
+    - Micah Kendall <micah@butane.dev>
+Implementors: [Micah Kendall, <you?>]
+Discussions:
+    - https://github.com/cardano-foundation/CIPs/pull/?
```
Ryun1, line 10:
```suggestion
    - https://github.com/cardano-foundation/CIPs/pull/832
```

CIP-SUBLEDGER/README.md
```diff
@@ -0,0 +1,179 @@
+---
+CIP: ?
+Title: A Plutus Subledger Enabling Native Asset Seizure
```
Ryun1, line 3:
```suggestion
Title: Arrestable native assets
```
probably don't need to use such charged wording, I would suggest changing to something like this

CIP-SUBLEDGER/README.md
```diff
@@ -0,0 +1,179 @@
+---
+CIP: ?
```
rphair, line 2:
```suggestion
CIP: 125
```

CIP-SUBLEDGER/README.md
rphair: My apologies @micahkendall @Ryun1 but I hadn't seen this yet when adding the proposed CIP title in the process of assigning the CIP number.  I'll set it to Ryan's suggestion pending Micah's approval; I'd previously chosen to the terms `Subledger Enabling Native Asset Freeze` to try to satisfy reservations from the CIP meeting while preserving some of the terms already used in the CIP (though I think Ryan's wording is better since technically correct while not subject to bias).

rphair (commented):
@micahkendall I argued at the CIP meeting tonight that this should be assigned a CIP number so that this can better progress in parallel with #444 and to help potential implementors choose which model suits them best.  Your recent updates & comments for both CIPs have been very helpful with that.  cc @michele-nuzzi

@Ryun1's earlier suggestions for the CIP title were based on suggestions from the meeting that:
- the term `Seizure` suggests an enthusiasm for doing so (rather than your stated modest goal of regulatory compliance)
- `Plutus subledger` is not "a thing" and neither really is `subledger` so to use those terms implies their usefulness in technical reference (currently none).

Please also change the containing directory to `CIP-0125`. 🎉

CIP-SUBLEDGER/README.md
rphair: p.s. this comment was entered with https://github.com/cardano-foundation/CIPs/pull/832#pullrequestreview-2111101780 but for some reason GitHub did not preserve it that way.  cc @micahkendall

rphair:
p.s. to https://github.com/cardano-foundation/CIPs/pull/832#pullrequestreview-2111101780 - @micahkendall after waiting for any disputes, I've recorded it in the CIP candidates list via https://github.com/cardano-foundation/CIPs/pull/844 with the most recent title correction.  If any further corrections please post in this thread and/or change in the branch source CIP title... otherwise please change the CIP title to match the PR title.

micahkendall:
> p.s. to [#832 (review)](https://github.com/cardano-foundation/CIPs/pull/832#pullrequestreview-2111101780) - @micahkendall after waiting for any disputes, I've recorded it in the CIP candidates list via #844 with the most recent title correction. If any further corrections please post in this thread and/or change in the branch source CIP title... otherwise please change the CIP title to match the PR title.

Hey cheers. The requested changes are mostly wording so I don't really care whether they are protested or not. Whether the CIP is enthusiastic or not doesn't meaningfully affect usage.

I would argue the term "subledger" is useful as a coined term because of how it can be useful for abstraction. From the CIP's glossary:
"A subset of the main ledger that tracks specific types of transactions or assets, often with additional rules or functionality."

CIP-SUBLEDGER/README.md
```diff
@@ -0,0 +1,179 @@
+---
+CIP: 125
+Title: Arrestable native assets
+Status: Proposed
+Category: Ledger
+Authors:
+    - Micah Kendall <micah@butane.dev>
+Implementors: [Micah Kendall, <you?>]
```
rphair, line 8:
```suggestion
Implementors: []
```
The implementors are the empty set for now; this set is understood to include any potential `<you?>` and no doubt you will be adding yourself (before or after merge) once you have the proposed Aiken implementation.

rphair (commented):
> I would argue the term "subledger" is useful as a coined term because of how it can be useful for abstraction

sure @micahkendall let's keep this term in mind whenever the question of the title comes up... I'll suggest reintroducing it myself if discussion indicates it would help with the abstraction.

CIP-SUBLEDGER/README.md
micahkendall: Thank you.

CIP-SUBLEDGER/README.md
```diff
@@ -0,0 +1,179 @@
+---
+CIP: 125
+Title: Arrestable native assets
+Status: Proposed
+Category: Ledger
```
rphair, line 5:
```suggestion
Category: Tokens
```

rphair (commented):
@micahkendall I noticed this PR wasn't categorised & also that the CIP was given a `Ledger` category.  That would only be appropriate if changes to the Ledger were going to be required — which doesn't seem so in this case — and since the software device you are calling a "subledger" interacts with native assets, I believe (pending possible further discussion) that this is really a `Tokens` category proposal:

rphair:
@micahkendall is this work by any chance also being absorbed into
* https://github.com/cardano-foundation/CIPs/pull/444

the way that
* https://github.com/cardano-foundation/CIPs/pull/944

has been?  It's been a long time since this proposal was updated (so marking `Waiting for Author` as an alternative to closing it as `Abandoned`) but perhaps, while candidate CIP-0113 is headed through validating its reference implementation towards confirmation & merging, some cooperation with that team would be beneficial?

In any case please let us know either where this is going or if you have discontinued work on this potential standard.

micahkendall:
Cheers for the follow up, don’t think the proposal is covered by any of the above but I don’t have time to push it through

rphair:
thanks @micahkendall for the update & for your ideas so far.  I'll give it a brief cooling-off period and then remove this candidate CIP link from [CPS-0003](https://github.com/cardano-foundation/CIPs/tree/master/CPS-0003) unless there is any objection.  cc @Ryun1 @Crypto2099 @perturbing

colll78:
I think CIP-113 after updating to CIP-143 does indeed cover all the proposed value of this CIP:

> this CIP maintains the same script hash across all UTxOs, hence saving budget, enabling programmability, allowing ease of interoperability, because we use prefixing to identify assets which are locked to the subledger, we may not only have one script hash across the same asset class, but across all asset classes (I believe making this the favourable CIP for DEX integration) again, because we use prefixing, you may put non-seizable assets into the subledger, and then withdraw them again, so no wrapping is necessary for example to swap $BTN (true native, unseizable) and $USDC (subledger native, seizable), because a subledger DEX can simply operate with both in one UTxO

CIP-143 addressed this. Now, all programmable tokens live at the same script hash. CIP-143 also is designed around the concept of a mini-ledger (referred to as subledger in this CIP) and allows non-seizable / non-programmable assets to enter and exit the programmable tokens mini-ledger. No wrapping is necessary.

> native assets are still used

CIP-143 also uses native assets, not virtual assets. 

> this CIP isn't a standard of implementations, it aims to provide a single implementation, to minimise fragmentation

CIP-143 also is not a standard for implementations, it is a single implementation of a mini-ledger for programmable tokens within which all programmable tokens must exist. Programmable tokens can never exit this mini-ledger, nor can they be minted outside of it.

> no account abstractions

CIP-143 (and thus CIP-113) doesn't use account abstraction.

> narrow scope, enabling a faster path to active.

CIP-143 is active on mainnet, with a minting platform and a management UI platform for viewing balances, freezing, seizing, and making transfers.

PR #833, opened 2024-06-01
CIP-0010 | Misspelled keyword for 1694 label

Godspeed-exe:
Noticed one entry had a typo, decided to fix it.

### DIFF
--- a/CIP-0010/registry.json
+++ b/CIP-0010/registry.json
@@ -64,7 +64,7 @@
     "description": "CIP-72: dApp registration & Discovery"
   },
   {
-    "transaction_metadataum_label": 1694,
+    "transaction_metadatum_label": 1694,
     "description": "Voltaire - additional metadata for governance, provided in a separate transaction"
   },
   {


rphair (approved):
thanks; have checked rest of file for similar keyword misspellings & found none.

Godspeed-exe:
@rphair probably not the right place for this discussion / question, but what's your opinion of adding a "cddl" attribute to the CIP10 registry which would be an array of CDDL definitions locations. 

CIP25 is used for metadata label 721, but there is no easy way to go from CIP10 registry JSON to CIP25 CDDL files - if you don't know where to look. So this CDDL definition would make that link easy / direct.. 

I'm building something that works with these metadata labels and I'm looking to automatically parse / validate the payload with their intended metadata schema. I was actually super glad that Ales decided to create those for CIP25 - very forward thinking!

Curious on your thoughts!

Ryun1 (approved):

rphair:
@Godspeed-exe I don't have experience with these implementations (@Crypto2099 would be better suited to this) but can see how what you're suggesting might be useful:
- If you feel confident that a specific format of the CIP-0010 registry would be useful for this, please submit a PR when ready against CIP-0010 with an update to the registry schema & explanation of how to use it.
- If not yet confident or just want to discuss the idea, please submit an issue on this repository and I'll add it to the agenda of whichever CIP Editors' meeting comes up next (and will tag you in case you want to be present for that discussion).

PR #834, opened 2024-06-03
CIP-1694 | Reword note on info actions

sgillespie:
I was trying to test ratification of info actions, before I was told that info actions can never be ratified or enacted. This change is intended to clarify that point.


Ryun1 (approved):
Nice edit to add clarity, thank you @sgillespie 💪

rphair (approved):
Seems like an straightforward improvement.

Issue #783, opened 2024-03-14
CIP 100 | Provide directions on how to create signatures for the body without circular dependencies.


kderme:
CIP 100 provides an algorithm to find and hash the body of governance metadata
```
Canonicalize the whole document according to [this](https://w3c-ccg.github.io/rdf-dataset-canonicalization/spec/) specification.
Identify the node-ID of the body node
Filter the canonicalized document to include the body node, and all its descendents
Ensure the file ends in a newline
Hash the resulting file with blake2b-256
```

However this assumes that in order to hash the body (last step), you first need to canonicalize the whole document (first step). This creates a circular dependencies problem, since the whole document already contains the signatures.

Ryun1:
> This creates a circular dependencies problem, since the whole document already contains the signatures.

Agreed this isnt clear.

But I dont think there is a dependency;
Since you filter out the signatures via; `Filter the canonicalized document to include the body node, and all its descendents`
you can just have placeholders, until you are able to actually build signatures with the hash digest as the payload

kderme:
The canonicalization algorithm depends in some way in sorting the generated N-Quads and also in the number of these N-Quads. Without knowing at least the context of the whole text it's not possible to have placeholders, because it's impossible to know keys that can be sorted before or after the body.

I must say my understanding of the [algorithm](https://w3c.github.io/rdf-canon/spec/) is limited to reach safe conclusions.

Quantumplation:
You know all of the keys ahead of time, so you should be able to generate all of the N-Quads. You just don't know the values for those keys.

kderme:
> You know all of the keys ahead of time

So each author is expected to know the total number authors and other information to predict the correct node-id of the body? It's a bit counter-intuitive that data that live outside the body actually affect the normal form of the body and as a result the signature.

Quantumplation:
If you have an alternative suggestion, this is one of the things we struggled with, there's not really a good solution; just taking the `body` for canonicalization doesn't quite work because it doesn't have the context, so your have to modify the body to inject a context, etc.

kderme:
Sorry for the late response, my suggestion would be to sign the body content as is, or follow the much simpler json canonicalisation algorithm. The json-ld canonicalisation algorithm has no protection over colision attacks, so even after fixing the circular logic problem, it's always possible for 2 different json-ld files to have the same canonical form. For example all 3 texts below have the same canonized format 

[...]

PR #796, opened 2024-04-19
CIP-0120? | Constitution specification

Ryun1:
This proposal aims to describe a standardized technical form for the Cardano constitution.
Hoping to enhance the accessibility and safety of the document.

---

[📄 Rendered proposal](https://github.com/Ryun1/CIPs/tree/cip-constitution-tech/CIP-0120)

PR #835, opened 2024-06-03
CIP-0100, CIP-0108, CIP-0119 | Hash raw content directly

kderme:
This pr tries to simplify and improve CIP 100 and its follow-ups 108 and 119, by avoiding the canonicalisation step when hashing the contents of the metadata. It is the result of a discussion [here](https://github.com/cardano-foundation/CIPs/issues/783) and [here](https://github.com/cardano-foundation/CIPs/issues/783#issuecomment-2094232404)

Even though the canonicalisation algorithm has been introduced to avoid ambiquities related to different json parsing, in fact it tries to solve a problem that it itself creates: Without the need to canonicalise the content, we don't even need to parse the content as json during validation.

Some other advantages of this change:
- it reduces the steps needed to create and validate the hash
- it follows the existing paradigm set by pool off-chain metadata
- It make CIP-100 a natural extention of CIP-1694. Application that only aim to be compatible with CIP-1694 and do a lightweight validation of the offchain data are not directly incompatible with CIP-100
- it avoids issues of the json-ld canonicalisation algorithm mentioned [here](https://github.com/cardano-foundation/CIPs/issues/783#issuecomment-2069660006)

Note that this dependency is not fully eliminated with this pr, since it still exists in the body signing algorithm. For this a separate pr will be created


Sworzen1:
Does hashing uncanonized jsons cause problems with comparing hashes and exchanging data between different systems ?

gitmachtl:
> Does hashing uncanonized jsons cause problems with comparing hashes and exchanging data between different systems ?

No, because its just handled like a binary file for the hashing. Like we do with all the other files that we hash to calculate a uniq stamp/identifier. A hash should be calculated on the raw binary version of a file IMO, so i fully support this PR.

Sworzen1:
Look on example in this RP: We have example.json and hash from raw file -> 04af1b48bccbf7cf9b3e3b7952dfbdde0cc851ccb87ae6643521672cc381b00d.
I replace the data in file, for example -> move body propertie  "comment" with value "This is a test vector for CIP-100" before "references" propertie. Result is -> the data is the same but hash is totaly different -> 5e40a477b7fd2f195371eb440063e0823b53fdefc2189a85a887c2b49746f29d

Sworzen1:
I do not agree with you. The data is the same, only a structure of the file changes. In my opinion, we should check the user data, because, the user data is most important, not file.

x = y;
y = x;
Its the same data man.

gitmachtl:
> I don't agree with you. The data is the same, only a structure of the file changes. In my opinion, we should check the user data, because, the user data is most important, not file.
> 
> x = y; y = x; Its the same data man.
> 
> Additionally, different languages, applications and systems may format files differently, but a data remains the same no matter how a file is read.

Well, i agree to disagree with you. :-) It makes it unnecessary complicated to check data integrity for tools. Is there a universal tool/lib available for all 3rd party applications? There is one which is doing just a blake2 hashing on binary data, thats an easy pick.

Sworzen1:
Yes, blake2 is great. We need to have the data in order, which is provided by canonicalising the json as described so far. Otherwise, the structure of the files matters, and as I wrote above, files can be interpreted in any way, which makes reordering the data affect hash validation.

gitmachtl:
Than why going with jsonld in the first place? We have a CDDL for every certificate, etc. on cardano. Why not directly going with CBOR binary data?

Sworzen1:
I think I didnt understand you. Please describe me which problem solve CBOR representation of json ?, is the same that but in another representation for optimization.

gitmachtl:
With all the tools i built like cardano-signer, i was coming from the CDDL cbor side. Where you have a numeric map entry for a specific data and thats it. These methods now trying to convert data over from readable text in any sorting order and charset to be converted to end up again like a cbor in canonilized format like we use all he time on cardano.

Sworzen1:
Yes  - we are using jsond-ld lib in JavaScript to canonization and URDNA2015 algorithm as default. Im using it in GovTool because of CIP-100 requirement. I dont know why - about "Than why going with jsonld in the first place?".  Naturally, if you have a better tool that will check the correctness of the data, I will be happy to check it. I know what cbor is and what parsing looks like, but without examples to your solution, I don't really know what you are talking about in the previous comment. Only thing I can comment on is the changes made to this PR, which I don't like very much for the reasons described above.

kderme:
> The data is the same, only a structure of the file changes. In my opinion, we should check the user data, because, the user data is most important, not file.
> 
> x = y; y = x; Its the same data man.

The example you provide contains the effect of an insignificant change to a file. However, at https://github.com/cardano-foundation/CIPs/issues/783#issuecomment-2069660006 I've provided examples that are not that harmless. They show how the content of a file can be maliciously manipulated and have a different meaning, while maintaining the same hash. A good hashing function is collision resistant, which URDNA2015 is not.

> Additionally, different languages, applications and systems may format files differently, but a data remains the same no matter how a file is read.

The only requirement is that the editor of the file has control over its content, which is a very reasonable requirement. Note that no json parser is involved in the validation, so there is nothing that can cause ambiguity.

> If it gets implemented within cardano-cli including the signing/verification part, i can update the spo scripts to reflect that change.

Afaik cli doesn't plan to support URDNA2015

disassembler:
I'm in agreement with @kderme and @gitmachtl . We don't want to canonicalize json because it can cause collisions. This was a major point @dcoutts made when we stopped using canonical json after Byron. It's a security risk to massage data before hashing.

gitmachtl:
may i ask a question: if the goal is to have a signature proof. why can't we just hash the binary file as it is. sign this hash with the wanted key - like the drep signing key - and put this information within the transaction aside with the action itself? can be added via metadata f.e. like we are doing catalyst registrations etc. this can all be done via cip8/30 message signing now. also hardware wallets support this with the upcoming firmware. 

using cose_sign1 format again would simplify it also for 3rd party tools.

gitmachtl:
oh and btw, hardware-wallets can't just sign a message or a hash. it must be embedded within the CIP8 structure. so relying on the function that signing keys needs to just sing a hash would not work for hardware-wallets. which would make them obsolete for dreps and committee members in combination with governance metadata signing.

gitmachtl:
about the anchor data-hash issue for conway ... can we settle down on the fact that the hash on the chain which is linked to a file should be the hash of the file and not some content hashing stuff? checking the validity of the content of a file has nothing to do with the filehash itself. to make a proof about the content can be done afterwards imo.

Sworzen1:
Yes no problem. The important thing is, that we want to check the correctness of the data or only file. If in fact it is enough to check the file, I agree. However, if the correctness and order of the data is important, another mechanism would be useful.

rphair:
We need to commit to a decision so related projects can decide how to proceed, so this will be top priority for final review at the CIP meeting tonight (https://hackmd.io/@cip-editors/91).

Thanks @kderme + @gitmachtl @Sworzen1 for your presentations so far... please come to the meeting to contribute to the discussion, with the intention of deciding there (cc @Quantumplation @Thomas-Upfield).

Note @Ryun1 for future reference this decision could also apply to the proposed method of constitution hashing here:

- https://github.com/cardano-foundation/CIPs/pull/796

cc @disassembler @Crypto2099

rphair (approved):
CIP meeting in last hour has reviewed the rationale for making this change & agreed without dispute to go ahead with the changes exactly as proposed.

Crypto2099 (approved):
The rationale for maintaining consistency and compatibility with downstream tooling such as db-sync make this a logical modification that I'm happy to see us get implemented and standardized before we move into the Conway ledger era.

PR #837, opened 2024-06-06
CIP-0010 | Validation by GitHub workflow

Godspeed-exe:
Hey guys,

I took some inspiration from the Cardano Token Registry Github repo and implemented a Github workflow which:
- on every PR will run automatically
- will validate the structure of the JSON schema (which I simplified as well).

You can check out how this works on my fork:
- https://github.com/Godspeed-exe/CIPs/pull/5 > this PR contains a broken entry in the registry.json file and the workflow is clearly showing this at the bottom of the PR.
- https://github.com/Godspeed-exe/CIPs/pull/8 > this PR does not contain any changes to registry.json and does not trigger the workflow
- https://github.com/Godspeed-exe/CIPs/pull/4 > this PR contains a valid new entry and passes the workflow.

Let me know if you have any questions / remarks / suggestions!


rphair:
thanks for your effort @Godspeed-exe - we can talk about this potential change at the next CIP meeting (I've put it on the agenda: https://hackmd.io/@cip-editors/90).

Processing these PRs is a low bandwidth task for editors and the textural regularity of entries does tend to make errors easily spotted by eye.  My main discussion point for other editors would therefore be, "Do we need this?"

Godspeed-exe:
No worries @rphair 

Recent events would argue that some of these errors do slip by occasionally and go unnoticed for months (very subtle mistakes like `metadautum`) and having this automated check would severely decrease chances of this happening again. 

I'm aiming / hoping to create many more standards in the upcoming future so I do expect this to be growing. From Enterprise Technology side we stress for registration and documentation of these standards. Good documentation is (as we know) key for growth and adoption.

I'm also building a component which would make pushing metadata to Cardano way more accessible for SME's and this is also syncing / checking with CIP-0010 for registration and preferably also using the JSON schema validators - otherwise people will be pushing useless rubbish. 

Thanks for the consideration!

Godspeed-exe:
It would also make it automagically visible to the person creating the PR
that this fails..

Win/win 😁



*Michiel Bellen*
Core Integrations Team Lead | Cardano Foundation

Website: www.cardanofoundation.org
Book a meeting <https://meetings-eu1.hubspot.com/michiel-bellen>

Crypto2099:
I like this idea, particularly for CIPs that use JSON or similar for "registries" to ensure that all changes are already 'correct' without the need for manual human intervention.

rphair (commented):
@Godspeed-exe we just discussed this at the CIP editors' meeting and for a couple reasons I wanted to retract my earlier lukewarm comment:
- this material sets a precedent & an initial test case for validating CIP changes in general... there are other validations like file & directory names which checks for in advance would help authors & editors avoid inappropriate merges.
- other registries besides CIP-0010 will have more complicated schema & syntax and so it would help to test on the much simpler case of CIP-0010 first.
- @Crypto2099 pointed out we'll also have Internet standardised CBOR tags in the registry, so integration with outside sites may be an issue going forward and we will need automation & validation for these.

So after a preliminary test & verification in an editor's fork is done, I would be in favour of merging this when we have an additional verification that it will perform as expected. 😎

Godspeed-exe:
@rphair thanks for the update, let me know if you have any questions or if anything needs to be improved! 💪

CIP-0010/registry.schema.json
```diff
@@ -1,64 +1,25 @@
 {
-    "$schema": "http://json-schema.org/draft-07/schema",
-    "$id": "https://github.com/cardano-foundation/CIPs/blob/master/CIP-0010",
-    "type": "array",
```
Ryun1, line 4: why are we removing things from the schema?

Ryun1: could removing this break anyone existing implmentations?

Godspeed-exe: Impossible to know I think? I don't think we have access to statistics on the usage of this file?
But as mentioned, there is no need to do this change, the main reason for this is to add the validation step to prevent breaking changes.

Crypto2099 (approved):
All of these changes look good to me and I think this one is ready for a real-world test.

CIP-0010/registry.schema.json
rphair: @Ryun1 @Godspeed-exe we are confirming now in CIP meeting that @Crypto2099 has assessed the consequences to existing implementations and ruled them out (my words, not his).

rphair (approved):
Willing to give this a try now that schema question has been resolved (https://github.com/cardano-foundation/CIPs/pull/837#discussion_r1636052192).

CIP-0010/registry.schema.json
```diff
@@ -1,64 +1,25 @@
...
+        "description": {
+          "type": "string",
+          "maxLength": 256
+        }
+      },
+      "required": [
+        "transaction_metadatum_label",
+        "description"
+      ],
+      "additionalProperties": false
```
Ryun1, line 23: Is there rationale for changing this?

CIP-0010/registry.schema.json
Ryun1: Understood thanks @Godspeed-exe

CIP-0010/registry.schema.json
Crypto2099: This actually makes sense in my opinion. We do NOT want potentially malicious or harmful content to be added to this registry file (like people advertising their URLs, etc) without at least needing to pass through the lens of the CIP Editor body and potentially an update being issued to the schema itself.

I'm in favor of leaving it so that any and all deviations from the schema get flagged and we can decide on them on a case-by-case basis.

Ryun1: Yup agreed!
appreciate the explanation

PR #842, opened 2024-06-16
CIP-???? | Cardano URIs - Authentication

Crypto2099:
[Rendered Version in  Branch](https://github.com/Crypto2099/CIPs/tree/web-cardano-auth-uris/CIP-XXXX)

This proposal aims to create a mobile-friendly URI standard for wallet authentication that can be used in a variety of situations.

### DIFF
--- a/CIP-XXXX/README.md
+++ b/CIP-XXXX/README.md
@@ -1,0 +1,98 @@
+---
+CIP: ????
+Title: Cardano URIs - Authentication
+Category: Wallets
+Status: Proposed
+Authors:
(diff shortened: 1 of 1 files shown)


rphair:
@Crypto2099 thanks for introducing this idea & it's only now (after almost 2 years) that I've noticed a Specification was never introduced for this.  I still think the Wallet space would have a need for this & would look forward to seeing it reopened (or the idea revisited with a CIP PR from another developer) if & when a properly specced CIP can be submitted.

Issue #836, opened 2024-06-04
CIP-0013: Current state of integration and further advancements


MadOrkestra:
With mobile wallets showing great progress over the past year, I'd like to bring some attention to CIP-0013 (Cardano URI Scheme), which makes quick payments by scanning a QR Code possible - a feature greatly needed to get a lot of use cases off the ground and improve the overall user experience on Cardano.

CIP-0013 was first proposed in 2020 but doesn't seem to have moved along since. There seems to be great support for CIP-0099 (Proof of Onboarding), which already is an addition to the non-active CIP-0013. Flint Wallet seems to have integrated their own deeplink solution already, so before other wallets go down this individual path, maybe lets have a discussion about what is missing and how to proceed, so we can put this CIP into action. I had great success working with BeginWallet on this for the past week, they will role out payment link integration shortly.

### Current state of Integration in mobile wallets

https://cip13.cardanothings.io/

### Advancements

From my conversations with multiple devs and projects, there seems to be a need for at least two improvements:

**1. Asset support for payment links**

Right now, there is no way to include Cardano native assets in a transaction, just ADA. Especially with stable coins being a thing now, this seems to need integrating. Or: Deciding to put this CIP live as is and have improvements in own CIPs like it was done with CIP99, also a way to go.

**2. Metadata support for payment links**

To match transactions on the backend of a payment system, there is the need to include some kind of metadata in payment links. This could already be achieved by adding a recognisable amount of Lovelace to the payable amount, but this is of cause more of a workaround than a solution. So it stands to reason to add functionality to include metadata in payment links 

_I will invite devs and wallets to join this discussion. Please spread the word, so we can hopefully work out the missing pieces and finally put this CIP into action._

### 🚀 2025 Update on current progress / further advancements
I'll leave these here for anyone joining the party. Currently in the works are the following advancements:
- https://github.com/cardano-foundation/CIPs/pull/843
- https://github.com/cardano-foundation/CIPs/pull/1058
- https://github.com/cardano-foundation/CIPs/pull/1069

rphair:
re: future plans for wallets... cc @Crypto2099

@MadOrkestra we had a brief review of some of these topics when I last rewrote CIP-0013 to have more flexibility for future protocols to be added.  Additional `//` schemes were suggested in this period for metadata and other chain queries (see https://github.com/cardano-foundation/CIPs/pull/625 - cc @Ryun1 @klntsky).

[...]

PR #843, opened 2024-06-16
CIP-0157? | Cardano URIs - Enhanced Payments

Crypto2099:
[Rendered Version](https://github.com/Crypto2099/CIPs/tree/cardano-uris-enhanced-payment/CIP-XXXX)

This proposal seeks to extend the CIP-13 payment URIs and address some of the issues raised here: https://github.com/cardano-foundation/CIPs/issues/836

We need an enhanced and more explicit `Payment` type of Cardano URI in the modern era of native assets, smart contracts, and transactional metadata.

Because of legacy integrations with existing URI payments (lacking a dedicated authority) I have chosen to break this extension out into a new CIP with an explicit authority to avoid having to change or update CIP-13.


rphair:
@Crypto2099 good to have a stub here for the current discussion of how a comprehensive and collision-free payment URI syntax would look.  Cross-referencing my https://github.com/cardano-foundation/CIPs/issues/836#issuecomment-2156442263 and https://github.com/cardano-foundation/CIPs/issues/836#issuecomment-2147763143 to be sure scenarios involving these boundary cases will be handled properly by any proposed syntax:
- keyword(s) used to specify metadata collide with token name(s)

... including these cases if payment handles are supported by this protocol (I would insist that they should be):
- payment handles (if supported by this protocol) are ambiguous as per https://github.com/cardano-foundation/CIPs/pull/605 (cc @SIDANWhatever)
- keyword(s) used to disambiguate payment handles collide with token name(s)
- payment handles collide with bare payment addresses (unlikely but possible)

Crypto2099:
Thanks @rphair excited to start the conversation and hoping to have some first-draft specs delivered over the course of the next week for discussion! These are great starting points for consideration though!

rphair:
you're welcome @Crypto2099 & if you feel potential specs could use community brainstorming then feel free to add it to the agenda in a week-plus time: https://hackmd.io/@cip-editors/91

MadOrkestra:
Quick heads up: beginWallet implemented a "CIP13-like" solution now (https://x.com/BeginWallet/status/1941614893691724100) - so I'll leave this here as starting point for a discussion to turn this into a standard CIP addressing at least _some_ of the downfalls of the original payment scheme.

the beginWallet URI scheme @francisluz already implemented looks like this:
`web+cardano:addr?amount=1&token=ADA&paymentID=123-ABC&note=abcdedafad`

**What it solves**
- Payments with other tokens than ADA (think USDM)
- Linking payments to offchain data (to link a payment to a shopping cart of whatever)

**What it doesn't solve / Downfalls**
- Transactions with multiple tokens
- Uses a ticker symbol instead of a policyId
- Not using the //payment authority

I think the biggest issue here is the ticker symbol and the centralized whitelists that have to come with it to convert the ticker into a policyId to build the transaction. But: I think we have to abandon the original idea of multiple assets to fit into a payment URL anyways, because while URLs can be quite lengthy today in most modern browsers (up to 10,000 characters or something), the bottleneck is the QR-Code with "only" 300-500 characters before the "scanability" decreases massively, which especially for the use case of providing a slick and quick payment method is a no-go.

If we were to go with policyIds instead of ticker symbols, the overall character length excluding the note/msg field would still perfectly fit within these limitations: 
- Identifier, params and stuff: 43+ Characters
- Amount: 1 to whatever (100bn are still only 12 digits)
- Bech32 Address: 104 Characters 
- PolicyId: 56 Characters
- PaymentID/Note: 32-36 Characters (for UUIDv4 e.g., could be shorter or longer ofc)

I think the dedicated paymentId field is debatable, as long as you can put _some_ metadata into a transaction via URL, we cover a lot of use cases.
 
Given that we've spent another year or so without proper definition of this very rudimentary functionality on Cardano, I'd rather see this moving quickly now and enabling stable coins or Hosky or whatever as possible payment method across the ecosystem instead of looking for the perfect solution while seeing wallets coming up with their own individual solutions, but thats just my take ofc.

rphair:
thanks @MadOrkestra - this covers everything I would have said.  Just one clarification though (cc @francisluz):
> **What it solves**
> - Payments with other tokens than ADA

When paying a non-Ada token, is the minimum amount of ada (to carry the tokens) implicit in the transfer?

francisluz:
Hi @rphair,

Yes, in our case it’s implicit transfer with the min ADA, we actually have
to make a improvement to it to show as part for the fee or something like
that.

Cheers.

Enviado via iPhone

rphair:
p.s. regarding the "unresolved" points (we can [discuss in person tomorrow](https://github.com/cardano-foundation/CIPs/issues/836#issuecomment-3046814048)):
> **What it doesn't solve / Downfalls**
> - Transactions with multiple tokens
> - Uses a ticker symbol instead of a policyId

These 2 points go together.  The discussion I raised in https://github.com/cardano-foundation/CIPs/issues/836#issuecomment-2147763143 about a general design to include multiple native token transfers could be trumped (after another year of scams) by some practical insecurity of using ticker symbols:
* We would have to assume, until demonstrated otherwise, that fake tokens with the same ticker as real ones might be used to spoof users through such a payment URL scheme.
* Even without the threat of malicious usage, we would still need a rigorously defined selection process to determine an exact Policy ID from a ticker name.

Such improvements, if we can make them, would provide a way around the URL length problem and fit poly-token transfer URLs into QR codes.

> - Not using the //payment authority

This alone currently makes the Begin Wallet / @francisluz solution outlined in https://github.com/cardano-foundation/CIPs/pull/843#issuecomment-3046101873 non-CIP-0013-compliant.  "We" (the general community?) could simply wash our hands of this and apply future discussion of a cooperative standard (as per https://github.com/cardano-foundation/CIPs/issues/836#issuecomment-3046814048) into a complete CIP for `//payment` links: especially since without that `authority` term, it can't comply either with CIP-0013 or the extension proposed in #843 for which `//payment` has been slated (though [not set in CPS-0016](https://github.com/cardano-foundation/CIPs/tree/master/CPS-0016#registered-uri-authorities) yet).

Crypto2099:
Thanks to @MadOrkestra and @francisluz for finally dragging this one out of the shadows and into the light for some proper definition and refinement. I've done my best to try and stay close to the version as implemented by Begin (as slurped from generating a payment request URI in their app) while still extending the definition a bit, particularly for supporting multiple native assets and moving away from a `ticker whitelist`.

Note that we do fall back to a bech32 representation of native assets because this allows for a predictable-length URI regardless of the token name (which can be an additional 32-bytes in addition to the policy ID, particularly tricky for NFTs) which should still allow wallets to maintain a "known whitelist" of their own for popular tokens such as USDM and HOSKY and others.

rphair (commented):
Looking forward to confirming this at the next meeting (https://hackmd.io/@cip-editors/116) and recording the feedback here from any interested parties especially @MadOrkestra @francisluz and any other implementors following this issue.

I might only be half-complete tagging reps from other wallets besides Begin, so please I will leave that to @MadOrkestra @Crypto2099 @Ryun1 - with invitations to that meeting & calls for review and suggestions all around this proposal.  (Normally we would wait 1 more meeting cycle for detailed Review after Triage first, but time seems essential because of the implementation in progress.)

CIP-XXXX/README.md
```diff
@@ -0,0 +1,263 @@
+---
+CIP:
+    ?
```
```suggestion
CIP: 157
```

rphair (commented):
Great work finally bringing this home @Crypto2099 - candidacy confirmed @ yesterday's CIP meeting & please update the document & directory name accordingly 🎉

Personally I expect to ✅ this soon but want to give it one more detailed read when settled again from travelling around the end of this week: especially cross referencing with CIP-0013 family members & #1058.

MadOrkestra:
Way more complex than I initially thought, but great job @Crypto2099

I have one thing to at least think about: As per [RFC3986](https://datatracker.ietf.org/doc/html/rfc3986) "|" (used to separate the asset count here) is not allowed as unencoded character in URLs. Not sure if this can cause problems with individual URL schemes like web+cardano, but it might be worth looking into?

Crypto2099:
> Way more complex than I initially thought, but great job @Crypto2099
> 
> I have one thing to at least think about: As per [RFC3986](https://datatracker.ietf.org/doc/html/rfc3986) "|" (used to separate the asset count here) is not allowed as unencoded character in URLs. Not sure if this can cause problems with individual URL schemes like web+cardano, but it might be worth looking into?

Thanks for the heads up on this, I will take a look and pivot the standard to use a URI-safe character

MadOrkestra:
> > I have one thing to at least think about: As per [RFC3986](https://datatracker.ietf.org/doc/html/rfc3986) "|" (used to separate the asset count here) is not allowed as unencoded character in URLs. Not sure if this can cause problems with individual URL schemes like web+cardano, but it might be worth looking into?
> 
> Thanks for the heads up on this, I will take a look and pivot the standard to use a URI-safe character

Sorry, didn't have too much time for this earlier... my understanding of RFC 3986 is, it it basically comes down to one of these: 

`- . _ ~ ! $ ( ) * + ;`

So unless we wanna drive every single dev out there mad with bringing back the good old tilde or cause confusion with a semicolon, I think `?t=asset1*amount1,asset2*amount2` is the logical way to go? Amounts could be optional (1 if not set - to save URL characters) and have a decimal point too (if needed) - with no URL encoding necessary at all.

Crypto2099:
> > > I have one thing to at least think about: As per [RFC3986](https://datatracker.ietf.org/doc/html/rfc3986) "|" (used to separate the asset count here) is not allowed as unencoded character in URLs. Not sure if this can cause problems with individual URL schemes like web+cardano, but it might be worth looking into?
> > 
> > 
> > Thanks for the heads up on this, I will take a look and pivot the standard to use a URI-safe character
> 
> Sorry, didn't have too much time for this earlier... my understanding of RFC 3986 is, it it basically comes down to one of these:
> 
> `- . _ ~ ! $ ( ) * + ;`
> 
> So unless we wanna drive every single dev out there mad with bringing back the good old tilde or cause confusion with a semicolon, I think `?t=asset1*amount1,asset2*amount2` is the logical way to go? Amounts could be optional (1 if not set - to save URL characters) and have a decimal point too (if needed) - with no URL encoding necessary at all.

The asterisk should do the job nicely without causing too much confusion (like the dash or plus sign would). I think we can stick to whole-integer representation since the ledger does not natively understand Native Asset decimals anyway and wallets need to build the transaction with the whole integer amount of the native asset in question. I do agree that removing `*1` also seems a logical default to save characters and space.

Crypto2099:
The devs from Vespr pointed out to me that very few wallets and none of the current data providers (Koios, Blockfrost, etc) support querying assets by their fingerprint. While I feel that the fingerprint is preferable due to being a fixed-length string so having predictable URLs lengths is ideal, for ease of implementation we should probably at least make an update to also support the traditional <policy_id>.<asset_id> format for sending tokens

AlexDochioiu:
Appreciate the update to support asset unit @Crypto2099 🙏🏻

AlexDochioiu:
@Crypto2099 I'd add an explicit remark in the CIP that token quantity **is not** decimals adjusted. (token quantity is always `int` / `bigint` ; never `float` or `decimal`)

Kingsleyioryoo (commented):
1DdfpmAw7QKogrSQ2VcDhqjtoXyHHPQq6f

CIP-XXXX/README.md
```diff
@@ -0,0 +1,301 @@
...
+* Providing one (or more) Native Assets to send
+* Providing a transaction metadata message to send
+
+## Path to Active
+
+### Acceptance Criteria
+
+* [ ] Community Feedback and Review Integrated
+* [ ] At least one wallet supports this payment standard
+* [ ] At least one project using this payment standard
```
rphair, line 284: As I also posted in https://github.com/cardano-foundation/CIPs/pull/1058#discussion_r2239029563 @Ryun1 (our usual parameter setter for this, or @perturbing or any other community reviewer) may think this should be increased from 1 to 2 or maybe more.

Personally I think 1 is fine on each side (producer + consumer) due to reasons of inevitable imitation that I mentioned in the other comment: but will agree to increase it if other editors request.

@Crypto2099 I do think the term `project` needs to be elaborated, since wallets are also "projects" — whatever it takes to establish that the "producer" and "consumer" of payment URIs are distinct projects.  As we mentioned at the meeting I think it would also help to specify that those 2 projects should be from distinct _vendors_... a statement that would also support the `Path to Active` credibility of https://github.com/cardano-foundation/CIPs/pull/1058.

rphair (commented):
As I just posted for candidate CIP-0158 (Browse Authority) (https://github.com/cardano-foundation/CIPs/pull/1058#pullrequestreview-3066295653) I believe this has all the elements of a proper CIP: with exhaustive detail & pre-emptive consideration of valuable use cases.

I think it is especially valuable that an ABNF grammar has been included: not because it will be automatically parseable or useful in validating URI syntax, but to eliminate ambiguities & edge cases that could otherwise emerge between different wallets & apps.

So as also posted for the parallel candidate CIP I am tentatively ✅ing this proposal as we begin the inevitable period of convergence between in-progress implementations (thanks @AlexDochioiu) and the standard expressed in ideal form.

If implementors ever think it would help to brainstorm in a "working group" setting around particular syntax / interpretation challenges for either of these proposals, please post to say so here and I'll put it up for Review on the CIP agenda.

CIP-XXXX/README.md
MadOrkestra: Just FYI, the OG CIP13 stated "one or more wallets" as acceptance criteria - no mention of projects whatsoever, so this is already a neat step up ;)

Maybe lets go with "At least one website using this payment standard" instead of project? Or maybe thats exactly Adams point, as a _project_ using QR-Codes for payments doesn't necessarily need to be a website (which is the beauty of this...)

MadOrkestra:
Just revisited the state of CIP13 integrations: Yoroi is the only mobile wallet supporting the web+cardano:addr scheme, Vespr went with web+cardano://addr (which - from what I hear - is actually needed for apps to enable proper deeplinking)

So yeah, can't stress enough how important it'd be to finish this CIP and pressure every wallet into supporting this, so we can maybe in 2026 pay with ADA by phone...

rphair:
@Crypto2099 you're the best candidate I can think of to bring together all possible approaches to this... so please let us know what you think needs to be done & editors will do anything possible to ensure it's merged promptly: with no time to waste, as @MadOrkestra says 😅

rphair:
> Vespr went with web+cardano://addr (which - from what I hear - is actually needed for apps to enable proper deeplinking)

@MadOrkestra can you explain why Vespr said this is "needed"... since **putting an "address" variable**, rather than a fixed `authority` token string (like `pay`), **after the `//` is a violation of the URI syntax**.  That means it would likely also break any widely-used URI parser: so how could it enable "proper deeplinking"?

Or can you please get them to summarise why they added simply the `//` instead of adding `//` + _authority_... which must always come as a pair, according to the URI standard itself (also linked from CIP-0013)?  https://en.wikipedia.org/wiki/Uniform_Resource_Identifier#Syntax

In any case 2025 seems to have been a regression for Cardano standardised payments: diverging from a common interest in following any posted standards or even suggesting any alternatives for discussion, beyond what @Crypto2099 was offering with this CIP candidate and @MadOrkestra's unification attempts with https://github.com/cardano-foundation/CIPs/issues/836.

Unless this changes ASAP, we're likely to end up with a fragmented and inconsistent Cardano payment infrastructure: a disgrace compared to other ecosystems.  It's too bad we have such brilliant UI/UX for Cardano wallets and yet nobody seems willing even to get together about standardising on a best practice for payment links. 🤢

I know in year 2026 we can do better and I promise the CIP editors will do whatever we can to arrange for a proper working group for this badly needed discussion & convergence.  cc @Ryun1 @perturbing

MadOrkestra:
> > Vespr went with web+cardano://addr (which - from what I hear - is actually needed for apps to enable proper deeplinking)
> 
> @MadOrkestra can you explain why Vespr said this is "needed"... since **putting an "address" variable**, rather than a fixed `authority` token string (like `pay`), **after the `//` is a violation of the URI syntax**. That means it would likely also break any widely-used URI parser: so how could it enable "proper deeplinking"?

You're right, it should work without `//` (just like `mailto:` does) - this might however be OS specific and maybe there are limitations on what individual URI schemes you can register on the OS level - wouldn't be the first time Apple strayed from the right path... I'll look into it.

mpizenberg:
For info, I just asked an llm what they thought about this scheme.

TLDR:
1. should not use "//" without a real naming authority after (server, domain name, ...)
2. might want to avoid "web+" saying it’s mostly for web apps/protocols. Not sure I’m super convinced by this argument. 

That being said, the mdn docs for "web+" uris contradicts point (1) with its examples so ... https://developer.mozilla.org/en-US/docs/Web/API/Navigator/registerProtocolHandler

Here is the response from Claude Sonnet:

---

Based on the RFC standards, here's what you need to know about URI scheme syntax:

## **When to use "//"**

The double slashes "//" should ONLY be used when the scheme contains a hierarchical structure with a naming authority. According to RFC standards, double slashes indicate that what follows is the top hierarchical element for a naming authority. (https://www.ietf.org/rfc/rfc2718.txt)

**Examples:**
- `http://example.com/path` - HAS "//" because it has an authority (example.com)
- `mailto:user@example.com` - NO "//" because it's non-hierarchical (no authority component)
- `tel:+1-555-1234` - NO "//" (non-hierarchical)
- `sms:+1-555-1234` - NO "//" (non-hierarchical)
- `file:///path/to/file` - HAS "//" for authority (empty authority = localhost)

## **Your Cardano scheme**

Your proposed scheme:
```
web+cardano://pay/v1/address
```

**Has issues:**

1. **The `web+` prefix**: Custom URI schemes for web apps must start with "web+" followed by at least one lowercase letter. This is specifically for web-based protocol handlers. If you're building a native app or want a proper registered scheme, you should drop the `web+` prefix and just use `cardano:`. (https://developer.mozilla.org/en-US/docs/Web/API/Navigator/registerProtocolHandler)

2. **The "//" usage**: You're using `//pay` where "pay" appears to be **part of the path, not an authority**. This is improper use of "//". The double slashes should not be used following the scheme when there's no conformant hierarchical naming authority.

(https://datatracker.ietf.org/doc/html/rfc7595)

## **Better alternatives:**

**Option 1 - Non-hierarchical (like mailto):**
```
cardano:pay?address=<address>&amount=<amount>
```
or
```
cardano:pay/<address>?amount=<amount>
```

**Option 2 - If you really need hierarchy:**
```
cardano://authority/v1/pay/<address>
```
Only use this if "authority" represents an actual naming authority (server, domain, etc.).

**Option 3 - Simple path-based (most common for payment URIs):**
```
cardano:address?amount=<amount>&memo=<text>
```

Similar existing payment schemes: `bitcoin:address?amount=0.1`, `ethereum:address@chainId`

The key principle: **Don't use "//" unless you have a real authority component** (like a server or domain name).

MadOrkestra:
right. so `web+cardano:{bech32_addr}?amount={ADA}` is still the way to go for the original CIP13 implementation and I can report that we finally got some consensus among wallets. Yoroi, Vespr and begin are fixing their implementations as we speak (or already have, but not yet pushed to live versions), I think eternl was already good.

for advanced payments we'll need the hierarchy I guess. and, as this not only involves native apps on mobile but browser extensions as well, we should stick to the web+cardano prefix.

rphair:
Implementation perhaps underway from @AlexDochioiu & team as noted in https://github.com/cardano-foundation/CIPs/issues/836#issuecomment-3961657610... thanks for whatever priority you can devote to this. 🙏

CIP-XXXX/README.md
```diff
@@ -0,0 +1,301 @@
...
+> sending quantities of 1 (one) of a native asset. If the quantity is not
+> specified then wallets should default to a quantity of 1.
+
+Example:
+
+* 100x of asset12ffdj8kk2w485sr7a5ekmjjdyecz8ps2cm5zed
+* 5x of asset17q7r59zlc3dgw0venc80pdv566q6yguw03f0d9
+
+  ```
+  t=asset12ffdj8kk2w485sr7a5ekmjjdyecz8ps2cm5zed*100,asset17q7r59zlc3dgw0venc80pdv566q6yguw03f0d9*5
```
colll78, line 157: I would recommend that CIP-157 encode the full ledger asset ID instead; either policyId.assetNameHex or plain policyIdAssetNameHex. 

A CIP-14 fingerprint is a 20-byte hash of policyId || assetName. It is not a reversible encoding, so wallets cannot turn it into the ledger unit they need to build the corresponding tx.

colll78 (changes requested):

CIP-XXXX/README.md
colll78: The issues I see:
- CIP-14 fingerprints are not reversible.
- The grammar requires *quantity, while the accompanying note says quantity is optional and defaults to one.
- The on-chain meaning of i= and n= is unspecified.
- Duplicate and unknown query-parameter behavior is unspecified.
- “64 characters” conflicts with the ledger metadata constraint, which is 64 UTF-8 bytes per text value.
- Zero is permitted by 1*DIGIT, despite the prose requiring positive quantities.

rphair: @francisluz how would you feel about changing your implementation to match what @colll78 is suggesting?  Please <ins>we need you to answer this</ins> because, at this point, this CIP draft should either be merged or discarded in favour of whatever can be written to match future implementations.  

So far Begin Wallet has the only credit for an implementation, while @AlexDochioiu has said in https://github.com/cardano-foundation/CIPs/issues/836#issuecomment-3961657610 that VESPR might support this someday.  Failing to merge this CIP because of lack of confirmation would invalidate this tentative progress: but would avoid the dev community wondering whether something called "CIP-0157" needs to be implemented and stalling any further attempts to standardise payment links (cc @Crypto2099).

If using the asset ID is the only way to future-proof this, then we need to make this change so other wallet devs can implement this in confidence.  If this is the _only_ issue @colll78 is bringing up in review, then I trust that doing it this way is important even if breaking some already issued links.

So please @francisluz if you think that Begin Wallet users (or any other Cardano payments) depend upon the original format then maybe this new suggestion can be written into a `v2` as encoded in the URL.  Again, I would rather see a general standard for payment links begin with Version 2 than never begin at all.

CIP-XXXX/README.md
```diff
@@ -0,0 +1,301 @@
...
+Title: Cardano URIs - Enhanced Payments
+Category: Wallets
+Status: Proposed
+Authors:
+    - Adam Dean <adam@crypto2099.io>
+Implementors:
+    - Begin Wallet <@francisluz>
+Discussions:
+    - https://github.com/cardano-foundation/CIPs/pull/843
+    - https://github.com/cardano-foundation/CIPs/issues/836
```
```suggestion
    - Original PR: https://github.com/cardano-foundation/CIPs/pull/843
    - Implementation discussion: https://github.com/cardano-foundation/CIPs/issues/836
```
New format (& workflow validation) requires labels for links (brief & descriptive by editorial standard).

CIP-XXXX/README.md
```diff
@@ -0,0 +1,301 @@
...
+---
+
+## Abstract
+
+This CIP will propose a new [CIP-13] Extension; introducing a new, dedicated
+`pay` authority and provide support for _Native Assets_ and transactional
+_Metadata_ as well as providing for extensibility and versioning. All features
+lacking in the original [CIP-13] payment URIs.
+
+## Motivation: why is this CIP necessary?
```
```suggestion
## Motivation: Why is this CIP necessary?
```
(1 of 2) fix long standing capitalisation glitch in standard template

CIP-XXXX/README.md
```diff
@@ -0,0 +1,301 @@
...
+#### Legacy Byron Address support
+
+``` 
+web+cardano://pay/v1/Ae2tdPwUPEZHSvyvbTYWCVUw3wBhDGenxhhH1BEJndFhaMcD5tm1R4RtuAr
+```
+
+Supports backward compatibility for Byron-era Base58 addresses. All query
+parameters work the same.
+
+## Rationale: how does this CIP achieve its goals?
```
```suggestion
## Rationale: How does this CIP achieve its goals?
```
(2 of 2) fix long standing capitalisation glitch in standard template

rphair (commented):
Given recent discussion of how payment links might accommodate "handles":
* https://github.com/cardano-foundation/CIPs/pull/1229#discussion_r3742436344

— and because we already have one implementing wallet (Begin) and another [willing to implement](https://github.com/cardano-foundation/CIPs/issues/836#issuecomment-3961657610) (VESPR, as per @AlexDochioiu) —  we should have this CIP confirmed by merging as our best effort to standardise payment links.  I've made some arguments for this in the course of our only recommended technical change so far (https://github.com/cardano-foundation/CIPs/pull/843#discussion_r3745455357).

We should also keep in mind the version currently `/v1` in the URL can always be bumped if mutual compatibility isn't perfect the first time, or needs to specify additional things like "handle resolvers" (cc @kingslavcho @papag00se).

Given @colll78's latest https://github.com/cardano-foundation/CIPs/pull/843#discussion_r3743952708 I would recommend this course:
1. a rewrite as he suggests... or however that discussion concludes (and hopefully he or @francisluz would please submit this as a correction & editors will commit it);
2. I'm submitting a format edit below to match our current CIP standards & pass validation (vs. [these failures](https://github.com/cardano-foundation/CIPs/actions/runs/31337336961/job/93305240387?pr=843));
3. once all edits are in, I would approve this & tag `Last Check` to get this merged ASAP — especially since it would allow further deltas to be submitted freely: with appreciation for how far @Crypto2099 has carried this already (cc @Ryun1 @perturbing);
4. if @Crypto2099 cannot rename the containing directory to `CIP-0157` as suggested long ago, the editors can fix the repository structure accordingly after this is merged.

Issue #564, opened 2023-07-29
CIP-1694: Careful thinking about extra bytes in DRep-ID and other governance elements


gitmachtl:
I wanna encourage everyone to take a deeper and careful look at all the used keys, hashes, ids, etc. For the
conway/votaire/1694 era.

As an example, the DRep-ID is currently designed to be just the hash of the Drep-Credential-Public-Key. No extra data is in there, on the hex level and on the bech level "drep_id". We added extra information for example for all the addresses in cardano by making the first byte the information-byte. 

The question is if such a concept could also be used for the dRep-ID (example parameter). As stated here
https://github.com/cardano-foundation/CIPs/blob/db70d91029a6512934d11c7eb4436acabe5db182/CIP-1694/README.md?plain=1#L381-L386
the dRep ID can come from a verification key or from a native script or plutus script. So it would be possible - if needed - to also add a byte at the front of the hash to specify the type before it gets converted into bech format. 

In the past we produced the issue for example with the NativeAssets and decimals. If we could have only a single byte more in the assetSubject, the decimals wouldn't be such an issue as they currently are. With strange online/offline solutions, problems with offline-wallets and hardware-wallets.

So if we could come together to take a look at all keys/ids/credentials/etc. for the governance stuff, and maybe adding a byte or two here and there to be future proof, that would be awesome! 🙏 

Maybe i am overthinking it, but we should learn from the past and include stuff that make us more future proof imo. Now we have the time to do it.

@disassembler @KtorZ @SebastienGllmt @rphair @crypto2099 @AndrewWestberg @SmaugPool  @janmazak @refi93 @Ryun1 @kevinhammond

Crypto2099:
This is interesting but would there be any way to force the use of this prefix byte in order to create a valid hash for a Plutus Script versus a key-pair? If not, is there a potential security vulnerability if someone mixes up or uses the wrong prefix?

I do like the idea of programmatic ability to detect/differentiate between a key-pair or Plutus Script based only on the provided ID.

gitmachtl:
The DREP-ID is only an example. Maybe we don't need extra bytes, but maybe they could be useful in other parts of the governance too.

Ryun1:
I have a pending CIP about DRep credentials, which should cover this. At the moment I had planned to model DRep IDs off of Pool IDs so no network tag or extra bits to indicate script or not. But I would be keen to hear feedback and use cases for extra bytes.

Ryun1:
[...]

PR #845, opened 2024-06-22
CIP-0005 | update drep and cc identifier prefixes

ashisherc:
Update drep hashes prefix to reflect the actual hash type, as dRep ID is not just a hash but a credential (unlike pool ID), we need a hex representation of the various keys, bech32 is supposed to be only for the human-readable representation.

eg.
- modify `drep` -> `drep_vkh`
- `cc_cold` > `cc_cold_vkh`
- etc

---
`drep`, `cc_cold` etc prefix to be used to represent the credential hex format, to be proposed in a dedicated CIP.

### DIFF
--- a/CIP-0005/README.md
+++ b/CIP-0005/README.md
@@ -92,35 +92,38 @@ We define the following set of common prefixes with their corresponding semantic
 
 ### Hashes
 
-| Prefix             | Meaning                                                               | Contents                                                                 |
-| ---                | ---                                                                   | ---                                                                      |
-| `asset`            | Fingerprint of a native asset for human comparison                    | See [CIP-0014]                                                           |
-| `pool`             | Pool operator verification key hash (pool ID)                         | blake2b\_224 digest of an operator verification key                      |
-| `script`           | Script hash                                                           | blake2b\_224 digest of a serialized transaction script                   |
-| `addr_vkh`         | Address verification key hash                                         | blake2b\_224 digest of a payment verification key                        |
-| `addr_shared_vkh`  | Shared address verification key hash                                  | blake2b\_224 digest of a payment verification key                        |
-| `policy_vkh`       | Policy verification key hash                                          | blake2b\_224 digest of a policy verification key                         |
-| `stake_vkh`        | Stake address verification key hash                                   | blake2b\_224 digest of a delegation verification key                     |
-| `stake_shared_vkh` | Shared stake address verification key hash                            | blake2b\_224 digest of a delegation verification key                     |
-| `req_signer_vkh`   | Required signer verification key hash                                 | blake2b\_224 digest of a required signer verification key                |
-| `vrf_vkh`          | VRF verification key hash                                             | blake2b\_256 digest of a VRF verification key                            |
-| `datum`            | Output datum hash                                                     | blake2b\_256 digest of output datum                                      |
-| `script_data`      | Script data hash                                                      | blake2b\_256 digest of script data                                       |
-| `drep`             | Delegate representative verification key hash (DRep ID)               | blake2b\_224 digest of a delegate representative verification key        |
-| `drep_script`      | Delegate representative script hash (DRep ID)                         | blake2b\_224 digest of a serialized delegate representative script       |
-| `cc_cold`          | Constitutional committee cold verification key hash (cold credential) | blake2b\_224 digest of a consitutional committee cold verification key   |
-| `cc_cold_script`   | Constitutional committee cold script hash (cold credential)           | blake2b\_224 digest of a serialized constitutional committee cold script |
-| `cc_hot`           | Constitutional committee hot verification key hash (hot credential)   | blake2b\_224 digest of a consitutional committee hot verification key    |
-| `cc_hot_script`    | Constitutional committee hot script hash (hot credential)             | blake2b\_224 digest of a serialized constitutional committee hot script  |
+| Prefix                 | Meaning                                                               | Contents                                                                 |
+| ---                    | ---                                                                   | ---                                                                      |
+| `asset`                | Fingerprint of a native asset for human comparison                    | See [CIP-0014]                                                           |
+| `pool`                 | Pool operator verification key hash (pool ID)                         | blake2b\_224 digest of an operator verification key                      |
+| `script`               | Script hash                                                           | blake2b\_224 digest of a serialized transaction script                   |
+| `addr_vkh`             | Address verification key hash                                         | blake2b\_224 digest of a payment verification key                        |
+| `addr_shared_vkh`      | Shared address verification key hash                                  | blake2b\_224 digest of a payment verification key                        |
+| `policy_vkh`           | Policy verification key hash                                          | blake2b\_224 digest of a policy verification key                         |
+| `stake_vkh`            | Stake address verification key hash                                   | blake2b\_224 digest of a delegation verification key                     |
+| `stake_shared_vkh`     | Shared stake address verification key hash                            | blake2b\_224 digest of a delegation verification key                     |
+| `req_signer_vkh`       | Required signer verification key hash                                 | blake2b\_224 digest of a required signer verification key                |
+| `vrf_vkh`              | VRF verification key hash                                             | blake2b\_256 digest of a VRF verification key                            |
+| `datum`                | Output datum hash                                                     | blake2b\_256 digest of output datum                                      |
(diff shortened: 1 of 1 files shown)


gitmachtl:
whats a drep address? there is just an ID

drep1xxx  
drep_script1xxx

ashisherc:
> whats a drep address? there is just an ID
> 
> drep1xxx drep_script1xxx
> 
> there is also no flag or prebyte that would differ from a mainnet or testnet "address". like with a pool id, there is just one. same for mainnet and testnets.
> 
> a drep-id is like a pool-id. you can delegate to it. i don't see how such an id would count as an address?

dRep-id is not same as pool id, pool ID is a single hash, whereas dRep is a credential with - type and key field, see here, https://github.com/IntersectMBO/cardano-ledger/blob/a44a5b5ac9f68c5d7a2ff38b040a5fcac8ccfc6a/eras/conway/impl/cddl-files/conway.cddl#L338

similarly, a stake address is defined with stake credential and an additional network flag is added to identify if a stake address is on mainnet or testnet.

Similarly a dRep address is combination of network flag + dRep Credential

gitmachtl:
there is no network flag for a drep id. i made an issue about that i think a year ago if such a thing should be implemented, did not get the attention of many, result was: "no we don't need a network flag for dreps"

gitmachtl:
that was the issue i opened back in time: https://github.com/cardano-foundation/CIPs/issues/564

ashisherc:
`1001 - 1101` are reserved for future address formats, which unfortunately is not enough to define all the addresses above. I'd also like to point out that a unique hex representation of the above credentials is also required.

ashisherc:
Since a network flag is probably not required for these addresses, should we use the network flag bytes instead for specifying the credential type, eg.

0-3 bits define the address type

0000 - cc_hot
0001 - cc_cold
0010 - drep

4-7 bits define the credential type

0000 - keyhash28
0001 - scripthash28

Ryun1:
Hey @ashisherc, thanks for opening this.

A few thoughts from me;

- The DRep ID definition was inherited from CIP-1694 [here](https://github.com/cardano-foundation/CIPs/blob/master/CIP-1694/README.md#registered-dreps).
- [CIP-105](https://github.com/cardano-foundation/CIPs/blob/master/CIP-0105/README.md#why-not-allow-network-tags) adds some extra context and thought around this, see rationale on [Why not allow network tags?](https://github.com/cardano-foundation/CIPs/blob/master/CIP-0105/README.md#why-not-allow-network-tags)
  - to add more to this, at the time there just was not really an appetite to add network tags (See [Martin's issue](https://github.com/cardano-foundation/CIPs/issues/564)), instead we decided to follow CIP-1694's definition and omit them.
- The current way of implementing DRep IDs, has already seen much adoption across tooling.
  - any such changes like this would be quite disruptive not just for implementers, but for those community members who already understand the current DRep ID mechanisms
- I would suggest that this pull request amend adjacent CIPs (CIP-1694, CIP-105 and CIP-19), to supply the rationale you are providing via comments in a more discoverable way.

ashisherc:
@Ryun1 
`The DRep ID definition was inherited from CIP-1694 [here](https://github.com/cardano-foundation/CIPs/blob/master/CIP-1694/README.md#registered-dreps).`

The CIP1694  does not define the bech32 prefix, I think there is a misunderstanding here, I am not suggesting a change to how a deep ID is represented, but infact suggesting a change that fits well with existing bech32 formats.

eg. the bech32 prefix for deep keyhash is just `drep` but I am suggesting it should be drep_vkh. Following the CIP-1694 the actually hex representation should be as I am suggesting or similar to it.

CIP-105 as well does not require any changes, it does define a vkey prefix precisely as `drep_vk`, again stating that this CIP should use `drep_vkh` for a key hash.

I would propose relevant amendments to cip-19 once this CIP makes the changes and uses the relevant bech32 format. As suggesting a deep address prefix will use `drep` as prefix for a deep address but has been already used here to define a v keyhash and not an address.

Ryun1:
@ashisherc 

As you have pointed out, when defining this we more or less ignored the "credential" nature of these DRep and CC credentials, and thus avoided defining addresses for them.

### Motivation for current setup

- We were modelling off of stake pool IDs, just as vote delegation (delegating to DReps) is modelled on top of delegation to stake pools.
  - This was just the easier and simplier.
  - At the time there was little to no community interest in pursing anything more than this.
 
### Benefits of proposed changes

- Conformity with existing Stake and Payment credential definitions

Im not sure of anymore?

### Drawbacks of proposed changes

- Adding DRep/CC addresses would, I believe, add confusion.
  - for me, the term address is something ada can be sent to
  - with these addresses ada cant be sent which I see as counter intuitive
- Existing implementers would have to upgrade or not - potentially leading to fragmentation and confusion

### Specific Responses

> I am not suggesting a change to how a drep ID spec should be, but infact suggesting a change that fits well with existing bech32 formats in this CIP.

My bad, I had misinterpreted you as wanting to remove DRep IDs.

> eg. the bech32 prefix for drep keyhash is just drep but I am suggesting it should be drep_vkh.

This is a fair take, the reason I chose `drep` over `drep_vkh` is that I was modelling the DRep IDs off of Pool IDs (`pool`).

> I would propose relevant amendments to cip-19 once this CIP makes the changes and uses the relevant bech32 format. As suggested a drep address prefix will use drep as prefix for a drep address but has been already used here to define a vKeyhash and not an address.

My suggestion is that this proposal becomes it's own CIP, then while editing CIP-19 the edits can point to the full definition and explaination within the new CIP.

Ryun1:
Also, not sure if intentional, but you seemed to have removed all significant file changes with the last force push.

ashisherc:
@Ryun1 thanks for the background, the idea is not to have drep/cc as addresses in the sense of sending/receiving ADA but a unique hex representation for sharing across the community, specially in tooling. While it was a good idea to base of pool id, but due to nature of these keys being a credential and not just a vkh unlike a pool id, it is a bit misrepresented.

I agree for a dedicated CIP for this, but I would like to use `drep` prefix for such identifier, similarly for other credentials as well, and hence update this current CIP to represent a vkh as such, hence this PR.

`Also, not sure if intentional, but you seemed to have removed all significant file changes with the last force push.`

I updated the PR to only keep the changes relevant for this CIP which is updating the prefix to use `_vkh` for the key hash values.

While this amendment might create confusion initially, it is not a user facing terminology yet, but will soon become once the explorers etc start showing this data, I believe it will be good to modify it at this time then to be stuck with it forever.

Ryun1:
@ashisherc 

> While it was a good idea to base of pool id, but due to nature of these keys being a credential and not just a vkh unlike a pool id, it is a bit misrepresented.

I completely agree, it is somewhat _wonky_ 😆

> I agree for a dedicated CIP for this, but I would like to use drep prefix for such identifier, similarly for other credentials as well, and hence update this current CIP to represent a vkh as such, hence this PR.

I look forward to seeing this proposal 💪

> While this amendment might create confusion initially, it is not a user facing terminology yet, but will soon become once the explorers etc start showing this data, I believe it will be good to modify it at this time then to be stuck with it forever.

understood, good points
ill wait for the proposal to discuss further

ashisherc:
@Ryun1 https://github.com/cardano-foundation/CIPs/pull/857

CIP-0005/README.md
```diff
@@ -92,35 +92,38 @@ We define the following set of common prefixes with their corresponding semantic
...
+| `req_signer_vkh`       | Required signer verification key hash                                 | blake2b\_224 digest of a required signer verification key                |
+| `vrf_vkh`              | VRF verification key hash                                             | blake2b\_256 digest of a VRF verification key                            |
+| `datum`                | Output datum hash                                                     | blake2b\_256 digest of output datum                                      |
+| `script_data`          | Script data hash                                                      | blake2b\_256 digest of script data                                       |
+| `drep_vkh`             | Delegate representative verification key hash                         | blake2b\_224 digest of a delegate representative verification key        |
+| `drep_script_vkh`      | Delegate representative script hash                                   | blake2b\_224 digest of a serialized delegate representative script       |
```
Ryun1, line 110:
```suggestion
| `drep_script`      | Delegate representative script hash                                   | blake2b\_224 digest of a serialized delegate representative script       |
```
script verification key hash?

ashisherc:
merged with https://github.com/cardano-foundation/CIPs/pull/857

PR #847, opened 2024-06-25
CIP-1694 | Updates and semantics

Hornan7:
- Replace "New Constitutional Committee" governance action by "Update Committee"
- Replace "Update to the Constitution" governance action by "New Constitution". This is how it is called on-chain when submitting this governance action and when querying its tag. I also changed it for semantic reasons. 
- Replace "Proposal policy" by "Guardrails Script". This wording is mainly used broadly by everyone and is also the wording used in the Interim constitution Article VIII, Apendix 1, (Automated Checking ("Guardrails Script")).
- Update each parameters name tags as seen on chain. (conway era)
- Replace "Pre-defined Drep" name by "Pre-defined voting option" at 3 different places in the CIP-1694 (Which already got me in big trouble, but lets not go there. =P)

### DIFF
--- a/CIP-1694/README.md
+++ b/CIP-1694/README.md
@@ -158,7 +158,8 @@ How any private companies, public or private institutions,  individuals etc. cho
   - [Constitutional committee keys](#constitutional-committee-keys)
   - [Replacing the constitutional committee](#replacing-the-constitutional-committee)
   - [Size of the constitutional committee](#size-of-the-constitutional-committee)
-  - [Term limits](#term-limits)
+  - [Terms](#terms)
+  - [Guardrails Script](#guardrails-script)
 + [Delegated representatives (DReps)](#delegated-representatives-dreps)
   - [Pre-defined Voting Options](#pre-defined-voting-options)
   - [Registered DReps](#registered-dreps)
@@ -224,10 +225,10 @@ The constitutional committee will use a hot and cold key setup, similar to the e
 #### Replacing the constitutional committee
 
 The constitutional committee can be replaced via a specific governance action
-("New constitutional committee", described below) that requires the approval of both
+("Update committee", described below) that requires the approval of both
 the **SPOs** and the **DReps**.
 The threshold for ratification might be different depending on if the governance is
-in a state of confidence or a state of no confidence.
+in a normal state or a state of no confidence.
 
 The new constitutional committee could, in principle, be identical to or partially overlap the outgoing committee as long as the action is properly ratified.
 This might happen, for example, if the electorate has collective confidence in all or part of the committee and wishes to extend its term of office.
@@ -236,14 +237,14 @@ This might happen, for example, if the electorate has collective confidence in a
 #### Size of the constitutional committee
 
 Unlike the Shelley governance design, the size of the constitutional committee is not fixed and can be any nonnegative number.
-It may be changed whenever a new committee is elected ("New constitutional committee and/or threshold").
+It may be changed whenever a new committee is elected ("Update committee").
 Likewise, the committee threshold (the fraction of committee `Yes` votes that are required to ratify governance actions) is not fixed and
 can also be varied by the governance action.
 This gives a great deal of flexibility to the composition of the committee.
 In particular, it is possible to elect an empty committee if the community wishes to abolish the constitutional committee entirely. Note that this is different from a state of no-confidence and still constitutes a governance system capable of enacting proposals.
 
 There will be a new protocol parameter for the minimal size of the committee,
-itself a nonnegative number called `ccMinSize`.
+itself a nonnegative number called `committeeMinSize`.
 
 #### Terms
 
@@ -269,16 +270,16 @@ The maximum term is a governance protocol parameter, specified as a number of ep
 During a state of no-confidence, no action can be ratified,
 so the committee should plan for its own replacement if it wishes to avoid disruption.
 
-#### Proposal policy
+#### Guardrails Script
 
 While the constitution is an informal, off-chain document, there will
 also be an optional script that can enforce some guidelines. This script
 acts to supplement the constitutional committee by restricting some
 proposal types. For example, if the community wishes to have some hard
 rules for the treasury that cannot be violated, a script that enforces
-these rules can be voted in as the proposal policy.
+these rules can be voted in as the guardrails script.
 
-The proposal policy applies only to protocol parameter update and
+The guardrails script applies only to protocol parameter update and
 treasury withdrawal proposals.
 
 <!---------------------------           DReps          ------------------------>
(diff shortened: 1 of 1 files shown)


rphair (commented):
Looks good to me & will endorse as soon as it passes one or two authoritative reviews.

@Ryun1 - cross-referencing proposed _CIP-0120? | Constitution specification_ #796 which still uses the term "updated version" - when according to the changes here it would be an entirely "new" constitution (please correct me if I've misunderstood).

WhatisRT:
Note that the proposal policy isn't restricted to being a guardrails script. In fact, the main motivation to introduce it was to put some restriction on the treasury which people _really_ cared about in the Edinburgh meetup and apparently everyone has forgotten about since then.

For the parameter names there are two versions of them, a short and a long version. The names as they are currently in the CIP should align with those written in previous ledger specs. I think we should really have some kind of dictionary of protocol parameters that is properly maintained. There are the CIPs for each era, but they haven't been maintained (I can spot some old names in there, and they match the naming in this CIP better than the proposed names here btw.) and they aren't particularly useful if you have a specific question about some parameter. So I'm somewhat hesitant to change those names that are actually still in use somewhere without having a proper source of truth.

@rphair do you think such a dictionary would make sense as a CIP? The issue is that it'd have to be actively maintained, so it wouldn't be static. But the upside would be that all those naming discussions would be a lot more public than they currently are.

rphair:
@WhatisRT I recall & know well what you mean about the parameter CIPs being neglected.  This effort ended with an internal disagreement (documented on one of these GitHub CIP PR threads: I think the one that's still unmerged) about whether these CIPs would be full updates (superseding the old ones) or cumulative.  So I don't want to push that approach since whatever impetus there may be this year to maintain such a documentation system might not exist _next_ year.

Likewise with CIPs we would have the problem of editorial process behind every single little parameter definition change: even without considering addition to editor workload (@Ryun1 @Crypto2099 & I would get in the habit probably of rubber-stamping the PRs that came in from "authoritative" sources)... but rather what happens whenever there's a disagreement from the community, which we are obligated to resolve by consensus... and those disagreements about wording, behaviour, or versioning haven't been promptly or authoritatively resolved in the past.

So I would recommend a GitHub structure that would allow updates by whatever sources the Ledger team considers "authoritative": a repo on the `IntersectMBO` or `input-output-hk` to only house a [GitHub Wiki](https://docs.github.com/en/communities/documenting-your-project-with-wikis/about-wikis): like https://github.com/input-output-hk/cardano-node-wiki/wiki - you can see that pages are automatically alphabetised onto a sidebar, which is a pain in the neck for navigation but I think would suit the parameter listings very well.

Note we cannot use `cardano-foundation/CIPs/wiki` since this will have to host the Wiki content I've been working on for the CIP process itself. 😇

WhatisRT:
Thank you! I think in that case I'll just put it somewhere in https://github.com/IntersectMBO/formal-ledger-specifications. It's still public and it makes sense for it to be there, it'll probably just be less visible.

Hornan7:
My goal here was to have something that reflect what the cummunity would have to deal with everyday in the governance state of the ledger on Conway era. Which explain my choice to use the most recent name for those params. 
I also made sure that the PR was editable by maintainers because I was expecting those discussions. 😃👍
I can change the wordings accordin GG to what you guys prefer but I would honestly try to stick with the most recent version of those names IMO since we are entering Conway era very soon. 🥰

Let me know if there is anything you would like me to change or remove, I will.make the modifications. 😃👍

WEEDisFOOD:
This is very cool Mike!

lehins:
Very nice! Thank you @Hornan7

rphair (approved):
I think this could well be approved before the next CIP meeting (https://hackmd.io/@cip-editors/92) but I've put it for `Last Check` there anyway to make sure we get CIP editor sign-off on this & merge no later than that date.

WhatisRT (approved):
I looked at the draft constitution and it also it a 'Guardrails Script'. So apparently that's considered the official name now and I withdraw that point. I think it's really stupid that no one involved with that bothered to make a PR here or let me know, but what's done is done.

As for the protocol parameter names, I started making a table and found a few more inconsistencies. I suspect that there'll be more renamings of them in the future, but maybe that doesn't really matter for this PR. Other CIPs usually had consistent names with the spec but I have a feeling that very few people actually care about that. So go ahead with it.

Ryun1:
Thanks for this @Hornan7 
Sharing this with stakeholders 💪

kevinhammond:
On the parameter name issue, we need canonical versions of the names.  The ones here should be consistent with those used in the guardrails document, which were intended to be the common names that users would need to interact with when making changes (i.e. not necessarily the ledger names, which are primarily internal).  There's more convergence on the newer names, fortunately, but some of the older ones can be rather confusing.  When looking, I found about 5 different conventions in use.

kevinhammond:
> I looked at the draft constitution and it also it a 'Guardrails Script'. So apparently that's considered the official name now and I withdraw that point. I think it's really stupid that no one involved with that bothered to make a PR here or let me know, but what's done is done.

The guardrails script is a proposal policy that enforces the automatable guardrails (which could include treasury withdrawals).  In principle, there could be different proposal policies, but this seems unlikely to happen on Cardano.  Some naming changes/fixes were proposed to the CIP but I don't know whether they made it as a PR?

kevinhammond:
> @WhatisRT I recall & know well what you mean about the parameter CIPs being neglected. This effort ended with an internal disagreement (documented on one of these GitHub CIP PR threads: I think the one that's still unmerged) about whether these CIPs would be full updates (superseding the old ones) or cumulative. So I don't want to push that approach since whatever impetus there may be this year to maintain such a documentation system might not exist _next_ year.

I ran out of energy on this one.  IMO, the sensible thing is to have a living document that doesn't involve looking at 4-5 unrelated CIPs to determine the current set of parameters.  This document aims to do that:

[https://docs.cardano.org/about-cardano/explore-more/parameter-guide/](https://docs.cardano.org/about-cardano/explore-more/parameter-guide/)

rphair:
thanks @kevinhammond - since we have no open CIP PRs for parameter definitions or changes, and since that approach hasn't been proposed for the Conway ledger era, and since the only unmerged one has recently been closed:
- https://github.com/cardano-foundation/CIPs/pull/337

... I personally feel fine about _not_ keeping the protocol parameters on the CIP repository and think that either of the approaches suggested above would work.  cc @Ryun1 @Crypto2099 @KtorZ

Crypto2099 (approved):
I think these changes make sense w.r.t. the context of CIP-1694 and making is more user-friendly to read and align between this document and the "guardrails script" that is included as part of the Interim Constitution document.

That said, I do believe that the lexicon and/or glossary of terms linking between the on-ledger names and the "human friendly" names does fall into the purview of the CIP Editors as, were someone post-Chang to propose that a protocol parameter should be renamed, we need some forum for that discussion and debate to take place in. cc: @rphair

Ryun1 (approved):
thanks Mike!

Ryun1:
We can move discussions of protocol parameter definitions/naming over to #852.
@kevinhammond @WhatisRT @Crypto2099 @rphair

PR #857, opened 2024-07-16
CIP-0129? | Governance Identifiers

ashisherc:
Introduces byte representation along with a bech32 prefixes for different Conway era credentials and identifiers.

This PR includes relevant updates to CIP - 0005, and 0105 (including updated test vector)

---

([rendered proposal](https://github.com/StricaHQ/CIPs/blob/gov-identifiers/CIP-0129/README.md))


CIP-XXXX/README.md
```diff
@@ -0,0 +1,139 @@
+---
+CIP: ????
+Title: Cardano Governance Identifier
+Status: Proposed
+Category: Wallets
```
rphair, line 5:
```suggestion
Category: Tools
```
I am sure `Wallets` will be a huge part of this but the _Motivation_ also suggests other tools will process these IDs.  But @ashisherc @Ryun1 if it can be shown that this CIP is only for how wallets will generate & interpret the keys then please let's establish this here and in the CIP.

CIP-XXXX/README.md
```diff
@@ -0,0 +1,139 @@
...
+
+
+TX ID in Hex - `0000000000000000000000000000000000000000000000000000000000000000`
+Gov Action index in HEX - `11` (number 17)
+
+gov action ID - `000000000000000000000000000000000000000000000000000000000000000011`
+
+bech32 - `gov_action1qqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqpzklpgpf`
+
+## Bech32 Encoding
```
rphair, line 85:
```suggestion
### Bech32 Encoding
```
(a probably inadvertent slip in the header level)

CIP-XXXX/README.md
```diff
@@ -0,0 +1,139 @@
...
+## Rationale: how does this CIP achieve its goals?
+
+By introducing a header and payload design for different keys, we can achieve specifying credentials in a hex format (bytes), being able to share and use across tooling is possible. We also specify bech32 prefixes for user friendly identification of the identifiers.
+
+
+## Path to Active
+
+### Acceptance Criteria
+
+- [] implemented by various tools
```
rphair, line 135:
```suggestion
### Acceptance Criteria

- [ ] Implemented by various tools

### Implementation Plan
```
At the very least it also needs an _Implementation Plan_ and more detail across **both** subheadings.  Checkboxes appear properly formatted when they have either a tick-character (`x` or `X`) or a blank space inside them. :)

CIP-XXXX/README.md
```diff
@@ -0,0 +1,139 @@
+---
+CIP: ????
+Title: Cardano Governance Identifier
```
```suggestion
Title: Governance Identifiers
```
For consistency with PR title at this time... removing redundant word `Cardano`, plus characterising a number of representations with plural in title.

rphair (commented):
@ashisherc thanks; looks like a great start & looking forward to introducing this at the next CIP meeting & hope you can be there for an initial discussion: https://hackmd.io/@cip-editors/93

CIP-XXXX/README.md
Ryun1: Maybe even "Governance Credential Identifiers"?

rphair: @ashisherc I think the latter is better because it's more specific and disambiguates from the many, many types of "identifiers" we will have in the Governance field.

CIP-XXXX/README.md
ashisherc: Tools is the right category

CIP-XXXX/README.md
ashisherc: This CIP also defines identifier and relevant bech32 prefix for a gov action, which is not a credential, I will keep it just `Governance Identifiers` for now.

CIP-XXXX/README.md
```diff
@@ -0,0 +1,148 @@
+---
+CIP: ????
```
Ryun1, line 2:
```suggestion
CIP: 129
```

Ryun1: and please remember to rename your CIP-XXXX directory

ashisherc: done

rdlrt:
Once this is merged, We would begin work for supporting it for CNTools and Koios API

gitmachtl:
SPO Scripts can support it too. But CLI and DB-Sync must also be onboard. For HW-Wallets, such a change will most likely not be implemented until the end of the year, the release cycle from the vendors (not the devs) is horribly slow nowadays.

rphair (commented):
@ashisherc it sounds like you can put the items provided by @rdlrt @gitmachtl on the Implementation plan, which will fill out that section to a bare minimum for continuing review.  It needed to be more specific anyway and I think it's even better to have specific ranges of work to focus on.

I would recommend tickboxes for the general categories (operator tooling, developer APIs) and then the projects above as specific examples in tickboxes nested under them.

Also based on @gitmachtl's assessment about hardware wallet support being slow, I think it would help to highlight that in the implementation plan: as a task to bridge with the responsible developers.  Therefore I think that section would also be better suited as tickboxes because of the list of things that are already mentioned there (anything done already can be pre-ticked).

If we have an updated _Path to Active_ before Tuesday we can get this again on the CIP meeting agenda & keep this moving along toward merge.  Thanks @rdlrt @gitmachtl for the community review & offers of support.

ashisherc:
@rphair updated the PR with the implementation plan, please provide context/recommendation for acceptance criteria.

CIP-0129/README.md
```diff
@@ -0,0 +1,162 @@
...
+## Rationale: how does this CIP achieve its goals?
+
+By introducing a header and payload design for different keys, we can achieve specifying credentials in a hex format (bytes), being able to share and use across tooling is possible. We also specify bech32 prefixes for user friendly identification of the identifiers.
+
+
+## Path to Active
+
+### Acceptance Criteria
+
+- [ ] implemented by various tools
```
rphair, line 139:
```suggestion
- [ ] Implemented by operator tooling, including:
  - [ ] StakePool Operator Scripts
  - [ ] CNTools
- [ ] Implemented in development libraries, including:
  - [ ] Koios API
```

rphair (commented):
> please provide context/recommendation for acceptance criteria

According to https://github.com/cardano-foundation/CIPs/pull/857#pullrequestreview-2215960341 (and pending further recommendations by @Ryun1 @Crypto2099)...

CIP-0129/README.md
ashisherc: this is added in implementation plan

ashisherc: https://github.com/cardano-foundation/CIPs/pull/857/commits/0f4cd78a424e19d7831225b79b53d7ca67ebc31d

rphair: @ashisherc the CIP process has a routine uncertainty here between Acceptance Criteria (criteria whose completion shows a CIP is `Active`) and Implementation Plan (the prerequisites & goals from CIP creation until declaring it `Active`).  We've tried to explain the difference here: https://github.com/cardano-foundation/CIPs/blob/master/CIP-0001/README.md#path-to-active

I made my last comment without realising that you had put your entire acceptance criteria (criteria list) as an implementation plan.  So you will have to
1. move that checklist up to Acceptance Criteria (where you will see it is in other CIPs);
2. produce a proper Implementation Plan.

Since # 2 depends on how you & your associates intend to work on the CIP, I think we should talk about the Implementation Plan at the next CIP meeting if you don't yet have any ideas to post here.  I've put it up for Review at the next meeting (https://hackmd.io/@cip-editors/94); if you can't attend it then I'll bring up the issue myself and note any suggestions back here.  cc @Ryun1

dmitrystas:
good stuff, AdaStat will definitely support this CIP

CIP-0129/README.md
```diff
@@ -0,0 +1,161 @@
...
+#### DRep Credential
+
+Key - `00000000000000000000000000000000000000000000000000000000`
+Type - `Key Hash`
+
+Identifier - `2200000000000000000000000000000000000000000000000000000000`
+
+Bech32 - `drep1ygqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqq7vlc9n`
+
+## Rationale: how does this CIP achieve its goals?
```
Ryun1, line 130: I believe it would be valuable to describe the ways that this differs from the existing CIPs, and why this approach is preferable.

CIP-0129/README.md
```diff
@@ -0,0 +1,161 @@
...
+- [ ] APIs
+  - [ ] Koios
+  - [x] Cardanoscan API
+  - [ ] Blockfrost
+- [ ] Explorers
+  - [x] Cardanoscan.io
+  - [ ] AdaStat.net
+- [ ] Wallets
+  - [ ] Eternl
+  - [x] Typhon Wallet
```
Ryun1, line 154:
```suggestion
  - [x] Typhon Wallet
  - [ ] Lace
```

Ryun1: Probably worth adding a few more

ashisherc:
> Id suggest we move to make this standard an "extension" to CIP105, so that it optionally adds to the existing standard, in a backwards compatible way

After extensive discussions on CIP105, it was determined that this specification should be its own CIP, as it is not related to CIP105 (which specifies the derivation for new Conway era keys). Confusing this CIP as a backward-compatible extension of CIP105 is a mistake. This CIP establishes a new specification for various Conway era credentials generated using CIP105.

> I believe at this stage of maturity for governance tooling having competing standards might not be ideal, potentially damaging user experience and compatibility.

As specified in the CIP, no existing toolings need to do extensive work to support this CIP, they would only have to switch to a proper bech32 prefix. I do not believe it is too much work just to replace `drep` -> `drep_vkh` for existing tools. I wouldn't want to create so much chaos but the suggested new prefix aligns with the specification of the CIP105 itself.

CIP-0129/README.md
ashisherc: @Ryun1 detailed in the Rational and Motivation

rdlrt:
> I believe at this stage of maturity for governance tooling having competing standards might not be ideal, potentially damaging user experience and compatibility.
I am worried that we reach the Chang hardfork (estimated end of August) and half of tooling represents DRep IDs one way and the other half of tooling represents it another way.

If this has to go through, sooner the better, to prevent the exact scenario mentioned. Currently the support for conway across tooling is a bit premature, with all exploration of edge cases requiring frequent fixes/changes. More mature the tooling becomes, more resistance there would be for changes later on.

CIP-0005/README.md
```diff
@@ -92,35 +92,38 @@ We define the following set of common prefixes with their corresponding semantic
...
+| `drep_vkh`             | Delegate representative verification key hash                         | blake2b\_224 digest of a delegate representative verification key        |
+| `drep_script`      | Delegate representative script hash                                   | blake2b\_224 digest of a serialized delegate representative script       |
+| `cc_cold_vkh`          | Constitutional committee cold verification key hash                   | blake2b\_224 digest of a consitutional committee cold verification key   |
+| `cc_cold_script`   | Constitutional committee cold script hash                             | blake2b\_224 digest of a serialized constitutional committee cold script |
+| `cc_hot_vkh`           | Constitutional committee hot verification key hash                    | blake2b\_224 digest of a consitutional committee hot verification key    |
+| `cc_hot_script`    | Constitutional committee hot script hash                              | blake2b\_224 digest of a serialized constitutional committee hot script  |
 
 ### Miscellaneous
```
Ryun1, line 116: need to add the `gov_action` definitions also

Ryun1:
@ashisherc 

> After extensive discussions on CIP105, it was determined that this specification should be its own CIP, as it is not related to CIP105 (which specifies the derivation for new Conway era keys). Confusing this CIP as a backward-compatible extension of CIP105 is a mistake. This CIP establishes a new specification for various Conway era credentials generated using CIP105.

Sorry, I wasn't clear what I meant by "extension", my bad.
I didn't mean that this shouldn't be its own proposal,
I meant that this proposal could be backwards compatible with CIP105, building on top of the existing standard directly.

>  I do not believe it is too much work just to replace drep -> drep_vkh for existing tools. I wouldn't want to create so much chaos but the suggested new prefix aligns with the specification of the CIP105 itself.

Would also need changes for tools who have implemented DRep scripts and CC key/script credentials,
although, point taken this isn't a massive lift.

There is a danger here of some tools who have implemented CIP105 not seeing these changes,
resulting in their tooling being broken.
I do worry about this,
we have seen this a few times in CIPs.

Ryun1:
@rdlrt 

> If this has to go through, 

This is a tricky one,
I would say it is up to the community to decide the necessarily

Personally, I think these changes do bring value
but I also see risk of them, as described

> sooner the better, to prevent the exact scenario mentioned.

This is a good point, I agree.

ashisherc:
@Ryun1 

> There is a danger here of some tools who have implemented CIP105 not seeing these changes, resulting in their tooling being broken.

This is part of upgrades, its like not upgrading to Chang bcoz existing dApps may not notice new changes. I know this is an exaggeration but the point is there are very limited no of tools that deal with drep/cchot/cccold key generation. Once this CIP is merged, we can do the needful to reach out to these tools and request for upgrades. Most of the tools which will support Chang upgrade are already listed in the acceptance criteria. We will also be linking to this CIP from Cardanoscan so that devs can easily find the spec used by us for generating drep and other IDs.

CIP-0005/README.md
ashisherc: done

gitmachtl:
HW-Wallet changes will take a longer time, not sure how to deal with it in the meantime. Because with the SPO Scripts, i always show the same data on screen like bech32 representations as it appears on the HW-Wallet gui. If those two differ, thats a problem.

ashisherc:
@gitmachtl added some details in the implementation plan for this. During the migration period, the software UI can show both the formats with a slightly greyed out font/color mentioning deprecated format. Essentially informing user to refer it for verifying on a HW wallet.

janmazak:
(comments from me and David Misiak for VL)

It's not great that the `drep` prefix will change its meaning.

Where is a DRep to be used in the form of whole drep credential, and where we should show drep_vkh instead? If one tool (e.g. a SW wallet) uses one representation, but another tool (e.g. a HW wallet) the other, it is confusing for the user at best --- one would be unable to compare the displayed data (to some extent --- the bytestrings encoded into bech32 will overlap). For instance, when ones has a script drep, should `drep_script` or `drep` (with the "script" type encoded in the subsequent bytestring) prefix should be used? This CIP should give unambiguous guidelines for UI design. 

The design in the proposed form is not coherent; it's inspired by addresses, but that's kind of a bad analogy; addresses are tied to network and serve a different purpose.
An already identified problem is that Cardano has no consistent notion of a "key / key hash / credential" object.
It transpires for instance in the "address" field in CIP-8 (message signing) where one wants to put in addr key hash, but instead a whole address goes in (i.e. one extra byte for address header + key hash), but for DRep key, only a key hash is used currently (i.e. no extra header byte). It would make sense to unify it all, but the proposed CIP does not do it (it ignores credentials from staking and multisig keys etc.). It also introduces unsystematic coupling ("We *reserve* the values 0 and 1, as these are used in the Cardano Address Network tag") --- if someone in the future adds another value for network tag (say, another specific kind of testnet), it could clash with this CIP.

I'd say that DApps / script authors / UI developers would welcome if Cardano had a unified notion of credential, which would cover all possible key hashes and script hashes, also those potentially added in the future, and a standardized way of displaying it (here I'd be happier with more than 16 possibilities as is currently suggested in the CIP by reserving just 4 bits). So this CIP might want to address not just drep and cc keys, but cover payment and stake keys, too. I have not thought it through at all, though, so I am not sure it is possible to design it in an easily implementable fashion.

hw-cli is not mentioned in the "path to active" list (maybe not affected? --- it does not do UI, but some command line arguments perhaps? but at least new version of ledgerjs/trezor connect would likely be required).

CIP-0129/README.md
```diff
@@ -0,0 +1,167 @@
...
+This specification describes the structure and the bech32 format of the identifiers for DRep, CC keys, and Gov Actions
+
+## Motivation: why is this CIP necessary?
+
+Cardano Conway era introduces new credentials for different actors, and introduces proposal procedures for governance actions. It is required to be able to find, link, share these identifiers in bytes form across the tooling and community.
+
+## Specification
+
+### Introduction
+We define a bytes representation for the various credentials and identifiers along with the their bech32 prefixes in this CIP. Taking inspiration from the Cardano addresses bytes format, we define an 8 byte header and a payload to define the key, which look similar to the reward address byte form but with a new specification and using the governance credentials.
```
Ryun1, line 27: What is the rationale for not just reusing how the Address headers as defined via [CIP-19](https://github.com/cardano-foundation/CIPs/tree/master/CIP-0019#binary-format)?

Why make a new definition for a header?

rdlrt:
FYI - This is now being used in Koios

rphair:
@rdlrt thanks for following up on https://github.com/cardano-foundation/CIPs/pull/857#issuecomment-2264592368 (I'm noting this to help editors stay abreast of current implementations)

Scitz0:
I can also add that Eternl will make this change, to be applied in next update. 

Not that affected like other tools might be but it will display CIP-129 id in UI but take any of the versions as input. So if you grab old style id from some other source to delegate to within Eternl, it should still handle it and extract hash correctly.

CIP-0129/README.md
ashisherc: 1 - gov identifiers are not addresses
2 - The current address spec header does not have enough space to accommodate these many identifiers

Jimbo4350:
> But CLI and DB-Sync must also be onboard.

I'm just seeing this CIP and I'm not a fan (at this point) because we are now straying from the CDDL. We've been down this road before in `cardano-api` and it caused confusion because we had an "intermediate tx body format" that differed from the ledger's CDDL.

Why do we need a header when we can piggy back on bech32 i.e [https://github.com/cardano-foundation/CIPs/tree/master/CIP-0105](CIP-0105)?. I know this describes wallet derivation however it introduces new prefixes for the governance credentials.

I can get behind this CIP (or some iteration of it) if the intention was to make it part of the CDDL. However, this route will require input from the ledger team.

rphair:
@WhatisRT @lehins in light of @Jimbo4350's https://github.com/cardano-foundation/CIPs/pull/857#issuecomment-2356924249 please review & post here so we can remain aware of whatever Ledger impact there would be for this CIP.

ashisherc:
https://github.com/IntersectMBO/cardano-ledger/issues/4441 some decisions were taken without considering future implications. this CIP does solve a few things, and major things for ecosystem tooling.

rdlrt:
The CIP was created well prior the date of the fork (we still had to have a date for fork on preprod) , if any chance was present to put it to ledger,  it'd have been at the time - I assume the suggestion from Jordan above is a bit more sarcastic way to say it wont be done (as now it not only will have a fork requirement, but will also make things muddy between pre/post fork) 🙂 

Having said that, for tooling - this did not even require dbsync/CLI/ledger to adopt.. since bech32 representation is best rendered on client-side.
I usually keep my expectation bar very low for any changes upstream due to the complicated jungle it has become - so the only question is , whether the builder/tooling community have an expectation to continue down this route, in spite of conflict - and how can we move forward (to get to approve/reject/modify stage)

gitmachtl:
this CIP is more on the frontend/tooling/user side imo. ledger stores it already differently for drep key hashes and drep script hashes. tooling/explorers/etc have already jumped on it because we saw the first issue on chain that people have used the wrong bech representation for a drep script. on the CLI side, we have the keyHash/scriptHash in the queries. there is not a single output that is showing a bech representation. so i don't see an issue here atm.

gitmachtl:
> > But CLI and DB-Sync must also be onboard.
> 

After months now implementing this on the tooling side, i don't see the need for a change on cardano-cli output side atm. We have the different outputs for keyHash/scriptHash. If it would be changed, than we would need to totally flatten it and only have a single Hash format with the infobyte infront. But it could be added for the input parameter option on cardano-cli, that would be not a different task for cardano-cli to recognize hashes and bechstrings that are that 1 byte longer and handle it internally to use it as a drep keyHash or scriptHash.

dmitrystas:
the same for db-sync. especially considering there are conversations going on now to remove the text (bech32) representations from tables, and leave only binary data

WhatisRT:
As pointed out by previous comments, there is nothing do to for ledger with this CIP. Any tooling could support this CIP tomorrow without a HF.

I agree with the sentiment that the CDDL is suboptimal for this purpose, and I'm supportive of @ashisherc's opinion that it might have been better to not inline the definitions to save a few bytes. We have no idea how much votes we'll actually get, it seems like a premature optimization to me. But it's done now, and the only way to change it would be community feedback.

CIP-0129/README.md
```diff
@@ -0,0 +1,169 @@
...
+**Example 1**
+TX ID in Hex - `0000000000000000000000000000000000000000000000000000000000000000`
+Gov Action index in HEX - `11` (number 17)
+gov action ID - `000000000000000000000000000000000000000000000000000000000000000011`
+bech32 - `gov_action1qqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqpzklpgpf`
+
+**Example 2**
+TX ID in Hex - `1111111111111111111111111111111111111111111111111111111111111111`
+Gov Action index in HEX - `00` (number 0) 
+gov action ID - `111111111111111111111111111111111111111111111111111111111111111100`
```
lehins, line 86: Gov action id is 2 bytes long
```suggestion
gov action ID - `11111111111111111111111111111111111111111111111111111111111111110000`
```

lehins:
> @WhatisRT @lehins in light of @Jimbo4350's https://github.com/cardano-foundation/CIPs/pull/857#issuecomment-2356924249 please review & post here so we can remain aware of whatever Ledger impact there would be for this CIP.

This CIP describes textual representation of hashes used in transactions. In CBOR we store hashes in binary format, so textual prefixes or any other offline representation (like bech32) have no business being encoded in the on-chain data.

> .. I'm supportive of @ashisherc's opinion that it might have been better to not inline the definitions to save a few bytes. 

When choosing between two legal isomorphic representations in the binary format we'll almost always choose the more efficient one. CBOR is meant to be decoded by tools, not read by people, so for any decent programmer it should not make any difference on how the data is encoded as long as it is clear how to decode it and CDDL spec describes it unambiguously. Binary represntation needs to be compact and efficient, it does need to be pretty. We will have millions of votes and these few bytes savings will manifest in megabytes of savings in the on chain data. So, @ashisherc and @WhatisRT I respectfully disagree with both of you.

My take on this CIP is that has nothing to do with Ledger or CDDL. This is 100% tooling issue.

CIP-0129/README.md
ashisherc: there is no fixed length limit for index bytes, it just needs to be a valid hex at the end of a tx ID (fixed) bytes

ashisherc:
@lehins thanks for your valuable inputs

Clarification—This CIP's goal was never to change the CDDL or ledger spec. It has always been an off-chain spec for byte representation and bech32 encoding for various government-related credentials and identifiers. We just realized that this spec may also be helpful at the ledger level.

Since you mentioned the **storage efficiency**, it has been long exploited by tx metadata, it is irrelevant I know. But let us see the existing CDDL for the gov credentials itself,

**Gov Action ID** - https://github.com/IntersectMBO/cardano-ledger/blob/a44a5b5ac9f68c5d7a2ff38b040a5fcac8ccfc6a/eras/conway/impl/cddl-files/conway.cddl#L148

Instead of having a list of tx ID, and index bytes, it should bring efficiency by just using this CIP action ID, since you mention there will be millions of votes, and each vote will also have an action ID. Exactly same data structure even for various voters.

**DREP/CC Credentials** https://github.com/IntersectMBO/cardano-ledger/blob/a44a5b5ac9f68c5d7a2ff38b040a5fcac8ccfc6a/eras/conway/impl/cddl-files/conway.cddl#L338

Storage efficiency is also impacted if we keep these creds the way they are right now, especially when delegating to these dreps. This CIP also makes it efficient to store such credentials.

Again, I am not suggesting we do the changes at the ledger level, as it isn't the goal of this CIP, it is only an off-chain tool spec.

lehins:
> This CIP's goal was never to change the CDDL or ledger spec. It has always been an off-chain spec for byte representation and bech32 encoding for various government-related credentials and identifiers. 

Glad to hear, I was worried there for a moment :sweat_smile: 

> We just realized that this spec may also be helpful at the ledger level.

Any offchain representation that does not affect CDDL has no use to ledger, since only input we ever care about is encoded in CBOR.

> it has been long exploited by tx metadata,

Not quite sure what you mean there about metadata.

> Instead of having a list of tx ID, and index bytes, it should bring efficiency by just using this CIP action ID, since you mention there will be millions of votes, and each vote will also have an action ID.

Why stop there, `transaction_input` would benefit from this optimization even more :sweat_smile: 
I agree, that it would be more efficient, but that would not be valid CDDL, so we can't really use it on chain.

Currently, I believe, there are only two types that use custom binary representation instead of CBOR, which also means that they cannot be described in CDDL: payment addresses and plutus scripts.
* With respect to plutus scripts, I think it is totally fine to have custom serialization.
* When it comes to addresses I can't tell what was the reason for using custom format. Looks like it was initially implemented with CBOR, but then replaced with custom format in https://github.com/IntersectMBO/cardano-ledger/pull/1412 Maybe it was motivated by efficiency, maybe by something else, I don't know because it predates everyone on the ledger team.

Unless there is a really really good reason to deviate from CBOR in favor of a custom binary format for some type(s) we are not even going to consider adding more cases to this exception list.

vsubhuman:
> It's not great that the `drep` prefix will change its meaning.

100% agreed. If this proposal has to exist at the very least it can extend the courtesy to not inflict standard conflicts that cause nothing but confusion for users. A custom bech32 prefix should rather be used for a custom bech32 encrypted content. An already merged convention defines `drep1<key>` meaning that the `<key>` value will be a plain `Ed25519KeyHash`. This proposal breaks it, while it doesn't have to. It can be changed to something like `drepx1<val>` meaning the `<val>` will be that "extended hex" value with encoding prefix byte.

Scitz0:
While drep bech32 prefix does conflict with merged CIP-105, they are distinguishable by the length of the bech32 drep id. I just wanted that said.

vsubhuman:
> While drep bech32 prefix does conflict with merged CIP-105, they are distinguishable by the length of the bech32 drep id. I just wanted that said. 

Ima tell users to make sure they check the length of the ID they are getting so they can know if all apps will accept it.

Scitz0:
> > While drep bech32 prefix does conflict with merged CIP-105, they are distinguishable by the length of the bech32 drep id. I just wanted that said.
> 
> Ima tell users to make sure they check the length of the ID they are getting so they can know if all apps will accept it.
> 
> Or we can just agree to use custom prefixes for custom values and make things immediately easier for everyone.

I have no problem with your suggestion, only meant that from a programmatic perspective, it can be used to distinguish them. But using unique prefix would make it even clearer, especially from user pov.

gitmachtl:
i think that CIP129 is ok as it is right now. also already adopted by a bunch of apps. CIP105 was rushed, not CIP129. i pointed out that we have to make this right 1,5 years ago. CIP105 failed to deliver it, but we used it anyways. a lot of tooling has CIP129 now in tegrated, i don't see a reason to change this all over again. imo it would be better to switch over to CIP129 and don't use CIP105 anymore.

vsubhuman:
> imo it would be better to switch over to CIP129 and don't use CIP105 anymore.

I will now just out of making a point say that IMO an opposite should happen. Even tho I don't really care. What now? We can keep having arguments about which thing is better or not and have this never merged. Or we can just stop having hard feelings about standards and agree that if some other proposal got merged first and reserved a use of a certain prefix for a specific case then other proposals that are not merged yet can adapt and make life easier for everyone including themselves , increasing the chances to get globally adopted, but at the very least being backward compatible with the existing state of the things. This proposal can then get merged and get adopted by all the projects that claimed they will be ready to implement it when that happens, and let CIP105 naturally die out over time, if that is what meant to be. Anyone who implements an unmerged proposal should always be ready it might change, that's the whole point and why so many projects are waiting for this to get merged before adopting it.

gitmachtl:
@vsubhuman @ashisherc @Ryun1 to keep this conversation going. is the only thing that is holding this pr up from being accepted the overlap in the bech prefix?

the `gov_action` prefix was introduced by this cip-129 draft. so there is no need to change that.
for the others we would have to change them like:

```
   drep -> drepx
 cc_hot -> cc_hotx / ccx_hot
cc_cold -> cc_coldx / ccx_cold
```
?

ashisherc:
@gitmachtl this PR updates the previous CIP prefixes. Hence once merged, it will avoid confusions.

gitmachtl:
@ashisherc i don't see any willing from IO side to update there tooling, so most likely this PR will go nowhere anytime soon if we overwrite the prefixes.

rphair (approved):
> is this PR going to get merged and accepted?

yes @gitmachtl I think it is about time: at the very least, to have a last call for "pressing the panic button" at Tuesday's CIP meeting (https://hackmd.io/@cip-editors/100) where I myself will argue in favour of finally merging it, and so marking this `Last Check` (*p.s.* @Ryun1 I can see you did that 5 days ago; so let's consider it "double" Last Check).

The feedback from Ledger (@WhatisRT @lehins) points out some anomalies but also confirms this would break nothing there.  According to my own multiple readings of the comment history, tooling at this time is mainly compliant with the proposal, and any quirks about key prefix or length can likewise be accommodated by those tools.

In the worst case we would have a mixed ecosystem in that regard, with tools having to advertise CIP-0105 and/or CIP-0129 compliance.  If this does happen, it wouldn't be the first or worst time such a thing happened in Internet standards & then relied upon time itself to smooth out any discrepancies.

Those are my main reasons for wanting to this on the books (though I have no technical or aesthetic preferences on these issues myself): especially now that any reservations have been recorded in the `Discussion:` link in the header.

HeinrichApfelmus:
I'm working on `cardano-wallet`, the business logic for Daedalus, and randomly came across this CIP as part of implementation questions.

To clarify: Am I understanding this right that merging this pull request will _retroactively change_ the active CIP-105 standard to no longer define the `drep` prefix?

I'm asking because this means that the phrase "our wallet is compliant with CIP-105" will mean different things at different points in time, causing confusion for users?

l-br1:
Hello everyone here :) 

Apologies for coming late to the party. I just wanted to share a perspective from a pure user experience side (I work as Head of product in Intersect and have facilitated the creation of Govtool so far).

### Context
We are seeing good participation in governance so far, considering also we are still in the bootstrapping phase, but there has been confusion on the DRep side about registration and how to successfully campaign. Tools (like also govtool) make it a bit easier, but of course, we need also to allow DReps to socialize their ID to people who will not use tools (arguably the delegators with bigger stakes). 

### Suggestion
Thinking again purely from a user experience, and so a participation point of view, I think it is critical, especially now that we are at the beginning and we need to increase DReps and delegation, to keep information clean and easy. To me this means aligning on one standard (whichever you all believe to be best) otherwise every DRep would need to share two IDs (some of which visible in some tools and some not) and every delegator would need to be aware of which ID to use where. **It becomes difficult to manage for DReps, confusing for delegators, and difficult to handle also for tools.**

### TDLR
If possible, I would recommend investing a bit more time to align all on a path forward that leads to one shared standard so that we can all support good participation in our Cardano government 🙏🏻

rphair:
@HeinrichApfelmus @l-br1 - I don't know to what extent it will be possible to satisfy everybody's reservations about the implications of merging this proposal.  There is a lot of commentary already in the thread above suggesting that it's worth the risk... or disqualifying the risk to some degree... but that doesn't mean it would be merged without considerations to the contrary, or that we would simply accept "majority rule".

All I can offer for now is that we will include all feedback up until that time, including the comments you have just made.  There will be 2 other editors involved in that decision who haven't expressed opinions recently, and you yourselves (as would all interested parties) are welcome to attend that meeting.

l-br1:
Hi @rphair, I fully understand it's not easy to make everyone happy, and I appreciate your reply and the invitation to the meeting 🙏🏻

If anyone has a proposal that could allow this new standard to be compatible with the existing one or if we see a safe way (and good reasons) to deprecate the existing one for the new one, that would be great. Of course, I am not aware of all the details related to the benefits of this change so I'll rely on the expertise of this group, if they see the value in it, to see if there is such a path forward 🙂

HeinrichApfelmus:
> the implications of merging this proposal.

I'd like to make my feedback more precise then: I don't care about the technical content of either this CIP or the other one, but I have an objection on the matter of _process_.

Specifically,

* merging this pull request

does two things in combination:

1. Accepting CIP-129, introducing a meaning for human readable strings of the form `drep1…` 
2. Revoking CIP-105, as the wording of the standard is changed to no longer define a meaning for `drep1…`.

The consequence will be that if I make the claim "our wallet is CIP-105 compliant", this claim is _true_ today, but will become _false_ after item 2 has happened. I will have to at least update my website, which however is unmaintained, so the false claim that was once true will persist.

I don't even have an objection against revoking standards, but I request that this be an _explicit_ action. One possible course of action is to do the following:

* Retire CIP-105
* Accept CIP-129
* Accept a CIP-130 (or some other number) which is equal to the now-retired CIP-105, except with the wording change here.

Anyone who wants to know more about the fact that "our wallet is CIP-105 compliant" will then have a chance of finding out that the wallet is outdated, as the standard that it is compliant has been retired.

CIP-0129/README.md
cryptophonic: I found the Abstract and Motivation might have been missing some background in order (for me) to to understand the issue here.  Let me try to summarize in order to verify my understanding:

1.  CIP-105 defines bech32 identifiers with the key (CC hot, CC cold, DREP) and credential (script hash, public key hash) defined by the bech32 prefix only.  In CIP-105 the binayr payload for the bech32 encoding is always just the key material itself.

2.  I am inferring from the discussion thread that various infrastructure and tools that use these identifiers decode the bech32 identifier and store the key material internally as the bytes themselves (presumably for space and efficiency reaons), thereby losing the metadata that was encoded in the bech32 prefix specified in CIP-105.

3.  This CIP-129 defines a single byte header to be prefixed to the key material prior to bech32 encoding that would include this metadata.  Note this would also be slightly redundant if the CIP-105 prefixes are used.

Since the primary purpose of bech32 encoding is for human-readable transmission, it does make sense to add a binary level encoding of the key type and credential metadata.  Unless I'm missing something it also seems that the single byte header can be inferred from CIP-105 -> CIP-129 if it is missing, through the bech32 prefix.

ashisherc:
@HeinrichApfelmus Regarding retiring 105, this proposal doesn't retire 105, only updates a few parts of it mainly regarding the dRep and CC IDs. Further merging this proposal will not affect existing toolings compatibility right away. Infact checkout the tools readiness and most of the tools support both CIP129 and 105 as backward compatibility. Once this proposal is merged, tools can slowly start removing 105 compatibility at their convenience.

Hence to summarise majority of the tooling rightnow are supporting both 129 and 105, and waiting for this proposal to be merged, to avoid maintaining 105 further and reduce confusions that's are being created.

gitmachtl:
Let me allow to make this comment: I know that this question is now driven from the Daedalus Update. Daedalus is always a bit slow when it comes to integrate new CIPs and functions that are picked up by other wallet way faster. Having no support for Transaction-Messages/Memos is only one of them. 

What i would like to see here is, that cardano-wallet as the underlying component powering Daedalus adobt such a new CIP faster. From what i have heard so far from CIP editors, the CIP129 is on a good path to be merged soon. No real blockers there, also a lot of adoption already. 

I konw that the dev works on various components are not that easy or not that highly rated when it comes to distribution dev power to it. Just wanna say that i have the fear if this will not be implemented right from the beginning, it will take months before we see it coming to the public.

HeinrichApfelmus:
> Once this proposal is merged, tools can slowly start removing 105 compatibility at their convenience.

Again: Once this proposal is merged, tools that were CIP-105 compliant will no longer be CIP-105 compliant, because their use of the `drep` prefix will no longer agree with the new wording of the CIP-105 standard. Tools cannot slowly remove compatibility even if they wanted to — it will be revoked by merging this pull request.

As I said, this is an issue with the standards process, not with any particular technical choices.

gitmachtl:
@HeinrichApfelmus If something is compliant with a CIP and a newer CIP comes along, than the product is still complaint with the original CIP. So thta does not invalidate it imo.

I think what we should do - and i already talked with @ashisherc (cip author) about this - is to simply add a seconf line in CIP005 describing the prefix with two different use cases. The prefix `drep` can exist in both domains. There is currently no need to define one CIP to rule them all.

So you can still keep your statement to be CIP105 complain for a product.

gitmachtl:
@ashisherc is about to update it like i described above.

Ryun1:
Hi all,
Appreciate the discussion on this.

### Context
- CIP105 provided definition for Conway credentials, including bech32 prefixes
  - Prefixes for DRep credentials and CC credentials.
- Some tooling/wallets have implemented the CIP105 definitions
  - CIP105 definitions have seemed to broadly worked
- A different approach to Conway credentials was proposed via CIP-129
  - These definitions provide a nicer representation for these credentials
- CIP-129 does suggest altering the CIP-105 definitions of bech32 prefixes
  - This is an issue as it requires a change to an active CIP

### Options
a) We change CIP-105 to not claim the prefixes.
b) We change CIP-129 to not claim to the prefixes.
c) We agree that CIP-105 and CIP-129 to both claim the prefixes.

### My opinion
I oppose a) as it would mean tooling which is "CIP-105 compliant" would now be broken.
Changing active CIPs has proven to be a mistake time and time again.

Myself and others have suggested b) a couple of times.
This is of course at the author's discretion, and they have chosen not to take this suggestion, I think that is fine.
Although I still think it is worth considering.

With the current context I think c) is the only possible action.
With the caveat that all CIP129 compliant tools should also recognise CIP-105 prefix definitions and handle them.
c) allows for tooling to migrate between the two standards, without outright breaking CIP105 compliant tools.
This means that things could only break when users are copying CIP-129 IDs to CIP-105 tools.
And this seems to be the consensus reached by @gitmachtl + @ashisherc whilst I was writing this response.

Does anyone object to allowing both CIP105 and CIP129 claim these bech32 prefixes?
(tagging a few discussion contributors @HeinrichApfelmus, @l-br1, @vsubhuman, @rphair, @Crypto2099)

rphair:
@Ryun1 - from my understanding of the discussion so far (and not as a tool developer) I have no objection to option (c) and believe it represents the best chance of moving forward together as a community over this issue.

vsubhuman:
> @Ryun1 - from my understanding of the discussion so far (though not a tool developer) I have no objection to option (c) [CIP-105 and CIP-129 both claim the prefixes] and believe it represents the best chance of moving forward as a community over this issue.

Same. Not perfect, but seems like a way forward. At the very least it makes the CIP recognize that prior state exists and cannot just be erased from documentation.

HeinrichApfelmus:
@Ryun1 Thanks! This makes a lot of sense.

I don't object to c), but I do want to note has the unfortunate and large downside of bringing confusing to users — "Hey, your drep doesn't work" — "Did you use a CIP-129 or CIP-105 compliant wallet?" — "What's a CIP wallet?"?

The only solution to that is to also retire one of the standards in the long term, e.g.

d) We retire CIP-105

Doesn't have to be now.

gitmachtl:
As i suggested above, i am also for c). 

But i would also go further and add in some information in the CIP-005 documentation for CIP105 entries, that there is now a newer version and that CIP105 might get deprecated in the future. So thats about d) @HeinrichApfelmus suggested.

l-br1:
Thanks @Ryun1 for the summary 🙏🏻 

From my side (considering participation in Cardano's governance, DReps and delegators' needs and so needs tools should support) I would still stress that effectively creating an 'alternative' DRep ID will require DReps and delegators to be aware of both (without any real benefit and reason for them) and it will add extra unnecessary friction to an already complex experience at its early stages. 

**So purely considering the perspective of users of our governance system I see this as a step back.**

Option C is definitely a way to unblock this easily (just looking at the CIP itself), but even from the rationale and summary in @Ryun1's comment, it doesn't seem the best option for the broader purpose.

gitmachtl:
@l-br1 in regards of user experience and that "Dreps" must be aware about an other format. In my oppinion its best to do this now, and not wait and sit it out until we have 1000+ DReps on chain. 

We should not artificially delay the depoyment of a better solution. CIP129 brings bech representation for ActionIDs too. Thats very nice.

It also provides the solution to not have to deal with 

drep
drep_script
cc_cold
cc_cold_script
cc_hot
cc_hot_script

credentials, but reduces that to only three from the six.

I am a bit biased about this whole topic, because i brought up this issue around 1,5 years ago. But noone reacted and it was passed and so only the direct hashes were implemented from the ledger/node/cli team. When the problem occured again, it was stated that "now its too late".

So, i hope we can move forward with this CIP and to provide a reseaonable upgrade path for the Tools that way.

CIP-0105/README.md
```diff
@@ -111,8 +111,8 @@ DRep keys and DRep IDs should be encoded in Bech32 with the following prefixes:
 | `drep_vk`     | CIP-1852’s DRep verification key                        | Ed25519 public key                                                |
 | `drep_xsk`    | CIP-1852’s DRep extended signing key                    | Ed25519-bip32 extended private key                                |
 | `drep_xvk`    | CIP-1852’s DRep extended verification key               | Ed25519 public key with chain code                                |
-| `drep`        | Delegate representative verification key hash (DRep ID) | blake2b\_224 digest of a delegate representative verification key |
-| `drep_script` | Delegate representative script hash (DRep ID)        | blake2b\_224 digest of a serialized delegate representative script |
+| `drep_vkh`    | Delegate representative verification key hash           | blake2b\_224 digest of a delegate representative verification key |
+| `drep_script` | Delegate representative script hash                     | blake2b\_224 digest of a serialized delegate representative script |
```
Ryun1, line 115:
```suggestion
| `drep` | Deprecated.... | |
| `drep_script` | Delegate representative script hash                     | blake2b\_224 digest of a serialized delegate representative script |
```

Crypto2099, line 115:
```suggestion
| `drep` | [DEPRECATED] Delegate representative verification key hash (DRep ID) | blake2b\_224 digest of a delegate representative verification key |
| `drep_vkh`    | Delegate representative verification key hash           | blake2b\_224 digest of a delegate representative verification key |
| `drep_script` | Delegate representative script hash                     | blake2b\_224 digest of a serialized delegate representative script |
```

CIP-0105/README.md
```diff
@@ -124,7 +124,7 @@ Constitutional cold keys and credential should be encoded in Bech32 with the fol
 | `cc_cold_vk`     | CIP-1852’s constitutional committee verification signing key          | Ed25519 private key                                                    |
 | `cc_cold_xsk`    | CIP-1852’s constitutional committee cold extended signing key         | Ed25519-bip32 extended private key                                     |
 | `cc_cold_xvk`    | CIP-1852’s constitutional committee extended verification signing key | Ed25519 public key with chain code                                     |
-| `cc_cold`        | Constitutional committee cold verification key hash (cold credential) | blake2b\_224 digest of a consitutional committee cold verification key |
+| `cc_cold_vkh`    | Constitutional committee cold verification key hash (cold credential) | blake2b\_224 digest of a consitutional committee cold verification key |
```
Crypto2099, line 127:
```suggestion
| `cc_cold`        | [DEPRECATED] Constitutional committee cold verification key hash (cold credential) | blake2b\_224 digest of a consitutional committee cold verification key |
| `cc_cold_vkh`    | Constitutional committee cold verification key hash (cold credential) | blake2b\_224 digest of a consitutional committee cold verification key |
```

CIP-0105/README.md
```diff
@@ -137,12 +137,12 @@ Constitutional hot keys and credential should be encoded in Bech32 with the foll
 | `cc_hot_vk`     | CIP-1852’s constitutional committee verification signing key          | Ed25519 private key                                                   |
 | `cc_hot_xsk`    | CIP-1852’s constitutional committee hot extended signing key          | Ed25519-bip32 extended private key                                    |
 | `cc_hot_xvk`    | CIP-1852’s constitutional committee extended verification signing key | Ed25519 public key with chain code                                    |
-| `cc_hot`        | Constitutional committee hot verification key hash (hot credential)   | blake2b\_224 digest of a consitutional committee hot verification key |
+| `cc_hot_vkh`    | Constitutional committee hot verification key hash (hot credential)   | blake2b\_224 digest of a consitutional committee hot verification key |
```
Crypto2099, line 140:
```suggestion
| `cc_hot`        | [DEPRECATED] Constitutional committee hot verification key hash (hot credential)   | blake2b\_224 digest of a consitutional committee hot verification key |
| `cc_hot_vkh`    | Constitutional committee hot verification key hash (hot credential)   | blake2b\_224 digest of a consitutional committee hot verification key |
```

CIP-0129/README.md
```diff
@@ -0,0 +1,170 @@
...
+
+- For key generation tools - To move faster to this CIP, current tools which implement CIP105 based bech32 prefixes do not have to implement this CIP specification. They can change the prefix according to the updates suggested in this CIP, updating bech32 prefix will be a find and replace sort of an updated and it will instantly become compatible with this CIP.
+
+- Tools like explorers and wallets - These tools can potentially support both formats to start with for the purpose of allowing users to search for drep, cold keys, hot keys as these 3 are impacted by this CIP upgrade. And can continue to show only this CIP specification, having an easy backward compatibility as well moving to this CIP standard.
+
+- This CIP does not require a hard fork for implementation, the goal is to use the identifies specified in this CIP for the UI, and as a medium of communication for sharing such keys and IDs (though it will be good if this CIP is also enforced at the ledger level, simplifying CDDL and bringing data storage efficiency).
```
lehins, line 166: Please remove this statement, since that would be a major breaking change and it would require a whole new CIP. Most importantly, I believe, it would a bad idea to implement something like this in ledger because:
* contrary to what it says, it would reduce storage efficiency since every credential would require one more byte in serialized form for the header and we have a lot of credentials on chain. There is a way to modify the existing one byte tag that we use for this in order to not incur new overhead, but there is definitely no way to improve storage efficiency when it comes to credentials.
* Changing serialization for an already existing type would have to span a few ledger eras and would result in a lot of tools being broken, **including all hardware wallets**.
* Such change is not needed for ledger, so we should not affect ledger semantics.
* There is no benefit in such change for the on chain representation or ledger users.
* bech32, prefixes or other credential identifiers are only useful for end users and tooling that faces them, so CDDL and ledger should be kept out of it. There is nothing wrong, from the ledger perspective, for users using the same cryptographic keys or same scripts for multiple entities.
```suggestion
- This CIP does not require a hard fork for implementation, the goal is to use the identifies specified in this CIP for the UI, and as a medium of communication for sharing such keys and IDs.
```

Crypto2099 (commented):
A couple of minor notes about leaving the previous prefixes in CIP-105 and flagging them as DEPRECATED instead so that we do not break any existing compatibility with this change. Everything else looks good

rphair (commented):
also @ashisherc there is a merge conflict (likely for the last few weeks, since I think it's from https://github.com/cardano-foundation/CIPs/pull/920) that should be resolved in your branch. **... p.s.** _please_ without a force push, so we can preserve all the pending changes & edits 🙏

CIP-0105/README.md
Ryun1: Might be worth referencing CIP-129 where we say depreciated

CIP-0105/README.md
```diff
@@ -200,6 +200,26 @@ For hardware implementations:
 | `ConstitutionalCommitteeHotHWSigningFile_ed25519`   | Hardware Constitutional Committee Hot Signing File     |
 | `ConstitutionalCommitteeHotVerificationKey_ed25519` | Hardware Constitutional Committee Hot Verification Key |
 
+### Deprecated Governance ID Definition
+The previous governance key IDs defined by this have been upgraded as per the new specification introduced in [CIP-0129]. Tools implementing the following specification should gradually adopt the updated ID format outlined in [CIP-0129]. Tools that already support [CIP-0129] maintain backward compatibility with the legacy formats specified below but should consider fully transitioning to [CIP-0129] to standardize key formats across the ecosystem. This will help avoid multiple formats and ensure consistency.
```
Ryun1, line 204:
```suggestion
The previous governance key IDs defined by this standard have been superseded by the definitions provided in [CIP-0129]. Tools implementing this standard are encouraged to consider adopting [CIP-0129]. Tools that already support [CIP-0129] maintain backward compatibility with the legacy formats specified below but should consider fully transitioning to [CIP-0129] to standardize key formats across the ecosystem. This will help avoid multiple formats and ensure consistency.
```
tweaks on wording

CIP-0105/test-vectors/test-vector-1.md
```diff
@@ -42,19 +42,19 @@ Hex: `f74d7ac30513ac1825715fd0196769761fca6e7f69de33d04ef09a0c417a752b1d84110299
 
 Bech32: `drep_xvk17axh4sc9zwkpsft3tlgpjemfwc0u5mnld80r85zw7zdqcst6w543mpq3q2vkjy3nw8x7n8asw4es78dyl4q7u7kwlwn7yy0sugxfrjs6z25qe`
 
-#### Verification key hash (DRep ID)
+#### Verification key hash (DRep VKH)
```
Ryun1, line 45:
```suggestion
#### [Deprecated] Verification key hash (DRep ID)
```

might be worth keeping the test vectors for the deprecated IDs as well as adding new ones

CIP-0129/README.md
gitmachtl: i agree, no need to change it in the ledger storage. this is a frontend "issue" not a ledger issue.

ashisherc: updated

CIP-0105/README.md
ashisherc: the deprecated tag links to the deprecation section

CIP-0105/README.md
ashisherc: updated

CIP-0105/README.md
ashisherc: updated

CIP-0105/README.md
ashisherc: done

CIP-0105/test-vectors/test-vector-1.md
ashisherc: done

ashisherc:
@rphair all conflicts resolved (Pun intended, perhaps?)

perturbing (approved):

Ryun1 (approved):
Great work!

gitmachtl:
yea, thanks everyone involved!

PR #858, opened 2024-07-17
CIP-???? | SpendMany script purpose (replace Spend)

matteocoppola:
This CIP proposes to replace the current Spend purpose with another one called SpendMany, removing the need of computational waste and higher fees when a transaction validates multiple inputs coming from the same script.

If adopted, Cardano smart contract development would be more efficient, less prone to security issues and easier from a Dev experience.

-------------------------

([rendered proposal in branch](https://github.com/matteocoppola/CIPs/blob/master/CIP-spendMany-script-purpose%20(replace%20Spend)/README.md))

### DIFF
--- a/CIP-spendMany-script-purpose (replace Spend)/README.md
+++ b/CIP-spendMany-script-purpose (replace Spend)/README.md
@@ -32,29 +32,83 @@ Even when a single utxo must be validated, all the recent smart contracts allow 
 
 Additionally, most of the times, the smart contract should also validate conditions of the transaction as a whole, for example the validity range or the sum of the Values of all the Inputs.
 Validating N inputs separately means wasting a lot of resources (computational and transaction fees) because these conditions must be checked N times.
-To circumvent this and get more performant smart contracts, the development community came up with the "Withdraw Zero" trick, which consists in using a separated WithdrawFrom script that will be executed only once for the full transaction. While this is a dirty hack used to circumvent the current status of the validator specification, it also still requires to check that the WithdrawFrom script is present for each script Input.
+To circumvent this and get more performant smart contracts, the development community came up with the "Withdraw Zero" trick, which consists in using a separated WithdrawFrom script that will be executed only once for the full transaction. While this is a dirty hack used to circumvent the current status of the validator specification, it also still requires to check that the WithdrawFrom script is present for each script Input, wasting resources.
 
 These problems have been briefly discussed in CPS-0004.
 
 It is therefore necessary to replace the existing Spend purpose with a new one that I called SpendMany.
-Please note that the name doesn't necessarily need to be changed, but we will use SpendMany here to refer to this CIP changes.
+The objective is to reduce the time complexity from O(n) to O(1) while keeping the development experience easy.
+Please note that the name doesn't necessarily need to be changed, but here we will use SpendMany here to refer to these CIP changes.
 
+The SpendMany purpose completely replaces the Spend purpose, because simply adding the new purpose would require extremely difficult changes in the ledger code and because, as discussed above, it covers 100% of the Spend validation use cases.
 
-TODO
+By using SpendMany we officially get rid of the need of using dirty hacks and unnecessary computation.
 
-TODO Alternative
+Even if it's not a dependency, this CIP works very well with the [CIP for Transaction Inputs as Unordered Set](https://github.com/cardano-foundation/CIPs/pull/758).
+
+Even if this CIP looks as an alternative to [CIP-112](https://github.com/cardano-foundation/CIPs/tree/master/CIP-0112), they can perfectly co-exist.
+
+If the changes formalized by this CIP can't be implemented within a reasonable timeframe, I at least suggest to implement the following change: adding in the script Context a new field called inputIndexesByScriptHashes of type ```Map<Hash, List<Int>>``` where the keys are the transaction's script hashes and the values are lists containing the indexes of the Inputs that come from these scripts.
+This change will allow to easily validate all the script inputs with either the CIP-112 or the already existing "Withdraw Zero" hack.
 
 ## Specification
 
-TODO
+The type signature of this script type will be the same of the usual Spend validator, except without any Datum, as it can be easily derived for each input, making the type signature the same of Minting and Staking validators:
+
+```Redeemer -> ScriptContext -> () ```
+
+The type signature of the newly introduced Purpose will be:
+
+```SpendMany ByteArray List<Int>```
+
+where:
+- ByteArray is the current script hash
+- List<Int> is the list of Input indexes that must be validated by this script. These are the usual inputs that would normally trigger the execution of Spend validation logic.
+
+### Alternative workaround:
+If SpendMany can't be implemented we should at least add the field inputIndexesByScriptHashes in the script Context :
+```
+-- | TxInfo for PlutusV3
+data TxInfo = TxInfo
+  { txInfoInputs                :: [V2.TxInInfo]
+  , txInfoReferenceInputs       :: [V2.TxInInfo]
+  , txInfoOutputs               :: [V2.TxOut]
+  , txInfoFee                   :: V2.Value
+  , txInfoMint                  :: V2.Value
+  , txInfoTxCerts               :: [TxCert]
+  , txInfoWdrl                  :: Map V2.Credential Haskell.Integer
+  , txInfoValidRange            :: V2.POSIXTimeRange
+  , txInfoSignatories           :: [V2.PubKeyHash]
+  , txInfoRedeemers             :: Map ScriptPurpose V2.Redeemer
+  , txInfoData                  :: Map V2.DatumHash V2.Datum
+  , txInfoId                    :: V2.TxId
+  , txInfoVotingProcedures      :: Map Voter (Map GovernanceActionId VotingProcedure)
+  , txInfoProposalProcedures    :: [ProposalProcedure]
+  , txInfoCurrentTreasuryAmount :: Haskell.Maybe V2.Value
+  , txInfoTreasuryDonation      :: Haskell.Maybe V2.Value
+  , inputIndexesByScriptHashes  :: Map ScriptHash [Haskell.Integer] -- ^ newly introduced list of inputs for each script 
+  }
+```
 
 ## Rationale: how does this CIP achieve its goals?
 
-TODO
+This CIP removes the need of unprofessional hacks to achieve whole-transaction validation while also achieving better computational efficiency.
+A simple chnage of the Spend purpose in SpendMany can simplify the developer experience, giving visibility of all the inputs to validate at once.
+
+### Alternatives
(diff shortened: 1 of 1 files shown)


CIP-spendMany-script-purpose (replace Spend)/README.md
```diff
@@ -0,0 +1,116 @@
...
+License: Apache-2.0
+---
+
+## Abstract
+
+This CIP proposes to replace the current Spend purpose with another one called SpendMany, removing the need of computational waste and higher fees when a transaction validates multiple inputs coming from the same script.
+
+SpendMany is executed only once for the whole transaction and it comes with 2 parameters:
+- The current script hash
+- The list of the indexes of the inputs that are being spent from this script
```
fallen-icarus, line 22: Some scripts allow composing with itself where inputs use *different* redeemers. It sounds like this CIP will require all inputs from the script in this transaction to use the same redeemer.

fallen-icarus (commented):
Would it be possible for users to specify multiple `spendMany` groups for the same script in a single transaction? For example, perhaps they want to sub-divide a group of inputs being spent by a single script (possibly by redeemer used).

CIP-spendMany-script-purpose (replace Spend)/README.md
matteocoppola: Yes, the current CIP version requires 1 redeemer for all the inputs from the smart contract.
Can't we just use a single redeemer that contains lists for each field, indexed by the order of the inputs?
We already do like that and it works smoothly: as we know the inputs and their order in advance, it's easy to setup such a redeemer.

CIP-spendMany-script-purpose (replace Spend)/README.md
```diff
@@ -0,0 +1,116 @@
+---
+CIP: ????
+Title: SpendMany script purpose 
+Category: Plutus
+Status: Proposed
+Authors:
+    - Matteo Coppola <m.coppola.mazzetti@gmail.com>
+Implementors: 
+    -
```
rphair, line 9:
```suggestion
Implementors: []
```
... nobody for now (i.e. the empty set)

CIP-spendMany-script-purpose (replace Spend)/README.md
```diff
@@ -0,0 +1,116 @@
...
+CIP: ????
+Title: SpendMany script purpose 
+Category: Plutus
+Status: Proposed
+Authors:
+    - Matteo Coppola <m.coppola.mazzetti@gmail.com>
+Implementors: 
+    - 
+Discussions:
+    -
```
rphair, line 11:
```suggestion
    - https://github.com/cardano-foundation/CIPs/pull/858
```

rphair (commented):
thanks @matteocoppola - putting on agenda to introduce at next CIP meeting, as titled in the document without the "replace" clause (which might be debatable): https://hackmd.io/@cip-editors/93 - you would be very welcome at the meeting.

We should eventually make the PR & document titles consistent, but in the meantime I think it helps to have `(replace Spend)` in the PR title to call attention to this fact.  I don't have the skill set to suggest whether `Spend` and `SpendMany` could or should coexist but I think it would be a good discussion point for the meeting.

CIP-spendMany-script-purpose (replace Spend)/README.md
fallen-icarus: >  Can't we just use a single redeemer that contains lists for each field, indexed by the order of the inputs?

I'd rather not use the redeemer like this since it must always be validated at run-time to verify all of the inputs are properly accounted for. If an input is accidentally/maliciously omitted from the redeemer, the smart contract can return true without actually validating that input. In other words, the security of this approach depends on the DApp developer. It is possible for one DApp to be secure while another DApp is missing checks.

Instead, the ledger can do the checks *before* running the scripts if separate spending groups are used. One of the plutus/ledger devs can correct me if I am wrong, but if the rule is that all inputs in a spending group must use the same redeemer, wouldn't this be a very cheap check that can be part of phase 1 validation? You can just hash the redeemer used for each input in the group and then compare the hashes. If an input is not tagged by a redeemer, the ledger will immediately catch it and fail the transaction.

I have not used the redeemer index pattern very much, so if I am misunderstanding something please correct me.

CIP-spendMany-script-purpose (replace Spend)/README.md
```diff
@@ -0,0 +1,115 @@
...
+The objective is to reduce the time complexity from O(n) to O(1) while keeping the development experience easy.
+Please note that the name doesn't necessarily need to be changed, but here we will use SpendMany here to refer to these CIP changes.
+
+The SpendMany purpose completely replaces the Spend purpose, because simply adding the new purpose would require extremely difficult changes in the ledger code and because, as discussed above, it covers 100% of the Spend validation use cases.
+
+By using SpendMany we officially get rid of the need of using dirty hacks and unnecessary computation.
+
+Even if it's not a dependency, this CIP works very well with the [CIP for Transaction Inputs as Unordered Set](https://github.com/cardano-foundation/CIPs/pull/758).
+
+Even if this CIP looks as an alternative to [CIP-112](https://github.com/cardano-foundation/CIPs/tree/master/CIP-0112), they can perfectly co-exist. Additionally this CIP drastically simplifies the development experience.
```
Ryun1, line 49:
```suggestion
Even if this CIP looks as an alternative to [CIP-112 | Observe Script Type](https://github.com/cardano-foundation/CIPs/tree/master/CIP-0112), they can perfectly co-exist. Additionally this CIP drastically simplifies the development experience.
```
just helps for those who struggle to remember every number 😆

CIP-spendMany-script-purpose (replace Spend)/README.md
```diff
@@ -0,0 +1,115 @@
...
+  , txInfoCurrentTreasuryAmount :: Haskell.Maybe V2.Value
+  , txInfoTreasuryDonation      :: Haskell.Maybe V2.Value
+  , inputIndexesByScriptHashes  :: Map ScriptHash [Haskell.Integer] -- ^ newly introduced list of inputs for each script 
+  }
+```
+
+## Rationale: how does this CIP achieve its goals?
+
+This CIP removes the need of unprofessional hacks to achieve whole-transaction validation while also achieving better computational efficiency.
+A simple chnage of the Spend purpose in SpendMany can simplify the developer experience, giving visibility of all the inputs to validate at once.
```
Ryun1, line 96:
```suggestion
A simple change of the Spend purpose in SpendMany can simplify the developer experience, giving visibility of all the inputs to validate at once.
```
typo!

rphair:
@matteocoppola following up from my https://github.com/cardano-foundation/CIPs/pull/858#pullrequestreview-2183768383 from CIP meeting today (re: does `SpendMany` "replace" `Spend`, add to it, modify it, supersede it and/or deprecate it...):

Consensus is that Ledger input (cc @lehins @WhatisRT) is needed to decide how to go ahead with this.  Everyone agrees the newly defined purpose is an improvement on the old one, with the old use cases being definable with the new script purpose, but can't be sure if there aren't "edge cases" that would break backwards compatibility: and therefore a requirement for versioning and/or supporting both the old & new script purpose.

Because of this question of scope, as well as feasibility of doing this in the Ledger, we've decided to wait for Ledger's initial feedback with some kind of endorsement before assigning this a candidate CIP number.  Once we get to this stage I'll be sure it's on the next CIP meeting agenda.

WhatisRT:
I'll need to think a bit more about this, but my first thoughts are that this is a great idea. The issue with the redeemer is a bit bigger though, since Ledger also needs to match redeemers to inputs. One idea would be to allow an input to be a list of inputs locked by the same script. Then one can still run the same script with different redeemers while the logic that matches redeemers could be unchanged.

lehins:
This is definitely a sensible proposal. That being said, we can't do this change for PlutusV1..PlutusV3, since we can't change the context for those versions and addition of a new script purpose is a change to the plutus context. So, we can't "replace Spend", as the title suggests. But, that doesn't mean we can't implement it for PlutusV4. In the matter of fact, I just made a [comment](https://github.com/cardano-foundation/CIPs/pull/758#pullrequestreview-2197339141) on another CIP that we could solve those two proposals with the same solution.

New script purpose could look something like this for plutus:
```haskell
PV4.SpendingMany [(Integer, TxIn)]
```
where `Integer` would refer to the index in the inputs

While for Ledger's redeemer pointer/plutus purpose (i.e. Ledger's script purpose) would look like this:
```haskell
SpendingMany (Set Word32)
```

There are a few issues that come with this approach off top my head, neither of which might be an actual problem, but something developers and users need to be aware of. In all of the examples below I assume all inputs are locked by the same script, which would be enforced by ledger:

* Same redeemer is expected for multiple inputs. Naturally, datum supplied and execution units must be the same for an identical execution of a script, so this is something to be expected: 
```haskell
Map.fromList [(SpendingMany [1,2,3], (_ :: Data, ExUnits 123 456))]
```
* When different behavior is desired it would still be possible to split it up into multiple purposes:
```haskell
Map.fromList
  [ (SpendingMany [1,2], (a :: Data, ExUnits 123 456)
  , (SpendingMany [3], (b :: Data, ExUnits 1230 4560)
  ]
```
* Incorrect usage could lead to execution of the same script for the same input multiple times:
```haskell
Map.fromList
  [ (SpendingMany [1,2], (a :: Data, ExUnits 123 456)
  , (SpendingMany [1,3], (b :: Data, ExUnits 1230 4560)
```
In this example the script that locks input `1` would be executed twice: once with redeemer `(a :: Data, ExUnits 123 456)` and the other time with redeemer `(b :: Data, ExUnits 1230 4560)`. That being said, there could be some use cases for this and people actually want this behavior or we could add some ledger rules that prevent duplicate execution. Something that would have to be decided.

rphair:
@matteocoppola it's noted by @fallen-icarus (https://github.com/cardano-foundation/CIPs/pull/418#issuecomment-2299286201) that this should be considered a solution for candiate CPS-0004.  By convention that CIP & CPS would be linked in the header so please keep that in mind for future revisions as these documents evolve.

matteocoppola:
Hello everyone, I hope to find some more time to re-activate this CIP, as the current dev experience on Cardano is still incredibly complex because of all the different validator types.
As said above, double satisfaction attacks are also very easy to create when devs don't have the necessary experience.

A single validator would simplify greatly the dev flow, helping to reason easily on possible security issues while reducing the computation units normally needed for redundant checks to avoid double satisfaction attacks or to link a minting validator action with a spend action, etc.

Ryun1:
> I hope to find some more time to re-activate this CIP

Hey @matteocoppola 
Since your last comment we will tag this proposal as `Waiting for Author` for the time being

rphair:
@matteocoppola @Ryun1 with no response to https://github.com/cardano-foundation/CIPs/pull/858#issuecomment-2734019023, progressing this to `Likely Abandoned`.  If this is indeed achievable for Plutus V4 then the author needs to make it ready for Ledger & community review soon.

rphair:
@matteocoppola with no updates in 5 months we have to assume this CIP is in fact `Abandoned` even if the concept may remain viable & still be considered useful.

PR #864, opened 2024-07-26
CIP-0122 | Typographical fixes & corrections

kozross:
This fixes some errors (mostly formatting, but some copy-paste issues), as well as some naming errors, in CIP-122.

In the process of implementation of CIP-122 (and CIP-123), the names of the functions to be implemented were changed, and `writeBits` ended up having a change in its signature, which the current versions of both documents don't reflect. I'm guessing that there is more of a process for non-typographical changes: if so, I'm happy to follow it, but decided to send this PR with typographical fixes only for now.


rphair (approved):
thanks @kozross ... have read through all changes & all seem straightforward except maybe some conceptual items that I am trusting your expertise on.

Ryun1 (approved):

PR #866, opened 2024-07-28
CIP-1694 | French - Updates and semantics

Hornan7:
- Correction to the French version related to PR #847


rphair (approved):
nice work @Hornan7 on both the original & the translation. 🙏  I compared the two diffs side-by-side so was also able to compare (by a good enough random sample) the machine readable (neither English nor French) terms that were updated in this revision. 🎉

Ryun1 (approved):

PR #870, opened 2024-08-01
CIP-???? | Votes & Certificates on Cardano

ghost:
This proposal specifies a cryptographic protocol for distributed and decentralised stake-based votes and compact certificates, based on the [Approximate Lower Bounds Arguments](https://iohk.io/en/research/library/papers/approximate-lower-bound-arguments/) research paper.

---

([rendered proposal](https://github.com/cardano-scaling/CIPs/blob/alba/CIP-0XXX/README.md))

PR #848, opened 2024-06-27
CIP-0102 | Reference Implementation

SamDelaney:
A reference implementation for CIP-0102 royalties, written in Lucid (offchain) and Aiken (onchain).

In this reference implementation you'll find examples of:
- Minting a CIP 102 compliant NFT with royalties
- Validators for storing royalty metadata
  - Always Fails (royalty cannot be modified)
  - Reducible (royalty cannot be increased, but can be decreased)
- Reading a CIP 102 NFT’s royalties off chain
- Reading and validating against CIP 102 NFT royalties on chain

This code has not been audited and is **NOT INTENDED FOR PRODUCTION**, but to provide code examples of expected behavior.

PR #872, opened 2024-08-02
CIP for Peras

bwbush:


CIP-0PRS/CIP-0PRS.lagda.md
```diff
@@ -0,0 +1,1355 @@
+---
+Title: Peras - Faster Settlement on Cardano
```
rphair, line 2:
```suggestion
Title: Ouroboros Peras - Faster Settlement
```
PR title (this) is better of the two especially because the word `Cardano` is always redundant in a CIP.

CIP-0PRS/CIP-0PRS.lagda.md
```diff
@@ -0,0 +1,1355 @@
...
+Category: Consensus
+Status: Proposed
+Authors:
+  - Arnaud Bailly <arnaud.bailly@iohk.io>
+  - Brian W. Bush <brian.bush@iohk.io>
+  - Yves Hauser <yves.hauser@iohk.io>
+  - Hans Lahe <hans.lahe@iohk.io>
+Implementors: []
+Discussions:
+    - https://github.com/cardano-foundation/CIPs/pull/?
```
rphair, line 13:
```suggestion
    - https://github.com/cardano-foundation/CIPs/pull/872
```

CIP-0PRS/CIP-0PRS.lagda.md
```diff
@@ -0,0 +1,1355 @@
...
+> - Tools    | A broad category for ecosystem tools not falling into any other category.
+> - Plutus   | Changes or additions to Plutus
+> - Ledger   | For proposals regarding the Cardano ledger (including Reward Sharing Schemes)
+> - Catalyst | For proposals affecting Project Catalyst / the Jörmungandr project
+
+* Look at [Leios spec](https://github.com/cardano-foundation/CIPs/blob/ouroboros-leios/CIP-0079/README.md) for inspiration
+* [CIP-001](https://github.com/cardano-foundation/CIPs/tree/master/CIP-0001) guidelines
+
+## Abstract
+<!-- A short (\~200 word) description of the proposed solution and the technical issue being addressed. -->
```
rphair, line 58:
```suggestion
<!-- A short (\~200 word) description of the proposed solution and the technical issue being addressed. -->
```
Now that the content is in place here you can (and should, before editing really begins) remove the comment scaffolding from the template: both here & in other places in the document.

CIP-0PRS/CIP-0PRS.lagda.md
```diff
@@ -0,0 +1,1355 @@
+---
+Title: Peras - Faster Settlement on Cardano
+CIP: ?
+Category: Consensus
```
rphair, line 4: We don't have a `Consensus` category, but are there stakeholders & representatives that could _give_ us one?  It would be forward-looking to add it, and especially useful if there are more CIPs coming from you guys 🤓 but first this category would need to be "enlisted" through dedicated industry contacts as described for the currently enlisted categories here (`Ledger`, `Plutus`. `Catalyst`):
- https://github.com/cardano-foundation/CIPs/blob/master/CIP-0001/README.md#categories
- https://github.com/cardano-foundation/CIPs/blob/master/CIP-0001/README.md#project-enlisting

If you think `Consensus` meets the criteria for enlistment (based on commitment to reviews & projects following the CIP process) then I would be happy to have it.  We haven't enlisted any new categories since @KtorZ was an active editor but if you're interested we'll reciprocate as best we can (cc @Ryun1 @Crypto2099).

rphair (commented):
apologies @bwbush if this is premature, but given the title I couldn't resist taking a look and noticed a few things which might be of help to consider early (among some standard editing chores).

CIP-0PRS/CIP-0PRS.lagda.md
```diff
@@ -0,0 +1,1355 @@
...
+This yields a vote with some _weight_ attached to it "randomly" computed so that the overall sum of weights should be around expected committee size.
+
+This method has also been [prototyped in Haskell](https://github.com/input-output-hk/peras-design/blob/73eabecd272c703f1e1ed0be7eeb437d937e1179/peras-vote/src/Peras/Voting/Vote.hs#L174).
+
+### Specification of certificates
+
+> [!CAUTION]
+> We need to commit solely to ALBA or to Mirthril, and edit this section accordingly.
+
+#### Mithril certificates
```
rphair, line 932: I noticed text here is currently duplicated with #870 (cc @abailly-iohk - we understand they're both only Draft & this is only to assure we have properly distinct documents when ready for review): https://github.com/cardano-scaling/CIPs/blob/alba/CIP-0XXX/README.md#certificates

rphair (commented):
(intended to include in last review)

CIP-0PRS/CIP-0PRS.lagda.md
bwbush: @rphair, My understanding is that a separate CIP will create a "Consensus" category. We discussed this at a recent Consensus Working Group meeting.

rphair: @bwbush it's good to hear that as a recommended strategy, since then likewise the first authors of `Consensus` CIPs would be the "enlisters" and effectively volunteering for review and advice about implementation.  cc @abailly-iohk

This is the best strategy I can think of to keep the CIP process decentralised with respect to supporting companies.  As this CIP moves forward through initial review I'll add `Consensus` to the official categories in CIP-0001.  cc @Ryun1 @Crypto2099

CIP-0PRS/CIP-0PRS.lagda.md
```diff
@@ -0,0 +1,1355 @@
...
+
+| Time offline | Blocks (GB) | Certificates (GB) |
+|--------------|-------------|-------------------|
+| 1 month      | 5.56        | 1.23              |
+| 3 months     | 16.68       | 3.69              |
+| 6 months     | 33.36       | 7.38              |
+
+#### Network costs
+
+![Typical node inbound & outbound traffic](https://peras-staging.cardano-scaling.org/assets/images/node-average-traffic-49a7cbcee8e4f05a7a42b61e445e5db9.jpg#scale50)
```
SmaugPool, line 1087: The link returns 401 Authorization Required

bwbush: Thanks. We're in the process of migrating the diagrams directly into the folder for this CIP.

rphair (commented):
@bwbush we can't keep a code repository within a CIP or use it as a build source, according to precedent and pending an explicit statement that will be added to CIP-0001 soon (last time it was confirmed: recently with https://github.com/cardano-foundation/CIPs/pull/848#pullrequestreview-2145300838).  The convention would be for the `.agda` files & folder to be kept in an external repository which is linked from this CIP. 🙏

bwbush:
> @bwbush we can't keep a code repository within a CIP or use it as a build source, according to precedent and pending an explicit statement that will be added to CIP-0001 soon (last time it was confirmed: recently with [#848 (review)](https://github.com/cardano-foundation/CIPs/pull/848#pullrequestreview-2145300838)). The convention would be for the `.agda` files & folder to be kept in an external repository which is linked from this CIP. 🙏

Thanks for the information. I'll consult with my colleagues and get back to you regarding this.

bwbush:
@rphair, I've removed the source files and added a section in the Appendix that explains how to typecheck the Peras specification.

This CIP has the unique quality that the specification is written formally in Agda so that it can be (1) machine checked and (2) verified that it avoids inconsistency and ambiguity. If at all possible, we'd like to keep three slightly non-standard aspects here:

- The specification's file is called `README.lagda.md`, which clarifies that it is an Agda-compatible form of markdown and can be type checked by the Agda theorem-prover.
- The file `peras.agda-lib` is required by Agda.
- The file `flake.nix` greatly simplifies the workflow for reviewers and implementers who'd like to verify the specification for themselves or proposed modifications.

Also note that there like will be additional machine-checkable Agda-based CIPs proposed in the future.

rphair:
@bwbush - to me that sounds like it justifies keeping the currently posted material mentioned above with the CIP... I'll invite @Ryun1 @Crypto2099 to confirm or say anything further about the issue, proceeding at your own discretion, given:

The prohibition of keeping implementation code in the CIPs — especially against using them as a compilation tree — was to keep the usual means of code development (issues, pull requests) outside of the CIP process... and of course to avoid the necessity of CIP editors reviewing changes to that code, which generally we will not be qualified to do.

If & when merged with your 3 keystone files, we would afterward need a quorum of CIP editors to make any updates to those files.  If you ever think the frequency or latency of these changes could be a problem for your project, that would indicate that the canonical versions of these files would be better kept in a separate repository.

CIP-0PRS/README.lagda.md
```diff
@@ -0,0 +1,1604 @@
...
+
+Under similar assumptions, we can estimate the storage requirements entailed by Peras: ignoring the impact of cool-down periods, which last for a period at least as long as $k$ blocks, the requirement to store certificates for every round increases node's storage by about **20%**. Votes are expected to be kept in memory so their impact on storage will be null.
+
+#### CPU
+
+The [Peras Technical Report #2](https://peras.cardano-scaling.org/docs/reports/tech-report-2#votes--certificates) provides some models and benchmarks for votes generation, votes verification, certificates proving and certificates verification, and votes diffusion. Those benchmarks are based on efficient sortition-based voting and ALBAs certificate, and demonstrate the impact of Peras on computational resources for a node will be minimal. Moreover, the most recent version of the algorithm detailed in this report is designed in such a way the voting process runs in parallel with block production and diffusion and therefore is not on this critical path. For ALBA certificates, assuming 1000 votes, a honest to faulty ratio of 80/20, and security parameter $λ=128$, one has the following typical measurements.
```
djetchev, line 1486: Could we clarify a bit more the details of the parameters of Alba (as described in the [original paper](https://eprint.iacr.org/2023/1655.pdf)) in order to obtain the certificate size of 47Kb. After reading the report above, I am having some difficulty mapping the parameters from the report to the parameters of the paper? 

Additionally, the estimated size of 47Kb is a bit unclear to me - we talk about honest / faulty ratio of 80/20 and (what seems to be assumed) is that $n_f$ is set to 20% and not the Peras quorum of 75% to obtain this value of 47Kb (I might be wrong here, so correct me if this is the case). This 20% coincides with the adversarial votes and hence, I am asking, what statement exactly is the certificate endorsing? Relating to the Alba abstract problem in the paper, Is it simply saying that "I know a set of voters and at least one of them is honest" ? If so, why is this sufficient for the consensus / boosting the block? I’d appreciate some clarifications on that.

abailly-iohk: The 47kB value might have come from an earlier version of ALBA prototype in Haskell, the latest value I computed (git hash 08cb09accf47093f53cf55e9b344f15cc9a3a408) gives me

```
 % cabal run alba -- prove --len 710
Generating random items
Generating proof Options {size = 100, len = 710, params = Params {λ_sec = 128, λ_rel = 128, n_p = 80, n_f = 20}, input = Nothing, output = "proof.alba"}
Written proof to 'proof.alba' (50278 bytes, 0 retries)
```

The parameters 80/20 used the computation came from my misunderstanding of what's actually guaranteed by ALBA certificates as outlined in [this discussion](https://github.com/cardano-scaling/alba/discussions/17) and as noted in one of the comments, this needs to be fixed to account for more serious assessment of security implication of ALBA parameters on Peras.
This is the purpose of https://github.com/cardano-scaling/alba/pull/22 and the discussion we've had on Friday. This CIP and the report will be corrected to tone down the enthusiasm and expectations around ALBA as it's perfectly possible we realise these are not suitable for Peras after all.

rphair:
@bwbush @abailly-iohk - applying freshly created label https://github.com/cardano-foundation/CIPs/labels/Category%3A%20Consensus - please chime in with this issue with any feelings about how you see the CIP process participation for this category developing:
- https://github.com/cardano-foundation/CIPs/issues/898

I've marked this `Unconfirmed` only because it hasn't been assigned a candidate CIP number yet.  In my opinion this could happen at any time (@Ryun1 @perturbing @Crypto2099 - shall we put on Review for next meeting to confirm this?).

CIP-0PRS/README.lagda.md
```diff
@@ -0,0 +1,1503 @@
...
+Peras requires a single addition, `peras_cert`, the [block](#blocks) data on the ledger.
+
+```diff
+ block =
+   [ header
+   , transaction_bodies         : [* transaction_body]
+   , transaction_witness_sets   : [* transaction_witness_set]
+   , auxiliary_data_set         : {* transaction_index => auxiliary_data }
+   , invalid_transactions       : [* transaction_index ]
++  , ? peras_cert               : alba_certificate
```
perturbing, line 1152: The occasional addition of a Peras certificate with size of the order of the max block size will impact the diffusion of these blocks. The [section](https://github.com/cardano-scaling/CIPs/blob/peras/CIP-0PRS/README.lagda.md#resource-requirements) on resource requirements focuses on localized factors, what about the emergent global impact?

I can imagine that slot leaders that produce these blocks, might diffuse them worse, will this create slot battles?

Also, some Dapps would like to see the max block size increase, the addition of Peras might limit this. Not sure what statement _can_ be made about this fact, but it might be worth adding a section on this balance for completeness?

CIP-0PRS/README.lagda.md
```diff
@@ -0,0 +1,1503 @@
...
+
+In no way does Peras weaken any of the security guarantees provided by Praos or Genesis. Under strongly adversarial conditions, where an adversary can trigger a Peras voting cool-down period, the protocol in essence reverts to the Praos (or Genesis) protocol, but for a duration somewhat longer than the Praos security parameter. Otherwise, settlement occurs at the blocks that Peras voting has boosted. The Peras protocol parameters can be tuned to adjust the settlement time or the non-settlement probabilities. Some stakeholder use cases might prefer shorter settlement times but with a higher probability of retries, or vice versa.
+
+
+### Resource Requirements
```
perturbing, line 1337: What is the risk of stake not participating in this protocol? I can imagine that the below additional cost might not be welcome to all SPO's, especially as there is no incentive to run this. 

So, if it is the case that SPO's can opt in for running this protocol (not sure how this will work), how will the security analysis move? Is this to be considered non-voting and thus adversarial stake? If so, how do we prevent this?

perturbing (commented):
Compliments on how well this CIP is written, it dissects a complex design into readable sections and the motivation of the CIP sets the perfect stage to understand the context of it all. It was a long read, but I learned a ton about Peras, thank you :+1: 

In general, I think this CIP would benefit from more structure due to its size, as is done in [CIP-1694](https://github.com/cardano-foundation/CIPs/tree/master/CIP-1694). Perhaps the addition of a table of contents for faster navigation and an occasional collapsed section (for example the Agda spec section), would be good.

CIP-0PRS/README.lagda.md
perturbing: Regarding incentives, can block producers use Alba to get a probability on whether a certain KES key (and thus SPO) contributed to the Peras protocol in past rounds by viewing their local votes? If so, can we use this measure in the reward calculation at the end of each epoch?

CIP-0PRS/README.lagda.md
abailly-iohk: The idea was that the peras certificate would be significantly smaller than the block size so that block producers would fill in the remaining bytes and just propagate the block normally, so there should not be an impact on block diffusion, only on tx inclusion in case this happens while the mempools are congested. 

Note that inclusion of certificates in blocks will be an extremely rare event. In the worst case, assuming nodes immediately enter cooldown when they are back from it, this means 1 Peras certificate roughly every 3000 blocks. In normal operations, this could be much rarer as we expect quorum will be reached on every round.

CIP-0PRS/README.lagda.md
abailly-iohk: It does not seem to us the resources requirements for Peras would incur significant additional costs to SPOs, but of course this should be assessed more carefully. We could imagine some specific incentive scheme to reward participation to Peras but the situation here is different than with Mithril: Peras changes the consensus and therefore has to be part of the core node operations.

That said, it's possible to start rolling out Peras as a sidecar and make blocks certificates informational, e.g something anyone connected to the network would observe and decide whether or not it's relevant. If you observe that 75% of nodes running Peras endorsed a specific block, even though this is not baked in the consensus you can already assume this block has a much stronger  likelihood of not being rolledback.

ALBA certificates sample the votes' set so we could not use that to assess whether or not a particular SPO participated in Peras.

perturbing: > To reply specifically to one of your questions: if consensus switches to weight-based chain selection, there's not opt-out from Peras and it must be rolled out to every node.

Hmmm, but the pessimistic case of Peras is Praos, so to opt out one “simply” forces the node in the pessimistic case? For example, by filtering messages the node receives?

abailly-iohk: I don't understand what you mean, but if you're saying a node can refuse to "play the game" honestly then sure, that's a possibility and the reason why there's a quorum and we need security proofs to understand the impact of such behaviour on the system. Or are you saying something else?

rphair:
A high level description published recently that may be of help in review: https://iohk.io/en/blog/posts/2024/10/14/ouroboros-peras-the-next-step-in-the-journey-of-cardano-s-protocol-1/

CIP-0PRS/CIP-0PRS.lagda.md
rphair: Resolved with adding `Consensus` category in https://github.com/cardano-foundation/CIPs/pull/924.

@dnadales it still would be nice to get a CIP with some guidelines for `Consensus` CIPs as per https://github.com/cardano-foundation/CIPs/pull/872#discussion_r1706965548 & to promote it somewhat from where it currently exists here ("tentatively enlisted"): https://github.com/cardano-foundation/CIPs/blob/master/CIP-0001/README.md#categories (cc @bwbush & @Ryun1 @Crypto2099 @perturbing)

CIP-0PRS/CIP-0PRS.lagda.md
bwbush: This duplication is removed in the latest version.

CIP-0PRS/CIP-0PRS.lagda.md
bwbush: URL fixed.

CIP-0PRS/README.lagda.md
bwbush: I'm resolving this because the ALBA-specific discussion was moved to the Voting CIP.

bwbush:
@rphair, I believe that b59c80e48373339d36f3e3b996473c82798ddcaf responds to all outstanding comments and suggestions.

In particular, I added subsections where there were long blocks of text and I included a collapsable table of contents. (The TOC might be redundant because github will not dislay a TOC for markdown documents if one pressed an icon on the upper left of the document. Feel free to remove the TOC if you see fit.)

rphair:
thanks @bwbush - according to our usual process that means this will be up for Review at the next meeting (https://hackmd.io/@cip-editors/101).  I believe all will be in favour of giving this a CIP number there, even if some remaining issues are found in the meantime.

Also, if you would please tag (for editors' reference) whom you believe has the most interest in reviewing and/or contributing to `Consensus` proposals, we'll keep track of these for future use.

CIP-0PRS/README.lagda.md
```diff
@@ -0,0 +1,1568 @@
+---
+Title: Ouroboros Peras - Faster Settlement
```
Ryun1, line 2:
```suggestion
CIP: 0PRS
Title: Ouroboros Peras - Faster Settlement
```
CIP number is above title via template

CIP-0PRS/README.lagda.md
```diff
@@ -0,0 +1,1568 @@
...
+CIP: 0PRS
```
Ryun1, line 3:
```suggestion
```

CIP-0PRS/README.lagda.md
```diff
@@ -0,0 +1,1568 @@
...
+   , auxiliary_data_set         : {* transaction_index => auxiliary_data }
+   , invalid_transactions       : [* transaction_index ]
++  , ? peras_cert               : votes_certificate
+   ]
+```
+
+[Votes](#votes) are serialized in the following CDDL.
+
+```cddl
+vote =
```
Ryun1, line 1218: tiny nit pick, but vote is already defined via [Conway CDDL](https://github.com/IntersectMBO/cardano-ledger/blob/master/eras/conway/impl/cddl-files/conway.cddl#L506C1-L506C14) :)

CIP-0PRS/README.lagda.md
rphair, line 3:
```suggestion
CIP: 140
```

rphair (commented):
Please rename containing folder to `CIP-0140` and update the leading pathname of the `README` file accordingly.  Also, eventually the trailing pathname will have to be renamed `README.md` (from `README.lagda.md`)as soon as convenient for you (given your own workflow & any upstream development).

bwbush:
@Ryun1 or @rphair, just checking on the status and verifying that you're not waiting on me for anything. Thanks.

CIP-0140/README.lagda.md
```diff
@@ -0,0 +1,1568 @@
...
+Title: Ouroboros Peras - Faster Settlement
+Category: Consensus
+Status: Proposed
+Authors:
+  - Arnaud Bailly <arnaud.bailly@iohk.io>
+  - Brian W. Bush <brian.bush@iohk.io>
+  - Sandro Coretti-Drayton <sandro.coretti@iohk.io>
+  - Yves Hauser <yves.hauser@iohk.io>
+  - Hans Lahe <hans.lahe@iohk.io>
+Implementors: Intersect
```
rphair, line 12:
```suggestion
Implementors:
  - Intersect
Proposed Solutions:
  - CPS-0017
```
In both cases entries are bulleted lists.  (I think it's possible that someday another solution might come along for this CPS.)

CIP-0140/README.lagda.md
```diff
@@ -0,0 +1,1568 @@
...
+### Implementation plan
+
+- [ ] Detailed node-level (as opposed to this protocol-level) specification.
+- [ ] Develop node-level conformance test suite.
+- [ ] Consider developing a "quick and dirty" implementation for large scale experiments.
+- [ ] Coordinate with related activities on other protocol enhancements (Genesis, Mithril, Leios, voting, etc.).
+- [ ] Triage by intersect Core Infrastructure and Consensus functions.
+- [ ] The upcoming CIP that establishes a *Consensus* category for CIPs may define additional requirements for the implementation plan.
+
+The following diagram summarizes a possible architecture for Peras highlighting its interactions with other components of the node.
```
rphair, line 1493: @bwbush this goes beyond the "checklist" format we prefer for the two parts of _Path to Active_ but some of this seems necessary due to dependencies in the implementation plan.  Editors try to maintain a "checklist" format to simplify the many assessments made over time by dozens of people on a CIP's way from `Proposed` to `Active`.

With that in mind, can you do this for each of the following paragraphs:
- if it's a "must do" item, put a tickbox in front of it (with nested tick-boxes under each of the subtasks in that paragraph);
- if it's a task based on a choice that will be made in the future (e.g. another implementation that hasn't evolved yet), leave it in prose form but please you & your team should keep updating the _Implementation Plan_ into a linear process as these possibilities are narrowed down.

rphair (commented):
thanks @bwbush - we had only been waiting for:

1 - confirmation from you that the commits to the document reflect your latest evolution of the CIP... unless you say otherwise, I'll assume from your last comment that evolution is complete.

2 - renaming the CIP README file to the canonical `README.md` (without the `.lagda`), currently necessary for a couple of reasons:
- derived web site parsers like the one on `cips.cardano.org` won't recognise the unconventional filename and so either won't show your CIP or will show blank text there
- the GitHub CIP directory page — where many users, visitors from search engines, and those clicking on the repository `README` will arrive — will only show a file listing & no CIP content... as we currently see here: https://github.com/cardano-scaling/CIPs/tree/peras/CIP-0140

Could point # 2 be resolved by renaming the file to `README.md` and including, in the Abstract itself where it will be seen immediately (from your https://github.com/cardano-foundation/CIPs/pull/872#issuecomment-2276976260)?
> This file `README.md` is in an Agda-compatible form of markdown and can be type checked by the Agda theorem-prover.

If point # 1 is true then I think further review will go quickly.  We have nobody on the CIP editors' team nor any community reviewer qualified to find fault in the specification, and so the only further changes would be to make formatting consistent with other CIPs.  I'll check through this below & then therefore we should be able to mark this `Last Check` for our next meeting (10 days' time) assuming everyone agrees the proposal is clearly explained (I do).

CIP-0140/README.lagda.md
rphair: p.s. I'm also now wondering if this proposal would also solve the candidate problem statement
- https://github.com/cardano-foundation/CIPs/pull/925

Please post if you think so and I'll suggest appropriate markup; if CIPs or CIPs aren't merged yet, the convention is to add a link to the PR like this (https://github.com/cardano-foundation/CIPs/pull/947/files) and then update after merge.

CIP-0140/README.lagda.md
```diff
@@ -0,0 +1,1568 @@
...
+
+| Parameter              | Symbol           | Units   | Value | Rationale                                                            |
+| ---------------------- | ---------------- | ------- | ----: | -------------------------------------------------------------------- |
+| Round length           | $U$              | slots   |    90 | Settlement/non-settlement in under two minutes.                      |
+| Block-selection offset | $L$              | slots   |    30 | Several multiples of $\Delta$ to ensure block diffusion.             |
```
amesgen, line 1359: This low value for $L$ (due to the lack of pre-agreement) allows relatively weak adversaries (e.g. weaker than 25%) to induce frequent cooldowns due to "vote splitting attacks", ie the adversary has a short fork with a different block $L$ slots before the start of a round, and then diffuses just barely before the round starts to (see "Honest vote splitting", in particular Figure A.1 in https://tweag.github.io/cardano-peras/peras-design.pdf for more details and concrete numbers), such that the honest vote is split and no quorum is reached.

One can make things a bit better by choosing a larger $L$, but it doesn't actually help much (or rather, one would have to make $L$ so large as for Peras to become much less useful).

I think it is enough to just point out somewhere in the CIP that one has to accept this downside for pre-alpha Peras.

CIP-0140/README.lagda.md
```diff
@@ -0,0 +1,1568 @@
...
+| Certification boost    | $B$              | blocks  |    15 | Negligible probability to roll back boosted block.                   |
+| Security parameter     | $k_\text{peras}$ | blocks  |  3150 | Determined by the Praos security parameter and the boost.            |
+| Certificate expiration | $A$              | slots   | 27000 | Determined by the Praos security parameter and boost.                |
+| Chain-ignorance period | $R$              | rounds  |   300 | Determined by the Praos security parameter, round length, and boost. |
+| Cool-down period       | $K$              | rounds  |   780 | Determined by the Praos security parameter, round length and boost.  |
```
amesgen, line 1364: How were these chosen? We have

 - $A = T_{\text{heal}} + T_{\text{CQ}}$, 
 - $R = \lceil A/U \rceil$ and 
 - $K = \lceil (A + T_{\text{CP}}) / U \rceil$. 

From the Praos parameterization (see e.g. [here](https://docs.google.com/document/d/1B8BNMx8jVWRjYiUBOaI3jfZ7dQNvNTSDODvT5iOuYCU/edit?tab=t.0#heading=h.qh2zcajmu6hm)), we have $T_{\text{CQ}} = k/f = 43200$ and $T_{\text{CP}} = 3k/f = 129600$, so we get

 - $A \ge 43200$,
 - $R \ge 480$,
 - $K \ge 1920$

for $U=90$. These values are already much larger than the ones listed here, and they have not yet considered $T_{\text{heal}}$.

CIP-0140/README.lagda.md
```diff
@@ -0,0 +1,1568 @@
...
+
+[^11]: https://pooltool.io/networkhealth
+
+[^12]: https://developers.cardano.org/docs/operate-a-stake-pool/hardware-requirements/
+
+[^13]: The AWS cost is quite hard to estimate up-front. The $150 base price is a rough average of various instances options in the target range. Google, AWS and Azure prices are based on 100% uptime and at least 1-year reservation for discounts. Cloud providers only charge _outgoing_ network traffic. The actual cost per GB depends on the destination, this table assumes all outbound traffic will be targeted outside of the provider which obviously won't be true, so it should be treated as an upper bound.
+
+#### Persistent storage
+
+Under similar assumptions, we can estimate the storage requirements entailed by Peras: ignoring the impact of cool-down periods, which last for a period at least as long as $k$ blocks, the requirement to store certificates for every round increases node's storage by about **20%**. Votes are expected to be kept in memory so their impact on storage will be null.
```
amesgen, line 1454: One thing to mention here is that we can't seem to be able to prevent adversaries from including useless certificates on chain when Peras is *not* in a cooldown (also see "Changes to the Ledger " in https://tweag.github.io/cardano-peras/peras-design.pdf for more details).

CIP-0PRS/README.lagda.md
amesgen: I think it makes sense to mention in the CIP that currently, it is not planned to have direct monetary incentivization to participate in Peras, in particular because it might be rather non-trivial how to do so on-chain.

CIP-0140/README.lagda.md
bwbush: @corettis?

CIP-0140/README.lagda.md
bwbush: @corettis?

CIP-0140/README.lagda.md
bwbush: @corettis?

CIP-0140/README.lagda.md
```diff
@@ -0,0 +1,1568 @@
...
+- $\mathsf{cert}^\prime \gets \mathsf{cert}_\text{genesis}$: the latest certificate seen
+- $\mathsf{cert}^* \gets \mathsf{cert}_\text{genesis}$: the latest certificate on chain
+
+A [*fetching*](#fetching) operation occurs at the beginning of each slot:
+
+- Fetch new chains $\mathcal{C}\_\text{new}$ and votes $\mathcal{V}\_\text{new}$.
+- Add any new chains in $\mathcal{C}\_\text{new}$ to $\mathcal{C}$, add any new certificates contained in chains in $\mathcal{C}\_\text{new}$ to $\mathsf{Certs}$.
+    -  Discard any equivocated blocks or certificates: i.e., do not add them to $\mathcal{C}$ or $\mathsf{Certs}$.
+- Add $\mathcal{V}\_\text{new}$ to $\mathcal{V}$ and turn any new quorum in $\mathcal{V}$ into a certificate $\mathsf{cert}$ and add $\mathsf{cert}$ to $\mathsf{Certs}$.
+    -  Discard any equivocated votes: i.e., do not add them to $\mathcal{V}$.
```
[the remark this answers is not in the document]
bwbush: @corettis?

CIP-0140/README.lagda.md
```diff
@@ -0,0 +1,1568 @@
...
+- $\mathcal{V} \gets \emptyset$: set of votes
+- $\mathsf{Certs} \gets \{\mathsf{cert}_\text{genesis}\}$: set of certificates
+- $\mathsf{cert}^\prime \gets \mathsf{cert}_\text{genesis}$: the latest certificate seen
+- $\mathsf{cert}^* \gets \mathsf{cert}_\text{genesis}$: the latest certificate on chain
+
+A [*fetching*](#fetching) operation occurs at the beginning of each slot:
+
+- Fetch new chains $\mathcal{C}\_\text{new}$ and votes $\mathcal{V}\_\text{new}$.
+- Add any new chains in $\mathcal{C}\_\text{new}$ to $\mathcal{C}$, add any new certificates contained in chains in $\mathcal{C}\_\text{new}$ to $\mathsf{Certs}$.
+    -  Discard any equivocated blocks or certificates: i.e., do not add them to $\mathcal{C}$ or $\mathsf{Certs}$.
```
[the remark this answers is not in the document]
bwbush: @corettis?

CIP-0140/README.lagda.md
bwbush: Fixed in f3e8326e8f2812648fafc657a0d374c9033976bb and 4b3786049b38cfbc4b3bd8f0793a744f0d38a27d.

CIP-0140/README.lagda.md
bwbush: Fixed in 8df81250d0d28dfa1a4cdaa8b42cb4cd054e3b13.

CIP-0140/README.lagda.md
```diff
@@ -0,0 +1,1568 @@
...
+| ----------------------- | --------------- | ------- | ----------------------------------------------------------------------------------------- | -------------------------------------------------------- | -------------------------------------------------------------------------------------------- |
+| Round length            | $U$             | slots   | The duration of each voting round.                                                        | $U \geq \Delta$                                          | All of a round's votes must be received before the end of the round.                         |
+| Block selection offset  | $L$             | slots   | The minimum age of a candidate block for being voted upon.                                | $\Delta < L \leq U$                                      | Rule VR-1B will fail if the candidate block is older than the most recently certified block. |
```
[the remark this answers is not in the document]
bwbush: Fixed in 66a58e8b87a2a4835f8fad8d6fbe5c8ef1500284.

CIP-0140/README.lagda.md
```diff
@@ -0,0 +1,1568 @@
...
+   [ header
+   , transaction_bodies         : [* transaction_body]
+   , transaction_witness_sets   : [* transaction_witness_set]
+   , auxiliary_data_set         : {* transaction_index => auxiliary_data }
+   , invalid_transactions       : [* transaction_index ]
++  , ? peras_cert               : votes_certificate
+   ]
+```
+
+[Votes](#votes) are serialized in the following CDDL.
```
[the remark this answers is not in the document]
bwbush: Fixed in a3b8c2e357fd9bca29ea5c5308c389599a5443f5.

CIP-0140/README.lagda.md
bwbush: Fixed in 2c4e4ef471ddca0f17973f7507de2cfc2437ae0d

bwbush:
@rphair , I revised in response to your comments on formatting etc.

I've also revised to resolve several of @amesgen's helpful comments and suggestions.

I've asked the [Sandro](@corettis) to chime in on four of @amesgen's remaining comments.

My preference is to finalize the CIP sooner rather than later, as it is in sync with our current view of Peras. There will likely be a Peras implementation architecture/design workshop late in March, but we could create a new PR against this CIP if needed.

rphair (approved):
thanks @bwbush for the maintenance & the great improvements along the way.  I'll invite the other CIP editors to re-review, and if favourable as I expect (and no further technical reservations expressed by other reviewers) then we might mark this `Last Check` for our next scheduled CIP meeting on the 18th (goal: to confirm & merge it in the course of that meeting).

CIP-0140/README.lagda.md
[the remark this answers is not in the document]
corettis: I agree with @amesgen:

1. One cannot ignore equivocated blocks. A sync mechanism has to be implemented that is "equivalent" to receiving all equivocated blocks. (In Praos this is achieved by following peers' chains via chain sync.)
2. The system has to be parametrized s.t. certificates are unique. In the future, one might consider a version of Peras that allows for smaller committee sizes, resulting in certificate equivocation. However, this requires a much more involved analysis.

CIP-0140/README.lagda.md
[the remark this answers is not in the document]
corettis: I again agree with @amesgen.

1. One cannot ignore equivocating nodes.
2. One can, however, instruct nodes to only download one vote per stakeholder and voting round and use certificate sync to resolve situations in which nodes have downloaded different equivocated nodes.

CIP-0140/README.lagda.md
corettis: Yes, this seems difficult to prevent. However, it shouldn't bother us too much if we can get small certificates (and hence not too much space is wasted).

Ryun1 (approved):

CIP-0140/README.md
```diff
@@ -0,0 +1,1572 @@
...
+The $L$ parameter is the minimum age of a candidate block for being voted upon, measured in slots.
+
+```agda
+        L : ℕ
+```
+
+<span id="#expiration"/>The $A$ parameter is the maximum age for a certificate to be included in a block, measured in slots.
+
+```agda
+        A : ℕ
```
amesgen, line 258: $A$ is used in the condition $r - \mathsf{round}(\mathsf{cert}^\prime) \leq A$, where the right hand side is a number of rounds, but $A$ is measured in slots as stated here. This is incorrect (confirmed with @corettis), $A$ should rather be measured in rounds (and therefore be equal to $R$).

I guess the Agda specification could benefit from using some unit system to rule out these kinds of bugs ("natural number blindness").

PR #862, opened 2024-07-23
Validation Zones

polinavino:
We propose a set of changes that revolve around validation zones, a construct for allowing certain kinds of underspecified transactions. In particular, for the Babel-fees usecase we discuss here, we allow transactions that specify part of a swap request. A validation zone is a list of transactions such that earlier transactions in the list may be underspecified, while later transactions must complete all partial specifications. In the Babel-fees usecase, the completion of a specification is the fulfillment of a swap request. We discuss how validation zones for the Babel fees usecase can be generalized to a template for addressing a number of use cases from CPS-15.

PR #873, opened 2024-08-05
CIP-0130? | Transaction Pieces

fallen-icarus:
This CIP suggests enabling users to create and sign pieces of transactions that can then be communicated and batched off-chain. This batch can then be submitted as a completed transaction without requiring any coordination among users. Due to the way it works, this CIP would make Cardano's basic transaction a natural intent settlement system which means it would enable direct support for things like Babel Fees and off-chain trading. Thanks to the eUTxO model, it is able to achieve this without sacrificing any of Cardano's determinism. All that is really required for the implementation is a re-organization of the cddl.

This CIP is a potential solution to [CPS-0015](https://github.com/cardano-foundation/CIPs/pull/779), and an alternative to [Validation Zones](https://github.com/cardano-foundation/CIPs/pull/862) and [Smart Transactions](https://github.com/input-output-hk/Developer-Experience-working-group/issues/47).

---
[rendered proposal](https://github.com/fallen-icarus/CIPs/blob/transaction-pieces/CIP-%3F%3F%3F%3F/README.md)

### DIFF
--- a/CIP-????/README.md
+++ b/CIP-????/README.md
@@ -5,7 +5,7 @@ Status: Proposed
 Category: Ledger
 Authors:
     - fallen-icarus <modern.daidalos@gmail.com>
-Implementors: N/A
+Implementors: []
 Discussions:
     - https://github.com/cardano-foundation/CIPs/pull/862#discussion_r1700138228
     - https://github.com/cardano-foundation/CIPs/pull/862#issuecomment-2266689423
@@ -424,7 +424,7 @@ transaction.
 > TODO: Does the collateral claiming mechanism (in the case of a phase 2 failure) need to be
 > changed? I do not believe so, but I am not entirely sure.
 
-## Rational: how does this CIP achieve its goal?
+## Rationale: how does this CIP achieve its goal?
 
 Thanks to the eUTxO model, transactions can safely be broken into pieces and aggregated off-chain
 without having to sacrifice any determinism. By leveraging this feature, Cardano would gain two
@@ -517,13 +517,13 @@ this is not always possible/feasible.
 
 ### Acceptance Criteria
 
-- [] It is approved by the community and the ledger/plutus teams.
+- [ ] It is approved by the community and the ledger/plutus teams.
 
 ### Implementation Plan
 
-- [] The ledger spec is updated to use transaction pieces.
-- [] Implementation of the Cardano node with a new era/hardfork.
-- [] Deployed on testnet/mainnet
+- [ ] The Ledger spec is updated to use transaction pieces.
+- [ ] Implementation in the Cardano node with a new Ledger era / hardfork
+- [ ] Deployed on testnets / Mainnet
 
 ## Copyright
 
--- a/CIP-????/README.md
+++ b/CIP-????/README.md
@@ -1,0 +1,530 @@
+---
+CIP: ????
+Title: Transaction Pieces
+Status: Proposed
+Category: Ledger
+Authors:
+    - fallen-icarus <modern.daidalos@gmail.com>
+Implementors: N/A
+Discussions:
+    - https://github.com/cardano-foundation/CIPs/pull/862#discussion_r1700138228
+    - https://github.com/cardano-foundation/CIPs/pull/862#issuecomment-2266689423
+Solution-To: 
+    - CPS-0015? | Intents For Cardano
+Created: 2024-08-05
+License: CC-BY-4.0
+---
+
+## Abstract
+
+Cardano currently requires cryptographic keys to sign the entire transaction. As a result,
+coordination is required whenever there are UTxOs being spent from pubkey addresses that belong to
+different parties. However, this is unnecessary since parties really only need to approve the parts
+of the transaction that are relevant to them. This CIP looks to enable users to create **pieces** of
+transactions that can then be signed by the user(s) relevant to that piece. Users that have nothing
+to do with a specific transaction piece *do not need to sign that piece*. These signed pieces can
+then be aggregated off-chain and submitted together as a single complete transaction, without
+requiring coordination or any other action from the respective users. By eliminating the need for
+users to coordinate signing transactions, combined with the fact that transaction pieces can be
+individually unbalanced, the basic Cardano transaction would become a natural intent settlement
+system. This would open up the possibility for use cases such as Babel Fees and high frequency
+off-chain trading. Furthermore, thanks to the eUTxO model, transaction pieces have the same level of
+determinism as whole transactions.
+
+## Motivation: why is this CIP necessary?
+
+A key assertion underlying this CIP is **a batch of properly signed transaction pieces is
+functionally equivalent to a single transaction signed by all parties**. To highlight the truth of
+this assertion, consider two complementary atomic swap transaction pieces:
+
+- Alice wants to swap 10 ADA for 10 DJED so she creates a transaction piece with:
+    - 1 input from her address with 10 ADA
+    - 1 output to her address with 10 DJED
+- Bob wants to swap 10 DJED for 10 ADA so he creates a transaction piece with:
+    - 1 input from his address with 10 DJED
+    - 1 output to his address with 10 ADA
+
+Individually, neither of these transaction pieces are balanced transactions which means they can't
+be submitted to the blockchain. But if they are put together, the combination *is* a balanced
+transaction. Transaction pieces must be included all-or-nothing which means Alice's input can only
+be spent if the transaction actually includes her required output. 
+
+Given the all-or-nothing requirement, does it matter that Alice did not sign Bob's piece or the
+overall transaction? No, it doesn't because spending Bob's input doesn't require a signature from
+Alice, and she was guaranteed to get the output she wanted. The same is true for Bob's signature.
+There is no reason why the batching of these two transaction pieces should invalidate the
+signatures. Requiring Bob to also sign Alice's transaction piece is like requiring a smart contract
+to validate the spending of all UTxOs in a transaction, even those that aren't protected by that
+smart contract.
+
+Transaction pieces are the perfect model of an intent since *they are inherently flexible* (ie, a
+single piece can encompass multiple inputs, outputs, mints, etc) and *they can be unbalanced*. Being
+able to create unbalanced intents is critical to a healthy intent ecosystem because most intents are
+fundamentally only pieces of a transaction. Look back at Alice's and Bob's pieces above. The atomic
+swap intents don't say anything about the counter-party to the swap, and they shouldn't need to.
+Requiring users to specify who the counter-party should be fundamentally limits who can actually
+fulfill the intent. Without being able to create unbalanced transaction pieces, users would be
+forced to specify inputs and outputs for the potential counter-party just to finish balancing the
+transaction. 
+
+Another benefit of not having to specify counter-party information is that these transaction pieces
+can be trivially batched together without requiring coordination. As a result, a single transaction
+can trivially satisfy the intents of many complete strangers. In short, being able to properly model
+unbalanced intents makes transaction pieces a very natural and efficient intent settlement system.
+
+> [!IMPORTANT]
+> Transaction pieces effectively give another entity the right to spend your UTxOs as long as
+> certain conditions are satisfied in the same transaction. The key thing that sets this proposal
+> apart from other similar proposals is the fact that **the rest of the transaction is irrelevant**.
+> Smart contracts can already operate this way which is why it is so easy to batch DeFi orders (aka
+> intents) of complete strangers; this CIP extends this functionality to cryptographic keys.
+
+### Babel Fees
+
+Imagine if Alice wants to buy a sword NFT worth 25 GEMS and she expects the transaction fee to be
+about 1.4 GEMS. Let's say she has a UTxO available with 30 GEMS. She can create a single transaction
+piece with:
+
+- 1 input from her address with 30 GEMS
+- 1 output to her address with 3.4 GEMS + the sword NFT
+
+Alice can sign this piece and send it off-chain to the store selling the NFT.
+
+> [!IMPORTANT] 
+> Alice does not actually know how much the transaction fee will be when she creates her transaction
+> piece since the fee depends on the ultimate batch of transaction pieces. However, this is not a
+> problem because Alice is effectively setting the fee she is willing to pay. The more she is
+> willing to pay, the faster the store will fulfill her intent since it means her intent is more
+> flexible in how it can be profitably batched with other intents. In other words, transaction
+> pieces enable Babel Fees to naturally support its own intrinsic fee market.
+
+The store can create its own complementary transaction piece to finish balancing Alice's piece. For
+example, its piece could be:
+
+- 1 input with the sword NFT + the amount of ADA actually required for the fee
+- 1 output with 26.4 GEMS
+
+The store can batch its transaction piece with Alice's to create a full transaction that can now be
+submitted to the blockchain. Since Alice already signed her piece, the store does not need Alice's
+permission to submit the batch. The store can even include transaction pieces from other users and
+adjust its own piece to finish balancing the batch.
+
+> [!IMPORTANT] 
+> Since the store actually has all of the pieces that will be submitted in the batch, it has all of
+> the information it needs to deterministically calculate the transaction fee (and collateral)
+> required for the batch. This means this CIP does not sacrifice the ability for submitters to know
+> how much a transaction will cost before actually submitting it. The store can use this information
+> to figure out the best batch of pieces to make a profit while still satisfying the users' intents.
+
+### High Frequency Off-Chain Trading
+
+Day-trading is a common occurrence on exchanges, and it involves users creating and updating asking
+prices with relatively high frequency. However, if these price updates were to happen on-chain,
+each update would incur a transaction fee and would also take up block space that would otherwise go
+to more critical activities. Using transaction pieces, it is trivial to support high frequency
+off-chain price updates followed by efficiently batched settlement on-chain.
+
+Consider Alice's atomic swap transaction piece again, this was effectively a limit order with an asking
+price of 1 ADA/DJED. Off-chain aggregators can enable users to submit new transaction pieces for a
+trade, and the aggregator can use the most recent transaction piece for that trade. For example, if
+Alice created a new transaction piece with:
+
+- 1 input from her address with 10 ADA (the same input from the prior transaction piece)
+- 1 output to her address with 20 DJED
+
+She can submit this new piece off-chain to effectively update her asking price to 0.5 ADA/DJED.
+
+Using the off-chain aggregator like this requires a level of trust since the aggregator can always
+choose to still use the older transaction piece. However, users can add "kill-switches" to their
+transaction pieces to securely and trustlessly cancel all prior pieces. There are two main ways to
+do this:
+
+- Have the transaction piece reference a particular input that can be spent by the user.
+- Have the transaction piece spend a particular input that can also be directly spent by the user.
+
+With either of the above choices, spending the input in question will invalidate all transaction
+pieces with this input since the UTxO would no longer exist. Both types of kill-switches can even be
+used within the same transaction piece. Using both would give users the flexibility of either
+canceling a particular limit order or a batch of limit orders (eg, all limit orders for a particular
+trading pair).
+
+The aggregator can take all (still valid) transaction pieces from users and batch them together,
+possibly with on-chain liquidity sources, to submit them in a single transaction. Since the
+transaction pieces did not need to specify a counter-party, these trades can be matched against
+**any** counter-party, no matter its form. It is even trivially possible to match a single large
+order against several smaller orders.
+
+With this approach, all high frequency activity can happen off-chain while settling the actual
+trades on-chain in the most efficient way possible.
+
+This approach can even be combined with Babel Fees to enable day-traders to pay the transaction fee
+in a native asset specified by the aggregator.
+
+## Specification
+
+Transaction pieces **must** be included all-or-nothing, and a transaction would become
+**exclusively** a list of signed pieces (ie, an output cannot be specified outside of a transaction
+piece). The current one-party style transaction can still be represented by a transaction with only
+one piece where that piece is a fully balanced transaction body.
+
+### CDDL
+
+#### Transaction
+
+Below is the current cddl representation of a transaction:
+
+```cddl
+transaction =
+  [ transaction_body
+  , transaction_witness_set
+  , bool
+  , auxiliary_data / null
+  ]
+```
+
+It contains a single transaction body and a single witness set. This will be changed to contain a
+list of transaction bodies, each paired with their own witness set.
+
+The transaction witness set is currently:
+
+```cddl
+transaction_witness_set =
+  { ? 0: [* vkeywitness ]
+  , ? 1: [* native_script ]
+  , ? 2: [* bootstrap_witness ]
+  , ? 3: [* plutus_v1_script ]
+  , ? 4: [* plutus_data ]
+  , ? 5: [* redeemer ]
+  , ? 6: [* plutus_v2_script ] ; New
+  }
+```
+
+Smart contracts will still be executed against the full transaction (this will be discussed in a
+later section), but all of the required information for an execution (eg, script, datum, and
+redeemer) can be found within the corresponding transaction piece. This enables the witness set to
+remain unchanged.
+
+The new transaction would then be:
+
+```cddl
+transaction_piece =
+  { 0: transaction_body
+  , 1: transaction_witness_set ; The witness set for this transaction body.
+  }
+
+transaction =
+  [ transaction_pieces    : [* transaction_piece]
+  , bool
+  , auxiliary_data / null
+  ]
+```
+
+The order of the transaction pieces list is significant since it represents the order in which the
+pieces must be combined, and ultimately determines the order of the inputs and outputs. How to
+combine the pieces will be discussed later in the spec. 
+
+#### Block
+
+The cddl for a block is currently:
+
+```cddl
+block =
+  [ header
+  , transaction_bodies         : [* transaction_body]
+  , transaction_witness_sets   : [* transaction_witness_set]
+  , auxiliary_data_set         : {* transaction_index => auxiliary_data }
+  , invalid_transactions       : [* transaction_index ]
+  ]; Valid blocks must also satisfy the following two constraints:
+   ; 1) the length of transaction_bodies and transaction_witness_sets
+   ;    must be the same
+   ; 2) every transaction_index must be strictly smaller than the
+   ;    length of transaction_bodies
+```
+
+Just like with the transaction cddl, the block's transaction bodies and the transaction witness sets
+will be merged, and the constraints must be updated accordingly:
+
+```cddl
+; This is re-posted here for reference.
+transaction_piece =
+  { 0: transaction_body
+  , 1: transaction_witness_set ; The witness set for this transaction body.
+  }
+
+; A specific transaction in the block is a specific batch of transaction pieces. 
+block_transaction = [* transaction_piece]
+
+block =
+  [ header
+  , block_transactions   : [* block_transaction] ; This is new
+  , auxiliary_data_set   : {* transaction_index => auxiliary_data }
+  , invalid_transactions : [* transaction_index ]
+  ]; Valid blocks must also satisfy the following constraint:
+   ; 1) every transaction_index must be strictly smaller than the
+   ;    length of block_transactions
+```
+
+### Transaction Size Limit
+
+A transaction broken into pieces should still have approximately the same total size as the whole
+transaction. There are a few things to note:
+
+- If two transaction pieces require the same witness (ie, both pieces require a signature from
+Alice), this witness would now effectively appear twice.
+- Certain items can appear in multiple pieces (eg, the same input can be spent/referenced, tokens can
+be minted/burned that would otherwise cancel out, etc).
+
+Despite the above, it is probably fine to leave the maximum transaction size as is, at least to
+start. The transaction fee would naturally incentivize users/aggregators to minimize the amount of
+duplication in a transaction.
+
+### Block Size Limit
+
+Due to the possible duplicate items in each transaction and the fact that a block is made up of
+multiple transactions, the issues in the Transaction Size Limit section are slightly exacerbated.
+However, it should still be fine to keep the current block size limit, at least initially. Just like
+with transactions, the transaction fee would naturally incentivize users to minimize duplications.
+
+### Smart Contract Execution
+
+While the cryptographic keys only need to sign the relevant transaction pieces, the smart contracts
+would be executed against the full transaction. This requirement enables the current smart contract
+interface to remain mostly untouched. The flow is essentially:
+
+1. Deterministically derive the whole transaction from the list of pieces.
+2. Generate the script context for the whole transaction.
+3. Evaluate the smart contract using this script context.
+
+When combining the transaction pieces, the required executions (ie, script + datum + redeemer) can
+be built up as well. Duplicate executions should be removed. For example, if two transaction pieces
+look to spend the same smart contract input using the same redeemer, there is no reason to evaluate
+this scenario twice since they are guaranteed to have the same result.
+
+### Transaction Fee and Collateral
+
+Both the fee and collateral fields of the transaction body can be set in any piece, but only the
+aggregator needs to set this (in their final piece that is used to finish balancing the batch). They
+have all of the pieces that will go into the batch which means they can deterministically calculate
+the required fee and collateral for the batch they are creating. Their final piece can be updated
+after determining the required fee and collateral.
+
+By having the aggregator put up the collateral, they are naturally incentivized not to include any
+transaction pieces with smart contract executions that will fail. They can know whether the smart
+contract execution will fail ahead of submission due to the determinism of the eUTxO model being
+preserved, despite the transaction body being broken into pieces. The same incentives apply for the
+transaction fee since the aggregator will want to minimize their own costs, and therefore, won't
+waste execution resources.
+
+### Transaction Piece Monoid Instance
+
+The monoid instance for transaction pieces defines the rules for combining the pieces. In order to
+be able to deterministically derive the full transaction, this instance must be standardized. The
+monoid instance requires answering two questions:
+
+1. What is an empty transaction piece?
+2. How are two transaction pieces combined?
+
+The first question is simple to answer: an empty transaction piece is an empty transaction body.
+Users fill out as much information as they wish when creating their transaction piece. The rest of
+the fields can remain empty.
+
+> [!IMPORTANT]
+> It doesn't really make sense to create a transaction piece with an input but no output, or an
+> output but no input. Therefore, these fields can remain required fields. The fee is also required,
+> but that can be set to zero in the pieces not paying the fee. All other fields in the transaction
+> body are already optional.
+
+The second question requires more consideration, but as a general rule, as long as there are no
+incompatibilities between transaction pieces, they can be combined into the full transaction. Any
+transaction that contains incompatible pieces should be rejected.
+
+#### Ordering
+
+The creator of the transaction piece gets to specify the order of inputs, outputs, reference inputs,
+and collateral inputs **within** their piece. Then, the aggregator (the one that batches pieces and
+submits the transaction) gets to decide the order **between** the pieces.
+
+For example, if Alice creates a piece with `input1` followed by `input2` and Bob creates a piece
+with `input3` followed by `input4`, then Charlie (the aggregator) only has two options:
+
+```
+[input1, input2, input3, input4] OR [input3, input4, input1, input2]
+```
+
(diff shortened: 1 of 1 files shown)


fallen-icarus:
@Quantumplation @michaelpj @polinavino This CIP elaborates on what I was talking about in the Validation Zones PR [discussion](https://github.com/cardano-foundation/CIPs/pull/862#issuecomment-2266689423). I think it covers the same use cases with a fraction of the complexity.

I'm going to tag other people that have expressed interest in intents:
@colll78 @AndrewWestberg @vlasin

CIP-????/README.md
```diff
@@ -0,0 +1,530 @@
...
+fulfill the intent. Without being able to create unbalanced transaction pieces, users would be
+forced to specify inputs and outputs for the potential counter-party just to finish balancing the
+transaction. 
+
+Another benefit of not having to specify counter-party information is that these transaction pieces
+can be trivially batched together without requiring coordination. As a result, a single transaction
+can trivially satisfy the intents of many complete strangers. In short, being able to properly model
+unbalanced intents makes transaction pieces a very natural and efficient intent settlement system.
+
+> [!IMPORTANT]
```
rphair, line 75:
```suggestion
> ⚠️ **Important**<br>
```
This looks great on GitHub there are sites derived from the CIP source material where this construction might not look very good, or at least will appear unstyled.  Currently no merged CIPs use this construction & you might be the first. 

Maybe you can tell from https://github.com/cardano-foundation/cf-cip-frontend how it would look (I searched for "admonition" on the source code & didn't find anything).  All things considered it might be suitable to leave these the way they are, but in general I think someday we will want CIPs with portable primitives (i.e. classic Markdown + the math extensions).

CIP-????/README.md
```diff
@@ -0,0 +1,530 @@
...
+
+### Phase 2 Validation
+
+Phase 2 validation remains unchanged since the smart contracts are still executed against the whole
+transaction. 
+
+> TODO: Does the collateral claiming mechanism (in the case of a phase 2 failure) need to be
+> changed? I do not believe so, but I am not entirely sure.
+
+## Rational: how does this CIP achieve its goal?
```
rphair, line 427:
```suggestion
## Rationale: how does this CIP achieve its goal?
```

CIP-????/README.md
```diff
@@ -0,0 +1,530 @@
...
+
+### Acceptance Criteria
+
+- [] It is approved by the community and the ledger/plutus teams.
+
+### Implementation Plan
+
+- [] The ledger spec is updated to use transaction pieces.
+- [] Implementation of the Cardano node with a new era/hardfork.
+- [] Deployed on testnet/mainnet
```
rphair, line 526:
```suggestion
- [ ] It is approved by the community and the ledger/plutus teams.

### Implementation Plan

- [ ] The Ledger spec is updated to use transaction pieces.
- [ ] Implementation in the Cardano node with a new Ledger era / hardfork
- [ ] Deployed on testnets / Mainnet
```
Fixing typography & some spelling.  I do think we need more detail in _Acceptance Criteria_ but I think these will emerge from discussion points over a long Ledger + community review.

rphair (commented):
@fallen-icarus thanks for continuing that premise.  I've added this PR for introduction to tomorrow's CIP meeting agenda: https://hackmd.io/@cip-editors/94

I would be really happy to keep this evolving along with _Validation Zones_ and any other approaches to intent-based transactions.  At some point soon we would need the Ledger team's opinion about feasibility so also tagging them (@lehins @WhatisRT).

This is not a detailed review, since the point at which I could assess all the technical propositions and/or compare/contrast with Validation Zones would be a long time coming: this is just a skim for format consistency, etc., while we get this into the pipeline.

vlasin:
I support this solution as a significantly less complex alternative.

A suggestion: smart contracts could get some extra efficiency from this CIP if there is a way to access the current transaction piece inside the script. Oftentimes, a script needs to find a specific input satisfying some conditions. Since transaction inputs are not sorted by the signers, the signers don't have control over the input order. So, in transactions with many inputs, a lot of ExUnits may be spent just on finding the right input, which is usually from the same transaction piece as the script itself.

CIP-????/README.md
```diff
@@ -0,0 +1,530 @@
...
+- Only UTxOs that exactly fit those placeholders can be used
+- The counter-party can't add or remove placeholders after the fact
+
+Imagine if Alice wants to swap 100 ADA for 50 DJED; this says nothing about how the counter-party
+will satisfy her swap. Perhaps Bob (the counter-party) wants to provide the DJED, but it is
+currently broken up across three UTxOs. And perhaps he wants the 100 ADA he receives to be stored in
+two separate outputs. If Alice specifies only one input placeholder and one output placeholder in
+her intent transaction, Bob might be unable to act as her counter-party. If Bob really wanted to
+still trade with Alice, he would have to find a way to merge his inputs and outputs into one input
+and one output. This may not always be possible and Bob may have to give up on being Alice's
```
WhatisRT, line 475: This is actually quite easy to do with validation zones. Bob just needs to spend his three outputs and Alice's Ada output and negative Djed output, and make two new outputs containing the Ada. The only extra effort is to spend Alice's two outputs.

CIP-????/README.md
```diff
@@ -0,0 +1,530 @@
...
+
+Again, the counter-party issue and extra complexity in Validation Zones are simply due to not being
+able to accurately model unbalanced intents. Since this CIP *can* model unbalanced intents, it can
+cover all of the same use cases while only adding a fraction of the complexity.
+
+It is also worth noting that, since the Validation Zones design uses otherwise unnecessary
+placeholders and extra transactions just to satisfy these placeholders, the impact Validation Zones
+would have on the average block size is likely equivalent to, if not larger than, the impact this
+CIP would introduce.
+
```
WhatisRT, line 502: Compared to the validation zones approach, this approach appears to be somewhat simpler indeed, but it also has some restrictions. For example, this approach doesn't allow for atomic batch transactions in general. A transaction piece cannot spend an output generated by another transaction piece in the same transaction. I see this as a major drawback, and it should be discussed here.

CIP-????/README.md
```diff
@@ -0,0 +1,530 @@
...
+and one output. This may not always be possible and Bob may have to give up on being Alice's
+counter-party. In other words, because Alice was forced to specify placeholders in her intent, she
+missed out on a potential trading partner.
+
+DApps would also be unable to be direct liquidity sources for these intents due to not always being
+able to match the value specified in each placeholder. If the input place holder is for 10.3 ADA but
+the liquidity source UTxO actually has 20.5 ADA, this liquidity source cannot be used despite it
+having enough liquidity. The Validation Zones CIP argues that DApps can just add support for
+breaking-off/merging UTxOs so that the liquidity source UTxOs can be forced into the proper forms
+for the required placeholders. However, this would dramatically increase the complexity and attack
```
WhatisRT, line 484: I don't see where this is discussed in the Validation Zones CIP. But I also don't see a big problem here, just by going on with the same logic from the last paragraph. A matching transaction can always spend the 'problematic' outputs and then construct the remainder of the transaction as it would construct the transaction piece. I don't see where this impacts 'dramatically increases complexity'.

fallen-icarus:
@gitmachtl A transaction piece is literally just a full transaction body that can be unbalanced. Therefore, hw wallets should already be (mostly) capable of signing pieces. If hw wallets can hash/verify/sign a transaction body, then it can hash/verify/sign a piece since they are the same thing.

>  But without the whole transaction in place the total hash of the transaction would not be correct.

When signing the piece, the hardware wallet would hash the piece itself. This is all the signer cares about anyway. Since the piece is included in its entirety on chain, users can easily verify their transaction piece was properly included using the hash of the piece. The off-chain tooling would definitely need to be updated to enable this verification, but I don't see anything that would make this difficult/impossible.

In this new paradigm, hw wallets only need to be able to sign transaction pieces. You can still have a piece be a fully balanced transaction so there is no loss of functionality with this change.

fallen-icarus:
> A suggestion: smart contracts could get some extra efficiency from this CIP if there is a way to access the current transaction piece inside the script. Oftentimes, a script needs to find a specific input satisfying some conditions. Since transaction inputs are not sorted by the signers, the signers don't have control over the input order. So, in transactions with many inputs, a lot of ExUnits may be spent just on finding the right input, which is usually from the same transaction piece as the script itself.

@vlasin I'd have to think about this. This would definitely add a lot of complexity and actually require a change to the smart contract interface. I agree there would be some efficiency gains, but since the aggregator can control the order of the pieces themselves, the ExUnits spent can already be minimized quite a bit. I'm not sure if the rest of the gains would be worth the extra complexity.

CIP-????/README.md
```diff
@@ -0,0 +1,530 @@
...
+
+Due to the possible duplicate items in each transaction and the fact that a block is made up of
+multiple transactions, the issues in the Transaction Size Limit section are slightly exacerbated.
+However, it should still be fine to keep the current block size limit, at least initially. Just like
+with transactions, the transaction fee would naturally incentivize users to minimize duplications.
+
+### Smart Contract Execution
+
+While the cryptographic keys only need to sign the relevant transaction pieces, the smart contracts
+would be executed against the full transaction. This requirement enables the current smart contract
```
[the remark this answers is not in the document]
fallen-icarus: > Without a good reason for this, it's probably better to properly reflect this structure in `TxInfo`.

Personally, whichever option minimizes added complexity is my preferred choice. If we preserve the transaction as you are suggesting, the smart contract interface would need a pretty large change and DApps would need to be re-architected. Perhaps keeping the structure in `TxInfo` would be better in the long run, but I'm not sure. I do not know what changes would be required on the ledger/plutus side to actually preserve the structure in `TxInfo`.

@vlasin If we did preserve the structure, I believe this would address your [suggestion](https://github.com/cardano-foundation/CIPs/pull/873#issuecomment-2270494230).

gitmachtl:
> A transaction piece is literally just a full transaction body that can be unbalanced. Therefore, hw wallets should already be (mostly) capable of signing pieces. If hw wallets can hash/verify/sign a transaction body, then it can hash/verify/sign a piece since they are the same thing.

maybe i am just confused, but how would an unbalanced transaction be ready for signing? if you change the lovelace/asset value of an output, the hash of the tx would not be the same again.

CIP-????/README.md
fallen-icarus: > A matching transaction can always spend the 'problematic' outputs and then construct the remainder of the transaction as it would construct the transaction piece. I don't see where this impacts 'dramatically increases complexity'.

I guess I'm just biased. Given there is a path that doesn't require these extra steps, why wouldn't you take that path? Why would you ever take the path of *more* resistance?

CIP-????/README.md
fallen-icarus: Atomic batch transactions seems orthogonal to intents (it isn't mentioned in the CPS), but I can still discuss it in the CIP.

@polinavino Perhaps there is a way to use transaction pieces as the foundation for zones?

fallen-icarus:
> if you change the lovelace/asset value of an output, the hash of the tx would not be the same again

You wouldn't change an output after signing. The transaction piece stays unbalanced. The aggregator can then fit unbalanced transaction pieces together, just like lego toys. The unbalanced transaction pieces end up balancing each other; no outputs need to be changed.

gitmachtl:
Got it, thx

CIP-????/README.md
fallen-icarus: See my other comment [here](https://github.com/cardano-foundation/CIPs/pull/873#discussion_r1705441782).

CIP-????/README.md
```diff
@@ -0,0 +1,530 @@
...
+
+1. Deterministically derive the whole transaction from the list of pieces.
+2. Generate the script context for the whole transaction.
+3. Evaluate the smart contract using this script context.
+
+When combining the transaction pieces, the required executions (ie, script + datum + redeemer) can
+be built up as well. Duplicate executions should be removed. For example, if two transaction pieces
+look to spend the same smart contract input using the same redeemer, there is no reason to evaluate
+this scenario twice since they are guaranteed to have the same result.
+
```
polinavino, line 302: This looks like a very promising design indeed! I am curious to understand how minting NFTs that require the spending of a specific UTxO entry would work, if two pieces are trying to mint the same NFT? Would you be able to locally test if your NFT contract validates?

CIP-????/README.md
```diff
@@ -0,0 +1,530 @@
...
+Title: Transaction Pieces
+Status: Proposed
+Category: Ledger
+Authors:
+    - fallen-icarus <modern.daidalos@gmail.com>
+Implementors: []
+Discussions:
+    - https://github.com/cardano-foundation/CIPs/pull/862#discussion_r1700138228
+    - https://github.com/cardano-foundation/CIPs/pull/862#issuecomment-2266689423
+Solution-To:
```
Ryun1, line 12:
```suggestion
```
adding to the header will break downstream tools

CIP-????/README.md
fallen-icarus: > I am curious to understand how minting NFTs that require the spending of a specific UTxO entry would work, if two pieces are trying to mint the same NFT? Would you be able to locally test if your NFT contract validates?

Hmm, I hadn't thought about testing just a piece, but it should be possible since a transaction piece is just a transaction body. This would just require a smart contract to be able to run against an unbalanced transaction body. Whether the smart contract would still succeed when aggregated with other pieces depends on the smart contract logic. So Alice can verify her piece will successfully validate *before* she signs, and the aggregator can verify it will still validate *after* adding it to the batch.

CIP-????/README.md
rphair: @fallen-icarus I'm going to withdraw this suggestion since for some reason the CIP author community has started to converge upon GitHub admonitions recently; e.g.: https://github.com/cardano-scaling/CIPs/blob/decentralized-message-queue/CIP-0XXX/README.md#mithril-extra-network-usage

If & when it appears awkwardly on derived sites then I'll argue it's those sites' obligation to add some support for this GitHub syntax & display it intelligently.  Please keep an eye on how this looks on the Dev Portal and `cips.cardano.org` after it is merged & raise a red flag if the admonitions make a poor showing there. :)

CIP-????/README.md
```diff
@@ -0,0 +1,530 @@
+---
+CIP: ????
```
```suggestion
CIP: 130
```
Please also retitle containing directory to `CIP-0130` and update the rendered proposal link in your OP 🎉

rphair (commented):
Agreed to assign number in today's CIP meeting mainly due to the good documentation & the specific, practical nature of the proposal.

AndrewWestberg:
I like this concept very much. Very elegant.

CIP-????/README.md
polinavino: In general, it opens up the possibility of an attack if the outcome of contract validation can be changed after you submit your transaction or transaction piece. I see two ways to deal with this :

1. restrict your design to only allow combining pieces that contain no Plutus smart contracts whatsoever

2. simplify Validation Zones to be more in line with what you are suggesting/be more implicit with value flow :
- revert changes to the UTxO rule, UTxO State, and Transaction data structures (leave them the same as in the current non-zone ledger design)
- remove new Plutus version requirement
- move the `produced = consumed` computation to the ZONE rule, and aggregate across all transactions in the zone, that is `Sum_{tx \in zone} consumed tx = Sum_{tx \in zone} produced tx`
- replace the cycle detection in the ZONE rule by in independency check, which checks that "for all inputs `(txid, ix)` in all transactions in a zone, `txid` is not the ID of another transaction in the zone
- only require the per-transaction component of the transaction fee to be paid once per zone

The validation zones simplification will all remove the need for the weird `requireTxs` song and dance, and an aggregator would be able to spend their liquidity pool once per zone to provide any missing value and claim any extra value within a single transaction. Scripts would operate without any changes. Flash loan attacks would be prevented by the independency check.

vlasin:
> @vlasin I'd have to think about this. This would definitely add a lot of complexity and actually require a change to the smart contract interface. I agree there would be some efficiency gains, but since the aggregator can control the order of the pieces themselves, the ExUnits spent can already be minimized quite a bit. I'm not sure if the rest of the gains would be worth the extra complexity.

Right, I think by controlling the order of pieces, we already get most of the efficiency gain I was talking about!

CIP-????/README.md
polinavino: 1. The difference between an account-based dApp doing things and the Boolean outcome of an EUTxO contract is that the dApp can (and has to) be written in a way that accommodates any possible account state to which it can be applied, eg price changes etc. An EUTxO contract fee cannot even be calculated without the complete fixed transaction data.  You can make any contract fail by bulking up the tx size. If you make a transaction that is valid, someone can always make your scripts fail by submitting it batched with whatever nonsense. To avoid this, you have to be able to prevent your transaction from being modified (this is currently possible only by signing it with a key that is required to sign it). Note that even if you yourself didn’t include collateral, and are relying on someone else to include it, the transaction batch they construct can also be messed up. This brings me to the Validation Zones modification - 

2. The change I propose actually allows you to model unbalanced intents by allowing whole transactions that are unbalanced, while the zones containing transactions must be balanced. It is somewhere in between the original design and what you propose. Instead of requests and fulfills, it simply allows assets to be missing from or be extra in a transaction. The preservation of value at the zone level ensures that regardless of who provides the missing assets and claims the extra ones , things balance out. Consider zone [Tx1; Tx2]

Tx1 - 
(Resolves) inputs: txIn -> (10 Ada, …)
Outputs:(5PolinaCoin, …)

Tx2 - 
(Resolved) inputs :txIn’ -> (5PolinaCoin, …)
Outputs :(10Ada, …)

The important thing is that txIn’ is not hash (Tx1), because of flash loans. No need to request 5 PolinaCoin or make an unlocked offer of 10Ada.

WhatisRT: The weakening to transaction balancing seems like a good middle-of-the-road approach. It means you can optionally have interdependencies but aren't forced to exactly specify everything. I suspect that in practice all it really adds is a minor optimization on transaction size, but that's nice as well.

fallen-icarus: I have a few questions:

- Transaction Fees - is it possible for unbalanced transactions within a zone to have a fee of zero? Particularly for babel fees, only one transaction in the zone needs to cover the fee for the entire zone. Multiple transactions in the zone could specify the fee (to pay their fair share) in which case the total fee paid would be the sum of all of the fees paid in the zone.
- Flash loans - I always thought flash loans implied no time passing; you borrow a loan and immediately pay it back "before the world notices". But that doesn't seem to be possible here since time is implied to pass: tx1 is processed before tx2. How is this any different than what currently happens where one transaction is processed after another in the same block? The only issue I can see is that the transaction fee would be lower now (assuming it is 0.15 ADA per zone now instead of 0.15 ADA per transaction). I don't see why `txIn'` can't be `= (hash (Tx1), _)`. If it could, it seems trivial to do an atomic batch of transactions since they are intrinsically linked.

polinavino: - i think it will not be possible to specify a fee of 0 (this is because the amount of fee specified in each transaction must go to the fee pot, and that's how the system knows how much to add), but it does not mean you have to provide the fee. Your transaction can simply be short that amount, and someone else's transaction needs to have that amount extra to make a valid zone

- the flash loan problem here really has to do with smart contracts : you should not be able to use a thread token or a role token in your transaction unless you spend an actual output containing one. If we allow transactions in a zone to be chained in this way, you dont actually need to spend a real output with a role token to use one, flash-loaning yourself one :

Tx1 -
(Resolved) inputs : txIn -> (out1)
Outputs: (1RoleToken, …) ,  (out2)

where out1 and out2 are some outputs containing the state of some contract which can only be updated if the transaction contains a RoleToken that allows such an update to be made

Tx2 -
(Resolved) inputs : (hash (Tx1), 1) -> (1RoleToken, …)

fallen-icarus: > you should not be able to use a thread token or a role token in your transaction unless you spend an actual output containing one.

Perhaps I misunderstood what you meant by `(hash (Tx1), _)`. I may also be misunderstanding what you mean by
> (Resolved) inputs: txIn -> (out1)`. 

The way I am thinking about it, the role token requirement is guaranteed to be true; there is no way to NOT spend the role token. No "resolving" is required.

A zone is just a list of transaction: `[tx1, tx2, tx3]`. But we can force a dimension of time on the list:
```
---- Time ---->
[tx1, tx2, tx3]
```

If we do this, then the above is just a mini-blockchain, and we can treat it as such. The zone can build up its own mini UTxO set, and transactions in the zone have access to the current state of `real UTxO set + zone UTxO set` at the time they are being processed. The zone can keep track of both the UTxOs spent from the real UTxO set and the UTxOs created/spent from the zone UTxO set. When all transactions in the zone have been fully validated, the net affect of the real UTxO set + the zone UTxO set is what actually gets applied to the blockchain.

In the above zone, `tx1` gets processed first. Imagine if `tx1` spends a real output `r1` and produces an output `z1`. The zone would log that `r1` has been spent from the real UTxO set and `z1` has been added to the zone UTxO set. `tx2` comes after `tx1` and can spend any remaining UTxO in the real UTxO set and the zone UTxO set. So `tx2` cannot spend `r1` because it was already spent by `tx1`, but it can spend both `r2` (another input from the real set) and `z1` to produce `z2`. `tx2` would call `z1` explicitly by output reference: `(hash(tx1), z1_index)`. `z1_index` is its index in `tx1`'s output list.

Now the new state is the real UTxO set minus `r1` and `r2`, plus the zone UTxO set of `z2`. `tx3` comes after `tx2` so it must use this new state. If it wants to spend `z2`, it must call it explicitly by output reference: `(hash(tx2), z2_index)`. `z2_index` is its index in `tx2`'s output list. Let's say `tx3` spends `z2` and produces `z3`. `tx3` cannot spend `z1` because it was already spent by `tx2`.

The net effect of this zone is `r1` and `r2` are both spent to produce `z3`. This is the change that would actually be recorded on the blockchain. 

With this method, if the role token was in `r1`, `tx1` moved it to `z1`. If `tx2` wants to execute the smart contract that requires the role token, it must now spend `z1` because that is where the token is at the time `tx2` is being processed. If `tx3` wants to use the role token as well, it must spend `z2` since that is where `tx2` moved the token to. Because time is enforced on the zone, the zone can know where the role token is at all times. When `z3` is actually recorded on-chain, it is the result of `r1` -> `z1` -> `z2` -> `z3`. The eUTxO model guarantees the result is the same as doing the updates separately (ie, in separate zones).

> [!Important]
> With this approach, building a transaction chain **requires** coordination between the transactions in the chain. However, I am making the assertion that: *coordination is always possible whenever transaction chaining is useful*.  If we accept this assertion, there is no need for any "resolving", and flash loans are impossible.

Strangers can't build up chains this way since they don't know each others' transactions (and therefore each others' output references), but why would strangers be chaining their transactions anyway? This is mainly useful for account style DApps that are controlled by batchers.

A batcher can personally build up the chain when trying to fulfill the user orders. They can create `tx1` first and use its hash in `tx2`, and then use `tx2`'s hash in `tx3`. Furthermore, since the batcher knows the order of the outputs in each transaction, they know what the resulting output references will be in advance. The batcher can even turn the user orders into chains (if the user transactions allow for it). If batchers need to build up a chain together, they can coordinate this off-chain.

Because the outputs are being called explicitly by output reference, it is trivial to check if the role token is actually being spent. The smart contracts don't even need to know they are being executed in a zone.

fallen-icarus: > i think it will not be possible to specify a fee of 0 (this is because the amount of fee specified in each transaction must go to the fee pot, and that's how the system knows how much to add)

I don't see why it is possible for the zone to know if it is balanced, but not know if it has enough ADA provided for the fee. The zone checks should have everything they need to calculate the total required fee for the zone, and it can compare this against the sum of the fee specified in each transaction within the zone. This would enable some transactions to have a fee of zero while still being part of a valid zone.

polinavino: > I don't see why it is possible for the zone to know if it is balanced, but not know if it has enough ADA provided for the fee. 

This is maybe possible? Removing the per-transaction fee check is likely possible, and if we do, entering a paid fee of 0 for a transaction would also be possible. I need to think more about this, though, and would be happy to discuss in CIP meeting.

I think i need a diagram for your previous comment ...

Here is a draft of the "Implicit Value Flow" validation zones version, it is much simpler than the original

https://github.com/cardano-foundation/CIPs/pull/862/commits/1fd1d6dc56fadca8b07009e795b3ec4eedc298cd#diff-560051716ec259fa49b3ef402bb6c1f4b7517d1bf400db3e735996986398bf96

polinavino: @AndrewWestberg @WhatisRT  if you feel like having a look also

CIP-????/README.md
```diff
@@ -0,0 +1,530 @@
...
+
+> [!IMPORTANT]
+> It doesn't really make sense to create a transaction piece with an input but no output, or an
+> output but no input. Therefore, these fields can remain required fields. The fee is also required,
+> but that can be set to zero in the pieces not paying the fee. All other fields in the transaction
+> body are already optional.
+
+The second question requires more consideration, but as a general rule, as long as there are no
+incompatibilities between transaction pieces, they can be combined into the full transaction. Any
+transaction that contains incompatible pieces should be rejected.
```
lehins, line 339: Any two transactions that have plutus scripts would be immediately incompatible.

Imagine a script that relies for there to be only a single input being spent or a single output being produced. As soon as you combine two transactions together they no longer validate. However, you can't know that until you try to combine them and run the scripts

There is another fundamental problem here. Imagine those inputs in two separate transactions are both locked by scripts. Those two transactions both provide script integrity hash that have potentially been signed. But whenever you combine those two transactions together you must change their redeemers and pass those redeemer modified in the script context to plutus scripts, which is totally screwing with the data that have been signed.

So, combining multiple pieces into one transaction idea is definitely out of the window

CIP-????/README.md
```diff
@@ -0,0 +1,530 @@
...
+> This would required [CIP-128](https://github.com/cardano-foundation/CIPs/pull/758) since the order
+> of the inputs and reference inputs would need to be preserved. There may be intents where the
+> order of these fields matter which is why the intent creator should get to specify their ordering.
+> The collateral inputs could be an exception, but if we are already making the change for inputs
+> and reference inputs, why not also make the change for collateral inputs?
+
+#### Duplicate Inputs
+
+It is valid for two separate transaction pieces in a batch to spend the same UTxOs (eg, a liquidity
+source UTxO). This UTxO can still only be spent once and the batch must still be fully balanced
```
lehins, line 368: This not valid. Their position in the input set can affect the execution outcome of a plutus script that locks them

CIP-????/README.md
```diff
@@ -0,0 +1,530 @@
...
+transaction =
+  [ transaction_pieces    : [* transaction_piece]
+  , bool
+  , auxiliary_data / null
+  ]
+```
+
+The order of the transaction pieces list is significant since it represents the order in which the
+pieces must be combined, and ultimately determines the order of the inputs and outputs. How to
+combine the pieces will be discussed later in the spec.
```
lehins, line 224: This is very true. However, I can't find any security mechanism that would preserve the order of transaction pieces. There is no guarantee that man in the middle or a nefarious pool will not reorder the transaction pieces in their favor

lehins (commented):
Unfortunately I see way too many problems with this proposal:

* change of transaction structure that would require a rewrite of half of the ledger implementation
* Idea of combining transactions (pieces) into one is just not sound. The common argument that is given in the CIP of rejecting any transactions that are incompatible in some way is way too dangerous and very hard to implement and foresee all the incompatibilities. Most importantly by changing the transaction structure through this combination process you'd be changing the content that users have signed. Which can't be allowed. In other words, transactions do not have a lawful `Semigroup` instance :wink:  
* No protection against transaction reordering

It might sound like I am bashing this CIP, but I am definitely not. I am pointing out the problems. There are few ideas I really like. For example the idea of having a list of transaction that are individually signed, but then combined together by a third part is pretty cool.

So, I've tried to understand this CIP as well as @polinavino's [Validation Zones | CIP-118](https://github.com/cardano-foundation/CIPs/pull/862) as much as I could. And here is my attempt at an alternative solution: [Transaction Swaps](https://github.com/cardano-foundation/CIPs/pull/880) It is a vastly different solution, but it has the same spirit, so hopefully you'll find it satisfactory.

CIP-????/README.md
```diff
@@ -0,0 +1,530 @@
...
+```cddl
+transaction_piece =
+  { 0: transaction_body
+  , 1: transaction_witness_set ; The witness set for this transaction body.
+  }
+
+transaction =
+  [ transaction_pieces    : [* transaction_piece]
+  , bool
+  , auxiliary_data / null
```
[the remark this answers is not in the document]
fallen-icarus: I thought I mentioned that, but I guess I accidentally deleted that part while editing. Each piece *can* specify it, but if multiple do, they must all agree. I wasn't anticipating users to use pieces for storing metadata on-chain. I thought they would build a personal transaction for that.

CIP-????/README.md
```diff
@@ -0,0 +1,530 @@
...
+   ;    length of block_transactions
+```
+
+### Transaction Size Limit
+
+A transaction broken into pieces should still have approximately the same total size as the whole
+transaction. There are a few things to note:
+
+- If two transaction pieces require the same witness (ie, both pieces require a signature from
+Alice), this witness would now effectively appear twice.
```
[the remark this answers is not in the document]
fallen-icarus: Yes, that is what I meant by "effectively", but I could have been more explicit.

CIP-????/README.md
```diff
@@ -0,0 +1,530 @@
...
+interface to remain mostly untouched. The flow is essentially:
+
+1. Deterministically derive the whole transaction from the list of pieces.
+2. Generate the script context for the whole transaction.
+3. Evaluate the smart contract using this script context.
+
+When combining the transaction pieces, the required executions (ie, script + datum + redeemer) can
+be built up as well. Duplicate executions should be removed. For example, if two transaction pieces
+look to spend the same smart contract input using the same redeemer, there is no reason to evaluate
+this scenario twice since they are guaranteed to have the same result.
```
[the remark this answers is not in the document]
fallen-icarus: > How can you spend the same input twice?

You wouldn't. You would remove duplicates when merging the pieces. If you have a list `[1,2,3,4]` and another `[2,5,6]` merging them (and removing duplicates) would give `[1,2,3,4,5,6]`. So input `2` is *listed* twice, but only *spent* once.

Smart contracts are a definite weak-point of this CIP, but the new Validation Zones revision seems to address this weak-point while still incorporating the main ideas from this CIP. I personally prefer the new Validation Zones CIP over my own for this reason.

CIP-????/README.md
```diff
@@ -0,0 +1,530 @@
...
+However, it should still be fine to keep the current block size limit, at least initially. Just like
+with transactions, the transaction fee would naturally incentivize users to minimize duplications.
+
+### Smart Contract Execution
+
+While the cryptographic keys only need to sign the relevant transaction pieces, the smart contracts
+would be executed against the full transaction. This requirement enables the current smart contract
+interface to remain mostly untouched. The flow is essentially:
+
+1. Deterministically derive the whole transaction from the list of pieces.
```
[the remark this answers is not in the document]
fallen-icarus: > - two transactions that withdraw from the same reward account. It is irrelevant that one of them will be zero, it will change semantics. Moreover if a stake credential is a scritp, then how can you ignore running one of them.

I would argue this transaction should be rejected. I think you are over-complicating things. IMO things should either exact match or be rejected. The only exceptions are when there is a very obvious way to merge them, like with mints and validity intervals. 

> - A DRep casting a yes vote on a proposal in one transaction and then a No on the same proposal in a separate transaction. You might think that we could then ignore the first vote, since the second one replaces the first. But, what if the DRep is backed by a script and it would actually fail if we didn't ignore it.

This transaction would be rejected as well because the proposal vote in one piece does not exact match the proposal vote in the other. The pieces are not applied to the ledger sequentially. They are folded into a single transaction and applied as a unit.

> - Two transactions minting the same asset. How do you run the script twice? You'll be forced to put squash those two mints together into one, which could affect the execution of the script.

The mints would be merged upon folding the pieces and the script would only be executed once. As I said before, smart contracts are a weak-point with this CIP. AFAIU, the new Validation Zone revision addresses this limitation by actually having smart contracts validate against the piece they are used in.

CIP-????/README.md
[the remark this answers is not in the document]
lehins: So it is the "must agree" thing sprinkled throughout the CIP that I really have a problem with. Because of those agreements would have to be enforced by the ledger rules. Which increases complexity of already very complex rules.

CIP-????/README.md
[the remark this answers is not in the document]
lehins: > You would remove duplicates when merging the pieces

So, as I said you can't remove duplicates because of smart contracts.

And with validation zones you can't have duplicates.

fallen-icarus:
@rphair @Ryun1 I would like to withdraw this CIP in favor of [Transaction swaps](https://github.com/cardano-foundation/CIPs/pull/880). @lehins correctly pointed out fatal flaws with this design:
- Potential man in the middle attack - since the list of transaction pieces itself is not signed, it can be reordered by a malicious party. This was an oversight on my part.
- Impossible to pair up pieces using smart contracts - I was not aware that the [redeemer](https://github.com/IntersectMBO/cardano-ledger/blob/47d5cd2795689094886e4b14e26b6509f9508a6d/eras/babbage/impl/cddl-files/babbage.cddl#L330) in the cddl actually includes the execution units. Then, the [`script_data_hash`](https://github.com/IntersectMBO/cardano-ledger/blob/47d5cd2795689094886e4b14e26b6509f9508a6d/eras/babbage/impl/cddl-files/babbage.cddl#L99) in the transaction body is calculated with these budgets. This means it is fundamentally impossible to use a transaction piece with a smart contract since the budgets are based on the whole transaction, and the intent creator must sign this budget. Thus, aggregating the piece with other pieces is guaranteed to invalidate the signature.

In my view, lehins' CIP is what I intended this one to be.

rphair:
@fallen-icarus thank you as always for your brilliant work, your insight, and your humility.  I'll mark this as `Likely Deprecated` and will confirm at next CIP meeting for closure if that still looks appropriate (you can also close it yourself if you wish).

CIP-????/README.md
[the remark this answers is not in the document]
lehins: I am not over-complicating! Those are all of the cases that would have to be addressed in the ledger rules. If we are to reject them, we need to implement specific predicate failures for each one of them that do the rejection. I can see this CIP adding more predicate failures than the total number of predicate failures that we already have today.

> Validation Zone revision addresses this limitation 

I am really confused here. [Validation Zones](https://github.com/cardano-foundation/CIPs/pull/862) is a totally different CIP. Are you trying to combine the two together somehow?

The only way I currently see combining multiple transactions into one for validation to work is if we restrict all those pieces to be **completely** disjoined, i.e. no common credentials.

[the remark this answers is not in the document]
fallen-icarus: > I am not over-complicating!

Sorry, my confusion came from not realizing the redeemers included the execution budgets and therefore you can't do an exact match on the script_data_hash across pieces. I was under-complicating it :sweat_smile:

> I am really confused here. Validation Zones is a totally different CIP. Are you trying to combine the two together somehow?

The new Validation Zone revision (the [README-implicit.md](https://github.com/cardano-foundation/CIPs/blob/1fd1d6dc56fadca8b07009e795b3ec4eedc298cd/CIP-0118/README-implicit.md)) is trying to incorporate the unbalanced transaction idea from this CIP. It uses them in an entirely different way than how this CIP used them.

lehins:
@fallen-icarus I am planning on holding a Ledger working group meeting next Monday. So please join it if you can, because we will be discussing CPS-15 and the suggested approaches at tackling it: https://calendar.google.com/calendar/event?action=TEMPLATE&tmeid=NXBhNmNocnBnaDZkMXJycThnb2pqZTBnaGNfMjAyNDA4MTJUMTcwMDAwWiBhbGV4ZXkua3VsZXNoZXZpY2hAaW9oay5pbw&tmsrc=alexey.kuleshevich%40iohk.io&scp=ALL

Anyone else interested in this topic is free to join.

fallen-icarus:
@lehins That link isn't working for me. It says: "Could not find the requested event."

lehins:
Thank you for pointing it out. I've changed the event's visibility to public

fallen-icarus:
The link still isn't working for me :sweat_smile:  You may need to also make the calendar itself public ([source](https://stackoverflow.com/questions/60683012/google-calendar-could-not-find-the-requested-event)).

lehins:
> You may need to also make the calendar itself public

That sucks. Google could have made this UX a bit better. Unfortunately, I can't make my calendar public, but it should not be a problem. Here is the location and time of the event:

Monday, August 12th 2024 [5pm - 6pm UTC]
Room https://meet.google.com/jer-vaqm-jbu

rphair:
@fallen-icarus @lehins I'm just assigning CIP # 131 to https://github.com/cardano-foundation/CIPs/pull/880 = `Transaction swaps`.  In my gut feeling for future scenarios I can imagine some circumstances in which this PR might be revived if a re-thinking of the security issue leads to a better evolved design... so I'm not giving the old number to the superseding proposal so the two could in fact exist literally side-by-side someday if they both came somehow to be equally considerable.

PR #874, opened 2024-08-05
CPS-???? | Full determinism of transactions is unnecessarily restrictive for DeFi

klntsky:
Determinism of transactions gives rise to valuable ledger properties, such as:

- atomicity (a transaction is either fully accepted or fully rejected without a fee)
- predictability of monetary changes and fees

However, it comes at a cost:

- Development becomes more complicated because of the need to "re-invent" non-determinism where it is required for the dApp logic
- UTxO contention limits the number of concurrent users a dApp can have

Full determinism *may* be unnecessarily strict as a way to achieve the "good" properties, as the design space for the alternatives haven't been fully explored.

[(Rendered version)](https://github.com/klntsky/CIPs/blob/klntsky/determinism-cps/CPS-XXXX/README.md)

### DIFF
--- a/CPS-XXXX/README.md
+++ b/CPS-XXXX/README.md
@@ -36,7 +36,6 @@ However, it comes at a cost: development becomes more complicated because of the
 
 Full determinism *may* be unnecessarily strict as a way to achieve the "good" properties, as the design space for the alternatives haven't been fully explored.
 
-
 <!-- A short (\~200 word) description of the target goals and the technical obstacles to those goals. -->
 
 ## Problem
@@ -54,20 +53,25 @@ Batchers establish ordering and execute concurrent orders from different users. 
 
 This solution introduces a whole set of new problems:
 
-- **Uncertain protocol liveness**. Batchers depend on someone initiating the executing transaction. The node does not provide the guarantee that a valid transaction will pass within a certain timeframe, therefore the risk of order getting stuck has to be accounted for by the user.
+- **Uncertain protocol liveness**. Batchers depend on someone initiating the executing transaction. The node does not provide the guarantee that a valid transaction will pass within a certain timeframe, therefore the risk of order getting stuck has to be accounted for by the user: *no immediate execution is provided*.
 - **Batcher fees and min-ada compensation**. The cost of initiating a transaction has to be compensated to the batcher. The batcher, no matter how efficient, can't take less than min-ada value (the cost of creating a UTxO), because that is the minimum ADA value a UTxO can have. The change can't be returned to the order initiator easily, because it is usually significantly lower than the min-ada.
 - **Doubled action confirmation time**. Two transactions take more time to pass, even assuming immediate order execution.
-
-
+- **Incentives for centralization**. Although decentralized batchers are possible, in reality designing them is more complex than building a centralized batcher bot.
 
 ## Use cases
 <!-- A concrete set of examples written from a user's perspective, describing what and why they are trying to do. When they exist, this section should give a sense of the current alternatives and highlight why they are not suitable. -->
 
-### AMM DEX
+### Automated market maker DEX
+
+AMM DEX requires many users to interact with the same pool, every interaction changes the price available for swap to the next user.
+
+Historically, the improvement of AMM DEXes over traditional order-book-style exchanges has been the ability to execute orders immediately.
+
+Currently, AMM DEXes on Cardano do not allow for high throughput without batchers. Either way, with or without batchers they are effectively no better than order books.
 
 ### Lending platform (liquidations)
 
-
+Liquidating many open positions at once in the presence of a dynamically changing price oracle is not reliable. Lending platform should allow a liquidation to proceed if the price is lower (or higher) than a known threshold, but in reality the need to refer to a particular UTxO containing the price datum makes it so that a liquidation may only proceed if the the price is *the same* as the one known to the liquidation bot during transaction building.
 
 ## Goals
 <!-- A list of goals and non-goals a project is pursuing, ranked by importance. These goals should help understand the design space for the solution and what the underlying project is ultimately trying to achieve.
@@ -77,8 +81,12 @@ Goals may also contain requirements for the project. For example, they may inclu
 Finally, goals may also serve as evaluation metrics to assess how good a proposed solution is. -->
 
 The goal of this CPS is to start a discussion about alternative ledger designs.
+There are no immediate plans to work on implementing any of the possible alternative ledger rules.
 
-There are no immediate plans to work on any of the possible alternative ledger rules.
+### A minimal non-deterministic ledger with fully-atomic transactions
+
+
+
 
 ## Open Questions
 <!-- A set of questions to which any proposed solution should find an answer. Questions should help guide solutions design by highlighting some foreseen vulnerabilities or design flaws. Solutions in the form of CIP should thereby include these questions as part of their 'Rationale' section and provide an argued answer to each. -->
--- a/CPS-XXXX/README.md
+++ b/CPS-XXXX/README.md
@@ -1,0 +1,92 @@
+---
+CPS: ?
+Title: Full determinism of transactions is unnecessarily restrictive for DeFi
+Status: Open
+Category: Ledger
+Authors:
+    - Vladimir Kalnitsky <klntsky@gmail.com>
+Proposed Solutions: []
+Discussions:
+    - https://github.com/cardano-foundation/CIPs/pull/?
+Created: YYYY-MM-DD
+License: CC-BY-4.0
+---
+
+<!-- Existing categories:
+
+- Meta     | For meta-CIPs which typically serves another category or group of categories.
+- Wallets  | For standardisation across wallets (hardware, full-node or light).
+- Tokens   | About tokens (fungible or non-fungible) and minting policies in general.
+- Metadata | For proposals around metadata (on-chain or off-chain).
+- Tools    | A broad category for ecosystem tools not falling into any other category.
+- Plutus   | Changes or additions to Plutus
+- Ledger   | For proposals regarding the Cardano ledger (including Reward Sharing Schemes)
+- Catalyst | For proposals affecting Project Catalyst / the Jörmungandr project
+
+-->
+
+## Abstract
+
+Full determinism of transactions gives rise to valuable ledger properties, such as:
+
+- "true" atomicity (the effects of a valid transaction are known before it is accepted)
+- fee predictability
+
+However, it comes at a cost: development becomes more complicated because of the need to "re-invent" non-determinism where it is required for the dApp logic.
+
+Full determinism *may* be unnecessarily strict as a way to achieve the "good" properties, as the design space for the alternatives haven't been fully explored.
+
+
+<!-- A short (\~200 word) description of the target goals and the technical obstacles to those goals. -->
+
+## Problem
+<!-- A more elaborate description of the problem and its context. This section should explain what motivates the writing of the CPS document. -->
+
+Let's consider two types of determinism:
+
+I. **execution layer determinism**: can we predict the exact outcome of a *transaction*, if it passes? Enforced on Cardano, does not exist on EVM.
+
+II. **application layer determinism**: can we predict the outcome of a *user action* within a dApp, if it succeeds?
+
+Whenever app layer (type-II) non-determinism is desired, the go-to solution for Cardano developers is to split one non-deterministic *app action* into multiple *transactions*, typically via on-chain orders and order batchers.
+
+Batchers establish ordering and execute concurrent orders from different users. Orders are typically implemented as transactions that create UTxOs with inline datums. Batchers would return the funds if it's impossible to execute an order, and the user has to compensate the tx fee for the batcher in advance.
+
+This solution introduces a whole set of new problems:
+
+- **Uncertain protocol liveness**. Batchers depend on someone initiating the executing transaction. The node does not provide the guarantee that a valid transaction will pass within a certain timeframe, therefore the risk of order getting stuck has to be accounted for by the user.
+- **Batcher fees and min-ada compensation**. The cost of initiating a transaction has to be compensated to the batcher. The batcher, no matter how efficient, can't take less than min-ada value (the cost of creating a UTxO), because that is the minimum ADA value a UTxO can have. The change can't be returned to the order initiator easily, because it is usually significantly lower than the min-ada.
+- **Doubled action confirmation time**. Two transactions take more time to pass, even assuming immediate order execution.
+
+
+
+## Use cases
+<!-- A concrete set of examples written from a user's perspective, describing what and why they are trying to do. When they exist, this section should give a sense of the current alternatives and highlight why they are not suitable. -->
+
+### AMM DEX
+
+### Lending platform (liquidations)
+
+
+
+## Goals
+<!-- A list of goals and non-goals a project is pursuing, ranked by importance. These goals should help understand the design space for the solution and what the underlying project is ultimately trying to achieve.
+
+Goals may also contain requirements for the project. For example, they may include anything from a deadline to a budget (in terms of complexity or time) to security concerns.
+
+Finally, goals may also serve as evaluation metrics to assess how good a proposed solution is. -->
+
+The goal of this CPS is to start a discussion about alternative ledger designs.
+
+There are no immediate plans to work on any of the possible alternative ledger rules.
+
+## Open Questions
+<!-- A set of questions to which any proposed solution should find an answer. Questions should help guide solutions design by highlighting some foreseen vulnerabilities or design flaws. Solutions in the form of CIP should thereby include these questions as part of their 'Rationale' section and provide an argued answer to each. -->
+
+
+
+<!-- OPTIONAL SECTIONS: see CIP-9999 > Specification > CPS > Structure table -->
+
+## Copyright
+
+This CPS is licensed under [CC-BY-4.0](https://creativecommons.org/licenses/by/4.0/legalcode).
--- a/CPS-XXXX/README.md
+++ b/CPS-XXXX/README.md
@@ -27,23 +27,23 @@ License: CC-BY-4.0
 
 ## Abstract
 
-Full determinism of transactions gives rise to valuable ledger properties, such as:
+Determinism of transactions gives rise to valuable ledger properties, such as:
 
-- "true" atomicity (the effects of a valid transaction are known before it is accepted)
-- fee predictability
+- atomicity (a transaction is either fully accepted or fully rejected without a fee)
+- predictability of monetary changes and fees
 
-However, it comes at a cost: development becomes more complicated because of the need to "re-invent" non-determinism where it is required for the dApp logic.
+However, it comes at a cost:
+
+- Development becomes more complicated because of the need to "re-invent" non-determinism where it is required for the dApp logic
+- UTxO contention limits the number of concurrent users a dApp can have
 
 Full determinism *may* be unnecessarily strict as a way to achieve the "good" properties, as the design space for the alternatives haven't been fully explored.
 
-<!-- A short (\~200 word) description of the target goals and the technical obstacles to those goals. -->
-
 ## Problem
-<!-- A more elaborate description of the problem and its context. This section should explain what motivates the writing of the CPS document. -->
 
 Let's consider two types of determinism:
 
-I. **execution layer determinism**: can we predict the exact outcome of a *transaction*, if it passes? Enforced on Cardano, does not exist on EVM.
+I. **execution layer determinism**: can we predict the exact result of a *transaction*, if it passes? Enforced on Cardano, does not exist on EVM.
 
 II. **application layer determinism**: can we predict the outcome of a *user action* within a dApp, if it succeeds?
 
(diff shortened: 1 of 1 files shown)


CPS-XXXX/README.md
```diff
@@ -0,0 +1,91 @@
...
+- Wallets  | For standardisation across wallets (hardware, full-node or light).
+- Tokens   | About tokens (fungible or non-fungible) and minting policies in general.
+- Metadata | For proposals around metadata (on-chain or off-chain).
+- Tools    | A broad category for ecosystem tools not falling into any other category.
+- Plutus   | Changes or additions to Plutus
+- Ledger   | For proposals regarding the Cardano ledger (including Reward Sharing Schemes)
+- Catalyst | For proposals affecting Project Catalyst / the Jörmungandr project
+
+-->
+
```
rphair, line 27:
```suggestion
```

CPS-XXXX/README.md
```diff
@@ -0,0 +1,91 @@
+---
+CPS: ?
+Title: Full determinism of transactions is unnecessarily restrictive for DeFi
+Status: Open
+Category: Ledger
+Authors:
+    - Vladimir Kalnitsky <klntsky@gmail.com>
+Proposed Solutions: []
+Discussions:
+    - https://github.com/cardano-foundation/CIPs/pull/?
```
rphair, line 10:
```suggestion
    - https://github.com/cardano-foundation/CIPs/pull/874
```

CPS-XXXX/README.md
```diff
@@ -0,0 +1,91 @@
...
+
+## Goals
+
+The goal of this CPS is to start a discussion about alternative ledger designs.
+There are no immediate plans to work on implementing any of the possible alternative ledger rules.
+
+## Open Questions
+
+<!-- A set of questions to which any proposed solution should find an answer. Questions should help guide solutions design by highlighting some foreseen vulnerabilities or design flaws. Solutions in the form of CIP should thereby include these questions as part of their 'Rationale' section and provide an argued answer to each. -->
+
```
rphair, line 83:
```suggestion
```

CPS-XXXX/README.md
```diff
@@ -0,0 +1,91 @@
...
+- What properties of the ledger that stem from determinism are really valuable?
+- TBD
+
+<!-- OPTIONAL SECTIONS: see CIP-9999 > Specification > CPS > Structure table -->
+
```
rphair, line 88:
```suggestion
```

CPS-XXXX/README.md
```diff
@@ -0,0 +1,91 @@
...
+CPS: ?
+Title: Full determinism of transactions is unnecessarily restrictive for DeFi
+Status: Open
+Category: Ledger
+Authors:
+    - Vladimir Kalnitsky <klntsky@gmail.com>
+Proposed Solutions: []
+Discussions:
+    - https://github.com/cardano-foundation/CIPs/pull/?
+Created: YYYY-MM-DD
```
rphair, line 11:
```suggestion
Created: 2024-08-05
```
(or an accurate composition date)

michele-nuzzi:
I'm sorry if I sound harsh, but I don't believe this CPS should be discussed.

> Development becomes more complicated because of the need to "re-invent" non-determinism where it is required for the dApp logic

I'm not sure I can think of a use case where non-determinism is **required** in principle.

Current designs that fall back to such "reinvented non-determinism" are the result of a previously immature ecosystem (plutus v1 + lack of specific knowledge on eUTxO), having to "borrow" non deterministic designs coming from other ecosystems.

But in principle, non determinism is NOT a requirement.

Yes determinism might need us to think a bit harder because there were no established patterns, but as the ecosystem is evolving we are coming up with more and more eutxo-friendly deterministic patterns.

And in doing so we are preserving a silver bullet in terms of security of the protocols running on this deterministic design.

Security that other ecosystems can not achieve this easily.

> UTxO contention limits the number of concurrent users a dApp can have

Again, this is mostly true only because so far the "established" designs tend to "borrow" designs NOT INTENDED for utxo. Paying the price of applying a synchronous model, where instead we should be thinking in a parallel model.

Many examples of contracts being able to handle concurrent users are present, the most naive I can think of is the NFT marketplace contract.

fallen-icarus (commented):
I have to agree with @michele-nuzzi (although I'm still open to discussing it). 

> I'm not sure I can think of a use case where non-determinism is required in principle.

I've built a DEX, lending/borrowing protocol, options trading protocol, and aftermarket protocol. This is a full DeFi stack, and **not once** did I wish for non-determinism.

I think all of the problems you've mentioned are actually due to most DApps developers still not knowing how to properly use the eUTxO model. This isn't something that can be figured out in a year or two. I literally just opened a [CIP](https://github.com/cardano-foundation/CIPs/pull/873) that shows the eUTxO model actually enables securely breaking apart transactions; doing so would trivially enable babel fees and high-frequency trading. I didn't realize it was possible until only recently. Perhaps this CIP can actually help solve your batcher problems?

Every DApp developer I've seen argue "we need non-determinism" is also doing something that I think could be better done another way. Is the problem that we don't have non-determinism? Or is the problem they are misusing the eUTxO model? My experience and understanding make mean lean very heavily towards the latter.

Personally, I wouldn't consider sacrificing **any** determinism for another few years. I really don't think it has been given a fair chance yet. If we can figure it out, determinism has way better security guarantees than non-determinism.

CPS-XXXX/README.md
```diff
@@ -0,0 +1,91 @@
...
+- **Batcher fees and min-ada compensation**. The cost of initiating a transaction has to be compensated to the batcher. The batcher, no matter how efficient, can't take less than min-ada value (the cost of creating a UTxO), because that is the minimum ADA value a UTxO can have. The change can't be returned to the order initiator easily, because it is usually significantly lower than the min-ada.
+- **Doubled action confirmation time**. Two transactions take more time to pass, even assuming immediate order execution.
+- **Incentives for centralization**. Although decentralized batchers are possible, in reality designing them is more complex than building a centralized batcher bot. Decentralized batchers *still* don't provide liveness guarantees: it *may just happen* that no executor bots will be available at any given time, despite the incentives.
+
+## Use cases
+
+### Automated market maker DEX
```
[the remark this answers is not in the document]
klntsky: > They directly undermine the security assumptions of Proof-of-Stake since users are forced to give up most, if not all, delegation control, voting control, and custody of their assets

It's possible to use own staking credentials with dApp's script as payment credential

> They do not allow users to specify their own prices which makes them extremely economically inefficient

Existing AMM DEXes allow placing orders as a side-feature. An order book existing alongside an AMM is how  [uniswap](https://app.uniswap.org/limit) does that.

> They are fundamentally centralized since updating/maintaining the liquidity pools is controlled by nothing more than a multisig; governance actions are not trustlessly enforced

Any governance scheme can be attached to any protocol.

Liquidity pools without an update mechanism are possible. An order book that lets the admins do something with the orders is also possible.

klntsky:
@fallen-icarus 

> Perhaps this CIP can actually help solve your batcher problems?

That CIP does not address UTxO contention, because even though Txs can be assembled piece-wise independently, the UTxOs they ought to consume in the end can only be consumed by a single transaction. There can only be as many swaps as there are UTxOs, while mutable shared state would allow for as many swaps at the same time as the settlement layer allows

klntsky:
@michele-nuzzi 

> Current designs that fall back to such "reinvented non-determinism" are the result of a previously immature ecosystem (plutus v1 + lack of specific knowledge on eUTxO), having to "borrow" non deterministic designs coming from other ecosystems.

The ledger changes since Plutus v1 did not let us use batchers any less. Do you have a counter-example?

> But in principle, non determinism is NOT a requirement.

Type-II non-determinism is in the product requirements of the core DeFi primitives: AMM DEXes, lending/borrowing, liquidations, etc. These are the cases where the user shouldn't and can't possibly know the outcome of his action within a dApp.

CPS-XXXX/README.md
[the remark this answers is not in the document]
klntsky: Anyway, AMM DEX is just an example here. An order book suffers from the same problem: an order can be matched only by one counter-party.

CPS-XXXX/README.md
```diff
@@ -0,0 +1,160 @@
+---
+CPS: ?
+Title: Full determinism of transactions is unnecessarily restrictive for DeFi
```
rphair, line 3:
```suggestion
Title: Relaxed Transaction Determinism
```
@klntsky @Ryun1 please post if you can think of something equally concise that encompasses the proposed changes without bias.

rphair (commented):
@klntsky I've retitled the PR based on the descriptive language that was used for this proposition in the last hour's meeting.  It's vital that we keep CIP titles not only concise (the original title would have been the longest) but free from bias: especially since the confining effect on Cardano DeFi has not been objectively established yet.

As I said at the meeting I'm happy to see this discussion as a counterpoint to the proposed design patterns that offer transaction flexibility (https://github.com/cardano-foundation/CIPs/pull/873, https://github.com/cardano-foundation/CIPs/pull/862) without sacrificing Cardano's "unique selling proposition" of determinism.  I would recommend promoting this to a CIP candidate if & when a use is documented that cannot be satisfied by a fully deterministic design pattern.

CPS-XXXX/README.md
```diff
@@ -0,0 +1,91 @@
...
+
+Determinism of transactions gives rise to valuable ledger properties, such as:
+
+- atomicity (a transaction is either fully accepted or fully rejected without a fee)
+- predictability of monetary changes and fees
+
+However, it comes at a cost:
+
+- Development becomes more complicated because of the need to "re-invent" non-determinism where it is required for the dApp logic
+- UTxO contention limits the number of concurrent users a dApp can have
```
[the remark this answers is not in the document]
klntsky: > Each seller can have their own UTxO which makes concurrency as high as the underlying market

The problem is that each buyer will try to use the order with the best price available. No matter how many UTxOs you create in an order book, it will be a probabilistic game without batchers, but it doesn't have to be in principle. By parallelizing you can only sidestep the problem, not solve it, which is not really viable, as EVM-based AMMs can just provide better guarantees: they can offer immediate execution no matter the number of users. Aiming at something less than that means losing, because the quality standards are not being set by you

CPS-XXXX/README.md
klntsky: I'll come up with something that mentions *mutable shared state*, because this is the term that was used in the eUTxO paper

klntsky: Renamed

CPS-XXXX/README.md
```diff
@@ -0,0 +1,152 @@
...
+However, it comes at a cost:
+
+- The need to "re-introduce" non-determinism where it is required for the dApp logic (which is commonly done by using multiple transactions for a single user action) makes immediate execution of user actions impossible.
+- UTxO contention limits the number of concurrent users a dApp can have, because a single UTxO can be consumed by a single transaction only.
+
+Completely avoiding mutable shared state *may* be a suboptimal way to achieve the desirable properties, as the design space for alternatives has not been fully explored.
+
+## Problem
+
+It is impossible to build a dApp that has these three properties at the same time on a fully-deterministic ledger:
```
klarkc, line 31: I think the key to scaling solutions on Cardano without compromising its security and determinism lies in building layered solutions. Bridges and other scaling tech can help us batch transactions, reduce load on the main chain, and improve overall performance without giving up on the ledger determinism. In my view this is not a problem, but a feature of Cardano, and we need to find ways and adapt our solutions with that goal.

zliu41:
cc @lehins

colll78:
In general I agree that our ecosystem would benefit greatly from some global state mechanism, after all the whole original pitch of Cardano was to be a hybrid ledger (this is why we already have global state in the form of our staking / rewards system). I just think the suggested approach in this CIP is wrong. Instead, we should just expand the existing account system (reward accounts) with the introduction of accumulators which provide the ability to store and manipulate data encoded types. Then we can simply expand the existing interface that we have for interacting with the account portion of our ledger i.e. we have

`TxCertRegStaking V2.Credential (Haskell.Maybe V2.Lovelace)`

We can add:

`TxCertRegAccumulator V2.ScriptCredential (Haskell.Maybe V2.Lovelace)`

Where `V2.ScriptCredential ` is a script credential derived from the script that manages the accumulator i.e.:

```
-- | Simple accumulator to support merging value to a data encoded integer
accumulatorScript :: BuiltinData -> BuiltinData -> BuiltinData
accumulatorScript acc new = BI.addInteger (BI.mkI acc) (BI.mkI new)
```

Then actions can be processed on the accumulator via:

`TxAccumulatorProcess V2.ScriptCredential BuiltinData`

Where `BuiltinData` is the new value to be processed by the accumulator.

The last piece required is `TxAccumulatorRead V2.ScriptCredential` which simply provides the accumulator value to the script context as an entry in a map (in a new field ex `accumulatorReads :: Map ScriptCredential BuiltinData`). 

I think the above covers everything this CIP seeks to achieve without introducing additional complexity to the UTxO portion of the ledger (we isolate this non-deterministic computation into the already non-deterministic account portion of the ledger). Likewise, we can use the `accumulatorScript` as a phase-1 validation check to preserve the "deterministic evaluation property" for the rest of the ledger. 

The detail is in the implementation. In case of Cardano that state is explicitly passed around from one transaction to another, while Ethereum actually does the global mutation.
In other words this is cardano: 

`foldl xor False [True, False, True, False]`

```
state = False
for v in [True, False, True, False]:
  state = xor v state
end
```

fallen-icarus:
What about how this impacts Ouroboros Leios? According to the original [paper](https://iohk.io/en/research/library/papers/ouroboros-leios-design-goals-and-concepts/), it depends on the assumption that the majority of transactions are independent of each other. 

> For general purpose ledgers where we can assume most in-flight transactions are independent ... With this approach, conflicting transactions will reduce the effective throughput, because they get processed only to be later discarded. However if the rate of conflicts is not too high then this is a reasonable trade-off.
-- Section 3.2.3

It also requires that checking for conflicts must be possible *without* having to run the actual smart contracts:

> Importantly, to preserve parallelism, the detection and discarding of conflicting transactions must be cheap relative to the cost of executing the transactions (e.g. running scripts and checking cryptographic signatures)
-- Section 3.2.3

Section 3.2 of the paper concludes with this:

> Fortunately UTxO-style ledgers are well placed to address these issues. This is due to the fact that transactions in UTxO ledgers explicitly identify all of their inputs and outputs up front, and those dependencies are complete. They are complete in the sense that there are no other ‘side effects’ of the transaction other than the explicitly identified inputs and outputs. This makes it straightforward to identify transactions that conflict with each other. It also means that – provided the dependencies between transactions are respected – the transactions can be reordered without changing the results. This makes it possible to do the serialisation procedure outlined above correctly, and to do so relatively cheaply. 
-- Section 3.2.4

While technically the rewards are a form of global state, they do not create any dependencies between transactions since the only possible thing you can do is withdraw the balance. Currently, smart contracts can't even check the available balance. AFAIU It sounds to me like the changes being proposed would make it possible for a large number of transactions to become dependent on each other in the sense that the order of execution now matters and requires running the smart contracts to determine if the specified order is actually valid. In other words, it would result in the assumptions underlying Leios being consistently violated.

colll78:
@fallen-icarus 
`While technically the rewards are a form of global state, they do not create any dependencies between transactions since the only possible thing you can do is withdraw the balance.`

This isn't quite correct. In addition to withdrawals, registration and deregistration also act on global state, and unlike withdrawals they can indeed create dependencies between transactions. For instance, consider the following:

```
dexOrderCS :: CurrencySymbol
dexOrderCS = "deadbeef"

dexOrderTN :: TokenName 
dexOrderTN = "OrderToken"

-- assume DEX order minting policy enforces that this script executes (i.e. by checking that a redeemer with the matching DepCredential is present in txInfoRedeemers)
globalStateCredential :: ScriptContext -> BuiltinUnit
globalStateCredential ctx = 
  case (scriptContextScriptInfo ctx) of 
    CertifyingScript 0 (TxCertRegDRep _ _) -> check dexOrderTokenMinted 
    TxCertUpdateDRep DRepCredential -> BI.unitval
    CertifyingScript 0 (TxCertRegDRep _ _) -> check dexOrderTokenMinted
    _ -> error 
  where 
    txInfo = scriptContextTxInfo ctx 
    dexOrderTokenMinted = valueOf (txInfoMint txInfo) dexOrderCS dexOrderTN == 1
 
 --- Return the DRepCredential derived from the above script
 globalStateDRepCred ::  DRepCredential
 globalStateDRepCred = ... 
 
-- Minting policy succeeds if and only if total number of DEX orders placed is odd.
readGlobalState :: ScriptContext -> BuiltinUnit 
readGlobalState ctx =
  case (scriptContextScriptInfo ctx) of 
    Minting _ -> 
      if (head certs == TxCertUpdateDRep globalStateDRepCred)
      then BI.unitval 
      else (traceError "Even number of DEX orders") 
    _ -> error 
  where 
    txInfo = scriptContextTxInfo ctx 
    certs = txInfoTxCerts txInfo                            
```

The success or failure of the `readGlobalState` minting policy depends on the order in which transactions are processed in the block. Sure, smart contract execution isn't needed to determine validity, but smart contract execution itself isn't the issue that Leios is concerned with, instead the important part is:
`the detection and discarding of conflicting transactions must be cheap relative to the cost of executing the transactions`
What this means in the context of our architecture is that detection of conflicting transactions must be done cheaply in Phase-1 validation. This property is preserved simply by imposing an extremely restrictive ex-unit budget on `accumulatorScript`s.

fallen-icarus:
@colll78 The even or odd example seems too contrived to be a real use case. Why would a dex care whether the total number of transactions in a block is even or odd? Are there any DApps that actually care about this?

Still, perhaps my wording could have been better since I concede it is *technically* possible to create dependencies using certificates. However, I don't think my conclusion is wrong since doing so doesn't seem useful (isn't the lack of usefulness the point of this CPS?). Even if there are *some* use cases for it, I think the overwhelming majority of transactions will not bother using certificates like this due to it being useless in most contexts. So IMO the assumption that most transactions will *not* depend on each other is still a safe assumption right now.

> What this means in the context of our architecture is that detection of conflicting transactions must be done cheaply in Phase-1 validation. This property is preserved simply by imposing an extremely restrictive ex-unit budget on `accumulatorScript`s.

You definitely know way more about software engineering than myself, but AFAIU I am skeptical restricting the ex-units is enough. Currently, isn't it enough to just check if the input still exists which is 0(1)? There is no need to do anything else aside from lookup the input in the UTxO set. But if you now need to check the smart contract for accumlator's, you also need to deserialize the smart contract before you can even run it. Wasn't this the problem with the reference script ddos you thwarted? I think someone from the consensus team should way in on this, but I'm not sure who to tag.

I don't mean to come across as universally against mutable state; I just want to make sure including it on L1 doesn't sacrifice anything meaningful. For you, the pull to cardano may have been the promise that it would be a "hybrid ledger", but I actually don't know why you think that was promised. The eUTxO paper quoted in this CPS literally says it "forgoes any notion of shared mutable state" due to creating difficulties when trying to scale the blockchain. I came to Cardano precisely because I thought there would be *zero* global state for 1) the scaling issues and 2) I don't think it is necessary for layer 1. Even TradFi has global state on layer 2 (ie, cash+coins are layer 1 and the banking system is layer 2).

colll78:
I agree the example I gave above is indeed contrived, purposely so, since it is learning material that I created to illustrate how the design pattern works. There are indeed practical applications of this design pattern, a single credential is essentially global state over a single bit. Thus the above example manages only a single bit, you can however, extend it to manage multiple account credentials which together form a sequence of bits of global state which can be used for more complicated examples. For instance, typically the maximum signers on a multi-sig in a single block is limited by the constraints of a single transaction. If you use this design pattern, you can create a multi-sig contract that is parameterized by a set of credentials and succeeds only if all credentials are registered, each credential script is parameterized by its own set of signers, then for each credential script registration succeeds if and only if a majority of its signers signatures are present then you have a very large multi-sig that can check for a majority signature of a large set of participants in a single block. You can similarly extend the odd, even example (modeling it as a bit sequence) to keep track of the total number of orders that were placed a the block, which you can then use to for instance adjust a lending rate or adjust protocol fees based on utilization.

Also keep in mind that this isn’t exactly “shared mutable state”, instead that state is explicitly passed around from one transaction to another. 

As for the concerns regarding scripts during phase-1 validation, O(1) just means constant upper bound time-complexity. Both executing a Plutus script and looking up an input in the UTxO set are O(1), since the running time of Plutus scripts are constrained by ex-units. The difference is the size of the constants. We already have a small efficient scripting language with low enough constants to fit in phase-1 validation that is native scripts! Deserialization and serialization are done to improve storage efficiency and preserve network bandwidth, if we impose very strict constraints on size and ex-units of these accumulator scripts I don’t see why they would need a compact representation, the storage cost already takes into consideration size in bytes, instead we can just store them directly in flat UPLC encoding.

fallen-icarus:
> For instance, typically the maximum signers on a multi-sig in a single block is limited by the constraints of a single transaction. ... If you use this design pattern, ... you have a very large multi-sig that can check for a majority signature of a large set of participants in a single block.

Unless you are the stake pool operator that is creating the next block, you don't have control over which transactions will be added to the block. It seems very likely for some transactions to be added to the block while others are omitted which means the multisig will likely fail even if it should succeed. Furthermore, because you don't have control over which transactions are added to the block, this approach is susceptible to a man-in-the-middle attack where the stake pool can deliberately omit certain transactions from the block to control the outcome of the multisig. Unless I am misunderstanding something, I don't see how this approach is practical at all.

Besides, you can *already* get around the constraints of a single transaction by using native assets instead of registration statuses. For example, you can group signatures into sub-transactions and mint a symbolic "Yes" token if the sub-transaction's multisig succeeds (the minting would be governed by a separate script to ensure all sub-transactions can mint the same "Yes" token). Then, you take a top-level transaction that spends each of the UTxOs with the "Yes" tokens from the sub-transactions and burns the "Yes" tokens. If there is a minimum number of "Yes" tokens being burned, the overall multisig would succeed. This approach is very similar to the registration approach except no global state is required at all, and the sub-transactions and top-level transaction can appear in separate blocks. This approach is also easily scalable since native assets are small enough to easily have 10s of thousands of "Yes" tokens in a single transaction (the tokens can be consolidated into a single UTxO before executing the top-level transaction).

> You can similarly extend the odd, even example (modeling it as a bit sequence) to keep track of the total number of orders that were placed a the block, which you can then use to for instance adjust a lending rate or adjust protocol fees based on utilization.

From an economic perspective, I think it a really bad idea to base prices off of the total number of orders processed in a previous block. Prices need to be based off of the *current* ratio of supply and demand. If a block had a sudden burst of 100 orders, using this approach, the prices should increase. But what if there are only 5 orders remaining to be filled (ie, there is very little demand left to fill)? This is actually the time when the prices should *decrease* since the supply now dwarfs the demand. So now you are incentivizing fewer orders at the exact time the DApp actually wants more orders. 

As another example, lets say there is a liquidity source that gets drained by 90% by one order in the previous block. It was only one order so the prices should decrease. But there is only 10% of liquidity remaining which means the prices should actually *increase* to disincentivize new orders and incentivize new liquidity. The total number of orders processed isn't relevant for determining prices. I don't see how this doesn't lead to economic instability and market distortions, so I don't consider this a practical use case either.

rphair:
@klntsky please post something practical here if you want to keep this problem statement under review: and also let us know if you think it can be closed.  We might already have seen all opinions on both sides expressed so far, with the material under review perhaps only serving as a position paper rather than a "Problem Statement" that could lead to CIP solutions.

The last 9 months have also given us more commercial consideration of P2P trading & DeFi arrangements that broaden Cardano's capabilities without relaxing its design principles (cc @fallen-icarus).  If this document is to be continued I hope these will also be brought into both the discussion and the CPS (some of which has already been suggested above).

rphair:
@klntsky closing this as `Abandoned` but we'll refer other discussions back to this discussion thread if & when this premise is presented again through the CIP process.

PR #875, opened 2024-08-06
CIP-0095 | Misspelling of `DeprecatedCertificate`

nilscodes:
Minor wording improvements to the standard committed alongside.

### DIFF
--- a/CIP-0095/README.md
+++ b/CIP-0095/README.md
@@ -40,11 +40,11 @@ For the many contributors to this proposal, see [Acknowledgements](#acknowledgem
 ## Motivation: why is this CIP necessary?
 
 CIP-1694 introduces many new concepts, entities and actors to Cardano;
-describing their implementation at the ledger level. Yet, for most ecosystem
-participants low level details are abstracted away by tooling. Creating a need
-for such tooling, to allow the utilization of new ledger features. This
-specification allows for creation of web-based tools for the utilization of
-CIP-1694's governance features.
+describing their implementation at the ledger level. This creates the need for
+new tooling with respect to governance. For the average ecosystem participant,
+the details should be abstracted away, enabling them to leverage the new ledger
+features more effectively. This specification allows for creation of web-based
+tools for the utilization of CIP-1694's governance features.
 
 Whilst CIP-30 facilitated the launch of dApp development on Cardano, it's
 functionality is limited in scope. It was written well before the emergence of
@@ -119,16 +119,6 @@ extending the API (as a plain integer, without padding). For example:
 
 #### CIP-95 Data Types
 
-#### DRepID
-
-```ts
-type DRepID = string;
-```
-
-A hex-encoded string representing a registered DRep's ID which is a Blake2b-224
-hash digest of a 32 byte Ed25519 public key, as described in
-[CIP-1694 Registered DReps](https://github.com/cardano-foundation/CIPs/blob/430f64d3e86dd67903a6bf1e611c06e5343072f3/CIP-1694/README.md#registered-dreps).
-
 ##### PubDRepKey
 
 ```ts
@@ -137,6 +127,16 @@ type PubDRepKey = string;
 
 A hex-encoded string representing 32 byte Ed25519 DRep public key, as described
 in [CIP-0105 | Conway Era Key Chains for HD Wallets](https://github.com/cardano-foundation/CIPs/blob/master/CIP-0105/README.md).
+
+#### DRepID
+
+```ts
+type DRepID = string;
+```
+
+A hex-encoded string representing a registered DRep's ID which is a Blake2b-224
+hash digest of the above mentioned 32 byte Ed25519 public key, as described in
+[CIP-1694 Registered DReps](https://github.com/cardano-foundation/CIPs/blob/430f64d3e86dd67903a6bf1e611c06e5343072f3/CIP-1694/README.md#registered-dreps).
 
 ##### PubStakeKey
 
--- a/CIP-0095/README.md
+++ b/CIP-0095/README.md
@@ -150,7 +150,7 @@ credential.
 ### Error Types
 
 For the methods described in
-[Governance Extension API](#governance-extension-api), we inherent APIError,
+[Governance Extension API](#governance-extension-api), we inherit APIError,
 DataSignError and TxSignError from
 [CIP-30's Error Types](https://github.com/cardano-foundation/CIPs/tree/master/CIP-0030#error-types).
 
@@ -191,13 +191,13 @@ type APIError {
 
 We repurpose this error type from CIP-30, extending it's functionality. We
 extend the `ProofGeneration` error code to also include cases where DRep secret
-key is not available. We also add one new error code `DepreciatedCertificate`.
+key is not available. We also add one new error code `DeprecatedCertificate`.
 
 ```ts
 TxSignErrorCode = {
   ProofGeneration: 1,
   UserDeclined: 2,
-  DepreciatedCertificate: 3,
+  DeprecatedCertificate: 3,
 };
 type TxSignError = {
   code: TxSignErrorCode;
@@ -209,8 +209,8 @@ type TxSignError = {
   unable to sign the transaction. This is because the wallet does have some of
   the private keys required.
 - `UserDeclined` - User declined to sign the transaction.
-- `DepreciatedCertificate` - Returned regardless of user consent if the
-  transaction contains a depreciated certificate.
+- `DeprecatedCertificate` - Returned regardless of user consent if the
+  transaction contains a deprecated certificate.
 
 #### [DataSignError](https://github.com/cardano-foundation/CIPs/tree/master/CIP-0030#datasignerror)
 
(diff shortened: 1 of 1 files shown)


CIP-0095/README.md
nilscodes, line 1: Lines 42-47 also have a confusing sentence in there, that I'd like to improve on, but I'd need to understand what the intent was.

```
CIP-1694 introduces many new concepts, entities and actors to Cardano;
describing their implementation at the ledger level. Yet, for most ecosystem
participants low level details are abstracted away by tooling. Creating a need
for such tooling, to allow the utilization of new ledger features.
```

The last sentence seems odd and disconnected.

rphair (approved):
It is definitely correct to use the technical term `deprecated` rather than [depreciated](https://en.wiktionary.org/wiki/depreciated) here, but now that this spec has been published for a while we also need to confirm with @Ryun1 that correcting the spelling won't cause compatibility issues.  cc @Crypto2099

Classifying this as an `Update` since we cannot be sure yet that this would not be a behavioural change.

rphair:
@nilscodes I'm backing out my previous ✅ for these changes until we can be sure you are done force-pushing to this branch.  Please post here when you're done making changes and then we'll get all the editors to make an assessment of it.

nilscodes:
Hey Robert - yeah, I had already reached out to @Ryun1 to talk about this. To be honest, wanted to make a draft into my own repo before making the official MR, but GitHub snuck up on me. There will be no further force pushing, and I'll update y'all here on the conversations with Ryan, he said he might reply directly, too.

Ryun1:
Great catches @nilscodes 
thanks for taking the time and creating this PR

I do worry that changing the error codes could break existing compatibility, although more likely to break dApp handelling of the errorcode than wallets
I will reach out to known implementors and make sure that this is okay (tagging a few @Scitz0, @refi93, @mchappell, @mirceahasegan)

nilscodes:
> Great catches @nilscodes thanks for taking the time and creating this PR
> 
> I do worry that changing the error codes could break existing compatibility, although more likely to break dApp handelling of the errorcode than wallets I will reach out to known implementors and make sure that this is okay (tagging a few @Scitz0, @refi93, @mchappell, @mirceahasegan)

Sounds good. I've already reached out to Eternl, since their non english speaking nature may not have alerted them to the misspell. Also, it may be that the implementers used the correct spelling and did not even realize the misspell in the CIP, leading to inconsistent implementations.

I just pushed some more changes, based on the clarity of the DRepID / DRepPublicKey ordering in CIP-0105 to clarify one depends on the other, and proposed a clarification in the first paragraph as indicated by my comment before. Let me know if that works.

Scitz0:
We will/have made the change in error response to `DeprecatedCertificate` in Eternl. We just used error codes as defined in CIP initially even though I fully agree with the change.

Scitz0:
> I just pushed some more changes, based on the clarity of the DRepID / DRepPublicKey ordering in CIP-0105 to clarify one depends on the other, and proposed a clarification in the first paragraph as indicated by my comment before. Let me know if that works.

How does this align with CIP-129 regarding DRep ID definition?

nilscodes:
> CIP-129

Yeah, I have not changed any of the definitions, just reordered them to match CIP-0105 (because the DRepID is derived from the DrepPubKey). And I don't believe CIP-0129 contains any references to the error codes. Thanks for getting back on this so quickly 🙏

~ Nils

Ryun1:
> How does this align with CIP-129 regarding DRep ID definition?
> I guess that's its own CIP and should not impact this one.

Since CIP95 returns just pub keys, without any encoding CIP129 (if adopted) shouldnt impact it

Crypto2099 (approved):
This looks alright to me as written. `Deprecated` is definitely the correct word to use in this instance and it doesn't seem like this would break many implementations and it looks like most of the requisite wallet and tooling creators have already been tagged in and could react to these changes rather easily (it would seem).

mirceahasegan:
> Great catches @nilscodes thanks for taking the time and creating this PR
> 
> I do worry that changing the error codes could break existing compatibility, although more likely to break dApp handelling of the errorcode than wallets I will reach out to known implementors and make sure that this is okay (tagging a few @Scitz0, @refi93, @mchappell, @mirceahasegan)

No issue from Lace side. Thank you for tagging me!

mirceahasegan (approved):
Nice work @nilscodes 👏

Scitz0:
As this one hasn't been merged yet, I propose another change to CIP-95 regarding signData. 

Its currently defined as:

> To construct an address for DRep Key, the client application should construct a [type 6 address](https://github.com/cardano-foundation/CIPs/blob/master/CIP-0019/CIP-0019-cardano-addresses.abnf#L7C8-L7C93). Using an appropriate [Network Tag](https://github.com/cardano-foundation/CIPs/blob/master/CIP-0019/CIP-0019-cardano-addresses.abnf#L13) and a hash of a public DRep Key.

This is NOT how for example Ledger has implemented signData. No address `%b0110 | NETWORK-TAG` is added to the protected header `address` field, just the raw key hash bytes. I think this makes sense. Why try to treat a key, be it pool, drep, cc_hot or cc_cold as an address when its not?

rphair:
that sounds like a good point (https://github.com/cardano-foundation/CIPs/pull/875#issuecomment-2275998500) @Scitz0 - can you submit a PR for it, or open an issue if you think the approach to such a change is debatable or needs clarification / confirmation first?

Scitz0:
> that sounds like a good point ([#875 (comment)](https://github.com/cardano-foundation/CIPs/pull/875#issuecomment-2275998500)) @Scitz0 - can you submit a PR for it, or open an issue if you think the approach to such a change is debatable or needs clarification / confirmation first?

I would, if it weren't for the fact that I leave for a 1½ week vacation in the morning. So if someone else has the time, feel free to do so... else it will have to wait a bit.

nilscodes:
> > that sounds like a good point ([#875 (comment)](https://github.com/cardano-foundation/CIPs/pull/875#issuecomment-2275998500)) @Scitz0 - can you submit a PR for it, or open an issue if you think the approach to such a change is debatable or needs clarification / confirmation first?
> 
> I would, if it weren't for the fact that I leave for a 1½ week vacation in the morning. So if someone else has the time, feel free to do so... else it will have to wait a bit.

I would do it, if I felt confident I knew exactly what to change, but I'm not deep enough in the material unfortunately. Maybe @Ryun1 can help out?

rphair:
@nilscodes it seems proper based on the last https://github.com/cardano-foundation/CIPs/pull/875#issuecomment-2284222377 to tag this `Waiting for Author` (and we do have to keep it tagged _something_) but @Ryun1 we are also waiting for you to advise 100% one way or the other about whether to merge this... avoiding compounded difficulties, which I think are not likely to get better as long as the misspelling persists... or disqualifying the correction in the name of backward compatibility.

nilscodes:
> @nilscodes it seems proper based on the last [#875 (comment)](https://github.com/cardano-foundation/CIPs/pull/875#issuecomment-2284222377) to tag this `Waiting for Author` (and we do have to keep it tagged _something_) but @Ryun1 we are also waiting for you to advise 100% one way or the other about whether to merge this... avoiding compounded difficulties, which I think are not likely to get better as long as the misspelling persists... or disqualifying the correction in the name of backward compatibility.

Since the comment from @Scitz0 was not primarily related to my changes, not sure this MR should wait for it - here we really should finalize if we change the error codes or not. Not a lot of tooling is using any of this yet, most affected would be wallets and potentially MeshSDK.

Scitz0:
> Since the comment from @Scitz0 was not primarily related to my changes, not sure this MR should wait for it - here we really should finalize if we change the error codes or not. Not a lot of tooling is using any of this yet, most affected would be wallets and potentially MeshSDK.

Agree, I created a separate PR for that change now here https://github.com/cardano-foundation/CIPs/pull/897.

rphair (approved):
thanks @nilscodes for that confirmation and @Scitz0 for separating & progressing that other issue.  We should indeed decide on this ASAP and commit to something soon so I've marked it `Last Check` for the CIP meeting in about 1 hour (https://hackmd.io/@cip-editors/96).

cc @Ryun1 (waiting for his approval re: side effects of making the correction)

CIP-0095/README.md
Ryun1: Good add, thanks!

CIP-0095/README.md
```diff
@@ -119,24 +119,24 @@ extending the API (as a plain integer, without padding). For example:
...
+#### DRepID
 
 ```ts
-type PubDRepKey = string;
+type DRepID = string;
 ```
 
-A hex-encoded string representing 32 byte Ed25519 DRep public key, as described
-in [CIP-0105 | Conway Era Key Chains for HD Wallets](https://github.com/cardano-foundation/CIPs/blob/master/CIP-0105/README.md).
+A hex-encoded string representing a registered DRep's ID which is a Blake2b-224
```
Ryun1, line 137: Why are we changing this?

Ryun1 (approved):
Thank you @nilscodes !

CIP-0095/README.md
Ryun1: I must have rewritten that part many times, could never get the right words

PR #880, opened 2024-08-09
CIP-0131? | Transaction swaps

lehins:
This CIP is a solution to [Intents for Cardano | CPS-15](https://github.com/cardano-foundation/CIPs/pull/779)

After reading [Transaction pieces - CIP-0130](https://github.com/cardano-foundation/CIPs/pull/873) and [Validation Zones](https://github.com/cardano-foundation/CIPs/pull/862) I realized there are too many downsides in both of those proposals fro my liking. Almost all of the downsides come out of introduction these strange new relationships between transactions. In order to retain sanity for all of the ledger team developers including myself, I would like to reduce the complexity and propose a different approach that hopefully satisfies all of the needs that other competing proposals would solve.

This is an initial attempt. Before I spend more time on polishing it up, I would like to get some feedback from the community.

---

([rendered proposal](https://github.com/lehins/CIPs/blob/lehins/transaction-swaps/CIP-0131/README.md))


lehins:
CC @polinavino, @fallen-icarus and @WhatisRT

rphair:
thanks; a timely submission @lehins... So we can link to a readable draft, please rename your document to `README.md`.

lehins:
> thanks; a timely submission

There was a lot of discussion on the mentioned CIPs and it was easier for me to write a CIP of my own that describes of how I would solve this problem, rather participate in lengthy discussion on existing CIPs trying to explain my point of view.

> So we can link to a readable draft, please rename your document to README.md.

Done. Sorry, my first CIP here :grin:

rphair:
Since this is not marked as a draft on GitHub, and because it's being referenced in other actively reviewed CIP pull requests, I think it's appropriate to "triage" it at the next CIP meeting even though you are still referring to it as a "draft" - so @lehins I've put it on the agenda () to help bring it up to review speed with other contemporary proposals of similar scope.

lehins:
@rphair I've changed the wording to "attempt", which actually better reflects my intent than a "draft"

fallen-icarus:
I like this approach. It seems simple and straightforward. AFAICT this doesn't support transaction batching, but I think it covers enough use cases to still be worth it.

polinavino:
Interesting idea! I would be very curious to understand the changes to the UTXO rule that this approach requires.

Also, the updated approach to Validation Zones does not require changes to the ledger state because there are no more requests and fulfills. There are also no changes to the transaction structure at all, so it is compatible with existing Plutus contracts (no new version needed). The main change is moving the `consumed = produced` check to the ZONE rule, allowing unbalanced transactions to perform swaps.

https://github.com/cardano-foundation/CIPs/blob/1fd1d6dc56fadca8b07009e795b3ec4eedc298cd/CIP-0118/README-implicit.md

CIP-xxxx/README.md
```diff
@@ -0,0 +1,164 @@
...
+
+Exclusion of all of the irrelevant features of a regular transaction from the swaps will reduce the complexity of implementation and decrease the chance of introducing bugs into the system. This proposal is designed in the type safe spirit of Haskell, because it specifies the feature explicitly, instead of trying to abuse existing ledger rules in order to accommodate desired features. Most importantly, all transactions will continue to be balanced, as crypto gods intended them to be.
+
+
+<!-- The rationale fleshes out the specification by describing what motivated the design and what led to particular design decisions. It should describe alternate designs considered and related work. The rationale should provide evidence of consensus within the community and discuss significant objections or concerns raised during the discussion.
+
+It must also explain how the proposal affects the backward compatibility of existing solutions when applicable. If the proposal responds to a CPS, the 'Rationale' section should explain how it addresses the CPS, and answer any questions that the CPS poses for potential solutions.
+-->
```
Crypto2099, line 146:
```suggestion

```

lehins:
> AFAICT this doesn't support transaction batching

We'd first need to define what transaction batching would actually mean for Cardano. What does it mean in the context of plutus scripts, etc.

This approach allows you to batch multiple swaps, which themselves can be thought of miniture transactions with limited functionality. So in some sense it is batching of transactions.
For example I can create a swap that spends 100ADA and outputs 101ADA. That is a basic transaction, which has 1ADA being reserved for fees.

> I would be very curious to understand the changes to the UTXO rule that this approach requires.

We would need to validate each individual swap (eg. verify signatures, ensure outputs satisfy minutxo requirement, etc.) We would be able to reuse various portions of existing rules to create this rule for swap validation. We would then need to enforce that that there are no duplicate inputs. Then for the purpose of balancing consumed and produced in the UTXO rule  we could pretend that all pieces of swaps are defined in the transaction itself, while enforcing there are no duplicate inputs.

> updated approach to Validation Zones does not require changes to the ledger state

I am not to worried about changing the ledger state, wed do it all the time.

> it is compatible with existing Plutus contracts (no new version needed).

This is not possible. Any existing script that for example expects a fee to never be `0` could now fail, where it wouldn't before. So I would advise against advertising that PlutusV1-PlutusV3 would be able to work with this feature.

polinavino:
> We would need to validate each individual swap (eg. verify signatures, ensure outputs satisfy minutxo requirement, etc.) We would be able to reuse various portions of existing rules to create this rule for swap validation. We would then need to enforce that that there are no duplicate inputs. Then for the purpose of balancing consumed and produced in the UTXO rule we could pretend that all pieces of swaps are defined in the transaction itself, while enforcing there are no duplicate inputs.

As far a rule changes, that is significantly more complicated than the (implicit/updated) validation zones proposal, but could possibly work. I still do not understand if this is built on top of the zones concept or transactions can be validated individually (without the need for zones)?

> This is not possible. Any existing script that for example expects a fee to never be 0 could now fail, where it wouldn't before. So I would advise against advertising that PlutusV1-PlutusV3 would be able to work with this feature.

Fair, a better approach would be - add a new field of type Value to the transaction that is computed at the time of deserialisation - the extra or missing value would be in it, and only let old version scripts run on a transaction if that field is 0.

fallen-icarus:
> We'd first need to define what transaction batching would actually mean for Cardano.

With the way I am thinking about it, and the way @WhatisRT [described](https://github.com/cardano-foundation/CIPs/pull/873#discussion_r1705359703) it, perhaps transaction "batching" is not appropriate. Transaction "chaining" may be more accurate.

Ergo [defines](https://docs.ergoplatform.com/dev/anatomy/transactions/chained/) transaction chaining as: *the sequential use of outputs from off-chain transactions*.

I personally don't have a use case for it, but a [CIP](https://github.com/cardano-foundation/CIPs/pull/508) was opened a while ago about it. I like this simplicity of this CIP so unless there is also a simple extension to enable chaining, I think I'd rather leave it as future work. IMO babel fees and off-chain trading are more critical than chaining.

polinavino:
- without swaps providing collateral, it could be the case that the work a transaction builder does to turn a swap into a transaction (which includes validating possibly very large scripts) is not compensated when scripts phase-2 fail. This is not the case for (implicit) Validation Zones - any phase-1 partially valid transaction can pay collateral

- if a swap includes, for example, a role token, a transaction builder can do whatever they want with it (e.g. interact with another contract that requires the same role token), and the swap-builder, if they want to engage with a contract (e.g. have the transaction builder sponsor the dApp fees for its execution), has no control over this. They dont even sign the final transaction in which their role token will be used. This is, again, not the case for (implicit) Validation Zones. What your transaction does is under your control. One transaction author is not surrendering the use of *all* their tokens for an unknown purpose (as one does within a swap) to another user because these tokens can only be used within a transaction signed by them.

CIP-xxxx/README.md
```diff
@@ -0,0 +1,160 @@
...
+swap_body = { 0 : oset<transaction_input> ;; Regular inputs
+            , 1 : [+ transaction_output]  ;; Outputs, that will have to be satisfied
+            , ? 3 : slot_no               ;; Validity interval start slot number
+            , ? 5 : withdrawals           ;; Regular withdrawals
+            , ? 8 : slot_no               ;; Validity interval end slot number
+            , ? 9 : mint                  ;; Mint field
+            , ? 14 : required_signers     ;; Required signers
+            , ? 15 : network_id           ;; Network Id
+            , ? 22 : positive_coin        ;; Treasury donation
+            }
```
fallen-icarus, line 45: I think this still needs something like the original `script_data_hash` to guarantee the proper `plutus_data` is used as the redeemer for each smart contract. Otherwise, there is a possible man-in-the-middle attack where a malicious party can change the supplied redeemer to one that benefits them. Wouldn't the following work?
```cddl
swap_redeemer = [ tag: swap_redeemer_tag, index: uint, data: plutus_data ] ; no ex_units
swap_redeemer_tag =
    0 ; inputTag "Spend"
  / 1 ; mintTag  "Mint"
  / 3 ; wdrlTag  "Reward"

swap_script_data_hash = $hash32
; The derivation is the same as script_data_hash except it uses 
; swap_redeemer instead of the original redeemer so that the execution units
; can be given later.
```

lehins: > I think this still needs something like the original script_data_hash to guarantee the proper plutus_data is used as the redeemer for each smart contract. 

The transaction swap is signed, so no-one can mess with the datum that will be passed to the swap scripts. Transaction builder will not have the capability to supply plutus datum's, it will only be the execution units that are supplied:

https://github.com/lehins/CIPs/blob/4fb4ba9d2b4a1164742607036aaa5e8566315819/CIP-xxxx/README.md?plain=1#L119-L130

fallen-icarus: Datums and redeemers are separate things. Datums are part of the UTxO set so the input itself has everything needed to verify the proper datum is being used. The same is not true for redeemers. The above `swap_body` does not contain the `plutus_data` (or hash of it) for the redeemers anywhere.

lehins:
@polinavino 

> without swaps providing collateral, it could be the case that the work a transaction builder does to turn a swap into a transaction (which includes validating possibly very large scripts) is not compensated when scripts phase-2 fail. This is not the case for (implicit) Validation Zones - any phase-1 partially valid transaction can pay collateral

The work of a transaction builder happens off chain, which is the work that no-one has to pay for. Even the largest and the most expensive of scripts are really cheap to run locally on a single computer. We only need to worry about cases like that for cardano-node that does validations of transaction. Maybe I misunderstood your concern, so if you can elaborate on it that will be great.

> if a swap includes, for example, a role token, a transaction builder can do whatever they want with it (e.g. interact with another contract that requires the same role token), 

No they can't. This is not a problem for swap transactions at all. When you have a script that locks something in a swap and that swap also contains a role token. That script will only succeed if that role token is present in that exact swap. Tokens from swaps are not placed in some pool that makes all of them equivalent. So, the fact that scripts from other swaps can see such tokens in other swaps, does not mean anything for the safety of the swap. It is up to the script writer to enforce where the role token is expected to be for the script to succeed. So, using such script that expects a token in the same swap will make it fail if it is missing, regardless if that token is present in other swaps in the transaction. You gotta remember that scripts will see the full transaction and they can make any decision the please on that transaction. 

> and the swap-builder, if they want to engage with a contract (e.g. have the transaction builder sponsor the dApp fees for its execution), has no control over this. They dont even sign the final transaction in which their role token will be used. 

There is no expected trust in this model between swap builders and the transaction builder. swap builders have full control of their portion of the transaction through signing the swaps and providing all the necessary data for execution of their scripts. So they *DO NOT NEED* to sign the transaction in which their token is used. In the matter of fact this is a feature that Validation Zones does not have: a script writer can look into other swaps and see who provided such tokens or make their own decision on such information. Hell, they even could even make the transaction to fail if it contains swaps with undesired tokens.

> What your transaction does is under your control. One transaction author is not surrendering the use of all their tokens for an unknown purpose (as one does within a swap) to another user because these tokens can only be used within a transaction signed by them. 

So, this is totally incorrect. No one is surrendering anything. In the matter of fact they are receiving more powers when comparing to Validation Zones: the see the origin of funds and everything that took place for those funds to balance the whole transaction. And they can make any decision they like on that information


> While this could be potentially architected around in new-version Plutus contracts, it seems like it would introduce a lot of additional cases to consider and difficultly in formal verification.

Plutus contracts will have to change, since the context will change. I would like to bring an important argument to this point. Formal verification is just a tool that we use. It should never get in a way of new features. If some feature requires more work on the formal methods side we will just have to add more people to solve it. Vast majority of work always happens on the implementation side and I don't see any blockers here.

CIP-xxxx/README.md
lehins: > Datums and redeemers are separate things. 

Datums was a terrible terminology. The way we use it internally in Ledger (not sure of that is the same outside) is there are spending datums and datums that are supplied in the redeemers. Both are just arguments that are supplied to plutus scripts through plutus context.

You are right. We'll need to move `swap_redeemers` into `swap_body`, so it can be signed. That is because there is no scriptIntegrityHash in the swap body.
```
swap_redeemers = { + [ tag   : swap_redeemer_tag
                     , index : uint .size 4
                     ] => plutus_data
                 }
```

fallen-icarus: > Datums was a terrible terminology. The way we use it internally in Ledger (not sure of that is the same outside) is there are spending datums and datums that are supplied in the redeemers.

It seems everyone uses the terms differently. The redeemer for a smart contract developer doesn't contain execution units, it is just the plutus data. The fact that the same term is used with a different definition by the ledger is really confusing... The terms should be standardized.

> We'll need to move swap_redeemers into swap_body, so it can be signed.

I don't think this is enough. If we have a list of signed redeemers, how do we know they are being paired up with the right scripts? I think we really need a script integrity check for the transaction swaps; this version just wouldn't care about the execution units.

fallen-icarus: The `swap_redeemer_tag` is also relevant for the integrity check since some scripts can be executed as spending, minting, or withdrawals. How do we know the script is being executed as the right type without the script integrity check?

polinavino:
> The work of a transaction builder happens off chain, which is the work that no-one has to pay for. Even the largest and the most expensive of scripts are really cheap to run locally on a single computer. We only need to worry about cases like that for cardano-node that does validations of transaction. Maybe I misunderstood your concern, so if you can elaborate on it that will be great.

The point of the two-phase validation mechanism is to make sure that when a node executes a script from an incoming (phase-1 valid) transaction, and the script fails (this is still off-chain/in the mempool), the node is still able to put that transaction in a block for the purpose of collecting its collateral, thereby getting paid for the work of running of the scripts in that transaction. The scripts are later also run as part of block validation, to make sure that they indeed fail, and collateral should be collected. This feature is a way to prevent DDoS attacks. Because swaps are not transactions, they force the node to run their scripts to determine if they *can* go into blocks, but do not provide any collateral. The two-phase validation guarantee that nodes will be compensated for running mempools is then void. 

Part (2) of my earlier comment is not so much an issue to solve, but rather I am just pointing out that script logic with swaps will have to be significantly more complicated, with more things to consider.

lehins:
@polinavino I know exactly how phase1/2 validation works and how, why and when collateral is collected. You are just missing my point. Transaction builder has to supply the collateral for all of the scripts in all of the swaps.

The person or application building the final transaction that contains all the swaps is ultimately responsible for its validity. They know ahead of time before submitting transaction will it result in a phase2 validationfailure or not. We are not dealing with Etherium here. It is the same as with ValidationZones, the last transaction must provide collateral for all prior transactions in the zone.

CIP-xxxx/README.md
lehins: > If we have a list of signed redeemers, how do we know they are being paired up with the right scripts?

Ledger will take of this part in the same we take of it right now.

The point of script integrity hash is that it takes pieces outside of the transaction body through a cryptographic hash and places it into the body, so it can be signed. In other words it is like placing redeemers, datums and current costmodels directly into the transaction body.

So, if we place redeemers (without execution units) into the swap body, then the user by signing the body prevents anyone messing with the arguments of all of the plutus scripts in the swap. Same applies to the purpose (redeemer tag), since that is the key in the redeemer map.

lehins: We could of course employ the same process and include script integrity hash in the swap transactions. Maybe that would be even better for consistency with regular transactions, albeit at the cost of slightly higher complexity

CIP-xxxx/README.md
```diff
@@ -0,0 +1,198 @@
...
+
+These two proposals have different levels of complexity in respect of implementation, which we are not going to discuss here, since the point of this section is to figure out the difference in features that users of Cardano could evaluate.
+
+
+#### Plutus Context
+
+This proposal has one huge difference from the Validation Zones proposal, namely all of the scripts in a transaction, uncluding the ones in the swaps will see all of the transaction swaps in their context, because they get access to the full transaction. This comes with a benefit of allowing plutus scripts to make decisions on all of the individually unbalanced pieces. That being said it would come at a higher cost for scripts, unless we would also implement [cardano-ledger#3124](https://github.com/IntersectMBO/cardano-ledger/issues/3124), which we have plans on doing anyways. The biggest cost is extra complexity for script writers, since now inputs and withdrawals and minting scritps could now appear in two different places: in regular transactions and in swaps.
```
fallen-icarus, line 155: I think the way the plutus context would work with this CIP needs to be explicitly stated in the specification section. I'm imagining the `TxInfo` would have a new field:
```haskell
data TxInfo = TxInfo
  { ...
  , swapTransactions :: [SwapTxInfo]
  }
```

All of the information related to the transaction swaps would *only* appear in their respective `SwapTxInfo` (eg, the signatures for a swap transaction are only found in that swap transaction's `SwapTxInfo`). 

Smart contracts also need a quick way to know which context is theirs. The `ScriptPurpose` would need to specify which `SwapTxInfo` this execution is for in addition to the rest of the information in the purpose. An index into the `[SwapTxInfo]` would likely be fine.

CIP-xxxx/README.md
```diff
@@ -139,7 +139,45 @@ The major advantage of the approach described in this proposal is in what it tak
...
+
+#### Dependencies of transactions and collateral
+
+Another major difference is that swaps are constructed completely independently and it is only the top level transaction that combines them all together. This allows for an unlimited number of swaps to be constructed concurrently, while Validation zones have inherent dependency in their design: every transaction depends on all of the preceding transactions in the zone. From my understanding this dependency comes from the design of how colateral is specified in the zones.
```
polinavino, line 159: With the new update to Validation Zones (implicit) , this dependency only exists for transactions that specify `requiredTxs`. That is, transactions that are performing exchanges without caring who the counterparty is are now not required to provide collateral for any other transactions, only themselves.

For `requiredTxs`, any DAG of dependencies can be formed. As soon as a failing script is encountered, only the failing transaction + the ones it depends on is entered into the block/zone. For this reason, each transaction is phase-1 checked to cover the size-based fee of transactions it depends on so that they can be included in the zone, and shown to Plutus scripts when needed.

CIP-xxxx/README.md
```diff
@@ -139,7 +139,45 @@ The major advantage of the approach described in this proposal is in what it tak
...
+```
+while this is the dependency of equivalent setup of four transactions in a zone:
+
+```
+tx1 <- tx2 <- tx3 <- tx4
+```
+
+The decision of who pays for the collateral in Validation Zones comes with a natural benefit of deterring users from constructing transactions with phase2 validation, since the first transaction that fails phase2 validation is the one that pays for all scripts in all of the preceding transactions in the zone.
+
+In case of swaps it is up to the transaction builder to figure out which swaps together make up a phase2 valid transaction, because ultimately they will be paying for the collateral if any of the scripts do not succeed. In my personal opinion it is totally reasonable to put this responsibilty on the transaction builder, since ultimately that is the entity that will be making the money in this process.
```
polinavino, line 176: After some consideration, it became clear to me that in Validation Zones implicit, transactions do not need to provide collateral for preceding ones. So, everyone provides their own collateral and it still functions for its intended purpose - to prevent DDoS attacks on nodes by forcing them to run arbitrary amounts of failing scripts.

In VZ design, is possible to allow for the special cases where some nodes would be willing to cover collateral of other users by signing users transactions that include the use of their collateral UTxOs, taking on the risk. This should be the exception and not the rule, however.

CIP-xxxx/README.md
```diff
@@ -139,7 +139,45 @@ The major advantage of the approach described in this proposal is in what it tak
...
+
+#### Full transaction vs a subset of features
+
+Validation Zones allow for full blown transactions that allow usage of features that are not relevant for the goal of solving unbalanced transactions and the feature of Babel Fees, for example voting, proposing, certificates etc. It makes no sense to include them in the swaps, since that would unnecessarily complicate the logic, while in Validation zones it would not make sense to exclude any of them for the same reason. This could be viewed as a benefit or a drawback, depending on one's point of view. One imporant thing to remember when concidering this point is that both of the approaches respet the same transaction size limit.
```
polinavino, line 180: I guess my overall comparison would be something like : 

VZ and Swaps do : 
- reduce duplication of witnesses 
- support exchanges with counterparty irrelevance 
- allow scripts to see other Txs 
- allow others to pay your fees

VZ negatives : 
- need to construct zones
- mempool needs to deal with zones
- more fiddly

VZ positives : 
- does not impose transaction dependencies unless specified by the user
- any DAG dependency graph is allowed rather than *-dependency , which is more versatile 
- collateral mechanism still works for DDoS
- collateral required for only your own scripts (unless dependencies specified) 
- individual collateral slightly cheaper because only the scripts in one transaction with failing scripts ever need to be run/covered by collateral
- Plutus scripts will be slightly cheaper to run, because other transactions' data is only shown to them when required by user

CIP-xxxx/README.md
```diff
@@ -0,0 +1,198 @@
+---
+CIP: ?
+Title: Transaction swaps
+Category: Ledger
+Status: Proposed
+Authors:
+    - Alexey Kuleshevich <alexey.kulehsevich@iohk.io>
+Implementors: []
+Discussions:
+    - https://github.com/cardano-foundation/CIPs/pull/?
```
Ryun1, line 10:
```suggestion
    - https://github.com/cardano-foundation/CIPs/pull/880
```

CIP-xxxx/README.md
```diff
@@ -0,0 +1,198 @@
+---
+CIP: ?
```
rphair, line 2:
```suggestion
CIP: 131
```

rphair (commented):
Thanks @lehins for the groundbreaking work so far on this proposal and vital review on other proposals within the same scope.  CIP meeting today acknowledged this & was keen to assign a number so the community can start referring to the proposal more concretely.

Please also change your directory in this branch to `CIP-0131` and update the link to (rendered proposal) in your OP. 🎉

Quantumplation:
@lehins it would be quite nice to mesh this with CIP-112, and provide a "required observers" field on each transaction swap, which specifies observer scripts that must be executed on the full transaction. That would allow me, for example, to add additional requirements that must be true of the whole transaction in order for the piece I'm signing to be valid.

CIP-xxxx/README.md
colll78: No please, calculating and working with script integrity hash is the worst part of offchain library development on Cardano. Frankly redeemers should have just been directly in the body originally.

lehins:
@rphair I think we can close this PR in favor of #862, since that PR incorporates many of the ideas for this CIP and improves on them with other cool ideas. I am myself now more in favor of CIP-118

rphair:
Thanks @lehins for developing the ideas here & for the update.  Looking forward to seeing you as a regular feature of CIP-0118 discussion and will mark this PR (informally) `Deprecated` in favour of #862.

Issue #793, opened 2024-04-02
Wiki: Initial topics + points to cover


rphair:
This issue is a call for anyone interested in the CIP editing process to suggest topics for a new Wiki introduced at [today's biweekly CIP meeting](https://hackmd.io/@cip-editors/85) to be posted at the standard repository URL: https://github.com/cardano-foundation/CIPs/wiki

- Suggestions can include items of any scale: from documentation sections to mere facts to be included or questions to be answered.
- A number of the items below overlap & it's not a problem to suggest further overlapping items.
- It may become a checklist at some point; depending on how many ideas are contributed.

I'm planning to post a draft of this content this week, so please update this issue ASAP with anything you think should be here.  Here are the main content items repeated from my originally proposed outline (posted [here on the CIP Discord](https://discord.com/channels/971785110770831360/992011119872970762/1224402038159314954) \[[invite](https://discord.gg/3HQayMq9gg) if needed\]):

- Workflow of CIP editors: both regular and irregular components
- Meetings: why we have them, what we do, how to prepare & follow up
- Community engagement: groups and suggestions for interacting with them
- Ethos: what the CIP process does and doesn't do (expanding on this from CIP-0001)
- Reviews: how specifically a GitHub review is processed when preparing to accept a CIP draft... this will also be directly useful to developer and community members engaging with the CIP process, including but not limited to those who may be "training themselves" for an eventual application as a CIP editor.

Much of the above will be re-tasked from my Medium article series beginning here (mainly from \# 1 of these 3 articles), omitting whatever might be considered personal, political, or speculative: https://rxphair.medium.com/cardano-improvement-proposals-cips-introduction-from-an-insider-7b2f7cc94d01

#### suggestions from today's CIP editors' meeting

From @Crypto2099

- What does being a "CIP Editor" mean
- Why do I need a rationale and how do I write one?
- I like Cardanzo \[sic\], how can I help?

From @Ryun1

- why do we seek extensibile standards?
- "your first proposal"
- how to review
- CIP process history / context  

Resolution responding to @rjharmon about target audience, allowing reader to "choose your own adventure":

- top level should have items of general interest & understandable by non-developers.
- navigation & questions should allow people to choose how they want to be involved:
  - as CIP readers?
  - as CIP writers?
  - as potential CIP editors?

My own reflection during the meeting:

- I have a good idea for something Cardano should do, or something I'm doing with it: should I post it as a CIP, a blog article, or a forum thread?

[...]

PR #881, opened 2024-08-14
Add top level CONTRIBUTING file

rphair:
This answers suggestion https://github.com/cardano-foundation/CIPs/issues/793#issuecomment-2033983895 from the initial consideration of the new CIPs Wiki (https://github.com/cardano-foundation/CIPs/wiki).

However, it would be a huge duplication of effort to [build & maintain a CONTRIBUTING.md file](https://mozillascience.github.io/working-open-workshop/contributing/) because it's already accessible & up to date in the Wiki itself.  Still, to carry on the tradition of offline access through these top-level files (e.g. `LICENSE`), I've included brief instructions to build an offline copy of the Wiki with `git clone`.

So I believe the file should just be these 2 pieces of information.  In my opinion it's even too much overhead to use Markdown, and as a text file it follows the convention set by the `LICENSE` file to omit the usual `.txt` extension.

### DIFF
--- a/CONTRIBUTING
+++ b/CONTRIBUTING
@@ -3,4 +3,4 @@ See project Wiki: https://github.com/cardano-foundation/CIPs/wiki
 To view the Wiki offline:
 
 1) git clone https://github.com/cardano-foundation/CIPs.wiki.git
-2) read Home.md first, then remaining pages in lexicographic order.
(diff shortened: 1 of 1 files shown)


Ryun1 (approved):
makes sense!

Crypto2099 (approved):
The rationale makes sense and this looks good

Issue #883, opened 2024-08-19
CIP PR queue overhaul, state vocabulary & tagging


rphair:
As one of the CIP process major tasks we talked about last year, I'm starting to:
1. go through the PR queue, starting with the oldest PR and going forward;
2. closing abandoned proposals (assuming we've given suitable notification in advance);
3. tagging the ones that remain, according to what we appear to be waiting for.

Currently there are 77 open CIP PRs in the queue (minus a couple that are "housekeeping") and it is time we closed off the ones quick are physically abandoned, relying on the submitter to reopen them & identify the pending issues if they wish to pursue that proposal.

As a result we will quickly arrive at a **state tagging vocabulary** that will tell us quickly which sets of PRs need a particular type of action for a proposal to progress (or for a PR to be closed, if that action remains undone with no reason given).

I've yet to start (and some more ideas may emerge for "intermediate" states by the time I get through the big list); here are the states I can imagine, including some we already have.  I'll rename the existing tags so each state tag begins with `State:` and will ensure that all open, non-Draft CIP/CPS PRs have __exactly one tag__ from the time they are first taken on by an editor:
- `State: Triaged` - applied editor actions for a new proposal PR: qualifying a submission, adding readable link, category tagging, adding to agenda, (optional) quick format review
- `State: Unconfirmed` - presented at CIP meeting, but not yet assigned a candidate CIP number (pending further review & update)
- `State: Confirmed` - CIP number has been assigned (usually at CIP meeting)
- `State: Last Check` (renamed from existing tag)
- TBD: any other intermediate states?
- `State: Waiting for Author` (renamed from existing tag)... any of these currently without CIP numbers should be tagged `Unconfirmed`
- `State: Abandoned` ... applies to `Candidate` PRs (if not assigned CIP number yet, they're simply closed if inactive too long)
- `State: Likely Deprecated` (renamed from existing tag) ... applies to `Candidate` PRs (if no CIP number yet, their closure doesn't "deprecate" anything)

This vocabulary will correspond to the existing `Category:` GitHub labels which also represent mutually exclusive categorisation matching the category in the CIP header.  Though these categories were intended to recruit editors and attention from enlisted projects, _the `State:` tags will be **mainly for editors**_ to keep the PR queue trim, readable & functional.

[...]

PR #886, opened 2024-08-24
Update top-level README: post meeting #95

rphair:
Merged:
- https://github.com/cardano-foundation/CIPs/pull/605
- https://github.com/cardano-foundation/CIPs/pull/758

Notes, as per https://github.com/cardano-foundation/CIPs/issues/883#issuecomment-2305540624:
* This doesn't update the dates of the "candidate" tables since they will be replaced with a state query in a forthcoming PR within the hour & I don't want to cause a merge conflict.
* Neither does it add this week's newly assigned CIP number(s)s, since from now on those will be automatically listed by the updated title according to the `State: Confirmed` tag that we apply when promoting at/after the CIP meeting.

### DIFF
--- a/CPS-0008/README.md
+++ b/CPS-0008/README.md
@@ -1,7 +1,7 @@
 ---
 CPS: 8
 Title: Domain Name Resolution
-Status: Proposed
+Status: Open
 Category: Tools
 Authors:
   - Hinson Wong <hinson@cns.space>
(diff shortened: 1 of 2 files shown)


rphair:
Also fixed a thing we missed in review (cc @HinsonSIDAN): the CPS had status `Proposed` instead of the proper initial CPS status `Open`.

Ryun1 (approved):

Crypto2099 (approved):

PR #887, opened 2024-08-24
Change top-level README to use dynamically generated lists

rphair:
Last step in implementing #883.

Before reviewing, please read (and perhaps comment on) description of the PR state tagging vocabulary that was applied through the whole PR queue this week: https://github.com/cardano-foundation/CIPs/wiki/301.-State-tagging

1 - The lists of candidate CIPs and CPSs can no longer be sorted by the candidate CIP number as they were in the manually generated tables.  Interestingly, [GitHub cannot do this](https://github.com/orgs/community/discussions/43151).  This is a toss-up: maybe before it was easier to find a particular proposal & now it's easiest to find the oldest or newest proposals (plus the user can change to a different sort ordering if they want, add more search terms, etc.).

2 - If any wrong items show up on these dynamically generated lists, it means somehow a preposterous set of tags has been applied.  Some care is applied to the queries to prevent this in many cases (i.e. filtering out any "administrative" tagged proposals, even though they generally are not supposed to have a "state" in the review process).

3 - The deleted lists had many likely-abandoned proposals even after I made regular cullings of closed proposals.  The list of "actively reviewed" CIP/CPS PRs is actually quite small & manageable.  Likewise the total PR queue will be rather small once the "Stalled" proposals are either closed _or_ the current & ongoing attention causes the authors to revive and progress them (I will be happy either way).

---

([updated README rendered in branch: changes begin here](https://github.com/rphair/CIPs/tree/replace-tables-with-state-queries#proposals-under-review-cip))

PR #891, opened 2024-08-27
Add Updates section to top level README

rphair:
After consideration in https://github.com/cardano-foundation/CIPs/pull/887#issuecomment-2310470656 — for increased visibility in the community, especially to attract attention from stakeholders & implementors for any affected CIPs.

This uses a low bar for query compared to the _Candidate_ list: only querying the `Update` tag without checking for a review `State:` tag... relying on that one bit of editor confirmation to keep spam or inappropriate update requests from being linked from the front page.

I don't see the harm in including likely "abandoned" or "deprecated" `Update` PR's here... in fact I believe it's important to include them because CIP/CPS updates aren't included in the _Stalled / Waiting For Authors_ list (since the latter is only for new CIPs/CPSs).

---

([newly added section, rendered in branch](https://github.com/rphair/CIPs/blob/readme-add-updates-section/README.md#updates-under-consideration))


Ryun1 (approved):

rphair:
@Crypto2099 this just needs to run through your review (no later than Tuesday's meeting) but I'll change the tag to `Last Check` since there are a bunch of other new CIPs.

Crypto2099 (approved):
LGTM

PR #897, opened 2024-09-03
CIP-0095 | signData correction

Scitz0:
- Clean up duplicate definitions over CIP-30. No need to re-define what is already defined in CIP-30, only how it is changed/how it extends the existing signData endpoint.
- `address` header should be the drep hash and not a type 6 constructed address.

Documentation isn't my best area so if you have suggestions for expressing it differently, feel free to provide suggestions. 😄 

---

([rendered: changes begin here](https://github.com/Scitz0/CIPs/blob/cip-0095-signdata/CIP-0095/README.md#supported-credentials))


gitmachtl:
Considering an Update in the future: Should we change CIP-0095 in a way that the CIP-129 Hash representation is used for the address field in the COSE_Sign1 structure? 

Its ok to use the Hash (which represents the DRep-ID/CC-Hot/CC-Cold) in the address field. However, you can't tell what type it really is. So switching to CIP-129 here would be handy, because it would allow a check on the type of hash/id returned.

However, this is not implemented yet in HW-Wallets or cardano-hw-cli or so...

Scitz0:
> Considering an Update in the future: Should we change CIP-0095 in a way that the CIP-129 Hash representation is used for the address field in the COSE_Sign1 structure?
> 
> Its ok to use the Hash (which represents the DRep-ID/CC-Hot/CC-Cold) in the address field. However, you can't tell what type it really is. So switching to CIP-129 here would be handy, because it would allow a check on the type of hash/id returned.
> 
> However, this is not implemented yet in HW-Wallets or cardano-hw-cli or so...

Though not a bad idea, for now I think we should stick to CIP-5 as CIP-129 haven't been accepted yet (hopefully soon). Or park this one as a dependency on CIP-129 and modify it to reference to that one for address payload. There is also the HW issue you mentioned that will cause issues. CIP-95 also only extend signData to take a DRep ID, not CC.

gitmachtl:
@Scitz0 i mean its all backward compatible, if the address field is only a 56char return, its a simple hash. if its 58char, its a payment or stake address (or goverment ids CIP-129)

Scitz0:
True, this is how we deal with it in Koios and Eternl. However, when verifying a signature through some of the signData libraries out there, it could cause a problem if those libraries dont deal with all formats. And if this specification only mention CIP-129 they might not add the backwards compatible support. I dont see a good way around that though and those lib creators will just have to adapt. 

To be fair, just checked one but CF version seem to expect an address and not just the raw hash as implemented by Ledger and now other, so wont work with just hash as is anyway without a change.
https://github.com/cardano-foundation/cardano-verify-datasignature/blob/main/index.ts

gitmachtl:
As they are using libs from strica (ashish), who are also the CIP-129 proposer, that might be an easy fix. But yea... lets check this out in the upcoming weeks.

gitmachtl:
@ashisherc would it be possible from your side to update the libs to also be cip-129 compatible without interfering with existing usage?

rphair (approved):
This seems sensible & well enough expressed to me (and removes the redundancy with CIP30), though I would leave it to @Ryun1 and/or @Crypto2099 to verify before merge (can briefly discuss this at CIP meeting in 1 hour).

rphair (commented):
This update is `Confirmed` by the last CIP meeting & awaiting further editor reviews & community discussion (thanks @gitmachtl @Scitz0).

CIP-0095/README.md
```diff
@@ -432,67 +432,9 @@ sign a payload conforming to said spec.
...
-| `APIError`      | `Refused`         | Returned if there is a refusal, could be wallet disconnection or extension is revoked.                                         |
-| `APIError`      | `AccountChange`   | Returned if wallet has changed account, meaning connection should be reestablished.                                            |
-| `DataSignError` | `ProofGeneration` | Returned if user has accepted to sign, but wallet could not sign the data; because the wallet does not have the required keys. |
-| `DataSignError` | `AddressNotPK`    | Returned if Address was not a P2PK address and thus had no SK associated with it.                                              |
-| `DataSignError` | `UserDeclined`    | Returned if the user declined to sign the data.                                                                                |
-
-<!-- prettier-ignore-stop -->
+The DRep hash [CIP-5](https://github.com/cardano-foundation/CIPs/blob/master/CIP-0005/README.md#hashes) should be used in the `"address"` field of `COSE_Sign1`'s `Sig_structure` header.
```
Ryun1, line 437: I do worry that this change could break existing implementations 
Could we allow this, in a backwards safe way?

Ryun1:
Since this PR is already open, @Scitz0 do you mind if I push an additional clarification for this CIP, to your branch?
Id like to add a note/warning around the `DRepID` definitions, explaining the CIP105 / CIP129 differences

Scitz0:
@Ryun1 feel free to do so.

CIP-0095/README.md
gitmachtl: This is in line with CIP008, i also asked @SebastienGllmt a long time ago if thats ok and it is. The address field is already the hash of the vkey, and so it is for DReps/CC keys too. I don't see a conflict there. Most of the tools don't even parse this entry, but for those who do it would be good to fill it up with the hash of the vkey. cardano-signer already has this implemented now for a while.

CIP-0095/README.md
```diff
@@ -123,24 +123,23 @@ extending the API (as a plain integer, without padding). For example:
...
 
 ```ts
-type PubDRepKey = string;
+type DRepID = string;
 ```
 
-A hex-encoded string representing 32 byte Ed25519 DRep public key, as described
-in [CIP-0105 | Conway Era Key Chains for HD Wallets](https://github.com/cardano-foundation/CIPs/blob/master/CIP-0105/README.md).
+A hex-encoded string representing a Blake2b-224 hash digest of a 32 byte
+Ed25519 public DRep key.
```
gitmachtl, line 133: shouldn't we make a comment here that this can also be in CIP129 format?

rphair:
was looking through the auto generated lists & this one stuck out as "old but nothing wrong with it"... @Ryun1 looks like you've already tuned this one up so @Crypto2099 @perturbing I'm marking this `Last Check` (cc @Scitz0 @gitmachtl)

**edit:** list of unresolved review points:
https://github.com/cardano-foundation/CIPs/pull/897#discussion_r2134957660

CIP-0095/README.md
ashisherc: This is nothing but breaking standards and introducing a lifelong wart in the ecosystem. This CIP should point to the CIP which defines DRep ID instead of defining drep ID here.

rphair: @gitmachtl @Ryun1 @ashisherc @Scitz0 please let's agree here on a way to do this properly even if it means having to update the text a bit.  No hurry to merge this unless everyone can agree a simple change will make it trouble-free (unless @ashisherc by `this` you mean the proposed change above & not the CIP itself).

Ryun1: I do intend to get back to this one and sort it out, amoung other CIP95 housekeeping

thanks for the nudge and the good points @ashisherc @rphair 🤝

rphair: OK... relaxing status from `Last Check` (a packed section in tomorrow's meeting) & let us know when it's finally ready 😎

rphair:
returning from `Last Check` based on https://github.com/cardano-foundation/CIPs/pull/897#discussion_r2136560300.

rphair:
@Scitz0 @gitmachtl @Ryun1 @ashisherc this has fallen through the cracks apparently (from my perspective) because of a decisive lack of agreement 😅 about whether or not this is a good idea:
* A quick skim through the comments in the unresolved threads above shows claims on either side that this is a confirmed safe, straightforward update _and_ a breaking change.
* There was a [suggestion](https://github.com/cardano-foundation/CIPs/pull/897#discussion_r2136560300) that the general problem could be resolved, but as far as I know that hasn't happened (@Ryun1 perhaps in another update PR?) and there's no indication that the potential issue would be addressed by the author of this update.

So we need some posting(s) in favour of confirming, denying, modifying, or closing this PR:
* Maybe @Crypto2099 would know if the problem has "gone away" somehow.
* I also recall there were some repository issues over `signData` and since we're overhauling the issue queue this week it would be additionally helpful to clarify what to do with this PR.
* In any case marking `Waiting for Author` so, without further comment, it would progress toward closure if we can't document a proper consensus over this update.

PR #915, opened 2024-09-18
CIP-0138? | Plutus Core Builtin Type - Array

ana-pantilie:
We propose an array builtin type for Plutus Core. This type will have constant-time lookup, which is a useful feature that is otherwise not possible to achieve.

([Rendered version](https://github.com/ana-pantilie/CIPs/blob/add-array-builtin/CIP-0138/README.md))


michele-nuzzi:
Happy to see onchain arrays being discussed.

As for the serialization, I think we can re-use the serialization of lists for on-chain constants

It would translate to less effort for independent implementations

ana-pantilie:
@michele-nuzzi 

> I think we can re-use the serialization of lists on-chain constants

I believe the `flat` encoding of arrays should be optimal for this data structure. Reusing the list encoding might not be ideal. Also, AFAIK we always use the standard `flat` encoding for builtins, and we would need a very good reason not to, since it could incur a lot of effort on our part to develop and maintain a custom version of `flat`.

I'm not sure what you mean by "on-chain constants", though.

rphair (commented):
@ana-pantilie this definitely considerable and will be introduced at the next CIP editors' meeting (https://hackmd.io/@cip-editors/98).

I would suggest dropping `New` from the title because it would become inappropriate as soon as this CIP was implemented:

CIP-array-builtin/README.md
```diff
@@ -0,0 +1,296 @@
...
+We can then naturally introduce a function which transforms lists into arrays: `listToArray`.
+
+Since the only current method of constructing arrays is via `listToArray`, one may ask
+whether the `lengthOfArray` builtin is at all necessary.
+We believe that, indeed, `lengthOfArray` is redundant in the current specification. The reason is
+that traversing the list to construct the array is enough to also compute its length. 
+However, we have decided that `lengthOfArray` should be kept as future work will depend upon its
+existence. Some components of Plutus Core could be natively represented as arrays,
+and there would be no other way to access their length in constant time other than
+through a builtin.
```
effectfully, line 219: Doesn't sound very convincing. We know that `dropList` for example is gonna definitely help some and it's still not added. But if `lengthOfArray` is described as "kept as future work", it just doesn't sound like a compelling argument within our current "builtins tradition".

I personally think that `lengthOfArray` is gonna be useful, because people will hate to pass the additional length parameter for array-polymorphic functions or even regular ones, calling `lengthOfArray` is easier than allocating a variable for the length, remembering what it was and referencing at each call site. Since arrays are static in a lot of cases we'll be able to optimize `lengthOfArray` calls away anyway.

CIP-array-builtin/README.md
```diff
@@ -0,0 +1,296 @@
...
+### Slices
+
+Many array-like data structures support cheap _slices_, e.g. a function with the following
+type signature: `slice :: Integer -> Integer -> Array a -> Array a`, which produces a _view_
+of the original array between the two indices (similarly, the same can be achieved using an
+`indexArray` and a `lengthOfArray`).
+
+We do not propose to add `slice` to our builtin set, since it is very easy to build a data
+structure that supports slices on top of `array`, simply by tracking some additional
+integers to track the subset of the array that is in view.
```
effectfully, line 231: Except if we do it as a builtin it'll be much much cheaper compared to implementing that logic in Plutus. Especially given that even just having that logic in the program is already costly in terms of size and that is our scarcest resource, unlike builtins real estate, which we can in theory expand a lot.

I don't mind not adding `sliceArray` right away, but let's make the underlying type `Vector` (which has fast slicing) anyway, since we'll probably end up adding `sliceArray` eventually.

CIP-array-builtin/README.md
```diff
@@ -0,0 +1,296 @@
...
+`Constr` as an `Array` in Plutus Core was presented as being more appropriate than
+the current list representation.
+
+A significant requirement in implementing this modification is maintaining backwards
+compatibility. Therefore, we cannot simply modify the internal representation of `Data`.
+
+One idea would be to add a new builtin such as the following:
+`unConstrDataArray :: Data -> (Integer, Array Data)`. However, this builtin will
+inevitably have linear time complexity since it is based on a list traversal, so
+it does not actually solve the original problem.
```
effectfully, line 245: It does very much solve the original problem! Because the problem is not the algorithmic complexity, it's the constants involved. `Constr` represents the constructor of a data type and chances are, it won't have many arguments, in almost all cases there's below 25 arguments, with the vast majority being below 6 (see [this](https://input-output-rnd.slack.com/archives/G01EE4NQ9U0/p1699520179968649?thread_ts=1699513075.953529&cid=G01EE4NQ9U0) slack discussion where Kenneth analyzed the hell out of it), so in practice it _is_ constant-time. But when handled on the Plutus side the constants are just really bad while as a builtin they would be fairly insignificant.

So we should definitely benchmark `unConstrDataArray`, maybe it'll perform perfectly well. Although maybe we should use the uncurried form: 

```
unConstrDataArray :: forall r. (Integer -> Array Data -> r) -> Data -> r
```

to save a builtin call (or two). Not entirely sure, needs benchmarking as well.

effectfully (commented):
Very well-written!

I'm mostly on board, but I think we need to discuss and experiment more before the design is settled. In particular, see below on `indexOfArray` and the underlying data type being slicing-friendly (all the other comments aren't really important within the scope of this CIP).

CIP-array-builtin/README.md
fallen-icarus: @effectfully The slack link is asking me to sign in with an IOG account (which I don't have). Can you give a public summary of the discussion?

CIP-array-builtin/README.md
```diff
@@ -0,0 +1,296 @@
...
+signature for `indexArray` would become `forall a . Integer -> Array a -> Maybe a` and it
+would return `Nothing` for out-of-bounds lookups.
+    - The obvious disadvantage is the necessity of adding another new builtin type (there is
+    no `Maybe` builtin type in Plutus Core), which would further increase the complexity of
+    the builtin language.
+    - Another disadvantage would be that this solution is the most costly: users will incur
+    additional costs in deconstructing the returned `Maybe`.
+
+2. Failed lookups return a default value provided by the caller:
+`indexArray :: forall a . Integer -> a -> Array a -> a`.
```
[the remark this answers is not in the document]
michaelpj: i.e. "return a scott-encoded Maybe"

Could be fine!

CIP-array-builtin/README.md
```diff
@@ -0,0 +1,296 @@
...
+    the user wants the function to fail. Since Plutus Core is strict, it is not possible
+    to pass `error` as the default value without it getting evaluated before the call and
+    terminating execution immediately.
+
+3. Include `lengthOfArray` and require the user to perform appropriate length guards before
+calling `indexArray`.
+    - This is a familiar option from other languages.
+    - By definition arrays are fixed-size and their length is available in constant time.
+    - It also allows users to omit bounds checks when they know a priori that the
+    index is in bounds.
```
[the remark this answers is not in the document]
michaelpj: Yeah, it lets people omit calls to `lengthOfArray`... but they wouldn't have to do those at all with the other solutions.

What they _would_ potentially have to do is check for the other bad cases, e.g. deconstruct the `Maybe`.  I think the only two that can allow the fastest operation on the PLC side are the one suggested here and your higher-order one.

CIP-array-builtin/README.md
michaelpj: Yeah, and I think the point about receiving arrays of unknown length is pretty fatal to this idea.

CIP-array-builtin/README.md
michaelpj: We can and should benchmark this. I wouldn't expect the overhead in PLC to be that high, since the slicing logic is very simple? But maybe it will be significant.

CIP-array-builtin/README.md
michaelpj: Yes, I think this is plausible. `unConstrDataArray` may well have to be linear in the length of the array, but it might be able to have a pretty gentle slope in comparison to the linear cost of processing a builtin list.

michaelpj (approved):

CIP-array-builtin/README.md
ana-pantilie: I completely removed this section because it's clear we will need to try the "future work" as part of this CIP.

bezirg (approved):

CIP-array-builtin/README.md
```diff
@@ -0,0 +1,294 @@
...
+    index is outside the bounds of the array.
+    - It uses constant time and constant memory.
+2. `lengthOfArray` of type `forall a . Array a -> Integer`
+    - It returns the length of the array.
+    - It uses constant time and constant memory.
+3. `listToArray` of type `forall a . List a -> Array a`
+    - It converts the argument builtin list into a builtin array.
+    - It uses linear time and linear memory.
+
+### Binary serialisation and deserialisation
```
perturbing, line 61: Does this cover [all](https://github.com/cardano-foundation/CIPs/blob/master/CIP-0035/README.md?plain=1#L225-L227) that CIP-035 requires of new types?

ana-pantilie: Thank you for reminding me of this.

I asked @kwxm about this before I opened the PR. It's not very clear to us what the second point there refers to ("A precise description of the measure used for the size of a value of that type.").

If it's about measuring size for costing, I believe his opinion is that given that the two of the operations in the proposal are constant time/space and the third one can be costed in terms of the length of the list there's not much to be said here. Maybe just that the size of an array is equivalent to the size of the list of all its elements?

If that point refers to something else, maybe @michaelpj can clarify.

CIP-array-builtin/README.md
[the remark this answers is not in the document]
effectfully: Scott-encoded `Maybe` would be more like

```haskell
indexArray :: forall a r. Integer -> r -> (a -> r) -> Array a -> r
```

But yeah, it's close enough.

CIP-array-builtin/README.md
effectfully: Well, pattern matching builtins [are](https://github.com/IntersectMBO/plutus/pull/5486#issuecomment-2324804683) 20-30% more efficient than encoding pattern matching using `chooseList`/`chooseData` , so the same logic gotta apply here. Not sure if 20-30% is considered "that high" or not though, but then array slicing is probably gonna have even more overhead. `vector` uses this within a call to [`slice`](https://hackage.haskell.org/package/vector-0.13.1.0/docs/src/Data.Vector.Generic.html#slice):

```haskell
checkSlice :: HasCallStack => Checks -> Int -> Int -> Int -> a -> a
{-# INLINE checkSlice #-}
checkSlice kind i m n x
  = check kind (checkSlice_msg i m n) (i >= 0 && m >= 0 && m <= n - i) x
```

and that's 6 builtin calls right there.

CIP-array-builtin/README.md
effectfully: @fallen-icarus 

> Here's what the constructor tags look like in the redeemers for ~2,600,000 recent transactions (tag on left, frequency on right): 24 seems to be unusually popular for some reason.  I'll add the info for datums (it stlll hurts to type that) and data constants inside scripts later on
>
> ```
>   0 3362344
>   1 356404
>   2 63254
>   3 15324
>   4 27958
>   5 2138
>   6 1650
>   7 123
>   9 27
>  20 851
>  21 4329
>  22 11349
>  23 198
>  24 104886
>  30 42
>  31 58
> ```
>
> Here's the distribution for datums.  No 24s, but some 123s.  Weird.
>
> ```
>   0 10133776
>   1 1021545
>   2 73704
>   3 37826
>   4 46090
>   5 3192
>   6 2381
>   7 123
>   9 1
>  22 88
>  23 52
> 123 40
> ```
>  
> Constructor tag frequencies in data constants in recent validator scripts.
>
> ```
>   0 1321627
>   1 1944159
>   2 2173
>   6 328
> ```
>
> No, I haven't forgotten anything: that's it.

effectfully: > It does very much solve the original problem! Because the problem is not the algorithmic complexity, it's the constants involved. 

I've spent some time thinking about this claim and I think it's not accurate. We do very much detect the performance difference between `chooseList` and `chooseData` and that is reflected in the costs of the builtins despite the only substantial difference between them is `chooseList` doing an extra pattern match and maybe a few other fairly insignificant things.

So either I don't understand why `chooseData` is faster than `chooseList` (which is very possible) or indeed we can detect small changes in performance of the Haskell denotation of a Plutus builtin.

But in any case, the original statement should still be true, the overhead from multiple builtin calls is going to vastly exceed whatever inefficiencies we can have on the Haskell side.

CIP-array-builtin/README.md
ana-pantilie: Then I'll add `slice` to the CIP unless anyone objects

CIP-array-builtin/README.md
Quantumplation: as a historical note, the 123's probably come from bad CBOR encoders / misunderstandings; in CBOR, "constructor 0" is encoded as 121; "constructor 1" is encoded as 122, and "constructor 2" is encoded as 123, etc.

so, someone was likely reading tools like https://cbor.nemo157.com/ and saw the 123, and then passed `123` when they should have passed `2`.

fallen-icarus:
I would personally appreciate it if the outputs field was also an array. But why not make all lists in the script context arrays? Isn't the script context already immutable? Constant lookups could also be useful for reference inputs and even extra signatories/observers. The script redeemers can be used to pass in the necessary indices.

As a slight extension of this topic, should the script purpose be updated to use indices for the input being validated by a spending script instead of the output reference? If the inputs list is changed to an array, wouldn't the index of the input be better?

ana-pantilie:
@Quantumplation @fallen-icarus In order to do that we would need to make changes to `Data`. We can't just replace `List` with `Array` without breaking backwards compatibility. What we can do is add an `Array` constructor to `Data`, but we've not yet fleshed out all the details and what that would entail. Anyway, any such changes will be the subject of another CIP.

CIP-array-builtin/README.md
```diff
@@ -0,0 +1,294 @@
...
+4. From the total set of newly implemented builtins, find a minimal but practical set of
+primitives which are indeed significantly faster in both real-time performance and modelled costs.
+5. If such a set does not exist, find out why. This means that the preliminary investigation
+was not successful. If it does, revise the specification to include
+the final set of primitives.
+
+If the preliminary performance investigation was not successful, this CIP should be revised
+according to the findings of the experiment. Otherwise, the implementation can proceed:
+
+5. Determine the most appropriate costing functions for modelling the builtin's performance
```
Ryun1, line 275:
```suggestion
5. Determine the most appropriate costing functions for modelling the builtin's performance
```
```suggestion
6. Determine the most appropriate costing functions for modelling the builtin's performance
```
list numbering slightly off

CIP-array-builtin/README.md
effectfully: @Quantumplation thank you very much for the explanation.

CIP-0138/README.md
```diff
@@ -0,0 +1,294 @@
...
+2. Failed lookups return a default value provided by the caller:
+`indexArray :: forall a . Integer -> a -> Array a -> a`.
+    - This solution is problematic whenever there is no sensible default value and
+    the user wants the function to fail. Since Plutus Core is strict, it is not possible
+    to pass `error` as the default value without it getting evaluated before the call and
+    terminating execution immediately.
+    - As of the time of writing, builtin functions cannot be higher-order. However, that is
+    subject to change in the near future when pattern matching builtins will be supported by
+    Plutus Core. This feature would allow a safe version of `indexArray` with type `forall a .
+    Integer -> (() -> a) -> Array a -> a` to be expressible in the builtins language.
```
effectfully, line 194: Sorry, my suggestion had a flaw, the ordering should be like this:

```haskell
indexArray :: (() -> a) -> Array a -> Integer -> a
```

This is so that we can

1. give a name to `indexArray (\_ -> error)` (e.g. `unsafeIndexArray`) and recover the failing version of `indexArray` that way without pretty much any performance overhead (that application will be computed once per execution)
2. give a name to `unsafeIndexArray arr` for a specific `arr`, so that this application also needs to be computed only once

I know it's weird to make `Integer` appear last (although that's kinda like the infix `(!)`), but I think efficiency concerns outweigh beauty ones here, plus compiler authors are free to provide versions of the builtin with arguments flipped.

Oh yeah, and pattern matching builtins have landed, so `indexArray` taking a function should be definable now.

CIP-array-builtin/README.md
```diff
@@ -0,0 +1,296 @@
...
+structures.
+
+Access to an array type would provide significant performance improvement opportunities to
+users of Plutus Core, since currently they must rely on suboptimal data structures such as
+lists for looking up elements inside a collection.
+
+## Specification
+
+We add the following builtin type: `Array` of kind `Type -> Type` representing
+a one-dimensional array type with indices of type `Integer`. Elements are indexed
```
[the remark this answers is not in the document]
effectfully: @rphair please resolve this comment.

CIP-array-builtin/README.md
```diff
@@ -0,0 +1,296 @@
...
+
+
+## Rationale: how does this CIP achieve its goals?
+
+### Choice of builtin functions and their specification
+
+The following section presents the reasoning behind the above specification of a Plutus
+core builtin array type.
+It is important to mention that we based our decisions on the desire to keep the builtin
+language as small as possible, i.e. to not introduce types or functions
```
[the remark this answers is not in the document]
effectfully: @rphair please resolve this comment.

CIP-array-builtin/README.md
```diff
@@ -0,0 +1,296 @@
...
+    index is outside the bounds of the array.
+    - It uses constant time and constant memory.
+2. `lengthOfArray` of type `forall a . Array a -> Integer`
+    - It returns the length of the array.
+    - It uses constant time and constant memory.
+3. `listToArray` of type `forall a . List a -> Array a`
+    - It converts the argument builtin list into a builtin array.
+    - It uses linear time and linear memory.
+
+### Binary serialisation and deserialisation
```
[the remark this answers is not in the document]
effectfully: @rphair please resolve this comment.

CIP-array-builtin/README.md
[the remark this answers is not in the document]
effectfully: @rphair please resolve this comment.

CIP-array-builtin/README.md
```diff
@@ -0,0 +1,296 @@
...
+`indexArray :: forall a . Integer -> a -> Array a -> a`.
+    - This solution is problematic whenever there is no sensible default value and
+    the user wants the function to fail. Since Plutus Core is strict, it is not possible
+    to pass `error` as the default value without it getting evaluated before the call and
+    terminating execution immediately.
+
+3. Include `lengthOfArray` and require the user to perform appropriate length guards before
+calling `indexArray`.
+    - This is a familiar option from other languages.
+    - By definition arrays are fixed-size and their length is available in constant time.
```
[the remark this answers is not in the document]
effectfully: @rphair please resolve this comment.

CIP-array-builtin/README.md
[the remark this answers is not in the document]
effectfully: @rphair please resolve this comment.

> And that is going to be extremely cheap compared to the overall cost of a builtin call.

Actually, I may be wrong about that. But it's not important.

CIP-array-builtin/README.md
```diff
@@ -0,0 +1,296 @@
...
+    - It also allows users to omit bounds checks when they know a priori that the
+    index is in bounds.
+
+After considering these three options, we concluded that the addition of
+the `lengthOfArray` builtin function is both necessary and sufficient for introducing a
+well-defined array type into the builtin language.
```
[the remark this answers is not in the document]
effectfully: @rphair please resolve this comment.

CIP-array-builtin/README.md
effectfully: @rphair please resolve this comment.

CIP-0138/README.md
```diff
@@ -0,0 +1,294 @@
...
+
+A more appropriate approach would be to construct the array in bulk,
+however that would require an intermediate representation of a collection of elements. Fortunately this already exists in the builtin language in the form of builtin lists.
+We can then naturally introduce a function which transforms lists into arrays: `listToArray`.
+
+
+### Slices
+
+Many array-like data structures support cheap _slices_, e.g. a function with the following
+type signature: `slice :: Integer -> Integer -> Array a -> Array a`, which produces a _view_
```
effectfully, line 224: Same point about the ordering of arguments.

CIP-array-builtin/README.md
effectfully: > Then I'll add `slice` to the CIP unless anyone objects

Have you done that? Not sure if something's changed or not.

I don't mind not including `sliceArray` right away, however I really want the underlying array representation to be `sliceArray`-friendly, i.e. I want `Vector`.

CIP-array-builtin/README.md
effectfully: @rphair please resolve this comment.

effectfully (approved):
Just a couple of nitpicks.

I'm OK with merging this as is, we can always bikeshed ordering of arguments elsewhere.

rphair (approved):
thanks @effectfully for collating the prior comments & marking the resolved ones.  If any of the remaining points raised (including the @michaelpj review) should block the merge (cc @ana-pantilie) please flag them so the editors can spot before merge.  Otherwise I've marked `Last Check` so we will likely merge around the next CIP meeting: https://hackmd.io/@cip-editors/100

perturbing (approved):

Ryun1 (approved):

colll78:
A huge number of use cases that would benefit from `BuiltinArray` involve selecting a list of elements based on an index list passed via the redeemer (often to select relevant inputs / outputs / withdrawals from the transaction to validate). Thus, I'd like to propose that we modify this CIP to introduce:
```haskell
mapIndexArray :: forall a . Integer -> BuiltinList Integer -> Array a -> a
mapIndexArray indices arrayList = ...
```
Example:
```
mapIndexArray [0, 3, 1] [a, b, c, d, e] == [a, d, b]
```

For reference, 
SundaeSwap @Quantumplation
Minswap @ljttl3q04t
Spectrum Finance
WingRiders @TechRiderWR
and Djed 
All of the above use some variation of the [redeemer indexing design pattern](https://github.com/Anastasia-Labs/design-patterns/blob/main/utxo-indexers/UTXO-INDEXERS.md#utxo-indices-in-redeemer) to reorder the inputs, and more generally to obtain certain inputs / elements in lists without needing to perform linear search. So `mapIndexArray` would be extremely relevant.  More generally, when you have certain criteria that you need to be satisfied by various elements throughout the script context, `mapIndexArray` would often trivialize the process and make it incredibly efficient.

rphair:
@ana-pantilie what do you think of https://github.com/cardano-foundation/CIPs/pull/915#issuecomment-2541805707 ... did you consider this possibility, or would you like to be involved in extending the definition as such? - cc @effectfully

colll78:
> Sorry @rphair @colll78 , I am only just seeing this notification.
> 
> 
> 
> @colll78 What you're suggesting sounds like you would have preferred a generalization of the `indexArray` builtin: 
> 
> ```.haskell
> 
> indexArray :: forall a . Array a -> BuiltinList Integer -> ??? a
> 
> ```
> 
> 
> 
> The reason I put `???` in the return type is because I'm not sure if `BuiltinList` or `Array` would be more suitable. However, after thinking about this from a language design perspective, `Array` is much more ergonomic, but we would have to add `arrayToList` as well. 
> 
> 
> Why not just return a `BuiltinList`? I think that if we are going to add slicing in the future, we could use the generalized `indexArray` to implement it, and the result of a slice is always an `Array`. Even if we don't want to use `indexArray` to implement slicing, I believe it would be more ergonomic for `indexArray` and slicing to compose, so `indexArray` should return an array.
> 

Yes agreed, the multiIndexArray function should return a BuiltinList since the whole point of the function is to extract a subset of elements in the array and return them in a given ordering as described by the index list (so they can be traversed linearly with low cost constants i.e. head list is much cheaper than indexArray 0 list and maintaining an index pointer).

> 
> So, we could modify `indexArray` to be more general and add `arrayToList`. We could implement the simple (original) indexing using the two, but if we think that will be a hit on performance, then we can add:
> 
> - `multiIndexArray :: forall a . Array a -> BuiltinList Integer -> Array a` which returns an array which contains a subset of the argument array (it is empty if none of the supplied indexes exist in the supplied array)
> 

I believe it's important to return a list because given result is sorted, the intended purpose is linear traversal which should be much cheaper on a list. I agree it should be empty if none of the supplied indices exist in the supplied array.

> 
> The above is mostly brainstorming, so I haven't thought too deep about any technical issues. Tagging @effectfully in case he spots anything.
> 
> 
> 
>

effectfully:
Adding

```haskell
multiIndexArray :: forall a . Array a -> BuiltinList Integer -> BuiltinList a
```

does make sense to me, but note that it has the overhead of requiring the input list of indices to be constructed and the output list of elements to be deconstructed. So if the the former is known statically, then the latter is basically wasted work as you could instead directly bind all elements to variables (that'd involve O(n) builtin calls, but so is deconstructing the output list, except in the latter case the builtin calls will likely be more expensive).

But if the input list isn't known, then at least you're saving on recursion costs, although if we speed up recursion, maybe it won't be a concern anymore.

Anyhow, I'm fine with adding such a builtin.

colll78:
Th

> Adding
> 
> ```haskell
> multiIndexArray :: forall a . Array a -> BuiltinList Integer -> BuiltinList a
> ```
> 
> does make sense to me, but note that it has the overhead of requiring the input list of indices to be constructed and the output list of elements to be deconstructed. So if the the former is known statically, then the latter is basically wasted work as you could instead directly bind all elements to variables (that'd involve O(n) builtin calls, but so is deconstructing the output list, except in the latter case the builtin calls will likely be more expensive).
> 
> But if the input list isn't known, then at least you're saving on recursion costs, although if we speed up recursion, maybe it won't be a concern anymore.
> 
> Anyhow, I'm fine with adding such a builtin.

The input list of indices never needs to be constructed because script inputs to Plutus are known during transaction construction (inputs are deterministic), thus for any subset of elements that you want to find, you already know the indices of those elements in the structures that contain them (ie lists or maps in the script context) when you are constructing the transaction, and thus you can construct the index list offchain and include it in the transaction redeemer. It isn't the recursion itself that is expensive, it is all the auxiliary information you need to compute non-natively to actually perform the recursion. The reason this will be so much more efficient is the same reason that `dropList` is so much more efficient than a function that recursively applies tail `n` times.

effectfully:
> The input list of indices never needs to be constructed because script inputs to Plutus are known during transaction construction (inputs are deterministic), thus for any subset of elements that you want to find, you already know the indices of those elements in the structures that contain them (ie lists or maps in the script context) when you are constructing the transaction, and thus you can construct the index list offchain and include it in the transaction redeemer. 

Yes, but what I'm saying is that you'll then need to deconstruct the resulting list to extract elements from it one by one and that is O(n) builtin calls as well, and likely more expensive ones than array lookups.

Or will it be common not to deconstruct the result of a `multiIndexArray` call?

PR #921, opened 2024-10-03
CIP-0139? | Plutus Core Builtin Type - ByteStringMap

ana-pantilie:
We propose a `ByteStringMap` type for Plutus Core. This type will have logarithmic time lookups, which are difficult to achieve in Plutus Core programs today.

([Rendered Version](https://github.com/ana-pantilie/CIPs/blob/bytestring-map-builtin/CIP-0139/README.md))

### DIFF
--- a/CIP-0139/README.md
+++ b/CIP-0139/README.md
@@ -1,5 +1,5 @@
 ---
-CIP: ????
+CIP: 139
 Title: Plutus Core Builtin Type - `ByteStringMap`
 Category: Plutus
 Status: Proposed
--- a/CIP-bytestring-map-builtin/README.md
+++ b/CIP-bytestring-map-builtin/README.md
@@ -9,7 +9,7 @@ Authors:
 Implementors: []
 Discussions:
     - https://github.com/cardano-foundation/CIPs/pull/921
-Created: YYYY-MM-DD
+Created: 2024-MM-DD
 License: CC-BY-4.0
 ---
 
--- a/CIP-bytestring-map-builtin/README.md
+++ b/CIP-bytestring-map-builtin/README.md
@@ -9,7 +9,7 @@ Authors:
 Implementors: []
 Discussions:
     - https://github.com/cardano-foundation/CIPs/pull/921
-Created: 2024-MM-DD
+Created: 2024-10-03
 License: CC-BY-4.0
 ---
 
--- a/CIP-bytestring-map-builtin/README.md
+++ b/CIP-bytestring-map-builtin/README.md
@@ -42,7 +42,7 @@ Unfortunately, the builtin language has a significant limitation: it does not al
 
 Why are higher-order builtin functions necessary? Map keys must be processed (or simply, just compared) in order to be able to build and operate over well-defined maps. For arbitrary key types, that would only be possible by passing the required processing or comparison functions as arguments to the builtin functions.
 
-The only way to implement a builtin map with a (builtin) polymorphic key type would be to modify the builtins language itself. A compiler would have to only allow the construction of maps with key types which are known a priori to posess some property (such as a total ordering). The Plutus Core evaluator would have to check for this as well. The implementation itself would not be too complicated, but we must first consider how this would change the formal semantics of the builtins language.
+The only way to implement a builtin map with a (builtin) polymorphic key type would be to modify the builtins language itself. A compiler would have to only allow the construction of maps with key types which are known a priori to possess some property (such as a total ordering). The Plutus Core evaluator would have to check for this as well. The implementation itself would not be too complicated, but we must first consider how this would change the formal semantics of the builtins language.
 
 Given the above, the only immediate solution is to implement a builtin map type which is specialised to `ByteString` keys but remains polymorphic in the value type. We believe that this solution should be good enough for most practical use-cases, and maybe even surpass a fully polymorphic map type in performance.
 
--- a/CIP-bytestring-map-builtin/README.md
+++ b/CIP-bytestring-map-builtin/README.md
@@ -61,7 +61,7 @@ We propose the following set of builtin functions to accompany the new builtin t
 1. `insertBSMap :: forall v . ByteString -> v -> ByteStringMap v -> ByteStringMap v`
     - it returns a map with the input key updated to the new value, silently discarding any previous value
     - it uses logarithmic time in the `ByteStringMap` argument, linear time in the `ByteString` argument, and logarithmic memory in the `ByteStringMap` argument
-2. `lookupBSMap :: forall a v . ByteString -> (() -> a) -> ByteStringMap v -> a`
+2. `lookupBSMap :: forall v . ByteString -> (() -> v) -> ByteStringMap v -> v`
     - it returns the value for a given key in the map; if the key is missing, it returns the default value of type `() -> a` 
     - its implementation depends on, as of the time of writing, ongoing work in supporting higher-order builtin functions; see the next section for the motivation behind this choice
     - it uses logarithmic time in the `ByteStringMap` argument, linear time in the `ByteString` argument and constant memory
@@ -99,7 +99,7 @@ Map lookups are essential, so we must provide `lookupBSMap`, but there remains t
 
 As mentioned earlier, we have decided to base the implementation of this builtin on an upcoming feature, because it is the least invasive way of providing a total implementation and it leverages future improvements to the builtins language.
 
-With the proposed signature of `lookupBSMap`, users will be able to provide any default value for when the lookup fails, including `error` or `False`. Therefore, `lookupBSMap` subsumes a membership check operation as well.
+With the proposed signature of `lookupBSMap`, users will be able to provide any default value for when the lookup fails, including `error`.
 
 ### Traversing `ByteStringMap`s
 
--- a/CIP-bytestring-map-builtin/README.md
+++ b/CIP-bytestring-map-builtin/README.md
@@ -1,0 +1,152 @@
+---
+CIP: ????
+Title: Plutus Core Builtin Type - `ByteStringMap`
+Category: Plutus
+Status: Proposed
+Authors:
+    - Michael Peyton Jones <michael.peyton-jones@iohk.io>
+    - Ana Pantilie <ana.pantilie@iohk.io>
+Implementors: []
+Discussions:
+    - https://github.com/cardano-foundation/CIPs/pull/?
+Created: YYYY-MM-DD
+License: CC-BY-4.0
+---
+
+## Abstract
+
+We propose a `ByteStringMap` type for Plutus Core. This type will have logarithmic time lookups, which are difficult to achieve in Plutus Core programs today.
+
+
+## Motivation: why is this CIP necessary?
+
+### Map data structures
+
+A _map_ is a collection of _values_ indexed by _keys_. We consider a map to be _well-defined_ if each key uniquely identifies each value. A basic map implementation provides, at the very least, a way to define an empty map, an operation to insert new key-value associations into an existing map, an operation to lookup values based on an input key and an operation to delete entries from the map.
+
+### Performance and Plutus Core implementations
+
+To provide map operations with good asymptotic performance, it is necessary to implement a fairly complex data structure such as a red-black tree. Implementing such data structures in Plutus Core programs themselves has been attempted multiple times in the past, but unfortunately we have found that the high overhead and large memory cost outweigh any theoretical algorithmic improvements (see [attempt 1][3] and [attempt 2][2]).
+
+The only practical way to implement maps in Plutus Core is by using an association list representation. Lists are known to have poor asymptotic performance for usual map operations. To make matters worse, maps are in many cases central to developing smart contract logic, specifically maps with `ByteString` keys. See [CPS-0013][1] for more details on how smart contracts heavily depend on `ByteString` maps.
+
+Another attempt to mitigate the performance restrictions imposed by the list implementation was to specialise the Plutus Core implementation [into an association list with `ByteString` keys][4]. Unfortunately, this approach also failed.
+
+We concluded that embedding maps into Plutus Core cannot be done without incurring high performance costs.
+
+### Map as a builtin type
+
+The only remaining solution is to consider introducing a builtin map type.
+
+Unfortunately, the builtin language has a significant limitation: it does not allow for higher-order builtin functions. As of the time of writing, builtin functions may not receive other functions as arguments, but there is ongoing work to lift this restriction. However, due to complex design considerations, repeated applications of function arguments will incur high performance costs. Since maps are recursive data structures, these high performance costs will be accumulated and such an approach will still not be feasible.
+
+Why are higher-order builtin functions necessary? Map keys must be processed (or simply, just compared) in order to be able to build and operate over well-defined maps. For arbitrary key types, that would only be possible by passing the required processing or comparison functions as arguments to the builtin functions.
+
+The only way to implement a builtin map with a (builtin) polymorphic key type would be to modify the builtins language itself. A compiler would have to only allow the construction of maps with key types which are known a priori to posess some property (such as a total ordering). The Plutus Core evaluator would have to check for this as well. The implementation itself would not be too complicated, but we must first consider how this would change the formal semantics of the builtins language.
+
+Given the above, the only immediate solution is to implement a builtin map type which is specialised to `ByteString` keys but remains polymorphic in the value type. We believe that this solution should be good enough for most practical use-cases, and maybe even surpass a fully polymorphic map type in performance.
+
+
+## Specification
+
+We propose the addition of the following builtin type: `ByteStringMap` of kind `Type -> Type`.
+- It represents a map associating keys of type `ByteString` with values of any builtin type.
+- The new builtin type should be implemented using a tree data structure allowing for logarithmic-time map operations.
+- We consider the costing size of a `ByteStringMap` to be equal to the size of its corresponding representation as an association list.
+- The binary encoding of `ByteStringMap`s is equivalent to its encoding as an association list.
+
+**Note**: Here `Type` is the universe of all builtin types, since we do not consider types formed out of applying builtin types to arbitrary types to be inhabited.
+
+We propose the following set of builtin functions to accompany the new builtin type:
+1. `insertBSMap :: forall v . ByteString -> v -> ByteStringMap v -> ByteStringMap v`
+    - it returns a map with the input key updated to the new value, silently discarding any previous value
+    - it uses logarithmic time in the `ByteStringMap` argument, linear time in the `ByteString` argument, and logarithmic memory in the `ByteStringMap` argument
+2. `lookupBSMap :: forall a v . ByteString -> (() -> a) -> ByteStringMap v -> a`
+    - it returns the value for a given key in the map; if the key is missing, it returns the default value of type `() -> a` 
+    - its implementation depends on, as of the time of writing, ongoing work in supporting higher-order builtin functions; see the next section for the motivation behind this choice
+    - it uses logarithmic time in the `ByteStringMap` argument, linear time in the `ByteString` argument and constant memory
+3. `deleteBSMap :: forall v . ByteString -> ByteStringMap v -> ByteStringMap v`
+    - it returns a map without the key-value association identified by the input key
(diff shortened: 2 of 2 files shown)


CIP-bytestring-map-builtin/README.md
```diff
@@ -0,0 +1,152 @@
...
+- We consider the costing size of a `ByteStringMap` to be equal to the size of its corresponding representation as an association list.
+- The binary encoding of `ByteStringMap`s is equivalent to its encoding as an association list.
+
+**Note**: Here `Type` is the universe of all builtin types, since we do not consider types formed out of applying builtin types to arbitrary types to be inhabited.
+
+We propose the following set of builtin functions to accompany the new builtin type:
+1. `insertBSMap :: forall v . ByteString -> v -> ByteStringMap v -> ByteStringMap v`
+    - it returns a map with the input key updated to the new value, silently discarding any previous value
+    - it uses logarithmic time in the `ByteStringMap` argument, linear time in the `ByteString` argument, and logarithmic memory in the `ByteStringMap` argument
+2. `lookupBSMap :: forall a v . ByteString -> (() -> a) -> ByteStringMap v -> a`
```
ana-pantilie, line 64: @effectfully - please confirm that we could actually implement this. If not, we can just revert the `a` to be `v` and add a `member` builtin function.

CIP-bytestring-map-builtin/README.md
```diff
@@ -0,0 +1,152 @@
...
+Given the above, the only immediate solution is to implement a builtin map type which is specialised to `ByteString` keys but remains polymorphic in the value type. We believe that this solution should be good enough for most practical use-cases, and maybe even surpass a fully polymorphic map type in performance.
+
+
+## Specification
+
+We propose the addition of the following builtin type: `ByteStringMap` of kind `Type -> Type`.
+- It represents a map associating keys of type `ByteString` with values of any builtin type.
+- The new builtin type should be implemented using a tree data structure allowing for logarithmic-time map operations.
+- We consider the costing size of a `ByteStringMap` to be equal to the size of its corresponding representation as an association list.
+- The binary encoding of `ByteStringMap`s is equivalent to its encoding as an association list.
```
ana-pantilie, line 56: @kwxm - please review. This is based on the standard `flat` implementation for Haskell's `Data.Map`.

CIP-bytestring-map-builtin/README.md
effectfully: You probably meant

```haskell
lookupBSMap :: forall a . ByteString -> (() -> a) -> ByteStringMap a -> a
```

? Yes, I believe we'll be able to once pattern matching builtins land (not before that).

ana-pantilie: Yeah I'll fix it, I don't know what I was thinking lol

CIP-bytestring-map-builtin/README.md
```diff
@@ -0,0 +1,152 @@
...
+    - it returns a map with the input key updated to the new value, silently discarding any previous value
+    - it uses logarithmic time in the `ByteStringMap` argument, linear time in the `ByteString` argument, and logarithmic memory in the `ByteStringMap` argument
+2. `lookupBSMap :: forall v . ByteString -> (() -> v) -> ByteStringMap v -> v`
+    - it returns the value for a given key in the map; if the key is missing, it returns the default value of type `() -> a` 
+    - its implementation depends on, as of the time of writing, ongoing work in supporting higher-order builtin functions; see the next section for the motivation behind this choice
+    - it uses logarithmic time in the `ByteStringMap` argument, linear time in the `ByteString` argument and constant memory
+3. `deleteBSMap :: forall v . ByteString -> ByteStringMap v -> ByteStringMap v`
+    - it returns a map without the key-value association identified by the input key
+    - it uses logarithmic time in the `ByteStringMap` argument, linear time in the `ByteString` argument and logarithmic memory in the `ByteStringMap` argument
+4. `unionBSMap :: forall v . ByteStringMap v -> ByteStringMap v -> ByteStringMap v`
```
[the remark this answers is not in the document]
zliu41: @colll78 Thanks for all the pointers! Sounds like we should have `insertWithBSMap` as a builtin. With higher order function support coming soon, this makes total sense.

I'm unsure about the following:
- Should we keep `insertBSMap`
- Do we need a `unionWithBSMap` builtin, or does `insertWithBSMap` suffice
- What can we do to speed up equality checks on Values? It's not simply map equality due to potential zero-value entries. Perhaps `filterBSMap` + `equalsBSMap`?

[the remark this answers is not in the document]
colll78: Map equality is cheap because we currently take advantage of `equalsData`. 

`equalsData (mergeValue ownInputValue tokensToAdd) ownOutputValue`

To handle zero-value entries `mergeValue` simply deletes them, I would expect that the union function would similarly be capable of deleting zero value entries.

CIP-bytestring-map-builtin/README.md
```diff
@@ -0,0 +1,152 @@
...
+
+Map lookups are essential, so we must provide `lookupBSMap`, but there remains the question of how to provide a total function with this functionality.
+
+As mentioned earlier, we have decided to base the implementation of this builtin on an upcoming feature, because it is the least invasive way of providing a total implementation and it leverages future improvements to the builtins language.
+
+With the proposed signature of `lookupBSMap`, users will be able to provide any default value for when the lookup fails, including `error`.
+
+### Traversing `ByteStringMap`s
+
+Unfortunately, due to the limitations presented earlier we cannot provide an effective builtin fold over maps.
```
effectfully, line 106: We could do it we implemented [Allow for returning application trees from builtins a.k.a. folding builtins](https://github.com/IntersectMBO/plutus/issues/6034), which is a straightforward generalization of pattern matching builtins and shouldn't be too hard to get done (famous last words).

effectfully (approved):
Very well-written, thank you.

Sorry for being a party pooper on `insertWithBSMap`/`unionWithBSMap`

CIP-bytestring-map-builtin/README.md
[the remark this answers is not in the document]
colll78: Then just a default `mergeValue` and `insertValue` (that you don't provide a function to, and only works on canonical ledger constructed `Value` as `BuiltinData`) would be just as impactful. Dealing with `Values` is the main use-case and one of the main efficiency bottlenecks in most protocols.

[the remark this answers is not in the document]
colll78: For reference, perhaps one of the most universally beloved smart contract platforms in the blockchain space defines a programming paradigm "Asset oriented programming", the entire language has been built around the notion that it natively understands assets and provides native first-class features for assets. The introduction of such builtins would also vastly enhance the security of the ecosystem because the implementation of such functions in onchain code are extremely error prone, especially when you look at all the black magic optimization that libraries are using (which wouldn't be necessary with such builtins).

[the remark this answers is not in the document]
effectfully: @colll78 sounds interesting. Wanna open an issue in the Plutus repo about that?

However I personally prefer to be as general as possible until we're sure that that's not enough. So I'd rather investigate `unionWithBSMap`, maybe we can do that and then we can keep domain-specific logic in the user space. If that's too hard or isn't going to fly, then I [agree](https://github.com/cardano-foundation/CIPs/pull/915#discussion_r1765832949) with you.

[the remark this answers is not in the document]
colll78: I think that is reasonable.
```haskell
mergeValue :: BuiltinData -> BuiltinData -> BuiltinData 
mergeValue v1 v2 = 
  let valueMap1 = listToBSMap (BI.unsafeDataAsMap v1) 
       valueMap2 = listToBSMap (BI.unsafeDataAsMap v2) 
       combinedVal = unionWithBSMap (\innerA innerB -> unionWithBsMap BI.addInteger (listToBSMap (BI.unsafeDataAsMap innerA)) (listToBSMap (BI.unsafeDataAsMap innerB))) valueMap1 valueMap2
   in toBuiltinData $ filterWithBSMap ((/=) 0) combinedVal
```

In this case `filterWithBSMap` and `equalsBSMap` are definitely required.

CIP-bytestring-map-builtin/README.md
```diff
@@ -0,0 +1,152 @@
...
+Title: Plutus Core Builtin Type - `ByteStringMap`
+Category: Plutus
+Status: Proposed
+Authors:
+    - Michael Peyton Jones <michael.peyton-jones@iohk.io>
+    - Ana Pantilie <ana.pantilie@iohk.io>
+Implementors: []
+Discussions:
+    - https://github.com/cardano-foundation/CIPs/pull/921
+Created: YYYY-MM-DD
```
Ryun1, line 12:
```suggestion
Created: 2024-MM-DD
```

Ryun1: add the date!

CIP-bytestring-map-builtin/README.md
```diff
@@ -0,0 +1,152 @@
...
+Title: Plutus Core Builtin Type - `ByteStringMap`
+Category: Plutus
+Status: Proposed
+Authors:
+    - Michael Peyton Jones <michael.peyton-jones@iohk.io>
+    - Ana Pantilie <ana.pantilie@iohk.io>
+Implementors: []
+Discussions:
+    - https://github.com/cardano-foundation/CIPs/pull/921
+Created: 2024-MM-DD
```
ana-pantilie, line 12:
```suggestion
Created: 2024-10-03
```

CIP-bytestring-map-builtin/README.md
```diff
@@ -0,0 +1,152 @@
+---
+CIP: ????
```
rphair, line 2:
```suggestion
CIP: 139
```

rphair (commented):
@ana-pantilie - a well received pair of proposals in general (with https://github.com/cardano-foundation/CIPs/pull/915) as confirmed at today's CIP meeting; please rename the containing directory accordingly & update the proposal link in the initial posting 🎉

CIP-bytestring-map-builtin/README.md
```diff
@@ -0,0 +1,152 @@
...
+
+Given the above, the only immediate solution is to implement a builtin map type which is specialised to `ByteString` keys but remains polymorphic in the value type. We believe that this solution should be good enough for most practical use-cases, and maybe even surpass a fully polymorphic map type in performance.
+
+
+## Specification
+
+We propose the addition of the following builtin type: `ByteStringMap` of kind `Type -> Type`.
+- It represents a map associating keys of type `ByteString` with values of any builtin type.
+- The new builtin type should be implemented using a tree data structure allowing for logarithmic-time map operations.
+- We consider the costing size of a `ByteStringMap` to be equal to the size of its corresponding representation as an association list.
```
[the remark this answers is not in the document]
effectfully: I think you could simply say "roughly equal" instead of "equal" and that'll be fine, this is all to be figured out during the costing phase anyway, it's fine not to have the exact answer now.

CIP-bytestring-map-builtin/README.md
```diff
@@ -0,0 +1,152 @@
...
+
+Why are higher-order builtin functions necessary? Map keys must be processed (or simply, just compared) in order to be able to build and operate over well-defined maps. For arbitrary key types, that would only be possible by passing the required processing or comparison functions as arguments to the builtin functions.
+
+The only way to implement a builtin map with a (builtin) polymorphic key type would be to modify the builtins language itself. A compiler would have to only allow the construction of maps with key types which are known a priori to posess some property (such as a total ordering). The Plutus Core evaluator would have to check for this as well. The implementation itself would not be too complicated, but we must first consider how this would change the formal semantics of the builtins language.
```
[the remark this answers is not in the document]
effectfully: My comment still stands, I don't believe you'd need to pass a comparison function around, keys can only be of built-in types and we either know how to compare those (in most cases) or don't (like those pesky crypto types) and in the latter case we can just throw. UPLC is a low-level language, it's fine for it to throw like that, and high-level languages will be able to provide a type-safe interface. No meta-theoretical properties are hurt. General maps are gonna be a bit less efficient though, but if done right, probably insignificantly so.

CIP-0139/README.md
```diff
@@ -0,0 +1,152 @@
...
+- The new builtin type should be implemented using a tree data structure allowing for logarithmic-time map operations.
+- We consider the costing size of a `ByteStringMap` to be equal to the size of its corresponding representation as an association list.
+- The binary encoding of `ByteStringMap`s is equivalent to its encoding as an association list.
+
+**Note**: Here `Type` is the universe of all builtin types, since we do not consider types formed out of applying builtin types to arbitrary types to be inhabited.
+
+We propose the following set of builtin functions to accompany the new builtin type:
+1. `insertBSMap :: forall v . ByteString -> v -> ByteStringMap v -> ByteStringMap v`
+    - it returns a map with the input key updated to the new value, silently discarding any previous value
+    - it uses logarithmic time in the `ByteStringMap` argument, linear time in the `ByteString` argument, and logarithmic memory in the `ByteStringMap` argument
```
effectfully, line 63: Could you please talk about `insertWith`/`unionWith` a bit? Specifically, that we want to have them, but cannot unless one of

- [Add support for limited higher-order builtins](https://github.com/IntersectMBO/plutus/issues/6587)
- [Add support for true higher-order builtins](https://github.com/IntersectMBO/plutus/issues/6563)

is implemented.

CIP-0139/README.md
```diff
@@ -0,0 +1,152 @@
...
+2. `lookupBSMap :: forall v . ByteString -> (() -> v) -> ByteStringMap v -> v`
+    - it returns the value for a given key in the map; if the key is missing, it returns the default value of type `() -> a` 
+    - its implementation depends on, as of the time of writing, ongoing work in supporting higher-order builtin functions; see the next section for the motivation behind this choice
+    - it uses logarithmic time in the `ByteStringMap` argument, linear time in the `ByteString` argument and constant memory
+3. `deleteBSMap :: forall v . ByteString -> ByteStringMap v -> ByteStringMap v`
+    - it returns a map without the key-value association identified by the input key
+    - it uses logarithmic time in the `ByteStringMap` argument, linear time in the `ByteString` argument and logarithmic memory in the `ByteStringMap` argument
+4. `unionBSMap :: forall v . ByteStringMap v -> ByteStringMap v -> ByteStringMap v`
```
effectfully, line 71: I wonder if we want `intersectionBSMap` and `differenceBSMap` then.

A horrible nit btw: I'd personally prefer a different capitalization: `unionBsMap`. But feel free to ignore.

CIP-0139/README.md
```diff
@@ -0,0 +1,152 @@
...
+    - it merges the two input maps, left-biased when there are key collisions
+    - it uses linear time in the sum of the first and second arguments, and linear memory in the sum of the first and second arguments
+5. `bsMapToList :: forall v . ByteStringMap v -> List (Pair ByteString v)`
+    - it transforms the map into an association list
+    - the resulting list should be ordered (in ascending order of the keys)
+    - it is linear in the size of the `ByteStringMap`
+6. `listToBSMap :: forall v . List (Pair ByteString v) -> ByteStringMap v`
```
effectfully, line 78: Well, this has a big [problem](https://github.com/IntersectMBO/plutus/issues/6592). I.e. at the moment this builtin is definable but basically unusable. Please reference that issue here.

CIP-0139/README.md
```diff
@@ -0,0 +1,152 @@
...
+
+## Rationale: how does this CIP achieve its goals?
+
+### Constructing `ByteStringMap`s
+
+The first concern we have to address is the reasoning behind not introducing a builtin function for constructing empty `ByteStringMap`s. Since maps are polymorphic in the value type, such a function would require receiving a type tag to know which type of concrete empty map to construct. This provides for an awkward interface to users, and it can easily be avoided by delegating the responsibility of constructing empty map constants to the compiler.
+
+`insertBSMap` and `deleteBSMap` remain fundamental operations which we must expose as builtin functions.
+
+One seemingly unnecessary addition is `unionBSMap`, as it may be implemented in Plutus Core by traversing the elements of one of the maps and calling `insertBSMap`. However, as outlined by [CPS-0013][1], `ByteStringMap` should provide an optimal implementation of the Plutus Tx `Value` type, and the operations on `Value`s depend heavily on map unions. Therefore, for performance reasons, we prefer to expose this functionality as a builtin function.
```
effectfully, line 92: I doubt that `Map`-backed `Value` is going to happen, it's way too much complexity just to get a slow version of a hypothetical built-in `Value`.

effectfully (approved):
A few more comments. Unfortunately, there's something that I didn't realize the first time and it's that `listToBSMap` is unusable, because at the moment it's not possible to construct arbitrary lists of pairs, only lists of pairs of `Data`. Neither the specification nor the formalization have this issue, but there's been zero work done to address it on the implementation side. And it is very non-trivial if we want to do it properly.

CIP-bytestring-map-builtin/README.md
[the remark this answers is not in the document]
kwxm: > @kwxm what should we say here?

@ana: sorry, I missed this.  We actually have to caculate the size of the map object before we carry out any operations on it, and that might be expensive if we have to traverse a tree to  findi ts size.  One way to alleviate this might be to have the maps carry their sizes about with them and make sure that the update operations update the size as well (and during deserialisation we'd have to check that maps embedded in scripts don't lie about their size).  We might even need two measures of size: one telling you the number of elements of the map (for converting to a list, for example), and one telling you the depth of the map (for insertion etc.). We could do this by using newtype wrappers to choose which size measure should be applied.  Maybe we could just store the size of the object and take the log instead of measuring the depth, since these should be proportional if the tree's always kept balanced.  This is all probably too detailed for a CIP though, so saying that the size should be something like the number of entries should be fine: we can work out the details later.

CIP-0139/README.md
kwxm, line 63: Would thememory usage relly be logarithmic in the size of the map?  In most functional data structures adding a new entry just requires a constant amount of space.  Maybe it's not that simple if you need to keep your structure balanced though: I should go and look up how you actually implement these things.

CIP-0139/README.md
```diff
@@ -0,0 +1,152 @@
...
+The only way to implement a builtin map with a (builtin) polymorphic key type would be to modify the builtins language itself. A compiler would have to only allow the construction of maps with key types which are known a priori to possess some property (such as a total ordering). The Plutus Core evaluator would have to check for this as well. The implementation itself would not be too complicated, but we must first consider how this would change the formal semantics of the builtins language.
+
+Given the above, the only immediate solution is to implement a builtin map type which is specialised to `ByteString` keys but remains polymorphic in the value type. We believe that this solution should be good enough for most practical use-cases, and maybe even surpass a fully polymorphic map type in performance.
+
+
+## Specification
+
+We propose the addition of the following builtin type: `ByteStringMap` of kind `Type -> Type`.
+- It represents a map associating keys of type `ByteString` with values of any builtin type.
+- The new builtin type should be implemented using a tree data structure allowing for logarithmic-time map operations.
```
kwxm, line 54: Would we have to implement all of this ourselves, or is there some existing library that we can use?

ana-pantilie:
After discussing internally, I've unfortunately decided to close this CIP.

The reason behind this decision is that we have discovered that it will not actually solve the original problem: introducing a builtin type for representing `Value` optimally. Together with the limitation pointed out by @effectfully in regards to `listToBSMap`, this makes a builtin `ByteStringMap` into an oddly specific and inflexible builtin with an awkward API.

@colll78 and @effectfully believe that all the overhead required to implement the `Value` logic on top of `ByteStringMap` will result in poor performance, so in this sense `ByteStringMap` is too general for this use-case.

To make matters worse, the limitation regarding `listToBSMap` further restricts its usage. Not having a dedicated `Pair` constructor in `Data` breaks a lot of type transformations between `ByteStringMap` and other types, resulting in a misleading and incomplete implementation.

Whether a builtin map could ever be useful, @colll78 believes a more general builtin `Map`, polymorphic in both input types does have some use cases, however a current implementation would suffer from even more limitations which we need to address first.

To conclude, the language isn't ready yet for a builtin `Map` type.

I'm glad that at least this CIP has generated discussions which have made that clear and we have a better idea of what to try next. @rphair thank you for your time considering and reviewing this CIP.

rphair:
Thank you as well @ana-pantilie for your honesty & comprehensive documentation, along with your reviewers, of this problem space. ⭐

PR #934, opened 2024-11-06
CIP-???? | On-Chain Parameter Application

keyan-m:
We propose a new builtin Plutus function, `builtinApplyParams`, to reduce the costs and complexity of instantiating parameterized scripts on-chain.

A sample use-case for performing this computation on-chain would be enabling validation by a minting policy for the destinations of its tokens, ensuring that only instances of a specific script can be the recipients.

Without this builtin, apart from development complexities, script instances need to be limited to hashed parameters that must be resolved via the redeemer in each transaction, and therefore bloat the chain with duplicate data.

---

([latest version in branch](https://github.com/keyan-m/CIPs/blob/param-application/CIP-param-application/README.md))

### DIFF
--- a/CIP-param-application/README.md
+++ b/CIP-param-application/README.md
@@ -102,10 +102,10 @@ credential.
 ## Path to Active
 
 ### Acceptance Criteria
-- [] Fully implemented in Cardano.
+- [ ] Fully implemented in Cardano.
 
 ### Implementation Plan
-- [] Passes all requirements of both Plutus and Ledger teams as agreed to improve Plutus script efficiency and usability.
+- [ ] Passes all requirements of both Plutus and Ledger teams as agreed to improve Plutus script efficiency and usability.
 
 ## Copyright
 This CIP is licensed under [CC-BY-4.0](https://creativecommons.org/licenses/by/4.0/legalcode).
--- a/CIP-param-application/README.md
+++ b/CIP-param-application/README.md
@@ -72,7 +72,7 @@ In short, we'll get:
 Implementing such a validation on-chain would involve a few steps.
 
 First, the whole target contract has to be wrapped within an outer function
-which the parameters will be applied to it. This leads to single occurances of
+which the parameters will be applied to. This leads to single occurances of
 the parameters throughout the script's CBOR.
 
 Secondly, each parameter will be required to have fixed lengths in order to
--- a/CIP-param-application/README.md
+++ b/CIP-param-application/README.md
@@ -23,18 +23,11 @@ Complex contracts tend to depend on multiple scripts, and in some cases,
 validation from scripts with some unknown parameters are required.
 
 As a simple example, assume a minting script which needs to validate that the
-destination of its token is a a parameterized spending that is uniquely
+destination of its token is a parameterized spending script that is uniquely
 instantiated for a user.
 
-The current solution involves wrapping the spending script with an inner lambda,
-expecting the parameters as hashed values to ensure their constant lengths, and
-expect each parameter to be provided via the redeemer. This approach allows
-slicing unapplied CBOR of the applied script at specific indexes in order to
-validate the presence of the expected hash.
-
-This, however, is a costly and relatively complex approach.
-
-With our proposed builtin function, the minting script of the example above can
+The [current solution](#alternatives), is costly and relatively complex. With
+our proposed builtin function, the minting script of the example above can
 be provided with the unapplied script (either directly or indirectly), and have
 arbitrary values applied to it as parameters.
 
@@ -42,7 +35,7 @@ arbitrary values applied to it as parameters.
 
 ### Function Definition
 
-we propose the Plutus builtin function `builtinApplyParams`, with the following
+We propose the Plutus builtin function `builtinApplyParams`, with the following
 type signature:
 ```hs
 builtinApplyParams :: BuiltinByteString -> [BuiltinData] -> ScriptHash
@@ -60,13 +53,49 @@ TODO
 
 ## Rationale: how does this CIP achieve its goals?
 
-With a builtin function, we'll get cheap and simple on-chain parameter
-validation. Consequently, this will enable easier implementation of advanced
-contracts with enhanced guarantees.
+Using the [provided example](#motivation-why-is-this-cip-necessary), the minting
+script can have access to the unapplied script:
+* Either by direct inclusion as a parameter (which can lead to a large script size)
+* Or providing a UTxO as a parameter that houses the unapplied script so that
+  the minting script can read it without becoming excessively large
+
+Furthermore, arbitrary parameters can be provided via the redeemer which can be
+used to generate an instance's script hash.
+
+In short, we'll get:
+* Easier and cheaper parameter verification on-chain
+* A more automation friendly build for such architectures
+* Less error prone builds
 
 ### Alternatives 
 
-Please see example above under the [motivation section](#motivation-why-is-this-cip-necessary).
+Continuing to use the [example above](#motivation-why-is-this-cip-necessary), implementing
+such a validation on-chain would involve a few key steps.
+
+First, the whole target contract has to be wrapped within an outer function
+which the parameters will be applied to it. This leads to single occurances of
+the parameters throughout the script's CBOR.
+
+Secondly, each parameter will be required to have fixed lengths in order to
+allow validation of the bytes surrounding parameters. This can be achieved by
+assuming these parameters are all hashes, and that their corresponding values
+will be provided via the redeemer.
+
+Next, in order to validate a given script is an instance of a known script,
+first few bytes of an instance must be provided (up until where the parameter is
+placed). With this "prefix" at hand, the instance's CBOR can be constructed
+on-chain as such:
+```hs
+let appliedCBOR =
+  prefix <> param0 <> paramHeader <> param1 <> paramHeader <> param2 <> postfix
+```
+
+Where `postfix` is similar to `prefix`, but for the rest of the script until its
+end. Note that `paramHeader` can be reused based on the presumption that all
+parameters are of equal lengths.
+
+Finally, this `appliedCBOR` can be hashed to attain its corresponding script
+credential.
 
 ## Path to Active
 
--- a/CIP-param-application/README.md
+++ b/CIP-param-application/README.md
@@ -43,7 +43,7 @@ builtinApplyParams :: BuiltinByteString -> [BuiltinData] -> ScriptHash
 
 Where the first argument is the unapplied script's CBOR, and
 the `BuiltinData` list is all the parameters that are to be applied in order.
-Finally the result will be hash of the applied script.
+Finally, output will be the hash of the applied script.
 
 
 ### Cost Model
--- a/CIP-param-application/README.md
+++ b/CIP-param-application/README.md
@@ -64,13 +64,12 @@ used to generate an instance's script hash.
 
 In short, we'll get:
 * Easier and cheaper parameter verification on-chain
-* A more automation friendly build for such architectures
+* A more automation-friendly build for such architectures
 * Less error prone builds
 
 ### Alternatives 
 
-Continuing to use the [example above](#motivation-why-is-this-cip-necessary), implementing
-such a validation on-chain would involve a few key steps.
+Implementing such a validation on-chain would involve a few steps.
 
 First, the whole target contract has to be wrapped within an outer function
 which the parameters will be applied to it. This leads to single occurances of
@@ -82,7 +81,7 @@ assuming these parameters are all hashes, and that their corresponding values
 will be provided via the redeemer.
 
 Next, in order to validate a given script is an instance of a known script,
-first few bytes of an instance must be provided (up until where the parameter is
+first bytes of an instance must be provided (up until where the parameters are
 placed). With this "prefix" at hand, the instance's CBOR can be constructed
 on-chain as such:
 ```hs
--- a/CIP-param-application/README.md
+++ b/CIP-param-application/README.md
@@ -1,0 +1,83 @@
+---
+CIP: ????
+Title: On-Chain Parameter Application
+Status: Proposed
+Category: Plutus
+Authors:
+    - Keyan M. <keyanmaskoot@gmail.com>
+Implementors: []
+Discussions: 
+    - https://github.com/cardano-foundation/CIPs/pull/999/
+Created: 2024-11-04
+License: CC-BY-4.0
+---
+
+## Abstract
+We propose a new Plutus builtin function `builtinApplyParams` with the
+signature `BuiltinByteString -> [BuiltinData] -> ScriptHash` in order to enable
+simple and low cost on-chain check for validating instances of unapplied
+scripts.
+
+## Motivation: why is this CIP necessary?
+Complex contracts tend to depend on multiple scripts, and in some cases,
+validation from scripts with some unknown parameters are required.
+
+As a simple example, assume a minting script which needs to validate that the
+destination of its token is a a parameterized spending that is uniquely
+instantiated for a user.
+
+The current solution involves wrapping the spending script with an inner lambda,
+expecting the parameters as hashed values to ensure their constant lengths, and
+expect each parameter to be provided via the redeemer. This approach allows
+slicing unapplied CBOR of the applied script at specific indexes in order to
+validate the presence of the expected hash.
+
+This, however, is a costly and relatively complex approach.
+
+With our proposed builtin function, the minting script of the example above can
+be provided with the unapplied script (either directly or indirectly), and have
+arbitrary values applied to it as parameters.
+
+## Specification
+
+### Function Definition
+
+we propose the Plutus builtin function `builtinApplyParams`, with the following
+type signature:
+```hs
+builtinApplyParams :: BuiltinByteString -> [BuiltinData] -> ScriptHash
+```
+
+Where the first argument is the unapplied script's CBOR, and
+the `BuiltinData` list is all the parameters that are to be applied in order.
+Finally the result will be hash of the applied script.
+
+
+### Cost Model
+
+TODO
+
+
+## Rationale: how does this CIP achieve its goals?
+
+With a builtin function, we'll get cheap and simple on-chain parameter
+validation. Consequently, this will enable easier implementation of advanced
+contracts with enhanced guarantees.
+
+### Alternatives 
+
+Please see example above under the [motivation section](#motivation-why-is-this-cip-necessary).
+
+## Path to Active
+
+### Acceptance Criteria
+- [] Fully implemented in Cardano.
+
+### Implementation Plan
+- [] Passes all requirements of both Plutus and Ledger teams as agreed to improve Plutus script efficiency and usability.
+
+## Copyright
+This CIP is licensed under [CC-BY-4.0](https://creativecommons.org/licenses/by/4.0/legalcode).
+
+[CC-BY-4.0]: https://creativecommons.org/licenses/by/4.0/legalcode
+[Apache-2.0]: http://www.apache.org/licenses/LICENSE-2.0
--- a/CIP-param-application/README.md
+++ b/CIP-param-application/README.md
@@ -4,11 +4,13 @@ Title: On-Chain Parameter Application
 Status: Proposed
 Category: Plutus
 Authors:
+    - Philip DiSarro <info@anastasialabs.com>
     - Keyan M. <keyanmaskoot@gmail.com>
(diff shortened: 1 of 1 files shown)


CIP-param-application/README.md
```diff
@@ -0,0 +1,113 @@
+---
+CIP: ????
+Title: On-Chain Parameter Application
```
rphair, line 3: It would be good if now and/or at the meeting we could settle on a CIP title the same as the PR title.  We've had to remove back-ticked expressions from CIP titles recently (#930, to fix portability problems) so the next best choice would either be this expression above or one that concisely describes the new builtin (instead of quoting its name).

rphair (commented):
thanks @keyan-m - I've tagged for `Triage` to put it on the agenda for our next CIP meeting (https://hackmd.io/@cip-editors/100).  @MicroProofs I hope you will also be with us this time to help introduce it & likewise @colll78 if you can make it.

colll78:
This proposed solution solves:

https://github.com/cardano-foundation/CIPs/pull/354
https://github.com/cardano-foundation/CIPs/pull/385

CIP-param-application/README.md
```diff
@@ -0,0 +1,113 @@
...
+arbitrary values applied to it as parameters.
+
+## Specification
+
+### Function Definition
+
+We propose the Plutus builtin function `builtinApplyParams`, with the following
+type signature:
+```hs
+builtinApplyParams :: BuiltinByteString -> [BuiltinData] -> ScriptHash
```
perturbing, line 43: Why do you return a `ScriptHash` and not a `BuiltinByteString`?  Besides this `ScriptHash` not being a plutus type, would it not be nice to return the applied CBOR, so that one can apply parameters over multiple transactions.

CIP-param-application/README.md
```diff
@@ -0,0 +1,113 @@
...
+```
+
+Where the first argument is the unapplied script's CBOR, and
+the `BuiltinData` list is all the parameters that are to be applied in order.
+Finally, output will be the hash of the applied script.
+
+
+### Cost Model
+
+TODO
```
perturbing, line 53: I wonder what the impact here would be, since this suggested built-in is effectively running the plutus machinery inside an already running instance of it.

My intuition tells me that this is okay, since the complexity of this calculation is always bound by the complexity of any possible other script. But looking forward to more insight from @kwxm :)

CIP-param-application/README.md
keyan-m: My bad :no_mouth: I confused the datatype with the one from Aiken (which is just an alias there).

colll78 (approved):

colll78 (commented):
The arguments changed to cborHex now allows SOP parameters to be applied.

CIP-param-application/README.md
perturbing: Also tagging @bezirg for his opinion

CIP-param-application/README.md
rphair: resolved with https://github.com/cardano-foundation/CIPs/pull/934#event-15240537419.

rphair:
@kwxm we decided the last CIP meeting 2 days ago to wait as per https://github.com/cardano-foundation/CIPs/pull/934#discussion_r1838146570 until you had gone over this to confirm feasibility & desirability of this change... once that's done we would confirm this as a candidate & assign a CIP number.  (cc @keyan-m @bezirg @perturbing)

colll78:
> @kwxm we decided the last CIP meeting 2 days ago to wait as per [#934 (comment)](https://github.com/cardano-foundation/CIPs/pull/934#discussion_r1838146570) until you had gone over this to confirm feasibility & desirability of this change... once that's done we would confirm this as a candidate & assign a CIP number. (cc @keyan-m @bezirg @perturbing)

Costing should just be the sum of:
1. the setup cost for the CEK machine with the script we want to apply arguments to
2. the cost of one `apply` 
3. overhead?

I think the above can be fully ignored though, and the costing process can be identical the process used to bench any other external functional call (ie. `validateSignature`). Just run large amounts of tests to see the impact the input size has on CPU/MEM and determine a function that accurately models that data as closely as possible.

kwxm:
Sorry, I completely missed this and only became aware of it a week or so ago.

Unfortunately I think that the proposal is infeasible.  If I understand correctly, the idea is that the new builtin should take a serialised script and a number of bytestring parameters and produce the serialised form of the script applied to those parameters.  For example, given the serialised version `s` of a script

```
program 1.0.0 (LamAbs x (LamAbs y (LamAbs z <body>)))
```
and three bytestrings `b1`,`b2`, and `b3`, the  builtin should return the serialised version of the script

```
program 1.0.0 (Apply (Apply (Apply (LamAbs x (LamAbs y (LamAbs z <body>))) b1) b2) b3)
```
(I'm using pseudocode based on the [Haskell datatype](https://github.com/IntersectMBO/plutus/blob/b012b8083a60693eb2d22064cded7a6f115eb296/plutus-core/untyped-plutus-core/src/UntypedPlutusCore/Core/Type.hs#L74) representing the basic constructors of the UPLC language here).  I found the explanation in the proposal somewhat unclear because it's not explicit about exactly what needs to be added, but I think that this is roughly the idea.  The important point is that it involves inserting new stuff near the start of a serialised script, which as we'll see later presents ceratin problems.

One way to perform the above transformation would be to deserialise `s` to get an abstract syntax tree, modify it to obtain the AST of the second program, then serialise that to get a new bytestring.  However that would require on-chain deserilaisation and reserialisation, and that would probably be too slow and expensive to be worthwhile.  As the proposal suggests, another way to obtain the same effect would be to directly modify the bytestring `s` containing the serilased version of the first program to transform it into the serialised version of the second problem.  The trouble with this is that the serialisation format is rather complicated.

The proposal refers to the "CBOR serialisation" of the script.  Serialised scripts are indeed represented on the chain as CBOR bytestrings, but these are just wrappers round a bytestring obtained by serialising the script using the `flat` format (see Appendix C of the [Plutus Core specification](https://github.com/IntersectMBO/plutus/tree/master/doc/plutus-core-spec) (PDF [here](https://plutus.cardano.intersectmbo.org/resources/plutus-core-spec.pdf))). CBOR is byte-oriented, and if we used CBOR for serialisation then every construct in a serialised UPLC progam would take up a whole number of bytes, and then it would be easy to implement the proposal by inserting new bytes into the serialised form.  However, because a whole number of bytes are required for everything there's a lot of wasted  spaceL for example you need one byte for every AST constructor despite the fact that there are only ten of them at present.   

It's important to keep serialised scripts as small as possible to reduce traffic on the network and the space required to store the blockchain on disk, so we use the _bit-oriented_ flat format instead, which reduces sizes significantly by using different numbers of bits to store objects of different sizes; for example, the ten basic AST constructors are each represented by four bits, and the tag of a built-in function is represented using seven bits.  In particular the `Apply` constructor requires four bits,  so to insert an odd number of them near the start of a serialised script we'd have to shift all of the rest of the script along by four bits, splitting each byte into two four-bit halves and then concatenating the right half of each byte with the left half of the following one.  This might be feasible, but it would be fiddly and possibly quite time consuming.  However, this still wouldn't work!  The problem is that bytestring constants in flat-encoded UPLC programs still have to be byte-aligned: this makes it very easy to encode them by padding whatever you've emitted so far up to a whole number of bytes with some zero bits and then just appending the bytesttring.  The difficulty then is that if you try to insert four bits for `Apply` near the start of a bytestring containing a serialised program you don't just have to shift the remainder of the bytestring along by four bits, you also have to maintain the byte alignment of any embedded bytestring constants.  Doing this would essentially require you to parse the whole of the rest of the script and you'd more or less be back to expensive on-chain deserialisation.

This problem only arises if you want to insert an odd number of `Apply`s intoa serialised program .  If you want to insert an even number then it really is just a matter of inserting a whole number of bytes (and also appending the script arguments at the end), which is reasonably easy and cheap.  This might make it tempting to suggest adding a dummy parameter when you have an odd number of arguments so that you always have to insert an even number of `Apply`s.  A potential difficulty is that in future we may add new term constructors in UPLC (we increased the number from 8  to 10 last year when we [added](https://github.com/IntersectMBO/plutus/pull/5240/) `Constr` and `Case` for example), and if the number ever exceeds 16 then we'll have to introduce a new version of the flat encoding that uses five bits for `Apply`, and then a  one-argument script would require seven dummy arguments to preserve alignment, which might be a bit excessive (and we'd need a new version of the builtin to deal with that).

Another possbility for working around this problem would be to allow incomplete serialised scripts which have the necessary `Apply` nodes near the start but don't have anything to apply them to: with this, application would simply become a matter of appending the bytestring arguments to the serialised script.  However, if we did that then the unapplied script wouldn't represent a valid UPLC AST and wouldn't be deserialisable on its own, but only when applied.  I think this would be quite a major change which might need modifications to the ledger/node, so I don't know if it's actually feasible.  Trying to get this implemented would probably require a lot of work and a comptely new proposal that would deal with the ramifications for the ledger.

In a completely different dierction, note the [proposed CIP](https://github.com/cardano-foundation/CIPs/pull/946) for modules in Plutus Core, which I think is very likely to be approved.  Would that be of any help with this issue?

--------------------------

In summary,  I don't think that this proposal is workable in its current form because of the complexities introduced by the `flat` encoding, which I suspect weren't appreciated by the author(s).   It may be possible to work around this in one way or another, but that'll probably require careful thought and a new proposal.

colll78:
The fact of the matter is that onchain parameter application (via the hacky bytestring method) is a popular design pattern for onchain applications. The issue is that as you have pointed out, due to the complexity of the serialization format, this bytestring manipulation is very error prone and often results in a number of edge-case issues. The proposal here is to introduce a builtin function that replaces this design pattern (currently complex bytestring manipulation) with a more reasonable native solution that handles edge-cases. Either way, the fact of the matter is this design pattern will continue to be used in onchain protocols due to its immense unique utility, so either it will continue to be used in its current very hacky error-prone state without native support, or we design and introduce a builtin that will make this process more high-level and less error prone. If you are claiming that we cannot design such a builtin, then the ecosystem is forced to continue this weird hacky approach which will be a permanent blemish on developer experience and onchain security.

keyan-m:
It seems my possibly ignorant use of "script CBOR" has caused some confusion. The encoding of the scripts is not the subject here. This proposed builtin would take a byte array, along with a list of byte arrays, and it'd return another byte array in a deterministic manner.

There are already implementations for achieving this—the one from `uplc` crate perhaps being one of the most widely used:
https://github.com/aiken-lang/aiken/blob/21c6a6a71d66fe7723f94698c023db9559ad70c8/crates/uplc/src/tx.rs#L157-L177

rphair:
@colll78 @keyan-m @kwxm we don't have many precedents so far for addressing such arguments of infeasibility of a change that informed members of the community would be highly in favour of.  All I can think of doing for now is to try to address any potential deadlock on the next CIP meeting agenda, for which I think @colll78 you might already be present due to the re-review of #444: https://hackmd.io/@cip-editors/106 - cc @Ryun1 @perturbing @Crypto2099

kwxm:
> There are already implementations for achieving this—the one from uplc crate perhaps being one of the most widely used:

Thanks!  It's very helpful to see that.  What is it used for though?  Is this something that happens off the chain?  If I understand the code correctly (and it's quite possible that I don't) it's deserialising the script and wrapping it in an `Apply` to apply it to its arguments, which is essentially what I mentioned in the paragraph beginning "One way to perform the above transformation..." [earlier](https://github.com/cardano-foundation/CIPs/pull/934#issuecomment-2649090800).  As I said there, I suspect that this would be prohibitively expensive todo on the chain.  If you could supply some realistic examples then we could do some benchmarking to see just how expensive it is.

colll78:
> > There are already implementations for achieving this—the one from uplc crate perhaps being one of the most widely used:
> 
> Thanks! It's very helpful to see that. What is it used for though? Is this something that happens off the chain? If I understand the code correctly (and it's quite possible that I don't) it's deserialising the script and wrapping it in an `Apply` to apply it to its arguments, which is essentially what I mentioned in the paragraph beginning "One way to perform the above transformation..." [earlier](https://github.com/cardano-foundation/CIPs/pull/934#issuecomment-2649090800). As I said there, I suspect that this would be prohibitively expensive todo on the chain. If you could supply some realistic examples then we could do some benchmarking to see just how expensive it is.

[See here](https://github.com/Anastasia-Labs/aiken-design-patterns/blob/develop/lib/aiken-design-patterns/parameter-validation.ak) for an example from our design-patterns repo that illustrates this being done in onchain code in Aiken.

CIP-param-application/README.md
```diff
@@ -0,0 +1,114 @@
...
+Finally, this `appliedCBOR` can be hashed to attain its corresponding script
+credential.
+
+## Path to Active
+
+### Acceptance Criteria
+- [] Fully implemented in Cardano.
+
+### Implementation Plan
+- [] Passes all requirements of both Plutus and Ledger teams as agreed to improve Plutus script efficiency and usability.
```
rphair, line 108:
```suggestion
### Acceptance Criteria
- [ ] Fully implemented in Cardano.

### Implementation Plan
- [ ] Passes all requirements of both Plutus and Ledger teams as agreed to improve Plutus script efficiency and usability.
```
These are not formatted as tick-boxes unless there is some space inside the box to put the tick 🤓

rphair (commented):
@keyan-m the result of discussing this at yesterday's CIP meeting was to confirm that the current presentation of proposed changes doesn't justify the impact it would have on the network.  I am not the best person to explain such consequences but to me it sounded like whatever "bloat" would be saved by not having to do the current workaround(s) would not justify the collective cost in execution time.

From my point of view there also seemed to be a parallel issue: that of developer convenience.  I can see how a natural feature to avoid a lot of developer complexity would be popular and well appreciated, but again the consensus of subject matter experts at the meeting was that existing "workaround" design patterns — however bulky or inconvenient — must be tolerated until this proposed change can be reformulated in a way that @kwxm and others can verify will be:
- not "clunky" or "hacky" (quotes from the meeting);
- precisely costed in both time and space, with those costs justifiable.

I hope this effort continues on GitHub since the essential question about how to settle https://github.com/cardano-foundation/CIPs/pull/934#issuecomment-2659818428 has already been answered in meeting review: therefore further meeting discussion can wait until online postings produce a better result.  Postings that promote this proposal & any further revisions to the idea should look in greater engineering detail: to obtain and then verify a more robust method for dealing with this issue.

cc @MicroProofs @michele-nuzzi @fallen-icarus (please everyone keep tagging _both_ developers that may be in favour of this change _and_ architects who may be critical of it)

MicroProofs:
> > Another possbility for working around this problem would be to allow incomplete serialised scripts which have the necessary Apply nodes near the start but don't have anything to apply them to: with this, application would simply become a matter of appending the bytestring arguments to the serialised script. However, if we did that then the unapplied script wouldn't represent a valid UPLC AST and wouldn't be deserialisable on its own, but only when applied. I think this would be quite a major change which might need modifications to the ledger/node, so I don't know if it's actually feasible.
> 
> @kwxm This sounds like a great idea, and I don't think it requires any change to the ledger whatsoever. There's no problem with the unapplied script not representing a valid UPLC AST, since it doesn't need to be deserialized to a valid AST.

It's starting to feel like this is a good case for (case  (constr  0 [..args]) [validator]) which allow for unlimited applies without other changes to the script (i.e. apply AST nodes). The one caveat is that params would need to be byte aligned which is the case  for both data and bytearray which is the most common params currently.

effectfully (commented):
I'm not really sure what exactly this CIP proposes.

But it seems we can address it in two ways:

1. allow partial scripts on the chain having the right `Apply` nodes, so that we don't need to put them in there, but no specific arguments
2. store fully applied scripts but the arguments are dummies that get stripped off by the built-in function, so that we can put the right ones there, like it's done in [here](https://github.com/Anastasia-Labs/aiken-design-patterns/blob/develop/lib/aiken-design-patterns/parameter-validation.ak) (except not by a builtin)

Both are probably doable with the former being a saner solution?

rphair:
> @kwxm https://github.com/cardano-foundation/CIPs/pull/934#issuecomment-2731377672: As required by [CIP-0001](https://cips.cardano.org/cip/CIP-0001) the proposal should give a precise specification of exactly what the new function should do.

Also @keyan-m you can use CIP-0035 as a checklist for what must be in the CIP (both now & when the CIP becomes `Active`), including a detailed checklist of requirements for builtin functions: https://github.com/cardano-foundation/CIPs/blob/master/CIP-0035/README.md#processes

rphair:
@keyan-m this has been tagged `Waiting for Author` for a couple of months.  We realise that it might take some effort to push through the mixture of responses from Plutus expert review about whether this change as presented would be a "good thing" as argued by @colll78 and I think also by @MicroProofs.

But without an author-advocate to respond to that review then we might not be able to take this on as a CIP.  If discussion is happening to help clarify this issue then the author, or someone in the CIP community, needs to post what measures are being taken — or will be taken —  to produce a proposal that the Plutus architects can implement.

keyan-m:
Hey @rphair, thank you for the reminder. I have been a bit preoccupied. I will try to clarify/refine the proposal as soon as possible.

rphair:
thanks @keyan-m - good to know it's not `Abandoned` ☀️

rphair (commented):
@keyan-m @colll78 @microproofs - after about a year we are going through the `Waiting for Author` queue & trying to decide which proposals should be more properly considered `Likely Abandoned`.

Several of the comments above suggest the flaws identified by @kwxm can be solved, perhaps accommodating suggestions by @zliu41, @perturbing and others.  But we would have to record some tangible progress on these updates here in this thread to keep this PR open... otherwise more time passing will make it more likely that the needs for this CIP (https://github.com/cardano-foundation/CIPs/pull/934#issuecomment-2468786649) will be addressed some other way & invalidate the work proposed here.

So we have to consider this potentially `Abandoned` and may close it at the next cleanup cycle unless we can document current validity somehow.  Also the Plutus experts tagged above should please let us know if there is a more recent solution for the identified problem(s) or if in fact something like this method or solution has already been included in Plutus.

keyan-m:
After considering this proposal further, supporting parameter validation with _arbitrary shape_ does not seem to be of great benefit (which is where [the current solution](https://github.com/Anastasia-Labs/aiken-design-patterns/blob/82bdbdfc727388d048f12c0a069d197bf9ffb477/lib/aiken-design-patterns/parameter-validation.ak#L170-L188) proves too clunky).

As a short re-iteration, said solution requires scripts to be applied with a hashed parameter, and have the pre-image provided via the redeemer in any endpoint that needs access to it.

However, I'm now thinking scripts that need to validate instantiations of other parameterized scripts will most likely be able to be smart about it, and just use a hash without ever requiring its pre-image.

Meaning, by [simple bytestring manipulations](https://github.com/Anastasia-Labs/aiken-design-patterns/blob/82bdbdfc727388d048f12c0a069d197bf9ffb477/lib/aiken-design-patterns/parameter-validation.ak#L83-L97) (detailed [here](https://github.com/Anastasia-Labs/design-patterns/blob/62d97c1a32a040b877f7d78cc915d9c788bc9a89/parameter-validation/PARAMETER-VALIDATION.md)) we have a cheap solution, which its only downside could be developer-unfriendliness.

To demonstrate "developer-unfriendliness," I can point to how we've used it in Midgard:
- [The parameterized script itself](https://github.com/Anastasia-Labs/midgard/blob/15361d94a1180c51d957cc8e4bc53a5231568b71/onchain/aiken/validators/user-events/witness.ak)
- [Its generated CBOR bytes prefix](https://github.com/Anastasia-Labs/midgard/blob/15361d94a1180c51d957cc8e4bc53a5231568b71/onchain/aiken/env/default.ak#L63-L64) and [a temporary AI script](https://github.com/Anastasia-Labs/midgard/blob/staging/scripts/generate-user-events-witness-script-prefix.sh) that generates it
- [Validation of a resulting script hash](https://github.com/Anastasia-Labs/midgard/blob/15361d94a1180c51d957cc8e4bc53a5231568b71/onchain/aiken/lib/midgard/user-events.ak#L136-L143)

All in all, it seems we can either close this CIP, or keep it open in the hopes that at some point implementing this would be easy enough that its benefits would out-weigh its effort.

Appreciate all the good people chiming in 🫡

rphair:
thanks @keyan-m — your last comment is a good result of the whole process of considering this CIP, and provides both practical & theoretical means of continuing to research and write about this issue in other contexts.

We'll therefore close this and CIP editors (cc @Ryun1 @perturbing) will note to refer authors & reviewers back to this thread any time parameter validation is presented as an issue or in another CIP/CPS draft.

And of course you or another advocate would be welcome to reopen this if there is enough inspiration & groundwork to continue this CIP with a more promising approach. 🙏

PR #935, opened 2024-11-07
CPS-??? | Social Governance

Crypto2099:
A new CPS to orchestrate solutions to social governance problems that cannot be explicitly addressed through on-chain procedures and may need to be more malleable versus the on-chain Constitution document. This gives us an alternative pathway to steer and orchestrate new standards and behaviors amongst the community.

### DIFF
--- a/CPS-XXXX/README.md
+++ b/CPS-XXXX/README.md
@@ -1,0 +1,111 @@
+---
+CPS: ????
+Title: Social Governance
+Status: Open
+Category: Governance
+Authors:
+  - Adam Dean <adam@crypto2099.io>
+  - Adam Rusch <adam@rusch.me>
+  - Kyle Solomon <>
+Proposed Solutions:
+  - TBD
+Discussions:
+  - https://lu.ma/constitution
+  - https://github.com/Crypto2099/draft-constitution/issues/5
+Created: 2024-11-04
+License: CC-BY-4.0
+---
+
+## Abstract
+
+As a blockchain ecosystem, Cardano is not only an immutable ledger of data but
+also a cooperative network of individuals and organizations engaged in economic
+collaboration, social impact, decentralized decision-making, and technological
+development. Therefore, Cardano's governance system cannot be limited to the
+source code of the software and on-chain effects of transactions but also must
+guide the interactions of individuals and organizations who choose to interact
+within the network.
+
+## Problem
+
+The Cardano Community is in need of guidance and procedures for dealing with
+governance issues. The Cardano Constitution is one document that helps
+adjudicate such issues. The Constitutional Committee - one of the three bodies
+designated in the Cardano Governance System - is bound to uphold this document
+when voting on the passage of Governance Actions.
+
+However, there are some issues or practices which the Constitution does not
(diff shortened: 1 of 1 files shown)


rphair:
@Crypto2099 it is a pretty big implication that this creates a `Governance` category.  As a devil's advocate I would have to say that the non-technical (i.e. non-`Metadata`) subject matter places it outside the boundaries of the CIP process as defined by CIP-0001 itself & confirmed all kinds of ways afterward.

If it's not considered off-topic for the CIP process, the next question is how it relates to CPS-0007.  Does it extend it, or deprecate it?  In either case, why doesn't the latter document refer to the earlier one?  Since the subject matter overlaps, how would people know which is relevant in the cases comment to both?  So, most importantly perhaps, why isn't this PR an update to CPS-0007?

https://github.com/cardano-foundation/CIPs/tree/master/CPS-0007

Crypto2099:
> @Crypto2099 it is a pretty big implication that this creates a `Governance` category. As a devil's advocate I would have to say that the non-technical (i.e. non-`Metadata`) subject matter might place it outside the boundaries of the CIP process as defined by CIP-0001 itself & confirmed repeatedly afterward: especially in the arduous GitHub discussion of CIP-1694 before it was merged.

I would argue that by lumping everything under `metadata` we then make the categorization itself worthless and if `Catalyst` can have its own category then certain on-chain and off-chain governance should qualify to have a dedicated categorization scheme (although that is not a hill for this CPS to die on).

> If it's not considered off-topic for the CIP process, the next question is how it relates to CPS-0007. Does it extend it, or deprecate it? In either case, why doesn't the latter document refer to the earlier one? Since the subject matter overlaps, how would people know which is relevant in cases where both might appear applicable? So, most importantly perhaps, why isn't this PR an update to CPS-0007?

I would argue that CPS-0007 related to on-chain governance (i.e. CIP-1694) and the technical implementation of governance whereas this CPS aims to address a fundamentally new issue that has only arisen this year during the drafting on the Cardano Constitution and that is the "off chain" processes that are necessary such as:
* Governance Actor Codes of Conduct
* Budget and other procedural patterns
* Election procedures

Also thank you Robert and looking forward to discussing this one (and the other CIPs to follow) during the next editor's meeting!

rphair:
Removing `Triage` label as per https://github.com/cardano-foundation/CIPs/issues/937#issuecomment-2468168150 until https://github.com/cardano-foundation/CIPs/issues/937 achieves CIP editor quorum & community consensus.

rphair:
At yesterday's CIP meeting we confirmed that @crypto2099 intends to create a repository for social governance propositions — roughly as per https://github.com/cardano-foundation/CIPs/issues/937#issuecomment-2476061813 — and that CIP editors will work on way to avoid possible conflicts between social and technical propositions in the same scope as per https://github.com/cardano-foundation/CIPs/issues/937#issuecomment-2575800139.

Therefore we also agreed to take these down to `Draft` review status, to allow them to live on the CIP repository until they can be migrated to such a new (`CGP`?) repository, and will leave it to the proposal authors to close these PRs at that time (or earlier, at their discretion).

PR #943, opened 2024-12-03
Terminada-CPS-block-delay-centralisation

TerminadaDrep:
Please consider this CPS be added to the existing list of CIPs.

### DIFF
--- a/CPS-XXXX/README.md
+++ b/CPS-XXXX/README.md
@@ -1,0 +1,82 @@
+---
+CPS: ?
+Title: Ouroboros not accounting for typical global network delays causes centralisation
+Status: Open
+Category: ?
+Authors: Terminada
+Proposed Solutions: []
+Discussions: https://github.com/cardano-foundation/CIPs/pull/?
+Created: 2024-10-22
+License: CC-BY-4.0
+---
+
+# CPS-XXXX: Ouroboros not accounting for typical global network delays causes centralisation
+
+## Abstract
+An underlying assumption in the design of Cardano's Ouroboros protocol is that the probability of a stake pool being permitted to update the ledger is proportional to the relative stake of that pool.  However, the current implementation does not properly realise this design goal.
+
+Minor network delays endured by some participants cause them to face a greater number of fork battles.  The result is that more geographically decentralised participants do not obtain participation that is proportional to their relative stake.  This is both a fairness and security issue.
+
+## Problem
+The [Ouroboros Praos paper](<https://eprint.iacr.org/2017/573.pdf>) states that it is the "*probability of being permitted to issue a block*" that "is proportional to the relative stake a player has in the system".  However, it is clear from the security analysis section that it is the _probability of contributing to the ledger_ that is supposed to be proportional to the relative stake of the participant.  This is an important distinction because it matters little if a player has stake weighted permission to make blocks if the protocol later disproportionately drops that players blocks despite their honest performance.
+
+With the current Ouroboros implementation where slots have 1 second duration, it is not uncommon to have consecutive slot leaders that are only 1 second apart.  Remote participants may be unable to receive the previous block within 1 second and so these block producers will make blocks that result in forks.  Ouroboros settles such forks by preferring the fork whose terminal block has the lowest VRF value.  This seems like a fair method because it is deterministic, neither player can alter the outcome, and each player has an equal chance of winning.  However, the problem is that the more geographically decentralised players will face a disproportionately greater number of "fork battles".  This in turn means that they will get more of their blocks dropped.  The net result is that more geographically decentralised participants do not receive stake weighted _probability of contributing to the ledger_.  This is not only unfair, but also represents a deviation from the Ouroboros security analysis assumptions.
+
+This might seem like a minor problem, but the effect is significant.  If the majority of the network reside in USA - Europe with close connectivity and less than 1 second propagation delays, then those participant pools will see 5% of their blocks suffering "fork battles" which will only occur when another pool is awarded the exact same slot (ie: a "slot battle").  They will lose half of these battles on average causing 2.5% of their blocks to get dropped, or "orphaned".
+
+However, for a pool that happens to reside on the other side of the world where network delays might be just over 1 second, this pool will suffer "fork battles" not only with pools awarded the same slot, but also the slot before, and the slot after.  In other words, this geographically decentralised pool will suffer 3 times the number of slot battles, amounting to 15% of its blocks, and resulting in 7.5% of its blocks getting dropped.  The numbers are even worse for a pool suffering 2 second network delays because it will suffer 5 times the number of "fork battles" and see 12.5% of its blocks "orphaned".  This not only results in an unfair reduction in rewards, but also the same magnitude reduction in contribution to the ledger.
+
+Even the high quality infrastructure of a first world country like Australia is not enough to reliably overcome this problem due to its geographical location.  But is it reasonable to expect all block producers across the world to receive blocks in under one second whenever the internet becomes congested, or if block size is increased following parameter changes?  Unfortunately, the penalty for a block producer that cannot sustain this remarkable feat of less than 1 second block receipt and propagation, is 3 times as many "fork battles" resulting in 7.5% "orphaned" blocks rather than 2.5%.
+
+Considering that most stake pools are competing over 1% or less in fees, these are big numbers.  The obvious solution for the remote pool is to move its block producer to a server housed in USA or Europe.  This illustrates not only the centralisation problem created, but also the reduction in security that flows from running a block producer on someone elses computing hardware.
+
+## Goals
+Cardano should live up to its [11 blockchain tenets](<https://iohk.io/en/blog/posts/2024/10/11/the-11-blockchain-tenets-towards-a-blockchain-bill-of-rights/>) proposed by Prof Aggelos Kiayias, in particular T8 and T11 which speak to treating participants fairly without asymmetries.
+
+## Possible Solutions
+1. Modify how stake pools calculate their slot leadership by including ```mod 2``` or ```mod 3``` in the formula so that only every second or third slot is a possible leader slot.  Then adjust other parameters so that the Cardano network still produces a block every 20 seconds on average.
+
+    A consequence of this change would be an increased number of true "slot battles", where two pools are awarded the exact same slot.  However such battles are fairly settled by preferring the lowest block VRF and cannot be gamed by a malicious actor.
+
+2. Increase the slot duration to 2 or 3 seconds.  However this could have consequences for dapps that have assumed a slot will always be 1 second.
+
+## Considerations
+1. What is a reasonable expectation for block producers housed across the world to achieve in terms of network delays?  Is 2 seconds of network delay enough time before block producers get penalised, or would 3 seconds be more appropriate?
+
+2. What internet infrastructure and geographical location does Cardano consider as a minimum requirement in order to fairly contribute to its ledger?
+
+3. It would seem to be an advantage to encourage fair participation from people residing in countries on the other side of the globe because this might provide resilience against political or other attacks localised to Europe or USA.
+
+4. Would there be any impact on the Ouroboros 5 second delta parameter?
+
+5. Would it be appropriate to make the window in which the block VRF tie breaker rule is applied also the same 2 or 3 seconds so that only blocks with the exact same slot number are deterministicly settled using the block VRF, otherwise the node would prefer its current block?  Such a change would remove the ability of malicious actors to deliberately cause a fork with the previous block when they know their block VRF is lower.
+
+6. What effect could any change have on the solution to [issue #2913: Consensus should favor expected slot height to ward off delay attack](<https://github.com/IntersectMBO/ouroboros-network/issues/2913>)?
+
+## Arguments against correcting this unfairness
+1. It doesn't matter where the block producer is warehoused because block production is like a virtual service that can be run from anywhere.  What really matters is ownership of the pledge and stake, not ownership of the computing hardware.  If you live on the other side of the world, just rent a virtual server in Frankfurt or Los Angeles for your block producer.
+
+- Centralising Cardano infrastructure to data centres potentially hands control over the software, as well as selective control over block propagation between nodes, to BigTech data centre owners.
+
+2. The internet infrastructure is centred in USA and USA is more politically stable and less likely to have its internet infrastructure compromised through acts of war.  If conflict between USA, China, and Russia ensues then the undersea cables to Japan, Australia and New Zealand could get damaged.  Therefore it makes sense that Cardano block production should be slightly advantaged in USA and slightly disadvantaged in Japan, Australia and New Zealand.
+
+- Japan, Australia and New Zealand are very politically stable and it might actually be a good idea to _fairly_ incentivise participation from people living in those areas.  If internet connectivity was to be affected by cyber warfare, the targets of such attacks may not be predictable today.
+
+- The bottom line is that if I have 0.001% of stake in the system then I should get 0.001% of access to the system.  Likewise with scaling that up, if good actors possess 51% of the stake then they should have 51% of the control, as this is a fundamental assumption protecting Cardano.  I hope that the Cardano community won't step away from this fundamental mathematics based approach in favour of some sort of subjective human inference about where they think block producers should be warehoused for political or other reasons.
+
+3. Internet transmission is improving so in the future 1 second might be sufficient for block propagation across the entire globe.
+
+- This unfairness problem doesn't seem hard to fix.  If / when internet speeds improve then it should be straightforward to recalibrate the software again.
+
+4. Won't halving or one-third'ing the number of potential leader slots reduce the number of blocks and therefore reduce the throughput?
+
+- Possible solution 1 also involves adjusting other parameters so that the target rate of block production remains unchanged.  Actually, it seems likely that the realised throughput would slightly increase.  The reason for this is that [possible solution 1](#Possible-Solutions) would eliminate forks caused by 1 second propagation delays which currently are wasted throughput.
+
+5. If my pool is located close to the majority then I am benefitting from this unfairness so why should I vote for this to be fixed?
+
+- See section: [Goals](#Goals).  Hopefully most people in Cardano want to build a fair system for everyone no matter where they live.
+
+## Copyright
+This CIP is licensed under [CC-BY 4.0](https://creativecommons.org/licenses/by/4.0/legalcode).
+
+****


CPS-XXXX/README.md
```diff
@@ -0,0 +1,82 @@
+---
+CPS: ?
+Title: Ouroboros not accounting for typical global network delays causes centralisation
```
rphair, line 3:
```suggestion
Title: Block Delay Centralisation
```
The originally presented title is a "motivation" rather than a title & we need to be more concise here to blend in with other proposals & problem statements... it's not just about fitting into smaller space in a table (though it helps).

The title you put into your branch, which became the initial title for this PR, is "just right" in my opinion: neither too general nor too specific.  I would have been tempted to say `Geographical` centralisation if not for the fact that the same problem (and same solution) you are talking about can also be caused by variations in network media rather than asymmetry in network topology.

CPS-XXXX/README.md
```diff
@@ -0,0 +1,82 @@
...
+Status: Open
+Category: ?
```
rphair, line 5:
```suggestion
Category: Consensus
```
As I understand this is a recommendation for the `Consensus` protocol... please @bwbush correct me if I'm wrong about this (e.g. if in fact this is a challenge for the `Network` layer: cc @jpraynaud).

CPS-XXXX/README.md
```diff
@@ -0,0 +1,82 @@
...
+Authors: Terminada
```
rphair, line 6:
```suggestion
Authors:
  - Terminada
```
In our standard format, the content of fields that might contain multiple values must either be an empty set or an unordered list.

CPS-XXXX/README.md
```diff
@@ -0,0 +1,82 @@
...
+Proposed Solutions: []
+Discussions: https://github.com/cardano-foundation/CIPs/pull/?
```
rphair, line 8:
```suggestion
Discussions:
  - https://github.com/cardano-foundation/CIPs/pull/943
```

CPS-XXXX/README.md
```diff
@@ -0,0 +1,82 @@
...
+Created: 2024-10-22
+License: CC-BY-4.0
+---
+
+# CPS-XXXX: Ouroboros not accounting for typical global network delays causes centralisation
+
```
rphair, line 14:
```suggestion
```
Titles for these documents are always generated from the tabular data in front & aren't part of the document itself.

CPS-XXXX/README.md
```diff
@@ -0,0 +1,82 @@
...
+- This unfairness problem doesn't seem hard to fix.  If / when internet speeds improve then it should be straightforward to recalibrate the software again.
+
+4. Won't halving or one-third'ing the number of potential leader slots reduce the number of blocks and therefore reduce the throughput?
+
+- Possible solution 1 also involves adjusting other parameters so that the target rate of block production remains unchanged.  Actually, it seems likely that the realised throughput would slightly increase.  The reason for this is that [possible solution 1](#Possible-Solutions) would eliminate forks caused by 1 second propagation delays which currently are wasted throughput.
+
+5. If my pool is located close to the majority then I am benefitting from this unfairness so why should I vote for this to be fixed?
+
+- See section: [Goals](#Goals).  Hopefully most people in Cardano want to build a fair system for everyone no matter where they live.
```
rphair, line 77:
```suggestion
### Arguments against correcting this unfairness
1. It doesn't matter where the block producer is warehoused because block production is like a virtual service that can be run from anywhere.  What really matters is ownership of the pledge and stake, not ownership of the computing hardware.  If you live on the other side of the world, just rent a virtual server in Frankfurt or Los Angeles for your block producer.
  - Centralising Cardano infrastructure to data centres potentially hands control over the software, as well as selective control over block propagation between nodes, to BigTech data centre owners.

2. The internet infrastructure is centred in USA and USA is more politically stable and less likely to have its internet infrastructure compromised through acts of war.  If conflict between USA, China, and Russia ensues then the undersea cables to Japan, Australia and New Zealand could get damaged.  Therefore it makes sense that Cardano block production should be slightly advantaged in USA and slightly disadvantaged in Japan, Australia and New Zealand.
  - Japan, Australia and New Zealand are very politically stable and it might actually be a good idea to _fairly_ incentivise participation from people living in those areas.  If internet connectivity was to be affected by cyber warfare, the targets of such attacks may not be predictable today.
  - The bottom line is that if I have 0.001% of stake in the system then I should get 0.001% of access to the system.  Likewise with scaling that up, if good actors possess 51% of the stake then they should have 51% of the control, as this is a fundamental assumption protecting Cardano.  I hope that the Cardano community won't step away from this fundamental mathematics based approach in favour of some sort of subjective human inference about where they think block producers should be warehoused for political or other reasons.

3. Internet transmission is improving so in the future 1 second might be sufficient for block propagation across the entire globe.
  - This unfairness problem doesn't seem hard to fix.  If / when internet speeds improve then it should be straightforward to recalibrate the software again.

4. Won't halving or one-third'ing the number of potential leader slots reduce the number of blocks and therefore reduce the throughput?
  - Possible solution 1 also involves adjusting other parameters so that the target rate of block production remains unchanged.  Actually, it seems likely that the realised throughput would slightly increase.  The reason for this is that [possible solution 1](#Possible-Solutions) would eliminate forks caused by 1 second propagation delays which currently are wasted throughput.

5. If my pool is located close to the majority then I am benefitting from this unfairness so why should I vote for this to be fixed?
  - See section: [Goals](#Goals).  Hopefully most people in Cardano want to build a fair system for everyone no matter where they live.
```
To the extent that you are keeping this content in such a block, making these changes will indent it properly (after demoting any non-standard heading to a subheading of `###` or less).

rphair (commented):
@TerminadaDrep thanks for getting this into the open after putting so much work into this issue & led so much constructive discussion about it in the last couple of years.  I'm marking this `Triage` to introduce it at the CIP meeting in a week's time (https://hackmd.io/@cip-editors/102) & you would be very welcome to attend if possible & field some initial questions.

CPS-XXXX/README.md
rphair: p.p.s. my changes don't have any effect on the formatting: my apologies for the unnecessary change (your document reads the same either way).

CPS-XXXX/README.md
```diff
@@ -0,0 +1,76 @@
+---
+CPS: ?
+Title: Ouroboros not accounting for typical global network delays causes centralisation
+Status: Open
+Category: ?
+Authors:
+  - Terminada
+Proposed Solutions: []
+Discussions:
+  - https://github.com/cardano-foundation/CIPs/pull/943
```
rphair, line 10:
```suggestion
  - https://github.com/cardano-foundation/CIPs/pull/943
  - https://forum.cardano.org/t/problem-with-increasing-blocksize-or-processing-requirements/140044
```
This field is also intended for community discussions, if technically significant, so feel free to add this link if there is any ensuing technical discussion there.

CPS-XXXX/README.md
```diff
@@ -0,0 +1,76 @@
...
+
+However, for a pool that happens to reside on the other side of the world where network delays might be just over 1 second, this pool will suffer "fork battles" not only with pools awarded the same slot, but also the slot before, and the slot after.  In other words, this geographically decentralised pool will suffer 3 times the number of slot battles, amounting to 15% of its blocks, and resulting in 7.5% of its blocks getting dropped.  The numbers are even worse for a pool suffering 2 second network delays because it will suffer 5 times the number of "fork battles" and see 12.5% of its blocks "orphaned".  This not only results in an unfair reduction in rewards, but also the same magnitude reduction in contribution to the ledger.
```
SmaugPool, line 27: This CPS does not provide data showing that 1s is not enough for blocks to propagate anywhere in the world with required hardware, connection and configuration. Without such data, it's impossible to determine if it's needed or not, and what solution would work.

CPS-XXXX/README.md
```diff
@@ -0,0 +1,76 @@
...
+
+Even the high quality infrastructure of a first world country like Australia is not enough to reliably overcome this problem due to its geographical location.  But is it reasonable to expect all block producers across the world to receive blocks in under one second whenever the internet becomes congested, or if block size is increased following parameter changes?  Unfortunately, the penalty for a block producer that cannot sustain this remarkable feat of less than 1 second block receipt and propagation, is 3 times as many "fork battles" resulting in 7.5% "orphaned" blocks rather than 2.5%.
```
SmaugPool, line 29: Some data is needed to prove the Australia's case and to be able to reproduce it and evaluate working solutions.

SmaugPool: For example this AWS datacenter to datacenter round trip latency map does not seem to be enough to prove the point:
image

CPS-XXXX/README.md
rphair: I often get 1000ms ping times just crossing India's border: though AWS Mumbai is probably exempt from such delays, which I imagine result from "great firewall" type packet inspection from the newly founded & somewhat ill-equipped surveillance state here.

TerminadaDrep: This block was produced by my BP earlier today. It was full at 86.57kB in size, containing 64 transactions, and 66.17kB of scripts: https://cexplorer.io/block/c740f9ce8b25410ddb938ff8c42e12738c18b7fd040ae5224c53fb45f04b3ba0

These are the delays (from beginning of the slot) before each of _my own relays_ included this block in their chains:

- Relay 1 (ARM on same LAN) → Delayed=0.263s
- Relay 2 (AMD on adjacent LAN) → Delayed=0.243s
- Relay 3 (ARM approx 5 Km away) → Delayed=0.288s
- Relay 4 (AMD Contabo vps in USA) → Delayed=2.682s
- Relay 5 (ARM Netcup vps in USA) → Delayed=1.523s

The average propagation delay by nodes pinging the data to pooltool was 1.67 seconds: https://pooltool.io/realtime/11169975

fallen-icarus: @TerminadaDrep Could you add the above delay metrics to the CPS? I think having empirical data would help strengthen the case of this CPS. Also, could you indicate whether your BP is locally controlled or in a vps? I'm guessing it is locally controlled.

CPS-XXXX/README.md
fallen-icarus: I agree with data helping, but I want to point out that using AWS as the benchmark doesn't seem appropriate since the goal is *not* to have AWS control most of the block producers.

P.S. I'm not suggesting you mean to use AWS as the benchmark, I just felt like this point should be made explicit.

rphair: Also we can't be sure these hop times aren't between AWS back-end networks: and therefore not inclusive of times to enter & exit backbone networks or the "last mile" of retail Internet services.

SmaugPool: > I agree with data helping, but I want to point out that using AWS as the benchmark doesn't seem appropriate since the goal is to _not_ have AWS control most of the block producers.
> 
> P.S. I'm not suggesting you mean to use AWS as the benchmark, I just felt like this point should be made explicit.

The point was that if AWS datacenter to datacenter round trip latency was already more than 1s from 2 points in the world, it would have been very likely enough to prove the CPS point. But that's not the case, so more data is needed. I didn't mean to say anything else, you are interpreting, and I think it's appropriate to show that more data is indeed needed. See my original quote:

> For example this AWS datacenter to datacenter round trip latency map does not seem to be enough to prove the point:

happystaking:
I can only speak for my own pool, but when I compare a high spec relay (A) to a low spec relay (B) the average time from slot begin to adoption on relays A and B is 240ms and 560ms respectively. 

Blocks are forged on node C which is on the same physical machine as relay A. Relay B is 5500km (~100ms) away, so that time should be subtracted from the 560ms to take network delay (same LAN, no hops) out of the equation.

This (perhaps unscientific) method leads me to believe that having the right hardware does more to improve propagation times than being close to all other nodes. Moreover pinging halfway around the world should take no more than 300ms in an optimal situation. That means you should easily have enough time to bypass any countries or areas with deep packet inspection (which cause high network delays) and still stay under 1 second.

CPS-XXXX/README.md
gufmar: Here's a visualisation of the block mentioned by terminada before produced by his TERM BP. I assume located in Australia. 
 
Screenshot_2024-12-03-13-16-57-65_40deb401b9ffe8e1df2f1cc5ba480b12

It's all mobile screenshots only, as I'm out of office this week.

But I tried to show the propagation experience as reported from many nodes around the globe (above screenshot shows only half of them). You can see almost all nodes take already a second to only hear about the existence of the new block, by receiving the header announcement. 
Only then the fetch request and body download starts.
And only as the last (blue) step the local validation happens.

What needs to be noted here: I sorted these reporting nodes by continent. First row shows a relay node from Sidney Australia. And this node already has a long waiting time until he heard about the header. 

This is also the case for the Asian nodes at the end of the list 

Screenshot_2024-12-03-13-21-18-59_40deb401b9ffe8e1df2f1cc5ba480b12

CPS-XXXX/README.md
gufmar: Unfortunately not this week, because involved in other things, but I can assure you, we have plenty of latency and propagation data. Not only general latency but actual mainnet block propagation times as transmitted and received via Ouroboros mini protocols.
And we have it for 2.5 years of history, covering a bunch of different normal and extraordinary network situations. For small and large blocks, with no, up to max script execution units.
The previously showed gantt chart here is just one visualization for one block. I'm happy to invite to a workshop call where we go through some of these data points, computed averages for predefined edge cases etc.

CPS-XXXX/README.md
TerminadaDrep: It is interesting that the delay is clearly in the Australian part of the internet.  Perhaps the Aussie national broadband network (NBN) was more congested than usual at this time.

Certainly this block had worse than usual propagation delays.

fallen-icarus: >  So I also disagree with one of the points in the CPS, in expectations of future transmission and latency improvements.

@gufmar I'm not an expert on networks, but I really don't think we should be relying on improvements to network throughput here due to the [rebound effect](https://en.wikipedia.org/wiki/Rebound_effect_(conservation)). Demand will very likely increase to use up any extra "slack" (eg, Leios, block size increases, and even non-blockchain demand). If network capacity doubles but so does demand, we can easily find ourselves here again.

gufmar: Yes sorry if I was unclear. I wanted to say I disagree with the argument made at https://github.com/cardano-foundation/CIPs/blob/e7bf9b4c103f3841f2d8364e78905c1183ee9526/CPS-XXXX/README.md#arguments-against-correcting-this-unfairness

because I don't expect we will see significant improvements in network latency which is the main limiting factor here with the TCP ACK back and forth packets, not so much the Throughput.

TerminadaDrep:
Coincidentally, the last two TERM blocks in a row happened to have a leader for the next slot.

1. [A full block which pooltool reported avg propagation time of 0.87s](<https://pooltool.io/realtime/11176750>) --> But despite the average reported propagation being less than 1 second, the next producer IOGP did not receive it in time and created a fork.  Unfortunately IOGP's block had the lower VRF so TERM lost the "fork battle" and got its block orphaned.
2. [A small block which pooltool reported avg propagation time of 0.62s](<https://pooltool.io/realtime/11179133>) --> Which fortunately was received by the next producer TLK in time to produce its block at the next slot.  So there was no fork and TERMs block did contribute to the chain.

TerminadaDrep:
Another important consideration is that it is possible to maliciously game these forks.

The block VRF only depends on the following inputs:
- Epoch nonce
- Slot number
- Pool private key

Therefore the block VRF is known ahead of time.

A malicious operator can run a modified version of cardano-node that inspects the previous block VRF, compares this to its own value, and decide whether to deliberately cause a fork or not if it knows it will win the "fork battle".  This would allow a malicious group of pools to deliberately "orphan" blocks of other competitors in order to earn a higher percentage of the reward pot, and gain more control over consensus.

kiwipool:
Hi

Here's some data from our onsite (our own site), NZ-based baremetal operation for comparison purposes. With the exception of perhaps 46S stakepool based in Invercargill we are likely to be the most remote stakepool in the ecosystem. We are never going to win any low latency awards from our location 

We run high-specification enterprize grade servers connected via gigabit fibre for our NZ relays and NZ primary BPs. In addition to NZ-based baremetal we operate cloud relays spread around the world with reputable providers on decent hardware. Our 'Plan B' cloud-based failover system is also very high specification and produces significantly lower latency numbers than our NZ-based baremetal. We choose to run our primary system on baremetal in NZ for philosophical, rather than performance, reasons.

See the below summarized pooltool data for the last 50 epochs (E476-525) for KIWI 
ID: 60397646d7d1ad6fe2ddccfe7efc9cba61f6d3d94d29e8f41de73240

Slots: 1,699
Height Battles: 16
Height Battle Wins: 4
Height Battle Losses: 12
Height Battles as % of Slots: 0.94%
Height Battle Wins as % of Slots: 0.24%
Height Battle Losses as % of Slots: 0.71%
Height Battle Wins as % of Height Battles: 25.00%
Height Battle Losses as % of Height Battles: 75.00%
Average Height Battles per Epoch: 0.32
Average Height Battle Wins per Epoch: 0.08
Average Height Battle Losses per Epoch: 0.24

Anecdotally we appear to be experiencing more height battles in the dynamic-p2p era

Hopefully this is useful for comparison purposes.

Matticus
🥝Kiwipool Staking

TerminadaDrep:
Is it reasonable to expect that geographically remote pools must use high QoS guaranteed priority fibre plans whereas the majority in USA/EU can use the "normal" internet?  Or is Ouroboros expected to function fairly with everyone using the "normal" internet?

CPS-XXXX/README.md
```diff
@@ -0,0 +1,109 @@
+---
+CPS: ?
+Title: Block Delay Centralisation
+Status: Open
+Category: Consensus
```
rphair, line 5:
```suggestion
Category: Network
```

rphair (commented):
@TerminadaDrep we discussed this at today's CIP meeting on Discord.  The consensus was that

1. the title & scope need to be more specifically defined as "fairness" relative to geographical limitations according to the observations that you've pointed out;
2. the CPS must be targeted towards specific goals to alleviate those discrepancies.  

Here are the recommendations that came up:

• The current title `Block Delay Centralisation` is ambiguous and maybe not accurate (since the delays are at perimeter, not the centre, of the network topology).  We should agree upon a CPS title that states "fairness" as something currently difficult to achieve for nodes on the perimeter of the Cardano network.

• Given that a huge component of node propagation delays derive from the speed of light and a currently unchangeable TCP/IP stack, we need the CPS to be written so CIPs that fix whatever is fixable _in Cardano_ can be attached to your problem statement: otherwise it might as well be an issue in the node repository issue queue (see further below).

• Other than altering the consensus mechanism, the only optimisations Cardano can therefore make are to address network inefficiencies.  @colll78 indicates that perhaps a 20-fold improvement in network efficiency is forthcoming, and I believe this CPS should make it possible to link any such improvements on the core roadmap to the symptoms of the problem you identify.

• Extending the slot duration to accommodate network propagation times was seen as be a "brute force" solution that would have dramatic effects on the Consensus protocol, and therefore less promising than _identifying points of research_ for improvements in network performance (hence also the category change highlighted below).

• Therefore the reviewers concluded the CPS should focus on _what needs to be **investigated**, and then **remediated**, to improve fairness of Cardano PoS performance for nodes at the topological boundaries of the network_.

My own question was (since I know you've been working steadily on getting visibility for these issues for a couple of years): do you have any Consensus or Node repository issues about these so far?  If so, I think the issues + any responses would help to fill out the proposal along these lines... and should be linked in the CPS and/or the discussion here.

fallen-icarus:
> Extending the slot duration to accommodate network propagation times was seen as be a "brute force" solution that would have dramatic effects on the Consensus protocol, and therefore less promising than identifying points of research for improvements in network performance (hence also the category change highlighted below).

Apologies for missing the meeting, but increasing the slot duration should still be seriously considered despite any required changes. According to the discussion in the consensus working group on discord, the designers did discuss 1 second vs 2 second slot lengths but they didn't really have any hard evidence to prefer 2 seconds. So they went with 1 second since it caused fewer slot battles.

Now that we have actual geographical data, there may be hard evidence to prefer 2 seconds. I'm not saying other avenues shouldn't be explored, but the above wording makes it seem like the slot duration option was downplayed too much in the meeting.

rphair:
cross-referencing a vital post from @karknu on network options, limitations & potential workarounds here: https://forum.cardano.org/t/problem-with-increasing-blocksize-or-processing-requirements/140044/7

rphair:
@TerminadaDrep now that discussion in the above Cardano Forum post with @karknu has progressed (up through [here](https://forum.cardano.org/t/problem-with-increasing-blocksize-or-processing-requirements/140044/16?u=cosdpool)) about tunable network parameters and other optimisations, could you refine this CPS with the points for investigation that were suggested in https://github.com/cardano-foundation/CIPs/pull/943#pullrequestreview-2493269284 to possibility give the related development teams something to work with?

rphair:
@TerminadaDrep I'm marking this `Waiting for Author` specifically about the review points above, beginning here (https://github.com/cardano-foundation/CIPs/pull/943#pullrequestreview-2493269284) and also hoping the progress referred to in https://github.com/cardano-foundation/CIPs/pull/943#issuecomment-2575848769 can be incorporated into the CPS somehow.

If you plan to continue with this CPS submission please post what you're working on or waiting for and/or respond to that feedback above.  I believe this is the community's best effort so far to qualify & quantify the targeted issues so I'm hoping to see work on this continue if you are able.

rphair:
@TerminadaDrep from public & private conversations my impression is that the problem has improved in latter versions of the node & that your successful efforts to get some analytical insight from core developers have better defined the remaining research & potential improvement areas by other means.

Please let us know whether you want to continue by the CPS route... since that would involve a deep dive into those findings made, since you posted this CPS, to be added to the CPS itself... and we will go back to leaving this tagged `Waiting for Author` instead.  In either case a response from you would be welcome.  cc @TerminadaPool

TerminadaPool:
Over 3 months ago the following changes were made to my stake pool setup:

1. On all relays and BP changed the Linux TCP configuration settings as follows:
  - InitialCongestionWindow=32, subsequently increased to =64
  - InitialAdvertisedReceiveWindow=32, subsequently increased to =64
  - net.ipv4.tcp_slow_start_after_idle=0
  - See: https://forum.cardano.org/t/problem-with-increasing-blocksize-or-processing-requirements/140044/12?u=terminada for further explanation.  Basically these TCP settings attempt to pipe the entire block (Cardano max block size=88KBytes) in one go without requiring intermediary Ack confirmation packets which incur round-trip transmission delays.  Such round-trip Ack confirmations are particularly time expensive when transmitting data from one side of the world to the other.
2. CardanoFoundation added one of my Australian relays as a localRoot to their relay warehoused in Australia.
3. Over the past 6 months, my pool has received more delegated stake (now around 20M Ada, up from 3M).  Through the P2P mechanism this will result in a greater number of other Cardano relays attempting to maintain connections with my registered relays.  Having more relays from other pools connected will likely result in blocks getting distributed more quickly to/from my pool.

The combination of these changes does anecdotally appear to have modestly reduced my pool's block transmission delays to/from the rest of the Cardano network.  However, the perceived improvement has been difficult to isolate from the simultaneous reduction in Cardano network load over the same period of time.

Regarding Cardano network load:  My pool has produced some 300 or so blocks over the past 3 months but literally none of these blocks has been completely full (Cardano max blocksize 88KBytes).  From memory, my pool has produced only 1 block that was around 80KBytes in size, but the vast majority of my pool's blocks have been nearly empty or less than 10-20KBytes in size.  In other words, these TCP setting changes haven't really been tested properly on full blocks.

Please correct me if I have some of my logic incorrect as I am not a TCP expert:
It is my understanding that my relays only control one side of the TCP window negotiation.  The other side is controlled by the other Cardano relay that is downloading my block.  If that other relay doesn't have similar TCP settings then it will request that the TCP window be smaller and therefore will require more round-trip time delay Ack packets to completely download the entire block.  Another consideration is that with longer network links there is a greater possibility for bit errors to occur during transmission and therefore if the TCP window is too large then this could result in a greater number of TCP packets requiring re-transmission.

I guess, there are only so many things that can be done to mitigate transmission delays and a pool on the other side of the world will necessarily be closer to the "1 second cliff".  As soon as that 1 second block delay threshold is crossed the number of fork battles for that pool increases from 5% to 15%.  I regularly monitor block delays and even with the Cardano chain minimally loaded at present, I still see occasional blocks delays to my BP in Australia that are longer than 1 second.

I apologise for holding off on collating the information on my previous few hundred blocks because I had been waiting to see if my pool would mint some full blocks.  As noted above, unfortunately small and empty blocks are not as likely to expose the problem outlined by this CPS.

Regarding fork battles caused by block delay of 1 second or more:
Since the above configuration changes, which as noted have incidentally coincided with a reduction in Cardano chain load, my pool hasn't experienced a single fork battle _across slots_ (IE: fork battle caused by delay in block receipt by 1 second or more).  And, my pool has produced some 300 or so blocks during this time.  My pool has of course had many "slot battles" (where the slot is identical), in line with an expected 5% true "slot battle" rate.

rphair:
thanks @TerminadaPool for the summary of your further research which I've at least partially observed has been delivered into the hands of the node & network engineers in the appropriate issue queues.  Your original writeup and further comments (some on the Cardano Forum instead) give node & pool administrators a lot to work with for their own diagnostics and troubleshooting.

Since there's no confirmation in your last comment that you intend to pursue the CPS — with no indication so far how this problem statement could lead to any CIP solutions — the only remaining way forward is to close this PR as inadmissible into the CIP repository.

Still I believe the information posted thus far might serve as a point of reference while verifying that these findings are all posted & updated in the issue queues of the node & network GitHub repositories. 🙏

PR #951, opened 2024-12-16
CIP-0108 | Ensure "CIP-0008" is in capital letters

### LINKED ISSUE #949: CIP-0108 - Inconsistent field name
I've noticed when addressing https://github.com/IntersectMBO/cardano-cli/issues/906 that, while the schema identifies the hash algorithm [here](https://github.com/cardano-foundation/CIPs/blob/d08e2671148a4f970ad9dab6ea82c9e94ddf32af/CIP-0108/cip-0108.common.schema.json#L55) as "CIP-0008" (consistently with the [CIP-0108](https://github.com/cardano-foundation/CIPs/tree/master/CIP-0108) text), in [this example](https://github.com/cardano-foundation/CIPs/blob/d08e2671148a4f970ad9dab6ea82c9e94ddf32af/CIP-0108/examples/no-confidence.jsonld#L70), it is written in lowercase ("cip-0008"). We probably should make this consistent. And I imagine we should favour "CIP-0008". Alternatively, if we want to allow both representations, we may want to update the schema to reflect that.

palas:
This PR addresses the issue https://github.com/cardano-foundation/CIPs/issues/949, by ensuring `CIP-0008` is used in capital letters in the `witnessAlgorithm` field of the `witness` type in CIP-0108, updating the "no-confidence" example and the test vector Markdown documentation.

Fixes #949.

### DIFF
--- a/CIP-0108/test-vector.md
+++ b/CIP-0108/test-vector.md
@@ -34,7 +34,7 @@ Blake2b-256 hash digest of canonicalized body: `68d6fe27087457acf0164e65414238c4
 ### Motion of No-Confidence
 
 Example metadata document file: [no-confidence.jsonld](./examples/no-confidence.jsonld).
-Blake2b-256 of the file content (to go onchain): `6c27e5bd0d7cdec7ddb30956be0b5eac892a8330e00689692d18f3815a71bf9f`
+Blake2b-256 of the file content (to go onchain): `87ba5c10c89484c52265b50d669b4acf116aaeb8a6fa35fbf06e5ec67cda9270`
 
 #### Intermediate files
 
--- a/CIP-0108/test-vector.md
+++ b/CIP-0108/test-vector.md
@@ -91,7 +91,7 @@ For [Treasury Withdrawal](#treasury-withdrawal), this will result in: `68d6fe270
 
 ### 4. Authors witness over the hash of canonicalized `body`
 
-Use the hash produced in [3.](#3-hash-the-canonicalized-body) as the payload for the witnessing. For a `witnessAlgorithm` of `ed25519` refer to [CIP-100 Hashing and Signatures](https://github.com/cardano-foundation/CIPs/blob/master/CIP-0100/README.md#hashing-and-signatures), for `cip-0008` refer to [CIP-108 New Witness Type](./README.md#new-witness-type).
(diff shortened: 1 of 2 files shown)


rphair (approved):
Can't think of any technical things that would fail because of this schema change but please @Ryun1 @Crypto2099 @perturbing @gitmachtl confirm if you can.

Ryun1:
Great spot @palas
ill reach out to tooling to double check that this change wouldnt break their implementations before we merge

CIP-0108/test-vector.md
```diff
@@ -34,7 +34,7 @@ Blake2b-256 hash digest of canonicalized body: `68d6fe27087457acf0164e65414238c4
 ### Motion of No-Confidence
 
 Example metadata document file: [no-confidence.jsonld](./examples/no-confidence.jsonld).
-Blake2b-256 of the file content (to go onchain): `6c27e5bd0d7cdec7ddb30956be0b5eac892a8330e00689692d18f3815a71bf9f`
+Blake2b-256 of the file content (to go onchain): `87ba5c10c89484c52265b50d669b4acf116aaeb8a6fa35fbf06e5ec67cda9270`
```
Ryun1, line 37: I pushed this update, because of the change introduced in no-confidence example changed the hash digest

Ryun1:
Spoke to a couple implementors, this change is all good to go.
Fortunately because governance actions are very expensive on mainnet, there isnt any examples of this on mainnet

Ryun1 (approved):
Thanks @palas 💪

gitmachtl:
Why do we use the public key for CIP-0008 method here and not the Cose_Key? 

Is there a written definition on the parameters used for this signature? Because CIP0008 allows for different ones. I guess we're going with the 
```
  kty (1) - must be set to OKP (1)
  kid (2) - Optional, if present must be set to the same value as in the Sig_structures protectedHeader_cbor via map entry (4)
  alg (3) - must be set to EdDSA (-8)
  crv (-1) - must be set to Ed25519 (6)
  x (-2) - must be set to the public key bytes of the key used to sign the Sig_structure
```
one?

If so, that should be specified in CIP108 too imo.

Ryun1:
@gitmachtl
the motivation for this inclusion in CIP8 is so that these signatures can be created over CIP30 connections, so using light wallets

gitmachtl:
@Ryun1 @palas the example file
https://github.com/cardano-foundation/CIPs/blob/72d499ffa0d6116c41b46276946bb5dc03f53165/CIP-0108/examples/no-confidence.jsonld
is missing the name field for the author

palas:
> is missing the name field for the author

@gitmachtl. It doesn't seem the `name` field is mandatory. CIP-0100 says "Each object MAY have a name property"

PR #942, opened 2024-11-21
CPS-0019? | Light Client Protocols

cleanerm5:
Hi,

There are a few activities ongoing that require the development of so called light client protocols for Cardano. The ones that came to my awareness are:
- the IBC implementation of CF (https://github.com/cardano-foundation/cardano-ibc-incubator) 
- the Bitcoin-Cardano bridge proposed by BitcoinOS and with the general approach described in this paper (https://assets-global.website-files.com/661e3b1622f7c56970b07a4c/662a7a89ce097389c876db57_BitSNARK__Grail.pdf) 
- the IOG research work on trust-minimized light client based wallets

There are various ways how to implement light clients on Cardano (whereas because of the core protocol not providing so called state proofs out of the box, there is no canonical light client design available for Cardano at the moment). In that sense, I wanted to create a common point of reference and problem description with this CPS to which concrete light client designs can refer to when they are documented as a CIP.

At least for the IBC implementation we plan to document the Mithril based light client as a CIP and the same would make sense for the respective light client protocol to be developed for a Bitcoin-Cardano bridge (in case the bridge follows a light client based security model), because both designs can be used for other use cases as well.

Any feedback is highly appreciated.

Kind regards

---

([rendered latest version](https://github.com/cleanerm5/CIPs/blob/master/CPS-%3F%3F%3F%3F/README.md))

PR #964, opened 2025-01-19
CIP-0178? | Merkle Root of Transactions in Block Header

aslesarenko:
Minotaur, a sub-project of Partner Chains, utilizes [light clients](https://github.com/input-output-hk/minotaur-connectors/pull/38) to follow state changes on integrated blockchains through a trust-minimized state bridge.

Previously, we implemented a proof of concept for Ethereum, using its existing features to enable this design. Our goal was to replicate a similar approach for Cardano using [Mithril](https://mithril.network/doc/). However, both Cardano and Mithril currently lack certain features necessary for achieving parity with the Ethereum implementation.

This ~~CPS~~CIP describes the missing component on the Cardano side.

---

([rendered proposal](https://github.com/aslesarenko/CIPs/tree/cps-cardano-header-hash/CIP-MPT-BLOCKHEADER/README.md))


Ryun1:
Hey @aslesarenko 

Thanks for this proposal, I really enjoyed reading it.

One high level scoping comment to add;

Is using merkle trees within block header the only solution to this problem?
I feel that this CPS may be re-scoped partly, to focus solely on the problem of creating compact transaction inclusion proofs
Then the 'solution CIP' can propose a specification of merkle proofs in block header.

CPS-xxxx/README.md
```diff
@@ -0,0 +1,108 @@
...
+Status: Open
+Category: Ledger
+Authors:
+  - Alexander Slesarenko <alex.slesarenko@iohk.io>
+Proposed Solutions: []
+Discussions:
+Created: 2025-01-17
+License: CC-BY-4.0
+---
+```
```
rphair, line 14:
```suggestion
---
CPS: ????
Title: Missing Merkle Root of Transactions in Cardano Block Header
Status: Open
Category: Ledger
Authors:
  - Alexander Slesarenko <alex.slesarenko@iohk.io>
Proposed Solutions: []
Discussions:
Created: 2025-01-17
License: CC-BY-4.0
---
```
This is a text block with YAML syntax, not a YAML code block.  The code tags need to be removed for it to be parsed correctly: both by GitHub and downstream filters / renderers.

rphair (commented):
coming up for introduction (not detailed review) at biweekly CIP meeting in a half-hour's time - @aslesarenko you are welcome to attend: https://hackmd.io/@cip-editors/104

aslesarenko:
> I feel that this CPS may be re-scoped partly, to focus solely on the problem of creating compact transaction inclusion proofs

@Ryun1, yes, this is the essence of what it will enable once implemented. And such proofs in turn will enable lots of use cases.

@rphair, just got to this message, too late for the meeting. BTW, I couldn't find the info on how to attend it (is it a scheduled Meets call?).

rphair:
@aslesarenko you can subscribe for notifications & scheduling here (https://lu.ma/cardano-cip) and follow this Discord invitation (https://discord.gg/J8sGdCuKhs) ... see [CIP Wiki > Key links & resources](https://github.com/cardano-foundation/CIPs/wiki#key-links--resources)

rphair (commented):
At today's CIP meeting we didn't have the opportunity to resolve the issue about "scope" in the last couple substantial comments above: since the content defines the problem (as per a CPS) but its title suggests a particular solution (as per a CIP solving that CPS).

I made the suggestion that only the title itself might be overly specific, and that the rest of the document roundly covers a number of solutions and should itself be considered a valid CPS already.

Rather than mark this `Unconfirmed` (as sounding too specific for a CPS) or `Confirmed` with the current title (perhaps requiring you to produce a CIP implementing the favoured solution), therefore I suggested:
- to **retitle** this PR & document with a more general title that covers all the implied solutions
- to leave this marked `Triage` to ensure we settle the scope & title questions no later than the next CIP meeting: where it will also be on the agenda & which you @aslesarenko are now subscribed to attend.

colll78:
Ergo has this (Merkle-root of transactions in block header), see [here](https://docs.ergoplatform.com/dev/scs/sigma/lang-spec/#data-types_1). This is incredibly useful for interoperability (especially the Bitcoin / Cardano BitVMX zero-knowledge bridge work), L2s (Midgard) and even state channels (Hydra). 

By having this on Cardano L1, it would allow Midgard instances to see / prove events that happened in other (whitelisted / known) Midgard instances via the Cardano L1 block information. This is an incredibly power feature that would allow us to design applications that are able to read, consume and react to events across the Cardano L1 (and all Cardano based L2s). Even more so, this allows us to easily create dApps that are interoperable with other dApps on the L1 across multiple transactions, (ie. An insurance protocol that automatically reinforces users' collateral positions if they are at risk of liquidations, or pays out an insurance claims if a token (ie. stablecoin) drops in value by X% over Y period of time).

rphair:
Following from @aslesarenko's introduction at tonight's CIP meeting and https://github.com/cardano-foundation/CIPs/pull/964#pullrequestreview-2565458152 with @Crypto2099's specific recommendations, we've agreed to give this a number pending the assignment of a more general title that doesn't overtly favour the "Merkle Root" specifically as the solution for the identified interoperability problem.

@aslesarenko we look forward to confirming a new title (maybe something simpler like "transaction inclusion proofs in block header") and then we can assign a number and change it to `Confirmed` status.

aslesarenko:
@rphair, abstracting from a concrete implementation the following key points nevertheless should be reflected in the title:
1. Header should contain short digest (hash) which authenticates transaction in the block body and cryptographically connects the header with the transactions.
2. Its presence should enable short inclusion proofs for any individual transaction of the block.

So, here are the variants, which don't assume any specific implementation approach:
- Header-Level Digest/Commitment to Enable Concise Inclusion Proofs
- Short Transaction Inclusion Proofs via Cryptographic Digests in Block Headers
- Missing Commitment in Block Header for Compact Transaction Inclusion Proofs
- Compact Transaction Inclusion Proofs Through a Header-Based Digest

rphair:
Thanks @aslesarenko for considering the naming issue.  The editors brainstormed — also considering the typical / maximum CIP world length you can observe [here](https://github.com/cardano-foundation/CIPs#cardano-improvement-proposals-cip) — and came up with (courtesy of @Ryun1):
- Lack of Concise Block Inclusion Proofs

This has the advantage of immediately getting the human reader's attention if they have encountered or considered this problem: whether centred upon Cardano or elsewhere.

aslesarenko:
@rphair I read "Block Inclusion Proofs" as "proofs that block is included somewhere" or "proofs about blocks", where as problem is to have "proofs about transactions" or "proof that transaction is included".

So, how about "Lack of Concise Transaction Inclusion Proofs"?

rphair:
@aslesarenko that's the same thing I was thinking.  Yes, let's use the latter 🙏

CPS-xxxx/README.md
```diff
@@ -0,0 +1,106 @@
+---
+CPS: ????
+Title: Missing Merkle Root of Transactions in Cardano Block Header
```
perturbing, line 3:
```suggestion
Title: Lack of Concise Transaction Inclusion Proofs
```

CPS-xxxx/README.md
```diff
@@ -0,0 +1,106 @@
...
+License: CC-BY-4.0
+---
+
+## Abstract
+
+Cardano’s block headers currently omit an explicit Merkle root (or equivalent data structure) for transactions. This omission prevents straightforward creation of compact transaction inclusion proofs, which are crucial for use cases that require trust-minimized verification. While existing solutions like Mithril can certify transaction data itself, they do not certify the header. Consequently, there is no proof that a transaction was included in a specific block at a specific time, unless the entire block is downloaded and verified. The absence of a transaction Merkle root in the header, in turn, reduces opportunities for flexible, lightweight proofs of transaction presence and position in time.
+
+## Problem
+
+Cardano headers do not expose a merkle root hash of all transactions included in a block.
```
perturbing, line 20: This problem statement already poses a solution, namely the addition of a data structure like a merkle tree to each block. 

I think what you are trying to abstract to is twofold:

1) The desire to proof if a block is onchain
2) Given a block and such a proof, easily extract data from it to prove usefull statements.

A usefull statement that you suggest is getting transactions, but this is just an example.

How much of this aligns with [this](https://github.com/cardano-foundation/CIPs/pull/942) CPS?

perturbing (commented):
I think this higly overlaps with [CPS-0019 ](https://github.com/cardano-foundation/CIPs/pull/942)

CPS-xxxx/README.md
rphair: committing from editor & @aslesarenko agreement from https://github.com/cardano-foundation/CIPs/pull/964#issuecomment-2639112833

CPS-xxxx/README.md
aslesarenko: > addition of a data structure like a merkle tree to each block
Nope, I only propose to add tree root hash to block header, the tree itself is not stored as it is not necessary. 

The tree root hash is different from a simple hash because it enables partial authentication of block content (with per transaction granularity, but other data can also be supported). 
In particular, having a full block it is possible to create a compact proof containing individual transaction + merkle path to the root hash. This path is the key, because it is compact and can be verified against what is stored in the header.

> How much of this aligns with https://github.com/cardano-foundation/CIPs/pull/942 CPS?

If implemented, this proposal will simplify implementation of both light clients for Cardano + many L2 dApps on top of Cardano.

colll78:
Please create a CIP for this instead of a CIP, and add the following:

Introduce a new field, `txInfoVerifiedRootHashes` to the `ScriptContext`, with the following type:
```haskell
txInfoVerifiedRootHashes :: [(TransactionRootHash, Maturity)] 
```
Where `Maturity` is an integer that represents how old the block corresponding to that transaction root hash is, ie. 1 would mean that it is the transaction root hash associated with the previous block. Alternatively, if this places too much of a burden on nodes (ie. querying the `TransactionRootHash` from 200 blocks ago) then we can instead just allow this list to contain up to 10 root hashes from the previous 10 blocks (ie. only allowed values for maturity would be 1-10). 

This provides smart contracts with access to the transaction root hashes associated with previous block, thus allows them to prove that some event occurred on the chain in the past ten blocks. This is extremely useful for sidechains, l2s, hydra, and general purpose DeFi protocols, and even governance (you can prove that a user voted on a proposal, delegated to a drep, or even delegated to a stake pool).

rphair:
> @colll78 [originally posted] Please create a CIP for this instead of a CIP

1. did you mean "instead of a **CPS**"?
2. why do you think "instead of" and not "in addition to"?  If I had to choose one or the other, I would prefer a direct solution over a CPS: but at our last CIP meeting there was general agreement there are a wider variety of approaches to the problem than the one currently favoured in the document.

ch1bo:
I'd like to second @perturbing and @colll78 in that this seems to be rather a starting point for a CIP than a CPS, because it basically is the concrete solution idea of including a merkle tree root hash of a block's transactions in the block header. There is a problem statement included, but this is also about enabling (or taking one step closer) to cryptographic light clients. The challenge of creating such light clients is described in more detail in [CPS-0019](https://github.com/cardano-foundation/CIPs/pull/942). Note though, that this might not be the only CIP needed to solve CPS-0019.

rphair:
@aslesarenko the CIP meeting in the last hour, after last meeting (2 weeks ago) inclined to agree with the expert review above suggesting that this submission be reformed as a CIP, has now established that as a consensus.  The material is strongly in favour of its most recommended solution, which makes much more sense than proposing roughly 5 parallel solutions when only adopting 1 of these would be a substantial, historically significant change.

So unless you point out a serious problem with this idea we would like you to please make the format & presentational changes to submit this as a CIP.  Please let's continue in this branch (i.e. don't open up a new PR) and try as best as you can to preserve the blocks of text now that several people have reviewed them & found them at least to be a good exposition of the approaches not taken (i.e. they would serve well in the _Rationale_).

Any other questions about recasting this as a CIP please ask them here ASAP.  As is always recommended on the CIP repository, incremental commits rather than a "force push" would be much appreciated.  I think once we see a properly formatted CIP for this material, editors would want to assign a candidate number: likely at the next CIP meeting (in 2 weeks' time) if ready by then.

aslesarenko:
@colll78 please review and make necessary corrections

rphair:
It sounds like the author agrees in recasting this as a CIP so I'll retitle the PR as such — to help keep review on track, with the title of the mainly proposed method (almost as originally submitted) — and will look forward to whatever rewrite @colll78 is able to provide that turns this into a CIP before its next review.

ch1bo:
Note that this CIP is similar to block header changes needed for Leios (and Peras) and we should keep it in mind when we sketch the block header changes for those.

rphair:
cc @colll78 on @ch1bo's https://github.com/cardano-foundation/CIPs/pull/964#issuecomment-2934826900 in hope the similarity to Leios / Peras constrains the forthcoming CIP so it can be rewritten more readily.

rphair:
@GeorgeFlerovsky tagging you for a quick review of this proposal's contents, based on @colll78's https://github.com/cardano-foundation/CIPs/pull/1110#issuecomment-3948705643.  This has effectively been abandoned by the author (and would be tagged this way soon if no further update) but it seems there is some substantial value to this proposal and I don't want it to get lost due to any incidental difficulties there may be.

What would it take to get this half-finalised draft (written half as a CPS before that apparent abandonment) and turn it with some update & advocacy into a viable CIP?  The editors (cc @Ryun1 @perturbing) will help with that in whatever way we can & would really want to get this back into proper review.

cc other participants: @aslesarenko @nhenin @ch1bo

colll78:
I have rewritten this as a CIP, hopefully someone else can take this forward (ex. the Mithril or light client team or anyone) I don't have capacity, but this is a necessary feature for the expansion of trustless and trust minimized interoperability between Cardano and other ecosystems.

CIP-MPT-BLOCKHEADER/README.md
```diff
@@ -0,0 +1,283 @@
+---
+CIP: TBD
+Title: Merkle Patricia Trie Root of Transactions in Cardano Block Header
```
rphair:
```suggestion
Title: Merkle Root of Transactions in Block Header
```
This is the level at which non-experts will be thinking about the concept & the particular data structure won't mean much to them: while the shorter title will emphasise the innovative part more effectively.  Also this is consistent with the PR title which will be familiar to reviewers from the past few months.

Even with the way current discussion of #1149 is going, I still think it's redundant to put the term `Cardano` in the title since it will be understood that compatibility with sidechains will always begin with the Cardano main chain by default.

CIP-MPT-BLOCKHEADER/README.md
```diff
@@ -0,0 +1,283 @@
...
+By committing to an MPT root in the header, Cardano blocks can support efficient **transaction
+inclusion proofs** (analogous to Bitcoin SPV proofs and Ethereum transaction/receipt proofs) without
+requiring full block body download. The MPT design is widely deployed in other blockchains (notably
+Ethereum), is compatible with nibble-based proof verification, and is amenable to compact proofs and
+incremental construction.
+
+This proposal specifies the header changes, CBOR encoding updates, consensus-critical computation of
+the root from the block’s transactions, and deployment via a Dijkstra-era hard fork.
+
+## Motivation: why is this CIP necessary?
```
rphair:
```suggestion
## Motivation: Why is this CIP necessary?
```
(emergent new format detail in the meantime since this was written)

CIP-MPT-BLOCKHEADER/README.md
```diff
@@ -0,0 +1,283 @@
...
+
+- A proof is a sequence of trie nodes (CBOR-encoded) along the path from the root to the leaf for
+  the transaction index key.
+- Verification recomputes hashes up the path and checks that the resulting root equals
+  `hbTxTrieRoot`, and that the leaf value equals the expected `TxId`.
+
+The on-wire/API encoding of proofs is not consensus-critical, but should be standardized by tooling
+to ensure interoperability.
+
+## Rationale: how does this CIP achieve its goals?
```
rphair:
```suggestion
## Rationale: How does this CIP achieve its goals?
```
(emergent new format detail in the meantime since this was written)

CIP-MPT-BLOCKHEADER/README.md
```diff
@@ -0,0 +1,283 @@
...
+CIP: TBD
+Title: Merkle Patricia Trie Root of Transactions in Cardano Block Header
+Category: Ledger
+Status: Proposed
+Authors:
+  - Alexander Slesarenko <alex.slesarenko@iohk.io>
+  - Philip DiSarro <philip.disarro@iohk.io>
+Implementors: []
+Discussions:
+  - https://github.com/cardano-foundation/CIPs/pull/964
```
rphair:
```suggestion
  - Original PR: https://github.com/cardano-foundation/CIPs/pull/964
```
New convention of labelling such as @Ryun1 is introducing, and also helps to answer question that (still might be) introduced early in the _Motivation_ (i.e. why this is a CIP & no longer a CPS).

CIP-MPT-BLOCKHEADER/README.md
```diff
@@ -0,0 +1,283 @@
...
+incremental construction.
+
+This proposal specifies the header changes, CBOR encoding updates, consensus-critical computation of
+the root from the block’s transactions, and deployment via a Dijkstra-era hard fork.
+
+## Motivation: why is this CIP necessary?
+
+> NOTE (CPS linkage): This proposal originated as a CPS in PR #964 (“CIP-???? | Merkle Root of
+> Transactions in Block Header”). The CIP process recommends that complex proposals link a CPS as
+> motivation; at the time of writing, the CPS identifier for this work is not finalized in the PR.
```
rphair:
```suggestion
> [!NOTE]
> This proposal originated as a CPS from a general recommendation that complex proposals should first consider a CPS to contain any solving CIPs: but community review determined that the single-solution nature of this problem suggested the purely practical approach of presentation as a CIP alone.
```
The discussion around the time of this https://github.com/cardano-foundation/CIPs/pull/964#issuecomment-2734185518 was that the utility of a CPS was limited since this would be a decision that could only be taken once... and therefore the review of alternative possibilities inventoried in a CPS would instead go into the _Rationale_ as the alternatives considered but not taken.

We have also agreed as editors not to push the requirement of a CPS on people, especially in cases where the CIP process seemed to be taking a long time to push through.  Personally I'm happy we now have a champion for this CIP and I wouldn't want to spend another year getting someone to write the CPS.

In any case I hope the suggested rewrite of this section will reflect this previous discussion (cc @ch1bo & @Ryun1 @perturbing @Crypto2099) ... or perhaps you could strike out this paragraph altogether.

For the particular syntax change: https://docs.github.com/en/get-started/writing-on-github/getting-started-with-writing-and-formatting-on-github/basic-writing-and-formatting-syntax#alerts

rphair (commented):
@colll78 this is great to be moving along again. 🙏 I'll propose at tomorrow's meeting (https://hackmd.io/@cip-editors/129) that this should be made a candidate & given a CIP number and then tagged for circulation more widely in the 2 weeks that follow.

If @ch1bo and others from Mithril & light client work can submit their contributions, then it could be merged soon afterward so that anyone capable of refining this could update it from that baseline as necessary.

* I understand there is no current implementation of this (correct me if that's wrong) so nothing should prevent us from merging a best-knowable specification, with refinements coming later from the relevant teams, [like we had with CIP-0137](https://github.com/cardano-foundation/CIPs/commits/master/CIP-0137/README.md).

@Ryun1 @perturbing I don't think we have to delete `CPS-xxxx/README.md` right away since 1) it still has some active commentary on it and 2) it may serve a purpose in review of the new document (for reviewers who've only seen this as a CPS so far, and to ensure all previously considered approaches are included in the _Rationale_).

@colll78 once that purpose appears to be served please delete that file & `CPS-xxxx` directory so we don't accidentally merge it later. 😅

GeorgeFlerovsky:
@rphair My bandwidth is limited this month, but I'll see if I can steal away some time to review this.

CIP-MPT-BLOCKHEADER/README.md
rphair: committed in https://github.com/cardano-foundation/CIPs/pull/964/commits/7371c8bb37654e291daf0dc900585ceba2a12117

CIP-MPT-BLOCKHEADER/README.md
rphair: no objection (yet): committed in https://github.com/cardano-foundation/CIPs/pull/964/commits/73bcb71dd1ae648d9decb4f2c4b18a165de6c2e1

CIP-MPT-BLOCKHEADER/README.md
```diff
@@ -0,0 +1,283 @@
+---
+CIP: TBD
```
rphair:
```suggestion
CIP: 178
```

rphair (commented):
thanks @GeorgeFlerovsky - and to @colll78 and others present at the CIP meeting today who confirmed the interest of this proposal to those working on light client projects.  We committed to tagging a broader range of people in the course of confirming that this is the correct approach to take.

@colll78 @ch1bo @GeorgeFlerovsky @lehins @perturbing - we will somehow have to determine a threshold of what level of apparent consensus is "good enough" to merge this.  Considering the implications and perhaps irrevocability of such a decision, I would think that bar would be set rather high: if not at unanimity.  Hopefully the next comments to come in will help tag others for a sufficiently broad distribution of this idea: especially now that there is a CIP number to consider.

In any case please rename the directory to `CIP-0178` and update the "rendered proposal" link in the OP 🎉
* It's still on the agenda to delete the artefact CPS directory and its contents from this branch... with anything still useful from that document confirmed to be moved to the CIP.
* If enough time goes by in review, I think @colll78 or one of the editors can just delete this if it appears not to be relevant in the review anymore (and definitely **before merge**).

CIP-MPT-BLOCKHEADER/README.md
rphair: @colll78 @Ryun1 since no alarms were raised about this proposed text at the CIP meeting today, I'll commit it to the working draft (feel free to modify if you think it can be explained more clearly or correctly).

Ryun1:
(**edited**) on agenda for next CIP editors call (number 131), two weeks from today (cc @colll78)
👉 https://hackmd.io/@cip-editors/131

rphair (approved):
We discussed this at the CIP meeting today & another editor mentioned the only prior issue of this lingering for a while as half-n-half CIP/CPS.  This has been fully fixed by @colll78 and in the last 2 weeks there's been no technical dispute from anyone on the tag list of protocol / sidechain developers and so I think it's fit for `Last Check`.

My (elementary) understanding is that if the Cardano community (including sidechain projects) later presents a different notion of facilitating this, then we'll either see a revision to this CIP or an alternative CIP written.  In either case the time has likely come for the block header change to be more seriously considered... so even if not yet confirmed by a chorus of reviewers we should merge it anyway.

I think @colll78's last changes have produced a functional and readable CIP and without anyone to say "wait, this won't work" in its current form then we are good to merge: cc @ch1bo @GeorgeFlerovsky @lehins @perturbing

jtranq:
Just wanted to leave a short comment here to share the relevance to the Cardano IBC Incubator project, as the incubator was mentioned several times in the initial CPS and following conversations, so I just want to clarify the practical implications for us:

Roughly that project is about restoring Cosmos-native cryptographic artifacts on Cardano. To keep the scope of discussion narrow, two important things that includes are Merkleized chain state roots in block headers, but also that block finality/consensus is directly linked to a validator voting mechanism, so once a block is committed, it can't be reversed, so when a header gets produced you can be safe under the assumption of this being the canonical chain tip. Two counterparty chains can very safely just keep exchanging headers and know very quickly and trustlessly what is happening on the other chain, as a factor not just of the merkle root but also finality dynamics.

We have a highly engineered solution for this in our Cardano implementation which I won't dig too deep into, but importantly, for this upgrade to be of use in this type of project, we still have the same finality issue. That is, in the longterm, I'm not sure if this upgrade independently unlocks any big doors in blockchain interop without improvements to finality, though it may have been framed that way earlier.

You can already establish a trusted Merkle root over a subset of UTxOs that your application cares about, especially clean to do this with an STT pattern. That is, developers can already construct on-chain Merkle roots for subsets of the ledger that they care about, which is what we've done for the incubator, but we can't "fake" finality over that same subset in the same way, so in practical terms this CIP-0178 upgrade wouldn't necessarily be covering any major pain point in the incubator project on its own.

colll78:
> Just wanted to leave a short comment here to share the relevance to the Cardano IBC Incubator project, as the incubator was mentioned several times in the initial CPS and following conversations, so I just want to clarify the practical implications for us:
> 
> Roughly that project is about restoring Cosmos-native cryptographic artifacts on Cardano. To keep the scope of discussion narrow, two important things that includes are Merkleized chain state roots in block headers, but also that block finality/consensus is directly linked to a validator voting mechanism, so once a block is committed, it can't be reversed, so when a header gets produced you can be safe under the assumption of this being the canonical chain tip. Two counterparty chains can very safely just keep exchanging headers and know very quickly and trustlessly what is happening on the other chain, as a factor not just of the merkle root but also finality dynamics.
> 
> We have a highly engineered solution for this in our Cardano implementation which I won't dig too deep into, but importantly, for this upgrade to be of use in this type of project, we still have the same finality issue. That is, in the longterm, I'm not sure if this upgrade independently unlocks any big doors in blockchain interop without improvements to finality, though it may have been framed that way earlier.
> 
> You can already establish a trusted Merkle root over a subset of UTxOs that your application cares about, especially clean to do this with an STT pattern. That is, developers can already construct on-chain Merkle roots for subsets of the ledger that they care about, which is what we've done for the incubator, but we can't "fake" finality over that same subset in the same way, so in practical terms this CIP-0178 upgrade wouldn't necessarily be covering any major pain point in the incubator project on its own.

The problem is that for multi-chain interoperability light IBC, you need lightweight state proofs that can be verified onchain, and the only state proof that Cardano has access to that can be feasibly verified onchain are mithril proofs. Mithril proofs attest to the block-header, so in Ethereum you can use them to prove that a given Cardano block-header is authentic (its authenticity is attested to by the mithril stake-weighted attestation). The problem is that as it stands now you cannot actually do anything useful with this certified block-header because it is impossible to prove anything about the transactions within the block that the block_header corresponds to. 

So, the current flow would be:
1, A Cardano block is produced `block_header_a`
2. At some point after, Mithril signers produce a stake-weighted commitment which attests to the authenticity of `block_header_a`, we call this `block_header_a_attestation`
3.  `block_header_a` and `block_header_a_attestation` are provided as inputs to an Ethereum smart contract, and the smart contract verifies the mithril attestation `block_header_a_attestation` at which point it knows that `block_header_a` is indeed a valid Cardano block. 
4. Now that we know that `block_header_a` is an authentic block, want to prove, on Ethereum, that some Cardano transaction happened in the block corresponding to `block_header_a`, the `block_header_a` has the `block_hash` which is the hash of the raw cbor of the block which contains the transactions, so to validate anything about the transactions that happened in that block we need to pass the preimage of that block hash (the actual block cbor) to the Ethereum validator validate it and then extract the relevant transactions from that cbor and validate them. The problem is that blocks can be 80kb+ when you consider the need to manipulate / deserialize the cbor in the consuming contract (ie. Ethereum) and this is completely infeasible within gas limits of the vast majority of chains (Ethereum, Solana, and every other major chain).

ch1bo:
As I see mithril mentioned: Mithril is not only attesting to block headers, but also block contents for which proofs can be queried. For example: https://mithril.network/doc/aggregator-api#/paths/~1proof~1cardano-transaction/get

So if an application relies on Mithril it may not need the proposed merkle root in the header (I believe this is also what @floor-licker said).

However, there are other applications which might be benefiting from these native merkle roots in headers already. Especially, if the finality improves through Peras and eventually we might be able to have true cryptographic light clients?

(Then we might only be missing succinct commitments to ledger state because sometimes one cares more about an output being available, than a particular transaction happened.)

colll78:
> As I see mithril mentioned: Mithril is not only attesting to block headers, but also block contents for which proofs can be queried. For example: https://mithril.network/doc/aggregator-api#/paths/~1proof~1cardano-transaction/get
> 
> So if an application relies on Mithril it may not need the proposed merkle root in the header (I believe this is also what @floor-licker said).
> 
> However, there are other applications which might be benefiting from these native merkle roots in headers already. Especially, if the finality improves through Peras and eventually we might be able to have true cryptographic light clients?
> 
> (Then we might only be missing succinct commitments to ledger state because sometimes one cares more about an output being available, than a particular transaction happened.)

https://mithril.network/doc/aggregator-api#/paths/~1proof~1cardano-transaction/get
This doesn't make sense to me. If this is the case, then it means that mithril signers are now doing stake-weighted threshold signatures of not just blocks, but of a merkle root of transactions that are separate from blocks? If the attested transaction roots are not attested to be associated with a specific block_header how do we know that the transactions attested to are final or recent or which block they are from? It appears that you give a list of tx hashes and the return value includes an attestation to those hashes and only the latest block number across all the txs, which doesn't solve the problem for the vast majority of use-cases because you only know that the transactions are at-least X blocks old if you also prove the progression of blocks using block attestations. More importantly this would also mean that mithril signers are redundantly producing stake-weighted threshold signatures over multiple different commitments (txs root and blocks, because mithril also supports attestations to blocks), which would mean a massive amount of redundant work (mithril signature consensus is very expensive offchain). This is infeasible for any light client design, especially if we want all nodes to be running mithril, we cannot have that level of insane wasted computation that redundantly attests to the same information. 

And indeed, yes, we are missing a commitment to ledger state in blocks. This is critical for mature light clients, Ethereum has it, every Ethereum L2 has it, every sufficiently interoperable ecosystem has it.

jpraynaud:
The Mithril team is currently implementing a new type of certification for blocks and transactions:
- They are signed in the same message to reduce the networking costs for SPOs
- They include the depth of the blocks/transactions at snapshot time, which allows the clients to decide if they are immutable enough for their use case
- This new certification will supersede the existing certification for transactions

As @ch1bo mentioned, by using the Mithril client, an application will be able to prove membership for any block or transaction hash (which also represents the block or transaction content) and decide if it was final enough at snapshot time depending on their security requirements.

rphair:
This had been marked for `Last Check` after my https://github.com/cardano-foundation/CIPs/pull/964#pullrequestreview-4041349048 a couple of weeks ago, but I think @colll78 putting this CIP in proper order ultimately prompted some recent review (with @ch1bo @jpraynaud @floor-licker) with attention to detail that might affect the content of the CIP... so editors shouldn't go ahead & merge this unless we can confirm the points directly above are also settled.

Especially since the spec of this CIP would affect suitability for Mithril I would wait for someone involved in the project — and/or other interested L2's or sidechains — to confirm with a statement like: "Yes, this is the kind of thing we're looking for and can use."  Otherwise the editors wouldn't have the expertise to declare that this is ready to merge just from technical discussion as above.

cc @Ryun1 @perturbing @GeorgeFlerovsky

GeorgeFlerovsky:
@rphair
Sorry. Unlikely I'll get to it this month.

CPS-xxxx/README.md
```diff
@@ -0,0 +1,106 @@
...
+
+2. **Security Implications**  
+   Are there new attack vectors introduced by splitting transaction verification across separate commitments (the existing block body hash vs. a new transaction root)? How would node operators and stake pool operators adapt?
+
+3. **Integration with Mithril**  
+   If headers carry transaction roots, what would a “header certificate” look like in Mithril? Would it become a new certified data type, and how would signers and aggregators handle it?
+
+## Copyright
+
+This Cardano Problem Statement is licensed under [CC-BY-4.0](https://creativecommons.org/licenses/by/4.0/legalcode).
```
rphair:
```suggestion
This CIP is licensed under [CC-BY-4.0](https://creativecommons.org/licenses/by/4.0/legalcode).
```
Updating & committing this to "trick" our validator into checking validation on the rest of the document (so we are not caught off-guard when trying to merge this).

rphair (commented):
@jpraynaud after a long period waiting for reviewers to follow up on this, I realised I have likely misinterpreted your last https://github.com/cardano-foundation/CIPs/pull/964#issuecomment-4225372645 — instead of a required compatibility with Mithril, I think you were suggesting that Mithril is doing something different to accomplish the same thing... as I think @colll78 also confirmed in https://github.com/cardano-foundation/CIPs/pull/964#issuecomment-4206949254.

Again, I'm sorry there hasn't been enough engagement about this proposal for editors to know what to do with it: but now that so much time has gone by — and especially with Phil's https://github.com/cardano-foundation/CIPs/pull/964/changes/5be3937be0715a992b9336d232c6d68c20bb66fc to translate this into a potential, practical solution (even if the resulting CIP might not be universally agreed or adopted) — I would rather merge this as-is than let it sit here through a second calendar year... especially since the original author has moved on.

@Ryun1 @perturbing - I do think the last review comments confirm this is suitable for `Last Check` as I originally arranged above (https://github.com/cardano-foundation/CIPs/pull/964#event-24096178656).
* If anyone has any strong feelings about how this should be changed (cc @ch1bo) then please comment ASAP to "blow the whistle" (before our CIP meeting in 15 hours: https://hackmd.io/@cip-editors/141).
* I feel safe with this strategy — given our low-bandwidth engagement — because anyone with sufficient interest but missing this time frame could then follow up with a PR via a branch that they own: and including any changes to the CIP-0178 text that might reflect current goals & practice.
* Alternatively, if _not_ agreed to merge this at tomorrow's CIP meeting, I would rather close this PR immediately — as a failed attempt to get consensus about how this should be implemented — than have it sit here with one of our "stale" tags which will just end up closing it anyway.

**p.s.** @Ryun1 @perturbing also note our usual problem of the CIP number not being applied to the containing directory (our [only remaining validation problem](https://github.com/cardano-foundation/CIPs/actions/runs/32087977308/job/95564344058?pr=964))... we would have to fix this in a quick administrative PR after merge.

Ryun1:
does this PR mean to have both a CIP and CPS being proposed?

rphair:
@Ryun1 as per the review trail above — and the basis for @colll78's intervention, confirmed beforehand in CIP meeting, to turn this from a CPS presentation to a CIP with the favoured solution (since nobody really wanted anything but the one with the current title) — this currently only exists as a CIP and I have never heard any interest in producing a CPS on this subject since then (and reviewers generally didn't think we needed one).

perturbing (commented):
So currently there is no review from Ledger or consensus on this.

Also note that adding this implies more load on the node for verifying blocks. 

For Plutus built-ins, we generally require benchmarks of new functionality, as complex built-ins might not fit the budget. Can you add a benchmark that shows this will not break block adoption?

rphair:
p.s. @perturbing has pointed out, from the CIP meeting review today, the essential constraints on merging this.  The main idea now seems acceptable but we would hope representative teams to confirm that all aspects of "cost" would be acceptable (or not)... can @lehins @ch1bo @SeungheonOh / @zliu41 please provide this in each of these 3 domains?

colll78:
Please guys, just add a merkle patricia trie of transactions to block header. This is like zero cost, insane value, it takes less than a week of development work. It reduces workload of every single interoperability application in the ecosystem. Every other relevant blockchain with programmability has this. 

There is no need to endlessly relitigate this or propose workarounds (ie. we have XYZ at home which can do this). Developers do not want to interact with workarounds; this is a primitive that is expected to be provided out of the box by any modern programmable blockchain.

rphair:
@perturbing maybe we aren't getting confirmations that this is "OK" from Ledger / Network / Consensus teams simply because of the potential taint in committing to an idea that isn't popular yet.  Hence I am in favour of merging this at least to avoid the bureaucracy that @colll78 is referring to... and I personally feel comfortable with his own assessment that all-around cost is insignificant (addressing our only remaining reason not to merge this).

Ryun1:
as agreed, ill push to the branch the editor fixes - dir name + CPS file

Ryun1 (approved):

PR #971, opened 2025-01-23
CIP-0146? | Multi-signatures wallet registration and discovery

Scitz0:
A new CIP for multi-signatures wallet registration and discovery handling. CIP-1854 describes the meat of multisig (native scripts) but leaves some areas in the peripheral untouched. 

Metadatum label **1854** is also added to the CIP-10 registry.

---

[Rendered Proposal](https://github.com/Scitz0/CIPs/blob/multisig-registration/CIP-0146/README.md)

### DIFF
--- a/CIP-0010/registry.json
+++ b/CIP-0010/registry.json
@@ -72,6 +72,10 @@
     "description": "Voltaire - additional metadata for governance, provided in a separate transaction"
   },
   {
+    "transaction_metadatum_label": 1854,
+    "description": "Multi-signatures wallet registration"
+  },
+  {
     "transaction_metadatum_label": 1870,
     "description": "Open Badges v2.0 compliant metadata"
   },
--- a/CIP-0146/README.md
+++ b/CIP-0146/README.md
@@ -45,11 +45,13 @@ The following rules apply for a multisig registration to be valid:
 - Auxiliary data with a non-empty native scripts array.
 - Auxiliary data with label 1854 metadata that at a minimum includes a mapping for script types.
 - `ScriptType` mapping array in metadata must match the length of, and directly corresponds to the elements in the 
-  `native_scripts` array of `AuxiliaryData`.
-- Public key hashes included in the native scripts must be derived in accordance with CIP-1854, ie using 
+  `native_script` array of transaction auxiliary data.
+- Public key hashes (credentials) included in the native scripts must be derived in accordance with CIP-1854, ie using 
   `purpose=1854'`. 
-- Key derivation limited to path `1854'/1815'/0'/x/0` for best cross-project interoperability and performance.
-- Additional **optional** metadata can be added to the 1854 metadatum label to describe the multisig.
+- Key derivation limited to path `1854'/1815'/0'/x/y`, ie account `0` for best cross-project interoperability and performance.
+- Key index (`y`) must be incremented by `1` for each publicly registered multisig wallet.
+- Key for role (`x`) must use the same index (`y`) for all native scripts in the registration transaction.
+- Additional optional metadata can be added to the 1854 metadatum label to describe the multisig.
 - Optional icon fields is a URL to a non-animated image file, maximum 40kb in size.
 
 ### Data Types
@@ -67,9 +69,9 @@ interface MultiSigRegistration {
 ```
 
 #### ScriptType
-Native script type (to be extended as needed)
+A number representing a specific type (role). This corresponds to the key derivation path role (`x`).
 ```ts
-type ScriptType = 'payment' | 'stake' | 'drep';
+type ScriptType = number;
 ```
 
 #### MultiSigParticipants
@@ -90,83 +92,12 @@ interface MultiSigParticipant {
 }
 ```
 
-#### AuxiliaryDataJSON
-The transaction auxiliary data as defined by CSL. Only the metadata, native_scripts, and prefer_alonzo_format fields 
-should be used but the entire interface is provided for completeness. 
+#### OfflineMultiSigRegistration
+The hex-encoded bytes for a transaction auxiliary data `metadata` and `native_script` array.
 ```ts
-interface AuxiliaryDataJSON {
-  metadata?: {
-    [k: string]: string;
-  } | null;
-  native_scripts?: NativeScriptJSON[] | null;
-  plutus_scripts?: string[] | null;
-  prefer_alonzo_format: boolean;
-}
-```
-
-#### NativeScriptJSON
-```ts
-type NativeScriptJSON =
-  | {
-      ScriptPubkey: ScriptPubkeyJSON;
-    }
-  | {
-      ScriptAll: ScriptAllJSON;
-    }
-  | {
-      ScriptAny: ScriptAnyJSON;
-    }
-  | {
-      ScriptNOfK: ScriptNOfKJSON;
-    }
-  | {
-      TimelockStart: TimelockStartJSON;
-    }
-  | {
-      TimelockExpiry: TimelockExpiryJSON;
-    };
-```
-
-#### ScriptPubkeyJSON
-```ts
-interface ScriptPubkeyJSON {
-  addr_keyhash: string;
-}
-```
-
-#### ScriptAllJSON
-```ts
-interface ScriptAllJSON {
-  native_scripts: NativeScriptJSON[];
-}
-```
-
-#### ScriptAnyJSON
-```ts
-interface ScriptAnyJSON {
-  native_scripts: NativeScriptJSON[];
-}
-```
-
-#### ScriptNOfKJSON
-```ts
-interface ScriptNOfKJSON {
-  n: number;
-  native_scripts: NativeScriptJSON[];
-}
-```
-
-#### TimelockStartJSON
-```ts
-interface TimelockStartJSON {
-  slot: string;
-}
-```
-
-#### TimelockExpiryJSON
-```ts
-interface TimelockExpiryJSON {
-  slot: string;
+interface OfflineMultiSigRegistration {
+  metadata: string;
+  native_scripts: string;
 }
 ```
 
@@ -174,7 +105,9 @@ interface TimelockExpiryJSON {
 
 After the multisig wallet has been defined according to the Cardano native script standard, and adhering to the
 [rules](#specification), either a registration transaction can be put on the blockchain or a JSON download provided for 
-offline sharing.
+offline sharing. 
+
+>Note that the registration **must** use previously unused keys in the scripts **if** registered in a transaction. 
 
 The transaction auxiliary data metadata should be formatted using NoConversions JSON schema.
 
@@ -192,18 +125,19 @@ by its participants.
 
 #### Offline download
 
-The offline JSON file should have the format of `AuxiliaryDataJSON`.
+The offline JSON file should have the format of `OfflineMultiSigRegistration`.
 
 ### Discovery
 
 Discovering registered multisig wallets on the blockchain that has public keys included for a participant wallet 
 can be done in the following way.
 
-- Derive Ed25519 verification keys from path `1854'/1815'/0'/x/0`.
+- Derive Ed25519 verification keys from path `1854'/1815'/0'/x/y`.
 - Create key credentials, `blake2b-224` hash digests of derived keys in previous step.
 - Search for multisig registration transactions on the blockchain that contain metadata with metadatum label **1854** 
-  and key credentials matching participant wallet.
+  and key credentials matching participant wallet. Only the first (oldest) encountered match should be returned.
 - Use `types` field in metadata to map native scripts and figure out its purpose
+- Repeat until no more matches are found, either sequentially or in bulk.
 
 ## Rationale: how does this CIP achieve its goals?
 
@@ -236,16 +170,20 @@ This specification doesn't restrict discovery to any specific role, depending on
 additional roles defined in the future through new CIPs. CIP-1854 define role 0 (payment) and role 2 (stake) for 
 script wallets. CIP-105 was created to extend key definition with additional governance roles 3-5.
 
-#### What about key index?
-This is another restriction in comparison to CIP-1854 specification that allow for multiple indexes. In theory, it 
-can add some limited privacy features. What we gain is simplicity and performance. The rationale here was that the 
-pros outweigh cons.
+#### What about key role and index restrictions?
+The main reason for the incrementing key index is to mitigate a possible attack vector where a malicious actor could 
+replay a publicly registered multisig wallet, modifying the script in a nefarious way. When the wallets connected to the
+users multisig key are discovered, both the valid and the invalid multisig registrations would be discovered and the 
+user might interact with the malicious wallet. By only allowing a key to be registered once, this threat can be 
+eliminated. 
+
+Keeping key index in sync for all roles for the same registration makes key handling more manageable.
 
 #### Is types mapping in metadata really necessary?
 It's true that on the wallet side when deriving purpose 1854 keys and scanning for registered wallets, it's known 
-from what path the keys where derived and thus one could figure out the type of native script based on key. However, 
-enforcing the transaction to include metadata with metadatum label 1854 and the types mapping make it clear that 
-this is a multisig registration and easier to scan for. 
+from what path the keys where derived and thus one could figure out the type of native script based on key credential. 
+However, enforcing the transaction to include metadata with metadatum label 1854 and the types mapping make it clear 
+that this is a multisig registration and easier to scan for. 
 
 ## Path to Active
 
@@ -258,7 +196,8 @@ this is a multisig registration and easier to scan for.
 ### Implementation Plan
 
 - [x] Author to engage with wallet providers and web applications for feedback.
-  - Discussion channel opened on [MeshJS Discord server](https://discord.gg/WvnCNqmAxy)
+  - Discussion channel opened on [Cardano Improvement Proposals Discord server](https://discordapp.com/channels/971785110770831360/1336823914671767663)
+  - [GitHub PR](https://github.com/cardano-foundation/CIPs/pull/971) for proposal.
 - [x] Author to implement said standard in Eternl wallet.
 - [ ] Collaborate with web applications and wallet providers to drive adoption of standard.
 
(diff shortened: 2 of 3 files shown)


Scitz0:
Thanks @Crypto2099 for raising the potential issue at today's CIP editors meeting.

To quickly describe the issue mentioned, we (Eternl) are adding multi-sig (native script) support to our upcoming v2 release. With this, we created this specification (harmonization between wallets/dapps) for how you can pre-announce this native script using the aux data field in tx together with metadata. This was the intention of the field in CDDL as described by CIP-1854. Each participant in the multi-sig somehow needs to know about the script, either through public or offline sharing. The discovery process described in CIP draft looks for TXs containing metadata with label 1854 and native scripts that contain the user's multi-sig key cred. 

The problem as mentioned by Adam is that some malicious user could look for these registration transactions and post their own, mimicking the real one but with a modified script that would allow them to drain the native script without the other participant's involvement. When the participant tries to discover multi-sig wallets that include his multi-sig key cred, they would find both the invalid (scam) wallet, including the real one, and might add and fund the wrong one. 

Is it a problem in real life 🤷‍♂️ ... but it's sort of similar to the scam tokens we have seen sent around. There isn't much you can do about those except be vigilant and educate yourself that scammers exist and what methods they use. 

The only solution I can currently think of is that a multi-sig key (cred) can only be used/registered once. All subsequent registrations would need to use the next key index. Thus a scammer could not replay a posted registration re-using an existing key cred. It complicates handling and syncing but currently the only thing I can think of to mitigate this issue. The other solution would be to acknowledge that it's a potential issue we need to educate users about.

Feel free to add other potential solutions to the problem.

CIP-XXXX/README.md
```diff
@@ -0,0 +1,268 @@
+---
+CIP: \?
```
```suggestion
CIP: 146
```

rphair (commented):
@Scitz0 we agreed at today's CIP meeting to assign this a CIP number regardless of how the issue mentioned in https://github.com/cardano-foundation/CIPs/pull/971#issuecomment-2634632453 may progress.

Please rename the proposal directory to `CIP-0146` and update the proposal link in the first comment above. 🎉

Scitz0:
Updates pushed according to the above-proposed solution for mitigation against attack vector on public multisig wallet registration. It also contains changes to metadata format and various tweaks to improve the standard.

QSchlegel:
Brief comment on Registration could you change to off Chain sharing instead of offline as we expect most pre- registrations to be shared online but off Chain, rather than offline.

Best Regards 
QS

QSchlegel:
Also what do you think about using CIP-0083 (encrypted message metadata) for a pre-registration tx to each potential wallet member.

Downside only works if member wallet addresses are known pre-registration.

To much chain bloat?

Or is this out of scope for CIP-0146?

rphair:
> @Scitz0 https://github.com/cardano-foundation/CIPs/pull/971#issuecomment-2634632453: Feel free to add other potential solutions to the problem.

> @QSchlegel https://github.com/cardano-foundation/CIPs/pull/971#issuecomment-2676809571: what do you think about using CIP-0083 (encrypted message metadata) for a pre-registration tx to each potential wallet member.

It sounds like this could be one solution (assuming those addresses are known in advance)... @Scitz0 can you confirm or refute?

Scitz0:
> Also what do you think about using CIP-0083 (encrypted message metadata) for a pre-registration tx to each potential wallet member.
> 
> Downside only works if member wallet addresses are known pre-registration.
> 
> To much chain bloat?
> 
> Or is this out of scope for CIP-0146?

Scitz0:
Clicked the wrong button :laughing:

gitmachtl:
How about adding another scope to CIP88? Like we did in the CIP88v2 for Pool-Daily-Key Registration?

Scitz0:
> > @Scitz0 [#971 (comment)](https://github.com/cardano-foundation/CIPs/pull/971#issuecomment-2634632453): Feel free to add other potential solutions to the problem.
> 
> > @QSchlegel [#971 (comment)](https://github.com/cardano-foundation/CIPs/pull/971#issuecomment-2676809571): what do you think about using CIP-0083 (encrypted message metadata) for a pre-registration tx to each potential wallet member.
> 
> It sounds like this could be one solution (assuming those addresses are known in advance)... @Scitz0 can you confirm or refute?

Encrypted metadata wouldn't solve issue with replay of multisig wallet registration. The only solution I can see is by rotating keys as currently described in CIP. 
AUX data CDDL for conway look like this:
```
;               metadata: shelley
;   transaction_metadata: shelley-ma
; #6.259(0 ==> metadata): alonzo onwards
; 
auxiliary_data = metadata
                  / [transaction_metadata : metadata
                    , auxiliary_scripts : [* native_script]]
                  / #6.259({? 0 : metadata
                           , ? 1 : [* native_script]
                           , ? 2 : [* plutus_v1_script]
                           , ? 3 : [* plutus_v2_script]
                           , ? 4 : [* plutus_v3_script]})
```
A MultiSig wallet registration contains two parts. The native scripts that define the MultiSig in auxiliary_scripts, alternatively in `1` native_script array for 259 tagged set. This is not encryptable and contains the pub keys of participants. The label `1854` metadata defined in CIP is mainly for additional data like describing the MultiSig. This could theoretically be encrypted according to  CIP-83 if you want to publically share but hide this additional data describing the wallet with a name, description etc. CIP-83 is however an addendum to CIP-20 for label `674` metadata and placing this additional MultiSig metadata under this label would make it harder/less performant to search for MultiSig registration transactions. It would also force participants to share this secret off-chain/offline to be able to reconstruct the metadata. 

If you feel encrypting metadata as a way of hiding 1854 metadata is worth looking into, I could add this as an optional feature. In this case, I would reference CIP-83 and not re-define it. So if `enc` field is present just like described in CIP-83, use that to determine if payload needs to be decrypted or not. In that case I would argue that each string value should be decrypted according to CIP-83, ie something like:
```
{ 
  "1854": { 
    "enc": "<encryption-method>",
    types: ScriptType[
      "base64-string 1", "base64-string 2" ...
    ],
    name: "base64-string",
    description: "base64-string",
    icon: "base64-string",
    participants: {
      [key: string]: {
        name: "base64-string",
        description: "base64-string",
        icon: "base64-string"
      }
      ...
    }
  }
}
```

Scitz0:
> How about adding another scope to CIP88? Like we did in the CIP88v2 for Pool-Daily-Key Registration?

Though I have read through CIP-88 (once) I'm not that familiar with how it works. Could you please in more detail go over what parts of this CIP you think CIP-88 could solve in a better way?

Note that it's important to be able to search for native scripts that include the user's 1854-derived pub key to find all MultiSig wallet registrations that it belongs to. So these must be available on-chain to search for to link to the native script. It's also not possible for participants of the MultiSig to all witness this registration as that would defeat the purpose of the registration/discovery flow and people could just off-chain share the registration. Because of this I currently don't see how CIP-88 in this case can solve it better but as I just explained, I haven't looked deep into CIP88 so I might just not see it.

Scitz0:
> Brief comment on Registration could you change to off Chain sharing instead of offline as we expect most pre- registrations to be shared online but off Chain, rather than offline.
> 
> Best Regards QS

Updated according to your comment wording for offline to off-chain :+1:

CIP-0010/registry.json
```diff
@@ -71,6 +71,10 @@
     "transaction_metadatum_label": 1694,
     "description": "Voltaire - additional metadata for governance, provided in a separate transaction"
   },
+  {
+    "transaction_metadatum_label": 1854,
+    "description": "Multi-signatures wallet registration"
```
Ryun1, line 76:
```suggestion
    "description": "CIP-146 Multi-signature wallet registration"
```
incase of competing standards, adding the CIP numbers aids differentiation

CIP-0146/README.md
```diff
@@ -0,0 +1,220 @@
+---
+CIP: 146
+Title: Multi-signatures wallet registration and discovery
```
Ryun1, line 3:
```suggestion
Title: Multi-signature wallet registration and discovery
```
someone can correct my grammar which is correct here? I have also assumed 'Multi-signature' is a plural

CIP-0146/README.md
```diff
@@ -0,0 +1,220 @@
...
+---      | ---
+Multisig | Shorthand for Multi-party signature scheme.
+CSL      | [Cardano serialization lib](https://github.com/Emurgo/cardano-serialization-lib) by Emurgo.
+
+Multisig wallets have the ability to communicate scripts and metadata ahead of time through a registration 
+transaction utilizing the transaction auxiliary data. This is a convenient way for the multisig participants to 
+discover the multisig wallet before it's first usage.
+
+A common structure of the auxiliary data is needed for interoperability between wallets and web applications. The 
+metadatum label **1854** is used to define the native script types added in the auxiliary data and optionally provide
```
Ryun1, line 38:
```suggestion
metadatum label `1854` is used to define the native script types added in the auxiliary data and optionally provide 
```

CIP-0146/README.md
```diff
@@ -0,0 +1,220 @@
...
+A common structure of the auxiliary data is needed for interoperability between wallets and web applications. The 
+metadatum label **1854** is used to define the native script types added in the auxiliary data and optionally provide 
+information about the wallet and its participants. 
+
+## Specification
+
+The following rules apply for a multisig registration to be valid:
+
+- Auxiliary data with a non-empty native scripts array.
+- Auxiliary data with label 1854 metadata that at a minimum includes a mapping for script types.
```
Ryun1, line 46:
```suggestion
- Auxiliary data with label `1854` metadata that at a minimum includes a mapping for script types.
```

CIP-0146/README.md
rphair: @Ryun1 @Scitz0 it's necessary in English & other languages I know to avoid a "double plural".  The prefix `Multi-` makes it a plural so adding the `s` at the end is redundant and therefore incorrect.  It should be an item on someone's to-do list to correct any such mistakes in CIP-1854 😅

CIP-0146/README.md
Scitz0: Datum being the singular form and data plural, in this case it's a reference to the whole metadata object and not the label or a single value. I could reword it to below if that's preferable?
> metadatum label `1854` metadata.

rphair: @Scitz0 I see what you mean.  In that case `metadata` makes sense because... like most popular uses of the word... it's an arrangement composed ultimately of `metadatum`s 😧 and therefore redundant & not necessary to use both the singular & plural.  TL;DR so I think the way you originally wrote it is fine.

CIP-0146/README.md
Scitz0: Updated all references for this spelling error :+1:

Scitz0:
Typos and spellings mentioned should now be resolved.

CIP-0146/README.md
```diff
@@ -0,0 +1,220 @@
...
+
+### Discovery
+
+Discovering registered multisig wallets on the blockchain that has public keys included for a participant wallet 
+can be done in the following way.
+
+- Derive Ed25519 verification keys from path `1854'/1815'/0'/x/y`.
+- Create key credentials, `blake2b-224` hash digests of derived keys in previous step.
+- Search for multisig registration transactions on the blockchain that contain metadata with metadatum label `1854` 
+  and key credentials matching participant wallet. Only the first (oldest) encountered match should be returned.
```
Ryun1, line 140: Is there the possibility for a replay attack here, where someone submits many of the same registration that they have observed from on-chain?

Scitz0: This is why the key rotation was added to the specification to prevent just that. So only the first registration where a key is used is valid. On creation of a multisig wallet you must first check for next unused key for each participant .

Scitz0:
According to the discussion at the last CIP editors meeting regarding encrypted metadata, a section for this has now been added.

CIP-0146/README.md
```diff
@@ -0,0 +1,249 @@
...
+However, enforcing the transaction to include metadata with metadatum label `1854` and the types mapping make it clear 
+that this is a multisig registration and easier to scan for. 
+
+## Path to Active
+
+### Acceptance Criteria
+
+- [ ] The specification is implemented by three wallet providers or dApps (web applications).
+  - [Eternl Wallet](https://eternl.io)
+  - ...
```
rphair, line 236:
```suggestion
- [ ] The specification is implemented by three wallet providers or dApps (web applications).
  - [x] [Eternl Wallet](https://eternl.io)
```
For consistency with other CIPs... for which a growing list of implementors implies that there will be more.  It looks like Eternl is already on the list; please untick if the implementation isn't completed to your satisfaction yet.

CIP-0146/README.md
```diff
@@ -0,0 +1,249 @@
...
+
+### Implementation Plan
+
+- [x] Author to engage with wallet providers and web applications for feedback.
+  - Discussion channel opened on [Cardano Improvement Proposals Discord server](https://discordapp.com/channels/971785110770831360/1336823914671767663)
+  - [GitHub PR](https://github.com/cardano-foundation/CIPs/pull/971) for proposal.
```
rphair, line 242:
```suggestion
- [x] Author to engage with wallet providers and web applications for feedback.
  - [x] Discussion channel opened on [Cardano Improvement Proposals Discord server](https://discordapp.com/channels/971785110770831360/1336823914671767663)
  - [x] [GitHub PR](https://github.com/cardano-foundation/CIPs/pull/971) for proposal.
```
For consistency with other CIPs.

rphair (commented):
thanks @Scitz0 - my understanding was that securing the metadata was the last thing we were waiting for.  The rest of this has always looked OK to me (though I don't have practical experience in the field) so I'm ready to approve this after a couple formatting adjustments.

CIP-0146/README.md
Scitz0: The implementation is "complete" in Eternl but lack the optional metadata encryption. Can easily be added later on without breaking current implementation so I would classify Eternl as ready in this regard. So I will accept your suggested change.

CIP-0146/README.md
rphair: Appears resolved in subsequent commits.

CIP-0146/README.md
```diff
@@ -0,0 +1,248 @@
...
+CIP: 146
+Title: Multi-signature wallet registration and discovery
+Category: Wallets
+Status: Proposed
+Authors:
+  - Ola Ahlman <ola.ahlman@tastenkunst.io>
+  - Marcel Baumberg <marcel.baumberg@tastenkunst.io>
+Implementors:
+  - Eternl <https://eternl.io/>
+Discussions:
```
Ryun1, line 11:
```suggestion
Discussions:
  - https://github.com/cardano-foundation/CIPs/pull/971
```

CIP-0146/README.md
```diff
@@ -0,0 +1,248 @@
...
+### Encrypted metadata (optional)
+
+To increase anonymity, encrypting the metadata following the specification of [CIP-83 | Encrypted Transaction message/comment metadata](https://github.com/cardano-foundation/CIPs/tree/master/CIP-0083/README.md)
+is supported. For this to be valid, the `enc` key should be added to the root metadata object and each value 
+(**except** `types`) within the metadata should be base64 encrypted according to CIP-83 specification.
+
+>Note that `types` mapping array within metadata should always be unencrypted! 
+
+Example structure:
+```
```
Ryun1, line 164:
```suggestion
```json
```

rphair (approved):
@Ryun1 - in summarising this for my April report I noticed that all our feedback from that period has been addressed: so this should have moved into the final review stages 🤔 cc @Scitz0

CIP-0146/README.md
```diff
@@ -0,0 +1,249 @@
...
+```json
+{ 
+  "1854": { 
+    "enc": "<encryption-method>",
+    "types": [0,2],
+    "name": "base64-string",
+    "description": "base64-string",
+    "icon": "base64-string",
+    "participants": {
+      "<pub-key-hash>": {
```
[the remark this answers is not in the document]
Scitz0: As you are referring to a section in the encrypted metadata for this CIP, I dont see how this would be affected by CIP-112. And in general, I dont think there is a conflict as the content of native scripts are not changed / handled by this CIP.

rphair:
@Scitz0 the feedback from the CIP meeting today is that this is ready to merge as soon as https://github.com/cardano-foundation/CIPs/pull/971#discussion_r2109278240 is addressed to everyone's satisfaction (assuming no further challenges are raised).  Therefore this is still in `Last Check` for the next meeting (https://hackmd.io/@cip-editors/113) unless most editors approve if for merge earlier.

QSchlegel:
## Additional Implementors
The multisig platform from Mesh is also in the process of implementing this CIP.

https://github.com/MeshJS/multisig
https://multisig.meshjs.dev/

rphair (commented):
thanks @QSchlegel (cc @Scitz0) ... proposing we add this to the Acceptance Criteria as a box to be ticked as soon as CIP-0146 compliance is demonstrated:

CIP-0146/README.md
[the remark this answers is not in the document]
Scitz0: Yeah please propose the change you would like to see to this CIP, and I'll review it.

Scitz0:
I made a minor adjustment: if metadata is encrypted, values need to be string arrays, as the length of the content might exceed the 64-character limit.

rphair (approved):
Added MeshJS as _Path to Active_ implementor in https://github.com/cardano-foundation/CIPs/pull/971/commits/0f9b649fb74be46014da47c0114f0d5b19a982df as per https://github.com/cardano-foundation/CIPs/pull/971#issuecomment-2915950092.

`Last Check` status for next CIP meeting (https://hackmd.io/@cip-editors/114) was punted from previous meeting when we ran out of time due to a long issue discussion.

Final approval to merge (as previously suggested in https://github.com/cardano-foundation/CIPs/pull/971#issuecomment-2914722532) should be contingent upon @perturbing @Scitz0 conversation https://github.com/cardano-foundation/CIPs/pull/971#discussion_r2113462669 being resolved... I believe it's currently resolved by default since no actual change has been suggested since then. 👀

perturbing (approved):

PR #973, opened 2025-01-26
CIP-0010 | Add 21325 label for PRISM

FabioPinheiro:
The Transaction Metadata Label to be registered is 21325.
This value 21325 represents the last 16 bits of 344977920845, which is the decimal representation of the concatenation of the hexadecimals 50 52 49 53 4d that form the word PRISM in ASCII.

https://github.com/input-output-hk/prism-did-method-spec

### DIFF
--- a/CIP-0010/registry.json
+++ b/CIP-0010/registry.json
@@ -112,6 +112,10 @@
     "description": "fida.finance transaction metadata"
   },
   {
+    "transaction_metadatum_label": 21325,
+    "description": "PRISM's Verifiable Data Registry for the PRISM DID method"
+  },
+  {
     "transaction_metadatum_label": 61284,
     "description": "CIP-0015 - Catalyst registration"
   },


FabioPinheiro:
I was playing around in making a new Indexer for PRISM. When I discovered the CIP-10.

But I'm not sure how to register the label so I created this PR.
There is already 6453 transactions with this label. I believe that all of them follow the PRISM DID method specs

CIP-0010/registry.json
```diff
@@ -41,7 +41,7 @@
   },
   {
     "transaction_metadatum_label": 839,
-    "description": "Agora - Proposal creation metadata"
```
FabioPinheiro, line 44: Should I undo this?

rphair:
@FabioPinheiro I can see the label number itself is correct: https://github.com/search?q=repo:input-output-hk/prism-did-method-spec%2021325 - but in tonight's biweekly CIP meeting we were looking to authenticate the need for the label from a current project representative.  Are you in a position to do this and/or indicate how the label will be used?

CIP-0010/registry.json
rphair: yes; please take it out: it's better to keep other lines the same to ensure each label has a unique PR associated with it.  If white space is a categorical problem, editors can submit a PR to clean it up or parse it out with some GitHub CI that is running on this file.

FabioPinheiro:
@rphair I'm part of the team that created the specification.

The DID PRISM is a Decentralized Identifiers https://www.w3.org/TR/did-1.0/.
It uses the Cardano blockchain as a Verifiable Data Registry (VDR).
Was developed by the Input Output Global Inc's "Atala Prism" group.
Its the main identity system of the [Hyperledger's Identus project](https://hyperledger.github.io/identus-docs/) and other projects.
The "[PRISM Node](https://github.com/input-output-hk/atala-prism)" is the official implementation of the [prism-did-method-spec](https://github.com/input-output-hk/prism-did-method-spec)

Here is an example of the metadata that is already on mainnet that I pre-computed
https://raw.githubusercontent.com/FabioPinheiro/prism-vdr/refs/heads/main/cardano-21325.csv

We are also creating a new specification just for the VDR part. For the Catalyst project#1300189 https://projectcatalyst.io/funds/13/cardano-open-developers/verifiable-data-registry-for-identus-in-cardano

rphair (approved):
Abundantly justified by https://github.com/cardano-foundation/CIPs/pull/973#issuecomment-2635220657.

FabioPinheiro:
I updated the link https://github.com/FabioPinheiro/prism-vdr/blob/main/mainnet/cardano-21325

CIP-0010/registry.json
FabioPinheiro: Yeah that make sense. Someone could just do a single commit to remove all the unnecessary white spaces and add the commit to `.git-blame-ignore-revs` file

perturbing (approved):

PR #974, opened 2025-01-29
CIP-???? | Removal of Epoch Boundary Blocks

nfrisby:
A CIP for the community to use to discuss removing Epoch Boundary Blocks (EBBs) from the historical Cardano chain.

To do so would ultimately simplify the reference node's specification and implementation, as well as that of any newly developed node etc and potentially various tooling as well --- EBBs are an onerous historical design mistake and it would be a relief to stop preserving them.

Rendered: https://github.com/nfrisby/CIPs/blob/nfrisby/EBBs-CIP/CIP-XXXX/README.md

### DIFF
--- a/CIP-XXXX/README.md
+++ b/CIP-XXXX/README.md
@@ -55,14 +55,37 @@ The removal will proceed in stages.
 
 The staging ensures interop between each successive stage, such that each Stage's node and its predecessor node will always be able to sync to/from one another.
 
+| Node Version | Proto Upstream | Proto Downstream | Storage |
+| - | - | - | - |
+| Stage Zero (ie today's node) | :red_circle: require EBBs | :red_circle: serve EBBs | :red_circle: store EBBs |
+| Stage One | :yellow_circle: treat EBBs as optional | :red_circle: serve EBBs | :red_circle: store EBBs |
+| Stage Two | :yellow_circle: treat EBBs as optional | :green_circle: skip EBBs | :green_circle: do not store EBBs |
+| Stage Three | :green_circle: reject EBBs | :green_circle: skip EBBs | :green_circle: do not store EBBs |
+
 As of Stage Three, all nodes would relay a chain that contains no EBBs and would never store them.
 EBBs could now be entirely removed from the node specification and implementation, except for the prev-hash overrides described in this CIP.
 
-It would be possible to accelerate this plan to merely two stages, at the cost of the additional complexity necessary for the Stage One node to conditionally send EBBs depending on the mini protocol version negotiated with each downstream peer.
-The simpler slower plan seems generally preferable in the absence of any unanticipated urgency — we've already suffered EBBs for several years.
-
 If a Stage One node is itself syncing, it might omit an EBB when serving its (volatile) chain to an un-upgraded downstream peer that requires all EBBs be sent.
 This is an acceptable failure case, because nodes/tools should not be syncing from syncing nodes.
+
+### Alternative Stages
+
+Instead of gating the next Stage on whether enough of the network has adopted the current Stage, the Stages could be gated by handshake versions.
+This allows Stage One and Stage Two to be combined, at the cost of the additional complexity necessary for the Stage OneAndTwo node to conditionally require/send EBBs depending on the mini protocol version negotiated with each upstream/downstream peer.
+
+| Handshake Version | Proto Upstream | Proto Downstream |
+| - | - | - |
+| X (ie today's latest handshake version) | :red_circle: require EBBs | :red_circle: serve EBBs |
+| Y (>X) | :green_circle: reject EBBs | :green_circle: skip EBBs |
+
+| Node Version | Allowed Handshake Versions | Storage |
+| - | - | - |
+| Stage Zero (ie today's node) |  ?≤v≤X | :red_circle: store EBBs |
+| Stage OneAndTwo | ?≤v≤Y | :red_circle: store EBBs |
+| Stage Three | Y≤v≤? | :green_circle: do not store EBBs |
+
+The simpler plan seems generally preferable in the absence of any unanticipated urgency — we've already suffered EBBs for several years.
+On the other hand, the handshake versions might make it easier for different implementations of the node coordinate progress on this CIP.
 
 ### Alternative Mitigations
 
--- a/CIP-XXXX/README.md
+++ b/CIP-XXXX/README.md
@@ -6,7 +6,7 @@ Category: Consensus
 Authors:
   - Nicolas Frisby <nicolas.frisby@iohk.io> <nicolas.frisby@moduscreate.com>
 Implementors:
-  - IOE Consensus Team, as tech debt/maintenance, Nicolas Frisby <nicolas.frisby@iohk.io> <nicolas.frisby@moduscreate.com>
+  - TODO IOE? Tweag? Nicolas Frisby either way, most likely <nicolas.frisby@iohk.io> <nicolas.frisby@moduscreate.com>
   - TODO various tooling authors throughout the community?
 Discussions:
   - Original tech debt Issue, https://github.com/IntersectMBO/ouroboros-consensus/issues/386
@@ -29,7 +29,6 @@ Relatedly, EBBs contain no transactions and are not signed — they contain even
 An EBB's only consequence is an additional step on the chain's sequence of block hashes and the various exceptions, side-conditions, and corner cases necessary to accommodate EBBs in the node's specification and implementation.
 The most egregious of which is that an EBB occupies the same slot as its successor and has the same difficulty (ie BlockNo) as its predecessor.
 As an old, exotic, and useless class of block, EBBs have been initially overlooked in our subsequent design work many times and have ultimately wasted many hours of work as developers troubleshoot and ultimately determine how to accommodate EBBs via corner-cases etc.
-(TODO it would be possible to track down Issues, PRs, etc to partially evidence this.
 But a great deal of the trouble with EBBs has been time lost coping with them in discussions that were not recorded in a manner that demonstrates the EBBs' obstruction.
 Several former and current Consensus developers can attest to the trouble with EBBs, if that is deemed necessary.)
 
@@ -101,25 +100,21 @@ On the other hand, the handshake versions might make it easier for different imp
 TODO This section needs to be resolved as part of this PR before the status of this CIP can transition from Draft to Proposed.
 
 - Deleting EBBs from the Storage Layer would spoil extant Mithril snapshots.
-  In general, how would Mithril handle any kind of change to the on-disk representation of the historical chain — are migration paths permanently required, effectively?
+  In general, how would Mithril handle any kind of change to the on-disk representation of the historical chain---are migration paths permanently required, effectively?
 
-- Which community tools will be disrupted by the node suddenly omitting EBBs when serving the chain?
-  Will separately publishing the set of 176 EBBs (or perhaps only their slots, hashes, and prev-hash fields) suffice for those tools' authors to mitigate the disruption?
+- Discussion on this CIP's [PR](https://github.com/cardano-foundation/CIPs/pull/974) revealed that community tooling authors do not anticipate much disruption.
 
 ## Path to Active
 
 ### Acceptance Criteria
 
-- (TODO, Path to Proposed: review from tooling authors such as db-sync, Mithril, etc has resolved the Open Questions.)
-
 - A release of the Haskell reference Cardano node implementation and documentation satisfies Stage Three above: EBBs are never stored and never exchanged with peers.
 
 - Optionally, the ledger specification and reference implementation removes all mention of EBBs, except for the prev-hash overrides.
 
-### Implementation Plan
+- Once they're removed from the chain, it would be preferable for the EBBs to still be somehow available, even just for reference.
 
-TODO Is this too detailed for a CIP?
-How certain do I need to be of these steps for inclusion here?
+### Implementation Plan
 
 - In Stage One, the node's initialization logic will delete any EBBs from volatile storage.
   (EBBs are typically only in the immutable storage, but upgrading in the middle of a fresh sync could result in an EBB in the volatile storage.)
@@ -138,9 +133,21 @@ How certain do I need to be of these steps for inclusion here?
 - In Stage Three, all changes for Stage One and Stage Two can be reverted except the database migration.
   Instead, the codec for Byron blocks will swap out the prev-hash field that identifies a known EBB with that EBB's prev-hash field, and moreover the codec will refuse to parse EBBs.
 
+- At any point, the existing EBBs can be archived
+  Any of the following three strategies would suffice.
+    - Some tool that is small enough and easy enough to maintain alongside the reference node's source code would be able to regenerate the EBBs from the corresponding historical ledger states (the list of 176 EBB hashes is a lightweight and easy test suite)---recall that each EBB is merely a function of the incoming ledger state.
+      The main difficulty would be rederiving the necessary logic, which has been irrelevant for several years (probably from <https://github.com/input-output-hk/cardano-sl/blob/1499214d93767b703b9599369a431e67d83f10a2/db/src/Pos/DB/Block/Lrc.hs#L244-L254>).
+    - The 176 EBBs (several megabytes in total, when compressed) are somehow archived off-chain.
+      The community would need to establish a best practice for this sort of archival, in a separate CIP presumably.
+      Such a CIP would likely also be useful for any future work towards truncating the historical chain.
+    - Since the 176 EBBs are relatively small, they could be chunked up and stored as UTxO datums on-chain.
+      This last option may seem wasteful, but an extra several megabytes of ledger state will go unnoticed in the long run and it's the only way to trivially ensure that the EBBs are always accessible if the chain itself is.
+      The main difficulty would be the fees for the necessary transactions and ensuring all the datums are easy to locate in the future.
+
 ## References
 
 - The hashes and prev-hashes of all known EBBs, <https://github.com/IntersectMBO/ouroboros-consensus/blob/534c2c2107dcdb2b962c1483bb02a14beae2e608/ouroboros-consensus-cardano/src/byron/Ouroboros/Consensus/Byron/EBBs.hs>.
+- The logic originally responsible for creating the payload of all known EBBs, <https://github.com/input-output-hk/cardano-sl/blob/1499214d93767b703b9599369a431e67d83f10a2/db/src/Pos/DB/Block/Lrc.hs#L244-L254>.
 
 ## Copyright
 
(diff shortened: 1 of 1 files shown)


nfrisby:
@kderme @jpraynaud As promised, I'm pinging you now that I've written up the basic plan for removing EBBs.

Please --- anyone whose reading this --- ping folks who author Cardano tools that would likely be affected by this retcon of the historical chain. Thank you!

Relevant summary so you can quickly judge who might care:

- No regular block's content nor hash would change.
- EBBs would simply be trimmed out of the chain --- nodes would no longer send (accept) them when serving (syncing) the historical chain.
- EBBs are useless blocks: they do not at all affect the protocol nor ledger state.
- As of the culmination of the CIP's plan, the resulting gaps in the prev-hash links would be patched up using a lookup table that contains the hashes and prev-hashes of all EBBs that have ever existed.
- The on-disk files would change, since they currently contain the EBBs.
- The code's various corner cases for EBBs could finally be removed; all consolidated down to that very isolated prev-hash fixup.

jpraynaud:
@scarmuega this will probably have an impact on Pallas

nfrisby:
@abailly @KtorZ @agaffney @AndrewWestberg I'd appreciate your feedback on how disruptive this CIP would be for the nodes/tooling you design and maintain. Thanks!

(This CIP made me starkly realize that Mithril snapshots are something the various nodes are going to have to be able to interoperate on, so I suppose there's a whole other conversation to have there too cc @jpraynaud .)

jpraynaud:
@nfrisby if the immutable files from the Byron era are rewritten at Cardano node startup, this could create problems with the certification made by Mithril for the Cardano database: 

It could be stopped because there is a split between signers (some use the digests of the previous immutable files to sign -because it is in their Mithril signer cache or they have not updated their Cardano node yet- and the others use the digests from the new immutable files), which prevents reaching the quorum to create a multi-signature.

Today the digest is computed as a hash of the files contents. There may be some other way to compute this digest of the immutable chunk files by using the blocks inside the file and by skipping EBB blocks during the first stage you propose. This would need a new Mithril era (i.e. all the signer nodes sign differently at a given epoch transition, which means that a very high majority of the signers run a compatible version) but this could work. However, I don't know if we could compute the digests for the primary and secondary indexes with the same trick.

Regarding the certification of the Cardano transactions, as there is no transactions in the EBB blocks, there should be no impact.

We also have plans to certify the blocks themselves, and in that case we would simply skip the EBBs (as we can detect them with the Pallas library). Depending on when this happens, this could also require a new Mithril era.

abailly:
The fact that mithril signs and serves the cardano-node DB is a historical artefact: it was the most useful and easiest thing we could do when we started the project, and it proved quite effective at boostrapping cardano-node. But of course, we would like to avoid depending on the specific implementation details of _one_ implementation. 

Ideally, I would like to:

* Have mithril provide a certified chain of _blocks_ in some _specified_ (in a CIP?) interchange format, perhaps something as simple as a zstandard concatenation of all blocks grouped by 100s or something 
* Have mithril provide a certified ledger snapshot at some point in some _specified_ interchange format
* Have a tool/command in cardano-node to `import` such raw data. This is the path we have started following with @KtorZ in Amaru (see [import command](https://github.com/pragma-org/amaru/blob/main/crates/amaru/src/bin/amaru/cmd/import.rs)) because it helps us

nfrisby:
> would this eventually end up leaving a gap in the block type IDs returned in the NtC chainsync payload? Could it go even further and get rid of that gap to make those NtC block type IDs the same as the NtN block header IDs and era ID?

I don't follow this question. What are the "block type IDs" and "era ID" etc? My guess is that my anwser is "no", since there would no observable change in behavior except for the fact that the ChainSync and BlockFetch protocols would not send EBBs.

> it kinda becomes a "tribal knowledge" sort of thing that will trip up people doing new implementations

I agree with this. I don't know how to handle it, other than to put up loud blinking marquees throughout the Byron ledger spec about this CIP.

> so it may be possible for an attacker to trick someone into using a "bad" list.

My current thought process is that the list will be baked in the source code, so if an attacker could manipulate my list then they could manipulate everything about my node 🤷. There is still the question of how to ensure the correct list of known EBBs is securely preserved indefinitely --- I don't know how to do that on the level of a long-lived community like this one. We'd need a "secure" version of this CIP repo, eg?

agaffney:
> > would this eventually end up leaving a gap in the block type IDs returned in the NtC chainsync payload? Could it go even further and get rid of that gap to make those NtC block type IDs the same as the NtN block header IDs and era ID?
> 
> I don't follow this question. What are the "block type IDs" and "era ID" etc? My guess is that my anwser is "no", since there would no observable change in behavior except for the fact that the ChainSync and BlockFetch protocols would not send EBBs.

In the node-to-client version of the chainsync protocol, each block payload has a header that includes the block type ID. This is 0/1 for Byron main/EBB blocks, 2 for Shelley, 3 for Allegra, etc.

The node-to-node version of the chainsync protocol also has a payload header a similar identifier, but instead it's just the era ID (0 for Byron, 1 for Shelley, 2 for Allegra, etc.). This ID also matches the TX type IDs used when submitting a transaction via LocalTxSubmission.

It's the difference in chainsync payload format combined with Byron having 2 block types that causes this discrepancy in the type IDs used for blocks in chainsync. It would be nice to get these synced back up, and it's relatively doable since it all happens at the protocol level.

abailly:
Perhaps the delineation of a common format for the _chain_ could be led by Mithril Team? It seems to me that simply reusing the existing node format, perhaps with some adjustments, but with a clear specification, would go a long way towards the goal of standardising. Then Mithril node, as a standalone node or component within a node would only need to follow the chain and construct that data himself, and each node would be free to store whatever they want the way they want.

abailly:
Also I wonder: Would it not be possible for the cardano-node to optimise its storage, yet keep a backward compatible behaviour by emitting those pesky EBB nodes "on-the-fly"?

nfrisby:
@agaffney Thanks for pointing this discrepancy out to me. I had never had cause to notice this; it was designed before my time and hasn't broke anything for my team since. I agree that it's burdensome and surprising.

I see the discrepancy in the code, but I do not ultimately know why it is this way. Specifically, I don't know why the (modern) HFC on-the-wire codec for blocks directly reuses the on-the-disk codec --- I agree with you that it doesn't have to. (The HFC instead does "the compositional thing" for headers.) I could imagine some reasons for why it was this way at first (eg the prevailing HFC design goal that a single-era HFC block is indistinguishable from that single era's block), but I don't see why it would need to still be this way.

> It would be nice to get these synced back up, and it's relatively doable since it all happens at the protocol level.

By "these", do you mean the codecs for the HFC envelope on blocks versus on headers and on txs? Or do you mean the tag in the envelope actually matching the era index (as is currently the case in headers)?

If this CIP were to go through, then there would be no EBBs. Unfortunately, the existing Byron codec uses the "era tag" 0 for EBBs and 1 for regular blocks. So if we left all bytes as-is (except for deleting the EBBs) the era tags would be 1-based instead of 0-based :shrug:. But at least it'd be one tag reserved per era, which is less surpising.

We could change the on-the-wire codec for Cardano blocks to have the same kind of envelope that headers and txs do today, ie even if this CIP is rejected. So I hesitate to bundle that change up with this CIP. If this CIP is accepted, then the codec for Byron itself could be simplified: its 0-versus-1 tag would be unnecessary. Moreover, if we _also_ want to simplify the Byron on-disk codec, then we'd need another database migration --- so maybe it would be nice to combine that migration with this CIP's migration.

(I'm so far assuming that this 0-versus-1 tag in Byron does not influence any of the identifying hashes, but that must be double-checked at some point.)

nfrisby:
@abailly

> It seems to me important to separate the official part (the protocol) from the unofficial part (what people do with the cardano-node storage).

I agree with this sentence.

> Also I wonder: Would it not be possible for the cardano-node to optimise its storage, yet keep a backward compatible behaviour by emitting those pesky EBB nodes "on-the-fly"?

But I disagree with this one. My goal with this CIP is not merely to simplify the storage. I also want to simplify the protocol itself. It just so happens those _most_ of the explicit cases for EBBs are in the storage layer. But EBBs themselves obstruct other aspects of the era-agnostic design as well, not just the storage layer. And it'd be really nice if _no one_ would ever have to consider EBBs in the future beyond "they were a historical mistake that necessitates today's Byron block codec doing the prev-hash swap thing with that one static table".

agaffney:
> > It would be nice to get these synced back up, and it's relatively doable since it all happens at the protocol level.
> 
> By "these", do you mean the codecs for the HFC envelope on blocks versus on headers and on txs? Or do you mean the tag in the envelope actually matching the era index (as is currently the case in headers)?

I was just referring to the block/header type tag/IDs in the chainsync payload header

abailly:
@nfrisby Understood, what I meant was to provide an incremental and smooth migration path, something like:

| Proto version | Feature | Node Version | Storage |
| -- | -- | -- | -- | 
| $v_n$ | serve EBB | $v_m$ | store EBB |
| $v_n$ | serve EBB | $v_{m+1}$ | does not store EBB |
| $v_{n+1}$ | does not serve EBB | $v_{m+1|}$ | does not store EBB |

nfrisby:
@abailly the CIP draft specifies the following in the Specification section. It's different than the table in your comment, but I'm not seeing which difference is crucial. What do you think?

| Node Version | Proto Upstream | Proto Downstream | Storage |
| - | - | - | - |
| Stage Zero (ie today's node) | require EBBs | serve EBBs | store EBBs |
| Stage One | treat EBBs as optional | serve EBBs | store EBBs |
| Stage Two | treat EBBs as optional | skip EBBs | do not store EBBs |
| Stage Three | reject EBBs | skip EBBs | do not store EBBs |

abailly:
I think that I should have read the CIP more carefully before commenting, thanks pointing that out @nfrisby

abailly:
On second thought, I think the difference is that I would not make a difference based on whether you are upstream or downstream, but based on version negotiation, so that an updated node can still serve/receive EBBs depending on the peer's version. But perhaps that's already covered in the CIP too ;)

CIP-XXXX/README.md
```diff
@@ -0,0 +1,124 @@
...
+- Finally, once Stage Two is sufficiently widespread, Stage Three would forbid upstream peers from sending EBBs.
+
+## Rationale: how does this CIP achieve its goals?
+
+The staging ensures interop between each successive stage, such that each Stage's node and its predecessor node will always be able to sync to/from one another.
+
+As of Stage Three, all nodes would relay a chain that contains no EBBs and would never store them.
+EBBs could now be entirely removed from the node specification and implementation, except for the prev-hash overrides described in this CIP.
+
+It would be possible to accelerate this plan to merely two stages, at the cost of the additional complexity necessary for the Stage One node to conditionally send EBBs depending on the mini protocol version negotiated with each downstream peer.
```
nfrisby, line 61: @abailly I think this paragraph is the extent of what the current draft says relevant to your suggestion in this comment https://github.com/cardano-foundation/CIPs/pull/974/#issuecomment-2628056362

nfrisby: Maybe I'm overlooking something? Let me know!

KtorZ:
> @nfrisby  I'd appreciate your feedback on how disruptive this CIP would be for the nodes/tooling you design and maintain. Thanks!

In my case, not disruptive at all. 

### Ogmios

Doesn't really care about EBB, as Ogmios will "simply" proxy answers from the (local-)chain-sync protocol down to clients. It will make my life _slightly_ easier as it removes the need [to document them](https://github.com/CardanoSolutions/ogmios/blob/master/docs/static/cardano.json#L66-L86), and explain why people should just ignore them. 

### Kupo

Kupo's interface with the chain via _transactions_ more than blocks. So I mostly leverage the [`extractTxs` method from the `HasTx` class](https://github.com/IntersectMBO/ouroboros-consensus/blob/0f852eee718295e948c160665045c9ff3be4b85d/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Ledger/SupportsMempool.hs#L156) defined in `ouroboros-consensus`, which yields an empty list for EBB. 

When plugged into Ogmios as a data-source instead of a Haskell node, I simply [default to an empty list](https://github.com/CardanoSolutions/kupo/blob/afb9b2854ae9db576441736fd54f465a0d358263/src/Kupo/Data/Ogmios.hs#L195) of transactions when running into an EBB (so, similar behaviour down the line)

### Aiken

No impact whatsoever, blocks aren't visible at the smart contract level, nor do they surface when simulating / evaluating transactions.

### Amaru

In Amaru, we always bootstrap from _somewhat recent_ snapshots. So we completely alleviate the problem: Byron doesn't even exist for Amaru. The strategy is to keep producing snapshots so that we always ever need to support only two eras: the current, and the next one. 

This is possible because Ouroboros gives us strong immutability guarantees after 2160 blocks. And by leveraging Mithril, we can hopefully circumvent dynamic availability attacks that come with it (lengthy discussion probably, but not the point of this CIP I reckon). 

---

So all-in-all, all good. But also, little impact too IMO. I don't know many dApps that are interested in re-validating old data; the snapshot strategy sounds like something we could make more generally adopted overall: 

- Make mithril a first-class citizen in the (all) node implementation
- Possibly add or revise incentives to accommodate for running Mithril signers as part of the SPO duties
- Bootstrap from stake-based-signed snapshots, and get rid of old eras altogether.

nfrisby:
@KtorZ thanks very much for the response! Glad to hear it wouldn't be very disruptive for you.

CIP-XXXX/README.md
nfrisby: FYI, I elaborated this section, adding the tables we were using in our other discussion.

kderme:
It would be nice to have an easy way to extract the lookup table of old EBBs from the ledger state

Quantumplation:
From my perspective, I don't see anything wrong with this for any of the projects I interact with, other than perhaps the impact on Mithril.

In fact I'm of the much more radical opinion that we should re-genesis the chain entirely to start at Shelley and forget Byron ever existed :sweat_smile:

I don't think there are any major "zkOuroboros" projects out there, but in the future when there are, they would likely care about having a *general* policy for this kind of historical change, as it could change the ability to construct a folded proof from scratch, etc.

nfrisby:
> that we should re-genesis the chain entirely to start at Shelley

I recall a conversation with some of the Praos/Genesis authors and they were surprised this wasn't the plan! So not _heretical_ by any means :D

nfrisby:
My notes from the  discussion in the Network Working Group meeting earlier today.

- The CIP should quote Section "7 Blockchain extension" of the Byron spec pdf.
- Karl point out that this CIP should definitely mention EBB's role in the "Lazarus event"!
- Brian Bush says on Marlowe work the indexer was once corrupted because they made the "obvious" assumption of at most one block per slot --- EBBs were a disruptive surprise. They figured it out, but it was a distraction and wasted some time.
- Brian Bush suggests indexers are a likely to care about this retcon.
- Arnaud recommends exchanges might have internal tools (eg indexers) that could be disrupted.
- Generally, the CIP should warn about DBs becoming "out of sync".
- Reminder: the CIP should discuss permanent storage, to address Kostas's idea above. There were some ideas brainstormed on Slack, but nothing seemed obviously best for how to permanently store ~12 megabytes _off chain_.

nfrisby:
For the permanent archive of the EBBs, I can think of two main options.

- Ironically: we put them back on the chain. We can compress them to the nines using a standard algo, and then chop them into small chunks which we attach to a series of TxO as auxiliary data. (Maybe there's a better way?)

- We ask the community to host redundant copies of them however they see fit. Ideally these are forever available, which doesn't seem prohibitive for 12 megabytes. IOG, CF, Emurgo, Tweag, etc would probably be willing to stash them on some simple URL? One of the copies could just be in this repository, even. A 12 megabyte file that _never changes_ seems rather harmless to include in the repo. It would inconvenience the first fetch anyone does, though, so we could instead leverage [GitHub's support for large files](https://docs.github.com/en/repositories/working-with-files/managing-large-files/collaboration-with-git-large-file-storage).

nfrisby:
In my opinion, the challenge posed by Mithril is currently the only thing keeping this PR in Draft status. I know Peras also raises similar challenges (since different honest nodes might have different historical Peras certificates), for example. So the general challenge is already being considered, but I don't see how to personally progress this CIP draft in the mean-time.

Quantumplation:
@nfrisby someone (me) jokingly suggested today at the Leios workshop that it might be convenient to have a special kind of block just at the boundary between epochs (I forget the context, something about avoiding conflicts with the treasury assertion field if Leios ends up introducing a form of sharding). If that makes it's way back to you somehow I just wanted to frontrun it so you know I was joking and don't die of anxiety :P

rphair:
@nfrisby after 13 more months, is there currently any way to address the Mithril challenge?  My non-expert understanding of the discussion above would be that this change wouldn't be disruptive, but also that the historical nature of this problem means a further waiting period in `Draft` mode shouldn't be much of a concern.

Still the editors would like to know what might be done with this (cc @Ryun1 @perturbing).  Rather than tagging as "under review" but `Waiting for Author` — as we generally would for a complete CIP waiting for a real-world event for confirmation — it seems like Implementors & Path to Active are still pending applicability to Mithril (and other light clients? see https://github.com/cardano-foundation/CIPs/pull/942) and so the indefinite period in `Draft` mode seems to make sense here.

In case there is more information from any quarter that would help finally give a proper formulation to this CIP, I'm re-tagging all review participants so far: @kderme @jpraynaud @scarmuega @abailly @agaffney @KtorZ @Quantumplation

(@nfrisby from this point onward — since moving this PR into the review process would require documentation of changes & correlating these with review comments — please avoid any further forced pushes to this branch unless they also contain the existing commit history shown here.)

rphair:
@nfrisby (cc co-editors @Ryun1 @perturbing) I considered closing this again because — noted after another 3+ months without comment — having this `Draft` here as an incompletely formulated CIP does not appear to be attracting the interest & participation of key developers... as @Quantumplation pointed out not long ago in https://github.com/cardano-foundation/CIPs/pull/974#issuecomment-2773927243.

The fact that the Leios reviewers (cc @ch1bo — pass it on please?) haven't commented here means that this `Draft` might likely _not_ be the best place to discuss this issue.  For all the CIP editors know, by now it could be: discussed somewhere else, concluded somewhere else (for or against), or incorporated into some other long-term migration plan for which CIP formulation might not be required at all.

Therefore I am tagging this `Likely Abandoned` unless someone (especially you Nicolas I hope) can establish that this is the _primary_ thread & technical proposal for the removal of EBBs.
* Without that, I think the CIP editors would have to consider this a distraction to the overall discussions of the evolving Cardano protocols and better off closed to avoid fragmentation or misinformation.

rphair:
@nfrisby closing for the reasons stated above in https://github.com/cardano-foundation/CIPs/pull/974#issuecomment-5310777490.  If something kick-starts the progression of this idea please feel free to comment — confirming that this is _the_ place to discuss the prospect of EBB removal — and with more detail about "where we will go from here".

PR #982, opened 2025-02-04
CPS-???? | Node Behavior during Low Participation

nfrisby:
A CPS for the community to elicit proposals for how the node should behave when the best chain it has access to is growing much slower than Cardano's Praos security argument anticipates when full participation is assumed.

The fact that this behavior is unspecified can obstruct node design and/or may be overlooked, such that nodes and community tooling may cause a failure cascade if the network ever does suffer low participation for some extended period (eg bug, solar flare, etc).

Rendered: https://github.com/nfrisby/CIPs/blob/nfrisby/low-participation-CPS/CPS-XXXX/README.md

### DIFF
--- a/CPS-XXXX/README.md
+++ b/CPS-XXXX/README.md
@@ -1,0 +1,134 @@
+---
+CPS: XXXX
+Title: Node Behavior during Low Participation
+Status: Draft
+Category: Consensus
+Authors:
+    - Nicolas Frisby <nick.frisby@iohk.io> <nicolas.frisby@moduscreate.com>
+Proposed Solutions: []
+Discussions:
+    - https://github.com/cardano-foundation/CIPs/pull/982
+Created: 2025-02-04
+License: CC-BY-4.0
+---
+
+## Abstract
+
+It is unclear how the Cardano node should behave when the best chain it has access to is so sparse that it violates the Praos security argument's lower bound on Chain Growth.
+This CPS motivates that question, lists some relevant concerns, and hopefully attracts some useful CIPs.
+
+The [Praos security argument](https://eprint.iacr.org/2017/573) assigns overwhelming probability to a lower bound on growth under the assumption that all honest parties are participating and together control more stake than the adversary.
+For Cardano specifically, that Chain Growth property has been instantiated as at least 2160 blocks of growth during each 36 hr interval, whereas the expected Cardano growth is 2160 blocks per 12 hr (assuming full participation).
+On the other hand, the [Disaster Recovery Plan](https://github.com/cardano-foundation/CIPs/tree/master/CIP-0135), intervenes when the chain does not grow at all for 36 hours.
+These two specifications do not address an observed growth rate less than 2160 per 36 hr but greater than zero, nor does any other specification.
+
+Such low growth rates become plausible when the assumptions of the Praos argument are violated for an extended period of time.
+Global-scale disasters, infrastructure attacks, local outages, eclipse attacks, bugs, negligent misconfiguration, etc could all cause it.
+
+## Problem
+
+The following mutually exclusive options seem to cover the design space, but each has unique advantages and disadvantages.
+
+- *AssumeImmutableAge* (AssumeIA).
+  The node will not switch to a chain if doing so would rollback more than 2160 blocks regardless of their age.
+  This limit is justified by the Common Prefix property of the Praos security argument as instantiated for Cardano.
+  (Note the absence of the Chain Growth property.)
+
+- *EnforceImmutableAge* (EnforceIA).
+  As a refinement of AssumeIA, the node will not switch to a chain if doing so would rollback more than 2160 blocks and/or a block that is more than 36 hr old (TODO: should age be with respect to the wall clock or the selection's tip?).
+  This limit is a disjunction of Common Prefix and Chain Growth.
+
+- *EnforceChainGrowth* (EnforceCG).
+  In addition to AssumeIA, the node rejects any chain that has less than 2160 blocks per 36 hr (even if selecting it wouldn't require rolling back any blocks).
+
+These options can be compared by considering the following scenarios.
+
+- *TwoThirdsInaccessible*.
+  Suppose more than 2/3rds of stake was knocked offline (eg by a solar flare) such that less than 1/3rd of all stake was active for more than 36 hr, but it all remained connected.
+
+    - The other scenarios all involve this one to some degree, so it serves as a baseline.
+
+    - If the adversary controls less than half of the surviving stake, then it still cannot defeat Praos during the outage.
+
+    - Note that the chain would have a growth rate less than 2160 per 36 hr but greater than zero.
+      
+    - If the whole network implemented AssumeIA, then the 2161th youngest block would be at least three times older than usual (36+ hr instead of 12 hr).
+      Since the nodes and/or other Cardano community tooling might have been designed around the guarantee of at most 36 hrs for 2160 blocks, they could fail outright or become more vulnerable to possible DoS attacks during the outage.
+
+    - If the whole network implemented EnforceIA, then the nodes would not face increased risks.
+      The 2160 disjunct in the EnforceIA definition is an optimization for when the chain is growing faster than 2160 blocks per 36 hr.
+
+    - If the whole network implemented EnforceCG, then the Recovery Plan would be unavoidable (once the affected stake was back online).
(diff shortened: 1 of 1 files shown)


nfrisby:
I'd appreciate your review @nc6, since we've discussed this in the past.

CPS-XXXX/README.md
```diff
@@ -0,0 +1,134 @@
...
+  The node will not switch to a chain if doing so would rollback more than 2160 blocks regardless of their age.
+  This limit is justified by the Common Prefix property of the Praos security argument as instantiated for Cardano.
+  (Note the absence of the Chain Growth property.)
+
+- *EnforceImmutableAge* (EnforceIA).
+  As a refinement of AssumeIA, the node will not switch to a chain if doing so would rollback more than 2160 blocks and/or a block that is more than 36 hr old (TODO: should age be with respect to the wall clock or the selection's tip?).
+  This limit is a disjunction of Common Prefix and Chain Growth.
+
+- *EnforceChainGrowth* (EnforceCG).
+  In addition to AssumeIA, the node rejects any chain that has less than 2160 blocks per 36 hr (even if selecting it wouldn't require rolling back any blocks).
```
nc6, line 42: It's not directly clear to the reader (or, this reader at least!) how these immediately relate to the question above, since they don't say "what does the node do if it has a sparse chain" but rather constrain some other operations. I would suggest moving the *TwoThirdsInaccessible* block up above the description of the options, in order to outline the things that the node _could do_, and then from there lead into the possibilities and how they relate to various security arguments

CPS-XXXX/README.md
```diff
@@ -0,0 +1,134 @@
...
+
+    - If the adversary controls less than half of the surviving stake, then it still cannot defeat Praos during the outage.
+
+    - Note that the chain would have a growth rate less than 2160 per 36 hr but greater than zero.
+      
+    - If the whole network implemented AssumeIA, then the 2161th youngest block would be at least three times older than usual (36+ hr instead of 12 hr).
+      Since the nodes and/or other Cardano community tooling might have been designed around the guarantee of at most 36 hrs for 2160 blocks, they could fail outright or become more vulnerable to possible DoS attacks during the outage.
+
+    - If the whole network implemented EnforceIA, then the nodes would not face increased risks.
+      The 2160 disjunct in the EnforceIA definition is an optimization for when the chain is growing faster than 2160 blocks per 36 hr.
```
nc6, line 59: I find the way that this is written to be slightly confusing. AIUI, the behaviour here would be identical to AssumeIA, since there's no rollback involved. So we'd still have the same concerns as under AssumeIA, right?

CPS-XXXX/README.md
```diff
@@ -0,0 +1,134 @@
...
+
+    - If the whole network implemented EnforceCG, then the Recovery Plan would be unavoidable (once the affected stake was back online).
+      The silver lining is that all nodes would be able to switch to the repaired chain automatically, without needing manual intervention.
```
nc6, line 62: Likewise, I would elaborate more on this case - would grow the chain up until the 36 hour point, when all nodes would stop growing the chain and we would need to effect disaster recovery. But since we hadn't grown more than 2160 blocks, each node would be happy to roll back and select the denser chain...

nc6 (changes requested):
I think fundamentally this looks good, but I would maybe try to make it slightly more accessible by spelling out the scenarios slightly more and possibly re-ordering some of the sections

CPS-XXXX/README.md
```diff
@@ -0,0 +1,134 @@
+---
+CPS: XXXX
+Title: Node Behavior during Low Participation
+Status: Draft
```
rphair, line 4:
```suggestion
Status: Open
```
@nfrisby please note marking `Draft` review status on GitHub is enough to indicate this is a draft; while the CPS itself will be `Open` until considered solved.  There is no `Draft` status for either CIPs or CPSs.

Likewise the CIP header of https://github.com/cardano-foundation/CIPs/pull/974 can be changed to `Proposed` since that is most likely the way it would be reviewed (once emerging from `Draft` review status) and merged.

CPS-XXXX/README.md
nfrisby: I added an intro and then added some connections between these options and the intro. Please let me know if that suffices. So far, I'm hesitant to introduce the scenarios first, but maybe that'd be better.

CPS-XXXX/README.md
nfrisby: I added a bit here. And a TODO... I haven't yet done the math to bound how many blocks might slowly arise before the rule actually kicked in.

CPS-XXXX/README.md
nfrisby: I hear you. I try to be as explicit as possible, so I like that the rendered documents says Draft for now.

rphair: sure, although other editors & reviewers are likely to call attention to the same item (though hopefully they will see your stipulation) since it would be a big headache if the document were inadvertently merged this way (i.e. web parsers like `cips.cardano.org` not working right & needing a further PR to clean it up).  As far as I'm concerned it's in your hands. 🙏

CPS-XXXX/README.md
```diff
@@ -0,0 +1,210 @@
...
+
+    - Note that the chain would have a growth rate less than 2160 per 36 hr but greater than zero.
+      
+    - If the whole network implemented AssumeIA, then the 2161th youngest block would be at least three times older than usual (36+ hr instead of 12 hr).
+      Since the nodes and/or other Cardano community tooling might have been designed around the guarantee of at most 36 hrs for 2160 blocks, they could fail outright or become more vulnerable to possible DoS attacks during the outage.
+
+    - If the whole network implemented EnforceIA, then the nodes would not face increased risks.
+      The 2160 disjunct in the EnforceIA definition is an optimization for when the chain is growing faster than 2160 blocks per 36 hr.
+
+    - If the whole network implemented EnforceCG, then the Recovery Plan would be unavoidable (once the affected stake was back online), since the honest nodes would eventually refuse to the newly minted blocks, since they are more than 36 hr younger than their 2161st predecessor.
```
nc6, line 118:
```suggestion
    - If the whole network implemented EnforceCG, then the Recovery Plan would be unavoidable (once the affected stake was back online), since the honest nodes would eventually refuse to select the newly minted blocks, since they are more than 36 hr younger than their 2161st predecessor.
```

CPS-XXXX/README.md
```diff
@@ -0,0 +1,210 @@
...
+    - The other scenarios all involve this one to some degree, so it serves as a baseline.
+
+    - If the adversary controls less than half of the surviving stake, then it still cannot defeat Praos during the outage.
+
+    - Note that the chain would have a growth rate less than 2160 per 36 hr but greater than zero.
+      
+    - If the whole network implemented AssumeIA, then the 2161th youngest block would be at least three times older than usual (36+ hr instead of 12 hr).
+      Since the nodes and/or other Cardano community tooling might have been designed around the guarantee of at most 36 hrs for 2160 blocks, they could fail outright or become more vulnerable to possible DoS attacks during the outage.
+
+    - If the whole network implemented EnforceIA, then the nodes would not face increased risks.
```
nc6, line 115:
```suggestion
    - If the whole network implemented EnforceIA, then the nodes would not face increased risks due to large numbers of slots remaining volatile.
```

nc6 (approved):
I think this is much more comprehensible now. I've made a couple of suggestions, and we should get a review from somebody who hasn't been involved in the discussion here.

rphair:
@nfrisby (cc @nc6 + newly tagging @ch1bo): I hope I am not wrong to infer, based on my non-expert understanding of the Consensus behaviour described here, that the goals of this CPS have been expressed in Leios in the meantime — or at the very least were included in its design considerations and therefore have more recent evolutions in the Leios document & its review, including the latest iteration:
* https://github.com/cardano-foundation/CIPs/pull/1167

If I am wrong about this, or if you wish to reopen this draft for any reason, please:
1. update this CPS so it brings discussion of all related Consensus matters up-to-date & so you feel it's ready to present for peer review to all Consensus experts & working teams;
2. reopen this PR, and we will set (or confirm your setting) full review status to properly present the resulting CPS into the usual review process.

nfrisby:
FYI, this is unrelated to Leios, but I don't currently have the bandwidth (:smirk:) to update the document.

Thanks for the ping, @rphair.

PR #991, opened 2025-02-19
CIP-0149? | Optional DRep Compensation

colll78:
A CIP that proposes a standard that enables users to elect to donate a percentage of their staking rewards to their DRep. This introduces an entirely optional DRep compensation scheme without requiring any changes to the ledger. The proposal was designed to maximize the simplicity of implementation and to bring the feature to mainnet as quickly as possible. 

---

[rendered proposal on fork](https://github.com/colll78/CIPs/blob/drep-comp/CIP-optional-drep-compensation/README.md)


gitmachtl:
aaaa... your metadata does not include the drep that this set ratio belongs too, what if the user changes the drep? what if he does it in another wallet or on the cli not being asked to correct the metadata? also, there is not max limit in the CIP, is it 100%?
if wallets are the last instance, you may ask the user if he really wanna spend xxx ada during the claim process. some might forget about such a setting during a claim every few months. a drep can provide a payment address, would be a bit easier to just send over a tip imo.

colll78:
> aaaa... your metadata does not include the drep that this set ratio belongs too, what if the user changes the drep? what if he does it in another wallet or on the cli not being asked to correct the metadata? also, there is not max limit in the CIP, is it 100%?

The metadata is included in the delegation transaction so including the DRep would be redundant, the expectation is that wallets query for the delegation transaction, and identify the DRep (and their payment address) from there as it is available in the drep's registration transaction. If you change DReps, your new delegation transaction will not contain this metadata (unless you explicitly opt-in again) and thus you will no longer be opted-in. The max limit is indeed 100%, I'll add that explicitly. 

> if wallets are the last instance, you may ask the user if he really wanna spend xxx ada during the claim process. some might forget about such a setting during a claim every few months. a drep can provide a payment address, would be a bit easier to just send over a tip imo.

The issue with sending over tips is that it is something you have to consciously remind yourself to do, and is not practical for that reason. The whole point of this is that it is an automated opt-in system that doesn't require users to regularly do anything outside of their normal activity. If people want to send tips they can already do that, but many want to compensate their DRep without needing to regularly do extra work just to send them nominal amounts of Ada.

rphair (commented):
@colll78 offhand I can't think of any reason why this shouldn't simply be laid out there as a `Proposed` CIP for the community to see what they might do with it... although if further specifications are made about the tooling over time we should keep an eye on the `Category:` and perhaps change it to `Tools` if so.

Marking `Triage` to introduce it at next CIP meeting: https://hackmd.io/@cip-editors/107

CIP-optional-drep-compensation/README.md
```diff
@@ -0,0 +1,57 @@
+---
+CIP: CIP-DRep-Comp
+Title: Optional DRep Compensation
+Category: Metadata
+Status: Proposed
+Authors:
+    - Philip DiSarro <info@anastasialabs.com>
+Implementors: []
+Discussions:
+    - https://github.com/cardano-foundation/CIPs/pull/?
```
Ryun1, line 10:
```suggestion
    - https://github.com/cardano-foundation/CIPs/pull/991
```

realdecimalist:
Providing feedback on behalf of Talos:

`thread_znY527qCoRSJ0O89mmQaLDQu (16095 tokens)`

> Talos
> First impressions: the CIP’s opt-in approach for compensating DReps via staking rewards is elegant in its simplicity and avoids ledger-level modifications—definitely a plus for quick deployment. However, a few considerations come to mind:
> 
> Immutability and DRep Changes
> • If a user re-delegates to a different DRep or revokes delegation altogether, that new delegation transaction presumably won’t carry any “compensation metadata,” thereby resetting the arrangement to zero. That’s good from a “nothing changes unless you explicitly set it” standpoint, but it does mean wallets need to detect and handle old references to DReps.
> • You might want to clarify what happens if a user simply uses a different interface (like CLI) or a second wallet that doesn’t preserve the compensation metadata. The CIP should specify how (or whether) to warn the user that continuing with that new non-opt-in transaction effectively cancels the arrangement.
> 
> User Experience in Wallets
> • The key promise is frictionless donation. To maintain user trust, wallets should implement a clear UI flow: someone opts in, sees the percentage, confirms, and then has an easy “Disable” or “Adjust Percentage” interface.
> • Even if it’s entirely optional, you’ll want to prevent “accidental generosity.” For example, if a user sets 35% to a DRep but months later forgets, they might be puzzled seeing smaller rewards. A quick reminder or user prompt—especially for higher percentages—could be helpful.
> 
> Maximum Percentage
> • Allowing up to 100% is bold but may be surprising to some. Indicating a recommended range (say “commonly 5–20%”) might be enough to steer new users away from unintentional extremes, without imposing a ledger-level restriction.
> 
> Metadata Context
> • One comment in the GitHub thread raises whether you need to store the actual DRep identifier in the metadata. Strictly speaking, yes, the delegation transaction already references the DRep. Still, for clarity and consistent wallet implementations, it might be worth explicitly matching the compensation ratio to that DRep in the metadata (so a wallet can parse everything from one place). That might forestall confusion if code changes or if, in future CIP updates, the delegation reference flows differently.
> 
> Edge Cases & Future Evolution
> • Over time, communities might want more sophisticated compensation splits if they are delegating to multiple DReps (a future CIP scenario). For now, the singular “opt in to one delegate” is fine, just be mindful that the next iteration could get complicated if multi-delegation to different DReps becomes an everyday scenario.
> • A ledger-level feature in the future might allow for programmatic reward routing. This CIP is more of a minimal, quick path—so it’s important to be explicit: it doesn’t preclude a more elaborate approach down the road.
> 
> Overall, the proposal’s main strengths are minimal friction and ease of implementation. The optional design means no forced overhead or ledger changes, which should streamline adoption. The caution is that wallet UIs need to do the heavy lifting for clarity and user consent. As long as the CIP includes guidelines for robust wallet UX (especially around re-delegation, un-delegation, or toggling the compensation ratio), you’ll likely see a smooth path to mainnet acceptance.

image

CIP-optional-drep-compensation/README.md
```diff
@@ -0,0 +1,57 @@
...
+
+The best way to approach feature development is to release an MVP, evaluate its reception in the market, and use that data to determine the best way to proceed. This also allows us to introduce the feature without imposing it on anyone, it is opt-in, if you do not wish to compensate your DRep you can simply not opt-in. If, at any time, you wish to adjust the percentage of rewards that you are donating to your DRep you can just issue a new delegation certificate and adjust the amount or opt-out entirely.
+
+## Specification
+This feature is extremely simple to implement. When a user delegates to a DRep, the user should receive a prompt asking if they want to donate a percentage of their staking rewards to that DRep (ie. if they want to opt-in to compensate their DRep).
+
+If they choose to opt-in, they enter the percentage of their staking rewards that they would like their DRep to receive. The DRep delegation transaction will include that percentage in the transaction metadata in the following format:
```
Ryun1, line 27:
```suggestion
If they choose to opt-in, they enter the percentage of their staking rewards that they would like their DRep to receive. The DRep delegation transaction will include that percentage in the transaction metadata, with metadatum label `3692`, in   the following format:
```
add specific detail on `metadatum_label`, this PR should also this label to the CIP-10 registry to reserve it

CIP-optional-drep-compensation/README.md
```diff
@@ -0,0 +1,57 @@
...
+
+```
```
Ryun1, line 29:
```suggestion
```json
```

CIP-optional-drep-compensation/README.md
```diff
@@ -0,0 +1,57 @@
...
+{
+    "3692": {
+        "int": PERCENTAGE_INTEGER_HERE
```
Ryun1, line 32:
```suggestion
        "donationPercentage": PERCENTAGE_INTEGER_HERE 
```
might be nice to be more specific here

CIP-optional-drep-compensation/README.md
```diff
@@ -0,0 +1,57 @@
...
+```
+{
+    "3692": {
+        "int": PERCENTAGE_INTEGER_HERE 
+    }
+}
+```
+Where the integer stored in the metadata is the thousandths place in the decimal so 1 would convey that the user is electing to donate 0.01% of their staking rewards to their DRep. This value should be obtained directly from the percentage the user described in the prompt (ie. the user put 0.5% in the prompt then the integer in the metadata would be 5).
+
+When a user claims their staking rewards, their wallet will look up their delegation transaction and determine if they have opted-in to compensate their DRep (by checking the tx metadata). If metadata in the format described above is found, then the wallet will add an additional output to the reward withdrawal transaction that sends the intended percentage of their rewards to their DRep. If there is not enough ada in their reward balance to pay the minAda for that output, the wallet will convey this to the user and let them decide how to proceed (or the user can establish how to handle this via wallet settings).
```
Ryun1, line 38:
```suggestion
When a user claims their staking rewards, their wallet will look up their delegation transaction and determine if they have opted-in to compensate their DRep (by checking the tx metadata). If metadata in the format described above is found, then the wallet will add an additional output to the reward withdrawal transaction that sends the intended percentage of their rewards to their DRep. If there is not enough ada in their reward balance to produce a output which satisfies minimum `lovelacePerUTxOWord`, the wallet will convey this to the user and let them decide how to proceed (or the user can establish how to handle this via wallet settings). 
```

Ryun1: someone can correct me, but I believe it is `lovelacePerUTxOWord` that must be satisfied now

Ryun1:
Hey @colll78 

Thanks for this proposal, im eager to see it develop.
I see a lot of value in such a proposal.

I have left a few review comments, 
primarily trying to make some of the text more specific.

Ryun1:
I do see one problem in that it is hard for wallets to know a DReps address, where to send the ada to?
Some DReps have added payment address in their metadata, as per CIP119, but not all.
Expecting wallets to index a DRep's latest metadata for this, I feel is a limitation of the design.

rphair:
@realdecimalist I've hidden the post above - a single dump of an AI critical summary of the CIP, apparently - since this posting and other like it will have no place in CIP reviews.  We need critical thinking and human experience, not linguistic analysis, and we definitely don't need conditioning from 100 different AIs in the global ecosystem each having a go at hundreds of reviewable CIPs in the next few years... so pending consensus from other editors we're trimming that branch of the tree preemptively.

The label of "Spam" is from the technically irrelevant mention of Talos as the review provider.  Note, functionally and rationally speaking, it's AIs that are supposed to act "on behalf of" humans: not the other way around.

colll78:
> I do see one problem in that it is hard for wallets to know a DReps address, where to send the ada to? Some DReps have added payment address in their metadata, as per CIP119, but not all. Expecting wallets to index a DRep's latest metadata for this, I feel is a limitation of the design.

The assumption is that DReps who do not have their payment address in their metadata as per CIP119, are opting out of this system entirely. There are valid reasons to do this, and many community members have reached out to me asking if there was a way to opt-out as a DRep (mostly because their local tax regulations would require them to track, claim and pay taxes on those payments as income). 

I agree it would simplify the implementation if we could obtain the payment address without requiring indexing the DReps metadata, but that would make it difficult to allow DReps to opt-out of this, so I will leave the decision up to community feedback here. 

Is it really valuable for this proposal to allow DReps to opt-out of compensation?

rphair:
@colll78 I know a couple of DReps that would want to opt out for social / ethical reasons alone.  In fact this preference would be part of their overall governance position, so I think they would consider the ability to opt out as integral to the DRep system as a whole.

TerminadaPool:
I recognise that DReps need to do work in reviewing proposals and deciding how to vote.  I also recognise that influence of voting behaviour can always be kept hidden through secret side deals.

Nevertheless, I feel uneasy about providing an official mechanism for tips or other forms of payments to DReps from constituents they represent because it might be perceived as encouraging or condoning financial influence.

colll78:
> @colll78 I know a couple of DReps that would want to opt out for social / ethical reasons alone. In fact this preference would be part of their overall governance position, so I think they would consider the ability to opt out as integral to the DRep system as a whole.

Sounds good, then I will leave the CIP as is, such that it expects wallets will obtain the DRep's payment address via their metadata and if it is not present there then the wallet will simply send the entire reward balance to the user upon withdrawal.

CIP-optional-drep-compensation/README.md
```diff
@@ -0,0 +1,58 @@
+---
+CIP: CIP-DRep-Comp
```
rphair, line 2:
```suggestion
CIP: 149
```

CIP-optional-drep-compensation/README.md
```diff
@@ -0,0 +1,58 @@
...
+    - Philip DiSarro <info@anastasialabs.com>
+Implementors: []
+Discussions:
+    - https://github.com/cardano-foundation/CIPs/pull/991
+Created: 2025-02-19
+License: CC-BY-4.0
+---
+
+## Abstract
+<!-- A short (\~200 word) description of the proposed solution and the technical issue being addressed. -->
```
rphair, line 16:
```suggestion
<!-- A short (\~200 word) description of the proposed solution and the technical issue being addressed. -->
```
(remove comment inherited from template scaffold)

rphair (commented):
This was well appreciated at Tuesday's CIP meeting — especially as an incremental step forward as a way of assessing a large potential change — and confirmed as a candidate.  Please @colll78 rename the directory to `CIP-0149` and update the proposal link in your original comment 🎉

Also, as resolved at that meeting, please tag representative wallet developers / architects for discussion of how to improve performance on wallets that don't use Blockfrost indexing as Lace does (as @Ryun1 pointed out) and maybe other potential optimisations discussed at the meeting.

colll78:
Thank you for the commits.

rphair (approved):
@colll78 I just looked this over in a monthly review of my own work & still think, as originally agreed with you & other editors, that this approach makes excellent sense to measure adoption & interest progressively so I think we should get it on the books so the community can begin with those live implementations.

I think with the recent format changes that everything else is good to go.  Apologies this was stale for a while but often if an author doesn't make noise about a pending CIP then we might not move it along to the final stages.

colll78:
> Hey @colll78 please can you make sure to rename the directory, to CIP-0149 so we can move this one along for merging

Done!

colll78:
Does anyone have a strong relationship with a major wallet that can get this integrated ASAP? 

If funding is an issue, please reach out to me and I will personally pay for the development hours required.

m-i-k-e:
This may be a great idea. I believe that some simple user research - asking DReps and delegators what they think of this idea - is the simplest way to start out.

Also, I think that it would be good to investigate several facets of this proposal:


1. We should try and estimate the amount of Ada that could be involved. In other words, will DReps be paid 100 Ada per month, or 1,000,000? Knowing this I think is important, as if the system is awash with cash, it will not look good.
2. DReps need to be able to both have a payment address and be able to opt out. If we don't provide that, then we are messing with a DRep's freedom of choice, and also this may be barred under EU regs (no automatic opt-ins).
3. When a Delegator changes DReps, GovTool (or other platform/wallet) needs to know about the 'donation' in place. The if the Delegator should have the option to pay the original DRep the accrued amount, transfer the payment to their new DRep, or cancel it altogether.
4. Delegators need to have a method of knowing (through GovTool or other interface) who they've delegated to, how much, and when it happened. Also, they need to know if the payment status/history of the donations, what the current accrued amount is, and the should be able to 'click to pay' that fee. If we make difficult UX, then: take up will be small, and the data on how popular the feature it is will not be reliable.
5. It would be good to know, in general, how often people reclaim their stake rewards. 
6. It would be bad for users if they are able to claim their stake rewards in a wallet that doesn't tell them that X percentage of it is going to a DRep. They could easily forget the donation.

aleeusgr:
I wonder why limit this feature to dRep?

As a delegator I want to be able to support any wallet so that I can help other members of the ecosystem.
It could be a developer, or an influencer, or a charitable organisation. why limit to dReps only?

rphair:
@aleeusgr the "support any wallet" semantic wouldn't be an extension of CIP-0149 because
* the specification takes the delegation transaction (not an account transaction) as a means of determining who the wallet holder might support;
* generally wallet holders haven't sent a transaction directly to their dReps.

So as far as this CIP is concerned, the addresses receiving funds from the wallet and the designated address of the wallet's delegated dRep are generally in two separate sets.  It would be a completely non-overlapping specification to provide a means for the user to deliberately choose an arbitrary address in a wallet: a different CIP with a different specification & different goals.

colll78:
> 2. DReps need to be able to both have a payment address and be able to opt out. If we don't provide that, then we are messing with a DRep's freedom of choice, and also this may be barred under EU regs (no automatic opt-ins).

The presence of a payment address in DRep registration metadata is explicitly an opt-in to accepting donations. The inclusion of a payment address in registration metadata for DReps is completely optional and users that choose to opt-in to include a payment address have explicitly consented to accepting donations. 

> 4. Delegators need to have a method of knowing (through GovTool or other interface) who they've delegated to, how much, and when it happened. Also, they need to know if the payment status/history of the donations, what the current accrued amount is, and the should be able to 'click to pay' that fee. If we make difficult UX, then: take up will be small, and the data on how popular the feature it is will not be reliable.

The point is that they don't ever 'click to pay' that fee, it just happens every time they claim their staking rewards (which typically if often not even done consciously it just happens whenever a user performs a transaction with their wallet and a reward balance is available). 

> 5. It would be good to know, in general, how often people reclaim their stake rewards.

It depends on the wallet that they use. Many popular wallets nowadays (Lace, Vespr, Eternl) automatically claim staking rewards whenever the user performs a transaction (ie. I click transfer, put my friends address and 5 ada, click send and the wallet automatically adds a staking balance withdrawal to the tx which is present in the transaction digest for the user to review before signing). So, users of these wallets claim much more frequently because for the most part, it isn't something they do consciously, it just happens whenever they perform a transaction with their wallet.  

> 6. It would be bad for users if they are able to claim their stake rewards in a wallet that doesn't tell them that X percentage of it 
is going to a DRep. They could easily forget the donation.

If a wallet doesn't have any tab / settings which conveys to the user their current DRep donation configuration then presumably the wallet wouldn't support this standard anyway and thus in such a wallet, no donation would occur anyway. But yes, if it needs to be said, wallets should be aware that they should present DRep donation configuration to users in some manner in their UI.

PR #1011, opened 2025-03-24
CIP-0153? | Plutus Core `MaryEraValue` Builtins

colll78:
We propose to introduce a number of functions for safely and efficiently working with `BuiltinData` as Value.

---

([current version](https://github.com/colll78/CIPs/blob/CIP-builtinvalue/CIP-0153/README.md))


Ryun1:
(sorry this is a little over my head)

Is this problem able to be solved in other ways, Im thinking if this could be more of a CPS rather than a CIP?

Ryun1:
For this proposal, please also consider the Plutus requires as defined via [CIP-0035](https://github.com/cardano-foundation/CIPs/tree/master/CIP-0035)

colll78:
> (sorry this is a little over my head)
> 
> Is this problem able to be solved in other ways, Im thinking if this could be more of a CPS rather than a CIP?

This issue cannot be solved in other ways, because the issue itself is the lack of a primitive type for `Value`, the introduction of which is the purpose of this proposal. The fundamental issue is that UPLC is not a general-purpose programming language. It is a language specifically designed to write validation logic to set constraints for the creation and transfer of Value and Data across the Cardano blockchain, and yet `Value` is not a primitive type, instead it is a standard library concept.

CIP-builtin-value/README.md
```diff
@@ -0,0 +1,138 @@
...
+4. The _keys_ in the inner map are ordered lexographically (`TokenName`s appear in ascending order)
+5. There are no empty inner maps (The _value_ associated with any given `CurrencySymbol` _key_ cannot be an empty map)
+6. There are no zero-quantity _values_ in the inner maps (The _value_ associated with any given `TokenName` _key_ in an inner-map cannot be zero).
+
+### Mary-era Value as a builtin type
+
+The only remaining solutions are to consider:
+
+1. Introduce builtin operations for efficient manipulation of `BuiltinData = Value` that take into consideration the invariants of ledger multi-asset values.
+2. Introduce a new builtin type `BuiltinMaryEraValue` with efficient Value operations that take into consideration the invariants of ledger multi-asset values.
```
ana-pantilie, line 56: What does `BuiltinData = Value` mean? A builtin Value type which is equivalent to `BuiltinData` would not be very efficient at all.

This CIP is missing some crucial information: if we are to introduce a new `BuiltinMaryEraValue` builtin type, how does it relate to `BuiltinData`, meaning I believe the specification is missing some transformations to and from `BuiltinData`.

We also need a proposal for the internal implementation of the `BuiltinMaryEraValue` type into the language. Based on this, the CIP needs to include a spec on the encoding and decoding of elements of this type to/from the `flat` encoding.

We need to also open a discussion on how this type and its builtins will be costed, and this also depends on the implementation.

Regarding the internal representation, @Unisay mentioned the https://github.com/jonathanknowles/monoidmap/tree/main/packages/monoidmap package which at first glance seems very interesting. I saw that there's also an example of a nested `MonoidMap`, which I guess could be appropriate in our case.

I believe we need to investigate multiple approaches on how to implement this type and discuss them on this CIP.

CIP-builtin-value/README.md
```diff
@@ -0,0 +1,138 @@
...
+2. Introduce a new builtin type `BuiltinMaryEraValue` with efficient Value operations that take into consideration the invariants of ledger multi-asset values. 
+
+
+## Specification
+
+We propose the addition of the following builtin functions: 
+
+All functions below work with:
+```
+type BuiltinMaryEraValue = BuiltinData -- or possibly (BuiltinList BuiltinCurrencySymbol (BuiltinMap BuiltinTokenName BuiltinData))
```
ana-pantilie, line 65: See above, we don't think that either representation (`BuiltinData`/`BuiltinList`) is suitable and this CIP needs to introduce `BuiltinMaryEraValue` as a new builtin type.

CIP-builtin-value/README.md
```diff
@@ -0,0 +1,138 @@
...
+
+Here are the important invariants of the PlutusCore representation of the Mary-era Value types from the Cardano ledger:
+1. All _keys_ in the outer map are unique (each `CurrencySymbol` appears only once)
+2. All _keys_ in the inner map are unique (each `TokenName` appears only once)
+3. The _keys_ in the outer map are ordered lexographically (`CurrencySymbol`s appear in ascending order)
+4. The _keys_ in the inner map are ordered lexographically (`TokenName`s appear in ascending order)
+5. There are no empty inner maps (The _value_ associated with any given `CurrencySymbol` _key_ cannot be an empty map)
+6. There are no zero-quantity _values_ in the inner maps (The _value_ associated with any given `TokenName` _key_ in an inner-map cannot be zero).
```
ana-pantilie, line 49: In the `plutus-ledger-api` there is a notion of `Value` and a very similar notion of `MintValue`. The invariants maintained by these types differ. How will we reconcile this?

We believe that adding multiple builtin types representing the different `Value` types would definitely not be ideal. @Unisay suggested implementing the difference on the ledger side: a new ledger language which would differentiate in the script context between the two, it would have a "minting" field and a "burning" field, both of the `Value` type.

This sounds fine to us, but I have a concern: since we introduced `MintValue` quite late, do you think that there are other kinds of "Value" which will need introducing? Is this specification of `Value` stable across multiple kinds of "Value"?

CIP-builtin-value/README.md
```diff
@@ -0,0 +1,138 @@
...
+3. `lookupCoin :: BuiltinCurrencySymbol -> BuiltinTokenName -> BuiltinMaryEraValue -> BuiltinQuantity`
+   - it returns the quantity of a given `Coin` in a Mary-era Value. 
+4. `unionValues :: BuiltinList BuiltinMaryEraValue -> BuiltinMaryEraValue`
+    - it merges all the values in the provided list
+    - when there are collisions it adds the quantities, if the resulting sum is zero it removes the entry to maintain the Mary-era Value invariants (no zero-quantity entries, no empty inner maps) such that the result is a normalized value. 
+    - this is not strictly necessary, however, it becomes extremely useful if we get limited higher order builtins, or even a `fieldMap` builtin. 
+5. `unionValue :: BuiltinMaryEraValue -> BuiltinMaryEraValue -> BuiltinMaryEraValue`
+    - it merges two provided values
+    - when there are collisions it adds the quantities, if the resulting sum is zero it removes the entry to maintain the Mary-era Value invariants (no zero-quantity entries, no empty inner maps) such that the result is a normalized value. 
+6. `valueContains :: BuiltinMaryEraValue -> BuiltinMaryEraValue -> BuiltinMaryEraValue`
```
ana-pantilie, line 85: In the `plutus-ledger-api`, we currently expose some [partial order comparisons on Value](https://github.com/IntersectMBO/plutus/blob/2596b5b93c100dc9c33c2848b37155cbb8503a21/plutus-ledger-api/src/PlutusLedgerApi/V1/Value.hs#L464). Is this enough to keep the same API? If we don't want the same API, why?

CIP-builtin-value/README.md
```diff
@@ -0,0 +1,138 @@
...
+
+## Rationale: how does this CIP achieve its goals?
+
+### Efficiency
+
+Builtins are strictly more efficient than CEK operations. 
+
+### Security and Abstraction
+
+The introduction of a canonical implementation of standard multi-asset value operations vastly improves ecosystem security and
```
ana-pantilie, line 98: I think this is missing the rest of the sentence.

ana-pantilie (changes requested):
Thank you for taking the time to write this up.

We (the Plutus Core team) have reviewed and discussed this CIP during our meeting today. I'm writing up some of my/our observations.

@zliu41 @Unisay @kwxm @effectfully please add anything I might have missed/you want to expand on.

CIP-builtin-value/README.md
colll78: I agree that a `BuiltinMaryEraValue` type would be preferable, but it would only be preferable if we can guarantee that we can have the values in script context all have this type when this `BuiltinMaryEraValue` is implemented. I opted for the `BuiltinData` definition because one of the issues I remember from the `BuiltinArray` discussion is that even if we get `BuiltinArray` the relevant script context fields will still be encoded as `BuiltinList` and we would need to use a builtin to first convert them to `BuiltinArray`. If we have to do this for `Value` in the script context I think this would be very very bad.

If we introduce `BuiltinMaryEraValue` can you confirm that the script context provided by the ledger would use this type where appropriate?

CIP-builtin-value/README.md
colll78: I was the one who made the feature request to add `MintValue`. 

Importantly, both `MintValue` and `Value` obey all the invariants I described above. I deliberately excluded the additional invariant "Every non-empty map must contain an entry for Ada, and the first key of any non-empty map must correspond to that Ada entry", I also excluded "all quantities must be positive` for similar reasons (and importantly for allowing subtracting values by unioning values with values that are negations of what you want to subtract). 

Both `MintValue` and `Value` will be represented by the `BuiltinMaryEraValue` builtin, in the same way that right now, they are both represented as newtypes around `BuiltinMap (BI.BuiltinList (BI.BuiltinPair BI.BuiltinData BI.BuiltinData))`. 

The only reason `MintValue` was introduced in the first place was because we needed a way at the type level to convey the difference in invariants between `Value` and `MintValue` and to discourage the use of certain operations on `MintValue`. When we introduce `BuiltinMaryEraValue`, this will not change. They will both have different newtypes which convey the difference in invariants to developers and expose operations accordingly.

The invariants of `BuiltinMaryEraValue` should not be all the invariants of the different values in the script context, instead they should be only those which must always be preserved across any operation on values. There is no benefit to having the `BuiltinMaryEraValue` impose invariants like `must contain ada` or `must not contain negative quantities of tokens`, because operations across values can break those invariants and then the result would not be representable by `BuiltinMaryEraValue`. If you union a value that contains only 10 ada with a value that contains only negative -10 ada the result should be an empty value. If you use the `deleteCoin` builtin to remove Ada from a `BuiltinMaryEraValue` the result should still be a `BuiltinMaryEraValue` (this is an extremely common use-case, ie you have some UTxO that only contains an minAda and NFT and you want the script to confirm the continuing output always must contain that NFT but allow the ada to change so that the script doesn't break from a protocol parameters update to Cardano, then you do `deleteCoin adaCS adaTokenName ownInputValue == deleteCoin adaCS adaTokenName ownOutputValue`).

>we introduced MintValue quite late, do you think that there are other kinds of "Value" which will need introducing? Is this specification of Value stable across multiple kinds of "Value"?

Yes, this specification is suitable for every type of value that currently exists, and any reasonable type of value that would exist in the future. The only way this would ever become unsuitable is if there is a fundamental change to the multi-asset model of the ledger, which is so deeply engrained into the code-base and architecture of Cardano at this point that such a change would never ever be considered.

CIP-builtin-value/README.md
colll78: I have never seen a production use-case that does any order comparison other than `valueEq`, `valueLessThan` or `valueGreaterThan` all of which can be easily expressed by `valueContains` and data equality on `BuiltinMaryEraValue`. Similarly, `checkBin` exists in Plutarch's standard library as-well, and there is not a single instance of it being used outside of to implement the order comparisons that I mentioned above. 

Not only is there no real use-case for `checkBin` but also implementing it as a builtin would be very difficult to cost and requires something akin to limited higher order functions which we will not get in PlutusCore for a very long time. I believe this just adds a ton of complexity with no real use-case. The only case I could see this being useful is if you want to check that a value is X amount greater than another value, but for this to be possible we need limited higher order builtins, and so I propose if this is something we want, we just propose this as an additional builtin down the line once we have limited higher order builtins implemented and leave it out of scope for this CIP.

rphair:
Thanks @colll78 and to @ana-pantilie + team for working out the details.  I couldn't attend this week's CIP meeting and hadn't heard that this was confirmed as a CIP there (i.e. "given a number") so I'll mark it as `Unconfirmed` unless indicated otherwise.

Please one or both of you post here when you think the review & proposal have converged on something the Plutus Core team would like to implement & then we'll get it on the CIP meeting review schedule again to confirm as a candidate.

effectfully (approved):
This is a great CIP: it addresses a vital issue and it's well-written. We (the Plutus team) should make it happen one way or another.

I feel like there should be mention of conversions to/from `Data` (which are trivial if `MaryEraValue` is represented as `Data`.

There's a bunch of terminology confusion like `BuiltinData` vs `Data`. The former is a Plinth concept, the latter is a UPLC and ledger concept, see e.g.

```haskell
evaluateScriptRestricting
    :: PlutusLedgerLanguage
    -> MajorProtocolVersion
    -> VerboseMode
    -> EvaluationContext
    -> ExBudget
    -> ScriptForEvaluation
    -> [Plutus.Data]
    -> (LogOutput, Either EvaluationError ExBudget)
```

So unless I'm missing something I don't feel like we need the `Builtin` prefixes. But I'm fine with them being in there, it's not very confusing and it's not worth nitpicking it too much.

Summarizing, we should implement this CIP.

CIP-builtin-value/README.md
```diff
@@ -0,0 +1,138 @@
...
+4. `unionValues :: BuiltinList BuiltinMaryEraValue -> BuiltinMaryEraValue`
+    - it merges all the values in the provided list
+    - when there are collisions it adds the quantities, if the resulting sum is zero it removes the entry to maintain the Mary-era Value invariants (no zero-quantity entries, no empty inner maps) such that the result is a normalized value. 
+    - this is not strictly necessary, however, it becomes extremely useful if we get limited higher order builtins, or even a `fieldMap` builtin. 
+5. `unionValue :: BuiltinMaryEraValue -> BuiltinMaryEraValue -> BuiltinMaryEraValue`
+    - it merges two provided values
+    - when there are collisions it adds the quantities, if the resulting sum is zero it removes the entry to maintain the Mary-era Value invariants (no zero-quantity entries, no empty inner maps) such that the result is a normalized value. 
+6. `valueContains :: BuiltinMaryEraValue -> BuiltinMaryEraValue -> BuiltinMaryEraValue`
+    - it strictly compares the two Mary-era Values and determines if the first value is a superset of the second.
+    - returns true if Value `a` contains each and every `Coin` in Value `b` with greater or equal quantities.
```
fallen-icarus, line 87: A common use case I have is to check if a UTxO has a specific quantity of multiple tokens from a specific policy ID. For example, the UTxO must have exactly 1 of: token_A, token_B, and token_C from minting_policy_1. If it has more than 1, zero, or an extraneous token from than minting policy, that would be a bug. So AFAIU `valueContains` is not useful to me since it does a `>=` check and it may allow extraneous tokens from a given minting policy. If other people have legitimate use cases for the current version of `valueContains`, then I would appreciate another builtin that makes it easy to verify the token quantities from a specific minting policy are exactly correct (eg, correct value, token names, and no extra tokens).

CIP-builtin-value/README.md
ana-pantilie: @effectfully @colll78 I don't understand how we could provide an efficient `Value` builtin type which is represented as `BuiltinList (BuiltinPair ... (BuiltinList (BuiltinPair ... ...))`. Other than the fact that an association list data-structure is a bad idea for representing maps, especially nested maps, each operation will also have to ensure that the `Value` invariants are maintained, which usually means traversing the nested lists multiple times. Sure, it will be more efficient than implementing them in the language, but even as builtins they're still going to be inefficient, so they will have high costs. Am I missing something here?

ana-pantilie: > I opted for the BuiltinData definition because one of the issues I remember from the BuiltinArray discussion is that even if we get BuiltinArray the relevant script context fields will still be encoded as BuiltinList and we would need to use a builtin to first convert them to BuiltinArray. If we have to do this for Value in the script context I think this would be very very bad.

Why? I would imagine translating `BuiltinList (BuiltinPair ... (BuiltinList (BuiltinPair .. ...)))` to some new `BuiltinValue` type would be better, because you only do it once. This translation would of course be a builtin function. Would the cost of this builtin function be worse than the costs of the other builtin functions combined?

michaelpj (commented):
I think this is okay. I would expect that we can implement some builtins to do these operations on `Data` pretty fast. I think the main question is whether or not we should have a separate builtin type for `Value`. I think that's basically preferable in all ways except for the problem of actually getting hold of them in the first place from the script context. We know already that people aren't willing to pay conversion costs. 

Possibly we could get away with a builtin that was internally something like:
```
data Value = 
   ValueFromData Data
   | StructuredValue (Map ByteString (Map ByteString Integer))

valueFromData :: Data -> Value
valueFromData d = ValueFromData d
```

and then if someone does an operation on a `Value` which requires us to traverse the whole thing, then we also take that opportunity to switch its representation. That gives us 0-cost conversions from `Data`, at the cost of some complexity.

colll78:
> I think this is okay. I would expect that we can implement some builtins to do these operations on `Data` pretty fast. I think the main question is whether or not we should have a separate builtin type for `Value`. I think that's basically preferable in all ways except for the problem of actually getting hold of them in the first place from the script context. We know already that people aren't willing to pay conversion costs.
> 
> Possibly we could get away with a builtin that was internally something like:
> 
> ```
> data Value = 
>    ValueFromData Data
>    | StructuredValue (Map ByteString (Map ByteString Integer))
> 
> valueFromData :: Data -> Value
> valueFromData d = ValueFromData d
> ```
> 
> and then if someone does an operation on a `Value` which requires us to traverse the whole thing, then we also take that opportunity to switch its representation. That gives us 0-cost conversions from `Data`, at the cost of some complexity.

I like this alot. My biggest issue with SOP / Data and conversions in general is that the heuristics around how / when conversion is worth it are an incredible amount of complexity that is imposed on developers who wish to become proficient in the domain of smart contracts. Additionally, the use of `conversion` functions like `unsafeFromBuiltinData` or similar (any function dedicated entirely to conversion of a given representation to another) is almost always bad practice. The reason is that if you are converting to a value faster representation (ie. you are willing to pay the cost of conversion) then you are likely doing so because you expect to perform many operations on the value, and you plan to access the value in its entirety (as opposed to only a few fields from it), and in such situations, you can avoid the vast majority of the cost of conversion, if instead of converting as a standalone operation, you perform the conversion simultaneously with the first operation you perform on that value. 

A common example is the `txInfoInputs`, their default representation is `BuiltinData ~ BuiltinList BuiltinData`. Often scripts will perform many operations on this list, for instance you might filter the list for all inputs from a given script `inputsFromCred :: (IsListLike list) => Credential -> list a -> list a`. You might think that because SOP list traversal operations are more efficient than current builtin traversal operations that you should convert to SOP first, but the conversion involves an entire traversal itself, and so often it might not be worth it to do `unsafeFromBuiltinData` to convert to SOP even if you plan to use the result frequently. However, in the above example we are already planning to traverse the entire list to filter it (`inputsFromCred`), and so in this case we can filter it AND perform the conversion in the same traversal (gets rid of redundant traversal and vastly reduces cost of conversion) ie with `inputsFromCred :: Credential -> BuiltinList BuiltinData -> [BuiltinData]`  (filters the list for inputs from a given script and converts the list to SOP in a single traversal). The issue with doing this at the standard library level is that it requires a massive amount of standard library code or some complex abstraction (which of-course sacrifices efficiency) to have functions for both operations and operations + conversions. For example, we define `map` and `mapWithSOPResult` and the same with `filter` and so forth. What you have proposed above, which abstracts this away at the builtin level, is a brilliant solution to this problem.

CIP-builtin-value/README.md
```diff
@@ -0,0 +1,138 @@
...
+
+1. Implement the new primitives according to the specification.
+2. Assign a preliminary cost to the new builtin functions. Consider similar operations and their current costs.
+3. Create variants of the [existing benchmarks][3] and potentially add some more.
+4. Check that the builtin operations over `BuiltinData = Value` are indeed significantly faster.
+
+If the preliminary performance investigation was not successful, this CIP should be revised
+according to the findings of the experiment. Otherwise, the implementation can proceed:
+
+5. Determine the most appropriate costing functions for modelling the builtin's performance
```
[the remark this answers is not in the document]
kwxm: The costing issue is an important one, and it might influence the design of the type and the functions operating on it.  We have to assign a costing function to each new builtin, and before a function is applied to some `Value` (`v`, say), we have to measure the size of `v` and estimate the CPU and memory costs of the function based on that size, and only proceed if there's sufficient budget left to pay those costs.  The nested structure of the `Value` type makes this very tricky.  Suppose you're using lists of pairs to implement maps, so `Value` would be something like (in Haskell notation) `[(CurrencySymbol, [(TokenName, Integer)])]`. If you want to look up the quantity `q` of token `t` for currency `c` , you first of all have to find `c` in the outer list, and you can work  out the potential cost of this by looking at the length of the list; however, you can't tell how long it'll take to find `t` in the inner list until you've found `c` in the outer list, so you'd have to do most of the work of looking up `q` in order to work out how much it'll cost to look it up, which kind of defeats the purpose.  I think this will need more thought, and we'll defintely need to take it into consideration when designing the functions.  One thing we could do would be to flatten the whole thing out and have a map of `(c,t)` pairs to quantities (so the cost would only depend on the number of such pairs), but would  it be possible to do this efficiently while still maintaining the invariants?   Maybe a Haskell `Data.Map` would be  suitable for this, in particular because it can tell you in constant time how big it is.  I haven't thought this through though, so it may well not be sensible.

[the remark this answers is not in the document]
kwxm: Anoother factor that might influence the design is what typical `Values` on the chain look like.  We could use some complex tree-based type, but if 99% of the `Values` on the chain have only 5 entries that it might be easier (and in fact more efficient) to just use lists.  I've just remembered that I analysed all of the scripts on the chain and looked at the structure of the `Values` they used: the results are described  [here](https://github.com/IntersectMBO/plutus/blob/master/plutus-ledger-api/exe/analyse-script-events/Values.md#values-appearing-in-transaction-outputs), and in fact most of them were pretty small. That was 18  months ago though, so it might be worth repeating the exercise.  Also, who can tell what might happen in the future?  If we implement something based on small numbers of tokens and then in a few years there are scripts that manipulate millions of them then it might be problematic.

kwxm (approved):
This is a very sensible proposal and I think we should very definitely do something along these lines.  Maybe it'll need some experimentation to converge on the best solution though.

I was also going to mention the blurring of the distinction between Plinth/Haskell and UPLC, and I see that @effectfully has already mentioned this.  Strictly this is about stuff that's supposed to be added to UPLC, but the notation is Plinth-like, using types and signatures that only appear in Haskell.  That's probably bearable though, and maybe it's actaully clearer, since eg, both `CurrencySymbol` and `TokenName` map onto the UPLC `bytestring` type where it's not clear what the bytestrings are supposed to represeent. A remark about this might be helpful though.

CIP-builtin-value/README.md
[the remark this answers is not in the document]
kwxm: Something else to worry about is how these things will be serialised.  If scripts are allowed to contain literal `Value`s then we have to  consider how they're converted into the `flat` format that gets used for serialised scripts. If you have enormous values in scripts then possibly the deserialisation cost could become large enough to be problematic. Maybe this isn't an issue, but we should think about it.  Perhaps we could prohibit serialisation of whatever internal format we end up using for `Value`s and require them to be serialised as `data` and then explicitly converted to `Value` during script execution with the conversion cost being paid for as normal rather than being done for free when the node deserialises scripts.

colll78:
> If you just use Data, though, you'll in general have no guarantee that a supposed Value object satisfies the invariants, or even that it doesn't contain irrelevant types of Data subobjects, like pairs. That would suggest that all of the operations would have to check that their arguments are well-formed, which might add a lot of overhead. Maybe we could supply a single function to check that a Data object is a well-formed Value and make the other functions less strict about checking, but then the overhead might be similar to converting to an opaque Value type.

For the vast majority of `Value` that a contract interacts with (`Value` obtained via the script context) all the invariants are guaranteed by the ledger (the ledger produces the script context, and in doing so guarantees the invariants of all values contained within it). The only time a `checkValueIntegrity :: BuiltinData -> Bool` function would be relevant (check if Data is a Value with all invariants) is if the Data in question comes from user provided content (ie. a datum or redeemer) and not from ledger provided content (ie. Value in all inputs / outputs / mint etc). We don't need to check the integrity of anything provided by the ledger. If the value is constructed within a script, the value will be correct by construction (otherwise the script author is malicious and there is nothing we can do about that no matter what fancy types we use). So the only case where such a function (that checks the integrity of a BuiltinData encoded value) is relevant is when the Value in question is from an external datum or redeemer, and this is very uncommon relative to interacting with ledger provided values and values that are constructed by the script itself.

CIP-builtin-value/README.md
fallen-icarus: Not equality on `Value`; I want equality on the inner `Token,Quantity` map. The `Value` would contain other `Token,Quantity` maps that I would ignore. So since `a` would contain tokens from other policy ids,  `a` and `b` will not actually be the same `Value` which means `a <= b && b <= a` won't work.

I guess there would need to be a builtin like:

```haskell
lookupCoinsFromPolicy :: BuiltinCurrencySymbol  -> BuiltinMaryEraValue -> [(BuiltinTokenName,BuiltinQuantity)]
```

And then I'd do an equality check on the result against what I'm expecting.

But thinking about my use case more, I usually want to check for equality of a specific inner `Token,Quantity` map while also seeing what else is in the `Value`. So I may still end up having to traverse the whole `Value` anyway...

rphair (commented):
Candidate status confirmed at today's CIP meeting... @colll78 please rename the containing directory to `CIP-0153` (and adjust the proposal link in the top comment accordingly) 🎉

michaelpj:
> this is very uncommon relative to interacting with ledger provided values and values that are constructed by the script itself.

I'm not sure if this is good enough. It depends what the consequences are. If they are bad enough then we are adding another subtle rule that all script authors need to remember: never put `Value`s in a datum or redeemer. It seems cheap to say people should just remember this, but these footguns add up over time, and generally it's a lot better if we can avoid them.

zliu41:
To summarize the options so far along with my thoughts:
1. No new builtin type - the builtin functions operate on lists. This feels like a non-starter - for instance implementing `unionValue` would require either converting the lists to Haskell `Map`s on the fly, or operating directly on lists, both of which are very expensive.
2. Adding a builtin type - this sounds quite reasonable. We should gather some data to confirm, but my intuition is that converting Data to Value with a builtin function isn't bad - it can be done in linear time, and it's certainly nowhere near the cost of converting Data to SOP ScriptContext with library functions. Sure, for large values the conversion is expensive, but operating on large Values without converting is also expensive. I think even just performing `unionValue` once, or a few lookups, on large values would be worth the conversion cost. (Also, note that the conversion isn't mandatory: the Values in ScriptContext are in the form of Data, so if you either don't use the Value at all, or only need to do something very cheap, you can still operate on Data directly)
3. @michaelpj's idea of switching representation from Data to Value only "if someone does an operation on a Value which requires us to traverse the whole thing". The problem is that without switching, every operation requires traversing the whole thing. So I don't know when it is a good time to switch.
4. Adding a builtin type, and adding a new constructor to Data, which elimintes the conversion cost. This may be doable - I'm not sure - but it sounds very tricky to add a new constructor to Data, while ensuring all existing things continue to work. Any idea @kwxm @effectfully @michaelpj @ana-pantilie regaring the feasiblity of this, and the steps required? Btw, if we are going to do this, we may as well also add an `Array` constructor for builtin arrays.
5. Adding a builtin type, and making a new Data type with a Value constructor. This sounds unappealing - we'll need to maintain two variants of Data, and it can only be used from Plutus V4 (that is if we can even do it in time for Plutus V4).

colll78:
> To summarize the options so far along with my thoughts:
> 
> 1. No new builtin type - the builtin functions operate on lists. This feels like a non-starter - for instance implementing `unionValue` would require either converting the lists to Haskell `Map`s on the fly, or operating directly on lists, both of which are very expensive.
> 

It would be significantly faster than current unionValue operations (which account for massive portion of budgets), but yes I agree this is a very bad option.
 
> 2. Adding a builtin type - this sounds quite reasonable. We should gather some data to confirm, but my intuition is that converting Data to Value with a builtin function isn't bad - it can be done in linear time, and it's certainly nowhere near the cost of converting Data to SOP ScriptContext with library functions. Sure, for large values the conversion is expensive, but operating on large Values without converting is also expensive. I think even just performing `unionValue` once, or a few lookups, on large values would be worth the conversion cost. (Also, note that the conversion isn't mandatory: the Values in ScriptContext are in the form of Data, so if you either don't use the Value at all, or only need to do something very cheap, you can still operate on Data directly)
> 

For me this is equally as bad as option 1. I agree the situation will be better than SOP for the reasons you mention, and that the heuristics of when to convert are much clearer, but there will still be situations where conversion is less efficient.

Imagine we have a value that contains 255 assets, and a given script execution only accesses the ada inside this value, then the cost of conversion + efficient lookup builtin will be incredibly expensive relative to just keeping the value as builtin data and only accessing the first element with
builtinLovelaceValueOf :: BuiltinValue -> Integer
builtinLovelaceValueOf b_v =
  BI.unsafeDataAsI $ BI.snd $ BI.head $ BI.unsafeDataAsMap $ BI.snd $ BI.head $ BI.unsafeDataAsMap b_v
Even if we aren't looking up ada, if we are only performing a lookup, in many cases it will still be more efficient not to convert.

It is a strong requirement for me that after we introduce this CIP, all value operations should always be done with whatever we introduce in this CIP and there should not be cases where people are still using builtin data nested maps to represent multi-asset-values. If the introduction of this CIP results in two representations for multi-asset-values, then it will have been a net negative. We do not want to introduce even more complexity, the whole point was to create a single canonical representation of multi-asset-values and the associated operations; if developers need to consider when to convert, and if there are situations where conversion would not be the most efficient option (which there will be) then we have just introduced more problems then we have solved. 

> 3. @michaelpj's idea of switching representation from Data to Value only "if someone does an operation on a Value which requires us to traverse the whole thing". The problem is that without switching, every operation requires traversing the whole thing. So I don't know when it is a good time to switch.
> 
Agreed.

> 4. Adding a builtin type, and adding a new constructor to Data, which elimintes the conversion cost. This may be doable - I'm not sure - but it sounds very tricky to add a new constructor to Data, while ensuring all existing things continue to work. Any idea @kwxm @effectfully @michaelpj @ana-pantilie regaring the feasiblity of this, and the steps required? Btw, if we are going to do this, we may as well also add an `Array` constructor for builtin arrays.
> 
If we can safely do this for both BuiltinArray and BuiltinValue this would in my opinion be the best option. Makes everything strictly more efficient, and creates a single canonical representation of multi-asset-values.

> 5. Adding a builtin type, and making a new Data type with a Value constructor. This sounds unappealing - we'll need to maintain two variants of Data, and it can only be used from Plutus V4 (that is if we can even do it in time for Plutus V4).

If 4 cannot be done, then from my perspective

michaelpj:
Thinking about the my representation-switching idea some more: it would be a limited win, because we would always need to cost our functions for the worst case, i.e. an unconverted value. So the advantages would be:
- Invariant-checking etc. is internalised and not the responsibility of script writers
- Processing of the (potentially large) `Data` objects happens entirely in Haskell, so will hopefully be fast and cheap even in the bad cases

> adding a new constructor to Data

I'm against this if we can possibly avoid it. The point of `Data` is to be a generic structured data type, and to map closely onto CBOR. Adding a constructor for `Value` makes it... I don't know what. Some kind of awkward union-of-anything type, where we just add random additional alternatives to cover random datatypes that turn out to be useful? I guess we could go that way, but it seems sad :shrug: 

> It is a strong requirement for me that after we introduce this CIP, all value operations should always be done with whatever we introduce in this CIP and there should not be cases where people are still using builtin data nested maps to represent multi-asset-values.

I agree that this is a good goal.

----

Here's another weird idea. The problem with having a builtin `Value` type is that it occurs inside the script context, which is `Data`, and `Data` can only contain nested `Data` objects. Well how about this?
- In addition to the script context object, we pass scripts a (builtin) list (or ideally array) of `Value`s.
- Inside the script context, where we would usually have a `Value`, we instead have an integer "value reference". The actual value is the corresponding entry in the aforementioned array of `Value`s.

Now, if we play our cards right, I think this means:
- All the advantages of having a builtin value type
- Constant-time "conversion" by indexing into the value array

zliu41:
> It is a strong requirement for me that after we introduce this CIP, all value operations should always be done with whatever we introduce in this CIP and there should not be cases where people are still using builtin data nested maps to represent multi-asset-values. If the introduction of this CIP results in two representations for multi-asset-values, then it will have been a net negative.

@colll78 I don't see it this way. I think it's par for the course in programming that the usual way of doing something is usually good enough, but in special cases, special things need to be done to make it more efficient. Examples: using different matrix representation/algorithms for sparse vs. dense matrix; using different sorting algorithms for small vs. large input.

This is also already the case with builtin array: you get a list from `Data`. You have to decide whether or not to convert it to Array - depending on what you are going to do with it.

zliu41:
> Here's another weird idea.

Weird indeed - this doesn’t really seem less weird than just giving Data a Value constructor.

The latter seems cleaner and more principled - expanding Data to fully represent script arguments by adding whatever constructors necessary.

The former sounds more hacky, since it makes Data a partial representation of script arguments, and compensates for the missing pieces (Value and Array) via references, just so that Data stays more close to CBOR, which I'm not sure why is particularly valuable.

> Adding a constructor for Value makes it

In the original EUTxO paper, it says Data "can be used to pass information into scripts", and adding Value/Array constructors fit this purpose. Again I'm not quite sure why it is important to make it resemble CBOR.

zliu41:
Another consideration is that if we go with option 5 (or Michael's alternative) for PlutusV4, then we should use option 2 for V1-V3, so we use the same builtin Value type and builtin functions. Option 5 may not be too bad especially since we can take the opportunity to also add a List constructor to Data.

colll78:
> > The BuiltinValue CIP is intended to introduce a single canonical representation of multi-asset values and its invariants. The goal isn't exclusively efficiency, it's also security, uniformity and a general improvement to the developer experience.
> 
> We'll have much more flexibility in Plutus V4, but for V1 through V3 we are basically stuck with options 1 & 2 (option 4 - changing the existing Data - sounds too fiddly to be feasible).
> 
> Between 1 and 2 I think 2 is less bad - yes accessing Ada only may be slower than option 1, but it at least offers an escape hatch for that, and there are many more scenarios where 1 is a lot slower than 2.
> 
> For Plutus V4 we can either introduce a new Data - but maintaining two builtin Data types can be a mess and point of confusion, or pass the Values separately as Michael suggested - but it feels like a hacky solution, and bundling it with a brand new Plutus version makes it worse. None of the options stands out and we'll need to weigh the tradeoffs carefully.

I don't mind at all if this `BuiltinValue` can only be used in Plutus V4 or newer. I would prefer that in-fact if it helps lead to uniformity.

kwxm:
>  Or we could simply add a new constructor to Data for storing Values

I'm starting to lose track of where we are since it's been rather a long conversation.  However it looks like we're converging on this being the preferred solution, along with a new `value` builtin type based on `Data.Map`.  Extending `Data` will complicate the costing process for `EqualsData` and (in particular) `SerialiseData`, which are already a bit problematic, but I don't think we should let that stop us: we'll just have to try it and see what happens.

If we're going with this plan then the CIP will have to be updated to reflect our conclusions (but it could maybe still mention the alternatives we've considered and why we've ended up with the eventual proposal), so I'll mark it as "changes requested".  Also, if we're changing `Data` then that'll have repercussions for the ledger code, so we'll need to get the ledger people involved as well in case they have any objections.

kwxm (changes requested):
The CIP should be updated to reflect the outcome of the lengthy (but productive!) discussions in the PR comments.

ana-pantilie:
Regarding the following:
 > Actually that needs to be included in the CIP, together with the encoding and decoding from flat and preliminary costing. See https://cips.cardano.org/cip/CIP-0035, section Additions to the Plutus Core Builtins.

We discussed this and I've opened a PR which should remove this requirement from Plutus CIPs: https://github.com/cardano-foundation/CIPs/pull/1038 So it's no longer necessary to provide this info in this CIP either!

As for the rest of the discussion, I think `Data`'s design does not suit the way it is used in general. I agree with making changes to it, however I'm not sure how it will be best to do this. I'm going to open a discussion with the team next week.

rphair:
Thanks @ana-pantilie @colll78 for pursuing the suggestions in the above https://github.com/cardano-foundation/CIPs/pull/1011#issuecomment-2883558850 and please keep us posted here... tagging `Waiting for Author` in the meantime.

ana-pantilie:
@colll78 I'm going to start working on the implementation in the second half of July. Do you have time to finalize the CIP until then?

colll78:
> @colll78 I'm going to start working on the implementation in the second half of July. Do you have time to finalize the CIP until then?

Yes, please reach out to me with any changes that I need to make to reflect what is being implemented.

ana-pantilie:
@colll78 I think the changes necessary are the following:
- we will add a constructor to `Data`: `Value BuiltinValue`
- we need to also add the following builtin functions: `valueData :: BuiltinValue -> BuiltinData` and `unValueData :: BuiltinData -> BuiltinValue`

colll78:
Updated accordingly!

ana-pantilie (approved):

CIP-0153/README.md
```diff
@@ -0,0 +1,179 @@
+---
+CIP: 153
+Title: Plutus Core Builtin Type - `MaryEraValue`
```
Ryun1:
```suggestion
Title: Plutus Core Builtin Type - MaryEraValue
```

Ryun1: remove backticks to follow title best practices :)

PR #868, opened 2024-07-31
CIP-0060 | v3 metadata restructuring

gstalamantes:
In this version, fields have been restructured to allow for improved indexing of token data.  Identifiers like `ipi`, `ipn`, and `isni` are now included within objects of their associated entity instead of separate fields.  Many optional fields are further defined to reduce interpretation.

PR #1018, opened 2025-04-09
CIP-0060 | Example track_number: change string to integer

arpradio:
The metadata in the example incorrectly used track_number as a string instead of the standard defined integer.  No changes needed to any of the other docs.

### DIFF
--- a/CIP-0060/README.md
+++ b/CIP-0060/README.md
@@ -130,7 +130,7 @@ Version 3 reorders identifiers like IPN, ISNI, etc into objects tied with the en
                         "song": {
                             "song_title": "<songName>",
                             "song_duration": "PT<minutes>M<seconds>S",
-                            "track_number": "<track#>",
+                            "track_number": 1,
                             "mood": "<mood>",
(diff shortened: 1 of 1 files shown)


rphair (commented):
Thanks @arpradio - This looks well justified to me but we should also get the approval of the last known person to update CIP-0060 and/or someone representing the stakeholders (maybe that could be you?) before merging - cc @sudoscientist @AndrewWestberg @Crypto2099

Strangely, at this moment, the last update to CIP-0060 is showing as a 404 page:
- https://github.com/cardano-foundation/CIPs/pull/868

Crypto2099 (approved):
This is just a grammatical change to make the examples match the written spec so this should be an easy merge.

Ryun1 (approved):
seems like a simple improvement to the example given, with no changes to the design, so no issues with backwards compatibility

rphair (approved):
Looks sensible & following @Crypto2099's confirmation that this doesn't change any behaviour.

PR #1021, opened 2025-04-13
CIP-???? | Orthogonal Certificates

yHSJ:
*NOTE: This CIP was initially suggested and discussed in person during the Node Diversity Workshop in Paris, France, April 7-9, 2025.*

---

The transaction is the most important interface in Cardano. Therefore, we should strive to make it as simple as possible without sacrificing functionality or security.

This CIP proposes a change to the ledger: replacing the certificates introduced in the Conway era with much orthogonal certificates, each scoped to a single action. This change significantly simplifies certificates, making them easier to use, document, and implement—which is especially important as alternative nodes are being implemented.

---
([rendered version](https://github.com/yHSJ/CIPs/tree/jshy/orthogonal-certificates/CIP-XXXX))

### DIFF
--- a/CIP-XXXX/README.md
+++ b/CIP-XXXX/README.md
@@ -1,0 +1,141 @@
+---
+CIP: ?
+Title: Orthogonal Certificates
+Status: Proposed
+Category: Meta
+Authors:
+    - Joshua Marchand <josh@securitybot.info>
+Implementors: N/A
+Discussions:
+    - https://discord.gg/tsr7H6ApdT
+Created: 2025-04-13
+License: CC-BY-4.0
+---
+
+## Abstract
+
+This CIP proposes a simplified design of stake managmenet certificate types focused on simplicity and usability.
+
+The current Conway era ledger specification includes multiple certificate types each representing different combinations of stake registration, delegation, and vote delegation. This proposal simplifies this specification by defining a set of orthogonal certificate types that can be composed to achieve the same functionality, making the ledger more intuitive and easier to document, use, and maintain.
+
+## Motivation: why is this CIP necessary?
+
+The Conway era introduces multiple certificate types for various stake operations including combinations of:
+
+- Registering a stake credential
+- Delegating stake to a pool
+- Delegating voting power to a DRep
+
+This approach results in many complex and convoluted certificate types, increasing complexity for developers and users. The number of certificate types would also grow combinatorially with new features, making it only harder to document, extend, and understand the certificates.
+
+By redesigning the certificate types with orthogonal certificates, we can:
+ - Make the system more intuitive for users and developers
+ - Simplify documentation by having self-describing, single-purpose certificates
+ - Make ledger implementations easier to maintain
+ - Avoid combinatorial growth of certificates with future features
+
+
+## Specification
+
+### Current Certificates CDDL
+The Conway era has multiple certificate types for stake related operations:
+```
+certificate = [stake_registration
+               // stake_deregistration
+               // stake_delegation
+               // pool_registration
+               // pool_retirement
+               // reg_cert
+               // unreg_cert
+               // vote_deleg_cert
+               // stake_vote_deleg_cert
+               // stake_reg_deleg_cert
+               // vote_reg_deleg_cert
+               // stake_vote_reg_deleg_cert
+               // auth_committee_hot_cert
+               // resign_committee_cold_cert
+               // reg_drep_cert
+               // unreg_drep_cert
+               // update_drep_cert]
+
+stake_registration = (0, stake_credential)
+stake_deregistration = (1, stake_credential)
+stake_delegation = (2, stake_credential, pool_keyhash)
+reg_cert = (7, stake_credential, coin)
+unreg_cert = (8, stake_credential, coin)
+stake_vote_deleg_cert = (10, stake_credential, pool_keyhash, drep)
+stake_reg_deleg_cert = (11, stake_credential, pool_keyhash, coin)
+stake_vote_reg_deleg_cert = (13, stake_credential, pool_keyhash, drep, coin)
+...
+```
+
+### Proposed CDDL
+While the current certificates may save a few bytes in the transaction when doing multiple actions in a single transaction, it greatly complicates certificates for both users and implementers.
+
+Instead, we propose a much simpler certificate design with orthogonal certificates.
+```
+certificate = pool_registration
+               / pool_retirement
+               / auth_committee_hot_cert
+               / resign_committee_cold_cert
+               / reg_drep_cert
+               / unreg_drep_cert
+               / update_drep_cert
+               / stake_registration_cert
+               / stake_deregistration_cert
+               / stake_delegation_cert
+               / vote_delegation_cert
+
+; New orthogonal certificates
+stake_registration_cert = (N, stake_credential, coin)
+stake_deregistration_cert = (N+1, stake_credential, coin)
+stake_delegation_cert = (N+2, stake_credential, pool_keyhash)
+vote_delegation_cert = (N+3, stake_credential, drep)
+
+; Others remain unchanged
+```
+
+### Semantics
+The orthogonal certificates creates a clear seperation of concerns, where each certificate has a clear, single responsibility:
+
+- `stake_registration_cert` - Registers a specified stake credential
+- `stake_deregistration_cert` - Deregisters a specified stake credential
+- `stake_delegation_cert` - delegates the stake controlled by a specified stake credential to a specified pool
+- `vote_delegation_cert` - delegates the voting power of a specified stake credential to a specified DRep
+
+### Handling Multiple Certificates
+To perform multiple operations, such as registration and stake delegation, multiple certificates would be included in a single transaction. The ledger would process certificates in the order that they are provided. If a stake delegation occurs before the registration of the specified stake credential, that delegation is invalid.
+
+
+### Example Usage
+To achieve the functionality of the current certificates:
+
+
+| Existing Certificate | Orthogonal Certificate Equivalent |
+| -------- | -------- |
+| `reg_cert`              | `stake_registration_cert`       |
+| `unreg_cert`            | `stake_deregistration_cert`     |
+| `stake_delegation`      | `stake_delegation_cert`         |
+| `stake_vote_deleg_cert` | `stake_delegation_cert` +  `vote_delegation_cert`   |
+| `stake_reg_deleg_cert`      | `stake_registration_cert` + `stake_delegation_cert`   |
+| `stake_vote_reg_deleg_cert` | `stake_registration_cert` + `stake_delegation_cert` + `vote_delegation_cert`  |
+
+
+## Rationale: how does this CIP achieve its goals?
+
+The orthogonal certificate provides **Conceptual Clarity** which has several benefits:
+1. **Usage:** The transaction is the fundamental interface for Cardano. It should be as simple as possible for users.
+2. **Documentation:** Each certificate is self-describing and encapsulates a single action, making documentation much easier to write.
+3. **Implementation:** Implementing the ledger should be as simple as possible–without losing features or security–to reduce the risk of an incorrect implementation resulting in a fork.
+
+While composite actions would require multiple certificates, multiple small certificates would only add a couple of bytes when compared to the composite certificates.
+
+## Path to Active
+
+### Acceptance Criteria
+
+- [ ]Acceptance of the Ledger team, as well as alternative node implementation maintainers (Amaru, Dingo, and others)
+
+## Copyright
+
+This CIP is licensed under [CC-BY-4.0](https://creativecommons.org/licenses/by/4.0/legalcode).


CIP-XXXX/README.md
```diff
@@ -0,0 +1,141 @@
+---
+CIP: ?
+Title: Orthogonal Certificates
+Status: Proposed
+Category: Meta
+Authors:
+    - Joshua Marchand <josh@securitybot.info>
+Implementors: N/A
+Discussions:
+    - https://discord.gg/tsr7H6ApdT
```
Ryun1, line 10:
```suggestion
    - https://discord.gg/tsr7H6ApdT
    - https://github.com/cardano-foundation/CIPs/pull/1021
```

CIP-XXXX/README.md
```diff
@@ -0,0 +1,141 @@
+---
+CIP: ?
+Title: Orthogonal Certificates
+Status: Proposed
+Category: Meta
```
Ryun1, line 5:
```suggestion
Category: Ledger
```

CIP-XXXX/README.md
```diff
@@ -0,0 +1,141 @@
...
+ - Simplify documentation by having self-describing, single-purpose certificates
+ - Make ledger implementations easier to maintain
+ - Avoid combinatorial growth of certificates with future features
+
+
+## Specification
+
+### Current Certificates CDDL
+The Conway era has multiple certificate types for stake related operations:
+```
```
Ryun1, line 42:
```suggestion
```cddl
```

CIP-XXXX/README.md
```diff
@@ -0,0 +1,141 @@
...
+stake_reg_deleg_cert = (11, stake_credential, pool_keyhash, coin)
+stake_vote_reg_deleg_cert = (13, stake_credential, pool_keyhash, drep, coin)
+...
+```
+
+### Proposed CDDL
+While the current certificates may save a few bytes in the transaction when doing multiple actions in a single transaction, it greatly complicates certificates for both users and implementers.
+
+Instead, we propose a much simpler certificate design with orthogonal certificates.
+```
```
Ryun1, line 76:
```suggestion
```cddl
```

CIP-XXXX/README.md
```diff
@@ -0,0 +1,141 @@
...
+2. **Documentation:** Each certificate is self-describing and encapsulates a single action, making documentation much easier to write.
+3. **Implementation:** Implementing the ledger should be as simple as possible–without losing features or security–to reduce the risk of an incorrect implementation resulting in a fork.
+
+While composite actions would require multiple certificates, multiple small certificates would only add a couple of bytes when compared to the composite certificates.
+
+## Path to Active
+
+### Acceptance Criteria
+
+- [ ]Acceptance of the Ledger team, as well as alternative node implementation maintainers (Amaru, Dingo, and others)
```
Ryun1, line 137:
```suggestion
- [ ] Acceptance of the Haskell Ledger team, as well as alternative node implementation maintainers (Amaru, Dingo, and others)
```

CIP-XXXX/README.md
```diff
@@ -0,0 +1,141 @@
...
+
```
Ryun1, line 138:
```suggestion

### Implementation Plan

N/A

```

Ryun1 (commented):
hey @yHSJ 

Thanks for the proposal
See a few minor editorial comments

Overall, I think this is well written and motivated proposal
One thing i may suggest adding is some numbers on how many times these 'combination certificates' have been used on mainnet
as I suspect that these certificates havent been widely adopted, maybe in-part because hardware wallets don't support them

CIP-XXXX/README.md
rphair: committing to ensure a properly indexed draft for meeting tomorrow.

CIP-XXXX/README.md
rphair: committing to ensure proper categorisation for meeting tomorrow.  @yHSJ - `Meta` is reserved for CIPs _about_ CIPs 🤓

CIP-XXXX/README.md
rphair: committing to ensure proper syntax display for meeting tomorrow.

rphair (commented):
@yHSJ @WhatisRT it's great to see that you've jump-started this with Ledger feedback already & looking forward to introducing this at tomorrow's CIP meeting (tagged `Triage`): https://hackmd.io/@cip-editors/110

CIP-XXXX/README.md
yHSJ: Thanks for the minor fixes @Ryun1!

Yes, @rphair, I obviously meant to change the category, but clearly forgot. I copied the header preamble from CIP-0001. Thanks for getting this committed

rphair: yes @yHSJ I thought it was an artefact like that 😅

CIP-XXXX/README.md
```diff
@@ -0,0 +1,146 @@
...
+Discussions:
+    - https://discord.gg/tsr7H6ApdT
+    - https://github.com/cardano-foundation/CIPs/pull/1021
+Created: 2025-04-13
+License: CC-BY-4.0
+---
+
+## Abstract
+
+This CIP proposes a simplified design of stake managmenet certificate types focused on simplicity and usability.
```
KtorZ, line 18:
```suggestion
This CIP proposes a simplified design of stake management certificate types focused on simplicity and usability.
```

CIP-XXXX/README.md
```diff
@@ -0,0 +1,146 @@
...
+reg_cert = (7, stake_credential, coin)
+unreg_cert = (8, stake_credential, coin)
+stake_vote_deleg_cert = (10, stake_credential, pool_keyhash, drep)
+stake_reg_deleg_cert = (11, stake_credential, pool_keyhash, coin)
+stake_vote_reg_deleg_cert = (13, stake_credential, pool_keyhash, drep, coin)
+...
+```
+
+### Proposed CDDL
+While the current certificates may save a few bytes in the transaction when doing multiple actions in a single transaction, it greatly complicates certificates for both users and implementers.
```
KtorZ, line 74: For completeness, it would be worth highlighting _how many bytes_ in the `Rationale` section, to show that the original motivation for introducing those numerous certificates was likely a bit misguided.

CIP-XXXX/README.md
```diff
@@ -0,0 +1,146 @@
...
+               / stake_registration_cert
+               / stake_deregistration_cert
+               / stake_delegation_cert
+               / vote_delegation_cert
+
+; New orthogonal certificates
+stake_registration_cert = (N, stake_credential, coin)
+stake_deregistration_cert = (N+1, stake_credential, coin)
+stake_delegation_cert = (N+2, stake_credential, pool_keyhash)
+vote_delegation_cert = (N+3, stake_credential, drep)
```
KtorZ, line 94: For backward-compatibility reason, it would be necessary here to specify the new indices, and makes sure that they don't clash with the current ones. This way, one could _virtually_ still deserialise a compound certificate into two orthogonal ones; and ensure that we don't needlessly break all transaction builders once again.

CIP-XXXX/README.md
```diff
@@ -0,0 +1,146 @@
...
+2. **Documentation:** Each certificate is self-describing and encapsulates a single action, making documentation much easier to write.
+3. **Implementation:** Implementing the ledger should be as simple as possible–without losing features or security–to reduce the risk of an incorrect implementation resulting in a fork.
+
+While composite actions would require multiple certificates, multiple small certificates would only add a couple of bytes when compared to the composite certificates.
+
+## Path to Active
+
+### Acceptance Criteria
+
+- [ ]Acceptance of the Ledger team, as well as alternative node implementation maintainers (Amaru, Dingo, and others)
```
KtorZ, line 138: This will also require a hard-fork, and the CIP can only be active once the hard-fork has been enacted.

CIP-XXXX/README.md
```diff
@@ -0,0 +1,146 @@
+---
+CIP: ?
```
```suggestion
CIP: 154
```

rphair (commented):
Confirmed as candidate at CIP meeting today, with some background given about how the proposal originated at a working group with the initial reviewers in common.  Therefore @lehins it would be great timing for you to also enter the discussion & review.

Please @yHSJ also rename the containing directory to `CIP-0154` 🎉 & update the proposal link in the top comment.

CIP-XXXX/README.md
lehins: There is no need to break existing specification. All you trying to do is remove some certificates, while leaving others unchanged.
```suggestion
stake_registration_cert = (7, stake_credential, coin)
stake_deregistration_cert = (8, stake_credential, coin)
stake_delegation_cert = (2, stake_credential, pool_keyhash)
vote_delegation_cert = (9, stake_credential, drep)
```
Otherwise we would have to keep old certificates for one more era to avoid breakage, just to replace certificates with equivalent ones

lehins (commented):
Here is my general view on this CIP: I am not really sure if it was worth writing the CIP, since this is fairly simple concern and a ticket on ledger repo would have probably suffice. That being said, there is nothing wrong with creating a CIP, since it will give more visibility to the proposed change.

In a nutshell, we already had plans on removing couple of old certificates, namely `stake_regestration` and `stake_deregistration`, since they were kept around temporarily for backwards compatibility with prior era and allowed users for easier transition to new ones that require a deposit. The only real new thing that that this CIP requests is to remove few redundant certificate types.

In other words, in the next era, we just want to remove all of these certificates: https://github.com/IntersectMBO/cardano-ledger/blob/49623962445143680dd725ebbf812c37e099b65c/eras/conway/impl/cddl-files/conway.cddl#L677-L687

Just to confirm, this is what we want plus some consistent naming for the next era from this CIP
```diff
certificate = [
-                stake_registration
-             // stake_deregistration
-             // stake_delegation
+                stake_delegation
              // pool_registration
              // pool_retirement
              // reg_cert
              // unreg_cert
              // vote_deleg_cert
              // stake_vote_deleg_cert
-             // stake_reg_deleg_cert
-             // vote_reg_deleg_cert
-             // stake_vote_reg_deleg_cert
              // auth_committee_hot_cert
              // resign_committee_cold_cert
              // reg_drep_cert
              // unreg_drep_cert
              // update_drep_cert]
```
If my understanding of this CIP is correct then you have my full support.

rphair (commented):
@yHSJ @lehins @WhatisRT I think the remaining Ledger confirmation was the main thing we were looking for so I'm marking this `Last Check` because I think @yHSJ all that remains are settling some cosmetic / clarity review points, given https://github.com/cardano-foundation/CIPs/pull/1021#pullrequestreview-2778302993.

Therefore this should be ready with plenty of time to spare, since due to Builder Fest overlap our next meeting isn't for 3½ weeks: https://hackmd.io/@cip-editors/111

I'm ready to approve this in the meantime if @yHSJ you can update _Path to Active_ at least according to these 2 suggestions (or equivalent):

https://github.com/cardano-foundation/CIPs/pull/1021#discussion_r2050699993 (time separation between relevance for Haskell vs. future nodes) 

https://github.com/cardano-foundation/CIPs/pull/1021#discussion_r2042264795 (clarity about whether / how a hard fork is required, even if "doing it anyway")

yHSJ:
> Here is my general view on this CIP: I am not really sure if it was worth writing the CIP, since this is fairly simple concern and a ticket on ledger repo would have probably suffice. That being said, there is nothing wrong with creating a CIP, since it will give more visibility to the proposed change.

Thanks for taking the time to review it! I agree, it's a rather simple change. That being said, I think it is important that, as an ecosystem, we move towards ledger changes being clearly documented in a CIP (or CIP like process) to help facilitate changes across all the different nodes. It's a big shift, but I truly believe it's important that the ledger is no longer owned by a single entity, but a shared resource. Also, this *should* be the first CIP in a family of "ledger simplification CIPs" that come about as a result of the node diversity workshop.

> In a nutshell, we already had plans on removing couple of old certificates, namely `stake_regestration` and `stake_deregistration`, since they were kept around temporarily for backwards compatibility with prior era and allowed users for easier transition to new ones that require a deposit. The only real new thing that that this CIP requests is to remove few redundant certificate types that allow register and delegate at the same time.

Yes, generally it's just removing redundant certificates to simplify the ledger (along with consistent naming). But I also (perhaps without stating it explicitly) suggested that we no longer wrap individual certificates in an array, as I don't see any reason to keep that, unless it can be justified.

> Just to confirm, this is what we want plus some consistent naming for the next era from this CIP

It looks like you're missing a few of the "composite" certificates in that diff. For clarity sake, I am suggesting:

removing:
```
stake_registration, (already happening)
stake_deregistration, (already happening)
stake_vote_deleg_cert,
stake_reg_deleg_cert,
vote_reg_deleg_cert,
stake_vote_reg_deleg_cert,
```

introducing (or renaming, the arrows here represent equivalency):
```
reg_cert -> stake_registration_cert
unreg_cert -> stake_deregistration_cert
stake_delegation -> stake_delegation_cert
vote_deleg_cert -> vote_delegation_cert
```

> There is no need to break existing specification. All you trying to do is remove some certificates, while leaving others unchanged. Otherwise we would have to keep old certificates for one more era to avoid breakage, just to replace certificates with equivalent ones

Well, if we are already going to break existing specification (i.e. remove the array wrapping each individual certificate), we could also break the existing certificates for the purpose of consistency. Having the indexes of each of these "account management" certificates be sequential just feels like a simpler API.

Ultimately, I am of the opinion it is better to have a more "painful" change in order to make the ledger as simple as possible for the future.

CIP-XXXX/README.md
```diff
@@ -0,0 +1,146 @@
...
+## Rationale: how does this CIP achieve its goals?
+
+The orthogonal certificate provides **Conceptual Clarity** which has several benefits:
+1. **Usage:** The transaction is the fundamental interface for Cardano. It should be as simple as possible for users.
+2. **Documentation:** Each certificate is self-describing and encapsulates a single action, making documentation much easier to write.
+3. **Implementation:** Implementing the ledger should be as simple as possible–without losing features or security–to reduce the risk of an incorrect implementation resulting in a fork.
+
+While composite actions would require multiple certificates, multiple small certificates would only add a couple of bytes when compared to the composite certificates.
+
+## Path to Active
```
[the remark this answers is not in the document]
yHSJ: I asked a clarifying question about this in the Cardano scaling discord, but I'm not sure why this has become the process only for some changes.

CIP-XXXX/README.md
yHSJ: Well, it really only requires implementation from the `cardano-ledger`. Since the haskell node is the majority of the network (as it's the only mainnet node currently), if it were implemented there, the rest of the nodes would be required to implement the change.

I added the rest of the nodes in an effort to include them in the decision process on this change (although it's a rather minor one) and, hopefully, to begin to include them in the decision process of future changes.

CIP-XXXX/README.md
yHSJ: Good point, added the *worst case* numbers.

CIP-XXXX/README.md
rphair: understood & looks great the way you have it after https://github.com/cardano-foundation/CIPs/pull/1021/commits/e893db97f89766f1130360f6ea848479512df40e ⭐

rphair (approved):
approved as per https://github.com/cardano-foundation/CIPs/pull/1021#pullrequestreview-2778722293 🚀

CIP-XXXX/README.md
lehins: I am weighing in. Renaming (changing the tag) will require old certificates to be retained for a whole new era. So, it makes no sense to introduce new identical replacement certificates. Please adjust the CIP to retaon the old tags.

Actual names of cddl bindings are irrelevant and can be discussed at a any point, since they do not require a HF.

lehins:
I just realized that we might never be able to remove those certificates, since some PlutusV3 scripts could depend on them. This means that if we completely remove these certificates, outputs that are locked by scripts that depend on these certificates being available:
```
stake_vote_deleg_cert,
stake_reg_deleg_cert,
vote_reg_deleg_cert,
stake_vote_reg_deleg_cert,
```
will become unspendable.

I am afraid that we might have to keep them around forever unless we have full support of all DApp writers out there that it is safe to remove these certificates.

The only safe thing that I can think of is to disable such certificates, unless a transaction uses PlutusV3 scripts. Which I am not sure is that much better, since tooling would still have to support all those certificates :cry:

lehins:
> please @lehins correct us if wrong about that, because otherwise we were ready to sign off on this).

Yes, I do confirm that if we remove these four certificates:

```
stake_vote_deleg_cert,
stake_reg_deleg_cert,
vote_reg_deleg_cert,
stake_vote_reg_deleg_cert,
```
there is a danger of locking funds in an unrecoverable manner, similar to what has happened in this issue: https://github.com/IntersectMBO/cardano-ledger/issues/5009 and was also discussed in #1028

We still can and will remove these two certificates:
```
stake_registration,
stake_deregistration,
```
since backwards compatibility is achieved through translation of newer certificates that have the deposit to plutus context. Same trick unfortunately cannot be applied to the aforementioned four certificates, since their equivalent representation requires multiple simpler certificates.

Naturally, we can and should also rename certificates in CDDL in more consistent manner.

>  if some mass deprecation warning were given to withdraw use of the affected certificates in advance of making this change...

There is certainly this option of communicating and warning the community about this. Question is whether this is worth the hassle. Naturally, this is not my call. My job is only to warn about the dangers and y'all deal with the community :grin:

WhatisRT:
I thought I had commented this earlier, but apparently I didn't. This backwards compatibility issue makes me think that the original suggestion from the node diversity workshop is better in this context because it doesn't suffer from that, being purely a CDDL change. I referenced it in a previous post already: https://github.com/cardano-foundation/CIPs/pull/1021#issuecomment-2801231871

To give a bit more detail, the original proposal there was to make the CDDL mirror the structure we have in the Conway spec for certificates, see e.g. https://github.com/IntersectMBO/formal-ledger-specifications/blob/conway-v1.0/src/Ledger/Certs.lagda#L76 This also avoids the issue of non-linear growth when we add features and has some other upsides that the proposal here as as well. It might be a less intuitive interface for people outside the functional programming space, but that seems like an ok price to pay for backwards compatibility.

Crypto2099:
Given that it's been more than two months without any movement against this CIP... can we get a status check from @yHSJ @WhatisRT and/or @lehins in regards to whether or not this CIP should be merged but marked invalid, changed to only reflect the renaming scheme, or if it still has a path to active?

cc: @rphair @Ryun1

rphair:
@Ryun1 @Crypto2099 -  @WhatisRT @lehins - just to make sure this proposal gets merged one way or another — according to the Ledger opinion about validity — since there is only that 1 issue remaining I've tagged this `Last Check` so we can keep reviewing that opinion biweekly, in case we don't get a ruling on that before the next meeting... with a note to change to `Status: Invalid` if necessary before merge: https://hackmd.io/@cip-editors/116

yHSJ:
I would argue there is still a path to active. I think this is still something worthwhile to invest some time into. While this particular change may be minor, I think it's part of a focus on simplifying the ledger that is important while we work on alternative node implementations (as brought up by @KtorZ in the Paris workshop).

As far as addressing the breaking of plutus scripts, I think there are a number of solutions. I know @KtorZ mentioned he had a solution to me in a private message, but never explained his solution. The naive approach would be to just translate the individual certificates into the old "complex" certificates when constructing a PlutusV3 script context, and then provide both the original certificates and the "translated" certificate(s). That would still potentially break scripts that rely on the presence of some combination of all of those certificates, but I think the likelihood such a script exists is rather low. Ultimately, it's a question for the community.

I will find some time early this week to write a query to find how many scripts have been executed on mainnet for the "complex" certificates. That will *overestimate* (as those scripts still may not rely on the complex transactions), but at least we'll have some rough numbers to consider.

lehins:
> The naive approach would be to just translate the individual certificates into the old "complex" certificates when constructing a PlutusV3 script context ...

That would be a pretty dangerous route.

Imagine a script that locks a staking credential and relies on presence of this multi-function certificate that can register and delegate at once. If you artificially combine a sequence of two certificates: `[cert_0, cert_1_register_A, cert_2_delegate_A, cert_3]` into `[cert_0, register_and_delegate_A, cert_3]`, you are now having serious issues:
1. Supplying a redeemer for all three `cert_1_register_A`, `cert_2_delegate_A` and `cert_3` is no longer correct, since index in redeemer pointer is off by one
2. Script is executed twice instead of once, unless we hack around it somehow
3. Script will not see what was actually supplied
4. We would inadvertently be preventing a potentially expected behavior, where such order of certificates: `[cert_0, cert_1_register_A, cert_2_delegate_A, cert_3]` was desired by the user, but ledger all of a sudden changed it from underneath them.

Most importantly it would have the opposite result: we would make the ledger implementation more complex with hacks like this, not simpler.

>  then provide both the original certificates and the "translated" certificate(s).

You can't do that, since that implies change to `PlutusV3` script context, which is not allowed.

> I will find some time early this week to write a query to find how many scripts have been executed on mainnet for the "complex" certificates.

The fact that you will be able to find all scripts that rely on such certificate on mainnet today, does not rule out all possible scripts like that from ever existing, because they might not have been used on chain yet. In other words someone could have registered a staking credential without providing the preimage of the hash, because today registration certificate does not require a witness. Also someone could have locked an input with a script that relies on a presence of such certificate, which you will not know until such input is spent and it might not happen for many years to come. So, this information cannot possibly serve as proof that your suggestion is sound.

I am now of a much stronger opinion that the change that this CIP proposes is dangerous.

yHSJ:
> Most importantly it would have the opposite result: we would make the ledger implementation more complex with hacks like this, not simpler.

Yes, this is a good point. That was just a stream of consciousness, trying to think of an example solution. You're absolutely right, that is a bad solution. I am curious to hear what @KtorZ's solution was.

> I am now of a much stronger opinion that the change that this CIP proposes is dangerous.

I think that is an unfair conclusion. Yes, the solution I just proposed was not a sound one, but that does not mean that a change that simplifies the certificates in general is dangerous.

WhatisRT:
> I am curious to hear what @KtorZ's solution was.

The original proposal was purely a CDDL change with no change in the ledger logic. Instead of having `stake_vote_deleg_cert, stake_reg_deleg_cert, vote_reg_deleg_cert` etc. there'd just be one certificate with optional fields. It would follow the [structure of the specification](https://intersectmbo.github.io/formal-ledger-specifications/site/Ledger.Conway.Certs.html#1812) just in CDDL form. This avoids the issue of combinatorial explosion (and in my opinion also improves on the other items in the motivation, but people are free to disagree).

rphair (commented):
@yHSJ the CIP editors resolved around a meeting a few weeks ago — when this was still up for `Last Check` but recently having the above irreconcilability pointed out — to give you and @lehins @WhatisRT some time to decide between yourselves how to go ahead with this proposal.

Our "backup plan" if this didn't happen — to document the premise of this CIP, while indicating that it can't (currently) be done & indexing this discussion to indicate _why_ and so avoid duplication of effort in other proposals — has been to merge this as `Inactive: (reason)` with a concise, mutually acceptable reason why it can't (currently) be done.

Please indicate whether you agree that we have gotten to this point.  I think it would be a great loss to simply close this PR: otherwise please all reviewers so far post your favourite short phrase for the "reason" this should be `Inactive` and editors will update it as such before merge, as planned for `Last Check` at the next meeting: https://hackmd.io/@cip-editors/118

lehins:
> The original proposal was purely a CDDL change with no change in the ledger logic. Instead of having stake_vote_deleg_cert, stake_reg_deleg_cert, vote_reg_deleg_cert etc. there'd just be one certificate with optional fields.

This would just introduce breakage for no good reason. Which means it would not be a simplification, but rather introduction of more work for everyone involved.

> It would follow the [structure of the specification](https://intersectmbo.github.io/formal-ledger-specifications/site/Ledger.Conway.Certs.html#1812) just in CDDL form

I get a 404. In any case, I can imagine what is in the link and again changing CDDL in a way that doesn't make it simpler and increases the size is counterproductive IMHO.

> Our "backup plan" if this didn't happen — to document the premise of this CIP, while indicating that it can't (currently) be done 

@rphair Some parts of this CIP can be done, in particular removal of registration and unregistration certificates that do not have a deposit in them are already in the process of being removed for Dijkstra era: https://github.com/IntersectMBO/cardano-ledger/pull/5137

yHSJ:
> Please indicate whether you agree that we have gotten to this point. I think it would be a great loss to simply close this PR: otherwise please all reviewers so far post your favourite short phrase for the "reason" this should be Inactive and editors will update it as such before merge, as planned for Last Check at the next meeting: https://hackmd.io/@cip-editors/118

I think in general the idea is one that is still worth exploring, but my proposed solution clearly isn't quite the right one. So whether that means we mark this as inactive or not is more of a organizational question.

rphair:
OK then @yHSJ if you wish to pursue this it sounds like we can merge a properly `Proposed` CIP (scratching the `Inactive` classification) if your proposal can be refactored according to what @lehins says in https://github.com/cardano-foundation/CIPs/pull/1021#issuecomment-3221891865:
> Some parts of this CIP can be done, in particular removal of registration and unregistration certificates that do not have a deposit in them are already in the process of being removed for Dijkstra era: https://github.com/IntersectMBO/cardano-ledger/pull/5137

I'll mark this `Waiting for Author` so this can persist with that footnote until such a reconciliation is made.  And if you have other ideas that would be helped with active review please post and/or recommend a CIP meeting discussion and we'll schedule it.

yHSJ:
@rphair It's pretty clear now 6 months later that this doesn't need to remain open. If a better solution comes along, it can be proposed separately and reference this (closed?) PR.

I don't want to be the one to close the PR, since I don't know if there is any additional process stuff to do before closing it, so I'll leave that for you

rphair:
Thanks @yHSJ for getting this technical scope on record.  If there ever comes a time when we might want to archive it more officially as an `Inactive` CIP as imagined in https://github.com/cardano-foundation/CIPs/pull/1021#pullrequestreview-3136380826, you or other interested parties can request re-opening this PR and we'll make those arrangements.

Issue #1017, opened 2025-04-07
Draft CIP process policy for AI generated content


Ryun1:
- currently there is no official guidance on the use of AI within writing CIPs and CPSs
- we should probably decide some guidance for authors

rphair:
Enthusiastic debate of the logistic and editorial questions behind using AI in the CIP writing begins for the first time publicly here:
* https://github.com/cardano-foundation/CIPs/pull/1009#issuecomment-2784700764

If that debate fails to cover any point that editors haven't made already in private Discord discussion in the last week... in response our first flag of AI writing in a submitted document... I have in my TODO to list those points here.  I'm not doing that straightaway since the vigour of this discussion looks like it will cover all of those points organically in the meantime.

We will transcribe a list of key recommendations one way or another soon enough.  #1009 is such an ideal test case for editorial standards & processes that I'd be inclined to generally follow whatever develops from the review of that CIP... at least in the first draft.

The result I think should be a Wiki page; probably item `22.` in this outline: https://github.com/cardano-foundation/CIPs/wiki/2.-CIPs-for-Reviewers-&-Authors

---

The relatively minor issue of what to do if AI output is presented as CIP **_review_** has already been suggested here; with so far no objection to the statement that it shouldn't be posted in CIP review threads:
* https://github.com/cardano-foundation/CIPs/pull/991#issuecomment-2672394318

nhenin:
In addition to what I outlined in https://github.com/cardano-foundation/CIPs/pull/1009#issuecomment-2784700764, I want to emphasize that, as someone deeply committed to advancing Cardano, I expect the CIP repository to uphold the highest possible standards of quality, both in CPSs and CIPs. Quality of Specification, Quality of Design, Quality of Communication.... If any tool allow you to improve something just use it... Editor should judge the quality at the end... 

Let’s challenge ourselves on the clarity of specifications, the simplicity of design, and the business value of each proposal. At the same time, let’s continuously improve the processes for writing these documents, as well as the diversity and talent of the people involved in reviewing them. Let's do things better, quicker and more efficiently ...

The situation is the same as before, only stronger... People didn’t wait for AI to start producing meaningless documents, they were already doing it. And those same individuals will continue producing such content, with or without AI. What’s changed is that technology amplifies human behavior. If someone creates value, AI can help them do it at scale. If not, it amplifies the noise.
So as before, we need to actively filter out this behavior...

In my opinion, there are several ways to detect  "BS" content in a document:

[...]

PR #1026, opened 2025-04-15
add note to CIP/CPS templates a optional disclaimer for AI/LLM application

Ryun1:
Attempts to partly address #1017

- added note to comments of template to nudge authors to self report AI/LLM application within their proposals

### DIFF
--- a/.github/CIP-TEMPLATE.md
+++ b/.github/CIP-TEMPLATE.md
@@ -51,9 +51,9 @@ It must also explain how the proposal affects the backward compatibility of exis
 <!-- OPTIONAL SECTIONS: see CIP-0001 > Document > Structure table -->
 
 ## Copyright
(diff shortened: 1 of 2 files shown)


Ryun1:
Happy to hear suggestions/ alterations to proposed wording, its difficult to succinctly capture the discussions from #1017

Crypto2099 (approved):
LGTM can probably wait a bit for further comment, if any

rphair (approved):
Definitely a step in the right direction & nothing incorrect about it: even if we choose to add things later this will be a help until then.

My only changes were to fix the grammar & make the reference to those HTML comments more explicit (which I think is better than moving the 2nd half of the instructions _below_ the copyright link choices).

Ryun1:
> My only changes were to fix the grammar & make the reference to those HTML comments more explicit (which I think is better than moving the 2nd half of the instructions below the copyright link choices).

good fixes, thanks Robert

PR #1029, opened 2025-04-24
CIP-0137 | Update local notification N2C mini-protocol

bolt12:
https://github.com/cardano-scaling/CIPs/pull/8


rphair (commented):
@bolt12 looks fine to me... @jpraynaud can you also confirm this?

jpraynaud:
> @bolt12 looks fine to me... @jpraynaud can you also confirm this?

Yes @rphair, we have made few improvements with @bolt12 in the **N2C Local Message Notification** mini-protocol in particular:
- to allow long lived connections (which is why we introduced a blocking mode in it) 
- and to inform the client if there are more messages to be notified for in the non blocking mode (with the addition of a `hasMore` flag in the response)

rphair (approved):

Ryun1 (approved):

PR #1031, opened 2025-05-01
CIP-0174? | Plutus Script Caching 

colll78:
## Summary

This CIP proposes introducing a memory-bound, Least Recently Used (LRU) in-memory caching mechanism within `cardano-node` to store deserialized Plutus scripts. The goal is to avoid redundant deserialization of scripts during transaction and block validation, particularly for frequently reused scripts (e.g., in DeFi or DAO contracts). 

By caching deserialized `PlutusScript` objects keyed by their script hash, and reusing them when present, nodes can significantly reduce CPU overhead without changing any ledger rules or validation outcomes. Scripts not found in cache will be deserialized on-demand (cold reload), preserving correctness.

This is a **runtime-only optimization** that applies at the `cardano-node` level. It does not require any changes to the `cardano-ledger` repository or protocol parameters, and is fully backwards compatible and deterministic.

The proposed cache would be:
- Bounded in memory (e.g., 256 MB)
- Eviction-based (LRU policy)
- Configurable via node settings

Importantly, if you consider the ledger history on mainnet over the past few months, the presence of such a caching mechanism with a bound of ~256MB would have reduced redundant deserializations to near zero. It also vastly increases the cost of spam attacks related to deserialization as you would need to execute unique scripts that are not in the cache which would require you to either submit each unique script as a reference script in an output to the chain (expensive large transactions) or to include the unique script in the witness set (increases the tx size and fee).

[Link to rendered proposal](https://github.com/colll78/CIPs/blob/0d95bf399bb58b718ef66ab58dad25533b05d3e5/CIP-Plutus-Script-Cache/README.md)


stevenj:
This looks like a very good idea.  Not sure if it should be in the CIP, but I would suggest there should always be some kind of metrics associated with LRU caches, tracking at least:

1. Actual space used. (Number of cached entries would vary depending on the size of the smart contract, and if its allocated 1G but only ever using 275MB then its probably too big).
2. Number of contracts cached (because they are different sizes so its not possible to know from the cache size).
3. Number of evictions (due to memory exhaustion of the cache, a big contract might need to evict multiple smaller ones to make space for itself in the cache).
4. Number of misses (a miss doesn't mean there is an eviction if there is memory available). 

So one can determine:
1. Is it effective.
2. Is it big enough.
Both of which may change over time.

I would also suggest a pre-load strategy when the node is first run, so the cache is already warm.  But again, it may not be necessary depending on how the cache behaves, so metrics would help one work out if pre-loading the cache is necessary or desirable.

CIP-Plutus-Script-Cache/README.md
```diff
@@ -0,0 +1,111 @@
+---
+CIP: ???  
+Title: Plutus Script In-Memory Caching  
+Category: Plutus  
+Status: Proposed  
+Authors:  
+  - Philip DiSarro <philip.disarro@gmail.com>  
+Implementors: []  
+Discussions:  
+  - https://github.com/IntersectMBO/cardano-ledger/blob/master/docs/adr/2024-08-14_009-refscripts-fee-change.md
```
rphair, line 10:
```suggestion
  - https://github.com/IntersectMBO/cardano-ledger/blob/master/docs/adr/2024-08-14_009-refscripts-fee-change.md
  - https://github.com/cardano-foundation/CIPs/pull/1031
```

CIP-Plutus-Script-Cache/README.md
```diff
@@ -0,0 +1,111 @@
...
+- Implementation merged into `cardano-node`
+- Benchmarks show measurable reduction in script deserialization cost under load
+- No deviation in transaction validation outcomes due to caching
+
+### Implementation Plan
+
+1. Modify Plutus script execution in `cardano-node` to check memory cache before deserialization
+2. Integrate cache via bounded LRU store
+3. Add config flags to control cache size and policy
+4. Test under synthetic load with high script reuse
```
rphair, line 107:
```suggestion
- [ ] Implementation merged into `cardano-node`
- [ ] Benchmarks show measurable reduction in script deserialization cost under load
- [ ] No deviation in transaction validation outcomes due to caching

### Implementation Plan

1. [ ] Modify Plutus script execution in `cardano-node` to check memory cache before deserialization
2. [ ] Integrate cache via bounded LRU store
3. [ ] Add config flags to control cache size and policy
4. [ ] Test under synthetic load with high script reuse
```
Adoption curve looks slow & progressive, and so I think checkboxes will be useful, but @colll78 that's up to you.  Note there is some talk the checkboxes on ordered lists [are browser dependent](https://stackoverflow.com/questions/78163737/how-do-you-create-a-numbered-checklist-aka-task-list-in-github-flavored-markdo#answer-78167926) but we could see how it goes (not tried in a CIP yet).

rphair (commented):
Looking forward to introducing this at the next CIP meeting (https://hackmd.io/@cip-editors/111) and hopefully the Plutus contingent will be there.

nau:
Quick analysis of evaluation of 1200 blocks of epoch 543 shows there were evaluated 504 unique scripts of all versions with total size of (flat serialized) scripts: 1,288,321 bytes. 
Guestimating 32 bytes per flat byte, we'd need cache of 32Mb for these scripts.

This cache is such an obvious optimization to do that I'm shocked it's never been implemented.

CIP-Plutus-Script-Cache/README.md
```diff
@@ -0,0 +1,112 @@
...
+
+```yaml
+PlutusScriptCache:
+  maxCacheSizeMB: 256
+  evictionPolicy: LRU
+```
+
+## Rationale: how does this CIP achieve its goals?
+
+This proposal follows the same spirit as performance optimizations in other smart contract chains (e.g. Solana's JIT caching model). It does not modify on-chain structures or Plutus semantics — only the **execution engine of node software**.
```
ch1bo, line 71: If it is not changing anything on-chain, is it even a CIP?

ch1bo (commented):
I love the idea, but not sure whether this is a CIP?

It's not changing how cardano works. While we _could demand_ such caching for all block producing node implementations because it is a CIP, other performance criteria are not covered by CIPs and some alignment needs to happen anyways in a multi-node implementation world.

colll78:
> I love the idea, but not sure whether this is a CIP?
> 
> It's not changing how cardano works. While we _could demand_ such caching for all block producing node implementations because it is a CIP, other performance criteria are not covered by CIPs and some alignment needs to happen anyways in a multi-node implementation world.

I created a CIP because I want to socialize this change and its impact to the Cardano developer community, and for it to become a hard requirement for all node implementations. 

The goal of this CIP is to be a first step towards getting rid of the non-linear scaling of fees based on the size of reference script bytes. Indeed, I could have proposed the reversion of that fee change directly in this proposal, however, given how risk adverse core developers tend to be, I figured it would be better to propose this as a standalone first step, and if accepted and integrated, then a second proposal which reverts the reference script byte fee scaling change with the justification being that this caching mechanism addresses the concerns that the fee scaling change was introduced for.

rphair:
When we created the `Network` and `Consensus` categories, each time there was a representative from that effort who had just submitted a CIP suited for that category... so we had the advantage of knowing that further CIPs coming in those categories could count on cooperation for review and to consider implementing them.  cc @bwbush @jpraynaud

If as @colll78 suggests (and I support the idea) this change should be presented for community endorsement, this could be given a category of `Node` (since according to my understanding it doesn't affect `Plutus` itself)... although like @ch1bo says we wouldn't yet have confirmation of "Node" cooperation like in the other two categories: further complicated by the fact that there will be multiple "nodes" in the future (though I'd propose any `Node` CIP category cover all of them).  cc @Ryun1

If there is any evidence that Haskell and/or Amaru node architects would generally be interested in what we used to require as [**enlisting**](https://github.com/cardano-foundation/CIPs/blob/master/CIP-0001/README.md#project-enlisting) in the CIP process, it would be helpful to see it here... and also to hear what more of them think about this particular proposal.

rphair:
@colll78 the CIP meeting today decided to leave this `Unconfirmed` until you can address these points...

We recognised @ch1bo's https://github.com/cardano-foundation/CIPs/pull/1031#pullrequestreview-2814023976 that particular software enhancements don't work well in the CIP process: generally because there's no opportunity for the community itself to confirm or support a "standard" and because such changes are more properly requested, documented & progressed as enhancements for that software project;

I mentioned that there could be grounds for including it as a CIP anyway given that Ouroboros improvements are also being documented as CIPs... but @Crypto2099 correctly pointed out that taking this approach on one node implementation wouldn't solve, or even be done the same way, on another (e.g. Rust vs. Haskell).

So it appears the "social" value of having this documented in a CIP is problematic since we wouldn't know how to make it into a usable standard.  Currently we can't even give it a category (not properly `Plutus` since it has nothing to do with the Plutus language itself) since `Node` seems a long way from being useful given the proliferation of node implementations.

Quantumplation:
IMO the CIP process likely needs an overhaul. We discussed this at the node diversity workshop, observing that Ethereum has:

EIP - improvements to the *protocol itself* that all nodes must agree on and adopt (if approved)
ERC - documenting community standards/patterns that are widespread or others might want to follow, but don't neccesarily impact the core node

(There might have been a third one, but I can't remember the details!)

So perhaps we need:

CPS - generic problem statements that might have a variety of possible solutions
CIP - specific changes to the *protocol* that all nodes would need to agree on and adopt (if approved)
CRC - community standards, best practices, etc.

The way I would expect Phil's goal to progress through these then is:
- a CPS, describing why the ref script bytes is highly detrimental and needs a solution to remove
- A CIP that outlines a *protocol level* performance benchmark that we propose all nodes must meet (regardless of how they do so); meeting that *protocol level* performance benchmark would allow us to make changes to remove the ref script bytes
- One or more CRCs with suggested architectural design patterns that nodes could use to meet the performance benchmark and make it safe to adopt the CIP

rphair:
thanks @Quantumplation ... let's continue that suggestion here:
* https://github.com/cardano-foundation/CIPs/issues/1040

colll78:
After discussion with @Quantumplation I'm on board with his approach, I will edit this CIP to provide concrete constraints on the amount of work nodes do for large adversarial payloads targeting reference script deserialization that ensures that the amount of work done by nodes to process such is negligible (can be absorbed into the network) thus we can justify the removal of the fee.

Likewise, I will introduce another CIP for the ledger category that proposes that the fee per reference script byte change is reverted once a node is conformant with the first CIP. 

I already maintain so many CIPs, so if possible, can someone else take the responsibility of drafting the CPS describing why the ref script bytes is highly detrimental and needs a solution to remove.

WhatisRT:
The issue with getting rid of the reference script fee is that you can still have attacks that maximize cache misses, in which SPOs would do a lot of work but aren't rewarded for it. I'd really like to get rid of this fee (especially that hideous recursive exponential) but the work in the worst-case scenario needs to be rewarded for somehow.

My suggestion to avoid this fee back then was to ensure that the operation required to get the script into a runnable state would be really cheap. This is relatively easy to do if you have a data structure without pointers because you can just read it into memory from wherever you store it, but Plutus scripts aren't such a data structure. I suggested a [GHC compact region](https://hackage-content.haskell.org/package/ghc-compact-0.1.0.0/docs/GHC-Compact.html) which sounds like a fantastic tool for exactly this purpose but apparently it doesn't quite work for Plutus scripts. I also don't know whether this problem is easy to solve in other languages like Rust (I guess you could just have some binary encoding with 'local pointers' and when reading it in all you do is allocate memory for every indirection & substitute in the pointer to that allocated memory).

So I think it is possible in principle to avoid this fee, but you really need to have fast loading of all scripts and not just a cache. At least if you want SPOs to be sufficiently paid for their work in the presence of attacks.

colll78:
Firstly, I agree with you 100% that we should work to reduce the cost of deserialization / setup time for scripts in the first place, that will benefit us regardless. 

However, I think we can get rid of the reference script bytes fee with a cache + a high one-time fee for reference script deployment. In Solana for instance, to deploy the equivalent of what would be a 100kb reference script on Cardano, the transaction fee incurred 89 SOL or roughly $14,906. 

On Cardano, instead of charging a lot for reference script deployment, the transaction fee to deploy a ~16kb reference script is roughly 1 ada, and we prevent bloat by requiring a large min-ada deposit to compensate for the low fee, the min-ada required for a UTxO with a ~16kb reference script is roughly ~68 ada. The issue is that users can simply reclaim this ada at any-time by sending the reference script UTxO back to themselves (very low tx fee). Instead, we should make the transaction fee for deploying such large reference scripts be non-trivial. Developers would unilaterally prefer to pay a very high one-time fee to deploy a contract in order to reduce the fee per transaction that uses their contract (reference script byte fee). 

With the above, it would cost hundreds of thousands of dollars for a malicious attacker to construct malicious payloads to attempt to create cache misses. 

Another future solution is to allow reference scripts to be submitted as blob transactions and cost those transactions accordingly. Blobs will likely be released in the next year or two and will allow publishing large amounts of data directly to the chain outside of the UTxO set but such that it can still be referenced in the execution layer and accessed by the ledger.

WhatisRT:
Sadly this doesn't quite work because an attacker doesn't need to deploy scripts themself to cause cache misses. As long as there are enough reference scripts around to exceed the size of the cache + what you need to fill up one transaction the attacker can just cycle through all of them. And since the attack doesn't require succeeding scripts then can all just fail them really quickly by providing `0` execution units leading to low costs. An attacker would prefer larger scripts because these lead to lower costs but it's not necessary.

Sorry for always being the guy saying that something doesn't work 😅 But I really like these suggestions and if you have more thoughts I'd like to hear them!

zliu41:
There are so many cache replacement policies that it's not obvious why LRU is the best.

If your goal is reducing fees, then I'd suggest a random replacement policy, or at least have some degree of randomness. Like @WhatisRT said - any deterministic algorithm would be subject to certain access patterns that make it perform poorly.

WhatisRT:
> I disagree. Why would the cache store failing scripts? A transaction with a script that fails in phase 2 consumes collateral, which can of-course be costed higher than a legitimate transaction (and should be, because there is no reason any honest actor would ever submit such a transaction).

The cache has to store failing scripts because the attack gets even easier if it doesn't. Remember that cache misses are the issue here, and so if I know my scripts aren't going to be cached I can just do many transactions with exactly the same scripts.

And yes, you could not cache the scripts and charge a higher collateral that makes up for this difference. That means that you're now moving the complexity from the fee calculation to the collateral calculation. This is a bit nicer for the users because the actual fees are lower, but it doesn't actually make things easier.

Another issue is that you can also do this attack with succeeding scripts. I don't know how viable it is, but I'm sure you could write some tooling that finds scripts on mainnet that you can use for your attack. So if this attack is viable and there are enough honest scripts that you can abuse then all of these optimizations won't save you.

> I think to state that this doesn't work, we need more concrete proof, because this is how Solana addresses the exact same problem and allows them to not cost deserialization. They use an LRU cache with a high degree of randomness baked in. If this doesn't work, then the Solana mainnet is vulnerable to the same exact deserialization attack.

Do you have a source for this? I wonder if Solana just has a massive cache size which they can do because they require beefy machines. Randomness also helps a lot here because you need a much larger pool of scripts to choose from to get a large amount of cache misses. And if you only have a couple of cache misses then there's no attack.

However, all of this requires benchmarking & simulation. Our standard of proof shouldn't be 'Solana does it so it must be fine'. They have different hardware requirements, different consensus and lots of other differences, so what works for them might not work for us.

colll78:
> The cache has to store failing scripts because the attack gets even easier if it doesn't. Remember that cache misses are the issue here, and so if I know my scripts aren't going to be cached, I can just do many transactions with exactly the same scripts.
>
> And yes, you could not cache the scripts and charge a higher collateral that makes up for this difference. That means that you're now moving the complexity from the fee calculation to the collateral calculation. This is a bit nicer for the users because the actual fees are lower, but it doesn't actually make things easier.

Introducing very high fees for transactions that fail phase 2 validation is in and of itself a good change that encourages the intended behavior of the network. It is impossible for an honest party to submit a phase 2 invalid transaction unwillingly. There are close to zero of such transactions throughout the history of the Cardano mainnet and all of them have been submitted deliberately for testing or otherwise. The "complexity" for collateral calculation should be simple relative to the complexity of fee calculation, because the goal for collateral calculation is just to make it very high, as we know that an honest user should never have their collateral consumed. 

>However, all of this requires benchmarking & simulation. Our standard of proof shouldn't be 'Solana does it so it must be fine'. They have different hardware requirements, different consensus and lots of other differences, so what works for them might not work for us.

I agree that shouldn't be proof that this solution will work for us, my point was that given this is the case, we cannot immediately conclude that this solution will not work for us. It provides an argument for why we should explore this as a potential solution on Cardano.

Ryun1:
hello!

Cool proposal
but im not sure how well it fits into the current CIPs processes, as mentioned a couple times above.

This proposal firmly seems to be more of a implementation specific detail,
rather than a standard that is needed for all core node implementations.

I see this in a similar vein to LSM Tree / UTxO-HD, where it is a optimisation that a node implementer may or may not implement.
There is no required changes to any part of Cardano's core categories which would require standardisation.
 
Although I feel this proposal doesn't exactly fit into the current CIP landscape
pending discussions on CIP process improvements via #1040 
a place for this sort of proposal could be made within CIPs

while discussions of CIP process evolution continue
i feel its best to leave this proposal as is, allow discussions to continue
but without a number

Crypto2099:
On the topic of whether or not this CIP is "on topic" I would like to say that I've loved the conversation that has already occurred within this thread but, it raises and important point, that being:

- How do we socialize potential ideas (maybe we should cache scripts) and then move into some amount of effort expended (testing, simulation) to justify further expense of effort to get to implementation because an idea is shown to have merit?

Also I do like the idea of creating a standard for `Node` implementers because it may be that we have identified this caching/serialization issue as problematic to block producers and therefore all must be able to handle a transaction with 27 different reference scripts within 500ms (whatever, making up numbers here) and that becomes a benchmarking test for whether or not a block-producing node is compliant (something for Cardano Blueprint maybe @ch1bo?).

I'm not satisfied with @colll78's argument that "Solana does it this way so we should too!" (I know that's reductive and not exactly what he said, don't kill me) but I'm also not convinced by @WhatisRT arguments that this one solution may lead to other issues so it's not worth considering. We've already rather poorly traded one issue (i.e. a DoS vulnerability from the deserialization of reference scripts) with another issue (extremely high transaction costs for transactions that have legitimate need for multiple reference scripts). 

Finding solutions that can bring the "per tx" fees for end users down are admirable and should be supported so I would love to see how we move this into a "next phase" for testing and simulation to see whether Phil's performance assumptions are true or false.

WhatisRT:
Ah, sorry, I didn't want to give the impression that some things aren't worth considering. We just need to make sure that we don't allow for attacks of this kind, which requires careful analysis.

Let me also suggest a hybrid approach: If the attack becomes harder to execute, we could lower the fees. Maybe a randomized cache combined with some other strategies makes this attack already expensive on the attacker, and then the fees could be lowered to compensate for this. I don't have the time to look into this, but it might be worth a shot if anyone is interested.

berewt:
> Also I do like the idea of creating a standard for `Node` implementers because it may be that we have identified this caching/serialization issue as problematic to block producers and therefore all must be able to handle a transaction with 27 different reference scripts within 500ms (whatever, making up numbers here) and that becomes a benchmarking test for whether or not a block-producing node is compliant (something for Cardano Blueprint maybe @ch1bo?).

I'd love those node standard to be expressed in term of requirements rather than implementation guidelines, for many  reasons actually:
1. You allow each node dev to pick the solution they want.
2. Don't have to change the spec if you find a better solution.
3. You can more easily test if you meet the standard.

I love the CIP, but it sounds like a glorified Cardano-node issue. And unfortunately it's probably a good idea to bring it here because it has more visibility and feedback than as an issue. So yes, we probably need to find a way to ease these types of discussion. My understanding was that it could have been the role of the intersect working group, but it might haven't succeed enough (yet?).

colll78:
> I'd love those node standard to be expressed in term of requirements rather than implementation guidelines, for many reasons actually:
> 
> 1. You allow each node dev to pick the solution they want.
> 2. Don't have to change the spec if you find a better solution.
> 3. You can more easily test if you meet the standard.
> 

Yes, I'm okay with allowing the standard to be expressed in terms of requirements, if the end result is that the costing of the implemented solutions is uniform such that the graph of the fee structure matches the work done in the implemented solutions.

> I love the CIP, but it sounds like a glorified Cardano-node issue. And unfortunately it's probably a good idea to bring it here because it has more visibility and feedback than as an issue. So yes, we probably need to find a way to ease these types of discussion. My understanding was that it could have been the role of the intersect working group, but it might haven't succeed enough (yet?).

This isn't a cardano-node issue, this is an issue for every node implementation (all alternative nodes also suffer the same issue).  I would agree that this proposal is definitely overly prescriptive (explicitly choosing a specific type of cache and size), and that it should instead have simply suggested the introduction of a cache and some requirements and leave the details of that decision to individual node implementations, 

for:
> 2. Don't have to change the spec if you find a better solution.

I think we don't actually want this. We want the spec to change such that it always reflects the best solution so that the best solution is clearly documented and socialized to all node implementations. The network fee costing of all work must always be set to match the least efficient implementation of that work across all nodes. This means we have a huge incentive to enforce that all nodes implement the most efficient solution to any given problem and if you don't socialize / include such solutions in a spec then nodes will significantly lag behind each other (as you can see in ethereum where some clients are 150% more efficient than others) because by default nodes have a huge incentive to not socialize these optimizations because by nature they are in competition with the other nodes (so each node wants to be the most efficient). An alternative solution is to always have the standard expressed in terms of requirements where the requirements match the efficiency achieved by the current most efficient solution, and to include documentation of that solution in the standard without prescribing it (such that if a node implementer finds an equivalently efficient solution that meets those requirements they can use that).

rphair:
@colll78 I've written this comment based on our discussion of this proposal when it came up at the 06 January meeting.  I think if we could come up with an extended CIP classification of "design document" as described here, we would be able to accept this into the CIP repository without leaving it entirely to project issue queue(s).  Your participation in this issue would be very welcome (possibly using this PR as an illustrative case):
* https://github.com/cardano-foundation/CIPs/issues/1076#issuecomment-3825818481

colll78:
Can someone just implement this in the node? It takes like a handful of days of engineering work max, and the benefit is absolutely massive. 

Why do we need to pontificate on this for so long?

rphair:
@colll78 I guess you are calling https://github.com/cardano-foundation/CIPs/issues/1076#issuecomment-3825818481 "pontificating" which would be pretty extreme as well as unproductive.  You'd be a top choice to help us settle how to create a CIP category for node (or network) performance (or architectural) issues: especially since this CIP would _either_ belong to such a category *or* no category at all & be inadmissible here.

Here & in the CIP as currently written, you're talking about "the" node as if assumed to be the Haskell node (unless I've missed something), and nothing about your document explains how other node projects would follow suit in implementing this.  If it's irrelevant outside the Haskell node or environment then your document needs to say so.

If other nodes _would_ benefit from doing this, you also need something in the _Path to Active_ how & when they would become involved.  If they're _not_ involved, I would wonder what the Haskell node devs would say when this spec showed up as a CIP and not as a feature request on their repository.

> Can someone just implement this in the node? 

I think that's a great idea and I would like to see it happen with a minimum of writing and bureaucracy.  Unfortunately the CIP editors can't implement it... we can only help you create a CIP.  But if you think a CIP draft pull request is a better way of getting attention on this improvement than using a "node" feature request, you're welcome to leave it here.

We might also be able to merge if updated for a multi-node environment _and also_ if we could assign a proper category for it.  We'd need to agree on a **new category** for it though — since `Plutus` is inapplicable — and so it would be unfortunate for all of us if you're written off that discussion because it doesn't suit your short-term goals.

rphair:
@perturbing @Ryun1 to explicitly review the questions of 1) whether we can give a CIP category to this (by creating one), and 2) whether there would be any adverse reaction for including this as a CIP rather than insisting it's a node(s) feature request, I've put this up for Review again at tomorrow's meeting: https://hackmd.io/@cip-editors/127

@colll78 I hope you can come especially if you think that my comment above or https://github.com/cardano-foundation/CIPs/issues/1076#issuecomment-3825818481 are not a useful description of the problems this PR is facing.  I would be happy to get more of your ideas on record here about what the life cycle of this CIP would be like & hopefully use this to assign some new, appropriately named category.

Ryun1:
>  whether we can give a CIP category to this (or create one)

Happy for this to get a new category

Traditionally, this sort of proposal hasn't been a CIP as it is more of a non-functional requirement for node implementations
but I feel it would be beneficial for this and similar proposals to move via the CIP process going forward

colll78:
> proposal hasn't been a CIP as it is more of a non-functional requirement for node implementations
> but I feel it would be beneficial for this and similar proposals to move via the CIP process going forward

My comment wasn't in response to the previous comment; it was just a general expression of pain from the fact that the node team has not implemented this yet into the Haskell node (to no fault of the CIP editors). 

I agree that we need a new category for nodes in general. I hope that getting this accepted as a CIP will cause the implementers of the only live mainnet node (Haskell node) to review this and implement it.

CIP-Plutus-Script-Cache/README.md
```diff
@@ -0,0 +1,112 @@
+---
+CIP: ???
```
rphair:
```suggestion
CIP: 174
```

CIP-Plutus-Script-Cache/README.md
```diff
@@ -0,0 +1,112 @@
+---
+CIP: ???  
+Title: Plutus Script In-Memory Caching
```
rphair:
```suggestion
Title: Plutus Script Caching  
```
It seems like the "In-Memory" term would be understood from the performance concept but we can keep working on this from both ends... for now setting it to the PR title which more people have agreed with at this point.

CIP-Plutus-Script-Cache/README.md
```diff
@@ -0,0 +1,112 @@
+---
+CIP: ???  
+Title: Plutus Script In-Memory Caching  
+Category: Plutus
```
rphair:
```suggestion
Category: Node  
```
Defined by #1145 and more discussion to come.

CIP-Plutus-Script-Cache/README.md
```diff
@@ -0,0 +1,112 @@
...
+CIP: ???  
+Title: Plutus Script In-Memory Caching  
+Category: Plutus  
+Status: Proposed  
+Authors:  
+  - Philip DiSarro <philip.disarro@gmail.com>  
+Implementors: []  
+Discussions:  
+  - https://github.com/IntersectMBO/cardano-ledger/blob/master/docs/adr/2024-08-14_009-refscripts-fee-change.md
+  - https://github.com/cardano-foundation/CIPs/pull/1031
```
rphair:
```suggestion
Discussions:  
  - https://github.com/IntersectMBO/cardano-ledger/blob/master/docs/adr/2024-08-14_009-refscripts-fee-change.md
  - Original PR: https://github.com/cardano-foundation/CIPs/pull/1031
```
@colll78 we have a new convention per https://github.com/cardano-foundation/CIPs/pull/1125 to say what these links are (I think it would be better for you to label the first link; the second label will be common across CIPs).

CIP-Plutus-Script-Cache/README.md
```diff
@@ -0,0 +1,112 @@
...
+Created: 2025-05-01  
+License: CC-BY-4.0  
+---
+
+## Abstract
+
+We propose introducing a memory-bound, Least Recently Used (LRU) in-memory caching layer within Cardano nodes to store deserialized Plutus scripts. This optimization will significantly reduce redundant deserialization operations and improve node performance, especially in high-throughput environments or dApps with widely reused scripts. Scripts not found in cache are lazily reloaded from transaction data or reference inputs via cold reload.
+
+## Motivation: why is this CIP necessary?
```
rphair:
```suggestion
## Motivation: Why is this CIP necessary?
```
(also standardising on grammatically correct capitalisation in https://github.com/cardano-foundation/CIPs/pull/1125)

CIP-Plutus-Script-Cache/README.md
```diff
@@ -0,0 +1,112 @@
...
+
+Add new parameters to the node configuration:
+
+```yaml
+PlutusScriptCache:
+  maxCacheSizeMB: 256
+  evictionPolicy: LRU
+```
+
+## Rationale: how does this CIP achieve its goals?
```
rphair:
```suggestion
## Rationale: How does this CIP achieve its goals?
```
(also standardising on grammatically correct capitalisation in https://github.com/cardano-foundation/CIPs/pull/1125)

rphair (commented):
At the CIP meeting today we finally laid out the tentative definition for a `Node` category — including discussion of all the points raised above & previously documented in https://github.com/cardano-foundation/CIPs/issues/1076#issuecomment-3825818481 — which can now be continued in #1145 pending a CIP that defines a `Node` category more properly.

Any reviewers of this PR (cc @ch1bo) with concerns or support for that `Node` category please give us some feedback on that issue.  The reflexive response that "there's nobody to handle 'Node' CIPs" came up in the CIP meeting today and was effectively answered by the decentralisation argument there.  Also I think we are just following the well known precedent that any open source project's standards will eventually outgrow its issue queue.

@colll78 please rename your containing directory to `CIP-0174` and re-link your "rendered" proposal link accordingly 🎉

CIP-0174/README.md
```diff
@@ -0,0 +1,112 @@
...
+
+```yaml
+PlutusScriptCache:
+  maxCacheSizeMB: 256
+  evictionPolicy: LRU
+```
+
+## Rationale: how does this CIP achieve its goals?
+
+This proposal follows the same spirit as performance optimizations in other smart contract chains (e.g. Solana's JIT caching model). It does not modify on-chain structures or Plutus semantics — only the **execution engine of node software**.
```
colll78: It is a pre-requisite to remove the reference script deserialization fee.

rphair: @ch1bo the original conception that CIPs needed to propose a change to _on-chain_ properties to be considered an improvement to "Cardano" (i.e. defining Cardano as its chain for CIP purposes) was thrown out by popular consensus around 2021.  We have a notable number of CIPs today that don't affect anything on the chain.

rphair (dismissed):
In order to start feeling out what happens to `Node` category CIPs, we might actually have to merge one of them & this seems like a good test case: unless some problem with the technical feasibility is pointed out along the way.

@Ryun1 @perturbing (and @colll78 as first advocate for a CIP in this category) I think we should next:
1. assemble a list of node projects & reviewers to tag for each of these;
2. test that tentative process by tagging them here;
3. accumulate that list in https://github.com/cardano-foundation/CIPs/issues/1145 from where we would start working on a process & category definition CIP for all `Node` CIPs & CPSs.

In any case I'm approving this on the merit of the proposal itself and based on my impression that the `Path to Active` makes sense without having to micromanage the response of the non-Haskell node teams in advance.  However I do think we should wait to apply `Last Check` until making some progress in getting that cross-section of reviewers across the multi-node ecosystem (including any statements that this proposal might only apply to the Haskell node).

wolf31o2:
This is describing an implementation detail of the Haskell cardano-node software and not something that really makes sense as a CIP. In Dingo, we already have a different multi-layered LRU for several objects (blocks, transactions, and UTxOs). I would not redesign my cache to adhere to this specific "Cardano" improvement proposal because it specifies a less featureful cache system then the one we've already built. We're not interested in nerfing our designs to adhere to limitations introduced by the internal designs of a specific node implementation.

Timing requirements, performance characteristics, etc. can be set, but prescribing how it's achieved? That's a bridge too far, IMO.

rphair:
@colll78 this effort may continue to get reactions like the above unless we can re-cast potential `Node` CIPs in a strict & widely agreeable context.  At the earlier CIP meeting when the editors' consensus was to try to accommodate this proposal, we did so because of the possibility you argued that Haskell Node performance issues would offer similar challenges for other nodes & therefore the community was best served by node teams considering these performance issues in parallel.

If you still think that can be done, we would have to rely on you to justify what impact this proposal could or should have on other nodes: and also in general while we're considering honouring these proposals in a new category (see https://github.com/cardano-foundation/CIPs/issues/1145#issuecomment-4001192652).

rphair:
@colll78 - as @wolf31o2 has said, and for reasons I've attempted to summarise here:
* https://github.com/cardano-foundation/CIPs/issues/1145#issuecomment-4381778010

... I also don't see how this can continue as a CIP.  To do so without a carefully scoped "extra" category would break a cardinal rule of the CIP process: that standards for a class of software should be applicable across that entire class & otherwise they are not "standards" but rather "design documents" that should be homed as repository issues, web pages, or research papers instead.

@colll78 I do realise this difficulty is only repeating your already given reasons for posting this as a CIP.  We likely cannot accommodate this as a CIP for the sake of other developers, because of the obligation it would imply if added to the proposed `Node` category... unless we created a hypothetical CIP category for such "design documents" whose main purpose would be to document principles only applicable to a narrow scope and therefore not to "Cardano" (arguably contrary to the CIP process).

Editorially therefore I have to ask you again to think about where else this document can be maintained & promoted from outside the CIP repository.  We can leave this `Waiting for Author` — or you can close it — until you and/or other developers update the issue above with a plan for a new category, for this & similar proposals, that preserves the scope & legitimacy of the CIP process itself... then we might categorise this proposal as such.

rphair:
@colll78 with an additional 3 months of no dispute to the summary that nobody can compel attention to design recommendations presented as CIPs like this — and without editors even having a way to categorise this to produce _any_ implementation pressure — we do have to close this after making a best effort to accommodate your request.

If there is any offer of support for this & similarly presented ideas across the Cardano multi-node world, I trust you and/or other representatives will comment here: and then somehow we might re-open this thread accordingly.  cc @Ryun1 @perturbing

colll78:
Can Cardano-Node developers just implement this? Use a random cache, it is an extremely small amount of work and would vastly improve node performance and network security.

PR #1033, opened 2025-05-08
SRV Registry

coot:
# Summary

This CIP proposes SRV prefix registry to foster discoverability of DAPs which
have their own decentralised network running along-side cardano block-chain.

### DIFF
--- a/CIP-xxxx/README.md
+++ b/CIP-xxxx/README.md
@@ -1,0 +1,111 @@
+---
+CPS: ????
+Title: SRV registry
+Status: Active
+Category: Network
+Authors:
+  - Marcin Szamotulski <marcin.szamotulski@iohk.io>
+Implementors: N/A
+Discussions: 
+  - [PR](https://github.com/cardano-foundation/CIPs/pull/xxx)
+Created: 2025-04-22
+License: CC-BY-4.0
+---
+
+## Abstract
+
+This CIP defines the procedure for assigning SRV prefixes for Cardano applications (like `cardano-node`, `mithril`, etc.).
+
+## Motivation
+
+The **Cardano Ledger** allows the use of SRV records in the **SPO** registration certificate.
+
+In this CIP, we propose to make them usable by decentralised applications running on `Cardano`.
+
+Having access to the ledger (either directly or through tools like `cardano-cli`) will allow decentralised applications (DAPS) developers to construct networks based on registered relays. 
+
+All involved parties (SPOs, DAP developers, etc) need to agree on SRV prefixes used by various DAPs and thus a public registry governed by a public CIP process is required to foster DAPs that co-exist with the **Cardano** network, like **Mithril** (see [CIP#137]).
+
+## Problem
+
+Bootstraping networks requires sharing information about available services.
+
+Historically, this was done by sharing IP or DNS names and PORT numbers, but this approach has its limitations.
+
+SRV records were designed to make deployment independent of hard-coded service end-points (see [RVC#2782]).
+
+Thus they can be instrumental in `Cardano`, since relays are registered on the blockchain through the registration certificate.
+
+By using SRV records in the registration certificate (which is supported by the `cardano-ledger`, but not by `cardano-node`), we wish to solve this problem not just for `Cardano` node implementation, but also for any DAP that requires constructing its own network.
+
+SRV provide a mechanism for exposing DAPs co-deployed with a node, like **mithril** or **hydra**.
+
+Making such services discoverable is one of the key features addressed by this CIP.
+
+In this CIP, we propose prefixes for both **Cardano** and **Mithril**; in the future, other services can be registered through a CIP process, thus starting a registry of prefixes used by the **Cardano** ecosystem.
+
+
+## Specification
+
+### SRV Prefix Registry
+
+| **service** | SRV prefix | Status |
+| _cardano_ | `_cardano._tcp` | Active |
+| _mithril_ | `_mithril._tcp` | Active |
+
+### SRV Prefix Semantics
+
+When a **cardano** node implementation reads an SRV record from a ledger, it must add the _cardano_ prefix from the table above before making a DNS lookup, e.g. it should do a DNS query for `_cardano._tcp.<srv-record-from-ledger>`.
+
+This design allows DAPs to use SRV records registered in the ledger for different purposes, e.g. a **mithril** node can use them to learn about end-points of its network.
+
+#### SRV Registry Rules
+
+* Each DAP can have at most one entry in the registry.
+* A CIP process assigns new entries, after a careful consideration and consultation with all the involved parties (see #acceptance-criteria below).
+* Entries cannot be removed, but can be revoked by assigning a `Revoked` status.
+  This can only happen if a DAP is no longer supported.
+
+### Example
+
+When registering a cardano pool on `example.com` domain using an `SRV` record, one should use:
+```
+cardano-cli latest stake-pool ... --multi-host-pool-relay example.com
+```
+(see [register-stake-pool]); and configure SRV record at `_cardano._tcp.example.com` to point to **Cardano** relays, `_mithril._tcp.example.com` to point to **Mithril** relays (see [srv], currently under development).
+
+A **Cardano** node implementation when retrieving information from a registration certificate from the ledger will receive `example.com` and it must prepend the `_cardano._tcp` prefix to make an SRV query.
+
+It may as well append `.` to indicate it is a top-level query, resulting in querying the `_cardano._tcp.example.com.` domain.
+
+## Path to Active
+
+### Acceptance Criteria
+
+The CIP should be accepted by:
+
+* [ ] [**Technical Steering Committee**][tsc]
+
+And when there's no major objection from one of the currently involved parties:
+
+* [ ] [**Amaru Team**][amaru] accepts the proposal
+* [ ] [**Cardano-Node Team**][cardano-node] accepts the proposal
+* [ ] [**Gouroboros Team**][gouroboros] accepts the proposal
+* [ ] [**Hydra Team**][hydra] accepts the proposal
+* [ ] [**Mithril Team**][mithril] accepts the proposal
+
+### Implementation Plan
+
+Each **Cardano** node implementation or other tools which rely on SRV records stored in the ledger should comply with this proposal.
+
+[CIP#137]: https://github.com/cardano-foundation/CIPs/tree/master/CIP-0137
+[register-stake-pool]: https://developers.cardano.org/docs/operate-a-stake-pool/register-stake-pool
+[RFC#2782]: https://datatracker.ietf.org/doc/html/rfc2782 
+[srv]: https://www.cloudflare.com/en-gb/learning/dns/dns-records/dns-srv-record/
+
+[amaru]: https://github.com/pragma-org/amaru
+[cardano-node]: https://github.com/IntersectMBO/cardano-node
+[mithril]: https://github.com/input-output-hk/mithril
+[gouroboros]: https://github.com/blinklabs-io/gouroboros
+[tsc]: https://docs.intersectmbo.org/intersect-overview/intersect-committees/technical-steering-committee-tsc
+[hydra]: https://github.com/cardano-scaling/hydra


CIP-xxxx/README.md
```diff
@@ -0,0 +1,111 @@
...
+In this CIP, we propose prefixes for both **Cardano** and **Mithril**; in the future, other services can be registered through a CIP process, thus starting a registry of prefixes used by the **Cardano** ecosystem.
+
+
+## Specification
+
+### SRV Prefix Registry
+
+| **service** | SRV prefix | Status |
+| _cardano_ | `_cardano._tcp` | Active |
+| _mithril_ | `_mithril._tcp` | Active |
```
ch1bo, line 54:
```suggestion
| **service** | SRV prefix      | Status |
|:------------|:----------------|:-------|
| _cardano_   | `_cardano._tcp` | Active |
| _mithril_   | `_mithril._tcp` | Active |
```

CIP-xxxx/README.md
```diff
@@ -0,0 +1,111 @@
...
+
+All involved parties (SPOs, DAP developers, etc) need to agree on SRV prefixes used by various DAPs and thus a public registry governed by a public CIP process is required to foster DAPs that co-exist with the **Cardano** network, like **Mithril** (see [CIP#137]).
+
+## Problem
+
+Bootstraping networks requires sharing information about available services.
+
+Historically, this was done by sharing IP or DNS names and PORT numbers, but this approach has its limitations.
+
+SRV records were designed to make deployment independent of hard-coded service end-points (see [RVC#2782]).
```
ch1bo, line 35: Should `RVC#2782` link to somewhere?

CIP-xxxx/README.md
```diff
@@ -0,0 +1,111 @@
...
+
+## Motivation
+
+The **Cardano Ledger** allows the use of SRV records in the **SPO** registration certificate.
+
+In this CIP, we propose to make them usable by decentralised applications running on `Cardano`.
+
+Having access to the ledger (either directly or through tools like `cardano-cli`) will allow decentralised applications (DAPS) developers to construct networks based on registered relays. 
+
+All involved parties (SPOs, DAP developers, etc) need to agree on SRV prefixes used by various DAPs and thus a public registry governed by a public CIP process is required to foster DAPs that co-exist with the **Cardano** network, like **Mithril** (see [CIP#137]).
```
ch1bo, line 27:
```suggestion
All involved parties (SPOs, DAP developers, etc) need to agree on SRV prefixes used by various DAPs and thus a public registry governed by a public CIP process is required to foster DAPs that co-exist with the **Cardano** network, like **Mithril** (see [CIP#137](../CIP-0137).
```

CIP-xxxx/README.md
```diff
@@ -0,0 +1,111 @@
...
+License: CC-BY-4.0
+---
+
+## Abstract
+
+This CIP defines the procedure for assigning SRV prefixes for Cardano applications (like `cardano-node`, `mithril`, etc.).
+
+## Motivation
+
+The **Cardano Ledger** allows the use of SRV records in the **SPO** registration certificate.
```
ch1bo, line 21: Can you link to some resource explaining this or if not available include an example how this is typically done?

ch1bo (commented):
I like the idea and Hydra could support this. I think it's most useful to open protocols though where we would want to enumerate all registered services using such certificates (info on how to register and query that registry through the ledger would be interesting). Currently, Hydra is using statically configured topologies only and this does not make much of a difference when sharing connection info (sharing a domain name vs. sharing a hostname/ip). Definitely useful for full on overlay networks where we also want to register services for longer (mithril?)

CIP-xxxx/README.md
crocodile-dentist: the rfc is https://datatracker.ietf.org/doc/html/rfc2782

CIP-xxxx/README.md
coot: `cardano-cli` supports it via the `--multi-host-pool-relay` flag, see: https://developers.cardano.org/docs/operate-a-stake-pool/register-stake-pool/

CIP-xxxx/README.md
coot: The link is included at the bottom, I'll update it to a relative one, though.

CIP-xxxx/README.md
```diff
@@ -0,0 +1,111 @@
+---
+CPS: ????
```
rphair, line 2:
```suggestion
CIP: ?
```

rphair (commented):
@coot I've titled this PR as a CIP because everything about it is following the CIP form... but if the `Draft` period suggests it is really a CPS then please feel free to change it.

In any case after you take it out of `Draft` mode we'll tag it `Triage` after a format double-check (there are currently some things missing) which will set it up for public discussion at the next CIP meeting.

coot:
It is a CIP, and I just wanted to gather initial feedback first - hence a draft.

CIP-xxxx/README.md
```diff
@@ -0,0 +1,117 @@
+---
+CIP: ?
+Title: SRV registry
+Status: Active
+Category: Network
+Authors:
+  - Marcin Szamotulski <marcin.szamotulski@iohk.io>
+Implementors: N/A
+Discussions: 
+  - [PR](https://github.com/cardano-foundation/CIPs/pull/xxx)
```
Ryun1, line 10:
```suggestion
  - [PR](https://github.com/cardano-foundation/CIPs/pull/1033)
```

CIP-xxxx/README.md
```diff
@@ -0,0 +1,117 @@
...
+Created: 2025-04-22
+License: CC-BY-4.0
+---
+
+## Abstract
+
+This CIP defines the procedure for assigning SRV prefixes for Cardano applications (like `cardano-node`, `mithril`, etc.).
```
Ryun1, line 17:
```suggestion
This CIP defines the procedure for assigning Service Record (SRV) prefixes for Cardano applications (like `cardano-node`, `mithril`, etc.).
```

Ryun1: might nice to have the definition in there

CIP-xxxx/README.md
```diff
@@ -0,0 +1,117 @@
+---
+CIP: ?
+Title: SRV registry
+Status: Active
```
Ryun1, line 4:
```suggestion
Status: Proposed
```

CIP-xxxx/README.md
```diff
@@ -0,0 +1,117 @@
...
+  - [PR](https://github.com/cardano-foundation/CIPs/pull/xxx)
+Created: 2025-04-22
+License: CC-BY-4.0
+---
+
+## Abstract
+
+This CIP defines the procedure for assigning SRV prefixes for Cardano applications (like `cardano-node`, `mithril`, etc.).
+
+## Motivation
```
Ryun1, line 19:
```suggestion
## Motivation: why is this CIP necessary?
```

CIP-xxxx/README.md
```diff
@@ -0,0 +1,117 @@
...
+
+The **Cardano Ledger** allows the use of SRV records in the **SPO** registration certificate (see [register-stake-pool]).
+
+In this CIP, we propose to make them usable by decentralised applications running on `Cardano`.
+
+Having access to the ledger (either directly or through tools like `cardano-cli`) will allow decentralised applications (DAPS) developers to construct networks based on registered relays. 
+
+All involved parties (SPOs, DAP developers, etc) need to agree on SRV prefixes used by various DAPs and thus a public registry governed by a public CIP process is required to foster DAPs that co-exist with the **Cardano** network, like **Mithril** (see [CIP#137]).
+
+## Problem
```
Ryun1, line 29:
```suggestion
```

please follow the template :) 
https://github.com/cardano-foundation/CIPs/blob/master/.github/CIP-TEMPLATE.md?plain=1

rphair (commented):
@coot from our reading at the meeting under the clock: aside from @Ryun1's format suggestions above which are universal for CIPs, we had to use our imagination to determine what this CIP was about... including the connection with SRV records that a few of us know from DNS experience in a different context.

So although you might have been hoping for technical feedback about the principle itself, the main feedback was that more background could be given about the terms used, and who would be using these methods, and for what.
- For instance: some SPOs who've used SRV records to configure their stake pool relays might look over this CIP without knowing whether it's applicable to them unless something is put in the Abstract to define the target audience and/or applications of the idea.
- If that's clear when this comes out of `Draft` mode then it would help us tag other potential reviewers for more detailed technical feedback.

I'll take the `Triage` label off now & we'll reapply it as soon as out of draft & ready for review in your opinion.  Personally I'd like to have a better technical understanding of your proposal but would prefer to try again when there's some more background info here.

jpraynaud:
@coot I believe that we should make a clear distinction between the DMQ node and Mithril nodes:

- The DMQ node could be utilized by various protocols, with SRV records used to advertise the DMQ host and relays for each protocol.
- Mithril nodes (signer and aggregator) could also use SRV records, but for different purposes, such as discovering available aggregators.

It may be necessary to define multiple entries per service in the SRV Prefix Registry:

| **service** | **scope** | SRV prefix      | Status |
|:------------|:------------|:----------------|:-------|
| _cardano_   |    | `_cardano._tcp` | Active |
| _mithril_   | _dmq_   | `_mithril._dmq._tcp` | Active |
| _mithril_   | _aggregator_   | `_mithril._aggregator._tcp` | Active |

Additionally, regarding DMQ: would this imply that the DMQ node also needs access to the ledger state (which is not currently planned)?

coot:
@rphair, thanks for your feedback. I'll update the CIP and clarify the target audience and how it can be useful for SPOs.

coot:
@jpraynaud As I understood `dmq` is universal, it might be used by different DAPs, so we shouldn't register it in the registry, instead each particular DAP should have it's own entry.  But maybe that's my aberration :wink:.

This registry is only helpful for DAPs that want to utilise the ledger peers.

CIP-xxxx/README.md
coot: This is meant to be a subsection of the `Motivation` section.

coot:
@jpraynaud what about:

| **service** | **scope**     | **SRV prefix**                       | **Status** |
|:------------|:--------------|:-------------------------------------|:-----------|
| _cardano_   |               | `_cardano._tcp`                      | Active     |
| _mithril_   | _dmq_         | `_dmq._mithril._cardano._tcp`        | Active     |
| _mithril_   | _aggregator_  | `_aggregator._mithril._cardano._tcp` | Active     |

CIP-xxxx/README.md
rphair: that's fine @coot - it'll work fine then if just bumped down a level:
```suggestion
### Problem
```

rphair (commented):
@coot as both an editor and an SPO I think what you have added in [this push](https://github.com/cardano-foundation/CIPs/compare/1ea92ff4d224bf3eb6671d758cf55e896e50eee6..c7c7a0ddce7dcb1dd76666dbf9fd0f1e3f3f37fe) clarifies the purpose of the CIP well enough.

I've marked it `Triage` to put it on the agenda for the next meeting (https://hackmd.io/@cip-editors/112) which you would be welcome to attend if it suits you.

I think this is a sound document & proposal so plan to suggest at the meeting that it be given a candidate CIP number.

CIP-xxxx/README.md
```diff
@@ -0,0 +1,127 @@
...
+  - [PR](https://github.com/cardano-foundation/CIPs/pull/1033)
+Created: 2025-04-22
+License: CC-BY-4.0
+---
+
+## Abstract
+
+This CIP defines the procedure for assigning Service Record (SRV) prefixes for Cardano applications (like `cardano-node`, `mithril`, etc.).
+
+This CIP constructs means of sharing authoritative information between SPOs who deploy services and DAP developers who write them.
```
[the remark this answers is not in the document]
coot: SPOs need to know how to deploy, DAP devs decide how this needs to be done - the CIP adding entries to the _SRV registry_ allows one group to agree & communicate with the other.

CIP-xxxx/README.md
```diff
@@ -0,0 +1,127 @@
...
+
+## Motivation: why is this CIP necessary?
+
+The **Cardano Ledger** allows the use of SRV records in the **SPO** registration certificate (see [register-stake-pool]).
+
+In this CIP, we propose to make them usable by decentralised applications running on `Cardano`.
+
+Having access to the ledger (either directly or through tools like `cardano-cli`) will allow decentralised applications (DAPS) developers to construct networks based on registered relays.
```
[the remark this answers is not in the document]
Ryun1: ahh I see later on these DAPs are different from dApps
in that case, can I recommend renaming DAPs to something else as this is confusing

CIP-xxxx/README.md
```diff
@@ -0,0 +1,127 @@
...
+
+The **Cardano Ledger** allows the use of SRV records in the **SPO** registration certificate (see [register-stake-pool]).
+
+In this CIP, we propose to make them usable by decentralised applications running on `Cardano`.
+
+Having access to the ledger (either directly or through tools like `cardano-cli`) will allow decentralised applications (DAPS) developers to construct networks based on registered relays. 
+
+All involved parties (SPOs, DAP developers, etc) need to agree on SRV prefixes used by various DAPs and thus a public registry governed by a public CIP process is required to foster DAPs that co-exist with the **Cardano** network, like **Mithril** (see [CIP#137]).
+
+Note that the notion of a DAP is specific in this context: these are applications, which construct their own network and relay on ledger peers published on the block-chain.   This include node implemtations, or 2nd-layer solutions like Mithril or Hydra.
```
Ryun1, line 31:
```suggestion
Note that the notion of a DAP is specific in this context: these are applications, which construct their own network and relay on ledger peers published on the block-chain. This include node implementations, or 2nd-layer solutions like Mithril or Hydra.
```

fixed spelling and removed extra space

CIP-xxxx/README.md
```diff
@@ -0,0 +1,127 @@
...
+
+Historically, this was done by sharing IP or DNS names and PORT numbers, but this approach has its limitations.
+
+SRV records were designed to make deployment independent of hard-coded service end-points (see [RFC#2782]).
+
+Thus they can be instrumental in `Cardano`, since relays are registered on the blockchain through the registration certificate.
+
+By using SRV records in the registration certificate (which is supported by the `cardano-ledger`, but not by `cardano-node`), we wish to solve this problem not just for `Cardano` node implementation, but also for any DAP that requires constructing its own network.
+
+SRV provide a mechanism for exposing DAPs co-deployed with a node, like **mithril** or **hydra**.
```
Ryun1, line 45:
```suggestion
SRV provides a mechanism for exposing DAPs co-deployed with a node, like **mithril** or **hydra**.
```

CIP-xxxx/README.md
```diff
@@ -0,0 +1,127 @@
...
+
+## Abstract
+
+This CIP defines the procedure for assigning Service Record (SRV) prefixes for Cardano applications (like `cardano-node`, `mithril`, etc.).
+
+This CIP constructs means of sharing authoritative information between SPOs who deploy services and DAP developers who write them.
+
+## Motivation: why is this CIP necessary?
+
+The **Cardano Ledger** allows the use of SRV records in the **SPO** registration certificate (see [register-stake-pool]).
```
Ryun1, line 23:
```suggestion
The **Cardano Ledger** allows the use of SRV in the **SPO** registration certificate (see [register-stake-pool]).
```
Since SRV includes abbreviation for 'record' do we need to say it again?
perhaps this is the convention, that I am unfamiliar with

CIP-xxxx/README.md
```diff
@@ -0,0 +1,127 @@
...
+|:------------|:--------------| ------------------------------------:|:-----------|
+| _cardano_   |               |                      `_cardano._tcp` | Active     |
+| _mithril_   | _dmq_         |       `_.dmq._mithril._cardano._tcp` | Active     |
+| _mithril_   | _aggregator_  | `_aggregator._mithril._cardano._tcp` | Active     |
+
+### SRV Prefix Semantics
+
+When a **cardano** node implementation reads an SRV record from a ledger, it must add the _cardano_ prefix from the table above before making a DNS lookup, e.g. it should do a DNS query for `_cardano._tcp.<srv-record-from-ledger>`.
+
+This design allows DAPs to use SRV records registered in the ledger for different purposes, e.g. a **mithril** node can use them to learn about end-points of its network.
```
Ryun1, line 69: Does this mean that someone could use a stake pool registration certificate to ONLY register a mithril node? and not a Cardano node

CIP-xxxx/README.md
```diff
@@ -0,0 +1,127 @@
...
+
+* Each DAP can have at most one entry in the registry.
+* A CIP process assigns new entries, after a careful consideration and consultation with all the involved parties (see #acceptance-criteria below).
+* Entries cannot be removed, but can be revoked by assigning a `Revoked` status.
+  This can only happen if a DAP is no longer supported.
+
+### Example
+
+When registering a cardano pool on `example.com` domain using an `SRV` record, one should use:
+```
```
Ryun1, line 83:
```suggestion
```shell
```

CIP-xxxx/README.md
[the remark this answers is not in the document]
perturbing: I agree, a DApp is built on top of a blockchain. I think this context is talking about peer to peer protocols or decentralized protocols (like hydra/mithril).

CIP-xxxx/README.md
```diff
@@ -0,0 +1,127 @@
+---
+CIP: ?
```
rphair, line 2:
```suggestion
CIP: 155
```

rphair (commented):
@coot as expected this was welcomed as a CIP candidate at today's CIP meeting.  Please update the CIP number in the text and rename the containing directory from `CIP-xxxx` to `CIP-0155` 🎉

We also are interested to see if the assignment of a CIP number can lead to involvement from the Hydra and SPO populations: somehow referring them here to give more feedback about how it would/will work in their respective operations.

Also we would like to please work with you more conventionally on this branch — rather than getting a "force push" every time you make changes — so that commits can be associated with ongoing review as we go forward.  This is the approach that works best for collaborative editing & a universal recommendation for PRs in this repo. 🙏

CIP-xxxx/README.md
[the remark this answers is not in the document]
coot: Agreed, I'll use _decentralised protocol_ instead of _dapp_.

CIP-xxxx/README.md
coot: It's often used this way.  Some examples: [_What is a DNS SRV record?_](https://www.cloudflare.com/learning/dns/dns-records/dns-srv-record/), or the original [RFC](https://www.rfc-editor.org/rfc/rfc2782)

CIP-xxxx/README.md
```diff
@@ -0,0 +1,127 @@
...
+When registering a cardano pool on `example.com` domain using an `SRV` record, one should use:
+```
+cardano-cli latest stake-pool ... --multi-host-pool-relay example.com
+```
+(see [register-stake-pool]); and configure SRV record at `_cardano._tcp.example.com` to point to **Cardano** relays, `_mithril._tcp.example.com` to point to **Mithril** relays (see [srv], currently under development).
+
+A **Cardano** node implementation when retrieving information from a registration certificate from the ledger will receive `example.com` and it must prepend the `_cardano._tcp` prefix to make an SRV query.
+
+It may as well append `.` to indicate it is a top-level query, resulting in querying the `_cardano._tcp.example.com.` domain.
+
```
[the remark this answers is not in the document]
coot: I don't know what should be put there other than something trivial, such as the fact that the SRV registry is being honoured by developers and SPOs.

CIP-xxxx/README.md
coot: It's a writing style to use double spaces between sentences.  It might be in other places...

CIP-xxxx/README.md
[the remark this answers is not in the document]
rphair: @coot I think you're right: we normally talk about comparative approaches & alternatives here, but I can't imagine anything else that would be fit for purpose.  Therefore a simple statement to that effect would be enough to fill in this required space. 😎

CIP-xxxx/README.md
rphair: @coot it's in other places & we have agreed (after last confirmed with @perturbing) that both double & single spaces after a full stop are used routinely in CIPs and acceptable either (or both) ways.  I think @Ryun1 had tagged the _triple_ space and then suggested removing one too many of the extra spaces 😅

rphair:
@coot it seems like this proposal is progressing favourably in parallel, so please post here after the proposal text has stabilised enough to commit to the repository as `Proposed` (or `Active` if there is justification for it) so we can review again & mark `Last Check` before merge.

CIP-0155/README.md
```diff
@@ -0,0 +1,133 @@
...
+
+When a **cardano** node implementation reads an SRV record from a ledger, it must add the _cardano_ prefix from the table above before making a DNS lookup, e.g. it should do a DNS query for `_cardano._tcp.<srv-record-from-ledger>`.
+
+This design allows DPs to use SRV records registered in the ledger for different purposes, e.g. a **mithril** node can use them to learn about end-points of its network.
+
+Each prefix SHOULD start with `_cardano._tcp` or `_cardano._udp`, to avoid clashes with services not related to `Cardano`.
+
+#### SRV Registry Rules
+
+* Each DP can have at most one entry in the registry.
```
Quantumplation: This this still relevant? In the example above, Mithril has two entries. Perhaps a bit of detail on what the definitions / intended use of `service` vs `scope` are would be helpful.

CIP-0155/README.md
```diff
@@ -0,0 +1,133 @@
...
+| _mithril_   | _dmq_         |       `_.dmq._mithril._cardano._tcp` | Active     |
+| _mithril_   | _aggregator_  | `_aggregator._mithril._cardano._tcp` | Active     |
+
+### SRV Prefix Semantics
+
+When a **cardano** node implementation reads an SRV record from a ledger, it must add the _cardano_ prefix from the table above before making a DNS lookup, e.g. it should do a DNS query for `_cardano._tcp.<srv-record-from-ledger>`.
+
+This design allows DPs to use SRV records registered in the ledger for different purposes, e.g. a **mithril** node can use them to learn about end-points of its network.
+
+Each prefix SHOULD start with `_cardano._tcp` or `_cardano._udp`, to avoid clashes with services not related to `Cardano`.
```
Quantumplation, line 66: I get the semantics of "start with" from context, but it's a bit unintuitive, as `_cardano._tcp` comes at the end. Would this be well understood well by someone more familiar with DNS than I am?

What should services that plan to be multi-chain use as a prefix? for example, if a decentralized bridge or multi-chain layer zero type protocol with Cardano support; Should they have their own prefix, and avoid this CIP process? or should they have different prefixes for different chains? etc.

CIP-0155/README.md
```diff
@@ -0,0 +1,133 @@
...
+
+### Example
+
+When registering a cardano pool on `example.com` domain using an `SRV` record, one should use:
+```shell
+cardano-cli latest stake-pool ... --multi-host-pool-relay example.com
+```
+(see [register-stake-pool]); and configure SRV record at `_cardano._tcp.example.com` to point to **Cardano** relays, `_mithril._tcp.example.com` to point to **Mithril** relays (see [srv], currently under development).
+
+A **Cardano** node implementation when retrieving information from a registration certificate from the ledger will receive `example.com` and it must prepend the `_cardano._tcp` prefix to make an SRV query.
```
Quantumplation: can you elaborate a bit this example, to show what the SRV query might return, and what the node might then do with it?

That's a little bit outside of scope for this CIP, as it's just repeating what DNS would do, but I think it would be helpful (and inspirational) to see the example carried through to its conclusion.

CIP-0155/README.md
```diff
@@ -0,0 +1,133 @@
...
+
+We consider DPs as applications, which construct their own network and relay on ledger peers published on the block-chain. This include node implementations, or 2nd-layer solutions like Mithril or Hydra.
+
+### Problem
+
+Bootstraping networks requires sharing information about available services.
+
+Historically, this was done by sharing IP or DNS names and PORT numbers, but this approach has its limitations.
+
+SRV records were designed to make deployment independent of hard-coded service end-points (see [RFC#2782]).
```
Quantumplation: A tiny bit more context here was needed for me to understand the value; in particular, the missing piece for me was that `A` records don't include a port, protocol, priority, or weight.

CIP-0155/README.md
```diff
@@ -0,0 +1,133 @@
...
+
+When registering a cardano pool on `example.com` domain using an `SRV` record, one should use:
+```shell
+cardano-cli latest stake-pool ... --multi-host-pool-relay example.com
+```
+(see [register-stake-pool]); and configure SRV record at `_cardano._tcp.example.com` to point to **Cardano** relays, `_mithril._tcp.example.com` to point to **Mithril** relays (see [srv], currently under development).
+
+A **Cardano** node implementation when retrieving information from a registration certificate from the ledger will receive `example.com` and it must prepend the `_cardano._tcp` prefix to make an SRV query.
+
+It may as well append `.` to indicate it is a top-level query, resulting in querying the `_cardano._tcp.example.com.` domain.
```
Quantumplation: Can you elaborate on why someone "may as well" do this?

CIP-0155/README.md
```diff
@@ -0,0 +1,133 @@
...
+
+* [ ] [**Amaru Team**][amaru] accepts the proposal
+* [ ] [**Cardano-Node Team**][cardano-node] accepts the proposal
+* [ ] [**Gouroboros Team**][gouroboros] accepts the proposal
+* [ ] [**Hydra Team**][hydra] accepts the proposal
+* [ ] [**Mithril Team**][mithril] accepts the proposal
+
+### Implementation Plan
+
+Each **Cardano** node implementation or other tools which rely on SRV records stored in the ledger should comply with this proposal.
```
Quantumplation: A little more detail here on "comply" would be helpful. What does it mean for Amaru, for example, to "comply" with this proposal?

CIP-0155/README.md
```diff
@@ -0,0 +1,133 @@
...
+
+By using SRV records in the registration certificate (which is supported by the `cardano-ledger`, but not by `cardano-node`), we wish to solve this problem not just for `Cardano` node implementation, but also for any DP that requires constructing its own network.
+
+SRV provides a mechanism for exposing DPs co-deployed with a node, like **mithril** or **hydra**.
+
+Making such services discoverable is one of the key features addressed by this CIP.
+
+In this CIP, we propose prefixes for both **Cardano** and **Mithril**; in the future, other services can be registered through a CIP process, thus starting a registry of prefixes used by the **Cardano** ecosystem.
+
+SPOs who deploy services need to configure their system according to the registry, e.g. SPO's cardano relays node MUST be available at `_cardano._tcp.<SPO_DOMAIN_NAME>`, as other nodes on the system will be looking at this address.
```
Quantumplation, line 50: It sounds like this sentence is describing the *current* state, correct? i.e. already the SRV records used by SPOs are expected to / required to begin with `_cardano._tcp`; and this proposal just changes that from a singular value to a "pre"fix, so it can be extended for other services?

Or is this intended to be a breaking change, where there was no enforcement on the SRV record format so far, and this invalidates those, requiring them to be more consistent?

CIP-0155/README.md
```diff
@@ -0,0 +1,133 @@
...
+
+Bootstraping networks requires sharing information about available services.
+
+Historically, this was done by sharing IP or DNS names and PORT numbers, but this approach has its limitations.
+
+SRV records were designed to make deployment independent of hard-coded service end-points (see [RFC#2782]).
+
+Thus they can be instrumental in `Cardano`, since relays are registered on the blockchain through the registration certificate.
+
+By using SRV records in the registration certificate (which is supported by the `cardano-ledger`, but not by `cardano-node`), we wish to solve this problem not just for `Cardano` node implementation, but also for any DP that requires constructing its own network.
```
Quantumplation: What does it mean that this is supported by the `cardano-ledger` but not the `cardano-node`? As in, there's a field in the SPO registration certificate for it, but the node doesn't actually use that to make SRV requests for peer discovery?

Quantumplation (commented):
Couple of things that were unclear to me on my first read

CIP-0155/README.md
```diff
@@ -0,0 +1,133 @@
...
+  - [PR](https://github.com/cardano-foundation/CIPs/pull/1033)
+Created: 2025-04-22
+License: CC-BY-4.0
+---
+
+## Abstract
+
+This CIP defines the procedure for assigning Service Record (SRV) prefixes for Cardano applications (like `cardano-node`, `mithril`, etc.).
+
+This means of sharing authoritative information between SPOs who deploy services and _decentralised protocol_ (DP) developers who write them.
```
abailly: This sentence seems to be lacking a proper verb, or perhaps it should be:
```
This means sharing authoritative...
```
?

I would not use `DP` abbreviation, I find it confusing in the rest of the text. Why not simply use `protocol`?

CIP-0155/README.md
```diff
@@ -0,0 +1,133 @@
...
+
+## Motivation: why is this CIP necessary?
+
+The **Cardano Ledger** allows the use of SRV records in the **SPO** registration certificate (see [register-stake-pool]).
```
abailly, line 23:
```suggestion
The **Cardano Ledger** already allows the use of SRV records in the **stake pool** registration certificate (see [register-stake-pool]).
```

I would distinguish a _stake pool_ from a _stake pool Operator_ but perhaps that's pedantic

CIP-0155/README.md
```diff
@@ -0,0 +1,133 @@
...
+
+In this CIP, we propose to make them usable by decentralised applications running on `Cardano`.
+
+Having access to the ledger (either directly or through tools like `cardano-cli`) will allow decentralised protocols developers to construct networks based on registered relays.
```
abailly: I would invert the 2 sentences above, it seems more logical to me.

CIP-0155/README.md
```diff
@@ -0,0 +1,133 @@
...
+
+## Motivation: why is this CIP necessary?
+
+The **Cardano Ledger** allows the use of SRV records in the **SPO** registration certificate (see [register-stake-pool]).
+
+In this CIP, we propose to make them usable by decentralised applications running on `Cardano`.
+
+Having access to the ledger (either directly or through tools like `cardano-cli`) will allow decentralised protocols developers to construct networks based on registered relays. 
+
+All involved parties (SPOs, DP developers, etc) need to agree on SRV prefixes used by various DPs and thus a public registry governed by a public CIP process is required to foster DPs that co-exist with the **Cardano** network, like **Mithril** (see [CIP#137]).
```
abailly:
```suggestion
All involved parties (SPOs, protocol developers, etc) need to agree on SRV prefixes used by various protocols and thus a public registry governed by a public CIP process is required to foster protocols that co-exist with the **Cardano** network, like **Mithril** (see [CIP#137]).
```

Also, why not use the blockchain itself to store that information?

CIP-0155/README.md
```diff
@@ -0,0 +1,133 @@
...
+
+We consider DPs as applications, which construct their own network and relay on ledger peers published on the block-chain. This include node implementations, or 2nd-layer solutions like Mithril or Hydra.
```
abailly:
```suggestion
We consider protocols as applications, which construct their own network and relay on ledger peers published on the block-chain. This includes node implementations, overlay networks like Mithril, or Layer 2 protocols like Hydra.
```

Mithril is not technically a _Layer 2_

CIP-0155/README.md
abailly: An extended example of what such a record look like would help. Even though I understand the reader can always look at the RFC, having more details immediately accessible avoids context switching.

CIP-0155/README.md
```diff
@@ -0,0 +1,133 @@
...
+
+### Problem
+
+Bootstraping networks requires sharing information about available services.
+
+Historically, this was done by sharing IP or DNS names and PORT numbers, but this approach has its limitations.
+
+SRV records were designed to make deployment independent of hard-coded service end-points (see [RFC#2782]).
+
+Thus they can be instrumental in `Cardano`, since relays are registered on the blockchain through the registration certificate.
```
abailly, line 40: I would group all the previous sentences in a single paragraph as it seems to me they represent a single logical argument: you need info about peers to bootstrap a network, this information is available in the ledger, therefore let's use the ledger for that.

CIP-0155/README.md
```diff
@@ -0,0 +1,133 @@
...
+
+By using SRV records in the registration certificate (which is supported by the `cardano-ledger`, but not by `cardano-node`), we wish to solve this problem not just for `Cardano` node implementation, but also for any DP that requires constructing its own network.
+
+SRV provides a mechanism for exposing DPs co-deployed with a node, like **mithril** or **hydra**.
```
abailly:
```suggestion
SRV provides a mechanism for exposing protocols co-deployed with a node, like **mithril** or **hydra**.
```

CIP-0155/README.md
abailly: I understand this is the _desired state_. While it's possible today to register a stake pool with a SRV record, there's no particular requirement on the content of this record.

CIP-0155/README.md
```diff
@@ -0,0 +1,133 @@
...
+
+Making such services discoverable is one of the key features addressed by this CIP.
+
+In this CIP, we propose prefixes for both **Cardano** and **Mithril**; in the future, other services can be registered through a CIP process, thus starting a registry of prefixes used by the **Cardano** ecosystem.
+
+SPOs who deploy services need to configure their system according to the registry, e.g. SPO's cardano relays node MUST be available at `_cardano._tcp.<SPO_DOMAIN_NAME>`, as other nodes on the system will be looking at this address.
+
+DP developers SHOULD submit proposals to the SRV Prefix Registry, so SPOs, who deploy them, can have an authoritative information how to do it.
```
abailly:
```suggestion
Protocol developers SHOULD submit proposals to the SRV Prefix Registry, so SPOs, who deploy them, can have an authoritative information how to do it.
```

That's where having the "prefix registry" on-chain would be quite helpful. For example, one could imagine a modification to the registry to be a governance action subject to SPOs voting because that's their remit.

CIP-0155/README.md
abailly: DNS entries are structured from the most particular to the most general, so `mithril.cardano.tcp` makes sense and we could imagine an entry for `mithril.ergo.tcp` for example.

CIP-0155/README.md
```diff
@@ -0,0 +1,133 @@
...
+When a **cardano** node implementation reads an SRV record from a ledger, it must add the _cardano_ prefix from the table above before making a DNS lookup, e.g. it should do a DNS query for `_cardano._tcp.<srv-record-from-ledger>`.
+
+This design allows DPs to use SRV records registered in the ledger for different purposes, e.g. a **mithril** node can use them to learn about end-points of its network.
+
+Each prefix SHOULD start with `_cardano._tcp` or `_cardano._udp`, to avoid clashes with services not related to `Cardano`.
+
+#### SRV Registry Rules
+
+* Each DP can have at most one entry in the registry.
+* A CIP process assigns new entries, after a careful consideration and consultation with all the involved parties (see #acceptance-criteria below).
```
abailly: Again, storing that information on-chain and allowing it to change through a governance/SPO vote would make a lot of sense to me.

CIP-0155/README.md
Quantumplation: If so, and there's not already a standard / requirement of using the `_cardano._tcp` prefix, then this represents a much more disruptive change for SPOs, and should be addressed in the implementation plan.

CIP-0155/README.md
coot: Right, I'd be keen to remove `scope`, `mithril` just has two discoverable services. @jpraynaud are you fine with it?

CIP-0155/README.md
coot: @Quantumplation for multi-chain services this process is focused only on cardano side of things, the other side ought to be managed by the other block-chain (unless it's also covered by the CIP process).

CIP-0155/README.md
coot: There are good resources for that, e.g. https://www.cloudflare.com/en-gb/learning/dns/dns-records/dns-srv-record. I'll add a link.

CIP-0155/README.md
jpraynaud: @coot, I guess that @Quantumplation asked for descriptions of the fields `service` and `scope` more than removing them. 

However, we could remove the scope and create multiple services for Mithril that would lead to the same DNS entries if this makes more sense:
| **service** | **SRV prefix**                       | **Status** |
|:------------| ------------------------------------:|:-----------|
| _cardano_   |                      `_cardano._tcp` | Active     |
| _mithril.dmq_   |       `_.dmq._mithril._cardano._tcp` | Active     |
| _mithril.aggregator_   | `_aggregator._mithril._cardano._tcp` | Active     |

CIP-0155/README.md
coot: We got this advice from @njd42, if I recall correctly, it will speed up a DNS query.

CIP-0155/README.md
coot: It means to interpret the entries stored in the ledger according to this CIP, e.g. one needs to pretend `_cardano._tcp` to multi host pool relay information stored in the ledger (how `srv`'s were called by ledger, don't ask me why the name is so complex).

CIP-0155/README.md
coot: > It sounds like this sentence is describing the current state, correct? i

If one has an SRV record that uses a different semantics than what this CIP proposes, it will be broken. However, this doesn't affect `cardano-node`, as it's not being used. So from the `cardano-node` perspective: what used to be broken remains broken :grin:.

CIP-0155/README.md
coot: > As in, there's a field in the SPO registration certificate for it, but the node doesn't actually use that to make SRV requests for peer discovery?

Yes, you got it right :).

CIP-0155/README.md
coot: I'll use _decentralised protocol_, _protocol_ alone is too vague for me.

CIP-0155/README.md
coot: This gives greater flexibility to the SPOs and developers.

CIP-0155/README.md
coot: Making such a link would be a mistake.  Consider a syncing node: depending on the era it would see invalid SRV records depending at which stage of syncing it is.  There is a fundamental difference between the stability of deployment configuration and the elasticity of blockchain-stored information.

CIP-0155/README.md
coot: I argued against it above.

CIP-0155/README.md
coot: I'd keep it simple and not introduce notions which we're not certain will be helpful in the future.  Once a need arises, we'll find an appropriate solution.

CIP-0155/README.md
coot: I'll go with _decentralised protocol_, _protocol_ alone is too vague for me.

CIP-0155/README.md
abailly: I don't understand your argument: No matter what, if a node is syncing then the information about its peers it can get from the ledger is not current, possibly outdated, etc, so it cannot be used anyhow even in its current form without SRV records. 
The information about the (valid) prefixes to use is orthogonal to this problem.

CIP-0155/README.md
abailly: Fair enough.

CIP-0155/README.md
coot: You're right that reconfiguration of relays does occur, but we don't see stake shifting much, which is even more important for a syncing node due to security reasons.

Note that once committed to a value, changing it would be impossible to do safely, as it would require reconfiguring DNS entries by all SPOs simultaneously.

abailly: I still don't understand how this invalidates my proposal, sorry. It's about storing and retrieving some information for the chain which is assumed to be safe. The DNS will need to be verified anyway hence my other proposal to sign those entries.

coot: It will matter for nodes that are syncing recent history (which is the most common scenario for syncing).  The value cannot change on-chain, as using a published genesis snapshot will result in invalid domains if SPO switches to the new value in the meantime, and Daedalus nodes won't be able to sync.  If they cannot change on-chain, why add the complexity of storing it there in the first place?

rphair:
Note to editors (cc @coot): we are still waiting for author confirmation (before merge) that commits have stopped, as requested in https://github.com/cardano-foundation/CIPs/pull/1033#issuecomment-2933326670.

coot:
@rphair I think this proposal should go to `Proposed` as is.  We haven't received a strong voice against it from any of the parties.  Soon, `cardano-node` will support this CIP.  Then, spo-s will be more exposed to how it works in practice. Then it can be marked as `Active` (although this is a bit backwards).

CIP-0155/README.md
```diff
@@ -0,0 +1,133 @@
+---
+CIP: 0155
```
rphair:
```suggestion
CIP: 155
```
"just a number" (no leading zeroes) in this context.

CIP-0155/README.md
```diff
@@ -0,0 +1,133 @@
...
+Title: SRV registry
+Status: Proposed
+Category: Network
+Authors:
+  - Marcin Szamotulski <marcin.szamotulski@iohk.io>
+Implementors: N/A
+Discussions: 
+  - [PR](https://github.com/cardano-foundation/CIPs/pull/1033)
```
rphair: This is YAML, not Markdown, and it's understood that the lowest numbered PR is the submission of the CIP (so no need to label any differently in this case).  If you wish to label a link in this context it would be, e.g. (note `PR` will likely not be distinctive going forward):
```suggestion
  - Submission: https://github.com/cardano-foundation/CIPs/pull/1033
```

CIP-0155/README.md
```diff
@@ -0,0 +1,133 @@
...
+
+### Implementation Plan
+
+Each **Cardano** node implementation or other tools which rely on SRV records stored in the ledger should comply with this proposal.
+
+## Rationale: how does this CIP achieve its goals?
+
+This CIP constructs a process to maintain SRV registry, and thus provides authritative information for SPOs and DP developers.
+
+
```
rphair: This section needs to be moved back before _Path to Active_. https://github.com/cardano-foundation/CIPs/blob/master/CIP-0001/README.md#structure

CIP-0155/README.md
```diff
@@ -0,0 +1,133 @@
...
+  - [PR](https://github.com/cardano-foundation/CIPs/pull/1033)
+Created: 2025-04-22
+License: CC-BY-4.0
+---
+
+## Abstract
+
+This CIP defines the procedure for assigning Service Record (SRV) prefixes for Cardano applications (like `cardano-node`, `mithril`, etc.).
+
+This means of sharing authoritative information between SPOs who deploy services and _decentralised protocol_ (DP) developers who write them.
```
Crypto2099:
```suggestion
This creates a means of sharing authoritative information between SPOs who deploy services and _decentralised protocol_ (DP) developers who write them.
```
The sentence is a little bit... unclear in its current form

CIP-0155/README.md
```diff
@@ -0,0 +1,133 @@
...
+
+SRV records were designed to make deployment independent of hard-coded service end-points (see [RFC#2782]).
+
+Thus they can be instrumental in `Cardano`, since relays are registered on the blockchain through the registration certificate.
+
+By using SRV records in the registration certificate (which is supported by the `cardano-ledger`, but not by `cardano-node`), we wish to solve this problem not just for `Cardano` node implementation, but also for any DP that requires constructing its own network.
+
+SRV provides a mechanism for exposing DPs co-deployed with a node, like **mithril** or **hydra**.
+
+Making such services discoverable is one of the key features addressed by this CIP.
```
Crypto2099, line 46: On the issue of making this process discoverable, I note that there is simply a table here in markdown. Would it make sense to follow the lead of other CIPs like CIP-10 and provide a JSON and a semantic schema for adding/removing entries to that document so that future additions or revocations can be programmatically consumed? 

In general the CIP Editors have been loathe to modify and change existing CIPs after they have entered "accepted" status due to possibly introducing breaking changes to unknown consumers of said CIPs. This would suggest that future additions to this table would need to construct a new CIP and relevant discussion, which would then ultimately end up in revoking or creation of a newly accepted/expected SRV prefix reservation.

I think it would be good to have this CIP, as it proposes to create this SRV registry, to contain information about how a future CIP will add a new prefix and the requirements and steps to do so.

CIP-0155/README.md
```diff
@@ -0,0 +1,133 @@
...
+
+## Specification
+
+### SRV Prefix Registry
+
+| **service** | **scope**     | **SRV prefix**                       | **Status** |
+|:------------|:--------------| ------------------------------------:|:-----------|
+| _cardano_   |               |                      `_cardano._tcp` | Active     |
+| _mithril_   | _dmq_         |       `_.dmq._mithril._cardano._tcp` | Active     |
+| _mithril_   | _aggregator_  | `_aggregator._mithril._cardano._tcp` | Active     |
```
Crypto2099: See earlier comment r.e. the process for modifying and/or changing entries within this table, I think this would be better served to be hosted via a JSON or other machine-readable file as part of this CIP process (documentation here doesn't hurt in templating how future CIPs will propose to revoke and/or add new SRV prefixes).

Example Schema:
```json
{
  "$schema": "http://json-schema.org/draft-07/schema#",
  "title": "SRV Prefix Registry",
  "type": "object",
  "patternProperties": {
    "^.+$": {
      "type": "object",
      "properties": {
        "service": {
          "type": "string",
          "description": "The name of the service, e.g., '_cardano_'"
        },
        "scope": {
          "type": "string",
          "description": "Optional scope qualifier, e.g., '_dmq_'",
          "default": ""
        }
      },
      "required": ["service"],
      "additionalProperties": false
    }
  },
  "additionalProperties": false,
  "description": "A mapping where each key is a unique SRV prefix, and its value describes the service and optional scope."
}
```

Example `srv.registry.json` file:
```json
{
  "_cardano._tcp": {
    "service": "_cardano_",
    "scope": ""
  },
  "_.dmq._mithril._cardano._tcp": {
    "service": "_mithril_",
    "scope": "_dmq_"
  },
  "_aggregator._mithril._cardano._tcp": {
    "service": "_mithril_",
    "scope": "_aggregator_"
  }
}
```

This would also allow us to add a CI/CD Action to the CIP registry ensuring that future values are unique and that service + scope are limited to a single entry, etc.

CIP-0155/README.md
Crypto2099: I don't know that it's the realm of a CIP to recommend "best practices for DNS lookup" and this may or may not have an impact depending on the language the implementation is using? I would advocate to remove this or add the context that it is a performance enhancement for implementers only.

CIP-0155/README.md
```diff
@@ -0,0 +1,133 @@
...
+When a **cardano** node implementation reads an SRV record from a ledger, it must add the _cardano_ prefix from the table above before making a DNS lookup, e.g. it should do a DNS query for `_cardano._tcp.<srv-record-from-ledger>`.
+
+This design allows DPs to use SRV records registered in the ledger for different purposes, e.g. a **mithril** node can use them to learn about end-points of its network.
+
+Each prefix SHOULD start with `_cardano._tcp` or `_cardano._udp`, to avoid clashes with services not related to `Cardano`.
+
+#### SRV Registry Rules
+
+* Each DP can have at most one entry in the registry.
+* A CIP process assigns new entries, after a careful consideration and consultation with all the involved parties (see #acceptance-criteria below).
```
Crypto2099: I would like to see more definition of what this future CIP process for adding, deprecating, or revoking SRV prefixes look like including versioning rules (note that the schema and example shared above have top-level versioning as well as per-entry versioning to record changes to statuses).

rphair:
@coot - in addition to the document nitpicks of my https://github.com/cardano-foundation/CIPs/pull/1033#pullrequestreview-3025480269 we have @Crypto2099's more substantial questions about versioning & updates in https://github.com/cardano-foundation/CIPs/pull/1033#pullrequestreview-3036097796 that will need to be accommodated or discounted before merging.

I think it's OK to keep this at `Last Check` in case all the latter issues can be addressed tomorrow or early the day after: but we will have to either assure this before merging or put back to `Confirmed` status to allow more time for these perhaps substantial changes (particularly the suggested enhancement of a JSON database as per https://github.com/cardano-foundation/CIPs/pull/1033#discussion_r2217890722).

CIP-0155/README.md
Quantumplation: I read it more like "this method of sharing"; i.e. means here was a noun

Crypto2099: If we're talking about the "means" of production here then it's still unclear as written... what about the means of sharing between the two parties?

rphair: @Crypto2099 is right... this is currently just a big long noun phrase with no verb.  Adding the verb `creates` solves the problem neatly.

CIP-0155/README.md
```diff
@@ -0,0 +1,133 @@
...
+
+### Acceptance Criteria
+
+The CIP should be accepted by:
+
+* [ ] [**Technical Steering Committee**][tsc]
+
+And when there's no major objection from one of the currently involved parties:
+
+* [ ] [**Amaru Team**][amaru] accepts the proposal
```
Ryun1, line 108:
```suggestion
* [ ] [**Amaru Team**][amaru] accepts and implements the proposal
```
perhaps to be more specific?

rphair:
Putting back to `Confirmed` status as per https://github.com/cardano-foundation/CIPs/pull/1033#issuecomment-3094701812 since the meeting will be starting in a few minutes without the above editorial comments being addressed... which therefore we would be doing at the meeting anyway.  @coot we look forward to resolving the above review points & getting this back into pre-merge status soon.

CIP-0155/README.md
coot: I find it strange to use pattern properties (e.g. keys as srv prefixes).  I'd suggest we'd go with:
```json
[
  { "service": "cardano",
    "prefix": "_cardano._tcp",
    "status": "Active"
  },
  { "name": "mithril.dmq",
    "prefix": "_dmq._mithril._cardano._tcp",
    "status": "Active"
  },
  { "name": "mithril.aggregator",
    "prefix": "_cardano._tcp",
    "status": "Active"
  }
]
```

and a schema:
```json
{
    "$schema": "http://json-schema.org/draft-07/schema#",
    "$id": "https://github.com/cardano-foundation/CIPs/blob/master/CIP-0155",    
    "title": "SRV Registry",
    "description": "JSON schema for the SRV registry",    
    "type": "array",
    "items": {
      "type": "object",
      "properties": {
        "service": {
          "type": "string",
          "description": "A unique identifier for the service name."
        },
        "prefix": {
          "type": "string",
          "description": "A unique SRV prefix.",
          "maxLength": 255
        },
        "status": {
          "type": "string",
          "description": "Status of the entry. Can be one of: \"Active\", \"Revoked\""
        }
      },
      "required": [
        "service",
        "prefix",
        "status"
      ],
      "additionalProperties": false
    }
  }
```

CIP-0155/README.md
coot: It is supposed to be very stable; the only thing that can be changed by a CIP process is `status` from `Active` to `Revoked`. I don't think we need versioning for it.

CIP-0155/README.md
coot: The CIP is not created to track implementation status, so I'd like to avoid that.

rphair: @coot you are not required to track any implementations going forward: but if you don't, someone else must do that, at least during the transition from `Proposed` to `Active`.  Therefore these criteria need to be agreed between author, reviewers & editors ahead of that time.

We have lots of CIPs already that haven't been tracked after that point: we'd only generally be logging implementations in the CIP itself when they provide _initial_ proof that a CIP is active in the ecosystem.

coot:
I pushed some changes requested by the reviewers.

CIP-0155/README.md
rphair: @coot please correct me & unresolve if I am wrong but this has been addressed by https://github.com/cardano-foundation/CIPs/pull/1033/commits/284cf67233931562216dfa6c3a7e84af654b2164.

rphair (approved):
Thanks @coot for the last round of commits which, as far as I can tell, have addressed all the editorial reservations so far except:
* [x] (checklist of implementations in Path to Active) https://github.com/cardano-foundation/CIPs/pull/1033#discussion_r2239237023
* [x] (more detailed versioning may be needed) https://github.com/cardano-foundation/CIPs/pull/1033#discussion_r2239207882
* [ ] (ticking boxes for current implementations) https://github.com/cardano-foundation/CIPs/pull/1033#discussion_r2243042033

Since I think the 2 issues above are optional (not required for the initial merge of the document) I'm approving this and tagging it for `Last Check`: anticipating these 2 issues might be resolved before the meeting or dismissed at the meeting (@coot you are welcome to join us as always, especially if not getting any other editor approvals before then): https://hackmd.io/@cip-editors/117

CIP-0155/README.md
coot: For me, this sounds like a good idea for a separate CIP, because one needs to track all the implementations (and they will be growing in numbers).  Tracking this in each CIP will duplicate work, while one CIP with a feature table would be much simpler to maintain.

rphair: @coot this is only for editorial purposes so that we can be assured of `Active ` status & we definitely don't need that degree of awareness about implementations.

Since generally it's editors who assess that (after getting input from everyone: and you would likely be the best judge of when it's `Active`) I'm going to suggest at the meeting we merge this even if the above suggestion isn't taken.  Yet I <ins>also</ins> think it would be OK to merge after @Ryun1 committing his suggestion (though I'll leave this to other editors).

Therefore I'm resolving this & ticking it off the checklist below (i.e. for our `Last Check` review) since as you say this is not a significant detail for the CIP.  Any future PR to upgrade the CIP to `Active` status will certainly involve you + editors and we'll make a comprehensive assessment at that time, including status of SRV Registry implementations on any viable node types.

coot: It's will be released in `cardano-node-10.6`, so it can be marked `Active`.

rphair: thanks @coot - the fact that the SRV Registry is "active" in the Haskell node is another reason I still think we can dismiss the suggestion above about confirming implementations: which I believe @Ryun1 intended as a proof of adoption.

For marking <ins>the CIP itself</ins> `Active`, we would have to wait until ticking the boxes currently set [in the Acceptance Criteria](https://github.com/coot/CIPs/blob/srv-registry/CIP-0155/README.md#acceptance-criteria).  Until then we are free to merge this as `Proposed`.

This will all be reviewed by other editors no later than Tuesday's CIP meeting next week: though it's OK the way it is as far as I'm concerned.

CIP-0155/README.md
```diff
@@ -0,0 +1,136 @@
...
+* [ ] [**Cardano-Node Team**][cardano-node] accepts the proposal
+* [ ] [**Gouroboros Team**][gouroboros] accepts the proposal
+* [ ] [**Hydra Team**][hydra] accepts the proposal
+* [ ] [**Mithril Team**][mithril] accepts the proposal
```
rphair:
```suggestion
* [x] [**Cardano-Node Team**][cardano-node] accepts the proposal
* [ ] [**Amaru Team**][amaru] accepts the proposal
* [ ] [**Gouroboros Team**][gouroboros] accepts the proposal
* [ ] [**Hydra Team**][hydra] accepts the proposal
* [ ] [**Mithril Team**][mithril] accepts the proposal
```
I assume the proposal must at least have been "accepted" if [already in `cardano-node-10.6`](https://github.com/cardano-foundation/CIPs/pull/1033#discussion_r2242501265).

rphair (commented):
@coot it seems a minor rearrangement of _Path to Active_ is in order based on your https://github.com/cardano-foundation/CIPs/pull/1033#discussion_r2242501265: what do you think? ...

CIP-0155/README.md
Crypto2099: Fair enough, I'll go ahead and mark this as resolved with the understanding that future additions or revocations and subsequent changes to the registry should only happen through their own, new CIP process and not via versioning or iterating on previous work.

rphair:
FYI @coot our next CIP meeting where we had hoped to merge this (https://hackmd.io/@cip-editors/117) has been postponed from tomorrow until 19 August due to mass attendance at Rare Evo.  If you need this to be merged for practical reasons please say so and perhaps @Ryun1 @perturbing can OK this for merge before that meeting.

coot:
On the 19th of August I am on Holidays :tada:

rphair:
OK @coot, given the lack of urgent objection by @Ryun1 and @perturbing, with affirmation by @Crypto2099 whose skills as an operations reviewer I have complete confidence in, I'll merge this considering the time delay we would suffer by waiting to do it in person.  Your comments above make it clear to the editors that any updates to the details would come in completely & promptly through future PR's. 😎

PR #1038, opened 2025-05-15
CIP-0035 | Remove serialization & costing requirements, apply to CIP-0138

ana-pantilie:
The Plutus Core team has discussed that requiring users to provide a detailed way to (de)serialize new builtin types and to propose costing functions is too technical for CIP authors and should be left for the team to figure out.

This PR also removes the serialization details from CIP-0138 as we have discovered that the implementation will differ a bit from what was proposed.

We have also discussed that technical details like the above have historically been documented in the [Plutus Core Specification](https://plutus.cardano.intersectmbo.org/resources/plutus-core-spec.pdf) and that should represent the source of truth in these regards.

CIPs should remain proposals from the community and we don't want to discourage the community from writing them.

cc: @zliu41 @effectfully @kwxm @Unisay


zliu41 (approved):
LGTM

rphair (approved):
thanks @ana-pantilie & @zliu41 - I agree with this rationale (having seen it as a big hurdle to submissions in the last 3-4 years) and can see what a difference it makes in the case that it is applied to.

effectfully (approved):
:+1:

Ryun1 (approved):

PR #1051, opened 2025-07-08
CIP-0010: Change Reeve metadata label after the official launch

Kammerlo:
We are launching Reeve today and we decided to move the label to `1447`. 
Our transactions are published under this label already. 

The documentation of the onChain format is here: [Link](https://github.com/cardano-foundation/cf-reeve-platform/blob/docs-updating-metadata-label-1/docs/onChainFormat.md)


Ryun1:
Hey @Kammerlo 

just to be safe -- is there any danger in changing this? could it break any implementations?

and the supplied documentation seems to have moved from the link

matiwinnetou:
It would make 2023 CF data (soft Reeve launch) maybe not invalid but it was using different label, the one that is being removed.

rphair:
@matiwinnetou (cc @Kammerlo) what about keeping the old tag number in there as a "legacy" metadata tag... titled something clearly not current like `Reeve - legacy (pre-launch) metadata (obsolete)`?

Kammerlo:
Yes as Mati said, the softlaunch data is published under the "old" label. But with the launch the new label is what matters. And the old one is obsolete. I will add it for documentar purposes, but will add that it's obsolete. 
@Ryun1 @rphair

Kammerlo:
@rphair Yes I think it's a good point from @Godspeed-exe 
When I submitted my first CIP-10 metadata label: I checked how others linked the metadata description and thought that having it in the PR note (as far I did it this way, since others did the same) is a little bit unhandy. So harder to find for users, if they want to understand it.

I like the approach, that it's optional. There might be usecases which want to just "block" the label, but if there is a usecase which wants others to really understand it this registry is the first place a user will check. Perhaps another field like an url to the project makes sense as well, but this would be more for human-readability.

PR #1056, opened 2025-07-10
CIP-108 CIP-136 | Small schema fixes

Ryun1:
### Changes
  
- fix naming of `@type` property within CIP108 and CIP136 schemas

### DIFF
--- a/CIP-0100/cip-0100.common.schema.json
+++ b/CIP-0100/cip-0100.common.schema.json
@@ -15,7 +15,6 @@
             "title": "Authors",
             "description": "The authors of this governance metadata",
             "type": "array",
-            "minItems": 1,
             "items": {
                 "$ref": "#/definitions/Author"
             }
--- a/CIP-0100/cip-0100.common.schema.json
+++ b/CIP-0100/cip-0100.common.schema.json
@@ -15,6 +15,7 @@
             "title": "Authors",
             "description": "The authors of this governance metadata",
             "type": "array",
+            "minItems": 1,
             "items": {
                 "$ref": "#/definitions/Author"
             }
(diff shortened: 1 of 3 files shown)


rphair (approved):
Structure of changes look fine to me but I would look forward to other editors' technical review of context & possible consequences.

gitmachtl:
I welcome this change, looks good.

Crypto2099:
Not sure that I understand the intent here? Having an empty array and thereby signalling no authors or anonymous authors was always a feature rather than a "bug" as far as I remember from the original workshops w/ @Quantumplation

Quantumplation:
Yes; I don't think it should be the purview of these specs to define what is required or not; it's to describe *how to interpret* the fields, if present.

It may make recommendations to tool authors about how to highlight deficiencies (ex: you might want to recommend that the lack of authors is highlighted clearly because of the anonymity, etc). But saying it's "required" implies the tool shouldn't show it at all if it's not present, or should mark it as invalid, which puts the thumb on the scale of governance.

rphair:
Will note  https://github.com/cardano-foundation/CIPs/pull/1056#issuecomment-3089431601 & https://github.com/cardano-foundation/CIPs/pull/1056#issuecomment-3089762749 for Tuesday's agenda & trust @Crypto2099 and @Ryun1 will debate both sides of the issue and include @Quantumplation's interpretation of the proposed change.

Ryun1:
Thanks for the feedback @Crypto2099 @Quantumplation 

you raise good points
and I agree

I have removed the requirement via [741ed74](https://github.com/cardano-foundation/CIPs/pull/1056/commits/741ed74747b7c9ba0d6a8894a0cc9eb9e67a091d)

Crypto2099 (approved):

PR #630, opened 2023-12-01
CPS-???? | Ada token supply exhaustion

xezon:
Hello. I hope this is not a duplicate and helps. At a glance I could not find a CPS for it.

Keep up the good work!

---

([rendered in PR branch](https://github.com/xezon/cardano-CIPs/blob/cps-ada-token-supply-exhaustion/CPS-XXXXX/README.md))

PR #1060, opened 2025-07-22
CPS-0022? | Lost Stake

Cerkoryn:
Completed some research put together this CPS to formally define a problem that we (the Cardano Incentives Working Group) have identified.

It is unfortunately difficult to measure with quantifiable numbers, but this should be explained well in the CPS.

---

([rendered](https://github.com/Cerkoryn/CIPs/blob/sticky-stake/CPS-0022/README.md))


Ryun1:
Cross referencing with past _similar_  [CPS-???? | Ada token supply exhaustion](https://github.com/cardano-foundation/CIPs/pull/630)

CPS-Sticky-Stake/README.md
```diff
@@ -0,0 +1,132 @@
...
+
+> **Note:** The exact amount of ADA that is lost but still delegated is difficult to quantify. We use a *conservative* working assumption of ~1.5% of the circulating ADA being lost per year, which is less than half the 4% per-year loss rate measured for Bitcoin by Ledger and Chainalysis.
+
+Lost stake and lost ADA have several detrimental effects on the Cardano network:
+
+- **Perpetual reward dilution:**  
+  Each epoch, a portion of the total ADA rewards is distributed to all staked ADA (active or lost). Rewards going to addresses with lost ADA are effectively removed from circulation permanently, which means active delegators and stake-pool operators (SPOs) receive a smaller share than they would if that lost stake were absent. In essence, active participants subsidize the lost stake. Over years and decades, the compounding nature of this lost ADA can significantly dilute the reward pool available to real users and operators.
```
rphair: I imagine in discussions — which, according to your premise, might go on for decades — you will want to deep-link to these statements.  Therefore we should use the built-in Markdown tech of headings that will provide links to each concept (for sharing on forum / social networking, code comments, other documents, etc.):
```suggestion
### Detrimental effects

Lost stake and lost ADA have several detrimental effects on the Cardano network:

#### Perpetual reward dilution
  Each epoch, a portion of the total ADA rewards is distributed to all staked ADA (active or lost). Rewards going to addresses with lost ADA are effectively removed from circulation permanently, which means active delegators and stake-pool operators (SPOs) receive a smaller share than they would if that lost stake were absent. In essence, active participants subsidize the lost stake. Over years and decades, the compounding nature of this lost ADA can significantly dilute the reward pool available to real users and operators.
```
... and so forth, all the way down through the _Goals_ section (note these topics are H4's so they don't seem in the same topic as the H3's further below).

CPS-Sticky-Stake/README.md
```diff
@@ -0,0 +1,132 @@
...
+    - Rich Manderino [ECP] 
+    - W Cataldo 
+    - Carlos Lopez de Lara 
+Proposed Solutions: []
+Discussions:
+Created: 2025-07-22
+License: CC-BY-4.0
+---
+
+## Abstract
```
rphair, line 15: This would likely be generally considered too long for an _Abstract_: in CIP/CPS documents working best as 1 or 2 short paragraphs.  Most of the material currently in this section is really the _Problem_ and I think your impact (and CPS-9999 compliance) would be improved by condensing this section into a brief introduction of the terminology used and then (in a second short paragraph) how the problem is defined.

There are conventionally no graphs in an abstract although this one would certainly also be welcome in the _Problem_ section.

CPS-Sticky-Stake/README.md
```diff
@@ -0,0 +1,132 @@
+---
+CPS: ????
+Title: Lost Stake
+Category: ?
+Status: Open
+Authors:
+    - Ryan Wiley <rian222@gmail.com>
+    - Rich Manderino [ECP] 
+    - W Cataldo 
+    - Carlos Lopez de Lara
```
rphair: SPOs and other community advocates that have been following this & similar issues for years will naturally know who/what ECP (@EarnCoinPool) is, but considering the general audience and proposed duration of this problem / message we will need something more generally recognisable at least during the review period & submission of whatever CIPs you think may follow in the near term.

Also @CarlosLopezDeLara could use a credential about his current work arrangements.  It is understood from the context that CIP submissions don't necessary reflect the views of the companies that the authors are working for at the time... and the open-source declaration is thankfully enough to pre-empt any issues of IP ownership. 😎

rphair (commented):
Thanks again @Cerkoryn for another academically informed submission.  It's already been tagged for `Triage` at next week's CIP meeting (https://hackmd.io/@cip-editors/117) & you and other authors (@EarnCoinPool, @CarlosLopezDeLara, ?) are welcome to talk about it there & observe initial responses especially if not already posted here.

My understanding is, as @Ryun1 points out in https://github.com/cardano-foundation/CIPs/pull/1060#issuecomment-3103734165, that this is a "horizon" problem and I would not want to make the mistake every generation makes about social, economic, political and human conditions always continuing on the same trajectory of "progress".  For instance, the assumption of a generally available 3% compounded interest continuing until year 2070 and beyond seems to ignore many disruptive scenarios that are likely to occur (and compound) long before then.

Still I would rather see this CPS on the books and see if what kind of bids we get in the next few years for CIPs that might provide a better outlook.  I think most in our community would agree that long-term responsibility leads to short-term equity, and therefore would like to see this CPS in place.

Cerkoryn:
@rphair 

> My understanding is, as @Ryun1 points out in https://github.com/cardano-foundation/CIPs/pull/1060#issuecomment-3103734165, that this is a "horizon" problem and I would not want to make the mistake every generation makes about social, economic, political and human conditions always continuing on the same trajectory of "progress". For instance, the assumption of a generally available 3% compounded interest continuing until year 2070 and beyond seems to ignore many disruptive scenarios that are likely to occur (and compound) long before then.

This is very true and part of the problem in doing research to determining the true depth of this problem.  It's extremely difficult (arguably impossible) to determine which addresses are actually permanently unreachable using on-chain data, so even our best, educated guesses end up being wild estimates.

However, there are some important things to note for why we ended up on the 3% number.  Historically, Cardano's rewards APY has been as high as 8%.  If ADA was lost during this time, that 8% would be compounding much faster than our 3% number.  As far as the future is concerned, it is true that disruptive scenarios are almost certain to occur.  We'd roughly anticipated that things like partner chains or other future changes will prop up the rewards rate at some point.  Otherwise the trend shows it going below 1% APY by 2030, which is already on the edge of sustainability.

Another point worth mentioning is that we tried to calculate just how much ADA is being sent into the void CURRENTLY based on our estimates for Lost ADA.  If you reference this chart:
image

Notice that the red line appears to be roughly crossing the 5B ADA mark around the current year 2025 (Note that this is also using the flat 3% compound instead of the historically higher APY numbers!).  

 - If ~50% of that ADA is still staked, then that represents 2.5B Lost ADA that is still earning rewards.   
 - Currently there is roughly ~21B ADA staked.
 - The rewards pot (after treasury cut, unmet pledge, pool performance, and the unstaked proportion is taken out) is around 7.5M ADA distributed each epoch
 - 2.5B out of 21B is about 12%.
 
Based on this very rough napkin math, that would mean 12% of the rewards pot is ALREADY being wasted each epoch.  This represents 900k ADA each epoch, or 65.7M ADA each year.  It's important to empathize that this is only an estimate, but it's also using conservative numbers.  The actual amount could be better or worse.  We wanted to include this in the CIP, but felt it might be too alarming or sensationalist and without enough hard support to justify mentioning directly.  So instead we used the language...

`Cardano already distributes a significant amount of ADA every epoch in staking rewards to addresses that are permanently inaccessible`

In my opinion, even if we accept more conservative estimates, it's very likely that this is a already a problem TODAY that is hurting our fee efficiency and staking rewards.  Not just a long term horizon problem.

rphair:
@Cerkoryn if you can put the essence of https://github.com/cardano-foundation/CIPs/pull/1060#issuecomment-3133741195 into the _Problem_ section of the CPS — under a heading where it can be easily spotted by readers who will by default consider this a far-term problem — I will call attention to it at our next meeting and suggest to fellow reviewers that it should be considered near-term if not already in effect & ready to address with tangible solutions.

**p.s.** even if this economic vulnerability is in fact already significant, I would much rather see it documented by Cardano's advocates than its adversaries.

TerminadaPool:
A couple of ideas for consideration:

#### Quantifying lost or inactive stake

We are going to gain insight into the scale of the problem after the Midnight glacier drop concludes.  The Night token will be independently valuable and thus it could be reasonably assumed that any Cardano users not claiming their Night tokens either aren't paying attention, don't care, or can't use their keys.  Even though we may not be able to identify which individual addresses represent lost or inactive keys, we will get an idea of that proportion of Ada.

#### Education problem: Staking rewards are not interest but rather reward for active participation

I think it would be good if this CPS attempted to address the education problem regarding why Cardano even has staking.  IE: What is the purpose of staking in the protocol and why do stakers earn rewards for "staking"?

Many users think they are entitled to staking rewards just like receiving interest on a bank account.  A kind of set and forget thing where they just need to do this initial extra step of selecting a stake pool, and now also a DRep, in order to receiving staking rewards.  However, if rewards were supposed to be like interest on a bank account then there would be no need for the selection of a stake pool and the "interest" could simply accumulate automatically.  The code could be simpler too because there would be no need for all this staking stuff.

We need to educate people that staking is voluntary, and those that choose to participate in staking are agreeing to be involved in an active process.  Staking is how the protocol decides which pools get to make blocks, and that underpins the security of the ledger.  Stakers are effectively performing a contract whereby they agree to monitor Cardano network health, and the performance of their chosen pool, in return for receiving a portion of the block production rewards.  Stakers are actually voting for a pool operator using a form of liquid democracy to determine block production and they are expected to change that vote should network conditions change through malicious behaviour or poor performance.  Users that are unable to actually change their chosen pool, should the need arise, are not fulfilling their side of the contract and therefore should not be entitled to rewards.

Lost or inactive stake is not actually contributing to securing Cardano and it would be better if it was somehow "unstaked" for two reasons:
1. It would be better for security since inactive stake provides some older pools with extra control over block production that they shouldn't be entitled to _forever_.  When eventually one of the OG multi-pool operators sells his pool keys to Blackrock then maybe people might understand the security threat from "sticky stake" better.
2. The limited pot of rewards could be directed to the real, active, participants so they receive more instead of throwing rewards away at undeserving wallets.

#### Ways to mitigate the problem

One possible mitigation strategy could be to automatically deem stake no-longer active after say 1yr, thus requiring re-staking or at least confirmation of current stake pool selection.  A kind of proof-of-life that the user is capable of signing with his keys.  This would at least place a limit on pools continuing to accumulate "sticky" or inactive stake.

rphair:
FYI @Cerkoryn the CIP meeting # 117 referred to above — where this is on the agenda for `Triage` — has been postponed 2 weeks (due to Rare Evo).

Cerkoryn:
@TerminadaPool Agree on all points.

The NIGHT airdrop may be a helpful way to gauge/estimate alive addresses.  I was toying with the idea of looking at ADA in reward account (non-UTxO) balances for addresses that have been inactive for a long time as well.

> Education problem: Staking rewards are not interest but rather reward for active participation

For this and for your suggested solution, I plan to address both issues in the CIP that I am currently writing to address this CPS.

Cerkoryn:
@rphair I made some changes based on the comments you left here for your review.

rdlrt:
> The NIGHT airdrop may be a helpful way to gauge/estimate alive addresses

Are you sure? Some of us dont really care about that token or believe in the management of it and wouldnt want to touch cold storage for it. It's value is very much opinionated and IMO - shouldnt count for much for representing 'activity'

rphair:
thanks @Cerkoryn - the writing looks really well organised now and I doubt there would be dispute that at least the problem statement is rational.  So I feel sure it will be accepted as a candidate at the CIP meeting tomorrow (https://hackmd.io/@cip-editors/117) though you might be interested in coming anyway since I think all editors would be curious to hear more about the CIP you have developing from this.

CPS-Sticky-Stake/README.md
```diff
@@ -0,0 +1,165 @@
...
+## Open Questions 
+- How can the protocol reliably identify truly lost stake keys?
+- What inactivity period (epochs/years) would be acceptable before stake is considered “lost”?
+- Which economic / social mechanisms can prevent reward dilution without violating property rights?
+
+## References
+
+- **Cardano Improvement Proposal 9999 (CIP-9999):** Cardano Problem Statements
+  [cips.cardano.org](https://cips.cardano.org/cip/CIP-9999)
+
```
rphair:
```suggestion
```
@Cerkoryn though I would be OK leaving this in at your discretion, the reference to CIP-9999 _is impicit_ for any CPS... just as CIP authors aren't expected to refer to CIP-0001 regarding the same choices in content & format that everyone else is making.  If you feel it supports your other writing, I would consider it justifiable to leave it in: even though editorially it's redundant.

Cerkoryn:
> Are you sure? Some of us dont really care about that token or believe in the management of it and wouldnt want to touch cold storage for it. It's value is very much opinionated and IMO - shouldnt count for much for representing 'activity'

@rdlrt TBH I am not sure that it will help all that much.  It is just another partial piece of information that may help us get a better idea for the open question of how to reliably identify lost stake keys.

CPS-Sticky-Stake/README.md
```diff
@@ -0,0 +1,162 @@
...
+License: CC-BY-4.0
+---
+
+## Abstract
+
+In everyday Cardano discussions the umbrella term “Sticky Stake” is used for any stake that stubbornly remains delegated, regardless of whether its owner is still around. This Cardano Problem Statement (CPS) zooms in on the most critical slice of that phenomenon that we dubbed “Lost Stake”: ADA that (a) remains delegated to a stake pool or dRep, yet (b) can never again be moved because the controlling private keys are irretrievably lost (e.g., seed-phrase loss, death of the sole key holder, catastrophic wallet failure).
+
+This CPS formalises the Lost Stake problem and quantifies its systemic impact: dilution of the reward pot available to active participants, distortion of pool-selection incentives, and ossification of governance power. Estimates suggest that as much as 900k ADA per epoch (67M ADA each year) may already be draining into permanently unreachable wallets. Left unchecked, compounding Lost Stake will siphon billions of ADA in rewards and an ever-growing share of voting weight into wallets that nobody controls, making future remediation far costlier and more contentious. 
+
+---
```
Ryun1:
```suggestion
```
remove extra syntax

CPS-Sticky-Stake/README.md
```diff
@@ -0,0 +1,162 @@
...
+    - Ryan Wiley <rian222@gmail.com>
+Proposed Solutions: []
+Discussions:
+Created: 2025-07-22
+License: CC-BY-4.0
+---
+
+## Abstract
+
+In everyday Cardano discussions the umbrella term “Sticky Stake” is used for any stake that stubbornly remains delegated, regardless of whether its owner is still around. This Cardano Problem Statement (CPS) zooms in on the most critical slice of that phenomenon that we dubbed “Lost Stake”: ADA that (a) remains delegated to a stake pool or dRep, yet (b) can never again be moved because the controlling private keys are irretrievably lost (e.g., seed-phrase loss, death of the sole key holder, catastrophic wallet failure).
```
Ryun1:
```suggestion
In everyday Cardano discussions the umbrella term “Sticky Stake” is used for any stake that stubbornly remains delegated, regardless of whether its owner is still around. This Cardano Problem Statement (CPS) zooms in on the most critical slice of that phenomenon that we dubbed “Lost Stake”: Ada that (a) remains delegated to a stake pool or dRep, yet (b) can never again be moved because the controlling private keys are irretrievably lost (e.g., seed-phrase loss, death of the sole key holder, catastrophic wallet failure).
```
a small suggestion on
ADA -> Ada

([this](https://github.com/cardano-foundation/CIPs/tree/master/CIP-0120#ada-vs-ada-vs-ada) is the guidelines i use on capatalization)

CPS-Sticky-Stake/README.md
```diff
@@ -0,0 +1,161 @@
...
+
+This scenario highlights the long-term risk of inaction. Even a modest annual loss rate combined with ongoing rewards can, in theory, lead to exponential increases in the amount of ADA effectively trapped as lost stake. Over decades, lost ADA (plus the rewards it continually accrues) could constitute an ever-growing share of the total circulating supply, potentially tens of billions of ADA by the 2050s–2070s in this model. However, it is important to note that the likelihood of this scenario fully playing out is low as the proportion of rewards and transaction fees paid to lost ADA grows, the situation would become increasingly unacceptable to active users. At some point, most users would likely abandon the ecosystem rather than continue subsidizing unreachable addresses, making such runaway growth of lost stake unsustainable in practice. Nevertheless, this projection magnifies all the aforementioned issues: reward dilution becomes more severe, stake pools and governance are heavily influenced by non-recoverable funds, and the active Cardano community would be supporting an increasing “dead weight” in the ecosystem until a breaking point is reached.
+
+---
```
Ryun1:
```suggestion
```

CPS-Sticky-Stake/README.md
```diff
@@ -0,0 +1,161 @@
...
+
+- **Economic Sustainability:** In the long term, as block-reward inflation tapers off, Cardano’s security will hinge on transaction fees and community participation. If a significant chunk of ADA is effectively out of economic circulation (yet still “consuming” rewards/fees), it means the active economy has to carry that burden. The security budget (total incentives for validators) would be partially drained to non-participants. This inefficiency will necessitate higher fees or protocol changes to compensate, which is undesirable for growth. In short, allowing lost stake and lost ADA to grow unchecked may undermine the sustainability of the network’s incentive model.
+
+---
```
Ryun1:
```suggestion
```
removing extra markdown formatting for consistency across CIPs/CPSs

CPS-Sticky-Stake/README.md
```diff
@@ -0,0 +1,159 @@
...
+CPS: ????
```
rphair:
```suggestion
CPS: 22
```

rphair (commented):
@Cerkoryn thanks for coming for preliminary review at yesterday's CIP meeting, which confirmed that the presentation of the problem is valid within the usual margin of error for predicting the future decades in advance. 😅

Please also rename the containing directory to `CPS-0022` and update the rendered proposal link accordingly in your OP 🎉

Other notes based on what was said at the meeting:
* We appreciate the limitation of the `Lost Stake` problem as a special case / subset of the `Sticky Stake` problem in the sense that the former problem may be be "solvable" while the latter problem may not: therefore this makes a good scope for a CPS.
* Since there were no identified flaws in these concepts or their presentation — and we understand a solving CIP to be on the way — please post whether you have any planned updates to this material or not.

If no further updates are planned, unless any objection from @Crypto2099 @Ryun1 @perturbing I can mark this `Last Check` for the next meeting (https://hackmd.io/@cip-editors/118).

Cerkoryn:
@rphair I made the changes to the directory name and rendered link as requested.

As far as further updates are concerned, we have nothing specific planned at this time.  We may try other methods to estimate how much lost stake there actually is, but due to the nature of the problem, any such efforts will require a large amount of guesswork no matter what.

I'm happy with keeping it at as it is for the foreseeable future.

rphair (approved):
thanks @Cerkoryn - on this last reading I don't see anything I would change.  Looking forward to seeing other editors' reviews and I'll mark this `Last Check` to make sure we get yea or nay on merging within this 2-week time frame.

CPS-0022/README.md
```diff
@@ -0,0 +1,159 @@
...
+
+## Use Cases
+
+The motivation for addressing lost stake and lost ADA is grounded in preserving fairness, efficiency, and the long-term health of Cardano’s proof-of-stake and governance mechanisms. At present, the problem may seem minor or largely theoretical, as lost ADA is not immediately visible on a small scale. But the impact compounds over time, and proactive understanding is crucial.
+
+**Figure 2** below illustrates a hypothetical projection of the growth of lost ADA if left unchecked:
+
+<img src="fig2.jpg" alt="Figure 2: Hypothetical exponential growth of lost ADA over time." width="100%" />
+
+> **Figure 2:** Hypothetical exponential growth of lost ADA over time, based on an assumed *conservative* 1.5% of ADA being lost per year (versus the 4% annual loss observed in Bitcoin), a 70% ratio of staked ADA, and a 3% annual staking-reward compounding. The blue curve (Lost ADA) shows the cumulative lost principal, while the red curve (Lost ADA + compounded rewards) shows how those lost funds are expected to grow with staking rewards, eventually representing a large fraction of total supply over decades.
```
lehins: It is unclear to me from this figure is why a static 3% APY is used in this projection, while rewards APY is projected to diminish over time well below that number.

Also, as mentioned earlier SPO retirement does remove such stake out of circulation.
This figure assume static SPOs for many decades to come, which isn't necessarily reflective of real life. If anything, stake pools are operated by real people who also have limited life spans. I think it would be worth at least mentioning the fact that some portion of the "lost stake" will naturally be removed or even better look at history how often stake pools retire with what proportion of total stake in order to attempt to do a more realistic projection in this figure.

Cerkoryn: @lehins that 3% APY was a compromise between the previous 8% yield from early the Shelley Era which has been compounding for longer, and also with the expectation that future things will be done to increase rewards such as Partner Chain AVS.  If the community does nothing to increase rewards, then the model has the estimate around 0.6% APY by 2030, which is likely unsustainable and another problem entirely.

SPO retirement does indeed remove stake from circulation, but an SPO with large amounts of immobile, lost stake is incentivized NOT to retire.  Even if their performance is poor or they are acting maliciously, by definition that stake will stay with them.  Additionally, with the model's projections it is likely that the problem will be quite bad within the lifetime of a single human who operates a stake pool.  And even then there's the possibility/likelihood of it being passed on.

As far as other methods of estimating or projecting lost stake, we also looked at...
- Residual stake left in pools after they retire
- Counting rewards (account) balances in wallets that have not had activity in X amount of time.
- Counting eligible wallets that don't claim their $NIGHT airdrop

The issue is that each of those things is a lot of extra, uncompensated work and at best they are only rough estimates as well.  We think it is more important to at least conclude that the problem exists even if we can't quantify the magnitude of it.

I do think our biggest saving grace for this problem is the existence of a saturation cap on each pool that sets on upper limit on lost rewards compounding infinitely.  But that is also very hard to estimate or model without a way to actually identify lost stake on chain.

CPS-0022/README.md
```diff
@@ -0,0 +1,159 @@
+---
+CPS: 22
+Title: Lost Stake
+Category: Ledger
+Status: Open
+Authors:
+    - Ryan Wiley <rian222@gmail.com>
+Proposed Solutions: []
+Discussions:
```
Ryun1, line 9:
```suggestion
Discussions:
    - https://github.com/cardano-foundation/CIPs/pull/1060
```

rphair: my apologies for also missing that one 😬

rphair (approved):
@Cerkoryn as you recall from last hour's meeting the editors that haven't submitted reviews yet will chime in... and if it appears we have consensus we might merge this well before the next meeting (will leave `Last Check` to cover all possibilities).

CPS-0022/README.md
```diff
@@ -0,0 +1,160 @@
...
+
+The lost-stake problem is the accumulation of this unreachable-yet-delegated ADA within the Cardano ecosystem. Such lost stake still contributes to stake-pool sizes and earns staking rewards every epoch, even though the rewards accumulate in an address that nobody controls. Over time, the lost ADA compounds. Rewards paid to these addresses increase their delegated stake, which in turn earns more rewards, and so on. Similarly, if the lost ADA was delegated to a governance representative (dRep) for Voltaire-era on-chain voting, that voting power remains with the dRep permanently (or until the dRep retires or is marked inactive). The original owner is no longer present to adjust their delegation in response to changing conditions. This creates a class of delegation that cannot be reallocated or corrected.
```
Ryun1:
```suggestion
The lost-stake problem is the accumulation of this unreachable-yet-delegated ADA within the Cardano ecosystem. Such lost stake still contributes to stake-pool sizes and earns staking rewards every epoch, even though the rewards accumulate in an address that nobody controls. For as long as the chosen stake pool produces staking rewards, the lost ADA compounds. Rewards paid to these addresses increase their delegated stake, which in turn earns more rewards, and so on. Similarly, if the lost ADA was delegated to a governance representative (dRep) for Voltaire-era on-chain voting, that voting power remains with the dRep permanently (or until the dRep retires or is marked inactive). The original owner is no longer present to adjust their delegation in response to changing conditions. This creates a class of delegation that cannot be reallocated or corrected.
```

A small point
lost stake only compounds if the chosen SPO continues to pay staking rewards

CPS-0022/README.md
```diff
@@ -0,0 +1,160 @@
...
+
+> **Figure 1:** Circles depict the Total Rewards Pot and its diminishing share with some rewards also flowing to Sticky Stake and Lost Stake addresses every epoch.
+
+Cardano already distributes a significant amount of ADA every epoch in staking rewards to addresses that are permanently inaccessible. This occurs when ADA is lost, such as when a holder loses their seed phrase or passes away without sharing their keys, rendering the funds permanently unreachable. In most other cryptocurrencies, lost coins simply exit circulation. For example, it is estimated that around 20% of all Bitcoin supply is lost forever [[investopedia.com](https://www.investopedia.com/tech/how-much-bitcoin-has-been-lost/)], with more granular analyses by Ledger Academy and Chainalysis both converging on roughly 4% of all Bitcoin being lost each year [[ledger.com](https://www.ledger.com/academy/topics/economics-and-regulation/how-many-bitcoin-are-lost-ledger); [chainalysis.com](https://www.chainalysis.com/blog/money-supply/)].
```
Ryun1: the investopedia.com and ledger.com links both don't seem to work

CPS-0022/README.md
```diff
@@ -0,0 +1,160 @@
...
+
+### Current Mitigations and Their Limits
+
+As noted earlier, governance proposals like [CIP-1694](https://cips.cardano.org/cips/cip1694/) include measures to limit the impact of inactive delegated stake on voting outcomes. These measures (such as marking inactive dReps) help prevent governance paralysis, but do not address the underlying issue of lost ADA still existing and, in some cases, continuing to accumulate rewards. However, when a stake pool with lost stake retires or shuts down, the lost ADA delegated to it is actually much less of a problem. That ADA effectively becomes undelegated and removed from circulation, meaning it no longer receives staking rewards or participates in governance. While the system currently has no direct way to reclaim or reassign lost ADA, the most persistent issues arise when lost stake remains delegated to active pools or dReps. Indirect mitigations only address symptoms (like governance quorum) rather than the root cause.
```
Ryun1:
```suggestion
As noted earlier, governance proposals like [CIP-1694](https://cips.cardano.org/cips/cip1694/) include measures to limit the impact of inactive delegated stake on voting outcomes. Additionally because DRep delegation is a new feature, all previously lost stake is unable to engage. These measures (such as marking inactive dReps) help prevent governance paralysis, but do not address the underlying issue of lost ADA still existing and, in some cases, continuing to accumulate rewards. However, when a stake pool with lost stake retires or shuts down, the lost ADA delegated to it is actually much less of a problem. That ADA effectively becomes undelegated and removed from circulation, meaning it no longer receives staking rewards or participates in governance. While the system currently has no direct way to reclaim or reassign lost ADA, the most persistent issues arise when lost stake remains delegated to active pools or dReps. Indirect mitigations only address symptoms (like governance quorum) rather than the root cause.
```

Ryun1: maybe worth noting that
because DRep delegation is new (and requires the use of private keys)
all previously lost stake is unable to delegate to DReps
which likely means that lost stake will have less of an impact on DRep voting power

CPS-0022/README.md
```diff
@@ -0,0 +1,160 @@
...
+
+### Why Ignoring the Problem Is Risky
+
+Some might argue that lost coins simply increase the value of the remaining ones (through scarcity) or that the effect is negligible for now. However, as the projections show, the effect is not static. It grows over time and can reach levels that materially impact network operation. Unlike in Bitcoin (where lost coins arguably don’t harm network security or functionality [[investopedia.com](https://www.investopedia.com/tech/how-much-bitcoin-has-been-lost/)]), in Cardano most lost coins still participate in consensus. Therefore, ignoring lost stake and lost ADA means accepting a slow-growing skew in the system that could ultimately undermine user trust and network performance. Early recognition allows for carefully researched, minimally disruptive solutions before the problem becomes too large and contentious to fix.
+
+---
```
Ryun1:
```suggestion
```

CPS-0022/README.md
```diff
@@ -0,0 +1,160 @@
...
+
+- **Economic Sustainability:** In the long term, as block-reward inflation tapers off, Cardano’s security will hinge on transaction fees and community participation. If a significant chunk of ADA is effectively out of economic circulation (yet still “consuming” rewards/fees), it means the active economy has to carry that burden. The security budget (total incentives for validators) would be partially drained to non-participants. This inefficiency will necessitate higher fees or protocol changes to compensate, which is undesirable for growth. In short, allowing lost stake and lost ADA to grow unchecked may undermine the sustainability of the network’s incentive model.
+
+
+## Open Questions 
+- How can the protocol reliably identify truly lost stake keys?
```
Ryun1:
```suggestion
- How can the protocol reliably identify truly lost credentials?
```
worth capturing all types of key / script

CPS-0022/README.md
```diff
@@ -0,0 +1,160 @@
...
+- What inactivity period (epochs/years) would be acceptable before stake is considered “lost”?
```
Ryun1:
```suggestion
- Could an inactivity period (epochs/years) be acceptable before stake is considered “lost”?
```
refocus the question to focus less on a single potential solution

CPS-0022/README.md
```diff
@@ -0,0 +1,160 @@
...
+- Which economic / social mechanisms can prevent reward dilution without violating property rights?
```
Ryun1, line 99:
```suggestion
- Which economic / social mechanisms can prevent reward dilution without violating property rights?
- Would the introduction of some solution to lost stake violate some of prior promises from the protocol?
```

when I stake my ada
the underlying promise from the protocol was that
as long as the pool kept paying staking rewards
my stake credential would earn the rewards
without me having to actively engage

I feel that any solution to this problem statement
would result in this original promise being broken

Cerkoryn: What was the exact promise though? Is there somewhere we can reference it so that we can work around it?

If it's just an implicit promise for anybody who signed a delegation transaction, then where do we draw the line? Is it a promise that the protocol would never change other than the exact things that were declared on the roadmap?

If so, then IMHO we've already broken it in worse ways.  Consider the hard requirement to select a dRep to withdraw your staking rewards.  The difference between this and that:

"You no longer earn any new staking rewards unless you do X"

vs.

"You can longer access staking rewards that are ALREADY YOURS unless you do X"

CPS-0022/README.md
Cerkoryn: The Ledger.com link worked fine for me, but the Investopedia link was broken.  I just updated the link and pushed the commit.

CIP-0163/README.md
```diff
@@ -0,0 +1,284 @@
...
+- **Ineligible for rewards.** For epochs in which a reward account is inactive, its epoch reward share is **not credited** to that account. Those uncredited rewards are instead distributed with the rest of the rewards pot (see next section).
+- **Migration.** On activation, determine the epoch that the last witness occurred for each rewards account.  Then initialize each account's `expirationEpoch` to that epoch plus `delegatorInactivity`.
+
+### 2) Full‑pot reward distribution
+
+- In each epoch, distribute the **full rewards pot** (monetary expansion + fees, less the treasury cut, blocks not produced [Eta], and unmet pledge) to eligible delegators and pools without returning any remainder to reserves **(i.e., eliminate the residual ∆r₂ described in SL‑D5 §11.10 and distribute that proportion as well)**.
```
perturbing: What is the reason to change this? As I always understood it, the failure to produce blocks means the rewards for that missed slot cannot be claimed by anyone, so they are put into the treasury.

That is, the motivation of the CIP does not warrant this change?

CIP-0163/README.md
```diff
@@ -0,0 +1,284 @@
...
+
+**Note:** These calculations do not take into account any collected transaction fees, pool performance, or pool fees.
+
+These effects are also modeled in the open source tool available at [https://spo-incentives.vercel.app/](https://spo-incentives.vercel.app/). Setting the rewards formula radio button to “Current” represents the current rewards calculation behavior. Setting it to “Full” represents the rewards calculation after applying this CIP. It is also recommended to experiment with the “Staked Ratio” slider under both of those settings to see how it will affect rewards.
+
+- **Runaway lost‑stake mitigation efficacy**
+
+  - Before:
+
+    ![Runaway lost-stake before mitigation (Fig. 2)](Fig2.png)
```
perturbing:
```suggestion
    ![Runaway lost-stake before mitigation (Fig. 2)](fig2.png)
```

CIP-0163/README.md
```diff
@@ -0,0 +1,284 @@
...
+
+  - After:
+
+    ![Runaway lost-stake after mitigation (Fig. 3)](Fig3.png)
```
perturbing:
```suggestion
    ![Runaway lost-stake after mitigation (Fig. 3)](fig3.png)
```

CIP-0163/README.md
```diff
@@ -0,0 +1,284 @@
...
+This dynamic couples network security to real participation and reduces the risk of active-stake dropping too low.
+
+### Evidence and modeling
+
+- **Reserves trajectory** will be closer to true Rho after this CIP is implemented
+  - Red = ADA in reserves if we followed Rho exactly
+  - Yellow = ADA in reserves account for actual blocks produced (Eta) and missing pledge. This is the track the reserves is expected to follow after the inclusion of this CIP.
+  - Green = Actual amount of ADA in reserves
+
+  ![Reserve Depletion)](Fig1.png)
```
perturbing:
```suggestion
  ![Reserve Depletion)](fig1.png)
```

CPS-0022/fig1.jpg
perturbing, line 1: The sizes of these circles are a bit misleading (I know that you cannot say exactly how big they are relative to each other).

CIP-0163/README.md
rphair: (committed typographical fix to keep the review field clear in case a `Last Check` is made... though review still seems active here)

CIP-0163/README.md
rphair: (same as https://github.com/cardano-foundation/CIPs/pull/1060#discussion_r2351795450)

CIP-0163/README.md
rphair: (same as https://github.com/cardano-foundation/CIPs/pull/1060#discussion_r2351795450)

CIP-0163/README.md
Cerkoryn: Currently, any rewards that are allocated during the rewards calculation but not distributed, are returned to the *reserve*, not the treasury.

The intent of CIP-163 is to make staking more dynamic by rewarding activity accordingly and curtailing rewards for inactive accounts.  The section you quoted works synergistically with the first section, `1) Reward-account inactivity expiry (proof-of-life)`.  The reasons for how it fulfills the motivation for the CIP are listed under the rationale section.

CPS-0022/README.md
Ryun1: gotcha, thanks

CPS-0022/README.md
Ryun1: > If it's just an implicit promise for anybody who signed a delegation transaction, then where do we draw the line?

100% you're right
it was/is a implicit promise

v good examples to the contrary

rphair:
@Cerkoryn as observed at the CIP meeting today we are leaving this in `Last Check` to allow all other editors enough time for further review and follow-up on existing review points.  We might merge this earlier than the next meeting (https://hackmd.io/@cip-editors/120) if we have consensus earlier here on GitHub.

rphair:
@Cerkoryn as you observed at the CIP meeting today (# 120) we had to postpone our `Last Check`s until next meeting (https://hackmd.io/@cip-editors/121) due to not having editor quorum at meeting (only 1 editor present) and the massive turn-out for emergency Leios review.

@Ryun1 @Crypto2099 @perturbing for these 2 proposals I'd offer to merge this in the in-between period if you can provide approval and if there are now objections... otherwise it's on to `Last Check` as planned in 2 weeks' time.

CIP-0163/README.md
```diff
@@ -0,0 +1,284 @@
+---
```
Ryun1: is this file intended to be included within this pull request?
I think it may be a mistake

Cerkoryn:
@Ryun1 @rphair @Crypto2099 

I removed the chart with the projections and any references to the speculative estimates throughout the document.  There were a couple of places where I re-phrased the wording to be clearer without relying on any of those numbers.  I think that it reads well now, so I didn't feel the need to add another graphic to replace the one that was there.  However, if any of you think that something needs to be added I'm happy to try and incorporate that to communicate the problem statement more clearly.

I also squashed the commits that had accidentally included CIP-0163 as well.  The git history should be clean now.

Please let me know what you think or if you require any further changes.

rphair (commented):
thanks @Cerkoryn - the updates certainly seem to me to support a more general & inclusive view of the problem.  I am now in favour of merging this without reservation & will check to see if further confirmation can be expedited so we don't have to wait another 2 weeks to merge this.

**edit, p.s.** before we do that, I assume the "offending" file `CPS-0022/fig2.jpg` was also intended for removal along with all the references to it.  If your intention was to leave it in then please indicate what purpose it serves: otherwise we'll wait for that file to be `git rm`'d in your fork before merging.

**p.p.s.**
> I also squashed the commits that had accidentally included CIP-0163 as well.

FYI we have the repository set up so that commits are squashed when PRs enter the repo... so committing the unmodified file into your branch will achieve the same effect, sometimes more simply (as long as the round trip didn't include any otherwise inapplicable co-authorship). 🤓

Cerkoryn:
My apologies, I had overlooked the image file. That is now removed as well.

PR #1066, opened 2025-07-25
Fix Typo in CIP-149

colll78:


Crypto2099:
Not sure what is going on with this PR and how we resolve the conflicts to be able to review and merge it

colll78:
whoops, I forgot to rebase. Okay now it is good to merge.

colll78:
or if you can commit directly just throw away this whole PR and commit directly to change that one line:
>Where the integer stored in the metadata is the thousandths place in the decimal so 1 would convey that the user is electing to donate 0.01% of their staking rewards to their DRep. This value should be obtained directly from the percentage the user described in the prompt (ie. the user put 0.5% in the prompt then the integer in the metadata would be 5).

To:
>Where the integer stored in the metadata is the thousandths place in the decimal so 1 would convey that the user is electing to donate 0.1% of their staking rewards to their DRep. This value should be obtained directly from the percentage the user described in the prompt (ie. the user put 0.5% in the prompt then the integer in the metadata would be 5).

rphair:
@colll78 for confirmation, please correct if wrong.: a single (erroneous) `0` is being removed from the `0.01%` term (which would be a "_ten_ thousandths" place): and no other changes.  I agree it's important enough to push out ASAP.

If incorrect please let me know.  There are too many commits on this branch for them to many any sense for correcting a single-character typo.  Maybe from your end that's also true for creating a new branch: but here in a public repository we need these PRs to be clean, especially for such simple changes.

rphair:
Resubmitted as https://github.com/cardano-foundation/CIPs/pull/1067.

PR #1063, opened 2025-07-25
Introduce GuardScripts CIP

colll78:
Introduces GuardScriptCredential, a new credential type that allows smart contracts to validate incoming UTxOs, not just spent ones.

This closes one of the most dangerous gaps in Cardano's smart contract model: today, anyone can send arbitrary UTxOs with spoofed datums to any script address — and the script can't stop it.

Guard scripts eliminate this vulnerability, removing the need for auth tokens, thread tokens, registry hacks, and complex off-chain filtering.

It’s a simple, native fix to a systemic security and UX problem. Huge reduction in attack surface and dApp complexity. 

Let's make Cardano best-in-class for DeFi applications.

PR #1072, opened 2025-08-06
CIP-???? | More Descriptive Script Purposes

michele-nuzzi:
This CIP proposes better script purposes to optimize for common operations for each purpose.

In particoular, it should standardize how to derive the hash of the script for all puropses, and should simplify the offchain by passing the index of ther relevant data in the `TxInfo` passed as context.

([Rendered proposal in branch](https://github.com/HarmonicLabs/CIP-better-script-purposes/blob/master/CIP-%3F%3F%3F%3F/README.md))

### DIFF
--- a/CIP-????/README.md
+++ b/CIP-????/README.md
@@ -7,7 +7,7 @@ Authors:
     - Michele Nuzzi <michele@harmoniclabs.tech>
 Implementors: []
 Discussions:
-    - https://github.com/cardano-foundation/CIPs/pull/?
+    - https://github.com/cardano-foundation/CIPs/pull/1072
 Created: 2025-08-06
 License: CC-BY-4.0
 ---
@@ -43,7 +43,8 @@ data ScriptInfo
       V2.ScriptHash
       -- | 0-based index of the input being spent in `txInfoInputs`
       Haskell.Integer
-      (Haskell.Maybe V2.Datum)
+      -- removed optional datum, since its presence (or not) is determined by the resolved input
+      -- (Haskell.Maybe V2.Datum)
       -- resolved input being spent
       V3.TxOut
       V3.TxOutRef
--- a/CIP-????/README.md
+++ b/CIP-????/README.md
@@ -1,0 +1,155 @@
+---
+CIP: ?
+Title: Better script purposes
+Category: Plutus
+Status: Proposed
+Authors:
+    - Michele Nuzzi <michele@harmoniclabs.tech>
+Implementors: []
+Discussions:
+    - https://github.com/cardano-foundation/CIPs/pull/?
+Created: 2025-08-06
+License: CC-BY-4.0
+---
+
+## Abstract
+<!-- A short (\~200 word) description of the proposed solution and the technical issue being addressed. -->
+This CIP proposes better script purposes to optimize for common operations for each purpose
+
+## Motivation: why is this CIP necessary?
+<!-- A clear explanation that introduces the reason for a proposal, its use cases and stakeholders. If the CIP changes an established design then it must outline design issues that motivate a rework. For complex proposals, authors must write a Cardano Problem Statement (CPS) as defined in CIP-9999 and link to it as the `Motivation`. -->
+
+Common operations such as deriving the hash of the script being executed, or finding the relevant data for validation could be optimized if the trusted data provided to the script was more complete.
+
+The same result could be achieved by including this data in the redeemer, however these alternative solutions ofter imply greater complexity in the offchain code, and do not come without overhead on-chian, since the redeemer cannot be trusted.
+
+This CIP would solve those problems, with great improvements on the development experience as well as the resulting onchain scripts complexity and efficiency.
+
+
+## Specification
+<!-- The technical specification should describe the proposed improvement in sufficient technical detail. In particular, it should provide enough information that an implementation can be performed solely on the basis of the design in the CIP. This is necessary to facilitate multiple, interoperable implementations. This must include how the CIP should be versioned, if not covered under an optional Versioning main heading. If a proposal defines structure of on-chain data it must include a CDDL schema in its specification.-->
+
+this CIP proposes to modify the definition of `ScriptInfo` and `ScriptPurpose` in [`plutus-ledger-api`](https://github.com/IntersectMBO/plutus/blob/618480658ec1750179c2832c6b619bc2d872b225/plutus-ledger-api/src/PlutusLedgerApi/V3/Contexts.hs#L428-L461) as follows
+
+```hs
+data ScriptInfo
+  = MintingScript
+      -- hash of the script being executed (same as polciy)
+      V2.CurrencySymbol
+      -- | 0-based index of the given `CurrencySymbol` in `txInfoMint`
+      Haskell.Integer
+  | SpendingScript
+      -- hash of the script being executed
+      V2.ScriptHash
+      -- | 0-based index of the input being spent in `txInfoInputs`
+      Haskell.Integer
+      (Haskell.Maybe V2.Datum)
+      -- resolved input being spent
+      V3.TxOut
+      V3.TxOutRef
+  | RewardingScript
+      -- V2.Credential -- removed, since implictily it must be a script credential
+      -- hash of the script being executed
+      V2.ScriptHash
+      -- | 0-based index of the withdrawal being verified in `txInfoWdrl`
+      Haskell.Integer
+  | CertifyingScript
+      -- hash of the script being executed
+      V2.ScriptHash
+      -- | 0-based index of the given `TxCert` in `txInfoTxCerts`
+      Haskell.Integer
+      TxCert
+  | VotingScript
+      -- hash of the script being executed
+      V2.ScriptHash
+      -- | 0-based index of the given vote in `txInfoVotes`
+      Haskell.Integer
+      Voter
+  | ProposingScript
+      -- hash of the script being executed
+      V2.ScriptHash
+      -- | 0-based index of the given `ProposalProcedure` in `txInfoProposalProcedures`
+      Haskell.Integer
+      ProposalProcedure
+  deriving stock (Generic, Haskell.Show, Haskell.Eq)
+  deriving anyclass (HasBlueprintDefinition)
+  deriving (Pretty) via (PrettyShow ScriptInfo)
+```
+
+```hs
+data ScriptPurpose
+  = MintingScript
+      -- hash of the script being executed (same as polciy)
+      V2.CurrencySymbol
+      -- | 0-based index of the given `CurrencySymbol` in `txInfoMint`
+      Haskell.Integer
+  | SpendingScript
+      -- hash of the script being executed
+      V2.ScriptHash
+      -- | 0-based index of the input being spent in `txInfoInputs`
+      Haskell.Integer
+      V3.TxOutRef
+  | RewardingScript
+      -- V2.Credential -- removed, since implictily it must be a script credential
+      -- hash of the script being executed
+      V2.ScriptHash
+      -- | 0-based index of the withdrawal being verified in `txInfoWdrl`
+      Haskell.Integer
+  | CertifyingScript
+      -- hash of the script being executed
+      V2.ScriptHash
+      -- | 0-based index of the given `TxCert` in `txInfoTxCerts`
+      Haskell.Integer
+      TxCert
+  | VotingScript
+      -- hash of the script being executed
+      V2.ScriptHash
+      -- | 0-based index of the given vote in `txInfoVotes`
+      Haskell.Integer
+      Voter
+  | ProposingScript
+      -- hash of the script being executed
+      V2.ScriptHash
+      -- | 0-based index of the given `ProposalProcedure` in `txInfoProposalProcedures`
+      Haskell.Integer
+      ProposalProcedure
+  deriving stock (Generic, Haskell.Show, Haskell.Eq)
+  deriving anyclass (HasBlueprintDefinition)
+  deriving (Pretty) via (PrettyShow ScriptInfo)
+```
+
+## Rationale: how does this CIP achieve its goals?
+<!-- The rationale fleshes out the specification by describing what motivated the design and what led to particular design decisions. It should describe alternate designs considered and related work. The rationale should provide evidence of consensus within the community and discuss significant objections or concerns raised during the discussion.
+
+It must also explain how the proposal affects the backward compatibility of existing solutions when applicable. If the proposal responds to a CPS, the 'Rationale' section should explain how it addresses the CPS, and answer any questions that the CPS poses for potential solutions.
+-->
+
+This modification would be reflected in the underlying data representation as always having the script hash of the given script as the first field of the purpose, and the index relevant to the purpose as second element.
+
+This allows to optimize for the very common operation of deriving the hash of the script itself onchain, greatly reducing onchain complexity and cost of execution.
+
+The change also propses to add the resovled input in the case of the very common spending purpose, where the majority of the relevant informations is present.
+
+## Path to Active
+
+Changes are reflected on the plutus-ledger-api
+
+### Acceptance Criteria
+<!-- Describes what are the acceptance criteria whereby a proposal becomes 'Active' -->
+
+The change is included in an hardfork
+
+### Implementation Plan
+<!-- A plan to meet those criteria or `N/A` if an implementation plan is not applicable. -->
+N/A
+
+<!-- OPTIONAL SECTIONS: see CIP-0001 > Document > Structure table -->
+
+## Copyright
+<!-- The CIP must be explicitly licensed under acceptable copyright terms. Uncomment the license you wish to use (delete the other one) and ensure it matches the License field in the header.
+
+If AI/LLMs were used in the creation of the copyright text, the author may choose to include a disclaimer to describe their application within the proposal.
+-->
+
+This CIP is licensed under [CC-BY-4.0](https://creativecommons.org/licenses/by/4.0/legalcode)
+<!-- This CIP is licensed under [Apache-2.0](http://www.apache.org/licenses/LICENSE-2.0). -->


michele-nuzzi:
On second thought, we could remove `Maybe Datum` if the resolved input is in there

michele-nuzzi:
ping @lehins @WhatisRT (Ledger)
ping @MicroProofs @fallen-icarus

Please tag anyone else who might be interested

CIP-????/README.md
```diff
@@ -0,0 +1,156 @@
...
+  = MintingScript
+      -- hash of the script being executed (same as polciy)
+      V2.CurrencySymbol
+      -- | 0-based index of the given `CurrencySymbol` in `txInfoMint`
+      Haskell.Integer
+  | SpendingScript
+      -- hash of the script being executed
+      V2.ScriptHash
+      -- | 0-based index of the input being spent in `txInfoInputs`
+      Haskell.Integer
```
lehins, line 45: This index cannot be added until [CIP-0128 - Preserving Order of Transaction Inputs](https://github.com/cardano-foundation/CIPs/pull/758) is implemented. I think it is important to point out this prerequisite in this CIP.

CIP-????/README.md
```diff
@@ -0,0 +1,156 @@
...
+
+this CIP proposes to modify the definition of `ScriptInfo` and `ScriptPurpose` in [`plutus-ledger-api`](https://github.com/IntersectMBO/plutus/blob/618480658ec1750179c2832c6b619bc2d872b225/plutus-ledger-api/src/PlutusLedgerApi/V3/Contexts.hs#L428-L461) as follows
+
+```hs
+data ScriptInfo
+  = MintingScript
+      -- hash of the script being executed (same as polciy)
+      V2.CurrencySymbol
+      -- | 0-based index of the given `CurrencySymbol` in `txInfoMint`
+      Haskell.Integer
```
lehins, line 40: This index can't be added, because the order in which multi-assets are listed in the serialized transaction will not necessarily match the order in which they will appear in plutus context. We would need a similar CIP to CIP-128 for minting before we can go this route.

CIP-????/README.md
```diff
@@ -0,0 +1,156 @@
...
+      -- hash of the script being executed
+      V2.ScriptHash
+      -- | 0-based index of the given `TxCert` in `txInfoTxCerts`
+      Haskell.Integer
+      TxCert
+  | VotingScript
+      -- hash of the script being executed
+      V2.ScriptHash
+      -- | 0-based index of the given vote in `txInfoVotes`
+      Haskell.Integer
```
lehins, line 67: Same issue as with minting. Order of votes is Haskell specific and the only way we would be able to report this index is if we preserve the order in which votes where submitted on the wire.
Note, that I am not opposing this, I am just pointing out an issue that would need to be resolved if we want to supply indices for maps.

CIP-????/README.md
```diff
@@ -0,0 +1,156 @@
...
+      -- (Haskell.Maybe V2.Datum)
+      -- resolved input being spent
+      V3.TxOut
+      V3.TxOutRef
+  | RewardingScript
+      -- V2.Credential -- removed, since implictily it must be a script credential
+      -- hash of the script being executed
+      V2.ScriptHash
+      -- | 0-based index of the withdrawal being verified in `txInfoWdrl`
+      Haskell.Integer
```
lehins, line 56: Same problem as for minting

CIP-????/README.md
```diff
@@ -0,0 +1,156 @@
...
+      -- | 0-based index of the given `ProposalProcedure` in `txInfoProposalProcedures`
+      Haskell.Integer
+      ProposalProcedure
+  deriving stock (Generic, Haskell.Show, Haskell.Eq)
+  deriving anyclass (HasBlueprintDefinition)
+  deriving (Pretty) via (PrettyShow ScriptInfo)
+```
+
+```hs
+data ScriptPurpose
```
lehins, line 81: I have the same comments for indices in the script purpose as I made for ScriptInfo

CIP-????/README.md
```diff
@@ -0,0 +1,156 @@
...
+<!-- The rationale fleshes out the specification by describing what motivated the design and what led to particular design decisions. It should describe alternate designs considered and related work. The rationale should provide evidence of consensus within the community and discuss significant objections or concerns raised during the discussion.
+
+It must also explain how the proposal affects the backward compatibility of existing solutions when applicable. If the proposal responds to a CPS, the 'Rationale' section should explain how it addresses the CPS, and answer any questions that the CPS poses for potential solutions.
+-->
+
+This modification would be reflected in the underlying data representation as always having the script hash of the given script as the first field of the purpose, and the index relevant to the purpose as second element.
+
+This allows to optimize for the very common operation of deriving the hash of the script itself onchain, greatly reducing onchain complexity and cost of execution.
+
+The change also propses to add the resovled input in the case of the very common spending purpose, where the majority of the relevant informations is present.
```
lehins, line 132: We've discussed this exact change in the latest Ledger Working Group in the context of [protected addresses CIP](https://github.com/cardano-foundation/CIPs/pull/1063).  Seems there is large demand for it.

CIP-????/README.md
```diff
@@ -0,0 +1,156 @@
...
+## Specification
+<!-- The technical specification should describe the proposed improvement in sufficient technical detail. In particular, it should provide enough information that an implementation can be performed solely on the basis of the design in the CIP. This is necessary to facilitate multiple, interoperable implementations. This must include how the CIP should be versioned, if not covered under an optional Versioning main heading. If a proposal defines structure of on-chain data it must include a CDDL schema in its specification.-->
+
+this CIP proposes to modify the definition of `ScriptInfo` and `ScriptPurpose` in [`plutus-ledger-api`](https://github.com/IntersectMBO/plutus/blob/618480658ec1750179c2832c6b619bc2d872b225/plutus-ledger-api/src/PlutusLedgerApi/V3/Contexts.hs#L428-L461) as follows
+
+```hs
+data ScriptInfo
+  = MintingScript
+      -- hash of the script being executed (same as polciy)
+      V2.CurrencySymbol
```
lehins, line 38: Why not `ScriptHash` for consistency sake? 
It is always possible to go from ScriptHash to CurrencySymbol
```suggestion
      V2.ScriptHash
```

CIP-????/README.md
```diff
@@ -0,0 +1,156 @@
...
+      -- | 0-based index of the given `CurrencySymbol` in `txInfoMint`
+      Haskell.Integer
+  | SpendingScript
+      -- hash of the script being executed
+      V2.ScriptHash
+      -- | 0-based index of the input being spent in `txInfoInputs`
+      Haskell.Integer
+      V3.TxOutRef
```
lehins, line 92: We could also add the resolved `TxOut` here as well. I think it would be useful for a script to know other outputs that are being spent in this transaction
```suggestion
      V3.TxOut
      V3.TxOutRef
```
With this suggestion, however, we are loosing any distinction between `ScriptInfo` and `ScriptPurpose`, so it would probably make sense then to remove `ScritpInfo` in favor of `ScritpPurpose`, like it was before.

lehins (commented):
For the most part this affects DApp developers and is non-issue for Ledger.
There was only one real concern that I have, which has to do with indices for minting, withdrawing and voting. If we are to change the semantics of ordering and preserve the order instead of what we do today, then it would be a bigger change for ledger and it would affect DApp developers quite a bit, because they would no longer get them (mints, withdrawals and votes) in a sorted order.

I don't really know exactly how this CIP would affect DApp developers, so I can't comment on that. I might be wrong, but from what I know deserialization of plutus context is a big part of execution cost. So, if I am right, then we should consider this kind of changes carefully. Although, there might have been new developments on this front, so there is a chance that this is no longer a problem. I'll let @zliu41 comment on that.

CIP-????/README.md
```diff
@@ -0,0 +1,156 @@
+---
+CIP: ?
```
```suggestion
CIP: "?"
```
We've found out by trial & error that quoting this character prevents the YAML error on GitHub preview `mapping keys are not allowed in this context at line 1 column 6` but haven't gotten this into the documentation & templates yet.

CIP-????/README.md
```diff
@@ -0,0 +1,156 @@
...
+Title: Better script purposes
```
rphair, line 3: If there is one dissenting comment about these script purposes not being unilaterally "better" I would propose one of the following:
* Updated Script Purposes (if reviewers are indifferent)
* Optimised Script Purposes (if reviewers are enthusiastic... and if this title is in fact accurate)

At this point that should be measurable by those already tagged (@lehins @WhatisRT @MicroProofs @fallen-icarus).

CIP-????/README.md
```diff
@@ -0,0 +1,156 @@
...
+    - Michele Nuzzi <michele@harmoniclabs.tech>
+Implementors: []
+Discussions:
+    - https://github.com/cardano-foundation/CIPs/pull/1072
+Created: 2025-08-06
+License: CC-BY-4.0
+---
+
+## Abstract
+<!-- A short (\~200 word) description of the proposed solution and the technical issue being addressed. -->
```
rphair, line 16:
```suggestion
<!-- A short (\~200 word) description of the proposed solution and the technical issue being addressed. -->
```
Please remove template comment scaffolding here & all the way down.

CIP-????/README.md
```diff
@@ -0,0 +1,156 @@
...
+
+This CIP would solve those problems, with great improvements on the development experience as well as the resulting onchain scripts complexity and efficiency.
+
+
+## Specification
+<!-- The technical specification should describe the proposed improvement in sufficient technical detail. In particular, it should provide enough information that an implementation can be performed solely on the basis of the design in the CIP. This is necessary to facilitate multiple, interoperable implementations. This must include how the CIP should be versioned, if not covered under an optional Versioning main heading. If a proposal defines structure of on-chain data it must include a CDDL schema in its specification.-->
+
+this CIP proposes to modify the definition of `ScriptInfo` and `ScriptPurpose` in [`plutus-ledger-api`](https://github.com/IntersectMBO/plutus/blob/618480658ec1750179c2832c6b619bc2d872b225/plutus-ledger-api/src/PlutusLedgerApi/V3/Contexts.hs#L428-L461) as follows
```
rphair, line 32:
```suggestion
This CIP proposes to modify the definition of `ScriptInfo` and `ScriptPurpose` in [`plutus-ledger-api`](https://github.com/IntersectMBO/plutus/blob/618480658ec1750179c2832c6b619bc2d872b225/plutus-ledger-api/src/PlutusLedgerApi/V3/Contexts.hs#L428-L461) as follows:
```
(small grammar errors at beginning & end)

CIP-????/README.md
```diff
@@ -0,0 +1,156 @@
...
+The change also propses to add the resovled input in the case of the very common spending purpose, where the majority of the relevant informations is present.
+
+## Path to Active
+
+Changes are reflected on the plutus-ledger-api
+
+### Acceptance Criteria
+<!-- Describes what are the acceptance criteria whereby a proposal becomes 'Active' -->
+
+The change is included in an hardfork
```
rphair, line 141:
```suggestion
The change is included in a hardfork.
```
This could use some elaboration but will wait until the Ledger responses are all in & Plutus reps have reviewed with respect to when this might be included (cc @zliu41).

rphair (commented):
Thanks again @michele-nuzzi & have tagged `Triage` for the next CIP meeting: https://hackmd.io/@cip-editors/117

michele-nuzzi:
@lehins 

>  the order in which multi-assets are listed in the serialized transaction will not necessarily match the order in which they will appear in plutus context. 

the same issue is already present in the current input ordering.

since we only care about the plutus ordering, would it be possible to derive the index from the tx info plutus data?

any tx builder that supports redeemer indexing already does it this way.

CIP-????/README.md
zliu41: The title should be more concrete. Better in what sense? Suggestion: adding script hash and indices in ScriptInfo and ScriptPurpose

CIP-????/README.md
zliu41: > the order in which multi-assets are listed in the serialized transaction will not necessarily match the order in which they will appear in plutus context

I'm not sure why this is a problem? This is just for indexing into `txInfoMint` - why does the order matter?

CIP-????/README.md
```diff
@@ -0,0 +1,156 @@
...
+
+## Rationale: how does this CIP achieve its goals?
+<!-- The rationale fleshes out the specification by describing what motivated the design and what led to particular design decisions. It should describe alternate designs considered and related work. The rationale should provide evidence of consensus within the community and discuss significant objections or concerns raised during the discussion.
+
+It must also explain how the proposal affects the backward compatibility of existing solutions when applicable. If the proposal responds to a CPS, the 'Rationale' section should explain how it addresses the CPS, and answer any questions that the CPS poses for potential solutions.
+-->
+
+This modification would be reflected in the underlying data representation as always having the script hash of the given script as the first field of the purpose, and the index relevant to the purpose as second element.
+
+This allows to optimize for the very common operation of deriving the hash of the script itself onchain, greatly reducing onchain complexity and cost of execution.
```
zliu41, line 130: The hash should be factored out of ScriptInfo and ScriptPurpose. Then you don't need a special optimization to get the hash.

CIP-????/README.md
```diff
@@ -0,0 +1,156 @@
...
+---
+
+## Abstract
+<!-- A short (\~200 word) description of the proposed solution and the technical issue being addressed. -->
+This CIP proposes better script purposes to optimize for common operations for each purpose
+
+## Motivation: why is this CIP necessary?
+<!-- A clear explanation that introduces the reason for a proposal, its use cases and stakeholders. If the CIP changes an established design then it must outline design issues that motivate a rework. For complex proposals, authors must write a Cardano Problem Statement (CPS) as defined in CIP-9999 and link to it as the `Motivation`. -->
+
+Common operations such as deriving the hash of the script being executed, or finding the relevant data for validation could be optimized if the trusted data provided to the script was more complete.
```
zliu41, line 22: For completeness it's good to briefly explain why obtaining the hash of the script being executed is a common and useful operation.

CIP-????/README.md
zliu41: This requires a new era and new ledger language version, not just a HF

zliu41 (commented):
Looks OK, except that the hash should be factored out of ScriptInfo/Purpose. Also, this is not really a change in plutus-ledger-api. It is the new Plutus V4 script purposes, which currently doesn't exist - I think the CIP should make this clear.

CIP-????/README.md
rphair: thanks @zliu41 - will highlight for discussion the title issue & your comment and brainstorm a "better" title at the next CIP meeting (https://hackmd.io/@cip-editors/117) where this is up for `Triage`.  @michele-nuzzi you are, as always, welcome to come & introduce the proposal and discuss this & other introductory questions.

fallen-icarus:
```haskell
| SpendingScript
      -- hash of the script being executed
      V2.ScriptHash
      -- | 0-based index of the input being spent in `txInfoInputs`
      Haskell.Integer
      -- removed optional datum, since its presence (or not) is determined by the resolved input
      -- (Haskell.Maybe V2.Datum)
      -- resolved input being spent
      V3.TxOut
      V3.TxOutRef
```
      
What is the point of having an index *and* the `TxOut` passed? I thought the whole point of the index was to get the `TxOut` from the input list. The same question applies to all the other purposes that have both the index and the actual item (e.g., `TxCert` under `CertifyingScript`).

Unisay:
The way this CIP is written makes it hard to understand for me:

-  "Better purposes" - I don't think there exists a universally agreed-upon notion of what "Better" is. I'd much prefer "Script Info and Purpose data structures extended with additional (trusted) attributes"
-  The new data structures aren't proposed in a way that makes it easy to see what's added, what is removed and what is left intact (Before/After, Diffs, etc.).
- I would be great to explicitly describe the use-cases that benefit from the proposed additions most of all.
- Are there any trade-offs? Are there only positive consequences or is there something that becomes harder/worse?

michele-nuzzi:
> What is the point of having an index _and_ the `TxOut` passed

the point is integrity.

keeping the same data types in the same fields allows field extraction without matching first.

regardless, the index could turn useful while filtering or checking relations with other elements of the list

colll78:
> I don't really know exactly how this CIP would affect DApp developers, so I can't comment on that. I might be wrong, but from what I know deserialization of plutus context is a big part of execution cost. So, if I am right, then we should consider this kind of changes carefully. Although, there might have been new developments on this front, so there is a chance that this is no longer a problem. I'll let @zliu41 comment on that.

If you mean deserialization within plutus execution (ie. Extracting fields out of BuiltinData) then this is not significantly expensive, and better yet, it's becoming leagues more efficient with the introduction of casing on BuiltinData and other improvements to Plutus. The reason historically this was a huge burden on ex-units was the use of full SOP / Scott conversion of types. If you operate directly on BuiltinData and only convent to SOP when you need, and only with the specific fields you need, then you won't experience these bottlenecks. PlutusLedgerApi now provides AsData variants for everything, including the ScriptContext. If you use the AsData variant, these additional fields will have no impact on execution budget unless you choose to use them, and even if you do the impact would be quite small and even smaller once we get these improvements (ie builtin array, casing on BuiltinData, and so forth).

colll78:
> Looks OK, except that the hash should be factored out of ScriptInfo/Purpose. Also, this is not really a change in plutus-ledger-api. It is the new Plutus V4 script purposes, which currently doesn't exist - I think the CIP should make this clear.

I agree with @zliu41 100% here, the script hash should be moved out of the script purpose directly into a field in the ScriptContext. So we would have: txInfo, redeemer, scriptPurpose and ownScriptHash as the four fields of the ScriptContext.

michele-nuzzi:
yes makes sense to move out the script hash

CIP-????/README.md
lehins: Because this index can be different from the index that was actually supplied in the redeemer in the serialized transaction. And that is dangerous in my opinion, since many people do not know that.

lehins:
> https://github.com/cardano-foundation/CIPs/pull/1072#issuecomment-3188726219

@colll78 That is really good to hear. Cause I do remember this being a problem in the past.

CIP-????/README.md
colll78: I don't think this is an unacceptable concern. The smart contract itself will still only succeed if whatever token indexed in the context matches what the contract expects, this would only impact offchain tooling which is for some reason relying on these indices for analytic purposes or similar non-value transfer related functionality.

I agree it would be better if they matched, but if ledger has other priorities to work on, or if this change to the ledger would take significant amount of time then I don't see any issues with introducing this now and then when you free up change the ledger in a subsequent hf to assure they indices provided in the serialized tx match the ones in the context.

lehins: > I don't think this is an unacceptable concern

It took me a moment to unwrap the double negation :smile: 

Yes, as far as the script execution is concerned it should be fine, unless the index in the redeemer pointer is also passed in the argument to the script, which wouldn't guarantee to match the index in the script info and the script purpose.

@colll78 and @zliu41 I'll then trust your judgement on this one and won't worry about.
We'll treat this as a separate concern that is only relevant to the Ledger

WhatisRT:
I really like the proposal to add the script's own hash into the context. It has come up many times that something is difficult to do because the script doesn't know it's own hash, and there's no reason it needs to be like that. Otherwise I don't have much to add other than I agree with the proposed changes as to how the data is passed exactly.

CIP-????/README.md
colll78: Adding `TxOut` would be great if we can make sure that in the ledger code that generates the script context we don't do any redundant work (right now, where the same data exists in the script context in multiple places, we do redundant work to get and `toPlutusType` it each time).

Otherwise, since the proposed `SpendingScript` includes the index of that input, and Plutus is introducing `BuiltinArray` which will allow O(1) access of elements by index, this isn't a huge concern, because getting this `TxOut` will have a trivial cost. (Right now getting the resolved input from the information provided in `SpendingScript` / `SpendingPurpose` is extremely costly, it requires O(n) decoding of BuiltinData (each tx input), O(n) head operations (getting first field of the decoded input, which is `TxOutRef`), O(n) equality comparisons for identifying the correct `TxOutRef`.

With `BuiltinArray` all of this can be done by invoking single O(1) builtin.

lehins: It might not be a huge concern, but even if you are indexing into an array with O(1), I presume you would need to decode every element in the array before you could index it. So, if all your script cares about is its own TxOut, then I suspect it would be less expensive to just decode that TxOut from script purpose, rather than decode the full array and lookup that TxOut, right?

CIP-????/README.md
rphair: Resolved by author in https://github.com/cardano-foundation/CIPs/pull/1072#event-19159347835.

CIP-????/README.md
colll78: No you don't need to decode any elements to index into it. You just need to call `BI.unsafeDataAsArray`, this is O(1), to get the array, which would be of type: `BuiltinArray BuiltinData`, and then you do `BI.indexArray idx txInputsArray` which is also `O(1)`, and then you have `BuiltinData` which is `ownInput`, which you can extract whatever fields you want from. 

(this assumes we have `BuiltinArrays` and a hardfork has occurred which changed the representation of the script context to have the inputs / outputs as builtinarray instead of builtinlist, and afaik the next HF isn't expected for a while)

rphair:
@michele-nuzzi @fallen-icarus @zliu41 @WhatisRT @lehins - we had some ambitious proposals on the `Triage` list at yesterday's biweekly CIP meeting and so didn't get a chance to confirm or withhold a CIP number.  Editors will keep following online discussion here in anticipation of a less impacted meeting next time: https://hackmd.io/@cip-editors/118

rphair:
@michele-nuzzi at the CIP meeting today we decided to leave this at `Triage` since it's still in a basically new state without your response to the reviews that have already come in:
(Ledger side) https://github.com/cardano-foundation/CIPs/pull/1072#pullrequestreview-3094352109
(CIP editing) https://github.com/cardano-foundation/CIPs/pull/1072#pullrequestreview-3095337164
(Plutus side) https://github.com/cardano-foundation/CIPs/pull/1072#pullrequestreview-3106832230

This has the potential to be a solid CIP if you can engage with the diverse feedback so far, so nobody wanted to mark it `Unconfirmed` today.  Can you please work through the review threads above so then we can reconsider giving this candidate status & a CIP number in 2 weeks' time?

rphair (commented):
@michele-nuzzi we are still in the same state as we were 8 months ago with this submission... with only selective responses to certain review comments, we still haven't done a comprehensive peer review: this might as well be a blog entry with a bunch of comments under it & some author responses.

Has your development in the last year exposed anything else about this problem & can you commit to documenting a solution that the Plutus community should be able to use?  If you want to finish this proposal, the CIP editors can help but — as already requested — you'd have to start going through the reviews as indicated 8 months ago in https://github.com/cardano-foundation/CIPs/pull/1072#issuecomment-3246227489.

Without further response this would be closed in about 4 weeks' time in the `Likely Abandoned` state.  Before or after that happens, reviewers so far in this thread are welcome to submit their thoughts about how this proposal might best be completed: @lehins @zliu41 @colll78 @fallen-icarus @Unisay @WhatisRT

rphair:
@michele-nuzzi closing due to abandonment with reasons already given 3+ months ago in https://github.com/cardano-foundation/CIPs/pull/1072#pullrequestreview-4240972093.  @lehins @zliu41 @colll78 @fallen-icarus @Unisay @WhatisRT please feel free to post any suggestions about how this might be continued & prepared for a proper peer review.

PR #1061, opened 2025-07-23
CIP-???? | Account Address Enhancement

fallen-icarus:
This CIP adds support for depositing ADA into account addresses (a.k.a. reward/staking addresses) which helps alleviate some of the pain-points from the `minUTxOValue` requirement. The CIP also provides a mechanism for plutus smart contracts to get a sense of the account's current balance using *account balance intervals* similarly to how time is specified.

**This CIP does not add support for depositing native assets into account addresses!** There are still open questions about how to safely support this feature, but being able to deposit ADA is critical enough and easy enough to implement to not wait on native asset support.

---
[Rendered Proposal](https://github.com/fallen-icarus/CIPs/blob/account-enhancement/CIP-%3F%3F%3F%3F/README.md)

PR #1077, opened 2025-08-22
CIP-???? | Time-Bound Delegation with Dynamic Rewards Distribution

Cerkoryn:
This is I think the most ambitious CIP that we (the community SPO Incentives Working Group) are proposing.

I think the expired delegations part is already mostly expected and well-understood. But the changes to the rewards calculation were necessary to include together and that required quite a bit more work. The documentation for how this "residual" ADA is returned from the rewards pot to the reserve was hard to find and difficult to follow. It's also highly likely that my math is off, but we included a link to our open source tool that duplicates the math (in a simpler form) using Typescript.

The references to the relevant sections in the Shelley Delegation spec and Shelley Ledger spec are referenced at the bottom and should help any implementors to understand what we are trying to achieve with this CIP.

([rendered](https://github.com/Cerkoryn/CIPs/tree/time-bound-delegation/CIP-time-bound-delegation))


trjate:
This CIP feels very comprehensive and I can't really think of any counter arguments why we wouldn't want to implement something like this. Social contract yada yada yada, but I think encouraging _slightly_ more active participation from delegators is the right way to go. I like the bank analogy.  Staking on Cardano is not opening a savings account.  

However, I read both the CPS and CIP, but I'm not sure I saw an accounting of the fact that lost stake still theoretically increases the number of blocks a pool produces, and therefore the amount of rewards that all active pool stakers receive.  

The CIP drills on the point that lost stake siphons off rewards that otherwise would go to active stakers, but doesn't seem to account for the increase in block rewards it generates, which are shared by other stakers in the pool.  Assuming I have this right I'm not saying I don't agree with CIPs solution to this problem, just that it still feels unclear how lost stake negatively impacts other pool stakers.

My understanding on all of this is certainly incomplete though.

CIP-time-bound-delegation/README.md
```diff
@@ -0,0 +1,277 @@
+---
+CIP: "?"
+Title: Time-Bound Delegation with Dynamic Rewards Distribution
+Category: Ledger
+Status: Proposed
+Authors:
+  - Ryan Wiley rian222@gmail.com
+Implementors: []
+Discussions: []
```
rphair:
```suggestion
Discussions:
  - https://github.com/cardano-foundation/CIPs/pull/1077
```
`this`

CIP-time-bound-delegation/README.md
```diff
@@ -0,0 +1,277 @@
...
+
+**Why this is acceptable to revisit now.** Today’s network conditions and incentives make sustained censorship of delegation renewals economically implausible and strategically dominated by other attack vectors:
+
+- **Healthy stake distribution & coordination threshold.** With broad dispersion of block‑producing power, a would‑be censor would need to withhold a large number of small delegation transactions across many blocks and coordinate with many other producers to move rewards meaningfully. Any unilateral attempt simply fails whenever other producers include those transactions.
+- **Externality of “success.”** Under full‑pot distribution, even a successful bout of censorship raises everyone else’s yields, not just the censor’s. This dilutes the payoff while increasing incentives for others to defect and include the transactions.
+- **Richer targets exist.** In a DeFi‑enabled ecosystem, **selective censorship of high‑value transactions (e.g., liquidations/arbitrages) offers a concentrated, private payoff from censoring few targeted transactions, whereas delegation censorship requires suppressing many of them for marginal gains.
```
rphair:
```suggestion
- **Richer targets exist.** In a DeFi‑enabled ecosystem, **selective censorship of high‑value transactions** (e.g., liquidations/arbitrages) offers a concentrated, private payoff from censoring few targeted transactions, whereas delegation censorship requires suppressing many of them for marginal gains.
```
I am guessing a missing `**` closure was here? (without closing that Markdown span, there may be problem with some renderings: though looks OK on GitHub)

CIP-time-bound-delegation/README.md
```diff
@@ -0,0 +1,277 @@
...
+
+$$
+f_{\text{std}} = R \cdot 0.0000845007526134465 = 1,414,612.80
+$$
+
+$$
+\text{ROI}_{\text{pool,std}} = \frac{f_{\text{std}}}{x + y} 
+= \frac{1,414,612.80}{36,000,000} 
+= 0.03929
+$$
```
rphair: For some reason the GitHub equation rendering breaks down for this item... from my technical level I can't see how it's different but you can currently see it in the rendered preview: only showing up as ordinary Markdown would.

CIP-time-bound-delegation/README.md
```diff
@@ -0,0 +1,277 @@
...
+\text{ROI}_{\text{epoch,full}} = \frac{f_{\text{full}}}{x + y} 
+= \frac{16,435.354078}{35,000,000} 
+= 0.0004695815415 \approx 1.40\%
+$$
+
+$$
+\text{ROI}_{\text{annual,full}} = (1 + \text{ROI}_{\text{epoch,full}})^{73} - 1 
+= 0.03486543814
+$$
```
rphair: These ones are also currently left unformatted in the GitHub preview. 🤔

CIP-time-bound-delegation/README.md
```diff
@@ -0,0 +1,277 @@
...
+
+    ![Runaway lost-stake before mitigation (Fig. 2)](Fig2.png)
+
+  - After:
+
+    ![Runaway lost-stake after mitigation (Fig. 3)](Fig3.png)
+
+**Note:** The baseline numbers in the previous two charts were rough estimates taken from CPS-0022. These estimates may be off by a wide margin, but the effects in preventing the compounding of lost ADA beyond their expiration date will apply generally regardless.
+
+## Path to Active
```
rphair, line 219: @Ryun1 @Crypto2099 @perturbing this is a little long & wordy for _Path to Active_ vs. most other CIPs but I don't see any other way of presenting it.  In other proposals when a lot of the specification depended on _Path to Active_ (e.g. https://github.com/cardano-foundation/CIPs/pull/946) it was fine to move those contingent pieces back into _Specification_ to keep the checklists cleaner.

I don't think that's possible in this case because these items refer to "real world" things happening that aren't part of the Specification.  Therefore our usual recommendation of tick-boxes for this section also goes out the window since we appear to be stuck in a narrative form.  Would be interested in also hearing what you (editors + interested reviewers) think about these matters.

**p.s.** @Cerkoryn in case there are no further comments on this matter — and people are wondering what to do with this suggestion thread — it can be "Resolved" after some reasonable time has passed (just wanted to call a bit of attention to it for now).

CIP-time-bound-delegation/README.md
```diff
@@ -0,0 +1,277 @@
...
+- Homer [AAA]
+- Chad [BBHMM]
+- Mark H [UPSTR]
+- Carlos Lopez de Lara [Input|Output]
+- Pedro Lucas
+- Seomon
+- OYSTR Pool
+
+## Copyright
+This CIP is licensed under CC‑BY‑4.0.
```
rphair:
```suggestion
This CIP is licensed under [CC-BY-4.0](https://creativecommons.org/licenses/by/4.0/legalcode).
```
Standard (social) practice is to make this a convenient-to-follow link 🤓

rphair (commented):
@Cerkoryn looking forward to introducing this in `Triage` at the next CIP meeting (https://hackmd.io/@cip-editors/118) especially since https://github.com/cardano-foundation/CIPs/pull/1060 is also up for `Last Check` there.

@lehins @WhatisRT since this proposal depends so heavily on the Ledger, I hope you can both give this your early assessment and keep contributing guidance on what would ultimately make it a target for implementation.

Since most of the _Implementation Plan_ is about Ledger interaction we need your help to make sure the language there is (or will eventually be) realistic, complete, and correct.  If you think it needs to be broken down more simply (see https://github.com/cardano-foundation/CIPs/pull/1077#discussion_r2293080202) please say so.

CIP-time-bound-delegation/README.md
```diff
@@ -0,0 +1,278 @@
+---
+CIP: "?"
+Title: Time-Bound Delegation with Dynamic Rewards Distribution
```
rphair:
```suggestion
Title: Time-Bound Delegation with Dynamic Rewards
```
p.s. this was the first review suggestion I thought of, but somehow left it out of https://github.com/cardano-foundation/CIPs/pull/1077#pullrequestreview-3143556803:

In the interest of brevity, can we eliminate this final long word after all the other long ones?  There are both social & technical reasons to have short CIP titles.  Does `Rewards Distribution` ever mean anything different than `Rewards` would?  Are rewards ever _not_ distributed: currently or according to this CIP?

WhatisRT:
From a Ledger POV this is a pretty clear proposal and it would be relatively easy to implement.

Have the contents of this CIP been discussed with incentives researchers? Incentive designs are quite tricky and non-intuitive, so this should definitely happen in the path to active. I'd also like to see a distribution of active stake that would be removed by this proposal depending on the initial value of `stakePoolDelegationLifetime`.

Removing a large amount of delegations at once might give an adversary quite a bit of power, so this approach seems quite dangerous to me. Old delegations may or may not be problematic for the network safety, but they add to the "momentum" of the system. If X% of the stake is removed, an attacker would require X% less stake to to the same amount of damage as today. Note also that exchanges (and other big players) are likely not affected by this - they would keep their delegations active. So this removal would likely increase the percentage of stake owned by them.

Requiring periodic review of delegations may also push active delegators away from the ecosystem, or lead to wallets that automate that chore. The former is clearly bad, while in the latter case nothing was gained in the long run.

Cerkoryn:
@rphair I made the formatting corrections and changed the name of the CIP.  The math should render correctly now.

Cerkoryn:
> However, I read both the CPS and CIP, but I'm not sure I saw an accounting of the fact that lost stake still theoretically increases the number of blocks a pool produces, and therefore the amount of rewards that all active pool stakers receive.
> 
> The CIP drills on the point that lost stake siphons off rewards that otherwise would go to active stakers, but doesn't seem to account for the increase in block rewards it generates, which are shared by other stakers in the pool. Assuming I have this right I'm not saying I don't agree with CIPs solution to this problem, just that it still feels unclear how lost stake negatively impacts other pool stakers.

@trjate the intent with expiring delegation certificates is that the lost ADA would become effectively unstaked eventually.  Which would mean they would not contribute to block production for any pools either.

Cerkoryn:
@WhatisRT 

> Have the contents of this CIP been discussed with incentives researchers? Incentive designs are quite tricky and non-intuitive, so this should definitely happen in the path to active. I'd also like to see a distribution of active stake that would be removed by this proposal depending on the initial value of `stakePoolDelegationLifetime`.

I would love to hear the feedback from more qualified incentives researches like @BTL_Edinburgh.

As for the distribution of active stake that would be removed at various parameter settings, that was intentionally left out of this CIP.  The intent is for that research to be presented to the parameter committee and community via governance and/or off-chain vote to determine what the best initial settings should be.

CIP-time-bound-delegation/README.md
```diff
@@ -0,0 +1,283 @@
...
+
+- **Indefinite delegations.** Delegation certificates today have no expiry. Once a stake credential delegates to a stake pool or a dRep, that delegation remains valid indefinitely until the holder submits a new delegation (or deregisters the stake key).
+- **Rewards return-to-reserves.** After monetary expansion and fees are combined and the treasury cut is taken, the distributable rewards pot is scaled by the network’s active‑stake ratio; only that fraction is paid out, and the remainder is returned to reserves. This occurs because rewards are computed using only the active stake snapshot and any residual R not assigned to reward accounts (∆r₂ = R − Σrs) is returned to reserves, per the epoch‑boundary and reward‑update rules (SL‑D5 §11.1, §11.5, §11.8–§11.10).
+
+This CIP intentionally limits itself to two changes to the ledger rules. All other reward mechanics (e.g., pledge influence, saturation, treasury cut) are unchanged:
+
+### 1) Delegation certificate expiration
+
+- Add **two protocol parameters** (measured in epochs), to be set via governance:
+  - stakePoolDelegationLifetime
```
Ryun1:
```suggestion
  - `stakePoolDelegationLifetime`
```
maybe a nice formatting idea

CIP-time-bound-delegation/README.md
```diff
@@ -0,0 +1,283 @@
...
+
+### Acceptance Criteria
+
+- **CIP Editor Approval** – Cardano CIP Editors must confirm that the specification is complete, unambiguous, and internally consistent with existing CIPs.
+- **Consensus on initial parameter values** – An initial values of for the new protocol parameters `stakePoolDelegationLifetime` and `drepDelegationLifetime` must be agreed upon before hard-fork combinator (HFC) activation. The choice should consider operational viability, empirical analyses (e.g., RSS results), and community feedback.
+- **Endorsement by Technical Bodies** – The Cardano Parameter-Change Proposals (PCP) Committee and the Intersect Technical Steering Committee (TSC) should both recommend the proposal as technically sound and aligned with the protocol’s long-term roadmap.
+- **Stakeholder Concurrence** – A majority of stake pool operators (SPOs), ecosystem tooling maintainers, dReps, and other infrastructure providers must signal readiness to upgrade.
+- **Governance Ratification** – The on-chain Hard-Fork Governance Action must pass the requisite dRep and Constitutional Committee thresholds, establishing legal-constitutional legitimacy and stakeholder support for the change.
```
Ryun1, line 227: It may also make sense to update the constitution, and specifically adding guardrails around these parameters

Cerkoryn: We have this mentioned just after this part in the fourth bullet of the implementation plan.

CIP-time-bound-delegation/README.md
```diff
@@ -0,0 +1,283 @@
...
+## Specification
+
+#### Current behavior (for context)
+
+- **Indefinite delegations.** Delegation certificates today have no expiry. Once a stake credential delegates to a stake pool or a dRep, that delegation remains valid indefinitely until the holder submits a new delegation (or deregisters the stake key).
+- **Rewards return-to-reserves.** After monetary expansion and fees are combined and the treasury cut is taken, the distributable rewards pot is scaled by the network’s active‑stake ratio; only that fraction is paid out, and the remainder is returned to reserves. This occurs because rewards are computed using only the active stake snapshot and any residual R not assigned to reward accounts (∆r₂ = R − Σrs) is returned to reserves, per the epoch‑boundary and reward‑update rules (SL‑D5 §11.1, §11.5, §11.8–§11.10).
+
+This CIP intentionally limits itself to two changes to the ledger rules. All other reward mechanics (e.g., pledge influence, saturation, treasury cut) are unchanged:
+
+### 1) Delegation certificate expiration
```
Ryun1: would it be more straightforward to have a 'stake account expiration'?

so you have to refresh your stake account registration every X epochs, I don't feel there is too much utility in having separate expirations for the different types of delegations
I feel it could be easier for users to track a single delegation expiration, rather than two

Cerkoryn: That was something that came up, but de-registering the stake address would potentially cause other problems especially with smart contracts or dApps that rely on the stake address to identify a wallet.

Aiming for the delegation certificates themselves is more straightforward and cuts to the heart of the issue.  Since they are already separate certificates on-chain, it made sense to have two separate parameters as well so that they could be adjusted independently as the community sees fit.  Especially as the issues with pool delegations (compounding rewards) and dRep delegations (governance ossification) are very different in nature.

It may be possible to have a single parameter like `delegationLifetime` that would put the same time constraints on both types.  But that would give the community less control to modify them independently via parameter changes.

lehins: I second @Ryun1 in his opinion. I believe it is wrong to track each delegation type separately. Moreover we don't need to require account owners to unregister in order to prove that the account is still alive. Any action that requires a witness would serve just fine for this purpose.

> Aiming for the delegation certificates themselves is more straightforward and cuts to the heart of the issue.

From my understanding of the CPS and this CIP the heart of the issue is inaccessible accounts due to lost keys or death of the owner. Retained delegations for such accounts are just a consequence, not the other way around. So, how is tracking individual delegations is relevant to being able and prove active ownership and liveness of an account?

lehins (commented):
My personal take on this is that I do not see these points that were mentioned in the CPS as problems:
* Stake delegation to an SPO indefinitely participating in block production. This I can see as a positive thing, since whenever more stake is involved in block production the more secure the network is. Also, whenever someone explicitly expresses desire to delgate their funds to some entity, I do not believe it is fair to retroactively take that intention away from ADA holders.
* Stake Delegation to a DRep. I totally see how laying on my death bed I delegate all my money to my favorite DRep. After all, that was my money and I was free to do with it as I please.

Another important point relevant to this is that most SPOs and DReps do not have infinite lifetime either. They are backed by real people who are not immortal. Any SPO that is inactive will eventually be dropped off and DRep will be marked as inactive.

However, the point that I can agree with is with rewards being occurred over time for lost ADA and as a consequence exacerbate the amount of lost ADA. Also, I don't think one should continue making money after death or due to being dumb and loosing their keys.

From technical perspective, my opinion is that this CIP could be drastically simplified. IMHO, there is no point in tracking expiration for individual delegations.
We could do something very similar to what we've done with DReps:

* Have an explicitly configured account inactivity period through a new protocol parameter that will apply to every account.
* Any action (eg, withdrawal, redelegation, etc.) that requires a witness on the account would prove liveness of that account and as such would reset the counter.
* Inactive accounts would either:
  * not participate in leader election and/or governance, as this CIP desires it to be
  * or my preferred outcome: accrued rewards for inactive accounts would go into treasury, instead of being added to the account. This would solve the exponential run-off without sacrificing the stability and security of the system. Also, from the perspective of the account owner it would be equivalent to "Expired delegations simply stop earning and can be renewed at any time."

This approach would not require anything extra from the majority of legitimately active participants, especially if inactivity window for accounts is set to be something on the order of a few years (which is totally reasonable when comparing to investments accounts in traditional banking). For those parties that are really inactive for many years on end, but still alive and do have access to their accounts' keys, would have to submit a transaction once in a blue moon, if they want to continue making rewards.
The main reason why I would advocate for this parameter to be set on the order of years instead of months or something, because ability to stake and forget on Cardano is one of the most amazing features. That being said, this part can be tuned after the feature  has been implemented.

WhatisRT:
> As for the distribution of active stake that would be removed at various parameter settings, that was intentionally left out of this CIP. The intent is for that research to be presented to the parameter committee and community via governance and/or off-chain vote to determine what the best initial settings should be.

I think this data is vital to the point this CIP is trying to make. I don't want to argue about the values of the parameters here, but without this data it is completely impossible to gauge the impact of this CIP.

> whenever more stake is involved in block production the more secure the network is

That's only true for _honest_ stake, i.e. stake delegated to an honest SPO. And when you don't watch your delegation, your own stake can become dishonest at any time. So this CIP would potentially increase the quality of the delegated stake, but since it also decreases the total stake at the same time it's unclear what the security implications are. If this CIP wants to make a security argument it should provide reasonable estimates for these things.

> Any action (eg, withdrawal, redelegation, etc.) that requires a witness on the account would prove liveness of that account and as such would reset the counter.

That makes sense, basically like what we do for DReps.

Cerkoryn:
@lehins I generally understand and appreciate your perspective here.  However, I have a question on this part:

> From technical perspective, my opinion is that this CIP could be drastically simplified. IMHO, there is no point in tracking expiration for individual delegations.
> We could do something very similar to what we've done with DReps:
> - Have an explicitly configured account inactivity period through a new protocol parameter that will apply to every account.
> - Any action (eg, withdrawal, redelegation, etc.) that requires a witness on the account would prove liveness of that account and as such would reset the counter.
>   - Inactive accounts would either:
>   - not participate in leader election and/or governance, as this CIP desires it to be
>   - or my preferred outcome: accrued rewards for inactive accounts would go into treasury, instead of being added to the account. This would solve the exponential run-off without sacrificing the stability and security of the system. Also, from the perspective of the account owner it would be equivalent to "Expired delegations simply stop earning and can be renewed at any time."

What exactly are you referring to by "accounts" here?  If you mean adding something like our current `dRepActivity` parameter that removes dReps from quorum if they are inactive, but for stake pools, then I don't think it would solve the problem.  Both dReps and SPOs are incentivized NOT to become inactive the more stake they control.  And this is especially true for SPOs who earn a yield based on that stake.

If accounts is referring to "delegators" instead, then I wonder how we could prove liveness for them specifically as that is exactly what this CIP set out to do.

> The main reason why I would advocate for this parameter to be set on the order of years instead of months or something, because ability to stake and forget on Cardano is one of the most amazing features.

FWIW I left this out of scope for the CIP because I didn't want to overly influence this from being a community-driven decision.  But personally I was leaning toward something like ~4 years as the maximum age for a delegation certificate.  And I think wallets could make this a relatively painless UX with reminders or bundling it with any other transaction.

Cerkoryn:
@WhatisRT 

>I think this data is vital to the point this CIP is trying to make. I don't want to argue about the values of the parameters here, but without this data it is completely impossible to gauge the impact of this CIP.

This is a fair point.  We will try and expedite this research and share it here once we have it.  The current intent is to be able to present an Ekklesia vote on both the CIP and the initial parameter setting for the community to vote on at the same time.  And we will try to link any and all relevant research for them to read over along with that if we can.

> So this CIP would potentially increase the quality of the delegated stake, but since it also decreases the total stake at the same time it's unclear what the security implications are. If this CIP wants to make a security argument it should provide reasonable estimates for these things.

I understand what you are saying here, but I'm not sure what kind of reasonable estimates we could come up with that wouldn't be pure speculation about future game theory.  Do you have any examples of what such evidence might look like?

lehins:
> What exactly are you referring to by "accounts" here?

I am referring to reward accounts, i.e. staking credentials, which are used to delegate as well as accrue and withdraw rewards.

> If you mean adding something like our current dRepActivity parameter that removes dReps from quorum if they are inactive, but for stake pools, then I don't think it would solve the problem. 

No I am not talking about stake pools. I am talking about actual delegators that who's ADA "got lost"

>  I wonder how we could prove liveness for them specifically as that is exactly what this CIP set out to do.

Yes, and what I am suggesting is simplifying what this CIP is trying to do.
What this CIP suggests is using delegation certificates to prove liveness of accounts/delegatorts, where I am suggesting that any action on the account constitutes as a proof of its liveness. In other words withdrawing rewards from an account requires exactly the same witness to be provided as if a user submitted a delegation certificate. Both actions prove that owner of that staking credential is alive and kicking.

> personally I was leaning toward something like ~4 years as the maximum age

Totally reasonable IMHO

Cerkoryn:
> ...I am suggesting that any action on the account constitutes as a proof of its liveness. In other words withdrawing rewards from an account requires exactly the same witness to be provided as if a user submitted a delegation certificate. Both actions prove that owner of that staking credential is alive and kicking.

@lehins I think we had a similar idea early on in our discussions, except we were saying any transaction at all from a wallet (not just the rewards account) would constitute its proof-of-life.  My understanding is that it even though the concept is simpler, the implementation would be more difficult.

Consider the node looking at each delegator on epoch turnover.  It would have to examine each wallet to figure out when the last transaction was made.  Once it has that information, it then calculates the difference between that epoch and the current epoch to see if too much time has past.  Whether or not it passes that check determines if it is eligible for rewards.  And it would have to do this check in future epochs as well, not just the one time when it expires.

On the contrary the node already has to look at the delegation certificate.  If we stamp that certificate with the "Expiration Date" then no wallet transaction history lookups need to be done which allows the delegation to be voided much sooner and with only a single check (i.e., `if currentEpoch > expirationEpoch`)

It's possible my understanding of this is incorrect though.  If so then it sounds like your method would indeed be better and I'd be happy to be enlightened.

lehins:
> I think we had a similar idea early on in our discussions, except we were saying any transaction at all from a wallet (not just the rewards account) would constitute its proof-of-life.

This was an incorrect direction. Even in theory you can't associate one wallet with one delegator. For example, Lace allows multi-delegations, which means they support multiple staking credentials, which from the ledger perspective are indistinguishable. 
Also you can have a wallet that doesn't have a staking credential at all.
So, this was conceptually a wrong direction.

> My understanding is that it even though the concept is simpler, the implementation would be more difficult.

What I am suggesting is quite a bit simpler from the implementation perspective. It is easier to track one expiration vs two separate ones for each account. We would not need to add two protocol parameters.
Most importantly the goal is to prove that an account is "active", not that a particular  delegation is active.

> Consider the node looking at each delegator on epoch turnover. It would have to examine each wallet to figure out when the last transaction was made. Once it has that information, it then calculates the difference between that epoch and the current epoch to see if too much time has past. Whether or not it passes that check determines if it is eligible for rewards. And it would have to do this check in future epochs as well, not just the one time when it expires.

I think you can trust me when I say that my suggestion will be simpler to implement in Ledger ;) Your assumption of how implementation would work is totally incorrect. In fact it would be impossible to implement the way you suggested.

> It's possible my understanding of this is incorrect though. If so then it sounds like your method would indeed be better and I'd be happy to be enlightened.

My attempt at enlightening :smile::

1. Each reward account has an expiration epoch number associated with it.
2. Whenever stake is being computed for voting and/or leader election, at that point expired accounts (ones who's expiration is in the past) are simply ignored.

1. Any action that requires a witness for an account moves expiration for that account into the future.

Done.

This has a huge benefit of simplicity of implementation for the Ledger team, as well as simplicity for the end user. Most notably wallets usually default to withdrawing rewards upon submission of any transaction from a wallet, which means as long as there is any transaction submitted by a wallet within the account inactivity period (of what you suggested should be 4 years), then the associated account does not expire. Delegation to a DRep or SPO would have the same affect.
Moreover, with the [CIP-159](https://github.com/cardano-foundation/CIPs/pull/1061) accounts will be used even more frequently, which means that this inactivity period will be a minor nuisance and only for those long term inactive HODLers.

Cerkoryn:
I trust your understanding over my own on these implementation details for sure.

> My attempt at enlightening 😄:
> 
> 1. Each reward account has an expiration epoch number associated with it.
> 2. Whenever stake is being computed for voting and/or leader election, at that point expired accounts (ones who's expiration is in the past) are simply ignored.
> 3. Any action that requires a witness for an account moves expiration for that account into the future.

I think this makes sense.  To simplify it in my own words, this is basically using the wallet rewards account like a global variable that gets updated each time it's used, is that correct?

If so I think you're correct that this achieves the same goals our CIP is trying to achieve, but in an even simpler way.  And this would allow us to have only a single protocol parameter enforce a minimum level of liveness on both dRep delegators and stake pool delegators.

I'll bring it up with our group at our meeting this week.  And if its alright with the other reviewers I may edit the CIP to reflect this superior implementation after that.

TerminadaPool:
> My personal take on this is that I do not see these points that were mentioned in the CPS as problems:
> 
> * Stake delegation to an SPO indefinitely participating in block production. This I can see as a positive thing, since whenever more stake is involved in block production the more secure the network is. Also, whenever someone explicitly expresses desire to delgate their funds to some entity, I do not believe it is fair to retroactively take that intention away from ADA holders.
> * Stake Delegation to a DRep. I totally see how laying on my death bed I delegate all my money to my favorite DRep. After all, that was my money and I was free to do with it as I please.
> 
> Another important point relevant to this is that most SPOs and DReps do not have infinite lifetime either. They are backed by real people who are not immortal. Any SPO that is inactive will eventually be dropped off and DRep will be marked as inactive.

Even though the humans running the stake pool or DRep are mortal, the stake pool keys or DRep keys themselves are immortal.  They can be transferred and sold.  Or, alternatively, the human operator(s) of the stake pool or DRep can change over time.  IE: When you staked with your chosen pool they might have had honest people faithfully implementing the Cardano protocol but some time later the pool keys, or DRep keys, might get sold to Blackrock and your stake would then be contributing to Blackrock's "control" over Cardano block production, or Cardano DRep voting power.

lehins:
>  some time later the pool keys, or DRep keys, might get sold to Blackrock and your stake would then be contributing to Blackrock's "control" over Cardano block production, or Cardano DRep voting power.

Your argument is based on the assumption that there is transparency from SPOs and DReps about what is going on behind the scenes, which is not necessarily true. Even today any SPO or DRep can pretend to be anyone they want. Moreover, any SPO or DRep can be compromised and no one would even know until there was some malicious action on their behalf. 
In other words, the only time when it can be visible whether an SPO or DRep is malicious is from the actions they perform. For example if a block producer is intentionally trying to fork  the network.

In either case, my presumption is that a malicious SPO or a DRep will not be backed solely by "lost ADA". If community learns that an SPO or DRep is malicious then most of live stake will be undelegated and only "lost ADA" will remain, in which case by definition of the CPS in will be a small percentage, which will not be enough to have any affect on the network.

In my opinion it is a lot more dangerous for a whole lot of stake to get undelegated at once, when all of the ADA that was lost over many years gets removed from block production.

I understand that you are acting with best intentions and you are trying to improve security of Cardano as a whole, but we need to make sure we do not make things worse then they are today! 

I would really love to see some decent research done before we do anything in this direction. So far I've seen only a lot of speculation that is based on a whole lot of assumptions. It is understandable that many of the people involved are doing this during their free time, but we do not want to make rush decisions based solely on opinions and incomplete research. We need proper research of historical data. We need some research from game theory specialists. We need more accurate predictions that are based on that research.

Even my personal opinion that I've expressed so far is also that, just an opinion. The only thing that I am certain about is the solution that I suggested. However, despite that I suggested the solution, it does not yet mean that I am convinced that the problem this CIP is trying to solve is indeed a real problem.

For everyone involved. Design of Cardano was always based on facts and thorough research. I recommend we do not start deviating from that all of a sudden.

That being said, I really do appreciate all the work that everyone involved has done on this so far. It is initiatives like this that either improve or confirm the overall security of Cardano blockchain!

TerminadaPool:
> Your argument is based on the assumption that there is transparency from SPOs and DReps about what is going on behind the scenes, which is not necessarily true. Even today any SPO or DRep can pretend to be anyone they want. Moreover, any SPO or DRep can be compromised and no one would even know until there was some malicious action on their behalf.
> In other words, the only time when it can be visible whether an SPO or DRep is malicious is from the actions they perform. For example if a block producer is intentionally trying to fork the network.

Totally agree, there is no transparency and nobody can tell how a pool operator or DRep will act in the future.  The Ouroboros protocol assumes that people can and will re-stake.  But lost / inactive stake won't and this stake provides extra voting / block production power.

How much does it cost to purchase voting / block production power?  If I buy Ada then we know how much that costs, but I will be able to buy that voting / block production power more cheaply by buying some DRep / pool keys that have a lot of lost / inactive stake _permanently_ "attached" to them.

There are two issues:
1. Obviously there is the security concern I am articulating.  We can debate how significant this is currently because we have no way of really knowing how much lost / inactive stake there really is.  But we do know that it will accumulate over time.
2. However there is also an unfair competitive advantage concern which can be compounded over time.  This is because OG pool operators have an unfair competitive advantage whereby they can create new pools charging lower costs and increase the costs on their old pool as a way to cross subsidise expenses.  They know the lost / inactive stake in their old pool can't / won't move.  This allows them to accumulate more stake in their new pools by undercutting competitors whilst still retaining the lost / inactive stake in their old pool.  Then rinse and repeat over time.  Lost / inactive stake will accumulate in all pools over time but arguably more in lower cost pools or ones that get high wallet rankings since time poor people are probably both more likely to simply click the highest ranked pool and also to lose their keys.

    So how do you get a high wallet ranking?  Well one way is to be cheap but another way is by having high pledge.  And, this lost / inactive stake also allows the OG operator to shift his pledge with less backlash from stakers.  This is because the lost / inactive stake isn't watching and doesn't care, but more pledge will attract new stakers to his new low cost pool.

    Now this leads back to the security issue as they accumulate lost / inactive stake over time _especially_ since we don't actually know whether a new low cost pool is actually run by an OG operator with other pools benefiting from lost / inactive stake.  Any OG operator can create any number of new pools and it is really impossible for the community to know who is actually pulling the strings behind the scenes.

The set and forget mechanism with no expiry that Cardano currently uses for staking to pools or DReps is flawed.

> In my opinion it is a lot more dangerous for a whole lot of stake to get undelegated at once, when all of the ADA that was lost over many years gets removed from block production.

I think that depends on assumptions about who is pulling the strings on the pool / DRep keys benefiting from the increased block production / voting power that lost Ada provides.

fallen-icarus:
I have a few thoughts that make me hesitant about this CIP:

Cardano's Proof-of-Stake launch was *extremely hyped up*. A lot of people had joined the community during that hype and staked their ADA then. Fast-forward to now, there is a lull from Cardano's disappointing DeFi applications. I think it is a sure bet that most people are not paying attention right now. If we implemented expirations during this lull period, I highly doubt the total delegated stake would recover post-expiration. We could see our stake participation drop from 70% to 40% or less.

> In Cardano, **staking is voluntary and active**.

This is not accurate. Passive vs active depends on how frequently actions need to be taken. If I only need to adjust my delegation once every few years, how is that active? If my pool is doing fine and acting honestly, I may never need to change my delegation in my lifetime. *PoS is designed to be passive when times are good*. It is like voting participation: most people only vote during troubled times.

---
> ... accrued rewards for inactive accounts would go into treasury, instead of being added to the account. This would solve the exponential run-off without sacrificing the stability and security of the system. Also, from the perspective of the account owner it would be equivalent to "Expired delegations simply stop earning and can be renewed at any time."

Given my above thoughts, @lehins approach is the only one that seems safe because:
1. It won't cause a significant drop in Cardano's total stake due to Cardano's popularity currently being at a low.
2. It won't punish rational actors who are *correctly* behaving passively in a PoS system.

Cerkoryn:
> I think it is a sure bet that most people are not paying attention right now. If we implemented expirations during this lull period, I highly doubt the total delegated stake would recover post-expiration. We could see our stake participation drop from 70% to 40% or less.

This is why we combined the expirations with the second part of the CIP that includes a change to distribute the full rewards pot.  It is a counter-balancing mechanism to mitigate this concern because with the change, the lower the stake participation gets, the higher the reward for staking.  This is easiest to see in our [tool](https://spo-incentives.vercel.app/) by experimenting with the `Staked Ratio` slider with rewards mode set to `Current` and `Full`.  The latter represents the change suggested by this CIP.

Unfortunately if we simply returned excess rewards from inactive accounts to the treasury, it would defeat this purpose because now there would be no counter-balancing mechanism to offset a worrisome drop in stake participation.

> Passive vs active depends on how frequently actions need to be taken. If I only need to adjust my delegation once every few years, how is that active? If my pool is doing fine and acting honestly, I may never need to change my delegation in my lifetime. PoS is designed to be passive when times are good. It is like voting participation: most people only vote during troubled times.

The **initial** act of delegating is active because it requires a delegator to select a pool and then submit a transaction.  From that point on, it requires regular re-assessment *somehow* to actually know whether or not your pool is doing fine and acting honestly (this is also technically active, though it isn't necessary reflected on-chain).  As it stands right now we have no way of knowing if users are doing that re-assessment and so they could be empowering malicious behaviors unintentionally.  Especially if those users *can't* move their delegation due to lost keys, death, etc.

This CIP's intention is to increase the *quality* of delegated stake (as @WhatisRT mentioned) by adding a mechanism to enforce that periodic re-assessment AND by rewarding such activity more directly.

Cerkoryn:
There is also the issue of reserve sustainability to consider (This relates to full-pot distribution. The second point of this CIP)

Yesterday the Leios CIP was published.  The specific area to draw your attention to is the charts under [Operating Costs](https://github.com/cardano-scaling/CIPs/blob/leios/CIP-%3F%3F%3F%3F/README.md#operating-costs).

Consider that in order to match our rewards rate by 2030 we'd need to sustain 41.9 TPS worth of transactions (actual transactions, not capacity).  If I'm understanding their wording correctly, they are also suggesting that it may be 5x that amount (209.5 TPS) because SPOs don't receive 100% of rewards, it's actually closer to 20% (I believe this is an average of pool margin, fixed fee, etc. but it's not quite clear what they mean).  Furthermore, in [CPS-22](https://github.com/Cerkoryn/CIPs/blob/sticky-stake/CPS-0022/README.md) we projected that the rewards APY is on track to be 0.6% by 2030.

I do not believe this is sustainable, and yet it is a hard problem to solve.  Simply adding partner chain AVS (or other rewards income) *WITHOUT* adding some sort of proof-of-life mechanism amplifies the compounding rewards problem and is also unlikely to be enough.  This CIP enhances sustainability by being slightly more selective about which wallets are eligible for rewards (i.e., active once every `delegationLifetime` epochs) therefore allowing us to stretch out our reserves even more.

fallen-icarus:
> This is why we combined the expirations with the second part of the CIP ... the lower the stake participation gets, the higher the reward for staking.

IMHO higher rewards is not going to meaningfully influence the total stake delegated right now. If everyone agreed that ADA was valuable, then offering them more ADA *would* incentivize action. But what if they don't value ADA? This is offering them more of an asset they don't care about. People are currently leaving the Cardano community because they don't see value in it anymore. The inactive stake is likely from this latter group which means offering slightly higher rewards is not going to work. The incentives are targeting the wrong group of people. The current inactive stake will only become active again if they actually see value in Cardano. It is sad to say, but I think there is a decent percentage of Cardano's security that is being propped up by users who have just "written-off" their ADA. (I'm open to data that proves this wrong.)

That being said, if the confiscated rewards from inactive delegators could go to the active delegators, I think that would be ideal. But I don't know how complicated that calculation would be for the ledger. My biggest concern is that we don't touch the total stake delegated for now.

fallen-icarus:
> There is also the issue of reserve sustainability to consider ...

This CIP doesn't really address this, especially if we just confiscate the rewards from inactive delegators and pay them out to active delegators. The same amount of the reserves are being spent. We could recycle the confiscated rewards back into the reserve, but then we aren't rewarding active delegators.

IMHO both the inactive stake and the sustainability issue are solved by making Cardano actually useful to the world. If its DeFi was actually useful, its tx volume (and collected fees) would increase and so would the value of ADA vs fiat. Right now, Cardano blocks seem to average about [90% empty](https://cexplorer.io/block) so increasing TPS to 50 won't do anything.

Cerkoryn:
> The current inactive stake will only become active again if they actually see value in Cardano.

I agree, but "Increasing the value of Cardano" is a much broader and harder problem to solve.  This CIP is much more narrowly-scoped.

> It is sad to say, but I think there is a decent percentage of Cardano's security that is being propped up by users who have just "written-off" their ADA. (I'm open to data that proves this wrong.)

This is good from the perspective of an EXTERNAL attacker, but bad from the perspective of an INSIDER threat who might already be accumulating lots of stake (or purchasing an established pool with it) with malicious intent.

> That being said, if the confiscated rewards from inactive delegators could go to the active delegators, I think that would be ideal. But I don't know how complicated that calculation would be for the ledger.

That's exactly what this CIP is aiming to achieve and I think it is arguably simpler than how the rewards are currently distributed.  Right now, the rewards pot is built based on TOTAL circulating ADA, but then only the ACTIVE-staked proportion is distributed, with the remaining going back to the reserves.  The CIP simplifies that to simply distribute the entire proportion to active, eligible stakers.  I haven't been able to find any good documentation on this other then section 5.5.1 of the Shelley Delegation spec:

image

However (as explained in the CIP) I think the censorship issue they describe there is no longer relevant.

> This CIP doesn't really address this, especially if we just confiscate the rewards from inactive delegators and pay them out to active delegators. The same amount of the reserves are being spent. We could recycle the confiscated rewards back into the reserve, but then we aren't rewarding active delegators.

The CIP touches on this in the Rationale under the section `Option to adjust Rho`.  However, we are *already* returning excess rewards to the reserves.  What we are suggesting is that we instead distribute that excess to active delegators, and then we could lower Rho accordingly.

Cerkoryn:
@lehins I did a minor re-write of the CIP to try and capture your method of using the most recent witness on the reward account to update `expirationEpoch`. Please let me know if I captured the idea correctly.  If not I will make any necessary adjustments.

Additionally, there is one complication related to this line:
```
- **Migration.** On activation, determine the epoch that the last witness occurred for each rewards account.  Then initialize each account's `expirationEpoch` to that epoch plus `delegatorInactivity`.
```

I'm not sure how complicated that would be to do in-practice.  Especially if all of those wallet lockups have to happen at the moment of the hard fork.  However, we also can't just set it to `currentEpoch + delegatorInactivity` because then that will likely cause a future incident when all inactive accounts expire at the same time.  It definitely benefits to have them staggered, and I also think it would be helpful to backdate to older accounts so that we can get the benefits of this CIP sooner.  (i.e., not have to wait 4 years after the hard fork for an even larger amount of expirations to start occuring).

Cerkoryn:
@fallen-icarus 

I also expanded on the `Option to adjust Rho` section to capture more of the reserve sustainability issues we talked about, because I believe that is very important to consider.

perturbing:
Just a thought: what is the impact of adding a transparent countdown to staking keys from a security perspective?

From the perspective of an adversary, by adding this countdown, I can now make statistical claims about the best window to launch an attack on the chain. I can imagine that the current design is a deliberate security design choice.

fallen-icarus:
Is there any data to quantify how much ADA is currently active? Given the transparency of the blockchain, it should be possible to figure out:

1. Since payment addresses are tied to staking addresses, what percentage of all staking addresses had their associated payment addresses send/receive ADA this past year? This could give an estimate of how many active users there are, but this metric will likely miss lurkers.
2. What percentage of all Cardano stake addresses have at least withdrawn rewards, changed pool delegation, and/or delegated to a DRep in the past year? Comparing this with the prior metric could give an estimate of the number of lurkers in Cardano.

>  As it stands right now we have no way of knowing if users are doing that re-assessment and so they could be empowering malicious behaviors unintentionally.

The above metrics can help. My hypothesis is that the lurkers will dwarf the active users because PoS is a passive mechanism. Yes, you have to set it up initially, but it is passive from then on as long as there are no issues. Without these numbers, it is hard to see if this is actually a problem right now or not.

Cerkoryn:
> Just a thought: what is the impact of adding a transparent countdown to staking keys from a security perspective?
> 
> From the perspective of an adversary, by adding this countdown, I can now make statistical claims about the best window to launch an attack on the chain. I can imagine that the current design is a deliberate security design choice.

@perturbing Are you aware of any explicit documentation of that as a design choice? I've talked to a couple of the writers of the Shelley Ledger/Delegation spec about the second part of this CIP (full-pot distribution) and seen the reasoning documented there. And that is what we cited in the rationale when we reasoned that those concerns are no longer an issue.

Cerkoryn:
@WhatisRT @fallen-icarus I have some preliminary data on how much delegation would be considered active/inactive as it pertains to this CIP.  

| Category                                 | Wallets (qty) | ADA Total (approx) |
|-----------------------------------------|--------------|-------------------|
| 4 years inactive (delegated or not) | 285,183      | 672,823,257 ADA   |
| **4 years inactive (currently delegating)**  | **180,152**      | **437,871,767 ADA**   |
| 1 year inactive (delegated or not)                          | 1,908,365    | 5,389,296,923 ADA |
| **1 year inactive (currently delegating)**        | **889,317**      | **3,752,708,944 ADA** |

The ones marked "currently delegating" are the ones relevant to security, block production, rewards, and voting power.  The others are included for informational reasons.  Considering that roughly 21B ADA is currently delegated, the delegated wallets that have been inactive for 4 years represents roughly 2% of stake and delegated wallets that have been inactive for only 1 year represents 17% of stake.  It's also worth noting that these would of course be upper limits as many of those wallets surely could be active, but simply don't have a reason to be without something like this CIP in place.

I'm working on both personally verifying this data and getting more visibility in increments of 1, 2, 3, 4, and 5 years inactivity.  Also I'd like to filter on retired vs active pools.  Unfortunately I don't have easy access to something that can query this data easily and am at the mercy of Koios rate limits by having to do this as a large triple-nested query (pools->wallets->transactions).

fallen-icarus:
@Cerkoryn How are you defining "inactive"?

Cerkoryn:
> @Cerkoryn How are you defining "inactive"?

No recorded transactions from that wallet during that period of time.

fallen-icarus:
Expiring 17% is a huge security hit. I recognize it is an upper limit but we don't know how much it over-estimates things. What about a tiered approach?

1. Inactive for 1 year and you no longer accrue rewards, but your ADA still counts for pool leader selection and governance.
2. Inactive for 4 years and your ADA is fully disregarded for calculations.
3. Inactive for 10 years and the ADA is reclaimed into the treasury (i.e., unlocking lost ADA).

Cerkoryn:
> Expiring 17% is a huge security hit. I recognize it is an upper limit but we don't know how much it over-estimates things.

I agree, 17% is way too much.  But 2% for as an upper limit for a 4-year expiry seems pretty reasonable.   My original guess was somewhere around 4-5 years (292-365 epochs) for the initial parameter setting.

> 1. Inactive for 1 year and you no longer accrue rewards, but your ADA still counts for pool leader selection and governance.
> 2. Inactive for 4 years and your ADA is fully disregarded for calculations.
> 3. Inactive for 10 years and the ADA is reclaimed into the treasury (i.e., unlocking lost ADA).

I feel like option 1 might create additional concerns related to separating block production from rewards, but maybe in this specific case it could be okay.

Option 2 by itself is more or less what I'm personally aiming for with this CIP, though OFC I believe that ultimately the decision for the initial parameter setting should be up to the community via governance.

Option 3 is something we've actually talked about at length in regards to demurrage/storage rent (see [this](https://x.com/Cerkoryn/status/1832399934798414232) thread.  However, we decided to table any discussions about demurrage because the idea of reclaiming ADA is both controversial AND technically complicated (imagine how to treat UTxOs that have been idle in smart contracts for many years.  It could break things in unexpected ways if the ledger simply yoinks them away suddenly.)

One day in the future if we get into discussions about fixed supply vs. tail emissions though, I'd be happy to bring it back to the table as a potential third option.

CIP-time-bound-delegation/README.md
```diff
@@ -0,0 +1,284 @@
+---
+CIP: "?"
```
rphair:
```suggestion
CIP: 163
```

CIP-0163/README.md
```diff
@@ -0,0 +1,284 @@
...
+
+**Balancing mechanism in this CIP.** The proposal deliberately pairs expirations with full‑pot reward distribution. Because the entire pot is distributed over the then‑active stake, the per‑unit yield rises automatically when participation drops and compresses when it rises. This creates a fast, transparent negative‑feedback loop:
+
+- If active stake decreases, epoch yield increases for active delegators/SPOs. Capital is attracted back until the security margin normalizes.
+- Likewise if active stake increases, epoch yield decreases. Though without changes to Rho the net result of this CIP will still increase rewards all the way up to 100% of active-stake ratio.
+
+This dynamic couples network security to real participation and reduces the risk of active-stake dropping too low.
```
Cerkoryn, line 98: @rphair I've made the changes to the directory and rendered link in the original comment as-requested. 

@perturbing 
> ... as per the meeting commentary we look forward to settling this question (https://github.com/cardano-foundation/CIPs/pull/1077#issuecomment-3245575675 re: potential insecurity from exposed populations of expiring delegations) regardless of how it has been already addressed in the literature & even if that consideration was ruled out by Cardano's initial academic work & design choices.

I think that the `Security consideration: participation shocks after large expirations` section touches on the issue in this comment a little bit, though I could add more about a specific "timing attack" if you think that would be helpful.  I think that pairing the expiration with the dynamic rewards mechanism as this CIP does works both ways.  If a large expiration were to occur, an adversary could potentially try to accumulate enough stake to pull of an attack at the same time.  But since these large expirations also come with additional rewards for other delegators, they are likely to increase their delegation in anticipation of these greater rewards, preempting any such attack.

I also think that if we choose a conservative initial value for `delegatorInactivity` such as the 4 years mentioned that affects the upper limit of 2% stake, the risk is probably relatively low.

CIP-0163/README.md
```diff
@@ -0,0 +1,284 @@
...
+- **Migration.** On activation, determine the epoch that the last witness occurred for each rewards account.  Then initialize each account's `expirationEpoch` to that epoch plus `delegatorInactivity`.
+
+### 2) Full‑pot reward distribution
+
+- In each epoch, distribute the **full rewards pot** (monetary expansion + fees, less the treasury cut, blocks not produced [Eta], and unmet pledge) to eligible delegators and pools without returning any remainder to reserves **(i.e., eliminate the residual ∆r₂ described in SL‑D5 §11.10 and distribute that proportion as well)**.
+- “Eligibility” remains exactly as today, except that stake controlled by an **expired reward account** does **not** earn rewards (and expired reward accounts contribute **no** governance voting power).
+
+**Note:** The initial numeric value for `delegatorInactivity` is intentionally **left undefined** here and will be chosen through community governance and deliberation. Potential jittering or lottery mechanisms to stagger per-epoch expiries are not mentioned in this proposal, but may be considered separately if operational data warrants it.
```
Cerkoryn, line 54: > It was also noted at the meeting that the gravity of this change may delay its implementation for quite some time: depending on the exact specification that is settled upon. This likely makes the deliberation in https://github.com/cardano-foundation/CIPs/pull/1077#issuecomment-3245916046 significant because we could either have a CIP that could be implemented early in the coming 5 years or something that might not be agreed upon for implementation for another generation or so.

It may also be worth noting the mention of jittering or lottery mechanisms here to add a bit of randomness (at the tradeoff of complexity) if we believe that a timing attack is a significant risk.

TerminadaPool:
If we are going to highlight a supposed security concern related to deregistering a small fraction of the currently staked Ada then it is only fair to also highlight the security concern that arises from _not_ deregistering this Ada.

The voting rights, or block production rights, for this inactive / inaccessible Ada can currently be purchased at a substantial discount to the Ada price by purchasing the pool keys or DRep keys to which this Ada is permanently attached.

In the case of a stake pool, the current operator can simply register new pools and encourage active participants to move to his new pools.  The inactive stake will stay in the old pool and he can sell these old pool keys to the malicious adversary.  Let's say his old pool has 5M Ada of inactive stake.  To a well intentioned pool operator this inactive stake is currently worth maybe 5% of 2.5% per year income (if pool fees are levied at 5%).  This equates to around 6,250 Ada per year in revenue, reducing over time.  An adversary might be able to purchase these pool keys for something like this amount up front, or maybe more, or maybe less, depending on the operator.  However there is no way that the cost of these pool keys in the hands of a rational willing seller will be anything close to the full cost of the 5M Ada.  The most that anyone can earn in staking rewards from this 5M inactive stake is currently 125,000 Ada per year if pool fees were increased to 100%.  However, these keys still provide the adversary with the full 5M worth of block production consensus power and since the adversary is interested in doing adverse things this is what he cares about.  Consequently he can purchase 5M worth of consensus power for _much less_ than 5M Ada and he knows this 5M Ada will always back his malicious activity because it is inactive and can't move _forever_.

Due to the fact that nobody can really know who is actually running a pool, there is no way that anyone can even know if an adversary hasn't already accumulated significant portions of inactive staked Ada by purchasing other pool keys.

An adversary could already be controlling significant portions of all the inactive stake.

CIP-time-bound-delegation/README.md
rphair: Resolving post co-editors' review without their putting any additional flags on this issue (the "wordy" path-to-active which I personally think is OK).

lehins:
@TerminadaPool Many holders of ADA delegate to a pool and do not look at that stake pool's activity, ever, despite that we would all still consider them active, since they might be occasionally transferring ADA, withdrawing rewards or buying NFTs or what have you! Point is, that if your concern is that an adversary buys up bunch of stake pools, then your concern should not be the "lost ADA", but all the people that do not pay attention, because I am more than sure they make up a much larger portion of the stake, when compared to "lost ADA". 

In my opinion, all this speculation that adversaries are buying up pools with a lot of stake delegated to them is just that, a speculation. However, if misbehavior of stake pools is your main concern, maybe we should try and attack this problem for the side of stake pools, instead of delegators? I mean, most of the assumptions here are based on the fact that all ADA holders are watching their stake pools as hawks, which could not be further from the truth. I personally have not looked at a stake pool that I am delegating to for a couple of years now.

lehins:
> If we are going to highlight a supposed security concern related to deregistering a small fraction of the currently staked Ada then it is only fair to also highlight the security concern that arises from not deregistering this Ada.

@TerminadaPool security concern of deregistering ADA can be backed up by math because it would be predictable, while "security concern that arises from not deregistering" is based on speculation that adversaries are buying stake pools up. Morever, former could be exploited without any cost to an adversary, while the latter actually has real cost to it for an adversary. There are various attack possible on Cardano, if one digs hard enough, but they are all prohibitively expensive! Is this attack that being speculated is cheap or expensive?

So, how much does it cost to buy a stake pool from some one? What is the cost per stake? How much stake does one need to mount an attack? What kind of an attack one can mount if they have enough stake? With all due to respect, but I would love to see some concrete answers to these kind of questions, instead of scare tactics of "An adversary could already be controlling significant portions of all the inactive stake."

rphair:
@TerminadaPool - "buying stake pools with significant amount of stuck stake" also has a self-limiting factor: the more such a hypothetical thing would happen, the more it would create a "market" that could be observed... since potential sellers would have to offer their non-performing pools to buyers who understood & were actively interested in such an unconventional proposition.  The protocol wouldn't currently be able to prevent these arrangements, but we could still therefore be assured that the problem wouldn't be invisible... since it were to emerge then we'd see that market emerge along with it.

perturbing:
And to add to what @lehins requested—please ground arguments in data. The Cardano chain is pseudo-anonymous: while stake pool operators identities are not directly visible, their on-chain actions inevitably leak information. For example, @SmaugPool has previously highlighted cases where centralized operations could be inferred from relay locations or operational patterns.

If there is a concern that adversaries are already controlling significant portions of inactive stake, it would be really valuable to see supporting evidence. For instance, analyses of relay clustering, correlated operational timing, or statistical linkage of stake pool behavior could help demonstrate and quantify this risk.

Furthermore, @TerminadaPool, a stakepool cannot genuinely be sold; it can only be leased. Because the previous owner can't simply forget the cold keys.

As for the idea of directing rewards from inactive stake to the treasury, I think that’s a solid approach. It also addresses my earlier concern about attackers being able to make statistical inferences about the optimal time to target the chain. The only caveat is that certain staking scripts might not support restaking. While such a limitation would point to a poorly designed Plutus script, it’s possible that some operators have relied on this assumption.

TerminadaPool:
> security concern of deregistering ADA can be backed up by math because it would be predictable, while "security concern that arises from not deregistering" is based on speculation that adversaries are buying stake pools up. Morever, former could be exploited without any cost to an adversary, while the latter actually has real cost to it for an adversary.

I don't know if owners of pool keys with significant inactive stake will have future malicious intent or not.  But I am pointing out that nobody can know how they will behave in the future, and that collecting the voting/consensus power of inactive stake will be cheaper than purchasing that same amount of Ada outright.  Thus buying (or leasing) keys with significant inactive stake will be a cheaper way to gain additional voting/consensus power.

As to how much Ada can make a difference, it is noteworthy that in the recent constitutional committee elections the difference between winning the 7th committee seat and 8th place was 142M Ada of voting power.

> Furthermore, @TerminadaPool, a stakepool cannot genuinely be sold; it can only be leased. Because the previous owner can't simply forget the cold keys.

Of course they can permanently delete all copies of the keys.  Though, I take your point that this might be difficult for the new owner to confirm depending on his power/economic relationship with the original owner, or other guarantees he can put in place.  In any case the adversary may only want to lease the DRep/Pool keys to mount his attack.

> "buying stake pools with significant amount of stuck stake" also has a self-limiting factor: the more such a hypothetical thing would happen, the more it would create a "market" that could be observed... since potential sellers would have to offer their non-performing pools to buyers.

:eyes: I don't think a visible market is necessarily inevitable.  There could be other methods or deals that allow an adversary to obtain pool or DRep keys other than by openly bidding for them.  Though I don't think it better to wait until it is obvious there is a market for pool keys with inactive stake before we implement an expiry for inactive stake, even if only after 5 years of no action?

It is also important to highlight the other problem with inactive stake forever providing consensus/voting power: and that is that it is simply anti-competitive against new entrants.  A new stake pool operator is unable to fairly compete on price due to the extra revenue inactive stake provides old "OG" operators.  This is even more of an economic advantage for multi-pool operators because they can set higher fees on old pools with inactive stake to cross-subsidise the running costs of the cheap new pools they create.  On this point, the Cardano Constitution is clear since it outlines how Cardano should be fair to all participants, ie: new or old.

Clearly my arguments ruffle the feathers of those closely associated with large old pools because these people realise that unstaking inactive Ada will force them to compete fairly on price; which is only fair.

Cerkoryn:
I think it's worth taking a step back and looking at it from the perspective of an insider threat instead of simply counting the cost to an external attacker starting with zero ADA.

An established SPO or pool group that already has large amounts of stake could, at any point, become adversarial for any number of reasons.  It could be that they sold their pool(s), they could become disgruntled at the state of Cardano, they could have their cold keys compromised, etc.  I don't think it is necessary or even possible to come up with an exhaustive list of reasons why someone could turn adversarial, but it is important to still recognize that the threat exists. Some pool groups might already have sufficient stake to pull off certain kinds of attacks, such as a long-range attack, a forced rollback, or governance attacks.  You could also consider things like setting margin to 100% suddenly, which is not necessarily an attack but it is something that may be considered unethical.

I think the rational expectation from a security perspective is for delegators to notice irrational or malicious behavior as soon as possible and then to delegate away to a more trustworthy SPO.  The specific subset of those delegators mentioned in [CPS-22](https://github.com/cardano-foundation/CIPs/pull/1060) are those that CAN'T delegate away, for whatever reason.  This particular CIP addresses that problem, but also comes with a host of other benefits listed in the rationale section that are independent from that CPS that should also be weighed accordingly.

Cerkoryn:
> As for the idea of directing rewards from inactive stake to the treasury, I think that’s a solid approach. It also addresses my earlier concern about attackers being able to make statistical inferences about the optimal time to target the chain.

I'm not sure I understand the reasoning here.  The second half of the CIP regarding full-pot distribution relies on the fact that delegators and SPOs will receive the rewards that was previously being sent to accounts that have since expired.  This is as a counter-balancing mechanism to help incentivize delegation if the total stake ratio drops too low.  If we send those rewards to the treasury instead, it defeats the purpose and increases the risks related to drops in total stake due to large expirations.

fallen-icarus:
> ... delegators and SPOs will receive the rewards that was previously being sent to accounts that have since expired. This is as a counter-balancing mechanism to help incentivize delegation if the total stake ratio drops too low.

I want to reiterate my opinion that this will not actually incentivize total stake to rise. If a person doesn't care enough about ADA to participate in the effectively effortless staking rewards right now, offering them more ADA likely won't change their behavior.

Cerkoryn:
@WhatisRT @fallen-icarus 

I've gathered good data on the count of wallets and sum of stake that would be affected by different initial settings of `delegatorInactivity`.

The first chart is counting every wallet currently delegated to an active stake pool, regardless of how much ADA is in the wallet:
image


The second chart is the same, but only looking at wallets with >5 ADA in them:
image

> [!WARNING]  
> The two above charts are not entirely correct.  The Koios `/account_txs` endpoint includes inbound-only transactions, which caused me to unintentionally mark some wallets as *active* when they shouldn’t be.  The true numbers are slightly higher than this and posted in a below comment.  -Cerkoryn


It is worth noting that of course there would be 0 stake and 0 wallets affected if we went up to 6 years, because the Shelley hard fork was less than 6 years ago.  However, the longer we wait to delay any kind of CIP like this one, the more stake that will build up in older pools over time and the harder it will be to introduce a proof-of-liveness mechanism without running into security concerns about stake expiration.

Looking at this data, if we choose an initial value in the range of 3-5 years, the *upper limit* of how much total stake that would be affected is 6.2%, 2.7%, and 0.5% of active stake (~21B) respectively.  I agree that the balancing mechanism will probably have a marginal effect at incentivizing new delegations, but based on these numbers I think that any tradeoffs to security that result from stake expirations will also have a similarly marginal effect.

In valuation of USD, Cardano is already `#1` in terms of market cap and `#4` in terms of raw USD valuation when it comes to economic security (see [this](https://x.com/Cerkoryn/status/1955712001625235782) post).  Even if this change were to result in a small 1-2% decrease in total stake we'll still be best-in-class.  And if a corresponding 1-2% increase in ADA/USD valuation accompanies this change due to increased investor confidence, then we will have actually RAISED the cost of an attack.

rphair:
@Cerkoryn I've just read through the review commentary all over again to see if there might be any grounds to progress this in our biweekly meeting cycle.  In my opinion we are still getting all the community ideas & expert review on the table, and the time to conclusively bind all these together into a single strategy for the CIP hasn't come yet.

If that's true, I would have no problem letting this continue just as it is now.  If it ever seems we are stalling on the blackboard forever, the 2 things you might do that could help are:
* if you (or an editor / reviewer that sees a more manageable number of possibilities than I do) would make a checklist of the issues that still need to be decided;
* if you think (and post) that putting this up for Review at a future CIP meeting would help either to settle upon that checklist or to address the items on it.

Cerkoryn:
Have to make a correction on the data.  I'll leave the above post up, but add a disclaimer explaining it.

TLDR, the Koios `/account_txs` endpoint includes transactions that are inbound-only, which was incorrectly marking some wallets as active when they shouldn't be.  After accounting for that, the upper limits for the amount of stake that could be affected by enacting this CIP raised a little bit from our earlier estimate.  Apologies for the error.

Wallets with less than 5 ADA were ignored for these queries:

image

```
=== Results ===
No tx in past 1y: 6,381,362,493 ADA across 583,961 accounts (29.36% of delegated stake in sample)
No tx in past 2y: 3,838,556,704 ADA across 454,497 accounts (17.66% of delegated stake in sample)
No tx in past 3y: 1,966,069,219 ADA across 361,726 accounts (9.04% of delegated stake in sample)
No tx in past 4y: 997,665,752 ADA across 167,816 accounts (4.59% of delegated stake in sample)
No tx in past 5y: 127,644,174 ADA across 2,820 accounts (0.59% of delegated stake in sample)
```

Reminder that 6y is also a viable choice, but it is obviously 0 ADA and 0 wallets affected because the Shelley Hard fork is not that old yet.

Other reminder: All of these numbers are still _upper limits_.  Many of those wallets surely would make a simple transaction and continue earning rewards without issue.

Cerkoryn:
I'll also add that there was another benefit to this CIP that we hadn't yet considered.  The portion of stake above the saturation limit in an oversaturated pool effectively reduces the pool’s rewardable stake and those forfeited rewards are currently sent back to the reserve. 

Under this CIP, when certain parameters change (e.g., raising k or lowering L per CIP-50), this sharpens the incentive to re-delegate quickly and converge to the new equilibrium. If some delegators lag (sticky stake, lost stake, etc.), the rewards they would have earned is redistributed to delegators who _are_ eligible for rewards instead of going back to the reserve.

Additionally, stake over the saturation limit currently still participates in block production even though it earns no additional rewards.  This effect carries over in the CIP so there shouldn't be any security tradeoffs to worry about in regards to this new information.

This effect is now modeled in our tool (previously it was notionally being sent to the reserve in the tool)
[https://spo-incentives.vercel.app/](https://spo-incentives.vercel.app/)

Notably, this requires no change to the CIP.

rphair (commented):
@Cerkoryn I wanted to see how this could be progressed, given it's been stalled for about a month: with editors skeptical about it and the methods & possible side effects challenged by the community.  I was hoping someone would follow what I recommended in https://github.com/cardano-foundation/CIPs/pull/1077#issuecomment-3288668966 but this is generally an editor's job so I've just done it myself.

This outline omits reservations (expressed by @TerminadaPool @fallen-icarus) that I believe were answered.  If those or any other reservations still remain, in anyone's opinion, feel free to emphasise them & I'll add them to this review outline so we can refer to it at meetings without sorting through the long prior comment history:

#### open questions to settle before final review
* [proposed method vs. "'stake account expiration"](https://github.com/cardano-foundation/CIPs/pull/1077#discussion_r2294670173)
  * [possibly resolved here](https://github.com/cardano-foundation/CIPs/pull/1077#issuecomment-3237842494) with https://github.com/cardano-foundation/CIPs/pull/1077/commits/636630c23fd8b63fc95a068182a5be37a1f23350
* [suggested simplification](https://github.com/cardano-foundation/CIPs/pull/1077#pullrequestreview-3153395544)
  * [response begins:](https://github.com/cardano-foundation/CIPs/pull/1077#issuecomment-3225120449)
  * [still pending working group response?](https://github.com/cardano-foundation/CIPs/pull/1077#issuecomment-3225917315)
* [continuity of stake vs. sudden drops](https://github.com/cardano-foundation/CIPs/pull/1077#issuecomment-3233607453)
  * [debate begins, over balancing incentives](https://github.com/cardano-foundation/CIPs/pull/1077#issuecomment-3233731118)
  * [possibly resolved with _participation shocks after large expirations_](https://github.com/cardano-foundation/CIPs/pull/1077#discussion_r2317078000)

@perturbing @Ryun1 @Crypto2099 I would personally be in favour of approving & merging this as a "workable solution" even if the above & any future points are still debated on both sides without _everyone_ being able to agree.

I would trust @Cerkoryn therefore to declare when this is the author's and collaborators' "best effort" to satisfy all community & Ledger feedback (cc @lehins @WhatisRT)... but we owe it to the community to get the best possible rendition so we should try to clear as many as possible of these points first.

Cerkoryn:
@rphair 

> * [proposed method vs. "'stake account expiration"](https://github.com/cardano-foundation/CIPs/pull/1077#discussion_r2294670173)
>   
>   * [possibly resolved here](https://github.com/cardano-foundation/CIPs/pull/1077#issuecomment-3237842494) with [636630c](https://github.com/cardano-foundation/CIPs/commit/636630c23fd8b63fc95a068182a5be37a1f23350)

That is correct.  I believe this was resolved when we switched to using @lehins method of using the last witness on the reward account instead.

> * [suggested simplification](https://github.com/cardano-foundation/CIPs/pull/1077#pullrequestreview-3153395544)
>   
>   * [response begins:](https://github.com/cardano-foundation/CIPs/pull/1077#issuecomment-3225120449)
>   * [still pending working group response?](https://github.com/cardano-foundation/CIPs/pull/1077#issuecomment-3225917315)

Similar to above.  The minor re-write was inspired by this simplification and included the things mentioned here with the exception of returning excess rewards to the treasury (since that forgoes many of the other benefits related to the full-pot distribution part of this CIP)

> * [continuity of stake vs. sudden drops](https://github.com/cardano-foundation/CIPs/pull/1077#issuecomment-3233607453)
>   
>   * [debate begins, over balancing incentives](https://github.com/cardano-foundation/CIPs/pull/1077#issuecomment-3233731118)
>   * [possibly resolved with _participation shocks after large expirations_](https://github.com/cardano-foundation/CIPs/pull/1077#discussion_r2317078000)

I believe this is more or less answered by the data presented in [#issuecomment-3358578912](https://github.com/cardano-foundation/CIPs/pull/1077#issuecomment-3358578912).  This shows the maximum amount of ADA that could be expired in such a way based on the current state of staked ADA. 

From the perspective of myself and others in our working group, we believe that this CIP is in a sufficient state such that it could be merged and presented to the community for a vote (on both the CIP itself and the initial value of `delegatorInactivity`).

Cerkoryn:
@rphair @Ryun1 @Crypto2099 

This one is fixed as well.  It was a much simpler change because the offending charts were able to be cleanly removed from the document without materially affecting anything important.  Also the git history for this one was already clean.

Please let me know if you guys recommend any further changes, thank you.

Cerkoryn:
@rphair 

My apologies, I had overlooked the image files.  Those are now removed as well.

EarnCoinPool:
@lehins 
> 3\. Any action that requires a witness for an account moves expiration for that account into the future.

Can you explain which actions require the witness needed here? Is it any transaction or transaction like withdrawal of rewards?

lehins:
> Can you explain which actions require the witness needed here? Is it any transaction or transaction like withdrawal of rewards?

@EarnCoinPool I am glad you asked :grin: 

There are three levels of ownership indication of an account:
1. Direct proof of ownership, which requires witness (i.e. cryptographic signature of a transaction or execution of a smart contract, depending on what type of account it is). This proof of ownership can be inferred from all of the actions below:

   * withdrawal of rewards
   * delegation to a stake pool and/or DRep
   * unregistration (which is a valid action to take into consideration, since account can always re-register)
   * From Conway era onwards registration certificate, whenever certificate also specifies the deposit amount. Classic registration certificate that was introduced in Shelley era did not require a witness, as such it cannot be used as proof.
   * Submission of registration or an update stake pool certificate for owners listed in stake pool parameters.
3. Indirect indication of ownership, since it does not require a witness for the account, but is closely tight to the account. Spending an output from an address that contains that staking credential. Since this action does not require a witness for that credential, it cannot be used as direct proof. However, because assets are being spent that used to be associated with the account is a good indicator that the wallet that created this staking account is still very much active, because only the owner of that wallet would be able to spent such an output.
4. Indirect hint at potential ownership:
   * There were funds send to an address with a staking credential. This is the weakest of them all, since anyone could send funds to a public address that contains a staking credential. However, it could be used to infer at least some activity. If someone is receiving a million ADA to an address, I would expect they will not be using someone else's staking credential, since they will be loosing money in the amount of rewards that will go to that account.
   * Stake pool registration or update certificate account is listed in the rewards account of stake pool parameters. Witness is not required for using account in that manner, but since funds received by that stake pool will go into that account it can be used as a vague indicator that the account is still active.

EarnCoinPool:
@lehins 

Thank you for your reply, this was very helpful.

Can I ask, if this CIP were to be implemented would all 3 of these be used to update the variable? Or would some be excluded? I assume the first item under 3 should be excluded as anyone could send some tokens to an address to game the system.

Thanks again for your help! 

> 
> There are three levels of ownership indication of an account
>

lehins:
> I assume the first item under 3 should be excluded as anyone could send some tokens to an address to game the system.

@EarnCoinPool Yes, I would definitely agree with that. I mostly included the third point for completeness sake and as something that could be considered as potentially useful information. For example if today one would analyze historical data, it could be useful information to know whether an address that might be considered inactive is still receiving funds. Maybe such account isn't really dead, especially if it is receiving large amounts of ADA :shrug:

PR #1084, opened 2025-09-04
CIP-0176? | Non-segregated Block Body Serialization

teodanciu:
Simplify serialization of transactions in a block

---

([latest version of proposal](https://github.com/teodanciu/CIPs/tree/td/block-serialization/CIP-0176))

### DIFF
--- a/CIP-0176/README.md
+++ b/CIP-0176/README.md
@@ -75,7 +75,7 @@ Note that we propose keeping invalid transaction indices separately, because:
 
 Also, note that `invalid_transactions` indices precede `transactions` in the proposed layout, so that consumers can determine the validity of transactions without having to scan through the entire transaction list.
 
-## Rationale: how does this CIP achieve its goals?
+## Rationale: How does this CIP achieve its goals?
 
 Serializing transactions in sequence directly supports more complex constructs - such as nested transactions - by eliminating the need to coordinate disjointed segments across different levels of structure.
 


CIP-0176/README.md
```diff
@@ -0,0 +1,84 @@
...
+CIP: XXXX
+Title: Non-segregated Block Body Serialization
+Category: Ledger
+Authors:
+  - Teodora Danciu <teodora.danciu@iohk.io>
+  - Alexey Kuleshevich <alexey.kuleshevich@iohk.io>
+Implementors: N/A
+Status: Proposed
+Discussions:
+  - https://github.com/IntersectMBO/cardano-ledger/issues/5046
```
rphair:
```suggestion
  - https://github.com/IntersectMBO/cardano-ledger/issues/5046
  - https://github.com/cardano-foundation/CIPs/pull/1084
```
add `this`

CIP-XXXX/README.md
```diff
@@ -0,0 +1,84 @@
...
+The intent was to enable static validation of a transaction without decoding its witnesses. However, this benefit conflicts with the need for strict field evaluation, which is essential to prevent space leaks caused by retaining the original block bytes for an indeterminate period.
+Moreover, the current segregated layout, if it was used for its intended purpose mentioned above, conflicts with incremental decoding of blocks being received over the wire - for which there is a need.
+Our decoders consume bytes on demand until the entire BlockBody structure is exhausted. As a result, the node decodes the full block - including all witness data - in a single pass. The segregated format therefore provides no practical benefit for partial or streaming validation.
+Moreover, time spent doing deserialization amounts to a small fraction of the whole time spent applying the whole block to the ledger state, even when it is done without validating transactions.
```
rphair: These lines are nicely broken into concepts, but in rendering are folded into a relatively big paragraph... could you add a blank line or two in between, so we get 2 or 3 shorter paragraphs with grouped ideas (for readability's sake)?

rphair (commented):
@teodanciu @lehins - looks good in format & readability; we'll introduce this in `Triage` at the next CIP meeting: https://hackmd.io/@cip-editors/119

CIP-XXXX/README.md
teodanciu: Good point, thank you!

rphair:
@teodanciu please let's avoid force pushing again, as requested here: https://github.com/cardano-foundation/CIPs/blob/master/CIP-0001/README.md#1a-authors-open-a-pull-request

This is a shared resource for commentary & editing, not a mirror of the author's code.  We need to keep track of review points and correlate these with changes to the branch, and force pushing effectively wipes out that history. 🙏

teodanciu:
That makes sense, I apologize. Thank you for the information!

CIP-XXXX/README.md
```diff
@@ -0,0 +1,87 @@
...
+
+## Abstract
+
+We propose changing the CBOR encoding of a block body from a segregated layout to a plain sequence of transactions.
+Current layout: all transaction bodies are concatenated and encoded first, followed by their witness sets, then followed by auxiliary-data hashes, and finally followed by validity flags.
+Proposed layout: each transaction is serialized in full before the next transaction is written to the stream.
+
+## Motivation: why is this CIP necessary?
+
+Segregated serialization of [Nested transactions](https://github.com/cardano-foundation/CIPs/pull/862) would be challenging both to specify and implement.
```
```suggestion
Segregated serialization of [CIP-0118? | Nested Transactions](https://github.com/cardano-foundation/CIPs/pull/862) would be challenging both to specify and implement.
```

rphair:
@teodanciu @lehins today's CIP meeting has not yet confirmed this as a CIP candidate mainly (as I understood from meeting discussion) because we can't verify that other node development teams have endorsed this.  We invited people at the meeting to tag relevant reviewers from those teams & so hopefully that will proceed next.

Including the Nested Transactions already mentioned in the proposal, we also noted that potential changes to the block body make this proposal considerable with these other CIPs (and so tagging authors & some reviewers of these proposals):
* https://github.com/cardano-foundation/CIPs/pull/862 - cc @polinavino @fallen-icarus @WhatisRT
* https://github.com/cardano-foundation/CIPs/pull/964 - cc @aslesarenko @colll78
* [CIP-0150 | Block Data Compression](https://github.com/cardano-foundation/CIPs/blob/master/CIP-0150/README.md) - cc @sierkov @Quantumplation

CIP-0176/README.md
```diff
@@ -0,0 +1,87 @@
...
+
+Segregated serialization of [Nested transactions](https://github.com/cardano-foundation/CIPs/pull/862) would be challenging both to specify and implement.
+Separating and concatenating components across nested and non-nested transactions introduces complexity that is error-prone and potentially inefficient, as it may require tracking offsets and performing additional buffering and copying at runtime.
+
+Currently, segregated serialization also complicates the logic used by Consensus to estimate transaction sizes when selecting transactions to fit within a block. Switching to a format where full transactions are encoded sequentially would simplify this process.
+
+Considering - in hindsight - the original motivation for adopting segregated serialization rather than the more natural and predictable way of serializing a list of transactions - it's unclear whether it provides any real benefit.
+The intent was to enable static validation of a transaction without decoding its witnesses. However, this benefit conflicts with the need for strict field evaluation, which is essential to prevent space leaks caused by retaining the original block bytes for an indeterminate period.
```
Quantumplation, line 31: I was always under the impression that this segmentation was instead so that you could have light clients that, for example, ignored the auxiliary data. and just downloaded the transaction bodies and witnesses.

CIP-0176/README.md
```diff
@@ -0,0 +1,87 @@
...
+  , transactions : [* transaction]
+  , invalid_transactions : [* transaction_index]
+  ]
+
+transaction = [transaction_body, transaction_witness_set, bool, auxiliary_data/ nil]
+```
+
+Note that we propose keeping invalid transaction indices separately, because:
+  * `isValid` flag - which determins validity -  is controlled by the block producing node, not by the the transaction creator
+  * it's more efficient: we serialize indices only for invalid transactions, which are a small minority.
```
Quantumplation, line 74: in fact, "a small minority" understates it; in asymptotically 100% of blocks, this list would be empty and could take up a single byte `80`.

Quantumplation: I wonder how this plays with the proposed changes in the Leios CIP

CIP-0176/README.md
```diff
@@ -0,0 +1,87 @@
...
+
+We propose changing the CBOR encoding of a block body from a segregated layout to a plain sequence of transactions.
+Current layout: all transaction bodies are concatenated and encoded first, followed by their witness sets, then followed by auxiliary-data hashes, and finally followed by validity flags.
+Proposed layout: each transaction is serialized in full before the next transaction is written to the stream.
+
+## Motivation: why is this CIP necessary?
+
+Segregated serialization of [Nested transactions](https://github.com/cardano-foundation/CIPs/pull/862) would be challenging both to specify and implement.
+Separating and concatenating components across nested and non-nested transactions introduces complexity that is error-prone and potentially inefficient, as it may require tracking offsets and performing additional buffering and copying at runtime.
+
```
aslesarenko, line 27: This CIP will also slightly simplify implementation of #964 as it requires each transaction bytes to be hashed. However this is always possible to do regardless of the format, so this CIP is not a strict dependency for #964.

CIP-0176/README.md
lehins: This segregation was done only because Bitcoin did it that way. This is what I've learned from others that were around during that time. From what I know, current structure is not utilized by anyone out there and it is definitely not utilized by the node.

It is absolutely possible to download transactions without the parts that you don't care about, regardless how transaction is serialized. Naturally this would prevent any sort of validation, but presumably that would not be a problem for light clients that trust the data they are downloading.

FTR. dropping auxiliary data prevents transaction validation, since its hash is in the body.

CIP-0176/README.md
lehins: > in asymptotically 100% of blocks, this list would be empty and could take up a single byte 80.

That is why value form `isValid` itself is not serialized, otherwise that list would always be as long as number of transactions.

FYI. There is also [a plan to remove `isValid` from transactions](https://github.com/cardano-foundation/CIPs/pull/1089), so those indices for `isValid=Flaslse` has to stay as a separate list.

> I wonder how this plays with the proposed changes in the Leios CIP

Leios does not care about how transactions are serialized in a block.
In the matter of fact, we've discussed with @nfrisby how this change could potentially be useful for speeding up mempool, which would be useful for Leios. Although this idea need to be properly experimented with.

CIP-0176/README.md
Quantumplation: You could still validate the transaction body; because only hash is in the body. You don't need to know what the preimage of the hash is to validate that the ledger state transition represented by / caused by the transaction body is valid.

What you can't validate is the auxiliary data, which is fine, because you didn't download it anyway, and its contents can't change the validity of the state transition to the ledger.

Maybe splitting hairs semantically about what "validating" encompasses, but the point is that there is a world where you can download just the tx bodies, validate them with respect to a ledger, without having to download the potentially much larger auxiliary data.

Whether that is useful / valuable to preserve is another question entirely :) just clarifying what I meant.

lehins: > You could still validate the transaction body; because only hash is in the body. You don't need to know what the preimage of the hash is to validate that the ledger state transition represented by / caused by the transaction body is valid.

Same applies to all of the witnesses. You don't need scripts or redeemers or signatures if you just want to reconstruct the ledger state.

> Maybe splitting hairs semantically about what "validating" encompasses, but the point is that there is a world where you can download just the tx bodies, validate them with respect to a ledger, without having to download the potentially much larger auxiliary data.

You cannot download just bodies and validate them with respect to the ledger! You either need a full transaction or you don't do any validation. You can't just be picking and choosing what you want to validate. You either trust the full block/transaction and skip validation altogether or you don't trust any of it and verify it yourself using the ledger logic.
Anyone who has a different opinion on that matter can do whatever the hell they want, but we do not need to care or support their use case.

> there is a world where you can download just the tx bodies

Even if you don't need to do validation this is not exactly true, since you also need `isValid` flag, otherwise you don't know which inputs have been spent or outputs created.

Moreover, with introduction of nested transactions, you will no longer be able to download "just the transaction bodies", since those transaction bodies will contain subtransactions with their own witnesses and their own auxiliary data, unless we also split them out into a separate part of the block.

My point is that, it is always possible to download transactions with any information stripped out from transactions, regardless of how they were serialized. In the matter of fact db-sync does just that and more. All one needs to do is discard the bytes that are not relevant for the end user (setting auxiliary data, witnesses, etc. to empty before transmission of transactions to the light client)

@Quantumplation In any case, your arguments are about hypothetical use cases that someone somewhere might or might not implement at some point. While my argument in favor of this CIP is that it will facilitate a safer and less complex implementation of Nested Transactions.
Imagine what we (and others who will need the full transaction from the block) would need to do if we are to keep segregated witnesses in presence of sub-transacitions! Today's logic for [combininng](https://github.com/IntersectMBO/cardano-ledger/blob/a703b69c0f4bb43fc1b190aed6e3dde99777bc54/eras/alonzo/impl/src/Cardano/Ledger/Alonzo/BlockBody/Internal.hs#L222-L259) and [splitting](https://github.com/IntersectMBO/cardano-ledger/blob/a703b69c0f4bb43fc1b190aed6e3dde99777bc54/eras/alonzo/impl/src/Cardano/Ledger/Alonzo/BlockBody/Internal.hs#L125-L212) is already a hairy mess and I do not want to make it worse. If we are to add another level to this monstrosity, there will be a much higher potential for bugs to be introduced, be it in ledger or in any other tool or node that has to deal with constructing full transactions.

So, I appreciate your concern, but unless we find someone that actually relies on the current structure, such speculative arguments do not carry much weight IMHO.

Quantumplation: I'm not making arguments either way, for what it's worth.

From my perspective, the stance that it was done this way "because bitcoin did it that way" is very incongruous with the thought and care that went into many other decisions of the early ledger. Usually things that seem arbitrary have at least some well thought out reasoning, if outdated or prematurely optimizing.

I had some recollection that Duncan Coutts had told me what that reason was at one point. I may certainly be mistaken on that, but I'm just sharing in case it's helpful.

Beyond making sure that that is actually understood, I don't have any particular opinion on this proposal (in fact, I'm slightly in favor!)

Since I'm still not 100% sure you understand the point I'm making, given some of your response, I'll clarify again, but this is not to make an argument that we should or should not change it now. I'm actually quite in favor of simplifying things since the benefit as I understood it never did materialize much.

> Same applies to all of the witnesses. You don't need scripts or redeemers or signatures if you just want to reconstruct the ledger state.

> You cannot download just bodies and validate them with respect to the ledger! You either need a full transaction or you don't do any validation. You can't just be picking and choosing what you want to validate.

This is not true. Scripts, Witnesses, and redeemers are fundamentally different with regards to the validity and trust assumptions compared to the auxilary data like transaction metadata.

If I don't ever need to look at the auxiliary metadata, and never depend on it for business logic, then I can get a fully trustless node by validating block headers and the transaction body + witness set. Regardless of what the preimage was, someone can't present me with fallacious data, because I'm not downloading the preimage, and consensus doesn't consider the auxiliary data for selecting a chain.

On the other hand, if I avoid downloading the witness set, then I have a fundamentally different trust model; I don't know the redeemers to apply to the scripts to even run them; I don't know whether the transaction is even authorized to spend the UTxOs.

There are ends of a spectrum, of course: I can have a trust-full node by ignoring the witness set, of course, and at the end of the day a full node needs to download the auxiliary data to make it available, I'm not arguing that that spectrum doesn't exist.

But I still maintain that there is a fundamental difference between these things; the body lets you trust-fully reproduce the ledger state; the block headers + tx bodies + witnesses let you trust-lessly validate and reproduce the ledger state; the block headers + tx bodies + witness + auxiliary data lets you fully replicate and serve all historical state, including non-ledger state like transaction metadata.

Again, I'm just sharing this as another point of data for what I thought someone had told me the motivation was, in case that shook forth additional context / memories / motivation for the people like yourself; I may certainly be misremembering, and I quite agree that it's likely not very relevant in todays landscape.

Beyond making sure that I'm not being misunderstood, I don't have a dog in this race on the proposal itself.

lehins: > I had some recollection that Duncan Coutts

Duncan was the person who told me that it was done that way "because bitcoin did it that way".

I get what you are saying, but without auxiliary data it is not only the hash that is not verifiable, but you also cannot validate the size of the transaction, thus you cannot validate the fees, or the correctness of collateral amount. So, a lot of validation goes out of window in the ledger rules if you don't have the full thing.
I understand your thought process from the perspective of DApp developer, you only care about general transaction validity and ability to verify scripts. That is, however, only one perspective. And your view on the level of validity of any transaction with missing auxiliary data is very much flawed. And that is the point I am trying to bring across.

lehins: Glad to hear though that you are in favor of this simplification :wink:

Quantumplation: Ah, I was missing tx size, that's indeed a big factor of it 😅

CIP-XXXX/README.md
```diff
@@ -0,0 +1,87 @@
...
+  ]
+```
+
+The proposal is to change it to:
+```cddl
+block =
+  [ header
+  , transactions : [* transaction]
+  , invalid_transactions : [* transaction_index]
+  ]
```
lehins, line 64: We should make a concrete distinction what the block body is. This will also be useful for Peras and Leios, since we will be adding their respective certificates into the block body:
```suggestion
  , block_body
  ]

block_body =
  [ transactions : [* transaction]
  , invalid_transactions : [* transaction_index]
  ]
```

CIP-0176/README.md
```diff
@@ -0,0 +1,87 @@
...
+```
+
+Note that we propose keeping invalid transaction indices separately, because:
+  * `isValid` flag - which determins validity -  is controlled by the block producing node, not by the the transaction creator
+  * it's more efficient: we serialize indices only for invalid transactions, which are a small minority.
+
+## Rationale: how does this CIP achieve its goals?
+
+Serializing transactions in sequence directly supports more complex constructs - such as nested transactions - by eliminating the need to coordinate disjointed segments across different levels of structure.
+
```
Ryun1, line 81:
```suggestion

### Ecosystem Impact

```
It may be worth discussing the potential downstream impacts of this change on existing tooling

rphair:
@teodanciu @lehins it's been suggested at today's CIP meeting (where this came up on a list of non-progressed proposals) that this could be done at an upcoming inter-era hard fork & the Haskell Ledger team might have this as a planned feature anyway... could we get this updated in the meantime to be sure the CIP itself is accurate by such a time?  Will leave `Waiting for Author` in the meantime.

CIP-0176/README.md
lehins: This change would only affect tooling that deals with (de)serialization of blocks (such as alternative nodes and indexers), which is a minority when looking at all of the Cardano related tooling. Those tools that do deal with block serializaion, will be affected in a positive manner by changes proposed in this CIP, since if we don't make this change, nested transactions will make serialization of a block incredibly complex and error prone.

With respect to any tooling that deals with transaction building/serialization/submission will not be impacted in any way by this CIP.

lehins:
> this could be done at an upcoming inter-era hard fork & the Haskell Ledger team might have this as a planned feature anyway...

@rphair No, unfortunately, this CIP cannot be implemented at an intra-era hard fork, since it introduces changes to serialization. This is however planned for Dijkstra era, since if we don't implement this CIP for Dikstra era, Nested Transaction implementation will make current approach of transactions serialization in a block incomprehensible.

CIP-XXXX/README.md
```diff
@@ -0,0 +1,87 @@
...
+
+transaction = [transaction_body, transaction_witness_set, bool, auxiliary_data/ nil]
+```
+
+Note that we propose keeping invalid transaction indices separately, because:
+  * `isValid` flag - which determins validity -  is controlled by the block producing node, not by the the transaction creator
```
ch1bo:
```suggestion
  * `isValid` flag, which determines validity, is controlled by the block producing node, not by the the transaction creator
```

CIP-0176/README.md
```diff
@@ -0,0 +1,87 @@
...
+### Acceptance Criteria
+
+- [ ] Block serializers and deserializers in [cardano-ledger](https://github.com/IntersectMBO/cardano-ledger) are implemented such that they follow the cddl specification described above, and reflected in the cddl specs
+- [ ] The feature is integrated into [cardano-node](https://github.com/IntersectMBO/cardano-node) and released as part of the Dijkstra era hard fork
+
+### Implementation Plan
+
+The implementation of this CIP should not proceed without an assessment of the potential impact on all the components that deserialise blocks.
+Leios and Peras R&D teams should also be aware of these changes.
```
ch1bo, line 107: Leios awareness can be assumed now as we discovered this while discussing validity flags for EBs in [this issue](https://github.com/input-output-hk/ouroboros-leios/issues/718)

CIP-0176/README.md
```diff
@@ -0,0 +1,87 @@
...
+  , transactions : [* transaction]
+  , invalid_transactions : [* transaction_index]
+  ]
+
+transaction = [transaction_body, transaction_witness_set, bool, auxiliary_data/ nil]
+```
+
+Note that we propose keeping invalid transaction indices separately, because:
+  * `isValid` flag - which determins validity -  is controlled by the block producing node, not by the the transaction creator
+  * it's more efficient: we serialize indices only for invalid transactions, which are a small minority.
```
ch1bo, line 74: Isn't this against the motivation of the proposal?

I understood the motivation to be simplicity, getting rid of premature optimization that did not pay off, and that streaming block data was not possible with the segregated parts. By suffixing all the validity information to the end of the block body, every consumer needs to seek through the 90k or so bytes to find out whether the tx at index 0 was valid or not. If we'd be prefixing the validity index, that might be a different story. But if we would really lean into the "hard to get wrong" argument of simplicity, we might be open to accept some inefficiency?

Especially, as using an array of 16bit indices is not the most compact way to serialize invalid indices. The most compact way I could think off would be compressed bitfields (e.g. run-length encoded) .. however, those are definitely non-standard in CBOR and raises the question:

How much inefficiency is acceptable at the cost of simplicity?

CIP-XXXX/README.md
rphair: Fixed a little differently in https://github.com/cardano-foundation/CIPs/pull/1084/commits/5e386f38ab3562b57144c06288e340fca9dc0ad4 since we also need "definite article" at beginning and punctuation at end (so all bulleted items read as a single sentence).

rphair:
@teodanciu I think we're just waiting for you to progress systematically through all unaddressed comments in the reviews above before removing the `Waiting for Author` tag & getting this on a steady review track again.

CIP-0176/README.md
teodanciu: Added the section,  with the description of @lehins , thank you.

CIP-0176/README.md
teodanciu: I don’t think this is at odds with the motivation of the proposal, which is to remove a segmentation scheme that turned out not to pay off. Doing so doesn’t imply that every remaining structure-aware choice should also be eliminated.

I agree that including an `isValid` flag directly in the transaction is simpler from the perspective of consuming transactions. At the same time, keeping the transaction serialization identical between the mempool and the block (i.e. no validity flag on the transaction itself) is also a form of simplicity, since it avoids having two representations of the same transaction depending on context.

Furthermore, [as noted above](https://github.com/cardano-foundation/CIPs/pull/1084/files#r2353294618), asymptotically all transactions are valid. Inlining the validity flag would therefore add hundreds of extra bytes per block (e.g. ~100 bytes for 100 transactions per block), paid by every node, despite the information being almost always redundant.

CIP-0176/README.md
teodanciu: Thank you, added the note under `Ecosystem Impact` section.

CIP-0176/README.md
lehins: > By suffixing all the validity information to the end of the block body, every consumer needs to seek through the 90k or so bytes to find out whether the tx at index 0 was valid or not. If we'd be prefixing the validity index, that might be a different story.

This will probably not have a big impact in practice, since it is rare that someone needs a single transaction, without ever looking at other transactions in a block. However, it doesn't mean that we should prevent such use case. I don't see any benefits whatsoever for having isValid flags at the end of the block body vs at the top. So, I am definitely in favor of moving the list of indices before the transactions. @teodanciu do you mind making this adjustment to CDDL and pointing this aspect out:
```cddl
block_body =
  [ invalid_transactions : [* transaction_index]
  , transactions : [* transaction]
  ]
```

> Especially, as using an array of 16bit indices is not the most compact way to serialize invalid indices. The most compact way I could think off would be compressed bitfields (e.g. run-length encoded) .. however, those are definitely non-standard in CBOR and raises the question:

There are a couple points here that I do not agree with:
* Simplicity. First and for most, it is not the submitting user that supplies the `isValid` flag, but the node that builds the block is the entity that sets this flag. Because of this `isValid` is not even part of the binary format of a transaction any more (see #1089). So, I totally agree with @teodanciu, it would be useful to have the same binary representation for transaction submission protocol as well as its representation in the block, since that approach results in the most overall simplicity
* Compactness:
  * I don't think using run length encoding would be the most compact way to represent this information in a block. Most importantly, I think such micro-optimization would introduce a totally unnecessary complexity and easy source of dangerous bugs. Moreover, as @teodanciu already pointed out, savings would be negligible if present at all, since 99.9% of transactions have `isValid = True`, which means for most blocks it is just an empty array. If anything a bit vector would be even more efficient way to represent this, since most of the time it would be just 0, but even that is not worth the complexity, despite having simpler and more efficient runtime and memory implementation.
  * Not that it is relevant, but just to have our facts straight this is not quite accurate "using an array of 16bit indices". It is not accurate because CBOR does not encode integral values in fixed width, so the overhead of each integral element in a CBOR array could be 1 byte for small values, even if the maximum accepted size is much higher like 64bits or even infinite.


> How much inefficiency is acceptable at the cost of simplicity?

I can measure inefficiency. It is only if you give me a metric on simplicity and its mathematical relationship to inefficiency that I will be able to answer that question :wink:

CIP-0176/README.md
teodanciu: Resolving, since - albeit an interesting conversation, there doesn't seem to be any action required by it.

CIP-0176/README.md
teodanciu: The change suggested by @lehins is already implemented, so I updated the CIP to reflect this. 
Resolving since there doesn't seem to be more action required by this conversation.

CIP-0176/README.md
teodanciu: Thank you for the note, resolving the comment.

CIP-0176/README.md
teodanciu: @lehins  Updated the cddl to invert `transactions` and `invalid_transactions` order, thank you.

rphair: @lehins since we intend to use resolved comments as a means of moving this CIP forward, please post here if you think https://github.com/cardano-foundation/CIPs/pull/1084/commits/06015ab9fdf7edeea94f268b4a03008b54335c8a resolves this conversation and/or resolve it yourself 🙏

ch1bo: Thanks for the responses both. I don't have a very strong opinion here and just wanted to mention my interpretation of a potential mismatch between motivation and specification, or something that could be improved even further on.

I want to clarify two things:
1. I was not suggesting to change the representation of the `transaction`, but wanted to discuss potential prefixing of it with that validity info. i.e.
  ```
  block_body =
    [* validated_transaction]
  validated_transaction = [bool, transaction]
  ```
2. I thought that `invalid_transactions : [* transaction_index]` would be a list of 16bit indices because of [this cddl definition](https://github.com/IntersectMBO/cardano-ledger/blob/144b207a7bf3981db96ca87c6ad800265c04d8e2/eras/conway/impl/cddl/data/conway.cddl#L731C1-L731C18) in the `cardano-ledger`. I'm not fluent in CDDL, but thought this would correspond to a 2 byte unsigned integer.

---

Moving the `invalid_transactions` array to the beginning is a good way forward and emphasizes that it shall not be ignored by anyone processing the blocks directly. It probably is the sweet spot between "keeping it simple" and not wasting a byte for 99.99% of transactions.

lehins: @rphair I am totally fine with resolving this thread. Sorry, I don't have permission to do it myself, since it was @ch1bo who created the thread. Reading @ch1bo's last comment makes me believe we are all in agreement. :+1:

CIP-XXXX/README.md
```diff
@@ -0,0 +1,111 @@
+---
+CIP: XXXX
```
```suggestion
CIP: 176
```

teodanciu:
Apologies, I missed this last comment. I renamed the folder and updated the description.

CIP-0176/README.md
```diff
@@ -0,0 +1,111 @@
...
+Title: Non-segregated Block Body Serialization
+Category: Ledger
+Authors:
+  - Teodora Danciu <teodora.danciu@iohk.io>
+  - Alexey Kuleshevich <alexey.kuleshevich@iohk.io>
+Implementors: N/A
+Status: Proposed
+Discussions:
+  - https://github.com/IntersectMBO/cardano-ledger/issues/5046
+  - https://github.com/cardano-foundation/CIPs/pull/1084
```
rphair:
```suggestion
Discussions:
  - Ledger issue: https://github.com/IntersectMBO/cardano-ledger/issues/5046
  - Original PR: https://github.com/cardano-foundation/CIPs/pull/1084
```
(an old feature of these YAML lists that has been [recently: since end of last year] universally recommended to produce more user-friendly lists of discussion links)

CIP-0176/README.md
```diff
@@ -0,0 +1,111 @@
...
+License: CC-BY-4.0
+---
+
+## Abstract
+
+We propose changing the CBOR encoding of a block body from a segregated layout to a plain sequence of transactions.
+Current layout: all transaction bodies are concatenated and encoded first, followed by their witness sets, then followed by auxiliary-data hashes, and finally followed by validity flags.
+Proposed layout: each transaction is serialized in full before the next transaction is written to the stream.
+
+## Motivation: why is this CIP necessary?
```
rphair:
```suggestion
## Motivation: Why is this CIP necessary?
```
1 of 2: another template change in the meantime (grammatically & aesthetically correct)

CIP-0176/README.md
```diff
@@ -0,0 +1,111 @@
...
+  ]
+```
+
+Note that we propose keeping invalid transaction indices separately, because:
+  * the `isValid` flag - which determines validity -  is controlled by the block producing node, not by the transaction creator;
+  * it's more efficient: we serialize indices only for invalid transactions, which are a small minority.
+
+Also, note that `invalid_transactions` indices precede `transactions` in the proposed layout, so that consumers can determine the validity of transactions without having to scan through the entire transaction list.
+
+## Rationale: how does this CIP achieve its goals?
```
rphair:
```suggestion
## Rationale: How does this CIP achieve its goals?
```
2 of 2: another template change in the meantime (grammatically & aesthetically correct)

PR #1086, opened 2025-09-15
CIP-0023 | Fair Min Fees

Cerkoryn:
2025 update to lean into new governance processes for the implementation plan and for parameter selection.

Additionally renamed `minPoolRate` to `minMargin` since that is the term that more people seem to be familiar with.


shawnim:
Thanks @Cerkoryn ! I appreciate you taking the time to update the CIP and your work in this area. I updated the assumptions in the fee calculator to what I could find as best estimates but if anyone has more accurate numbers, that would be great.

CIP-0023/README.md
```diff
@@ -144,14 +150,33 @@ See the minfees.php code to test different potential values of the parameters.
 
 ### Acceptance Criteria
 
-- [ ] The new parameter `minPoolRate` is implemented in the protocol and enacted through a hard fork.
-- [ ] The minimum value of the existing parameter `minPoolCost` is adjusted in the protocol parameters (not requiring a hard fork).
-- [ ] Code must be written in the cardano-node to check the provided variable rate when creating pool certificates to make sure it meets the new requirements.
-  - [ ] If backward compatibility option 1 is chosen it would require some new code in the cardano-node to treat the below minimum rate as the minimum rate.
+- CIP Editor Approval – Cardano CIP Editors must confirm that the specification is complete, unambiguous, and internally consistent with existing CIPs.
```
rphair:
```suggestion
```
This is implicit for any CIP or CPS because no pull request will be merged unless all these things have been considered.

shawnim:
Hi @rphair !
I'm happy to see this topic getting renewed interest.
It's fine to add @Cerkoryn as an author.
I will try to join the meeting tomorrow morning.
I saw the agenda on hackmd.io but I don't know how to join.
Can you provide instructions on how to join?

rphair:
@shawnim I can see you're on the CIP Discord: you'll see all CIP meetings on the sidebar Events tab ([today's event link](https://discord.com/events/971785110770831360/1370460084454293625/1417540534272000000)).

rphair:
> @shawnim: It's fine to add @Cerkoryn as an author.

@Cerkoryn I tried to submit a PR to your branch adding you as an author, but GitHub said "the file cannot be edited" ... it only allowed me to edit the file in your branch directly. 🤪

This is due to the substantial nature of the updates in this case.  I won't object if you retract your name but it makes editorial sense to have your name on the authors list for reasons of accountability & sustainability for this CIP.

Ryun1 (approved):
really nice work 💪

rphair (approved):
These changes could only make the proposal even more applicable & practical than the original.

rphair:
Based on today's CIP meeting consensus (thanks @shawnim @Cerkoryn for joining), we're changing to `Confirmed` status in recognition of this as valid update & allowing some time for the authors to
* updated the reward calculation based on the possibilities considered for parameter adjustment & backward compatibility
* perhaps adjusting the text of the proposal to reflect the discussion today of staged migration & backward compatibility (e.g. setting the minimum percentage to zero and/or applying a "min function" to out-of-range pool parameter settings).

@Cerkoryn when you can confirm the text is again the way you want it please say so and we'll try to get the remaining editors to apply their own feedback & perhaps tag `Last Check` if not ready to merge outright.

CIP-0023/README.md
```diff
@@ -28,21 +29,28 @@ This creates a more fair marketplace for all stakepool operators and increases d
...
-If option 1 is possible, it would provide a better user experience for both stakepool operators and delegators and would cause less disruption to the network.
+To maintain compatibility for existing pool certificates whose current margin is below the new `minMargin`, the rewards calculation logic should apply a `min` function so that the protocol parameter `minMargin` overwrites any individual pool margin that is smaller than the chosen value. This lets legacy pool certificates remain valid while ensuring the ledger enforces the new minimum fee during reward distribution.
+
+It is also recommended to introducce the hard-fork with `minMargin` initially set to `0`. Doing so minimizes migration friction for stake pool operators and gives governance time to raise the parameter to its target value through the normal paramater change governance action process.
+
+Should this clamping approach prove infeasible, pool certificates with a margin lower than `minMargin` would need to be re-registered with compliant values, but the goal is to avoid disruption as much as possible.
```
Cerkoryn: @rphair @shawnim 

I've updated the backwards compatibility section to reflect what we talked about during the CIP editors meeting.

Please let me know if I missed anything.

rphair: @Cerkoryn that explains the proposition & considerations well enough that I think the CIP can be accepted with these — even though there could be other considerations involved (though I think the Path to Active is written completely & broadly enough to accommodate these).

rphair:
With https://github.com/cardano-foundation/CIPs/pull/1086/commits/1cfe7136880ddf73f948ae43c15b99056ad2243c and https://github.com/cardano-foundation/CIPs/pull/1086#discussion_r2355525366 I think we've addressed all feedback from the CIP meeting, and that we've also upheld the original view that this update could only be an improvement regardless of the governance & Ledger decisions that would follow.

Therefore I'm marking `Last Check` to ensure this update doesn't get forgotten (since we have no way to automatically progress PRs tagged `Update` from the agenda otherwise) although this should not rush other editors through their own opportunity to review this to their hearts' content.

shawnim:
Looks good to me with perhaps one minor clarification:
"...the rewards calculation logic should apply a `min` function..."
might change to
"...the rewards calculation logic should apply an `ensureMinMargin` function..."
to avoid any possible confusion with the math function min() which would have a reversed effect.

rphair:
... yes: technically I think it would be a `max` function 😅

Cerkoryn:
@shawnim @rphair 

Good catch, I think I'd written that poorly.  I just pushed an update to hopefully clarify the language even further.  Please let me know if this is sufficient:

> To maintain compatibility for existing pool certificates whose current margin is below the new `minMargin`, the ledger's rewards calculation should treat the protocol parameter `minMargin` as the effective margin for those pools. In other words, if a pool's margin is less than `minMargin`, the protocol-level `minMargin` overrides the pool's registered `margin` during reward calculation. This minimizes disruption and lets legacy pool certificates remain valid while ensuring the ledger enforces the new minimum fee during reward distribution.

shawnim:
Thanks @Cerkoryn !
I agree with @rphair that your new description is clear.

Two more thoughts:

1 Should the name of the parameter by minPoolMargin so it goes with the existing minPoolCost and is distinguishable from any other (potentially future) parameters that are also minimum margin values? If you both think it is fine to stay minMargin, I am ok with that too.

2 Does anyone know how the current implementation would handle minPoolCost if in the future it wa raised, creating the same situation we now have with pool certificates having invalid values? Seems like both situations should be handled in the same way but not sure if that belongs in this CIP.

rphair:
`minPoolMargin` was routinely suggested in discussion of CIP-0023 over the years, so much that I was focusing on the other wording & technical changes without paying attention to the naming issue... so thanks @shawnim, indeed we had missed this.

Also, we incidentally lost the positioning of these terms side-by-side when the [initial table in _Specification_](https://github.com/cardano-foundation/CIPs/tree/master/CIP-0023#specification) was removed that would have made the asymmetry obvious to everyone. 😅 So yes of course @Cerkoryn let's ensure both terms are "namespaced" with `minPool`.

BTW I think consideration of namespacing is limited here, because the top-level domain of these parameters would appear to be `min` — rather than ideally using `pool`, since some future scenario might set a `maxPoolCost` — but I believe this is the best we can do under compatibility constraints.

Cerkoryn:
@shawnim @rphair that wording makes sense to me.  I just updated the CIP to use `minPoolMargin` as the new parameter name.

While we're on the topic of parameter names, this is a bit of a side-line but I think `minPoolCost` is badly named, since it doesn't actually represent a pool's cost (which varies based on MANY different factors).  It simply sets a minimum per-epoch fixed fee.  So something like `minPoolFixedFee` would be much more accurate of a description.

Nonetheless, `minPoolMargin` makes perfect sense.

rphair:
I think `minPoolCost` is OK since it represents a "cost" to the set of delegators vs. the rewards they would otherwise collectively earn... and lots of things in economics considered "cost" from one point of view will be characterised differently from others. 😎

perturbing (approved):

rphair:
@Crypto2099 this has only been out for a week so I don't want to put it through without your opinion.  Anything you would add to the changes (or not)?

rphair:
@Cerkoryn as you observed at the CIP meeting today (# 120) we had to postpone our `Last Check`s until next meeting (https://hackmd.io/@cip-editors/121) due to not having editor quorum at meeting (only 1 editor present) and the massive turn-out for emergency Leios review.

@Ryun1 @Crypto2099 @perturbing for these 2 proposals I'd offer to merge this in the in-between period if you can provide approval and if there are now objections... otherwise it's on to `Last Check` as planned in 2 weeks' time.

WhatisRT:
This CIP now has the proposed value for the minimum cost removed, but still advocates for a 'lower' value. I don't think this is actionable - if a concrete number isn't a good thing to put in a CIP then maybe proposing a parameter change is generally not a good thing to have as the contents of a CIP?

Adding `minPoolMargin` is a relatively small change on the ledger side of things, but it's not clear to me why you would want this.

In general, this seems to be mostly an incentives issue. Incentives design is quite difficult, and they shouldn't be touched without getting the opinion of the researchers.

rphair (commented):
@Cerkoryn at first I couldn't find the source for the last observation above:
> This CIP now has the proposed value for the minimum cost removed, but still advocates for a 'lower' value.

... since I didn't recall it from our review of the proposed changes.  Then I found it in the _Motivation_, which hasn't yet been changed in this PR: apparently an artefact of the original CIP (cc @shawnim) which doesn't fit the updated specification; especially with its current statement consistent with review consensus:
> This parameter \[`minPoolMargin`\] is distinct from the existing `minPoolCost`, which represents the minimum fixed fee a pool can set. Both limits are enforced independently by the ledger. ... This CIP does not prescribe specific values for either parameter.

But I can't propose a correction to material unaffected by the PR, and don't want to commit a change directly since it's a decision for the co-authors — after removing the text "Reducing the minimum fixed pool fee" — to adjust the motivation / incentives accordingly: because as @lehins says some of these can't be substantiated (e.g. "reduces the imbalance between stakepools with less or more stake to a more reasonable range").

(BTW "stake pools" is 2 words so that can be fixed while we're at it.)

I still think only this 1 further update is needed for a proposal that's otherwise ready to merge.  I'll reapply my favourable review as soon as this recommendation is committed somehow... and since this is a simple, logical change we can keep it at `Last Check` in the meantime.

Cerkoryn:
> In general, this seems to be mostly an incentives issue. Incentives design is quite difficult, and they shouldn't be touched without getting the opinion of the researchers.

@WhatisRT Our Incentives Working Group that (re)proposed this CIP performs independent research on precisely those difficult incentives questions.  Unfortunately the original RSS paper and related Shelley Design/Delegation specifications are very sparse in terms of documentation for the existing `minPoolCost` parameter and it has been a hotly debated topic amongst the community for years.

@rphair I just pushed the change to that section and corrected the spelling of `stakepool` to `stake pool` throughout the document.

shawnim:
Thanks @WhatisRT @Cerkoryn and @rphair !
I appreciate the efforts to clarify and move this forward.

PR #1087, opened 2025-09-15
CIP-0166? | Efficient scalars for BLS12-381

euonymos:
While working on the [Hydrozoa](https://github.com/cardano-hydrozoa/hydrozoa) project, we decided to use pair-based accumulators as part of ZKifing proofs of membership for the L2 utxo set in the rule-based regime. We are pretty happy with the results, since KZG commitments are easier to work with than Merkle tree-like structures we wanted to use previously.

Operations over scalars for BLS12-381 in Plutus _might_ require some improvements; probably it will become the next bottleneck for this particular use case after MSM makes it to Plutus, as our benchmarks suggest.

By drafting this CIP, I mostly wanted to share those results with the community and establish a place for a potential discussion on the topic, since the current state of things feels a bit rough around the edges, at least it questions `Scalar` module in Aiken and other frameworks (initially we ported it to Scalus, but now we likely will use unbounded integers as this approach proves to be less expensive).

I am happy to hear from everyone who has a take or related ideas. Thanks.

Link for reading current version: [here](https://github.com/euonymos/CIPs/tree/bls-efficient-scalars/CIP-0166)


jmagan:
I think this CIP will add considerable value to the ZK capabilities in Cardano L1. The rationale is correct and spot-on regarding one of the biggest use cases where we encounter these Plutus/Plinth limitations. It's quite unfortunate that we almost exhaust the Plutus execution budget with just a few elements using bilinear accumulators, and it's even worse to consider that most of the budget is spent on unoptimized 'simple' multiplication and addition operations on Fr.

I have a couple of observations:

- The introduction of a new type to represent Fr could be avoided since we currently abstract it using Integer in other BLS operations. This is already mentioned in the CIP, and I think it would be more flexible to handle the conversion in the binding layer.
- As @euonymos mentioned, I would add all the Fr-related operations natively supported in blst. These could be useful later for creating algorithms in embedded curves like JubJub, namely inversion.

CIP-?/README.md
```diff
@@ -0,0 +1,433 @@
+---
+CIP: ?
+Title: Efficient scalars for BLS12-381
+Status: Proposed
+Category: Plutus
+Authors:
+    - Ilia Rodionov <hey@euony.me>
+Implementors: []
+Discussions:
+    - https://github.com/cardano-foundation/CIPs/pull/?
```
rphair:
```suggestion
    - https://github.com/cardano-foundation/CIPs/pull/1087
```
`this`

CIP-0166/README.md
```diff
@@ -0,0 +1,433 @@
...
+the relative share of resources spent on operations over the scalars 
+might become even bigger.
+
+In practice, some cryptographic protocols need to perform quite extensive arithmetic
+over the _scalar field_ $F_r$, particularly when working with _polynomials in KZG commitments_.
+This CIP presents a motivational example, discusses related benchmarks,
+and considers different ways of evolving Plutus 
+toward efficient implementation of scalars for BLS12-381 curve.
+
+## Motivation: why is this CIP necessary?
```
rphair: Most of the subsections her actually belong in the _Rationale_ section, in which implementation details and comparative approaches are considered.  This includes performance considerations for what you have simply described _the **need** for doing_ here in the _Motivation_ section.

CIP-0166/README.md
```diff
@@ -0,0 +1,433 @@
...
+
+Not surprisingly, the first benchmark gives the best result of `1,67B cpu / 5,19 M mem`:
+only addition and multiplication are used, and although costs for both operations depend on the size of arguments,
+even over integers up to _10k-bit_ length, the total is relatively cheap:
+
+```aiignore
+ (a, 1) final_poly_int
+
+   memory units                                           cpu units
+   ⡁⠈⠀⠁⠈⠀⠁⠈⠀⠁⠈⠀⠁⠈⠀⠁⠈⠀⠁⠈⠀⠁⠈⠀⠁⠈⠀⠁⠈⠀⠁⠈⠀⠁⠈⠀⠁⠈⢀⠕⡁ 5189564.0    ⡁⠈⠀⠁⠈⠀⠁⠈⠀⠁⠈⠀⠁⠈⠀⠁⠈⠀⠁⠈⠀⠁⠈⠀⠁⠈⠀⠁⠈⠀⠁⠈⠀⠁⠈⠀⠁⠈⢀⠕⡁ 1669321088.0
```
rphair, line 139: I think it would serve this CIP & the repository well, by assuring portability / readability / responsiveness, if you could implement these graphs in Mermaid (I think the same set of points that produced the text renderings can be applied to a graph).  And if & when someday it _is_ possible for an AI to interpret these graphs, that would make it more achievable.

rphair (commented):
Tagging `Triage` to introduce at the CIP meeting in a few hours: https://hackmd.io/@cip-editors/119

@euonymos it's nice to see that Plutus expert review has already started & as usual we would be happy to hear from @zliu41 @effectfully @kwxm and others about this.

rphair:
@euonymos we decided to leave this in `Triage` because this was submitted shortly before today's CIP meeting and therefore to allow time for the already tagged Plutus reviewers to contribute their assessment about whether this change is doable (if not planned already).

A potential benefit for prompt review was also suggested at the meeting: the upcoming mid-term hard fork, in which some similar improvements for curve primitive support is planned, might also be able to include this change.

euonymos:
> * The introduction of a new type to represent Fr could be avoided since we currently abstract it using Integer in other BLS operations. This is already mentioned in the CIP, and I think it would be more flexible to handle the conversion in the binding layer.

Yes, this would be ideal since otherwise, we would end up having two sets of functions - old for integers and new for Fr. But how can we decide whether and `Integer` arguments belong to a field? Another thing to consider is that if we munge to/from Montgomery representation on every operation, it will be prohibitively wasteful.

And speaking more broadly, do we want to support only the scalar field of BLS of prime fields in general, at the level of Plutus API, in case we add another curve in the future, for example?

euonymos:
It would be nice to hear the thoughts of the Plutus team @effectfully @ana-pantilie @zliu41 @SeungheonOh.

rphair:
Still hoping for Plutus participation... there were some likely reviewers at the CIP meeting today, but discussion went entirely towards Leios CIP review due to massive turnout of those stakeholders & concerned devs.  I think we are on track to consider this as a candidate at the following CIP meeting: https://hackmd.io/@cip-editors/121

kwxm:
> Still hoping for Plutus participation... 

I'll try to take a close look at this by the end of the week.  Sorry for the delay.

CIP-0166/README.md
```diff
@@ -0,0 +1,433 @@
...
+and [CIP-133](https://cips.cardano.org/cip/CIP-0133). 
+Basically, we would like to expose several additional functions from this library in the Plutus API.
+
+### New types definition
+
+To represent a scalar in $F_r$ stored in the Montgomery form, a new opaque type `bls12_381_fr`
+(which corresponds to `blst_fr` type) can be used along with introducing and eliminating from/to a _byte string_: 
+
+```
+bytestring_to_bls12_381_fr: [bool, 𝚋𝚢𝚝𝚎𝚜𝚝𝚛𝚒𝚗𝚐] -> bls12_381_fr
```
SeungheonOh, line 369: is it
```
bytestring_to_bls12_381_fr: bool -> bytestring -> bls12_381_fr
```
or 
```
bytestring_to_bls12_381_fr: [(bool, bytestring)] -> bls12_381_fr
```


I assume it's former. It's just a bit confusing; please update to clarify. Same goes with other ones below.

CIP-0166/README.md
```diff
@@ -0,0 +1,433 @@
...
+The provided figures for Montgomery multiplication _include_ time for converting the initial
+vector of coefficients into the Montgomery form and back to integers in the end for the results.
+
+### Summary
+
+The way Plutus treats BLS12-381 scalars (and prime fields in general, for that matter) 
+is understandable but not quite satisfying for several reasons.
+
+The main argument showcased by this use case is that the significant share of work 
+(20% using pointwise multiplication, which will be bigger with MSM)
```
SeungheonOh, line 313: 20% of onchain exbudget I assume?

CIP-0166/README.md
```diff
@@ -0,0 +1,433 @@
...
+In addition to the conversion functions mentioned in the previous section,
+we propose to define the only function for Montgomery modular multiplication
+**bls12_381_fr_mul** as follows:
+
+```
+bls12_381_fr_mul :: [bls12_381_fr, bls12_381_fr] -> bls12_381_fr
+```
+TBD: Scalar and multi-scalar multiplication in $G1$ and $G2$ are typical downstream functions 
+for `bls12_381_fr` values, but in the current Plutus, they use integers, i.e., double conversion is needed:
+`bls12_381_fr -> bytestring -> integer` to call them.
```
SeungheonOh, line 388: nitpick: this looks like a type signature. It'd be better to write out

"`bls12_381_fr` to `bytestring` to `integer`"

CIP-0166/README.md
```diff
@@ -0,0 +1,433 @@
...
+The last part (c) that performs the actual pairing doesn't depend on the input size 
+and has a constant cost of `1B / 5K mem`. 
+
+Overall, we have the following distribution of cpu exunits for the biggest input in the benchmark:
+
+|                 Part | Unbounded (1) | Percent | Reduced (2)   | Percent | Diffence     |  
+|---------------------:|---------------|---------|---------------|---------|--------------|  
+| **(a)** Coefficients | 1_669_321_088 | 19%     | 1_847_736_704 | 21%     | +178_415_616 |  
+|   **(b)** Commitment | 5_908_456_448 | 68%     | 5_892_875_264 | 67%     | −15_581_184  |  
+|      **(c)** Pairing | 1_098_158_336 | 13%     | 1_098_158_336 | 12%     | -            |
```
SeungheonOh, line 260: How come "percent" for 1 is 13% and for 2 is 12% when they have exactly same CPU budget? 

also diff is 0 for consistency.

CIP-0166/README.md
```diff
@@ -0,0 +1,433 @@
...
+can be done much faster and with less resource usage in terms of real performance.
+Even for on-chain-plausible sizes of inputs, Montgomery multiplication can give
+at least 30–40 times speed-up.
+
+Additionally, there are other reasons to consider:
+
+* It's counterintuitive since it clashes with the mental model of the finite field.
+* Representation of field elements as `Integers` (and even worse, peculiarities of the cost model as we saw)
+pushes developers to use unbounded integers, which may lead to misunderstanding and unexpected bugs.
+* It complicates the cost model by (mis)using `Integer` type for scalars.
```
SeungheonOh, line 323: Cost of integer operation depends on the sizes of integers used. So it doesn't really complicate cost in perspective of VM. Not sure how else using `Integer` can complicates the cost model.

CIP-0166/README.md
SeungheonOh: also these belong in the below section in "Function definition"

SeungheonOh (commented):
Left some comments. Thanks for the nice writing and nice plots. These seem like a reasonable and very useful addition. 

With addition of new type for scalar in $F_r$, I think we should change existing builtins(i.e. scalar multiplication in $G1$ and $G2$) to use the new $F_r$ scalar. This does require changing existing builtins similar to how we have different semantics variant.

CIP-0166/README.md
perturbing: Given the shared correspondence between the underlying `blst` lib it is 
```haskell
bytestring_to_bls12_381_fr: bool -> bytestring -> bls12_381_fr
```

perturbing:
@jmagan 

can you explain what exactly you mean with 

> The introduction of a new type to represent Fr could be avoided since we currently abstract it using Integer in other BLS operations. This is already mentioned in the CIP, and I think it would be more flexible to handle the conversion in the binding layer.

jmagan:
@perturbing

I initially thought that $F_r$ was a simpler type — essentially an `Integer` reduced modulo r, similar to `Scalar`. However, that’s not the case: it’s actually an opaque type that stores the value in Montgomery form.

My first idea was to use plain `Integer` instead, abstracting away the $F_r$ type, much like we do in the scalar point multiplication primitive of BLS.

Mainly because I think using the blst conversion interface approach could cause problems within Plutus scripts. For instance, in polynomial multiplication (e.g., when constructing a polynomial from terms like $(x - a_n)$, we’d need to convert every $a_n$ term to $F_r$, perform the multiplication, and then convert back to Integer. This could be significantly more budget-expensive than performing the operations directly on integers.

Moreover, the use of Fr here doesn’t seem to provide any additional benefit, since the Montgomery representation is mainly an internal optimization for arithmetic efficiency.

That’s why I was considering keeping Fr as an internal detail, only using it at the interface between Plutus and the blst library. Exposing it more widely to end users might not add much value. However, as @euonymos pointed out, this could also become computationally expensive, since every function argument and return value would require a conversion between Integer and Fr.

CIP-?/README.md
```diff
@@ -0,0 +1,433 @@
+---
+CIP: ?
```
rphair:
```suggestion
CIP: 166
```

rphair (commented):
Accepted as CIP candidate at today's CIP meeting based mainly on positive response from Plutus stakeholders.  @euonymos please rename this directory to `CIP-0166` and update your "Link for reading current version" 🎉

CIP-0166/README.md
```diff
@@ -0,0 +1,433 @@
+---
+CIP: 166
+Title: Efficient scalars for BLS12-381
+Status: Proposed
+Category: Plutus
+Authors:
+    - Ilia Rodionov <hey@euony.me>
+Implementors: []
+Discussions:
+    - https://github.com/cardano-foundation/CIPs/pull/1087
```
rphair:
```suggestion
    - Original PR: https://github.com/cardano-foundation/CIPs/pull/1087
```

CIP-0166/README.md
```diff
@@ -0,0 +1,433 @@
...
+the relative share of resources spent on operations over the scalars 
+might become even bigger.
+
+In practice, some cryptographic protocols need to perform quite extensive arithmetic
+over the _scalar field_ $F_r$, particularly when working with _polynomials in KZG commitments_.
+This CIP presents a motivational example, discusses related benchmarks,
+and considers different ways of evolving Plutus 
+toward efficient implementation of scalars for BLS12-381 curve.
+
+## Motivation: why is this CIP necessary?
```
rphair:
```suggestion
## Motivation: Why is this CIP necessary?
```

CIP-0166/README.md
```diff
@@ -0,0 +1,433 @@
...
+
+The computational impact of Montgomery multiplication is straightforward, since the values of
+type `bls12_381_fr` are statically limited to 255 bits, so for the newly added `bls12_381_fr_mul`
+function we can use a static cost model.
+
+TDB: Introduction of this type potentially allows simplifying cost models for some other functions.
+Currently, scalars have to be reduced modulo the order of the group before being passed to the `blst` 
+functions, see [cardano-base](https://github.com/IntersectMBO/cardano-base/blob/6f9c20abdd3010e5a25356580cc968ba430101ad/cardano-crypto-class/src/Cardano/Crypto/EllipticCurve/BLS12_381/Internal.hs#L521).
+
+## Rationale: how does this CIP achieve its goals?
```
rphair:
```suggestion
## Rationale: How does this CIP achieve its goals?
```

rphair (commented):
@euonymos as I recall we were waiting late last year for either your response to @SeungheonOh's https://github.com/cardano-foundation/CIPs/pull/1087#pullrequestreview-3292410548 or the engagement from a more general Plutus community the editors first invited in my initial https://github.com/cardano-foundation/CIPs/pull/1087#pullrequestreview-3229232470.

We are trying to find how to progress PRs that haven't been advocated by their authors (as requested in CIP-0001) and this one looks promising to continue because most of the review responses, more clearly in hindsight, were questions rather than criticisms.  So if it can be confirmed:
1. @SeungheonOh - that you would find acceptable to merge this as-is
2. @euonymos @jmagan - that the CIP text would still accurately describe work in this field

then I can approve it for merge myself, and put on the last-check list to merge at the next CIP meeting.
* Since sometimes either authors and reviewers can drift in & out of interest, I believe it will be OK to get 1 of the 2 confirmations above (cc @Ryun1 @perturbing) to consider this valid for merge.
* Conversely, if nobody can confirm that the current state is mergeable, the "waiting" stage would eventually cause this PR to be closed.

FYI the changes below would help ensure this CIP can pass the format validation introduced in the meantime: hopefully the editors have the ability to commit to this branch so we can finish this.

CIP-0166/README.md
rphair: FYI @euonymos if we are really going to merge this after such a long refractory period without review response — as I've offered in https://github.com/cardano-foundation/CIPs/pull/1087#pullrequestreview-4955655808 — then this unusually & impractically long _Motivation_ section should be trimmed & contents moved as recommended.

(cc'ing @perturbing in case he may think this is not an issue — and in any case it would not block merging this, but would be a courtesy to readers & maintainers)

PR #1089, opened 2025-09-17
Add initial version of CIP for Transaction serialization without `isV…

teodanciu:
…alid` flag


CIP-XXXX/README.md
```diff
@@ -0,0 +1,65 @@
+---
+CIP: XXXX
+Title: Standalone transaction serialization without the `isValid` flag
+
```
```suggestion
Title: Remove the `isValid` flag from standalone transaction serialization
```
Of course @teodanciu @lehins if you can make it even shorter (without it being ambiguous) it would be even better: since CIP titles tend to be short & usually shorter than this (e.g. if `standalone transactions` would be understood as specifically as `standalone transaction serialization` then it would be nice).

@Ryun1 @Crypto2099 @perturbing I don't think we have to remove the literal `` ` `` delimiters from CIP titles anymore.  This was mainly to prevent bugs in Docusaurus when the content was copied onto the Developer Portal... now not a problem there anymore because it links to `cips.cardano.org` which interprets the backticks correctly (I think).  If indeed it's not a problem after this is merged & rendered there, I'll remove the backtick prohibition from CIP-0001 (and CIP-9999 if it's also there) at the next all-around update.

rphair (commented):
thanks again @teodanciu @lehins - tagging `Triage` for introduction at next meeting (https://hackmd.io/@cip-editors/120).

@lehins @WhatisRT I think we're now seeing a greater bandwidth of `Ledger` CIPs for atomic changes as I think was requested in some statements I attempted to summarise in [Ledger participation and bandwidth](https://github.com/rphair/cip-processes/blob/main/ledger.md).
* Assuming we stay with the usual (and originally recommended) process of CIPs for all Ledger changes, is there are standard "tag list" we can apply to confirm these PRs before what will usually be, if no objection, a summary merge?
* @Crypto2099 @Ryun1 @perturbing we should also collect names into this list with other node Ledger implementors to be sure it's evenly distributed across all node projects.

@teodanciu I've suggested a rephrase of the CIP title to:
* use the briefest possible phrasing of the change, as appearing at the beginning of the _Rationale_;
* since this CIP represents a _change_, like `Update` CIPs it's more clear to express it as an _action_ (i.e. a verb phrase: whereas proposed standards for new objects & features are generally nouns).

CIP-XXXX/README.md
rphair: Mainly resolved with https://github.com/cardano-foundation/CIPs/pull/1089/commits/202edf555133a797383f6747c539b464bd907ad8 (@teodanciu you are also welcome to commit the changes from here in the UI so the resolution happens automatically... FYI there is still also the extra blank line in the header that would also have been corrected if the suggested change had been committed this way.)

lehins:
> is there a standard list of people we can tag to confirm these PRs 

I think the list of people should be decided on CIP by CIP basis, because not everyone is affected equally. For example, this CIP could use input from the Consensus team, while other ledger related CIP might need input from the Plutus team instead.
That being said, as you know always feel free to ping me for any CIP that is Ledger related. @teodanciu also expressed interest in participating in Ledger related CIPs too.

CIP-XXXX/README.md
```diff
@@ -0,0 +1,65 @@
+---
+CIP: XXXX
+Title: Remove the `isValid` flag from standalone transaction serialization
```
lehins: I think this will be succinct enough.
```suggestion
Title: Remove the `isValid` from transaction
```

CIP-XXXX/README.md
```diff
@@ -0,0 +1,65 @@
...
+Discussions:
+  - https://github.com/cardano-foundation/CIPs/pull/xxx
+Created: 2025-09-01
+License: CC-BY-4.0
+---
+
+## Abstract
+
+We propose removing the `isValid` boolean from the CBOR encoding of standalone transactions (e.g. for mempool).
+This would not affect the serialization of the transactions within blocks.
```
lehins:
```suggestion
This would not affect the serialization of the transactions within blocks, since isValid flag is already stored separately from the transaction
```

CIP-XXXX/README.md
```diff
@@ -0,0 +1,65 @@
...
+
+## Motivation: why is this CIP necessary?
+
+The `isValid` flag in standalone transaction CBOR is not intrinsic to the protocol or to the lendger-consensus boundary:
+  * for block validation: the flag is not involved
```
lehins:
```suggestion
  * it is not signed by the transaction creator, so anyone can set it to any value they like.
  * for block validation: the value that is used is the one that was set by the consensus protocol
```

CIP-XXXX/README.md
```diff
@@ -0,0 +1,65 @@
...
+  * for remote submissions from untrusted nodes: the node ignores the incoming flag, evaluates the transaction as if `isValid = True`, if phase-2 fails for that reason alone - admits it to the mempool with `isValid = False` so collateral can be collected.
+
+The only remaining use is local submission from trusted clients (like cardano-cli), where the node reads the flag to avoid unintended collateral burn.
+This use is important, but perhaps the transaction bytes are not the best layer to encode that intent.
```
lehins, line 29: I don't believe this is important for the local submission either. cardano-node can always assume that it accepts only valid transactions and warn the user when it is not valid (eg. has a submission flag of `--ignore-is-valid-protection` or something). There should never be a reason why user would need to submit an invalid transaction, outside of testing. Alternatively this `isValid` flag can be moved into the envelope that consensus expects the transaction to be wrapped in upon, since this is indeed an implementation detail of the submission protocol, rather than transaction itself.

CIP-XXXX/README.md
```diff
@@ -0,0 +1,65 @@
...
+The proposal is to change it to:
+```cddl
+transaction = [transaction_body, transaction_witness_set, auxiliary_data/ nil]
+```
+
+## Rationale: how does this CIP achieve its goals?
+
+Removing the `isValid` flag from standalone transaction serialization simplifies the wire format without changing consensus or ledger semantics.
+
+The trusted local client submission use case might be better expressed in a different way (for example, as a Node-to-Client submit parameter), rather than embedded in the transaction bytes.
```
lehins, line 50: Yep, that is what I meant in my comment about the submission "envelope"

lehins (approved):
Love it. Hopefully community will see it as a nice simplification.

CIP-XXXX/README.md
rphair: Continued in https://github.com/cardano-foundation/CIPs/pull/1089#discussion_r2360519458.

CIP-XXXX/README.md
rphair: @lehins the succinct version is nice, but I want to make sure we get it 100% right:
* since `isValid` is already specific, would we not say "Remove `isValid`" without the "the"?
* yes we're removing it from individual transactions, but aren't we therefore removing it from all of them... and therefore "from transactions" (plural)?

If these things are true it would be updated as follows (will amend as necessary if not):
```suggestion
Title: Remove `isValid` from transactions
```

rphair:
> @lehins https://github.com/cardano-foundation/CIPs/pull/1089#issuecomment-3308799209: this CIP could use input from the Consensus team

Apologies I don't have a complete list: I'll do my best to tag `Consensus` authors & reviewers here; from memory plus those in this thread:
* https://github.com/IntersectMBO/ouroboros-consensus/pull/1559

@bwbush @dnadales @nhenin @nfrisby @jasagredo @amesgen @fraser-iohk @geo2a @abailly

CIP-XXXX/README.md
amesgen: FTR we actually broke this behavior in the past, meaning that it was impossible to submit `isValid=False` transactions via LocalTxSubmission. This was caught by the node integration tests (https://github.com/IntersectMBO/ouroboros-consensus/issues/947 / https://github.com/IntersectMBO/ouroboros-consensus/pull/949), where this exact scenario is tested to make sure that collateral-collecting txs *can* make it on chain.

Adding `isValid` as a separate parameter to LocalTxSubmission sounds feasible to me :+1:

CIP-XXXX/README.md
teodanciu: Is it implicitly understood that we mean serialization, if we remove this word from the title?

rphair: cc @lehins:
> Is it implicitly understood that we mean serialization, if we remove this word from the title?

lehins: > since isValid is already specific, would we not say "Remove isValid" without the "the"?

I'll trust you on this one, since I can never get it right. (There are no `the` or `a` in Russian language :see_no_evil:)

> yes we're removing it from individual transactions, but aren't we therefore removing it from all of them... and therefore "from transactions" (plural)?

That sounds fine with me.

> Is it implicitly understood that we mean serialization, if we remove this word from the title?

Yes. It is implicitly understood from "serialization", since serialization is the public interface and that is what lives on chain. In other words, we don't need to specify how any CIP is actually implemented, since that is not relevant information for the rest of the community.

rphair: thanks @lehins ... I'll set that to the PR title & @teodanciu I'll commit the same text here... but feel free to adjust further as necessary.

dnadales:
@rphair, thank you for the ping. @amesgen is part of the Consensus team :) I'll ask the rest of the team if there are any other remarks besides what Alex already mentioned :+1:

WhatisRT:
The reasoning of this CIP seems sensible and it doesn't affect any specifications that we currently have, so I thumbs up from me.

rphair:
@teodanciu @WhatisRT we had a shift in meeting focus in favour of Leios CIP review, and also due to not having a quorum of editors present we'll have to try confirming this as `Candidate` again at the next meeting: https://hackmd.io/@cip-editors/121

CIP-XXXX/README.md
```diff
@@ -0,0 +1,66 @@
+---
+CIP: XXXX
+Title: Remove `isValid` from transactions
+
```
```suggestion
```
remove empty line

CIP-XXXX/README.md
```diff
@@ -0,0 +1,66 @@
...
+Category: Ledger
+Authors:
+  - Teodora Danciu <teodora.danciu@iohk.io>
+  - Alexey Kuleshevich <alexey.kuleshevich@iohk.io>
+Implementors: N/A
+Status: Proposed
+Discussions:
+  - https://github.com/cardano-foundation/CIPs/pull/xxx
```
```suggestion
  - https://github.com/cardano-foundation/CIPs/pull/1089
```

CIP-XXXX/README.md
```diff
@@ -0,0 +1,65 @@
+---
+CIP: XXXX
```
```suggestion
CIP: 167
```

CIP-XXXX/README.md
```diff
@@ -0,0 +1,65 @@
...
+### Acceptance Criteria
+
+- [ ] Transaction serializers and deserializers in [cardano-ledger](https://github.com/IntersectMBO/cardano-ledger) are implemented such that they follow the cddl specification described above, and reflected in the cddl specs
+- [ ] The feature is integrated into [cardano-node](https://github.com/IntersectMBO/cardano-node) with necessary adjustments made to [ouroboros-consensus](https://github.com/IntersectMBO/ouroboros-consensus) and released as part of the Dijkstra era hard fork
+
+### Implementation Plan
+
+The implementation of this CIP should not proceed without an assessment of the potential impact on all the components that deserialise standalone transactions.
+
+## Copyright
```
neilmayhew, line 63: This isn't your problem, @teodanciu, since all the CIPs are like this, but I'm mentioning it in the hope that someone from the CF sees it.

A license isn't the same as a copyright. My understanding is that there needs to be a copyright holder in order for a license to have any meaning, because there has to be someone/entity that grants the license. I couldn't find a statement of a copyright owner anywhere in the repo. I suppose it can be inferred that it's the Cardano Foundation since the repo belongs to the CF's GitHub org, but it might be helpful to state this explicitly somewhere.

rphair: @neilmayhew it's not the CF's problem either, since they're not the authors of individual CIPs: although they could be, if a document they submitted & its GitHub history established that they were.  Original authorship is established for every file on GitHub by its user account with precise production dates of any posted original material.

If you think that's invalid because no "copyright" is obtained for that material, you are likely the only one who feels that way.  So the burden of proof would be on you to demonstrate that this GitHub history isn't a clear definition of authorship that could be used in any legal dispute... as well as defining the unique ability to grant a license.

Nobody is interested in copyrighting CIPs as far as I know: none of the authors so far, nor the CF, nor any of the CIP editors who maintain this repository.  If someone attempted to use CIP material in a way that's not allowed by its license, "my understanding" is that this would expose them to legal action depending on what consequences could be proven for that improper use.  I wouldn't want to be such an antagonist having to stake their defence on a claim like "there's no copyright holder, and the GitHub history is legally meaningless, so nobody lost anything by me stealing this material."

If you still want to pursue your criticism of our licensing strategy of the last 5 years (which was rigorously reviewed by the CF legal team in the early days of the CIP process - cc @KtorZ) then please open up a repository issue on that subject.  But if you want that issue to be seriously considered I would suggest you post more than unreferenced supposition about how open-source licenses are vulnerable without copyright filings.

neilmayhew: Thank you for explaining. I was actually trying to be helpful rather than contentious, but perhaps it didn't come across that way.

I can see that the GitHub history is likely to be sufficient to establish ownership in the event of a legal claim. I still think it's unfortunate that the sections titled "Copyright" are actually about "License", but others may see it differently.

rphair: @neilmayhew the wording may still be appropriate because it calls attention to the Licensed state of the material for anyone interested in its copyright.  Still, maybe that part can be improved and I'll try to kick that idea around to see what comes up & see if I can find out why in particular the section was titled that way (it might simply have been copied from BIPs or EIPs). 🤔

rphair:
@teodanciu somehow an incidental discussion got this off track (my mistake I think).  I don't see any obstacle to what would be a relatively simple, universally agreed-upon change: so am marking `Last Check` which would be at our next CIP meeting in 4 weeks (since the next biweekly interval would be at the Cardano Summit).  @Crypto2099 @Ryun1 @perturbing if there is consensus to merge this in the meantime I will handle the extra paperwork. 😎 cc @lehins

colll78:
This must be done very carefully because it has the potential to break a lot of bridges and light client protocols. A mithril bridge proves inclusion of a transaction in a block, if the transaction body as present in the block does not include the validity flag, then there is no way for such a bridge to determine whether the transaction succeeded or if the collateral was taken.

rphair:
@colll78 the editors will definitely ensure https://github.com/cardano-foundation/CIPs/pull/1089#issuecomment-3559791992 is addressed before merging this & take any response from @lehins @teodanciu into account about how we might define "carefully" in the CIP itself.

lehins:
> This must be done very carefully because it has the potential to break a lot of bridges and light client protocols. A mithril bridge proves inclusion of a transaction in a block, if the transaction body as present in the block does not include the validity flag, then there is no way for such a bridge to determine whether the transaction succeeded or if the collateral was taken.

`isValid` flag will still be present in the block. It is very important that it is retained, precisely because the block producer that includes transactions in the block sets that flag. This CIP is about removing `isValid` flag for transactions that are being submitted into the mempool. It is totally sound to do that because today that flag is completely ignored by the node.

lehins:
> @colll78 the editors will definitely ensure https://github.com/cardano-foundation/CIPs/pull/1089#issuecomment-3559791992

@colll78 Does my explanation above clarifies your concern? Please let me know if it doesn't and I'll be happy to discuss it more. We have this statement in the CIP, please let me know how we can make it clearer:

> This would not affect the serialization of the transactions within blocks, since isValid flag is already stored separately from the transaction

colll78:
> > @colll78 the editors will definitely ensure [#1089 (comment)](https://github.com/cardano-foundation/CIPs/pull/1089#issuecomment-3559791992)
> 
> @colll78 Does my explanation above clarifies your concern? Please let me know if it doesn't and I'll be happy to discuss it more. We have this statement in the CIP, please let me know how we can make it clearer:
> 
> > This would not affect the serialization of the transactions within blocks, since isValid flag is already stored separately from the transaction

Okay so it does not impact how bridges validate whether a transaction in a block succeeded or failed? If so then yes it addresses this concern.

lehins:
> Okay so it does not impact how bridges validate whether a transaction in a block succeeded or failed?

Yes, that is correct. This CIP only affects mempool submission, not block validation

rphair:
@teodanciu @lehins at `Last Check` review at the CIP meeting today all present editors were in favour, so we're just waiting for 1 or more additional editors to sign off on it before it can be merged anytime.  cc @Crypto2099 @Ryun1

PR #1090, opened 2025-10-01
CIP-0168? | More `BuiltinValue` Functions

fallen-icarus:
[CIP-0153](https://github.com/cardano-foundation/CIPs/tree/master/CIP-0153) added a new `BuiltinValue` type to make working with on-chain values more efficient. However, it added only a few builtin functions for working with this new `BuiltinValue` type which limits its real world usability. Adding and maintaining builtin functions is costly, but the importance of validating values in the eUTxO model justifies a wider range of builtin functions for this purpose. With this in mind, this CIP proposes a few extra builtin functions to improve the usability of this new `BuiltinValue` type while still trying to minimize the overall maintenance burden.

---
([Rendered Version](https://github.com/fallen-icarus/CIPs/blob/value-builtins-ext/CIP-0168/README.md))


fallen-icarus:
Tagging people who might be interested in weighing in:

@lehins @WhatisRT @zliu41 @effectfully @MicroProofs @colll78 @ana-pantilie @Unisay @Quantumplation @rvcas

fallen-icarus:
@effectfully With respect to your reply [here](https://github.com/cardano-foundation/CIPs/pull/1088#issuecomment-3354295285) about `intersectionValue` and `flattenValue`:

I agree with you on `flattenValue` just being `valueData`. I usually only use aiken's `flatten` for pattern matching and `valueData` covers that use case.

However, for using `intersection` for `lookupTokens`, I use aiken's `tokens` *a lot* in my [CIP-89](https://github.com/cardano-foundation/CIPs/blob/master/CIP-0089/README.md) protocols so the extra overhead would quickly add up. I'm not a software engineer, but IMO `lookupTokens` is just a simple key lookup so I don't see there being much of a burden from supporting it as a separate builtin. How is it much different than `lookupCoin`?

zliu41 (commented):
- We'll need `mkEmptyValue` as a primitive, otherwise in Plutus V4 there will be no way to construct an empty value.
- What's the motivation for `negateValue`?
- It seems useful to generalize `policies` to return `[(CurrencySymbol, [(TokenName, Integer)])]`?
- What's the motivation for `intersection`? I think it can be done without a builtin, assuming we add all other builtins proposed, so this will need strong justification.

fallen-icarus:
> We'll need `mkEmptyValue` as a primitive ...

I had assumed this was covered by `unValueData $ List []`. What would `valueData` return if used on an empty `BuiltinValue`?

> What's the motivation for `negateValue`?

Subtracting values. `unionValue` only adds two values. I can use `negateValue` on one and then `unionValue` will do subtraction. This is useful for calculating the change to return at runtime.

> ... generalize `policies` to return `[(CurrencySymbol, [(TokenName, Integer)])]`?

`policies` is meant to only return the policy ids in the `BuiltinValue` like aiken's [version](https://aiken-lang.github.io/stdlib/cardano/assets.html#policies). It is effectively just the `keys` function from `Data.Map` in Haskell. What you are suggesting just seems like `valueData` which would be more costly since it is processing extra unneeded information.

> What's the motivation for `intersection`?

I may have gotten ahead of myself on this one. The idea was to have a single builtin support arbitrary higher-level filtering functions like aiken's [`restricted_to`](https://aiken-lang.github.io/stdlib/cardano/assets.html#restricted_to). But I'm not sure now if it will actually work the way I was thinking...

zliu41:
> I had assumed this was covered by `unValueData $ List []`. What would valueData return if used on an empty BuiltinValue?

It's actually `unValueData $ Map []`. But that's only the case for Plutus V1 through V3. From V4, there will be a new `Value` constructor for `Data`, so `Value` will no longer be encoded as a `Map`. That said, I don't think `emptyValue` is actually necessary since we can just use a constant, so I take it back.

> What you are suggesting just seems like valueData

This is also only the case for Plutus V1 through V3. From V4, `valueData` will essentially be a no-op: just wrapping a `Value` within `Data`'s `Value` constructor.

The broader problem with `Value` is this: there are a _lot_ of useful operations on nested maps, any subset of them can be made primitives, and the possibilities are almost endless. So I'd want to proceed very carefully, and give people enough time to think it through and reach consensus that it is the right primitive.

For example, if we have a `valueToList` primitive that returns `[(CurrencySymbol, [(TokenName, Integer)])]`, then many other primitives can be done without primitives. So to make a case that a particular primitive should be a primitive, it will be very helpful to, for example, lay out some concrete use cases where the primitive leads to significant performance improvements.

CIP-????/README.md
```diff
@@ -0,0 +1,149 @@
+---
+CIP: ??
+Title: More `BuiltinValue` Functions
+Category: Plutus
+Status: Proposed
+Authors:
+    - fallen-icarus <modern.daidalos@gmail.com>
+Implementors: []
+Discussions: []
```
Ryun1:
```suggestion
Discussions: 
    - https://github.com/cardano-foundation/CIPs/pull/1090
```

rphair (commented):
Accepted as CIP candidate at today's CIP meeting based mainly on initial Plutus stakeholder response being queries about the specifics, rather than any claim of infeasibility or disinterest... and because support for this idea was also shown here although not included with the recent CIP-0153 update:
* https://github.com/cardano-foundation/CIPs/pull/1088

@fallen-icarus please update the link to the current rendered document & change the directory name to `CIP-0168` 🎉

CIP-????/README.md
```diff
@@ -0,0 +1,150 @@
+---
+CIP: ??
```
```suggestion
CIP: 168
```

CIP-0168/README.md
```diff
@@ -0,0 +1,150 @@
...
+unValueData :: BuiltinData -> BuiltinValue
+
+-- New functions
+negate :: BuiltinValue -> BuiltinValue
+intersection :: BuiltinValue -> BuiltinValue -> BuiltinValue
+policies :: BuiltinValue -> List [BuiltinCurrencySymbol]
+lookupTokens :: BuiltinCurrencySymbol -> BuiltinValue -> BuiltinValue
+```
+
+From these builtin functions, most of Aiken's stdlib `Value` functions can now make use of the
```
michaelpj, line 176: To be a bit annoying: it would make this proposal much more concrete to have an appendix listing all the interesting functions you want to implement with their implementations using the new primitives. I think they're presumably all quite simple, and it would make the case quite compelling.

_Ideally_ we'd also be able to see how this improves the costing behaviour of these functions.

CIP-0168/README.md
```diff
@@ -0,0 +1,150 @@
...
+intersection :: BuiltinValue -> BuiltinValue -> BuiltinValue
+
+-- | Returns all policy ids found in the value.
+policies :: BuiltinValue -> List [BuiltinCurrencySymbol]
+
+-- | Return all tokens and their quantities for a given policy id. It returns a `BuiltinValue` so
+-- that the result can make use of the `lookupCoin` builtin added in CIP-0153. It can always be
+-- converted to a `List` for pattern-matching using the `valueData` builtin added in CIP-0153.
+-- See the Rationale section for why `intersection` isn't used instead.
+lookupTokens :: BuiltinCurrencySymbol -> BuiltinValue -> BuiltinValue
```
michaelpj: I find the naming a bit confusing since it returns the same type it is given. How about `restrictPolicyTo`?

CIP-0168/README.md
```diff
@@ -0,0 +1,150 @@
...
+
+## Specification
+
+This CIP proposes adding the following new builtins functions:
+
+```haskell
+type BuiltinCurrencySymbol = BuiltinByteString
+
+-- | Negate all values in a `BuiltinValue`.
+negate :: BuiltinValue -> BuiltinValue
```
michaelpj: Adding `negate` to `union` gives us a group on `Value`. The obvious missing thing to me is scalar multiplication (to give us a module). Does anyone need that?

CIP-0168/README.md
```diff
@@ -0,0 +1,150 @@
...
+
+-- | Intersection of two `BuiltinValue`s. Returns data in the first value for the (policy ids, token
+-- names) existing in both values.
+intersection :: BuiltinValue -> BuiltinValue -> BuiltinValue
+
+-- | Returns all policy ids found in the value.
+policies :: BuiltinValue -> List [BuiltinCurrencySymbol]
```
michaelpj: What about getting the token names?

michaelpj (commented):
This seems pretty reasonable to me. It would be nice to see the motivation expanded with examples.

CIP-0168/README.md
fallen-icarus: I don't personally have a use for scalar multiplication (right now), but I could see it being useful to others.

CIP-0168/README.md
fallen-icarus: That doesn't seem as useful to me since assets could have the same token name despite having different policy ids.

CIP-0168/README.md
fallen-icarus: Perhaps it is better to make `restrictPolicyTo` be:

```haskell
restrictPolicyTo :: [BuiltinCurrencySymbol] -> BuiltinValue -> BuiltinValue
```

This would support a higher-level `lookupTokens` as well as aiken's `restricted_to` ([here](https://aiken-lang.github.io/stdlib/cardano/assets.html#restricted_to)).

colll78:
I believe I made a huge mistake not including `negateValue` in the original CIP. `negateValue` is required for a huge number of use-cases of `unionValue` (to subtract one value from another). Can anyone evaluate the scope of this builtin? Is it possible we could introduce it? It is undoubtably the single most critical value builtin I neglected.

Without `negateValue`, the implementation of the programmable tokens CIP will not be possible, so this is very high priority for me. 

Even if we need time to evaluate the demand and consider alternatives for some of the value operations proposed in this CIP, I believe `negateValue` is a special case where the need for it is very clear, and we have very high confidence that this is indeed a primitive value operation that we will need and it is very general.

CIP-0168/README.md
effectfully: Is `negate` just scalar multiplication with the scalar being `-1`? In that case I think instead of adding `negate` we should just add `scale`?

rphair: @colll78 could this do what you need to do for https://github.com/cardano-foundation/CIPs/pull/1090#issuecomment-3427255658?

zliu41:
@colll78 in Plutus V1 through V3 the builtin Value will be encoded as a Map in Data, so `negateValue` (and indeed anything else) can be done without a builtin. So even if it doesn't make it into the intra-era HF, this shouldn't be a deal breaker for most applications, should it? We can certainly add something like `negateValue` for V4.

As you can see, there's a suggestion above regarding generalizing `negateValue` into `scaleValue`. So it may take some time to decide what the best option is, even though the implementation of `negateValue` itself may not be difficult. It's also not clear whether there are more options.

CIP-0168/README.md
Unisay: Did you mean
```suggestion
policies :: BuiltinValue -> List BuiltinCurrencySymbol
```
or 
```suggestion
policies :: BuiltinValue -> [BuiltinCurrencySymbol]
```
?

fallen-icarus: Yes, but I'm not sure what is the right syntax for `Data` . I went off of [CIP-0153](https://github.com/cardano-foundation/CIPs/tree/master/CIP-0153#mary-era-value-as-a-builtin-type).

colll78: It should be `BuiltinList BuiltinCurrencySymbol` where `type BuiltinCurrencySymbol = BuiltinData`

colll78:
> @colll78 in Plutus V1 through V3 the builtin Value will be encoded as a Map in Data, so `negateValue` (and indeed anything else) can be done without a builtin. So even if it doesn't make it into the intra-era HF, this shouldn't be a deal breaker for most applications, should it? We can certainly add something like `negateValue` for V4.
> 
> As you can see, there's a suggestion above regarding generalizing `negateValue` into `scaleValue`. So it may take some time to decide what the best option is, even though the implementation of `negateValue` itself may not be difficult. It's also not clear whether there are more options.

`negateValue` can indeed be done without a builtin, but it is prohibitively expensive. In collaboration with IOG I have been working to introduce programmable assets to Cardano (https://github.com/input-output-hk/wsc-poc), the framework itself is complete, all features are implemented, and it is fully functional; the only reason that the feature is not live in production for end users is that the contracts are prohibitively expensive for mainnet production use. I have done an extensive amount of benchmarking, and the bottleneck is the value operations, particularly `unionValue` and `negateValue`, these are the two most expensive operations currently. The issue with the non-builtin implementation of `negateValue` is that it must traverse the entire value structure:

```haskell
-- | Negate the quantity of every token in a Value
negateValue :: BuiltinData -> BuiltinData 
negateValue = BI.mkMap . negateOuter . BI.unsafeDataAsMap
 where
  negateOuter  value =
    BI.caseList'
      BI.mkNilData                        
      (\pair restCS ->                                 
        let newPair =
              BI.casePair pair $ \(cs, tkMapData) ->
                BI.mkPairData cs (negateTokens $ BI.unsafeDataAsMap tkListData)
         in BI.mkCons newPair (negateOuter restCS)
      )
    value
  negateTokens tks =
      BI.caseList'
        BI.mkNilData                        
        (\tkPair restTk ->
            let newTkPair =
                  BI.casePair tkPair $ \(tkName, amtData) ->
                     BI.mkPairData tkName (BI.mkI $ 0 - (BI.unsafeDataAsI amtData))
             in BI.mkCons newTkPair (negateTokens restTk)
        )
        tks
```

The above blows up the budget when applied to even eight or nine values with less than ten unique tokens each. This alone makes it impossible for programmable tokens to go to production.

zliu41:
`caseList'` and `casePair` currently expand to large and inefficient terms: https://github.com/IntersectMBO/plutus/blob/7ebd44f7c83ba4909a66ab3466802ad6b9714fd2/plutus-tx-plugin/src/PlutusTx/Compiler/Builtins.hs#L390-L527

After the HF, the terms under the `_BuiltinCasing` case will be used, which should be a lot cheaper. Of course, they may still not be cheap enough for your use case.

CIP-0168/README.md
colll78: It could, but it would be significantly less efficient than `negateValue` (roughly 64% more expensive) if we are to assume there is a similar costing difference to `multiplyInteger -1 n` and `subtractInteger 0 n`

effectfully: These two builtins only perform the respective arithmetic operations, `scaleValue` will have to traverse the given `Value`, which is going to make the discrepancy less pronounced.

There's a different point however: `multiplyInteger` was surprisingly tricky to cost, so that's gonna propagate into `scaleValue` (so much so that it may not even be feasible to implement that with the current costing machinery, not sure), while `negateValue` would be very straightforward. @kwxm what do you think?

If Kenneth agrees costing `scaleValue` is hard, then I personally don't mind adding `negateValue` right away, given your reasoning [here](https://github.com/cardano-foundation/CIPs/pull/1090#issuecomment-3434552637), because even with faster `caseList` and `casePair`, the derivative `negateValue` is probably going to be at least an order of magnitude more expensive than a builtin one (not least because `negateValue` is guaranteed to produce a well-formed `Value` so we don't need to check any of the invariants (except for quantities being in their range), while a derivative implementation will have to, because we don't provide an unsafe builtin that could create a `Value` while avoiding the checks).

zliu41: `multiplyInteger -1 n` is 64% more expensive than `subtractInteger 0 n`? If that's the case I'd be surprised, and it would suggest that either `multiplyInteger` or `subtractInteger` is probably not costed properly.

Regarding costing of `multiplyInteger` in general - since we cap the range of quantities, we should be able to consider it constant time.

Whether it's `negateValue` or `scaleValue`, it will be linear whether it's a builtin or not. Certainly the constant factors will differ greatly, but a builtin would be much more useful if it is asymptotically more expensive without it. If we are going to add one more builtin for the intra-era HF, are we absolutely sure that it should be `negateValue` / `scaleValue`?

kwxm: > There's a different point however: `multiplyInteger` was surprisingly tricky to cost

I think that actually doesn't matter.  We're now restricting the `Quantity` type to be in the range `[-2^127, 2^127-1]` (although it might make sense to make the lower bound `-2^127 + 1` to avoid surprises), so if were scaling the value `v` by `n` and `v`  is nonempty and `n` is outside the `Quantity` range then the scalar multiplication is bound to fail for at least one of the tokens and we might as well fail right at the start.  This means that we never need to consider scalars of more than size 2, so we can probably just ignore the size of the scalar and base the cost on the total number of tokens.

kwxm: We can also try both `negateValue` and `scaleValue` and if scaling by -1 isn't significantly more expnsive than negation (and I suspect that it won't be) then we may as well go for `scaleValue`.   We could also implement `scaleValue` and check if the scalar is -1 and use negation internally in that case if it's faster, although it might be a little tricky to reflect any improvement from doing that in the costing.

zliu41:
Based on the discussion it seems sensible to implement `scaleValue` for the intra era HF, so we'll do that. @colll78 @fallen-icarus if you can add `scaleValue` to the original CIP, that would be helpful, since it makes it clear what the exact scope for the intra era HF will be. Making a separate CIP just for `scaleValue` is fine, too.

fallen-icarus:
I'm fine with `scaleValue` being added to CIP-153. Using a new CIP for `scaleValue` seems like a bad documentation practice. I'll update this CIP once CIP-153 has been updated.

rphair (commented):
@fallen-icarus now that the plan to reconcile `negateValue` vs. `scaleValue` has been settled — which I think resolves the only pending issue for this update PR — I think this would be ready to merge unless @colll78 has anything to add or amend, pending some uncertainty about where the change is going.  I'm just not making sense of the last comment:
> I'm fine with `scaleValue` being added to CIP-153. ... I'll update this CIP once CIP-153 has been updated.

Since it seems appropriate to add  `scaleValue` to _this_ CIP update PR, what CIP-0153 update would we be waiting for in the second sentence?

If in fact we're ready to make the necessary changes here, please confirm once that's done so we can get @colll78 @zliu41 to OK it and then we might arrange for a CIP editor review online so we don't have to wait until the next CIP meeting in the last week of November: especially if _this_ PR has to be merged before continuing with any more updates (?).

fallen-icarus:
This CIP isn't an *update PR*. It is a brand new CIP. AFAIU CIP-153 is for the builtins that will be added in the next intra-era hardfork. Meanwhile, this CIP is for the builtins that will be added after that hardfork, but for the new era.

rphair:
Thanks @fallen-icarus... I had a feeling I was missing something.  I was a little disoriented because [all CIP-0153 updates have currently been merged](https://github.com/cardano-foundation/CIPs/pulls?q=is%3Apr+CIP-0153+in%3Atitle) so I didn't know what updates you might be talking about.  Therefore, as I understand it, you're _already_ free to submit your own CIP-0153 update adding `scaleValue` (please correct again if that seems wrong).

CIP-0168/README.md
effectfully: Great points, thank you, I withdraw my concerns.

fallen-icarus:
Changes in latest commit:
- Added an example in the Motivation section (based on this [comment](https://github.com/cardano-foundation/CIPs/pull/1090#pullrequestreview-3353186611)).
- Added a new [Supported Higher-Level Functions](https://github.com/fallen-icarus/CIPs/blob/value-builtins-ext/CIP-0168/README.md#supported-higher-level-functions) section (based on this [comment](https://github.com/cardano-foundation/CIPs/pull/1090#discussion_r2442100559)). I'm not experienced enough to know their before/after Big O notations... @KtorZ @MicroProofs @rvcas I'd appreciate it if you could check this section.
- Changes to the builtins list:
    - `scaleValue` is now mentioned as part of CIP-153. (I'm leaving the CIP-153 update to @colll78)
    - removed `negateValue` since it is replaced by `scaleValue`
    - added `emptyValue` constant (based on this [comment](https://github.com/cardano-foundation/CIPs/pull/1090#issuecomment-3373057162))
    - added `valueToList`
    - renamed `lookupTokens` to `restrictValueTo` and it now takes a `[BuiltinCurrencySymbol]` to make it a more general filtering function.
    - added `filterOutPolicies` as the inverse to `restrictValueTo`
    - removed `intersection`
    - added `policyCount` (is this necessary?)
    - added `assetCount` (is this necessary?)
    - added `valueEquals` (in case `==` isn't automatically supported)

---
Lmk if I misunderstood anything in this discussion so far.

zliu41:
> scaleValue is now mentioned as part of CIP-153. (I'm leaving the CIP-153 update to @colll78)

An update to CIP-153 soon would be great - `scaleValue` has been implemented, and hard fork communications will start referencing it soon.

colll78:
> > scaleValue is now mentioned as part of CIP-153. (I'm leaving the CIP-153 update to @colll78)
> 
> 
> 
> An update to CIP-153 soon would be great - `scaleValue` has been implemented, and hard fork communications will start referencing it soon.

Will do.

KtorZ:
@fallen-icarus I had a brief look at the Haskell version of the Aiken's stdlib functions; this looks interesting. Note sure what exactly you wanted me to check in this section, but a few remarks:

- `policyCount` and `assetCount` seems unused (and a bit useless 😅?)
- the `valueToList` looks helpful, but is likely an `O(n)` (depends on how the underlying map is implemented). One would typically then traverse again the list to do stuff with it. So as a builtin, it would perhaps makes more sense to have a more generic fold operation? Something along the lines of [`reduce`](https://aiken-lang.github.io/stdlib/cardano/assets.html#reduce). With it, we can implement quite many other functions, and it would be extremely handy to have an efficient fold.

fallen-icarus:
@KtorZ Thanks for looking. I just wanted to confirm aiken's stdlib for `Value` can be optimized using the new builtins.

I typically use `valueToList` (or aiken's `flatten`) for pattern matching where I effectively count the number of entries like:

```
[_, _] -> True
_ -> False
```

So I could use `policyCount` and `assetCount` instead of pattern matching.

> it would perhaps makes more sense to have a more generic fold operation? Something along the lines of [reduce](https://aiken-lang.github.io/stdlib/cardano/assets.html#reduce).

I thought there was [no support](https://github.com/IntersectMBO/plutus/issues/6587) for higer-order builtin functions; the linked PR was never merged. It would be great to support generic filtering and `reduce`, but AFAIU it is hard to cost them. @effectfully 

IMO `valueToList` is a catch-all for anything we can't support directly.

colll78:
Higher-order builtins are very difficult to cost, but extremely high value. I think there was a proposal for limited higher order builtins which restrict the arguments to be compositions of builtins which was considered much lower scope to cost and implement, however, there is still quite a bit of complexity.

colll78:
```haskell
emptyValue :: BuiltinValue
valueToList :: BuiltinValue -> BuiltinList ???
policies :: BuiltinValue -> [BuiltinCurrencySymbol]
restrictValueTo :: [BuiltinCurrencySymbol] -> BuiltinValue -> BuiltinValue
filterOutPolicies :: [BuiltinCurrencySymbol] -> BuiltinValue -> BuiltinValue
policyCount :: BuiltinValue -> BuiltinInteger
assetCount :: BuiltinValue -> BuiltinInteger
valueEquals :: BuiltinValue -> BuiltinValue -> BuiltinBool
```

I think some of these should be removed:

`emptyValue` would not be a new builtin, instead it is possible to construct empty values as constants via:
`(Constant () (Some (ValueOf DefaultUniValue Value.empty)))`

```haskell
policyCount :: BuiltinValue -> BuiltinInteger
```
I'm not sure what the purpose of the above is, it cannot be used to prevent dust token attack, and you have no information about what the actual policies are. 
1. The value contains only tokens relevant to the protocol logic, ie. only tokens that the smart contract interacts with in some way (performs some validation against those tokens or policy ids or token names) then dust token attack is prevented by explicitly checking for the presence of those tokens and the absence of unrecognized tokens. 
2. The value must contain tokens relevant to the protocol logic, but can also obtain some number of tokens that are irrelevant to the protocol logic (to facilitate smart contract composability, ie they may not be relevant to the logic of this contract, but the address of the output to which these tokens go to may be another smart contract which does care about these tokens). 

In case two, we only care about the total number of unique tokens, in which case the use-case can be satisfied by the presence of your proposed `assetCount` builtin. However, it is also likely that the introduction of `BuiltinValue` in their current form, and even more-so in their Dijkstra form where they are added as a new constructor to data (thus removing the overhead of `unValueData` and `valueData`) will make dust token attacks a non-issue, as currently a worst case call of `quantity_of` on a single maximum size value (limited by ledger) will consume roughly ~65%+ of the execution budget, with the `lookupCoin` builtin on multi-asset-values even those at the maximum size limit should only consume a fraction of the execution unit budget, and so the burden for an attacker to successfully deadlock a utxo with a dust token attack would require that the smart contract in question is performing a massive amount of these value operations.  

I think `assetCount` would be a very trivial builtin to add given that the current implementation of Value already stores:
```haskell
{-| Total size, i.e., the number of distinct `(currency symbol, token name)` pairs
contained in the `Value`. -}
totalSize :: Value -> Int
totalSize (Value _ _ size _) = size
{-# INLINE totalSize #-}
```
so `assetCount` could simply extract this field from the value and return it without performing any additional work. 

I think `valueToList` goes against the purpose of introducing `BuiltinValue` which is to introduce a primitive type for value and functions for the operations on that type. If there is a use-case where you would currently need `valueToList` then instead support for that use-case should be added to `BuiltinValue`. Also the type signature of `valueToList` is incomplete, what elements do you want to be in the BuiltinList? I assume you would want `BuiltinList BuiltinData` where `BuiltinData` is a nested data encoded pair (or `BuiltinPair` which may be difficult due to no direct constructor)? I don't think there is ever a case where this should be used in any capacity. Functionally, anything you can do with `flattenValue` you can do with a non-flattened value (nested map representation which you can obtain with `unsafeDataAsMap . valueData`) more efficiently after accounting for conversion costs. flattenValue is a dev-ex convenience, as many developers feel that operating on nested map representations is more complicated than operating on the flattened representation. Though again, if there is an operation that is possible and useful on values as nested maps or values as flattened lists, then we should just support that operation on `BuiltinValue`. Eventually, far in the future, we can revisit limited higher order builtins.  


For now `valueEquals` is expected to be represented as `equalsValue x y = equalsData (valueData x) (valueData y)`. In Dijkstra, a constructor for `BuiltinValue` will be added to `BuiltinData` at which point `equalsValue` can be implemented as `equalsValue = equalsData`, so this builtin is also unnecessary. 

TDLR: 
You should make the proposed builtins:
```haskell
restrictValueTo :: [BuiltinCurrencySymbol] -> BuiltinValue -> BuiltinValue
filterOutPolicies :: [BuiltinCurrencySymbol] -> BuiltinValue -> BuiltinValue
assetCount :: BuiltinValue -> BuiltinInteger
policies :: BuiltinValue -> BuiltinList BuiltinCurrencySymbol

-- where:
type BuiltinCurrencySymbol = BuiltinData 
```

rphair:
@colll78 @fallen-icarus since we are already introducing the new proposed update https://github.com/cardano-foundation/CIPs/pull/1134 at the next CIP meeting (https://hackmd.io/@cip-editors/126) we could try to gather some Plutus/Ledger reviewers and put this PR up again for a dedicated Review and look at both of these in the same light... just post as such & we'll add it.

At the last CIP meeting editors were wondering what to do with some of these medium-age `Confirmed` proposals & it might be a way to keep this one moving forward (cc @Ryun1 @perturbing).

fallen-icarus:
> `policyCount :: BuiltinValue -> BuiltinInteger`
>
> I'm not sure what the purpose of the above is, it cannot be used to prevent dust token attack, and you have no information about what the actual policies are.

@colll78 It can be used for ensuring no extraneous assets are found in the UTxO. The datum tells me everything I need to know. Consider creating a limit order looking to swap ADA for DJED. It must have the beacon tokens (which all share a policy id) and ADA; no other assets can be present in the creation step. So `policyCount` should yield 2. (If DJED was being offered then it should be 3 since ADA must always be present.) Considering it seems very simple to implement (seems basically the same as `assetCount`), I think we should include it.

> I think `valueToList` goes against the purpose of introducing `BuiltinValue` which is to introduce a primitive type for value and functions for the operations on that type. If there is a use-case where you would currently need valueToList then instead support for that use-case should be added to BuiltinValue.

I don't agree with removing `valueToList`. This would mean that the use case is basically impossible *until* we add the builtin. So yes, I agree we *should* add the builtin if it is needed; but we should not block use cases just because we didn't foresee them. An inefficient implementation is sometimes better than no implementation. The type signature in my mind is as you say:

```
valueToList :: BuiltinValue -> BuiltinList BuiltinData
```

I'm fine with removing `emptyValue`, and `valueEquals`.

colll78:
> @colll78 It can be used for ensuring no extraneous assets are found in the UTxO. The datum tells me everything I need to know. Consider creating a limit order looking to swap ADA for DJED. It must have the beacon tokens (which all share a policy id) and ADA; no other assets can be present in the creation step. So `policyCount` should yield 2. (If DJED was being offered then it should be 3 since ADA must always be present.) Considering it seems very simple to implement (seems basically the same as `assetCount`), I think we should include it.
> 

What if the DJED policy ID can also mint tokens of arbitrary token names besides the token name "DJED"? Even if the policyCount yields two, then the UTxO could have 255 tokens in it. Instead wouldn't you want to use `assetCount` which would tell you in `O(1)` the exact amount of tokens? 

> I don't agree with removing `valueToList`. This would mean that the use case is basically impossible _until_ we add the builtin. So yes, I agree we _should_ add the builtin if it is needed; but we should not block use cases just because we didn't foresee them. An inefficient implementation is sometimes better than no implementation. The type signature in my mind is as you say:

What use-case would `valueToList` support that would not be possible without it as a specialized builtin? As it stands right now, you can implement `valueToList` as `unsafeDataAsMap  . valueData ` which returns `BuiltinList (BuiltinPair BuiltinData BuiltinData)`. I imagine there is something you want to do when you have the list of assets in a flattened representation (ie. the use-case) but whatever that is you should be able to do it on the result of `unsafeDataAsMap . valueData`, and if that is too inefficient then whatever use-case you need the underlying list for should be supported directly as a builtin operation on value right? 

```haskell
-- You can choose to flatten the result of this or just work with the nested map representation directly.
-- This is how you could achieve this with the current interface. 
valueToListNow :: BuiltinValue -> BuiltinList (BuiltinPair BuiltinData BuiltinData)
valueToListNow =  unsafeDataAsMap . valueData

-- This converts the BuiltinValue to an SOP??? list of SOP tuples? (keep in mind Aiken does not support SOP types natively)
valueToListAsProposedByThisCIP :: BuiltinValue -> [(CurrencySymbol, (TokenName, Integer))]
valueToListAsProposedByThisCIP = ...

-- or perhaps what was intended was:
valueToListAsProposedByThisCIP2 :: BuiltinValue -> BuiltinList BuiltinCoinTuple 

type BuiltinCoinTuple = BuiltinData 
-- where builtin coin tuple is builtin data of the following structure:
-- `BuiltinList BuiltinData` of size three where the elements are:
-- `[CurrencySymbolAsData, TokenNameAsData, IntegerAsData]`
type CurrencySymbolAsData = BuiltinData 
...
```

`valueToListAsProposedByThisCIP` I don't think is advisable as SOP encoded types are quickly becoming very bad in comparison to BuiltinData types due to the new builtin functions (ie. `dropList`) and builtin pattern matching support. For `valueToListAsProposedByThisCIP2` I think this doesn't warrant its own builtin because it doesn't provide any concrete advantage over what we already have with the BuiltinValue interface which is `valueToListNow`.

fallen-icarus:
Good points. I'll remove both `policyCount` and `valueToList`.

fallen-icarus:
@zliu41 What is the latest with this CIP? Are these extra builtins not planned for PlutusV4?

rphair:
@fallen-icarus @zliu41 does the query above suggest this proposal is headed for implementation without the CIP draft being merged yet? 😟  If so then we have a breach of the usual standards process: as previously commented above, this was still waiting on implementation details to be settled.

We can't merge these documents unless authors, advocates, and implementors confirm that they're finalised.  Please (cc @colll78 @KtorZ) post whatever you can to help editors decide whether to fast-track merging this: especially if indeed it's already an item on the development board. 🙏 cc @Ryun1 @perturbing

zliu41:
@fallen-icarus We do plan to add additional value builtins in V4 but I'd like to see the specification and motivation strengthend. For example
- "Returns all policy ids found in the value" - it should more explicitly say "in ascending order".
- "filterOutPolicies" - the use case given is "withoutLovelace". Is this a common use case? Why?
  - By the way, "restrictValueTo" and "filterOutPolicies" are duals. It's better to name them "keepPolicies" and "dropPolicies".
- "assetCount" - no use case is given. Why is it needed?

fallen-icarus:
@zliu41 I could use help with the spec since I'm not an actual software engineer. Maybe there are better ways than what I'm suggesting. I can explain my use cases but I'm not the right person to judge if the currently listed builtins are the right ones.

---
[CIP-89](https://github.com/cardano-foundation/CIPs/tree/master/CIP-0089) uses Cardano native assets to tag UTxOs to make filtering the UTxO set off-chain trivial. Just like how you can filter Github issues by "Feature", "Bug", etc, we can do the same for "limit order", "loan request", etc. So for my protocols, my biggest concern is making sure these UTxO tags are not misused. I call these tag tokens "Beacon Tokens".

Consider a limit order UTxO that wants to trade ADA -> DJED. In addition to having ADA, it needs **exactly** three Beacon Tokens.

1. Pair beacon token which encodes that this limit order is for ADA -> DJED.
2. Offer beacon token which encodes this limit order is offering ADA.
3. Ask beacon token which encodes this limit order is asking for DJED.

If the UTxO contains any other beacon token, that is a security issue! There should also only be 1 copy of each beacon token in a given UTxO.

My use cases for `restrictValueTo`, `filterOutPolicies`, and `assetCount` all follow from the above. Today, I treat the `Value` as a list and just traverse the list once, pattern matching on the expected beacon tokens and DeFi-relevant tokens as I go. **This is error prone.** I had issues where I missed a pattern match and an edge case allowed withdrawing beacon tokens to invalid UTxOs.

A less error prone (though possibly slightly less efficient) approach is to split the `Value` into beacon tokens and everything else, hence `filterOutPolicies` and `restrictValueTo`. I then do separate checks on each portion. I know the beacon token policy id (it is the smart contract's hash) so I can easily segregate them. Then I use `assetCount` on the `Value` with only beacon tokens to make sure there are no extra tags present. Consider the limit order above. What if it had 4 beacon tokens where 3 were correctly configured but the 4th was to DJED -> HOSKY? That would destroy the fidelity of the off-chain filtering. I always need to make sure the tags are *exactly* correct. If I skipped the `assetCount` check, the only way for me to be sure that extra beacon tokens are not present is by converting the `Value` to a list and pattern matching. This is possible, but again it is less clear and more error prone than:

```
-- 1. Ensure exactly 3 tags exist
assetCount beaconValue == 3

-- 2. Ensure the specific required tags are the ones present
lookupCoin beaconValue pairBeaconName == 1
lookupCoin beaconValue offerBeaconName == 1
lookupCoin beaconValue askBeaconName == 1
```

AFAIU you already have the asset count available, so why not just expose it? Doesn't pattern matching also result in larger script sizes than just using the above? So while execution budget might be less efficient, there is an efficiency gain in script size in addition to the legibility gain for human auditors. Would this actually also be more efficient for the execution budget than pattern matching during a runtime list traversal?

As a related use case, I sometimes need to burn all beacon tokens which are no longer used across many input UTxOs. So the optimized flow is:

1. Get the beacon values from each UTxO using `restrictValueTo`
2. Add them up using `unionValue`
3. Negate them using `scaleValue`
4. Check that the `txInfoMint` (after also passing it through `restrictValueTo`) exactly matches this negated value to ensure the burn is perfectly accurate.

Wouldn't this be significantly more efficient than today:
1. Traverse each UTxO's `Value` as a list to get the beacon tokens
2. Add them up at runtime
3. Check the tx mint matches the value (after traversing it as well to extract the beacon tokens)

colll78:
@zliu41 

> * "Returns all policy ids found in the value" - it should more explicitly say "in ascending order".
> * "filterOutPolicies" - the use case given is "withoutLovelace". Is this a common use case? Why?
>   
This is indeed the common use-case. Because there are many cases where smart contracts are managing the lifecycle of, and thus only care about the value of non-ada tokens. Due to the requirements of the ledger these UTxOs, by necessity, will have ada in them (to satisfy min_ada requirement), however, the ada quantity itself in such cases is irrelevant to the scope of the contract. Say for instance we have a state UTxO that contains an authentication token and some dApp state in its datum. The sole purpose of this UTxO is to provide an authenticated source of this dApp state. The contract at which it lives supports a number of operations which represent state transitions of that dApp state. All of those operations typically (at a very abstract level) will look like:

```
let ownInput = elemAtWithAuthTokenNFT authTokenCS ownInputIdx txInputs 
     ownOutput = elemAt ownOutputIdx txOutputs
 in txOutAddress ownInputs == txOutAddress ownOutput
     -- ^ The output is indeed a continuing output
       && withoutLovelace (txOutValue ownOutput) == withoutLovelace (txOutValue ownOutput) 
       -- ^ The state token is preserved 
       &&  allowedStateTransition (txOutDatum ownInput) == (txOutDatum ownOutput)
       -- ^ The state in the continuing output is exactly the result of the state of the input after applying the allowed state transition
```

Typically when working with the data representation of Value, `withoutLovelace` currently this is efficiently achieved by something like:
```
pstripAdaH ::
    forall (v :: AmountGuarantees) (s :: S).
    Term s (PValue 'Sorted v) -> Term s (PValue 'Sorted v)
pstripAdaH value =
    let nonAdaValueMapInner = ptail # pto (pto value)
     in pcon (PValue $ pcon $ PMap nonAdaValueMapInner)
```
Which simply drops the first entry in the outer `Value` map. This works for any ledger provided value (any `Value` in the script context, namely in inputs and outputs) because the ledger guarantees that the first entry of a `Value` in a UTxO must always be Ada. 

>   * By the way, "restrictValueTo" and "filterOutPolicies" are duals. It's better to name them "keepPolicies" and "dropPolicies".
> * "assetCount" - no use case is given. Why is it needed?

`assetCount` is necessary to provide reasonable limits in the size of an output value to prevent multi-step locking / bricking vulnerabilities. IE. a script requires producing an output, we can call this output `destination_utxo`, to some destination address such that `destination_utxo_value >= some_value`. If we allow `destination_utxo_value` to be unbounded, then it's possible to for a satisfying transaction to be maliciously compliant, it sends `destination_utxo_value` that indeed contains `some_value` but it also contains 245 other garbage tokens. This makes subsequent spends of that output more expensive, especially if the destination address is a script address. There are legitimate reasons why the contract would want to allow users to include other tokens in addition to `some_value` in that output, however there is typically a reasonable upper bound that makes sense for honest use-cases (ie. `assetCount destination_utxo_value <= 10`)

See also:
https://github.com/cardano-foundation/CIPs/pull/1090#issuecomment-3762289802

zliu41 (changes requested):
I'd like to see the following changes being made, before merging this PR:
- The motivations stated in the above comments should be summarized in the CIP
- Describe the semantics of the builtins more precisely as I mentioned previously
- Rename restrictValueTo/filterOutPolices to keepPolicies/dropPolicies
- The comment above `assetCount` says "Is this more efficient than pattern-matching on the result of `restrictValueTo`???" Remove this, and instead explicitly state that the addition of this builtin will be conditional upon the Plutlus team verifying that the efficiency gain justifies it.

fallen-icarus:
@colll78 Isn't the `withoutLovelace` use case better served by using `insertCoin` from CIP-153? `insertCoin` overrides the existing integer (and drops the entry if the new integer is 0).

```Haskell
insertCoin "" "" 0 inputValue == insertCoin "" "" 0 outputValue
```

I still think `dropPolicies` is important for splitting validation of protocol tokens from DeFi tokens, but `withoutLovelace` is a weak motivation for this specific builtin because it can be covered by `insertCoin`.

colll78:
> @colll78 Isn't the `withoutLovelace` use case better served by using `insertCoin` from CIP-153? `insertCoin` overrides the existing integer (and drops the entry if the new integer is 0).
> 
> ```haskell
> insertCoin "" "" 0 inputValue == insertCoin "" "" 0 outputValue
> ```

It is served by that; however, I would imagine that `insertCoin` which has to have its costing function designed for the worst-case outcome (full traversal of the value), will be significantly more expensive than something like:
```haskell
withoutLovelace = BI.mkMap . BI.tail . BI.unsafeAsMap 
```
Because the worst case outcome for this is O(1) with a very small constant cost. 

I would not suggest a new specialized `withoutLovelace` builtin, what I would suggest is special costing for operations interacting with Ada, because such operations will be very common, and the ledger invariants guarantee that ada, if present, will always be the first entry in the value. The ledger `MultiAssetValue` representation separates Ada explicitly for this reason, it allows for specialized efficiency for operations on Ada.

zliu41:
> withoutLovelace = BI.mkMap . BI.tail . BI.unsafeAsMap 

@colll78 The builtin Value is backed by the Haskell Map which is a balanced binary tree. Removing the smallest key isn't cheaper than removing any other key.

We can potentially optimize it by storing the lovelace entry outside the map in a separate field, then make the cost of `insertCoin` cheaper if currency = token = 0.

In any case, if removing lovelace is the only use case for, then `dropCurrencies` is indeed unnecessary.

fallen-icarus:
> In any case, if removing lovelace is the only use case for, then dropCurrencies is indeed unnecessary.

@zliu41 It is not the only use case. Splitting protocol tokens from DeFi tokens is my main use case. They require separate validations.

fallen-icarus:
@colll78 @zliu41 Take a look at the updates when you get a chance.

zliu41:
Since Value is going to be completely opaque, there also needs to be an `emptyValue` primitive, otherwise there will be no way to construct an empty Value. Otherwise looks good.

zliu41:
I'm now leaning against adding Value constructor to Data for Plutus V4. And if so we won't need `emptyValue`.

- The topline feature for the Dijkstra HF is Leios. And that dictates the timeline. Currently the expectation is all development should finish by end of Sep for a Dec HF. And this cannot be pushed back because of Plutus, especially for something far from critical like this.
- It is possible that some of the new Value primitives in this CIP won't be ready by Sep, as our commitment was to land them by Dec. If we make Value opaque and only manipulatable by Value primitives, then not having all primitives could make some logic not expressible.
- I've experimented with adding Value to Data - it is a quite invasive change and the code will need to be carefully written and audited, plus ledger integration. End of Sep may not be enough for this alone.
- Also, adding Value to Data is not among the plutus proposal's deliverables. It would be better proposed separately, with a specific target date and sufficient resource allocated.

That said, I'll open a PR adding Value to Data and have it reviewed, in case the Leios work is delayed, which will delay the HF. cc @colll78 @kwxm @Unisay @SeungheonOh

fallen-icarus:
So do I still add `emtpyValue` to this CIP?

rphair:
a note from @colll78 about the timeline that may also be relevant to the last question (cc @zliu41): https://github.com/cardano-foundation/CIPs/pull/1238#issuecomment-5316448582

zliu41 (approved):
I think we can merge it as is. If we decide that we need `emptyValue` later, we can amend it.

rphair (approved):
thanks @zliu41; I think that answers the one remaining review point & we'll check with other editors / reviewers to see if they're OK merging this from `Last Check` tomorrow (cc @fallen-icarus): https://hackmd.io/@cip-editors/141

Ryun1 (approved):

PR #1095, opened 2025-10-14
CIP-0153 | Add range limits to amounts in builtin Value

zliu41:
This makes costing much easier, since `unionValue` involves integer addition. If we allow unbounded integers, then the integer sizes would need to be taken into account.

The Int128 range is chosen because the ledger allows amounts up to 2<sup>64</sup>-1. However builtin Values may be used for intermediate computation where the amounts temporarily exceed that limit. So to be safe we propose using [-2<sup>127</sup>, 2<sup>127</sup>-1].

cc @colll78 @kwxm @ana-pantilie @effectfully @Unisay @basetunnel @fallen-icarus @SeungheonOh @MicroProofs

### DIFF
--- a/CIP-0153/README.md
+++ b/CIP-0153/README.md
@@ -91,6 +91,7 @@ We propose the following set of builtin functions to accompany the new builtin t
     - it returns a Mary-era Value with the `Coin` inserted, silently discarding any previous value.
       If the `BuiltinInteger` argument (the quantity) is zero, the `Coin` is removed.
     - Both `BuiltinCurrencySymbol` and `BuiltinTokenName` must be no longer than 32 bytes (unless the amount is zero).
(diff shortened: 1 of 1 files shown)


rphair (approved):
I'm approving this as a matter of course (and tagging `Last Check` which can be repealed upon any objection) following from https://github.com/cardano-foundation/CIPs/pull/1088#issuecomment-3402629916 on the already merged update, while also looking forward to review from those tagged.  With sufficient confirmation we can put this though by the next CIP meeting: https://hackmd.io/@cip-editors/122

colll78 (approved):

Crypto2099 (approved):
Simple change, looks good

Ryun1 (approved):
thanks for the update @zliu41

PR #1102, opened 2025-10-20
CIP-0023 | Fix Abstract Phrasing

Cerkoryn:
This simply updates the abstract that was still outdated after our recent modernization of this CIP.

Old:
"This proposes to create a more fair marketplace for stake pools by reducing the minimum fixed pool fee and adding a minimum variable pool fee."

New:
"This CIP introduces a new protocol parameter, `minPoolMargin`, which specifies a lower bound on the variable fee (margin) a stake pool may set. The parameter is introduced initially set to `0` to avoid disrupting existing pool certificates.  This proposal does not change or reduce the existing minimum fixed pool fee (`minPoolCost`)."

### DIFF
--- a/CIP-0023/README.md
+++ b/CIP-0023/README.md
@@ -16,7 +16,7 @@ License: CC-BY-4.0
 
 ## Abstract
 
-This proposes to create a more fair marketplace for stake pools by reducing the minimum fixed pool fee and adding a minimum variable pool fee.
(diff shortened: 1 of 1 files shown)


Cerkoryn:
I wasn't sure if I should create a new branch or just pull from the remote on the same branch, edit, and submit again.  I went with the latter not realizing it would bring the entire commit history with it 😅

PR #1103, opened 2025-10-26
CPS-0023? | Cardano Multi Asset Treasury (CMAT)

HinsonSIDAN:
This CPS outlines the motivation of turning Cardano treasury from `lovelace` only into capable of holding multi assets including various Cardano Native Tokens.

Our co-authors (me plus Felix & Nicolas) would continue seeking for perspectives and feedback to update the draft until we confirm, mostly by circulating this draft and workshops to be held during upcoming Cardano Summit.

Also tagging #1061 for cross referencing, as the progress / update on the proposed CIP-159 will affect how the community should approach this CPS.

---

[branch directory, including rendered document](https://github.com/sidan-lab/CIPs/blob/cardano-multi-asset-treasury/CPS-0023)

PR #1083, opened 2025-09-01
Proposal of canonical ledger state format

qnikst:
This proposal defines the Simple Canonical Ledger State (SCLS), a stable, versioned, and verifiable file format for representing the Cardano ledger state. It specifies a segmented binary container with deterministic CBOR encodings, per-chunk commitments, and a manifest that enables identical snapshots across implementations, supports external tools (e.g., Mithril), and future-proofs distribution and verification of state.

PR #1109, opened 2025-10-29
CPS-0024? | Canonical CBOR Serialization

HinsonSIDAN:
This CPS emerged from community feedback at events like Cardano Builder Fest 2025, where developers identified CBOR serialization fragmentation as a critical pain point hindering ecosystem maturity.

This CPS addresses the growing interoperability challenges caused by non-deterministic CBOR serialization across Cardano's tooling ecosystem. The same logical transaction or script can be encoded in multiple valid ways, leading to different hashes and breaking multi-signature workflows, cross-tool transaction building, and script reference consistency.

Ideally we can establish a canonical CBOR serialization standard that would be adopted across major libraries and wallets, ensuring predictable behavior and reducing the development friction that currently exists when working across different tools.

---

([rendered latest document](https://github.com/sidan-lab/CIPs/blob/cps-canonical-cbor-serialization/CPS-%3F%3F%3F%3F/README.md))


fallen-icarus:
Thanks for opening this discussion! This topic is outside my area, but I came across this [comment](https://github.com/IntersectMBO/cardano-ledger/issues/2943#issuecomment-1203982266) (and this [one](https://github.com/IntersectMBO/cardano-ledger/issues/2943#issuecomment-1203989504)) by @dcoutts a while ago which AFAIU explains the reasoning at the time. The entire Issue thread might be worth a read.

GeorgeFlerovsky:
FYI @nau @fernweh0

yHSJ:
Thank you for drafting a CPS about this. You're right, a lack of canonical encoding has definitely been frustrating for many of us Cardano developers, especially those of us working on alternative node implementations (I can speak to my experience working on Amaru primarily). I have only done a quick skim of the content so I won't comment on the particulars in the CPS here right now, but hopefully I will have more time later to sit down and reread it carefully.

But, to give some context:
At the first node diversity workshop, canonical representation of ledger state was heavily discussed since it's needed (or at least very helpful) for many things, including testing and bootstrapping.

Sundae Labs has started work on a conformance testing suite which currently uses the `NewEpochState`, encoded as CBOR, from the Haskell node but we're eager to conform with a reasonable and appropriate standard there. They currently live in the Cardano Blueprints project, where I assume a canonical ledger state specification would eventually reside, along with a CIP: https://github.com/cardano-scaling/cardano-blueprint/tree/main/src/ledger/conformance-test-vectors (cc: @Quantumplation)

Tweag has received treasury funding to work on exactly that actually, and they have a draft PR open already: https://github.com/cardano-foundation/CIPs/pull/1083 (cc: @nc6 and @qnikst)

CPS-????/README.md
```diff
@@ -0,0 +1,137 @@
+---
+CPS: ?
+Title: Canonical CBOR Serialization Standard
+Category: Tools, Wallets, Ledger
```
rphair: @HinsonSIDAN it's not too early to decide on a single category that you think represents the ecosystem sector where the most work would need to be done to cover this problem statement.

If  don't think that possible, or would like to promote the idea of multiple categories in CIPs and CPSs, you could read this recent comment:
* https://github.com/cardano-foundation/CIPs/pull/749#issuecomment-3441294312

and contribute your experience & professional opinion here, if you feel multiple categories would be better than single choices, and also if any other conservatively chosen additional tags could provide good coverage of what CIPs and CPSs are really about:
* https://github.com/cardano-foundation/cf-cip-frontend/discussions/55

HinsonSIDAN: I have chosen the `Tools` category for now. Thanks for reminding that

CPS-????/README.md
```diff
@@ -0,0 +1,137 @@
...
+CPS: ?
+Title: Canonical CBOR Serialization Standard
+Category: Tools
+Status: Open
+Authors:
+    - Hinson Wong <hinson.wong@deltadefi.io>
+    - Tsz Wai Wu <tszwai@deltadefi.io>
+Proposed Solutions: []
+Discussions:
+    - https://github.com/cardano-foundation/CIPs/pull/?
```
```suggestion
    - https://github.com/cardano-foundation/CIPs/pull/1109
```
`this`

rphair (commented):
@HinsonSIDAN now that this is marked "Ready for Review" we'll add it as `Triage` which puts it on the agenda for our next CIP meeting (https://hackmd.io/@cip-editors/123).

In the meantime I think it'll be useful to get some co-review with this current CIP candidate (cc @qnikst @lehins @nc6 @Ryun1 @Crypto2099):
* https://github.com/cardano-foundation/CIPs/pull/1083

CPS-????/README.md
```diff
@@ -0,0 +1,137 @@
...
+
+The Cardano ledger accepts any valid CBOR encoding for transactions and on-chain data. While this flexibility was intentional to encourage ecosystem diversity, it has created significant interoperability challenges as the tooling landscape has matured. The same logical data can be encoded in multiple ways (map key ordering, integer encoding, definite vs. indefinite length, etc.), leading to different byte representations and transaction hashes.
+
+### Core Issues
+
+**Transaction Hash Instability**: When a transaction is passed between tools or wallets for signing, each may re-serialize it differently. Since transaction hashes are computed over CBOR bytes, logically identical transactions produce different hashes. This breaks:
+
+- Multi-signature workflows where each signer's wallet may re-serialize the transaction
+- Cross-tool transaction building where fee calculations depend on exact byte size
+
```
Ryun1, line 30:
```suggestion
- Hardware wallets, which require the keys in every map to be sorted from lowest value to highest.

```
this is one Ive encountered, which was frustrating to figure out

CPS-????/README.md
rphair: resolved with https://github.com/cardano-foundation/CIPs/pull/1109/commits/fae3905253f814620322137c749c8f2b8ab76dc4

rphair (commented):
@HinsonSIDAN we're leaving this `Unconfirmed` for now due to lack of current agreement & statements about why such serialisation would be considered "dangerous" (as per https://github.com/cardano-foundation/CIPs/pull/1109#issuecomment-3463024833) even though the node itself is doing this all the time (cc @fallen-icarus @Crypto2099).

@colll78 we collectively recalled that you've posted some things about this issue on X... can you give us the gist of how you stand on the issue & any other points that the community has made about it?

HinsonSIDAN:
@rphair I am drafting this CPS mostly due to the common perspective I gathered in 2025 Vietnam Builderfest and 2025 UPLC conference in edinburge. I believe it is a pain to vast majority of builders.

I am happy to keep this CPS unconfirmed indeed if that's not a too pain of an issue that people interested to put more comments in. 

Tho one more comment / observation. It seems like this issue has bugged so many Cardano developers from junior to middle level, that scares away many potential talents for our community. However, when we get to more senior in Cardano dev, usually we are experience enough to figure out workaround on issues caused by this problem. Like we in SIDAN Lab & DeltaDeFi built some stupidly complicated dev tools for it, where we shared in [2025 Edinburgh UPLC conference](https://www.youtube.com/watch?v=HEAAShtOM9Q&list=PLnPTB0CuBOBzSg9YF5ECeFpkQixm4Tv2Q&index=20)

image

Eventually, no one got interested enough to solve it for others. 

Sorry for this week CIP meeting I cannot squeeze time to attend, I would try to share more in next editor meeting. Appreciate the moderation!

CPS-????/README.md
HinsonSIDAN: @Ryun1  I did not put hardware wallets in since there is an active standard on it - https://cips.cardano.org/cip/CIP-0021. Perhaps you are referring to this one

rphair:
@HinsonSIDAN since you've already volunteered to attend the next meeting (https://hackmd.io/@cip-editors/124) as per https://github.com/cardano-foundation/CIPs/pull/1103#issuecomment-3589138386, I've also added this one to the agenda: keep in mind that this agenda will be packed due to requests to recap the Leios status, but at least we should have time for you to review the relationships above & reassure the technically oriented editors & observers about the "safety" of CBOR conversion.

qnikst:
Hello, sorry for jumping late on the party. From the perspective of the team who works on canonical ledger state serialisation, supporting serialised CBOR standard is a very good initiative, that we will strongly support and ready to participate in.

This CPS and CIP-0165 do overlap, or no be more concrete CIP-0165 can use the results of this intiative what ever they are. In CIP-0165 we want to use canonical representation, however as CIP-0165 does not put any restrictions on the ledger and on-wire data we just used what current CBOR RFC defines with small restrictions. 

In case if community align on the concrete implementation of the CBOR serialisation we will join the efforts and going to use that in the reference implementation and fix the rules in CIP. As for the current state we relied on the 
[RFC7049:3.9](https://datatracker.ietf.org/doc/html/rfc7049#section-3.9) and [RFC8949:4.2](https://datatracker.ietf.org/doc/html/rfc8949#section-4.2.2). With additional restrictions such as we forbid indefinite length structures, and this restriction works in the scenarios for the canonical state, I'm not sure if this restriction can be accepted in general, as it may harm streaming properties of the protocols.  Anyway that solution was not fixed in either CIP or merged implementation, so can discussed and changed.

From my personal perspective in addition to defining canonical and deterministic serialisation rules it would be nice to agree in the document on using patterns and tag, like [#258](https://github.com/input-output-hk/cbor-sets-spec/blob/master/CBOR_SETS.md) for set. This way it would be possible to rely more on CDDL files and validation tools, that at the moment can't enforce properties like "values in the set should be ordered".

The discussion about the safety and concrete examples should be held, especially because there is state of the art in CBOR RFCs about how to define deterministic and canonical CBOR.

colll78:
> @HinsonSIDAN we're leaving this `Unconfirmed` for now due to lack of current agreement & statements about why such serialisation would be considered "dangerous" (as per [#1109 (comment)](https://github.com/cardano-foundation/CIPs/pull/1109#issuecomment-3463024833)) even though the node itself is doing this all the time (cc @fallen-icarus @Crypto2099).
> 
> @colll78 we collectively recalled that you've posted some things about this issue on X... can you give us the gist of how you stand on the issue & any other points that the community has made about it?

My opinion is:
1. Best if we could do everything from scratch: Use protobuf or another efficiency optimized serialization format, or a completely ground up custom serialization format that we consider canonical and enforce that all information must be in this standard. Unfortunately, cbor is so engrained in the ecosystem and tooling that it would probably be impossible to get away from it now. 
2. Best given current circumstance: We should create a custom version of CBOR that we specialize for Cardano, standardize it and consider it canonical and enforce that everything must be encoded in this format. 

Any argument that enforcing canonical cbor is insecure is mute because every dapp & wallet and application on Cardano must enforce it anyway for hardware wallet compatibility, and simply deciding to ignore hardware wallet compatibility is not an option as a product. 

As for why CBOR is bad, just look at:
```
CBOR | 10 ms | 6.7 ms | 1691 KB
Protobuf | 10 ms | 7.7 ms | 890 KB
```

The size of data serialized with cbor is double the size of data serialized with protobuf, and the deserialization time differs by only 1 ms. That is not even close to a reasonable tradeoff. The size of serialized data is perhaps the single largest factor in terms of blockchain scalability. This data means that if everything was encoded with protobuf, the maximum throughput would be 2x whatever it is now. Even in a future when Leios exists this doesn't cease to matter, in-fact as throughput goes up this matters even more, because higher throughput means waste per message becomes much more important (because we have more messages).

Quantumplation:
Please don't make it custom; make it the existing canonical cbor standard that we've all been forced to use already 😅

rphair:
@HinsonSIDAN @fallen-icarus @colll78 @qnikst @Quantumplation editors would be ready to consider the candidacy of this CIP again, especially according the lesser-of-two-evils explanation in https://github.com/cardano-foundation/CIPs/pull/1109#issuecomment-3628811880.  If there's anything else we should be considering, please come to the CIP meeting tomorrow, if possible, and we'll get everything on record in any case... and @colll78 if there's any other suggestion of protobuf around CIPs going forward I'll connect it back here.

qnikst:
I'm not sure that two evils are as simple as stated.

Yes, CBOR is not the most efficient in the size and efficiency of encoding/decoding protocol, as least because it's schema-less and has to keep type tags. On the other hand it does allow interspectability and allows to define global types and rules for them (read tags).

However just schema is not enough to bring canonicity with either CBOR and protobuffers, as we should be careful enough how we store the data inside the fields and if it's canonical. For example reading the documentation of the protobuf I find the following property for the map type: `Wire format ordering and map iteration ordering of map values is undefined, so you cannot rely on your map items being in a particular order.`
And this is a killer for the canonicity, and what we are trying to solve in this CPS! So if we chose protobuf as a basis we would have to define canonical representation anyway, and it may be a way harder comparing to CBOR, as as far as I know there are no canonical protobuf efforts (on the contrary to CBOR).

In addition there are other sources for the protobuf criticism (e.g. https://reasonablypolymorphic.com/blog/protos-are-wrong/) whether to agree with all the points or not, choosing a protobuf as an option one may not be an easy choice that just works (tm) in our scenarios. As here we need a serialisation protocol that is consistent, canonical, allows to express required data structures and efficient enough. And CBOR is not a bad choice in this setting :)

As for defining our own CBOR, not sure about the phrasing here, as canonical/deterministic CBOR definition exists in its RFCs. So likely that we should carefully check if it meets our requirements, agree on contradicting points (e.g. one RFC proposes length-ordered keys order in maps and another value based), possibly set an additional restrictions e.g.  if we allow indefinite structures or not and verify that are all our choices are safe. As a result it may happen that the existing resulting rules can be fully based on existing RFC (so it's not a new CBOR). In addition we can define some additional rules like, always use sets tag for sets in does not change encoding and libraries compatible with existing RFCs.

Ryun1:
@qnikst 

> From my personal perspective in addition to defining canonical and deterministic serialisation rules it would be nice to agree in the document on using patterns and tag, like [#258](https://github.com/input-output-hk/cbor-sets-spec/blob/master/CBOR_SETS.md) for set. This way it would be possible to rely more on CDDL files and validation tools, that at the moment can't enforce properties like "values in the set should be ordered".

an effort for this has been made via CIPs towards this [CIP-114 | CBOR Tags Registry](https://github.com/cardano-foundation/CIPs/tree/master/CIP-0114)

colll78:
> I'm not sure that two evils are as simple as stated.
> 
> Yes, CBOR is not the most efficient in the size and efficiency of encoding/decoding protocol, as least because it's schema-less and has to keep type tags. On the other hand it does allow interspectability and allows to define global types and rules for them (read tags).
> 
> However just schema is not enough to bring canonicity with either CBOR and protobuffers, as we should be careful enough how we store the data inside the fields and if it's canonical. For example reading the documentation of the protobuf I find the following property for the map type: `Wire format ordering and map iteration ordering of map values is undefined, so you cannot rely on your map items being in a particular order.` And this is a killer for the canonicity, and what we are trying to solve in this CPS! So if we chose protobuf as a basis we would have to define canonical representation anyway, and it may be a way harder comparing to CBOR, as as far as I know there are no canonical protobuf efforts (on the contrary to CBOR).
> 

It will be harder for sure; my point wasn't that this is a good option now, but that it's something we should have done from the start (in which case we would already have made progress in creating a canonical representation by now). As I said, at this point cbor is far too engrained in our ecosystem and all downstream tooling for us to even really consider using this approach. 

Why is this approach the best option if we could start from scratch? The proof for this can be seen from observing other high performance blockchains. Solana uses a custom canonical binary wire format built from its Rust structs + shortvec varints; Aptos uses Binary Canonical Serialization, which as implied by its name is a custom canonical binary serialization format. Sui also uses BCS. Near uses Binary Object Representation Serializer for Hashing, **Borsh**, a canonical, non-self-describing binary format. Cosmos and all Cosmos SDK chains use protobuf. Algorand uses Canonical MsgPack.

There is a consistent trend amongst all of these high performance blockchains:
 1. Binary encoding - They avoid JSON/CBOR-with-tags style self-description for the performance-critical path.
 2. Canonical – Deterministic encodings (sorted map keys, minimal integer sizes, fixed field order) to prevent tx malleability.
 3. Non-self-describing – The node/wallet must already know the schema/struct definitions in order to decode.
 
The benefit of having the serialization format be self-describing is relatively useless given the abundant of modern techniques for circumventing the need for interspecability entirely (these techniques were mostly pioneered by other blockchains explicitly to solve this problem).  For example, even though Aptos' BCS requires the data type of a serialized value to the enforced by the application, this requirement is easily fulfilled by using unique hash seeds for each data type (a technique that they inherited from the Diem blockchain). Each of the high performance chains listed above employ their own techniques to circumvent any need for self-describing properties in their serialization formats.

So my conclusion is that either:
1. We picked the optimal serialization format back in 2017 and every other smart contract platform blockchain released since has been wrong. 
2. We did not pick the most optimal serialization format back in 2017, and since then blockchain research has advanced significantly, and techniques have become available that make some of the reasons we didn't choose non-self-descriptive serialization formats, obsolete. 

Personally, I believe that when every other blockchain on the market since 2017 is using non-self-descriptive serialization formats, and we are the only ones using a self-descriptive format, then it's more likely than not that we are the ones who made the non-optimal choice.

rphair (commented):
Thanks @colll78 @qnikst for helping us understand the CBOR debate, even if there are some unhappy consequences today that we can't do anything about.  My understanding & the general impression from yesterday's CIP meeting is that this statement of what to do about contemporary treatment of CBOR is still well-presented and usable in this CPS draft itself.

After further review at the meeting we've decided to progress this as a CPS candidate — @HinsonSIDAN please rename this directory to `CPS-0024`. 🎉

Some requests for expert review were emphasised at the meeting to try to keep this proposal concrete and usable (i.e. something that we can feel good about merging, not just a "wish list"):
* @qnikst sounded happy to review this as a complete document & the editors would look forward to ensuring all statements & recommendations are in order... and especially that the CPS is written so it can lead to CIPs that satisfy it.  @colll78 whatever you can provide as review would be equally appreciated for the same reasons.
* @lehins though this is not about any structure in the Ledger, we did think it would be nice to have your opinion about anything the CPS should contain that would make resulting CIPs more interoperable with the Ledger, and with each other, where CBOR is concerned... if there's someone else you'd prefer we tag about this particular issue, please feel free to pass it on.

CPS-????/README.md
```diff
@@ -0,0 +1,137 @@
+---
+CPS: ?
```
```suggestion
CPS: 24
```

CPS-????/README.md
```diff
@@ -0,0 +1,137 @@
...
+
+<!-- A more elaborate description of the problem and its context. This section should explain what motivates the writing of the CPS document. -->
+
+The Cardano ledger accepts any valid CBOR encoding for transactions and on-chain data. While this flexibility was intentional to encourage ecosystem diversity, it has created significant interoperability challenges as the tooling landscape has matured. The same logical data can be encoded in multiple ways (map key ordering, integer encoding, definite vs. indefinite length, etc.), leading to different byte representations and transaction hashes.
+
+### Core Issues
+
+**Transaction Hash Instability**: When a transaction is passed between tools or wallets for signing, each may re-serialize it differently. Since transaction hashes are computed over CBOR bytes, logically identical transactions produce different hashes. This breaks:
+
+- Multi-signature workflows where each signer's wallet may re-serialize the transaction
```
lehins, line 28: I see this as problem in this CPS, since reserialzaition should never happen for the purpose of signing and if it is then it is because tooling is doing incorrectly. I would concider it a bug  in the software regardless if canonical vs non-canonical serialization is used.
Whenever a payload is given to a program for the purpose of signing, that payload should not be mocked with. Same applies to transactions, if all you need is to sign the transaction either for the purpose of multi-sig or any other, transaction hash and serializaiton should not be recomputed and the original bytes must be retained and that is what needs to be signed.

CPS-????/README.md
```diff
@@ -0,0 +1,137 @@
...
+- Cross-tool transaction building where fee calculations depend on exact byte size
```
lehins, line 29: I don't quite follow this one. If there is any change to a transaction body then the bytes might change which could affect the fee. That is quite normal and expected.

If you mean that some tooling might want to change some part of the transaction body that it itself should not change the size if canonical CBOR was used (eg. change a required signer, i.e. swap one hash for another), then I can see it as an argument, but I don't understand why such tooling couldn't just recompute the minimum fee?

CPS-????/README.md
lehins: The biggest issue with ordering is that Haskell (or any other programming language) ordering is not guaranteed to always match the CBOR ordering.

In other words, I suspect enforcing CBOR ordering would make Ledger implementation a bit more complicated.

For example What is the ordering for a key that is a tuple?
What is bigger `-1` or `10`?
According to CBOR ordering that would be `-1`, while in Haskell it would be `10`.

lehins: The biggest issue with ordering is that Haskell (or any other programming language) ordering is not guaranteed to always match the CBOR ordering.

In other words, I suspect enforcing CBOR ordering would make Ledger implementation a more complicated and error prone.

For example What is the ordering for a key that is a tuple?
What is bigger `-1` or `10`?
According to CBOR ordering that would be `-1`, while in Haskell it would be `10`.

Not saying it is impossible, but something that needs to be taken under. consideration

CPS-????/README.md
```diff
@@ -0,0 +1,137 @@
...
+
+**Script Inconsistencies**: Smart contracts suffer from unpredictable script hashes, reference script mismatches across tools. The same compiled script may produce different hashes depending on the library used to apply parameters or cbor serialize the script.
```
lehins, line 31: I don't believe compiled scripts actually use CBOR serialization. @zliu41 will have a definitive answer, but from I know Plutus uses `flat` library for serialization.

CPS-????/README.md
```diff
@@ -0,0 +1,137 @@
...
+
+**Development Friction**: Developers face increased testing burden across multiple libraries and wallets, library-specific test fixtures, vendor lock-in risks, and debugging challenges that require logical rather than byte-level comparison.
+
+### Ecosystem Impact
+
+The lack of standardization creates:
```
lehins, line 37: CBOR is a standard, so this is not phrased correctly.

CPS-????/README.md
```diff
@@ -0,0 +1,137 @@
...
+
+- **Security risks**: Hard-to-diagnose bugs and complex audit requirements due to multiple serialization paths
```
lehins, line 39: In ledger we deal with serialization all the time and non of the bugs that we ran into had to deal with the fact that serialization is not canonical. I also never found debugging serialization to be that complicated.
Diagnosing logic bugs in ledger rules are are on order of magnitude more complicated than serialization, since by definition serialization does not have complication relationships within large types.
I don't buy that this is a security risk. I believe making a massive change of changing how we deal with serialization will open up the system to security risks because such transition have a huge potential for new bugs!

CPS-????/README.md
```diff
@@ -0,0 +1,137 @@
...
+- **Community overhead**: High support burden in addressing serialization issues and maintaining multiple strategies
```
lehins, line 40: This is only relative to all the tooling that cares about the current era, any tooling and the node itself will incur higher burden because of this CIP, because we do have to support current serialization forever while also introducing support for new serialization.

A great example is Byron era, which actually tried to use canonical CBOR and failed. We are now stuck with that logic forever without ability to reuse almost any of it!

CPS-????/README.md
```diff
@@ -0,0 +1,137 @@
...
+- **Adoption barriers**: Unpredictable behavior discourages enterprise adoption and increases new developer friction
```
lehins, line 41: I don't understand what is unpredictable in CBOR. It is a standard that 100% predictable.

CPS-????/README.md
```diff
@@ -0,0 +1,137 @@
...
+
+This problem has become urgent as sophisticated DApps require cross-tool interoperability, multi-signature usage grows, and community feedback (e.g., Cardano Builder Fest 2025) has identified this as a critical pain point.
```
lehins, line 43: It will take on the order of years for everyone to switch to canonical CBOR.
Ledger has already the whole of 2026 workload planned out. Even in optimistic case it would be sometime in late 2027 that this CIP could hypothetically could be implemented and then a whole lot more time for all the tooling to catch up. So, if this issue has urgency to it, I'd recommend putting more effort in fixing the tooling, rather than rushing dangerous changes like this on the chain itself.

CPS-????/README.md
```diff
@@ -0,0 +1,137 @@
...
+
+## Use cases
+
+<!-- A concrete set of examples written from a user's perspective, describing what and why they are trying to do. When they exist, this section should give a sense of the current alternatives and highlight why they are not suitable. -->
+
+**Cross-Library DApp Development**: A DApp developer builds transactions with Lucid in their frontend, but users sign with various wallets built on cardano-serialization-lib or pycardano. Canonical serialization ensures the transaction built equals the transaction signed.
```
lehins, line 47: Canonical serialization does not ensure it. It is only when everyone involved implements implementation of canonical serialization correctly.
Considering the situation today that tooling cannot correctly implement a standard today, I wouldn't bet that canonical serialization that "the transaction built equals the transaction signed". The only thing that can guarantee it is if original bytes are signed! And this CIP is actually promoting the opposite direction.

CPS-????/README.md
```diff
@@ -0,0 +1,137 @@
...
+
+**Script Hash Consistency**: A developer publishes a reference script on-chain, then references it from their off-chain code. Currently, locally computed script hashes may not match the on-chain version due to encoding differences. Canonical serialization guarantees hash consistency across compilation and deployment pipelines.
+
+**Library Maintainers**: Serialization library authors currently must support multiple encoding strategies for compatibility. With a standard, they can focus on a single canonical implementation, reducing maintenance burden and improving deserialization reliability.
```
lehins, line 51: Again, serialization libraries will have to support both non-canonical for historical data and canonical CBOR for new data. So, IMHO serialization library authors will be impacted negatively by this CIP

CPS-????/README.md
```diff
@@ -0,0 +1,137 @@
...
+
+1. **Establish a canonical CBOR standard**: A CIP that specifies deterministic encoding rules for all Cardano transaction and on-chain data structures, with clear guiding principles for choosing between encoding alternatives.
+
+2. **Achieve ecosystem adoption**: Widespread implementation across major serialization libraries (cardano-serialization-lib, pycardano, Lucid, Aiken, etc.) and wallets (Nami, Eternl, Lace, Yoroi, etc.), ensuring cross-tool interoperability.
+
+3. **Provide implementation guidance**: Comprehensive documentation including test vectors, reference implementations or validation tools, and migration paths for existing tooling.
+
+### Optional Goals
+
+4. **Ledger-level enforcement**: If community consensus supports it, implement validation rules in the ledger to guarantee compliance (requires hardfork and backward compatibility strategy).
```
lehins, line 65: Deserialization in Ledger is not part of the Ledger rules. It is a totally separate stage, when compared to transaction validation.
```suggestion
4. **Ledger-level enforcement**: If community consensus supports it, canonical deserilization must be correctly implemented in ledger to guarantee compliance (requires hardfork, backward compatibility and forward migration strategies).
```

CPS-????/README.md
```diff
@@ -0,0 +1,137 @@
...
+
+### Success Criteria
+
+This CPS is successfully resolved when:
+
+- A canonical CBOR serialization CIP reaches "Active" status with clear specifications
+- At least 80% of major libraries and wallets demonstrate compliance
```
lehins, line 72: Is there a way to quantify major libraries today?
Also, there is no mention of Ledger actually have it implemented:
```suggestion
- At least 80% of major libraries and wallets demonstrate compliance
- cardano-node has a hard fork ready foe a new Ledger era that changes its deserializers to canonical CBOR.
```

CPS-????/README.md
```diff
@@ -0,0 +1,137 @@
...
+- **Evolution mechanism**: Process for handling future hardforks and new ledger types
+- **Verification**: Test vectors or validation tools to verify implementation compliance
+
+## Open Questions
+
+### What are the guiding principles for choosing the canonical form?
+
+When multiple valid CBOR encodings exist, how should we decide which becomes canonical?
+
+- **Efficiency**: Minimize transaction size (e.g., smallest integer encoding, definite over indefinite length)
```
lehins, line 93: Definite and indefinite length encoding are have more efficient depending on number of elements. When there are less than 23 elements in an array definite length encoding is more efficient, while large count benefits from indefinite length encoding.
So, canonical will make encoding less efficient.

CPS-????/README.md
```diff
@@ -0,0 +1,137 @@
...
+- **Technical clarity**: Unambiguous encoding rules for all covered data structures
+- **Guiding principles**: Clear rationale for choosing specific encodings (efficiency, simplicity, adoption)
+- **Comprehensive scope**: Address transactions, scripts, datums, redeemers, and specify what is out-of-scope
+- **Path to Active**: Detailed adoption strategy including timeline, stakeholder coordination, and migration tooling
+- **Evolution mechanism**: Process for handling future hardforks and new ledger types
+- **Verification**: Test vectors or validation tools to verify implementation compliance
+
+## Open Questions
+
+### What are the guiding principles for choosing the canonical form?
```
lehins, line 89: If a custom canonical CBOR standard is to be designed then there is much higher chance that it will be implemented correctly by all of the tools. Grabbing an existing standard and accepting any potential drawback it could have (eg. non-optimal size) would be a safer bet IMHO.

lehins (commented):
I am very much against canonical CBOR to be enforced at the Ledger level and in general. Here is my reasoning:
* Drastically increases complexity and opens up possibilities for new bugs to be added due to such transition
* Non-canonical CBOR deserialization that we have today will have to be supported indefinitely anyways due to the on chain history.
* It is impossible to fully switch to canonical CBOR for on chain data because of plutus data. There are likely transaction outputs that are locked on chain today requiring data that is expected to have a non-canonical form. I.e. switching to canonical CBOR would lock those funds forever. 
* It is not solving any issues, since proper solution should be correct support of tooling of an existing standard.
* Promotes reserialization of transactions that ought to be immutable for the purpose of signing. Re-serialization for the purpose of signing is just a bad idea in general.

This is my professional opinion, which means if Cardano community as a whole is willing to sacrifice safety to make Cardano more user friendly, then Ledger team will have no chance but implement it regardless of my opinion.

Disclaimer: I am not a big fan of CBOR standard in general and if this discussion had happened at the design stages of initial version of Cardano my opinion maybe would have been different or at least I could have been persuaded to change my opinion much easier. Today, however, I very much believe this ship has sailed  and it would be much safer to support full CBOR instead of enforcing a drastic change like this.

CPS-????/README.md
```diff
@@ -0,0 +1,137 @@
+---
+CPS: 24
+Title: Canonical CBOR Serialization Standard
```
```suggestion
Title: Canonical CBOR Serialization
```
If we had a "Standard" for how to deal with this problem, we wouldn't need to make a Problem Statement about it. 😆  The word might be applicable in CIPs (though rare, since most CIPs _aren't_ standards, since generally merged when not yet `Active`), but even then perhaps not... if the Ledger & different realms of software end up treating CBOR standardisation differently.

CPS-????/README.md
```diff
@@ -0,0 +1,137 @@
...
+---
+
+## Abstract
+
+There is no canonical CBOR serialization standard in Cardano. While this is a delibrate design choice initially, standardizing it has growing popularity in Cardano developer community as evidenced by developer meetups such as Cardano Builder Fest 2025 hosted in Vietnam. This CPS outlines the motivation of the growing concern of fragmented CBOR serialization patterns across in the community.
+
+## Problem
+
+<!-- A more elaborate description of the problem and its context. This section should explain what motivates the writing of the CPS document. -->
+
```
```suggestion
```
(artefact from CPS template)

CPS-????/README.md
```diff
@@ -0,0 +1,137 @@
...
+- **Security risks**: Hard-to-diagnose bugs and complex audit requirements due to multiple serialization paths
+- **Community overhead**: High support burden in addressing serialization issues and maintaining multiple strategies
+- **Adoption barriers**: Unpredictable behavior discourages enterprise adoption and increases new developer friction
+
+This problem has become urgent as sophisticated DApps require cross-tool interoperability, multi-signature usage grows, and community feedback (e.g., Cardano Builder Fest 2025) has identified this as a critical pain point.
+
+## Use cases
+
+<!-- A concrete set of examples written from a user's perspective, describing what and why they are trying to do. When they exist, this section should give a sense of the current alternatives and highlight why they are not suitable. -->
+
```
```suggestion
```
(artefact from CPS template)

CPS-????/README.md
```diff
@@ -0,0 +1,137 @@
...
+**Library Maintainers**: Serialization library authors currently must support multiple encoding strategies for compatibility. With a standard, they can focus on a single canonical implementation, reducing maintenance burden and improving deserialization reliability.
+
+## Goals
+
+<!-- A list of goals and non-goals a project is pursuing, ranked by importance. These goals should help understand the design space for the solution and what the underlying project is ultimately trying to achieve.
+
+Goals may also contain requirements for the project. For example, they may include anything from a deadline to a budget (in terms of complexity or time) to security concerns.
+
+Finally, goals may also serve as evaluation metrics to assess how good a proposed solution is. -->
+
```
```suggestion
## Goals

```
(artefact from CPS template)

rphair (commented):
@lehins I appreciate your detailed review and your providing a realistically complete look at the problem.  I trust @qnikst / Tweag will also respond about their own considerations of all these issues.

Even if Ledger considerations prevent further progress, I don't regret the editors' assignment of a number and considering this a CPS "candidate" — since the standardisation problem will be discussed in the dev community no matter what, and this review thread should avoid fragmenting that discussion & might produce some acceptable resolutions here.

In any case @HinsonSIDAN the document would need a lot more detail & depth to reflect the considerations above.  I don't think it would be proper to rely on CIPs to address these concerns: otherwise we would get a dispersion of approaches, instead of the unified response we would need to produce particular CIPs from (perhaps by category: `Ledger`, `Tools`, `Wallets`).

We should also be prepared for the possibility that this CPS may remain unmerged if the Ledger and community developer points of view can't be reconciled.

CPS-????/README.md
yHSJ: This is correct. They're wrapped in a `script` type defined in the [CDDL](https://github.com/IntersectMBO/cardano-ledger/blob/master/eras/conway/impl/cddl/data/conway.cddl#L264-L275) within a transaction, but the compiled scripts themselves are just flat encoded bytes.

HinsonSIDAN: Not sure how to describe it more precisely, the issue we faced is that the different `uplc` libraries in different languages bahave differently. i.e. the Aiken `uplc` in Rust and HLabs `uplc` npm package implemented differently such that in `Mesh` users  get different script cbor if using different cores. It is a massive devexp issue

CPS-????/README.md
HinsonSIDAN: In general I agree with most of the points you mentioned so far & I agree the standardization should not built in ledger, so when selecting category of this CPS, we decided to make it `Tools` but not `Ledger`. Simply introducing a standard where tool builders can follow can solve the issue also

> I also never found debugging serialization to be that complicated.

btw you might not aware how far a distinct in technical knowledge amongst engineers in ledger team vs typical app builders😅 From tool builder perspective aka maintainer of Mesh where I face normal devs everyday, only less than 5% of devs understand how our CBOR system works and most of them can build DApps for Cardano

CPS-????/README.md
HinsonSIDAN: Yes, indeed this CPS is not intended to affect ledger

HinsonSIDAN:
@lehins Appreciate a lot of your review! I agree with most of your concerns if we see the canonical form is built in ledger. Originally I created this CPS with multiple categories since it is possible to be built / supported in different parts of community - https://github.com/cardano-foundation/CIPs/pull/1109#discussion_r2517078568. But eventually I think the `Tools` category is more appropriate since for example if it is enforced in ledger it might create some issues like permanently locking DApps which using hacks from different form of CBOR.

If we think about the standard is applied only to the level of tools, we can still solve the devexp issue without the need to change any ledger implementation. And the list, quantifying as if collecting most active transaction building libraries and its dependencies for example, is not too difficult to obtain for any CIP candidate also, such as:

```md
- Mesh
  - whisky
    - Aiken uplc
    - csl / pallas 
  - cardano-sdk-js
  - HLabs uplc
- Lucid evo
- pycardano
- apollo
- CSL
- CML
- etc ...
```

So eventually we can have a map of libraries who are actively following such standard, at least by default, so it becomes safe for 99% of DApp builders to never care CBOR again.

If think about the standard is built in Tools level only, I think some of the concerns above becomes irrelevant. Let me know how it feels if using this perspective to see this CPS, appreciate that!

CPS-????/README.md
lehins: > users get different script cbor if using different cores.

@HinsonSIDAN plutus core is not using CBOR for serialization. So, this issue albeit looks related, it has nothing to do with CBOR! It has to do with the fact that there is no standard for plutus core serialization at all.

In other words, if you'd like to use canonical or non-canonical CBOR for Plutus core serialization you could create a separate CIP for it, but I suspect that there will be some pushback there as well, since custom serialization that is currently in use is likely more efficient that CBOR.

CPS-????/README.md
lehins: > Simply introducing a standard where tool builders can follow can solve the issue also

I certainly have no say in this domain. If tooling developers in the community believe they can use another standard, I am not here to stop you.

That being said, if you don't believe they can understand current standard, what makes you believe they will understand a new standard?

CPS-????/README.md
lehins: Well, there is a section about it:

> Should the standard be enforced at the ledger level?

As I already pointed out, I would not want this CPS to affect Ledger either. I am a bit skeptical about standards that aren't enforced by the chain itself, but there are standards like these that proved themselves to work. So, at the very least, I would suggest adding all the drawbacks that I've mentioned to the section that suggests that this "standard should be enforced by the ledger level"

lehins:
>  I trust @qnikst / Tweag will also respond about their own considerations of all these issues.

@rphair With respect to canonical ledger state, it is a feature that is totally new, which makes making such decision much easier. Because of that and because @qnikst is working on a **canonical** ledger state, I believe using **canonical** CBOR would be a great idea.

> Even if Ledger considerations prevent further progress, I don't regret the editors' assignment of a number and considering this a CPS "candidate" — since the standardisation problem will be discussed in the dev community no matter what, and this review thread should avoid fragmenting that discussion & might produce some acceptable resolutions here.

I 100% agree with you. I've seen this idea of canonical CBOR being thrown around in all sorts of inappropriate places like twitter and such. This is a much better place to have a technical discussion about this

CPS-????/README.md
```diff
@@ -0,0 +1,127 @@
...
+License: CC-BY-4.0
+---
+
+## Abstract
+
+There is no canonical CBOR serialization standard in Cardano. While this is a delibrate design choice initially, standardizing it has growing popularity in Cardano developer community as evidenced by developer meetups such as Cardano Builder Fest 2025 hosted in Vietnam. This CPS outlines the motivation of the growing concern of fragmented CBOR serialization patterns across in the community.
+
+## Problem
+
+The Cardano ledger accepts any valid CBOR encoding for transactions and on-chain data. While this flexibility was intentional to encourage ecosystem diversity, it has created significant interoperability challenges as the tooling landscape has matured. The same logical data can be encoded in multiple ways (map key ordering, integer encoding, definite vs. indefinite length, etc.), leading to different byte representations and transaction hashes.
```
qnikst, line 22: As was discussed during  the editor meeting, the main priority (as far as I understood) is definition of the encoding that can be used in community and external tooling. And we do not expect to change onchain part and the ledger, because onchain data will remain immutable anyway and tools will have to support that anyway. 

So for the tools we can set a number of requirements or rules that will help library implementor. E.g.:
- when returning on-chain data (the one that may have historical or non-canonical format we can return it as a `bytes` blob (with cbor inside) or prefix it with a tag. This way the tools will be able to persist exact byte-size, order and keep hashes as is without additional effort no matter of the quality of the cbor libraries.
- if work with onchain data directly tools have to be able to read any format
- tools should use canonical format for storing data on chain

All those rules are possible examples, and should be discussed and defined in the future CIP.

Also those rules should not put an additional restrictions on the onchain data or existing ledger implementation. 
Though It would be nice if `external` interfaces of the ledger (debug or if ever there will be a non-debug) will talk to the world using canonical cbor, and of cause canonical ledger state have to follow the rules will be defined.

I think we should restate this paragraph to explicitly tell what part of the problem is defined here

CPS-????/README.md
qnikst: I would set this as the lack of standard canonical serialization implementation.
There are rules for dCBOR and cannonical CBOR that differ (for example in the order of the keys)

CPS-????/README.md
qnikst: I would say that it should not be a new standard but rather a set of the references to existing standards and tags that should (or for the tools must?) be used, like https://datatracker.ietf.org/doc/draft-mcnally-deterministic-cbor/10/ and mentioning the https://github.com/cardano-foundation/CIPs/tree/master/CIP-0114 and possibly adding more tags there. For example we can define optimal binary representation of utxos form ledger define binary serialisation format and introduce a tag.

If we do that then can apply those validations in verification tools and generate code for working with the encoding. 

Even if CPS solution will be CIP that just states that tools should use: dCBOR for encoding, should support non-deterministic input, use tag+binary for the on-wire data, that would be a good shift from the current status quo.

CPS-????/README.md
qnikst: I would say that there is a scenario: storing the data that is not on the chain that later should have exact byte-to-byte encoding, example of such scenario is canonical ledger state dump, we have CDDL for the format, but exact representation is not fixed by being on chain. So two separate implementation may write different binary encoding for example for coin if one uses canonical format while another doesn't, as a result binary representation will be different. 

I expect that there are other similar scenarios in tooling communication.

Just to note, I do not expect those changes to affect the ledger and fully agree with the problems that happen if doing so that were mentioned.

CPS-????/README.md
qnikst: It should be a part of the CIP discussion but I think there are solutions in the problem space for this problem. 

As far as I understand if we currently define an array in CDDL, it can have.either definite or indefinite length and that would lead to a different encoding and hashes of the value.

But if we add a tag, telling that it's an indefinite structure then we will bypass the problem, and such encodings will match. There can be other approaches to the problem.

CPS-????/README.md
```diff
@@ -0,0 +1,127 @@
...
+- Cons: Breaks backward compatibility; requires hard fork; existing tools and transactions may become invalid; potentially impractical migration burden
+
+**Not enforcing** (off-chain standard only):
+
+- Pros: No backward compatibility concerns; existing transactions remain valid; easier initial adoption
+- Cons: Voluntary compliance may be insufficient; fragmentation may persist; no guarantee of universal adoption
+
+### How should canonical serialization evolve with hardforks?
+
+When a hardfork introduces new ledger types or transaction fields, the CBOR encoding for these new structures must be decided. This raises critical workflow questions:
```
qnikst, line 112: I think that result of the canonical CBOR format should provide the set of rules, implementation tips, testing vectors that does not depend on exact types and transaction fields.

CIP will have to provide a process **that takes CDDL as an input and returns exact canonical representation**. 

So I think that success for CPS resolution will be if we can  make zero changes or types definition over the current existing processes where CDDL updates are defined.

(On the contrary to the canonical ledger state CIP where those new types should be registered as CDDLs)

CPS-????/README.md
colll78: It's not entirely true that they are just flat encoded bytes. They are flat encoded bytes wrapped in cbor. 

The reason the applyParams utilities are returning different script cborHex bytes across different offchain frameworks is likely because some of the tools actually apply the parameters via cek as an optimization (ie. inline at usage), and the other just applies the parameters via cborHex byte manipulation.

rphair (commented):
@HinsonSIDAN we have had less than 7 weeks of discussion on this, which ended in December of last year: more than 8 months ago.  The only facts that were settled in the discussion appear to be (cc @lehins @qnikst @yHSJ @colll78 @Quantumplation):
1. any suggestion of standardising CBOR can only be accomplished from a tooling perspective and can never hope to have any consistency with the Ledger;
2. there should be a common place to discuss this issue (with resolutions summarised in this CPS candidate) so DevX pain points are not scattered in blogs, social network posts, and PR/issue comments across many repositories.

However neither of these goals are being accomplished because (respectively):
1. there has been no further detail posted about how developers & agencies who would agree on a common serialisation would: depend upon it, use it, certify it, etc.;
2. feedback such as what @HinsonSIDAN you mentioned in your OP (the outcome of Cardano Builder Fest 2025) is not being accumulated & posted from: working groups, further events / seminars, or similar discussions on other channels.

So we do have a point of common interest but don't have a potential CPS because it is not evolving to accumulate the results of discussion and peer review:
* @HinsonSIDAN and or any other advocates: somebody would have to continue or take over this role of advocacy and start supplying the above information... and then I believe we could obtain a CPS addressing this "pain point" without it simply being one person's conception of how a solution would be characterised.

Without that, or without further comment, we would have to assume this is either just abandoned or perhaps beyond the author's (or anyone's) capability to draft a universally agreeable & productive problem statement: in either case, closing this PR. - cc @perturbing @Ryun1

colll78:
In future eras, only a single canonical cbor format should be allowed for transactions. There should not be multiple representations for the same data. 

> any suggestion of standardising CBOR can only be accomplished from a tooling perspective and can never hope to have any consistency with the Ledger;

I disagree. Txs in new eras should be enforced to use canonical encoding by the ledger.

rphair:
@colll78 then if you put the resolutions above with the your last statement, doesn't that make the Ledger serialisation of CBOR itself "canonical" with no good reason for anyone to deviate from it... and therefore this CPS just a "wish list" for a different serialisation that might be used between apps that _aren't_ creating Txs?

If so then I would be inclined to close this unless @HinsonSIDAN and/or other supporters of this idea can justify the existence of a second serialisation of CBOR — and commit to a related CIP that would very strictly define it, to exist side-by-side with the "Canonical" CBOR serialisation of the Ledger.
* If so then maybe another word than "canonical" should be used: since unless I've misunderstood it would be a **non**-canonical variant chosen for convenience, efficiency, interoperability, etc.

As you know I am not an expert in this domain but I _do_ need to define exactly what we have really been waiting for these last 8 months as the discussion has been left hanging.  The dev, Ledger, & CIP communities are gaining nothing by leaving this CPS draft in a pending state... unless anyone following the discussion so far thinks it would be helpful & truthful to simply merge it as-is (?) cc @qnikst @yHSJ @Quantumplation

@lehins you also have a very clear view of the motivations & pitfalls of this proposal and maybe we could get unstuck by your answering the question (adding whatever the tagged dev comunity experts think): _What exactly would the Cardano community lose if this CPS PR were closed & no CIPs solving this 'problem' were ever produced?_

PR #1110, opened 2025-10-29
CIP-???? | Sign transaction IDs together with guards

polinavino:
We propose the following change to the ledger signature checking mechanism : instead of signing just the transaction ID, users will now be required to sign the hash of the pair of (i) transaction ID and (ii) the hash of the guards listed in the transaction body. This way, key holders will have no need to inspect the transaction they are signing in 
order to be sure that the transaction satisfied its guards (if posted on-chain). 

---

([rendered latest proposal version](https://github.com/polinavino/CIPs/blob/polina/light/CIP-%3F%3F%3F%3F/README.md))


CIP-????/README.md
```diff
@@ -0,0 +1,131 @@
+---
+CIP: ????
+Title: Sign transaction IDs together with guards
+Status: Proposed
+Category: Ledger
+Authors:
+    - Polina Vinogradova <polina.vinogradova@iohk.io>
+    - Nicolas Henin <nicolas.henin@iohk.io>
+Implementors: []
+Discussions:
```
```suggestion
Discussions:
    - https://github.com/cardano-foundation/CIPs/pull/1110
```

CIP-????/README.md
```diff
@@ -0,0 +1,131 @@
...
+### CLI
+
+To support this change, the CLI will have to be modified to implement the new signing strategy.
+
+### Backwards Compatibility
+
+This change will be best suited for a hard fork (e.g. alongside Nested Transactions) that already does not have backwards compatibility 
+since it will not itself be backwards compatible. 
+
+## Future Intent DSL Development
```
```suggestion
### Future Intent DSL Development
```
I can see how this could be either a pending part of the _Specification_ or, as a thing not done (yet), kicked forward into the _Rationale_.  In either case this heading does need to be demoted to a subsection.

rphair (commented):
thanks @polinavino - new reviewers please refer back to the original submission of this idea & why it's distinct from CIP-0112:
* https://github.com/cardano-foundation/CIPs/pull/1105

@colll78 the CIP editors at this week's meeting were keen to you get your audit / security view of this CIP once prepared.  Also @GeorgeFlerovsky you'd expressed an interest and of course @lehins we would look forward to your view on how all this fits together.

CIP-????/README.md
```diff
@@ -0,0 +1,151 @@
...
+
+## Abstract
+We propose the following change to the ledger signature checking mechanism : instead of signing just the transaction ID, users
+will now be required to sign the hash of the pair of (i) transaction ID and (ii) the hash of the guards listed in the transaction body.
+This way, key holders will have no need to inspect the transaction they are signing in 
+order to be sure that the transaction satisfied its guards (if posted on-chain). 
+
+## Motivation: why is this CIP necessary?
+
+This CIP (which depends on CIP-0118) serves both a practical and a conceptual purpose.
```
Ryun1, line 24:
```suggestion
This CIP (which depends on [CIP-0118 | Nested Transactions](https://github.com/cardano-foundation/CIPs/pull/862)) serves both a practical and a conceptual purpose.
```

CIP-????/README.md
```diff
@@ -0,0 +1,151 @@
...
+During transaction deserialization, the `MemoBytes` mechanism (an abstraction for a data type that encodes its own serialization) will be 
+used to store the bytes of the CDDL field containing the guards of the transaction. 
+The following values will be computed in the process of transaction processing : 
+
+- `guardsHash`, which will be computed by hashing the bytes of the guards, and 
+- `txidAndGuards`, which will be computed by hashing the concatenation `(txid ++ guardsHash)`
+
+Signature checking will now check that each key signed `txidAndGuards`.
+
+** Note ** There are other options for how exactly to compute the final hash of the data that will be signed. E.g.
```
Ryun1, line 64:
```suggestion
**Note** There are other options for how exactly to compute the final hash of the data that will be signed. E.g.
```

fallen-icarus:
The motivation for this CIP seems to be that users will remove the fee from a tx built by the service provider. But this doesn't seem to be true in reality. Wallets like Vespr [charge fees](https://docs.vespr.xyz/vespr/user-guide/faq/swap/swap-fees-explained) already and users do not remove this fee. For non-technical users, they don't even have the necessary know-how to remove the fee. Technical users do, but they would likely have their own infrastructure anyway and have the know-how to get around *any* service fee. 

So is this CIP actually solving a real problem? If the target user is the non-technical group, what is wrong with the status quo? Vespr seems to have no issue charging fees.

polinavino:
@fallen-icarus i thought about this [point](https://github.com/cardano-foundation/CIPs/pull/1110#issuecomment-3576549592) a bit and the following comes to mind :
- with the feature proposed here, the user does not need to deserialize a transaction. They do need to be able to construct and serialize an observer, and they also need to be able to hash.
- without this feature, the user would need to deserialize the entire transaction just to check that the right guard script is included. This is not only making it more complicated and error-prone for a non-technical user, but also making it way easier to get around the fee (as they have to deserialize the transaction anyways)

fallen-icarus:
@polinavino I think you are confusing "who is doing what". Wallet software needs to do what you say *on behalf of* non-technical users. 

> They do need to be able to construct and serialize an observer, and they also need to be able to hash.

No non-technical user can do this...

So your actual threat model is a malicious wallet provider removing the fees *on behalf of* its non-technical users. But again, there is no evidence of this happening today on Cardano. Aggregators like SteelSwap add fees into the txs they build for wallets like Vespr, **and users keep the fees in the txs**.

Sorry, but I really don't think this is solving a real problem. There seems to be nothing wrong with the status quo where a service can add its fees directly to the tx and send that tx to the user's wallet unshielded. There is no evidence of users messing with these transactions to remove fees.

rphair (commented):
@polinavino @fallen-icarus as per the post- CIP meeting routine I'm updating this thread & the PR status based on our meeting resolutions, and I think the last few comments outline the considerations fairly well.

Since it's still not agreed how weighty these considerations are, I think our resolution not to confirm this CIP as a candidate yet still stands... which I think can be resolved by addressing all the possibilities in the CIP document itself (including manual work for fee avoidance, "malicious" wallet apps, etc.) because it will be likely still serve as a useful CIP even if less necessary to deploy based on social / commercial factors.

polinavino:
It seems like the architectures/trust models you (@fallen-icarus ) point to are sufficiently different from the design we propose here. They seem to be examples of ways a single provider can design an app such that the provider gets paid when this app is used. I can't tell what exactly is happening, but this is how i understand things 
- perhaps the app makes it quite difficult for the user to mess with the transaction the app produces before signing it. This is ok, because the user only needs to trust code they themselves run locally
- the user can check that they are running an authentic version of the app (and therefore trust that code)
- the app queries the chain using blockfrost (trust needed here)

The protocol we propose here is fully decentralized and doesnt not require any trust : 
- requires the service provider (could be anyone with a node, not necessarily trusted), run on a different computer, to trust the LC. The step of concealing the transaction body from the LC is to protect one untrusted party from another. 
- The signature on the txid ++ guard protects the LC from a malicious SP. 
- The thing you have to pay for is not the use of the app, but rather the query answer. As this would be open source, it would be extremely easy for someone to modify the code once so that nobody pays for the tx-construction service ever again (as opposed to, e.g. someone writing an app *like* VESPR but that doesnt pay VESPR)

The design we propose is very much in line with blockfrost decentralization (e.g. via icebreakers infrastructure). It provides an alternative to icebreakers that is suited for a different purpose.

> > They do need to be able to construct and serialize an observer, and they also need to be able to hash.

> No non-technical user can do this...

Thats the entire design of the client. The workflow is
- user selects intent template from supported ones (eg "purchase a ticket")
- user fills out template (their spending keys, change address, etc.) 
- LC outputs serialized guard `g`
- LC sends it to SP

Once LC receives a `txid` in response from SP
- LC computes `tx_g = hash (txid ++ hash g)` 
- LC signs `tx_g` and sends it to SP

fallen-icarus:
> They seem to be examples of ways a single provider can design an app such that the provider gets paid when this app is used.

This is not correct. It is multiple service providers each adding a fee to the user's tx. Consider a DeFi swap:
1. A user of Vespr wants to make a swap so they get an offer from a DEX aggregator like SteelSwap. This offer is effectively a tx that contains a service fee for SteelSwap.
2. Vespr looks at the tx and adds their own service fee to the tx. The tx now has fees from two different service providers.
3. Vespr presents the final tx to the user to sign.
4. The user signs it and Vespr submits it to the network.

This one tx is paying both Vespr and SteelSwap despite neither service fee being enforced by a smart contract.

fallen-icarus:
> The design we propose is very much in line with blockfrost decentralization

So is the status quo! If Vespr added a third service fee (for a decentralized Blockfrost) to the swap tx, you are assuming users will remove this fee. But there is no evidence to support that concern, and I personally do not share that assumption. You are asking to make a change to the Cardano ledger without first proving that removing service fees is actually an issue.

polinavino:
Ah, i see. Thanks for clarifying how VESPR does things. I am really surprised by the argument that "since a vulnerability has not yet been exploited for social reasons (maybe social pressure, or moral convictions, or laziness?), it is therefore fine to not address it". 

VESPR chose to take that risk and it is working fine for now. In our model, we would be forcing every single service provider to take on that risk. The strategy we propose allows a very easy way to eliminate this issue with an extra hash computation. We are working towards creating sustainable economics for blockchain infrastructure services through cryptography rather than trust.

fallen-icarus:
> ... "since a vulnerability has not been exploited yet ... it is therefore fine not to address it".

You seem to be missing my point. You are **assuming** this is a vulnerability! I do not agree with you at all. Don't forget that Ouroboros is based on a 51% honest majority. By your logic, Ouroboros has an unpatched vulnerability. Why is it okay for Ouroboros to rely on an honest majority assumption but not okay for the funding of a decentralized LC?

Instead of relying on an honest majority assumption, you are trying to use cryptography that creates a terrile UI/UX. It is so bad, that I doubt anyone will actually use the LC you are trying to create. You are asking people to be okay with signing txs they can't see inside of with the promise that observer scripts will protect them. But IMHO this is not realistic. Marlowe failed because end-users do not want to write custom smart contracts; making it more user-friendly did not change that. End-users want to use standardized smart contracts. Can general standardized observer scripts be used to protect end-users from all possible malicious behavior? I doubt it.

I strongly believe your LC is better off relying on an honest majority assumption than trying to completely eliminate the need for trust with its current approach. What I keep trying to stress is that this honest majority assumption may actually be safe to make. If it is, we don't need to change anything about Cardano.

polinavino:
A repercussion-free opportunity for a single user to act dishonestly and gain Ada as a result is not the same thing as 51% agreement that it is ok for them to do the dishonest action. 

As for a "script protecting end-users" - is this a general argument against the idea that it is possible to translate a template into write reliable Plutus code? Or that some malicious or poorly written software will do this badly? The latter applies to VESPR too - their software could easily be showing the user a transaction that's different than what they're actually signing. Of course, this needs to be tested and analyzed. 

The transaction templates themselves will look mostly like conditions on produced and consumed values. The goal here is much much simpler contracts than marlowe, and automated Plutus code generation from single-expression templates such as this `Registration` (rather than general-case financial contracts), which the participant needs to fill out with they own PK hashes, etc., 
`GetTicket : Participant x EventOrganizer x Price x DREp -> Registration`
We are still discussing the details here, but it will be much simpler than Marlowe (in particular, because these contracts are not concerned with the evolution of the state of an on-chain program).

A more general goal of this work is to support intents on Cardano. We already have a way to express an intent - a Plutus script. Now, we need a mechanism that separates intents from their fulfilment, i.e. a way to decouple a transaction from the Plutus script intent.

fallen-icarus:
@polinavino Let me start over because your reply makes me think there is still a misunderstanding.

Imagine a decentralized Blockfrost needs to make 1,000,000 ADA per year to be financially sustainable. Let's also imagine an average of 5,000,000 txs (that use Blockfrost) per year. Naively, we could just say that every tx needs to pay a Blockfrost fee of 0.2 ADA to be sustainable, but this won't work because free-loaders will avoid the fee. 

However, if we assume that at least 51% of txs pay the fee, then charging 0.4 per tx will make Blockfrost financially sustainable: 49% of txs will strip the fee but the other 51% will pay it. The following table shows the relationship between the percent of txs that pay and the required fee/tx for Blockfrost to be financially sustainable:

| Honesty | Required Fee/Tx |
|:--:|:--:|
| 100% | 0.20 ADA |
| 66% | 0.31 ADA |
| 51% | 0.40 ADA|
| 33% | 0.61 ADA |

If you expect only 33% of users to respect the fee, then Blockfrost would still be financially sustainable by charging 0.6 ADA without requiring your method of obscuring the tx. I personally expect greater than 51% honesty which means charging 0.3-0.39 ADA would work. I keep pointing out evidence that honesty is currently quite close to 100%.

You are calling the lack of cryptographic enforcement a *vulnerability* but this is just being cynical of humanity. You think users will eventually start being dishonest. I do not share this cynicism. While free-loaders certainly exist, I believe the majority of people will act collaboratively and honestly. So I do not see the lack of cryptographic enforcement as a vulnerability. I do not think we need to make the UX worse for everyone just to protect against the free-loaders who are a small minority. It is very similar to taxes, some people do evade taxes, but most people are actually in favor of paying taxes; it is the *magnitude* of the taxes they disagree on.

> ...  is this a general argument against the idea that it is possible to translate a template into write reliable Plutus code?

No. As I said, I don't think users want to create custom plutus scripts for every tx. It doesn't matter how user-friendly you make it - they won't use it.

---
To be clear, I'm not against your LC implementation. But I am against making any changes to Cardano before you prove your LC works and is accepted by users. If we make this change *before* you validate the idea with the market, what happens if the market rejects your LC like I am predicting it will? We can't easily undo this change. It is significantly better to first prove your model works *before* making a change to Cardano for it.

If you come back a year from now with proof I'm wrong, I will gladly back this CIP. But until then, I'm against this CIP since its only use case seems to be an LC model that I think will be rejected by the market. 

FYI You are not the only person working on financial sustainability of decentralized services. I am also working on it (with CIP-159) and I'm relying on the concepts in the table above. How about we do experiments to see which method is better before we make an irreversible change?

rphair:
@lehins this PR ended up `Unconfirmed` because basically reviewers couldn't get consensus that there was a motivation for it.  Editors had tagged you for review early on and would still look forward to it; not only to know whether it covers the same CIP-0118 / CIP-0112 issue recently identified here (it [sounded to me like it might](https://github.com/cardano-foundation/CIPs/issues/1148#issuecomment-3875508110)):
* https://github.com/cardano-foundation/CIPs/issues/1148

... but also generally to know what to do with this CIP itself & how to get it back into review.

lehins:
> not only to know whether it covers the same CIP-0118 / CIP-0112 issue recently identified here (it https://github.com/cardano-foundation/CIPs/issues/1148#issuecomment-3875508110):

That issue is unrelated to what is being discussed here and in fact there is no real issue there as I mentioned in my comment: https://github.com/cardano-foundation/CIPs/issues/1148#issuecomment-3876372936

> but also generally to know what to do with this CIP itself & how to get it back into review.

I personally do not see any usability for the changes proposed in this CIP and their relevance to CIP-112 or CIP-118, but I do not want to be a sole decider whether feature should or should not go into Ledger. For this reason I suggested @polinavino to create a separate CIP, instead of including these changes into CIP-118, so that a larger audience can review the changes proposed and see if anybody else out there can see any merit in changes being proposed. I will be happy if the community decides one way or another. If it is indeed useful, then we'll implement it, but as far as I am concerned, I am siding with @fallen-icarus, I don't see a real use case for this CIP.

rphair:
Thanks @lehins for confirming the separation of these issues, since your https://github.com/cardano-foundation/CIPs/issues/1148#issuecomment-3875508110 only stated that issue was related to CIP-0112 itself (already obvious from the context).

After 2+ months of not hearing a commonly accepted argument to proceed, @polinavino I think it is appropriate to label this `Waiting for Author` until that can be established.  If we can help by putting this on the Review agenda at a CIP meeting please say so: if so, it would help to list the points of disagreement (since editors might be bad at digesting this) so we can present them to other reviewers & observers.

lehins:
> since your https://github.com/cardano-foundation/CIPs/issues/1148#issuecomment-3875508110 only stated that issue was related to CIP-0112 itself (already obvious from the context).

Maybe I should rephrase my comment somehow, but my comment stated that there is no issue at all (I'll be discussing with @carlostome a need to close that ticket) and the fact that @carlostome was talking about the original CIP-112, i.e. it had nothing to do with this #1110 or #1105.

polinavino:
The light client design this CIP was instrumental for is not being developed currently. I will follow up on this in a few days to make sure, but I will probably just close this PR.

GeorgeFlerovsky:
I think that this CIP's motivation aligns with broader needs in several areas of Cardano, but its solution is too ad-hoc and burdensome for the other stakeholders besides users writing local Plutus scripts to express their intent.

The essential problem, as I see it, is:

- Hashes are simple and sufficient to pin-down the contents of a block body, tx body, tx metadata, tx script witnesses. etc.
- However, using hashes to pin contents makes it impractical for anyone to concisely prove something about blocks, txs, tx metadata, or other entities of interest without processing those entities entirely in the proof.

I think there is a general trend of usecases needing Cardano to gradually move away from hashes toward cryptographic commitments (e.g. Merkle trees, BLS/KZG commitments, etc.) that facilitate concise proofs about select elements of interest in the larger data structure summarized by the commitments. 

### Light wallets

@polinavino and her colleagues encountered this in her light wallet architecture, where she found it hard to facilitate users who want to concisely express their intent about select parts of a transaction.

She didn't ask for something like a Merkle tree to be used instead of a hash to pin transaction contents, probably because it's a more significant refactor of the Cardano ledger. Instead, she asked for the bare minimum that would solve her problem: add another hash at the top level that pins the guard scripts. This solves her problem, but it doesn't help the others, and now everyone has to worry about signing a pair of hashes -- as @lehins and @fallen-icarus highlighted.

If Cardano transactions were to have crytographic commitments that allow concise membership proofs of transaction elements, then @polinavino's could just get such concise proofs from servers that the intent guard scripts are included.

### Tx metadata in Plutus

@SebastienGllmt's Paima Studios had, a couple of years ago, needed to access some of the CIP-25 metadata of a transaction in a Plutus script. To do so, their Plutus script reconstructed the tx body hash from all the serialized components that go into it, which then allowed them to prove something about the metadata. As mentioned in their blog post about this, there are significant trade-offs and limitations:

- https://blog.paimastudios.com/cip25-plutus/

### Optimistic rollups

Midgard (led by @colll78) is an optimistic rollup that processes blocks of Cardano transactions offchain and then posts just those blocks' headers onchain. Midgard's onchain scripts require collateral bonds from the people posting those blocks, and they make the block header wait in a queue for 7 days to give anyone from the public a chance to prove the headers fraudulent.

Midgard leverages Cardano's determinism to reduce the size of fraud proofs: they only need to show a path from the defendant's block header to a transaction, and some transaction component, and then they demonstrate that the component is fraudulent. As Paima Studios' brute-force approach above shows, this is impractical to do with the traditional hashes of blocks, transactions, etc.

Therefore, Midgard had to introduce its own data structure for L2 blocks, where everything is merkelized:

- The block header contains Merkle roots of its transactions. (Also the deposits and withdrawals from/to L1, but that's not relevant to Cardano L1).
- Each transaction is augmented with additional hashes/merkle roots to make it easier to prove that a particular input, output, script, or other element is contained in the transaction.
- The execution trace of each script in the transaction is merkelized to make it easier to prove that a particular execution step is part of the trace.

Midgard is a bit extreme as an example of Merkelizing the Cardano ledger, but it's unavoidable if Midgard wants its onchain Plutus scripts to be able to verify concise proofs about blocks that are only available offchain.

- https://github.com/Anastasia-Labs/midgard

### Inter-blockchain bridges

@matteocoppola @nau and others are working on bridges between Cardano and other blockchains, in particular Bitcoin.
As I recall from discussing with them, to reduce some of their trust assumptions, they need the ability to concisely prove that a certain transaction occurred in Cardano. Perhaps, similar to Midgard, they may also need to prove that something specific happened in that transaction.

Following the entire chain of block hashes from genesis is impractical. Mithril is an improvement because it's a shorter chain of block hashes, but it's still recursive to genesis, which perhaps also makes it impractical for these bridges.

I don't know what the solution is for these bridges, but I'm bringing them up because they too feel the pain from hashes.

- https://github.com/FluidTokens/ft-bifrost-bridge

### Conclusion

Overall, I think there needs to be a broader discussion about these painpoints with hashes, which could hopefully lead to some systematic solution that helps many of the different stakeholders/usecases without burdening any of them too much.

It might be the case that what works to replace tx hashes is different from the replacement for block hashes, etc. However, we should still consider the problem holistically.

P.S. sorry for not participating in this issue thread for so long.

colll78:
> https://github.com/cardano-foundation/CIPs/pull/1110#issuecomment-3934863427 (**ed:** omitted entire comment directly above, for readability of discussion)

https://github.com/cardano-foundation/CIPs/pull/964 

The above CIP is the answer to many of the challenges that builders are facing that you enumerated so nicely above.

rphair (commented):
@polinavino following up on your https://github.com/cardano-foundation/CIPs/pull/1110#issuecomment-3894884188 (nearly 3 months ago) the editors have to assume that you intended to close this — due to abandonment of the related design — and just never got around to it.

Of course feel free to reopen this if that work is proceeding & you believe it will lead to a refinement of this CIP & a continuation of its review.  In any case the latest comment by @GeorgeFlerovsky, plus @colll78's response, confirm that use cases relevant to this proposal could also be served by this CIP still under review:
* https://github.com/cardano-foundation/CIPs/pull/964

Issue #1111, opened 2025-11-02
Proposal to Match Parity with CIP-0025 in CIP-0068 (New CIP?)


thaddeusdiamond:
Wanted to solicit known contributors' opinions (have reached out on Twitter as well).  I'm running into a problem with large collection mints (5-10k and over) where the amount of ADA being locked by minUTxO in the smart contract is gargantuan with CIP-0068.  Now that ADA is closer to $1 my artist clients are getting frustrated by the increased costs of their mints (10-20% tax sometimes for cheap big collection mints).

I want to allow the metadata standard for CIP-0068 to optionally match CIP-0025 through a new version where metadata can either be:
1) The direct metadata for the NFT (backwards compatible)
2) A 721-style CIP-0025 metadata (using some sort of binary reserved key or extra data indicator to avoid naming conflicts)

This would also have the added benefit of easier migration from CIP-0025 to CIP-0068 (an occassional use case I support for my artists).

What do people think: should this be a new CIP or version 5 of CIP-0068?  It would be pretty cross-cutting and apply to all of 222/333/444 tokens.

Tagging @rphair @Quantumplation @Crypto2099 @KtorZ @alessandrokonrad @SmaugPool for feedback

thaddeusdiamond:
See #1112 for an example of how we could do it without a new CIP

rphair:
@thaddeusdiamond I can't contribute technically to decide the best path forward, but I'd like to see active discussion on this issue and so we'll try to highlight it at the next CIP meeting (https://hackmd.io/@cip-editors/123)... though this will be the first meeting in 4 weeks so we may have a full agenda.  But one of our CIP editors has a very active interest in these issues and I expect to get some very productive discussion from there (if we don't have it already by then).

> What do people think: should this be a new CIP or version 5 of CIP-0068? It would be pretty cross-cutting and apply to all of 222/333/444 tokens.

Our convention is to require _both_ a new version in CIP-0068 (so that existing mints can guarantee compatibility with the new label) _and_ a new CIP even if only a new label is added: https://github.com/cardano-foundation/CIPs/blob/master/CIP-0068/README.md#extending--modifying-this-cip — and a v5 of the CIP would also be an opportunity to indicate any changes to the already existing labels beyond just a version bump.

rphair:
[...]

PR #1112, opened 2025-11-02
CIP-0068 | Support multi-asset metadata a la CIP-0025

thaddeusdiamond:
Add version 4 metadata support with 721-ERC-style mappings to NFT (222), FT (333) and RFT (444) standards, ensuring consistency across all asset classes.

Changes:
- Add metadata_field union type supporting both direct and 721-map formats
- Add __RESERVE_KEYWORD_721_V4__ marker for 721-style detection
- Update retrieval steps to handle both metadata formats
- Support version 4 in FT standard
- Support version 3/4 in RFT standard
- Add 721-style JSON example for NFT standard
- Update pattern descriptions with asset_name terminology
- Fix URI support to include [* bounded_bytes] for RFT

### DIFF
--- a/CIP-0068/README.md
+++ b/CIP-0068/README.md
@@ -699,6 +699,45 @@ make use of this off-chain and on-chain.
 The security for the link is derived from the minting policy itself, so it's important to write the validator with the
 right constraints and rules since this CIP solely defines the interface to keep flexibility as high as possible.
 
+### Rationale: Version 4 Nested Map Format
+
+The current design of CIP-0068 inadvertently creates a significant cost barrier for large NFT collections due to minimum-ADA requirements. As ADA approaches parity with USD, artists and platforms minting 5,000–10,000+ on-chain assets are increasingly impacted by the **minUTxO amplification effect**: the metadata-bearing reference scripts and datums mandated by CIP-0068 can cause **10–20% of project budgets** to be consumed purely by ADA locked in outputs, rather than by art production or ecosystem growth.
+
+This cost profile wasn't as visible when ADA was significantly cheaper, but it has now become an operational friction point across multiple creator workflows.
+
+The proposal aims to **reduce the minUTxO footprint** for CIP-0068 assets by allowing an **optional alternate metadata shape** that mirrors CIP-0025 (721-style metadata) while preserving CIP-0068's semantics. Concretely:
+
+#### 1. Dual-format Metadata Reduces Unnecessary ADA Locking
+
+Allowing a CIP-0025-shaped metadata payload means:
+
+* The on-chain metadata can consolidate across multiple reference tokens.
+* The minUTxO requirement is correspondingly reduced.
+* Large-scale mints return to economically sensible territory, especially for low-cost or promotional collections where the metadata weight provides no additional value to creators.
+
+This addresses a real-world pain point reported by multiple production-level minting pipelines, not just isolated cases.
+
+#### 2. Improved Migration Path from CIP-0025 to CIP-0068
+
+A sizeable portion of the ecosystem still lives on CIP-0025, particularly older marketplaces and archived collections. When artists revisit or reissue older collections, the transition path to CIP-0068 is currently friction-heavy, requiring metadata restructuring and toolchain updates.
+
+Supporting a CIP-0025-compatible metadata envelope inside CIP-0068:
+
+* Enables forward migration without metadata rewrites.
+* Preserves backward recognizability.
+* Reduces engineering cost for wallets, indexers, and minting tools.
+
+#### 3. Strict Optionality Preserves Backward Compatibility
+
+This proposal does **not** modify or deprecate existing CIP-0068 patterns.
+Instead, it introduces an explicitly flagged alternate metadata payload, using a reserved binary key or indicator to avoid collisions with existing fields.
+
+This allows:
+
+* Full backward compatibility.
+* No disruptive changes for current indexers or marketplaces.
+* Opt-in use where economic relief is necessary.
+
 ### Backward Compatibility
 
 To keep metadata compatibility with changes coming in the future, we introduce a `version` field in the datum.
--- a/CIP-0068/README.md
+++ b/CIP-0068/README.md
@@ -165,10 +165,10 @@ Example:\
 
 This is either:
 
-1. A low-level direct representation of the metadata, following closely the structure of CIP-0025
-2. A 721-ERC-style token mapping exactly matching the structure of CIP-0025
+1. A low-level direct representation of the metadata, following closely the structure of CIP-0025 for individual asset metadata (name, image, files, etc.)
+2. A nested map structure following CIP-0025's organization with policy_id and asset_name as keys, where the outer map contains a "721" key whose value maps policy_id to asset_name to metadata
 
-All UTF-8 encoded keys and values need to be converted into their respective byte's representation when creating the datum on-chain.
+All UTF-8 encoded keys and values need to be converted into their respective byte's representation when creating the datum on-chain. In the nested map format (option 2), the "721" key, policy_id keys, and asset_name keys must be encoded as bounded_bytes (following CIP-0025 version 2 conventions).
 
 ```
 files_details =
@@ -194,23 +194,18 @@ metadata =
 
 metadata_map = 
   {
-    __RESERVE_KEYWORD_721_V4__ : bool, ; Long name to avoid conflict
-
-    "721":
+    "721" : bounded_bytes => ; The "721" key encoded as UTF-8 bytes (0x373231)
     {
-      ; As many of the below as needed based on which reference tokens are added
-      "<policy_id>":
+      * policy_id : bounded_bytes => ; Policy ID encoded as bytes
       {
-        "<asset_name_1>" : metadata,
-        ; ...
-        "<asset_name_n>" : metadata,
+        * asset_name : bounded_bytes => metadata ; Asset name encoded as bytes (without label prefix)
       }
     }
   }
 
 metadata_field = 
-  metadata          ; For backwards-compatibility in v4 onward, a map itself is valid metadata
-  / metadata_map    ; Multi-metadata is now supported as in CIP-0025
+  metadata          ; Direct metadata format (versions 1-3)
+  / metadata_map    ; Nested map format with "721" key (version 4)
 
 ; A valid Uniform Resource Identifier (URI) as a UTF-8 encoded bytestring.
 ; The URI scheme must be one of `https` (HTTP), `ipfs` (IPFS), `ar` (Arweave) or `data` (on-chain).
@@ -260,7 +255,7 @@ Example datum as JSON (direct metadata):
 }
 ```
 
-Example datum as JSON (721-style):
+Example datum as JSON (nested map format):
 
 ```json
 {
@@ -307,9 +302,9 @@ to lookup. The steps are
 1. Construct `reference NFT` from `user token`: `d5e6bf0500378d4f0da4e8dde6becec7621cd8cbf5cbb9b87013d4cc.(100)TestToken`
 2. Look up `reference NFT` and find the output it's locked in.
 3. Get the datum from the output and lookup metadata by going into the first field of constructor 0.
-4. Determine whether this is a direct metadata or the CIP-0025 map style using the `__RESERVE_KEYWORD_721_V4__` map value (if present, default false)
+4. Determine whether this is direct metadata (map without "721" key) or nested map format (map with "721" key).
 5. Convert to JSON and encode all string entries to UTF-8 if possible, otherwise leave them in hex.
-6. For multi-map style metadata, return only the JSON values inside the matching path: `721`->>`policy_id`->>`asset_name`
+6. For nested map format metadata, return only the metadata inside the matching path: map["721"][policy_id][asset_name], where policy_id and asset_name are matched without the asset_name_label prefix.
 
 ##### Retrieve metadata from a Plutus validator
 
@@ -350,9 +345,9 @@ Example:\
 This is either:
 
 1. A low-level direct representation of the metadata, following closely the structure of the Cardano foundation off-chain metadata registry
-2. A 721-ERC-style token mapping exactly matching the structure of CIP-0025
+2. A nested map structure following CIP-0025's organization with policy_id and asset_name as keys, where the outer map contains a "721" key whose value maps policy_id to asset_name to metadata
 
-All UTF-8 encoded keys and values need to be converted into their respective byte's representation when creating the datum on-chain.
+All UTF-8 encoded keys and values need to be converted into their respective byte's representation when creating the datum on-chain. In the nested map format (option 2), the "721" key, policy_id keys, and asset_name keys must be encoded as bounded_bytes (following CIP-0025 version 2 conventions).
 
 ```
 ; Explanation here: https://developers.cardano.org/docs/native-tokens/token-registry/cardano-token-registry/
@@ -374,23 +369,18 @@ metadata =
 
 metadata_map = 
   {
-    __RESERVE_KEYWORD_721_V4__ : bool, ; Long name to avoid conflict
-
-    "721":
+    "721" : bounded_bytes => ; The "721" key encoded as UTF-8 bytes (0x373231)
     {
-      ; As many of the below as needed based on which reference tokens are added
-      "<policy_id>":
+      * policy_id : bounded_bytes => ; Policy ID encoded as bytes
       {
-        "<asset_name_1>" : metadata,
-        ; ...
-        "<asset_name_n>" : metadata,
+        * asset_name : bounded_bytes => metadata ; Asset name encoded as bytes (without label prefix)
       }
     }
   }
 
 metadata_field = 
-  metadata          ; For backwards-compatibility in v4 onward, a map itself is valid metadata
-  / metadata_map    ; Multi-metadata is now supported as in CIP-0025
+  metadata          ; Direct metadata format (versions 1-3)
+  / metadata_map    ; Nested map format with "721" key (version 4)
 
 ; A valid Uniform Resource Identifier (URI) as a UTF-8 encoded bytestring.
 ; The URI scheme must be one of `https` (HTTP), `ipfs` (IPFS), `ar` (Arweave) or `data` (on-chain).
@@ -448,9 +438,9 @@ to lookup. The steps are
 1. Construct `reference NFT` from `user token`: `d5e6bf0500378d4f0da4e8dde6becec7621cd8cbf5cbb9b87013d4cc.(100)TestToken`
 2. Look up `reference NFT` and find the output it's locked in.
 3. Get the datum from the output and lookup metadata by going into the first field of constructor 0.
-4. Determine whether this is a direct metadata or the CIP-0025 map style using the `__RESERVE_KEYWORD_721_V4__` map value (if present, default false)
+4. Determine whether this is direct metadata (map without "721" key) or nested map format (map with "721" key).
 5. Convert to JSON and encode all string entries to UTF-8 if possible, otherwise leave them in hex.
-6. For multi-map style metadata, return only the JSON values inside the matching path: `721`->>`policy_id`->>`asset_name`
+6. For nested map format metadata, return only the metadata inside the matching path: map["721"][policy_id][asset_name], where policy_id and asset_name are matched without the asset_name_label prefix.
 
 ##### Retrieve metadata from a Plutus validator
 
@@ -495,10 +485,10 @@ Example:\
 
 This is either:
 
-1. A low-level direct representation of the metadata, following closely the structure of CIP-0025 with the optional decimals field added
-2. A 721-ERC-style token mapping exactly matching the structure of CIP-0025
+1. A low-level direct representation of the metadata, following closely the structure of CIP-0025 for individual asset metadata (name, image, files, etc.) with the optional decimals field added
+2. A nested map structure following CIP-0025's organization with policy_id and asset_name as keys, where the outer map contains a "721" key whose value maps policy_id to asset_name to metadata
 
-All UTF-8 encoded keys and values need to be converted into their respective byte's representation when creating the datum on-chain.
+All UTF-8 encoded keys and values need to be converted into their respective byte's representation when creating the datum on-chain. In the nested map format (option 2), the "721" key, policy_id keys, and asset_name keys must be encoded as bounded_bytes (following CIP-0025 version 2 conventions).
 
 ```
 files_details =
@@ -525,23 +515,18 @@ metadata =
 
 metadata_map = 
   {
-    __RESERVE_KEYWORD_721_V4__ : bool, ; Long name to avoid conflict
-
-    "721":
+    "721" : bounded_bytes => ; The "721" key encoded as UTF-8 bytes (0x373231)
     {
-      ; As many of the below as needed based on which reference tokens are added
-      "<policy_id>":
+      * policy_id : bounded_bytes => ; Policy ID encoded as bytes
       {
-        "<asset_name_1>" : metadata,
-        ; ...
-        "<asset_name_n>" : metadata,
+        * asset_name : bounded_bytes => metadata ; Asset name encoded as bytes (without label prefix)
       }
     }
   }
 
 metadata_field = 
-  metadata          ; For backwards-compatibility in v4 onward, a map itself is valid metadata
-  / metadata_map    ; Multi-metadata is now supported as in CIP-0025
+  metadata          ; Direct metadata format (versions 1-3)
+  / metadata_map    ; Nested map format with "721" key (version 4)
 
 ; A valid Uniform Resource Identifier (URI) as a UTF-8 encoded bytestring.
 ; The URI scheme must be one of `https` (HTTP), `ipfs` (IPFS), `ar` (Arweave) or `data` (on-chain).
@@ -615,9 +600,9 @@ to lookup. The steps are
 1. Construct `reference NFT` from `user token`: `d5e6bf0500378d4f0da4e8dde6becec7621cd8cbf5cbb9b87013d4cc.(100)TestToken`
 2. Look up `reference NFT` and find the output it's locked in.
 3. Get the datum from the output and lookup metadata by going into the first field of constructor 0.
-4. Determine whether this is a direct metadata or the CIP-0025 map style using the `__RESERVE_KEYWORD_721_V4__` map value (if present, default false)
+4. Determine whether this is direct metadata (map without "721" key) or nested map format (map with "721" key).
 5. Convert to JSON and encode all string entries to UTF-8 if possible, otherwise leave them in hex.
-6. For multi-map style metadata, return only the JSON values inside the matching path: `721`->>`policy_id`->>`asset_name`
+6. For nested map format metadata, return only the metadata inside the matching path: map["721"][policy_id][asset_name], where policy_id and asset_name are matched without the asset_name_label prefix.
 
 ##### Retrieve metadata from a Plutus validator
 
@@ -696,7 +681,7 @@ versions of the affected tokens. `asset_name_labels` **MUST** only be marked obs
 
 #### version 4
 
-- Add backwards-compatible support for multi-asset 721-ERC-style metadata mappings
+- Add support for nested map format following CIP-0025's policy_id -> asset_name structure, allowing multiple assets' metadata in a single datum. This format is backwards-compatible with versions 1-3, which only support direct metadata format.
 
 ## Rationale: how does this CIP achieve its goals?
 
--- a/CIP-0068/README.md
+++ b/CIP-0068/README.md
@@ -155,7 +155,7 @@ The `user token` represents an NFT (non-fungible token).
 
 ##### Pattern
 
-The `user token` and `reference NFT` MUST have an identical name, preceded by the `asset_name_label` prefix.
+The `user token` and `reference NFT` MUST have an identical name (heretofore entitled `asset_name`), preceded by the `asset_name_label` prefix.
 
 Example:\
 `user token`: `(222)Test123`\
@@ -163,8 +163,12 @@ Example:\
 
 ##### Metadata
 
-This is a low-level representation of the metadata, following closely the structure of CIP-0025. All UTF-8 encoded keys
-and values need to be converted into their respective byte's representation when creating the datum on-chain.
+This is either:
+
+1. A low-level direct representation of the metadata, following closely the structure of CIP-0025
+2. A 721-ERC-style token mapping exactly matching the structure of CIP-0025
+
+All UTF-8 encoded keys and values need to be converted into their respective byte's representation when creating the datum on-chain.
 
 ```
 files_details =
@@ -188,6 +192,26 @@ metadata =
     ; ... Additional properties are allowed
   }
 
+metadata_map = 
+  {
+    __RESERVE_KEYWORD_721_V4__ : bool, ; Long name to avoid conflict
+
+    "721":
+    {
+      ; As many of the below as needed based on which reference tokens are added
+      "<policy_id>":
+      {
+        "<asset_name_1>" : metadata,
+        ; ...
+        "<asset_name_n>" : metadata,
+      }
+    }
+  }
+
+metadata_field = 
+  metadata          ; For backwards-compatibility in v4 onward, a map itself is valid metadata
+  / metadata_map    ; Multi-metadata is now supported as in CIP-0025
+
 ; A valid Uniform Resource Identifier (URI) as a UTF-8 encoded bytestring.
 ; The URI scheme must be one of `https` (HTTP), `ipfs` (IPFS), `ar` (Arweave) or `data` (on-chain).
 ; Data URLs (on-chain data) must comply to RFC2397.
@@ -198,12 +222,12 @@ uri = bounded_bytes / [ * bounded_bytes ] ; UTF-8
 ; and needs to be at least Unit/Void: #6.121([])
 extra = plutus_data
 
-datum = #6.121([metadata, version, extra])
+datum = #6.121([metadata_field, version, extra])
 
-version = 1 / 2 / 3
+version = 1 / 2 / 3 / 4
 ```
 
-Example datum as JSON:
+Example datum as JSON (direct metadata):
 
 ```json
 {
@@ -236,6 +260,45 @@ Example datum as JSON:
 }
 ```
 
+Example datum as JSON (721-style):
+
+```json
+{
+  "constructor": 0,
+  "fields": [
+    {
+      "map": [
+        {
+          "k": { "bytes": "373231" },
+          "v": {
+            "map": [
+              {
+                "k": { "bytes": "5370616365427564204131" },
+                "v": {
+                  "map": [
+                    {
+                      "k": { "bytes": "6E616D65" },
+                      "v": { "bytes": "5370616365427564" }
+                    },
+                    {
+                      "k": { "bytes": "696D616765" },
+                      "v": { "bytes": "697066733A2F2F74657374" }
+                    }
+                  ]
+                }
+              }
+            ]
+          }
+        }
+      ]
+    },
+    {
+      "int": 4
+    }
+  ]
+}
+```
+
 ##### Retrieve metadata as 3rd party
 
 A third party has the following NFT `d5e6bf0500378d4f0da4e8dde6becec7621cd8cbf5cbb9b87013d4cc.(222)TestToken` they want
@@ -244,7 +307,9 @@ to lookup. The steps are
 1. Construct `reference NFT` from `user token`: `d5e6bf0500378d4f0da4e8dde6becec7621cd8cbf5cbb9b87013d4cc.(100)TestToken`
 2. Look up `reference NFT` and find the output it's locked in.
 3. Get the datum from the output and lookup metadata by going into the first field of constructor 0.
-4. Convert to JSON and encode all string entries to UTF-8 if possible, otherwise leave them in hex.
+4. Determine whether this is a direct metadata or the CIP-0025 map style using the `__RESERVE_KEYWORD_721_V4__` map value (if present, default false)
+5. Convert to JSON and encode all string entries to UTF-8 if possible, otherwise leave them in hex.
+6. For multi-map style metadata, return only the JSON values inside the matching path: `721`->>`policy_id`->>`asset_name`
 
 ##### Retrieve metadata from a Plutus validator
 
@@ -274,7 +339,7 @@ The `user token` is an FT (fungible token).
 
 ##### Pattern
 
-The `user token` and `reference NFT` MUST have an identical name, preceded by the `asset_name_label` prefix.
+The `user token` and `reference NFT` MUST have an identical name (heretofore called `asset_name`), preceded by the `asset_name_label` prefix.
 
 Example:\
 `user token`: `(333)Test123`\
@@ -282,9 +347,12 @@ Example:\
 
 ##### Metadata
 
-This is a low-level representation of the metadata, following closely the structure of the Cardano foundation off-chain
-metadata registry. All UTF-8 encoded keys and values need to be converted into their respective byte's representation
-when creating the datum on-chain.
+This is either:
+
+1. A low-level direct representation of the metadata, following closely the structure of the Cardano foundation off-chain metadata registry
+2. A 721-ERC-style token mapping exactly matching the structure of CIP-0025
+
+All UTF-8 encoded keys and values need to be converted into their respective byte's representation when creating the datum on-chain.
 
 ```
 ; Explanation here: https://developers.cardano.org/docs/native-tokens/token-registry/cardano-token-registry/
@@ -304,6 +372,26 @@ metadata =
     ; ... Additional properties are allowed
   }
 
+metadata_map = 
+  {
+    __RESERVE_KEYWORD_721_V4__ : bool, ; Long name to avoid conflict
+
+    "721":
+    {
+      ; As many of the below as needed based on which reference tokens are added
+      "<policy_id>":
+      {
+        "<asset_name_1>" : metadata,
+        ; ...
+        "<asset_name_n>" : metadata,
+      }
+    }
+  }
+
+metadata_field = 
+  metadata          ; For backwards-compatibility in v4 onward, a map itself is valid metadata
+  / metadata_map    ; Multi-metadata is now supported as in CIP-0025
+
 ; A valid Uniform Resource Identifier (URI) as a UTF-8 encoded bytestring.
 ; The URI scheme must be one of `https` (HTTP), `ipfs` (IPFS), `ar` (Arweave) or `data` (on-chain).
 ; Data URLs (on-chain data) must comply to RFC2397.
@@ -314,9 +402,9 @@ uri = bounded_bytes / [ * bounded_bytes ] ; UTF-8
 ; and needs to be at least Unit/Void: #6.121([])
 extra = plutus_data
 
-datum = #6.121([metadata, version, extra])
+datum = #6.121([metadata_field, version, extra])
 
-version = 1 / 2 / 3
+version = 1 / 2 / 3 / 4
 ```
 
 Example datum as JSON:
@@ -360,7 +448,9 @@ to lookup. The steps are
 1. Construct `reference NFT` from `user token`: `d5e6bf0500378d4f0da4e8dde6becec7621cd8cbf5cbb9b87013d4cc.(100)TestToken`
 2. Look up `reference NFT` and find the output it's locked in.
 3. Get the datum from the output and lookup metadata by going into the first field of constructor 0.
-4. Convert to JSON and encode all string entries to UTF-8 if possible, otherwise leave them in hex.
+4. Determine whether this is a direct metadata or the CIP-0025 map style using the `__RESERVE_KEYWORD_721_V4__` map value (if present, default false)
+5. Convert to JSON and encode all string entries to UTF-8 if possible, otherwise leave them in hex.
+6. For multi-map style metadata, return only the JSON values inside the matching path: `721`->>`policy_id`->>`asset_name`
 
 ##### Retrieve metadata from a Plutus validator
 
@@ -403,9 +493,12 @@ Example:\
 
 ##### Metadata
 
-This is a low-level representation of the metadata, following closely the structure of CIP-0025 with the optional
-decimals field added. All UTF-8 encoded keys and values need to be converted into their respective byte's representation
-when creating the datum on-chain.
+This is either:
+
+1. A low-level direct representation of the metadata, following closely the structure of CIP-0025 with the optional decimals field added
+2. A 721-ERC-style token mapping exactly matching the structure of CIP-0025
+
+All UTF-8 encoded keys and values need to be converted into their respective byte's representation when creating the datum on-chain.
 
 ```
 files_details =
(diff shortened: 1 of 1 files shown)


rphair:
@thaddeusdiamond I was answering your comments by subject in alphabetical (not chronological) order, so I posted this response without seeing you've already done the proposed update:
* https://github.com/cardano-foundation/CIPs/issues/1111#issuecomment-3495360543

I think these changes are consistent with your suggestion of a new best-practice approach, the editorial guidelines, and the requirements of CIP-0068... but we'll follow the related issue in case other feedback indicates that a different approach should be taken.  In any case we'll tag this `Triage` so this can all be introduced at the next CIP meeting together (to include any review or resolutions in the meantime): https://hackmd.io/@cip-editors/123

rphair:
@thaddeusdiamond I'm recording this meeting resolution to leave this submission `Unconfirmed` due to a conversation that you & @Crypto2099 reportedly had about some unresolved details of implementing such a change... I'm not the best person to document this & in fact didn't catch all of it at the meeting, so I'll leave it to you & Adam to record the details here and/or discuss them in the "container" issue https://github.com/cardano-foundation/CIPs/issues/1111.

thaddeusdiamond:
I think @Crypto2099 was onboard as is since it's backwards compatible but he can comment here.  Will close #1111

rphair:
@thaddeusdiamond we might not get @Crypto2099's participation on this issue... so if there are any review points you could record here about that conversation please do... otherwise we'll continue on with this without it.

If that's the case... @Ryun1 @perturbing if you can recall any other blocking issues to consider this a CIP candidate, perhaps you can post them here?  Otherwise we might mark this as `Confirmed` and assign a CIP number at a CIP meeting in the near future (or by editor consensus outside the meeting).  @thaddeusdiamond feel free to comment again if time goes by without you observing progress along either of these lines.

SmaugPool:
1. I'm not sure to understand the proposal. As the reference token asset_name must match the user token one modulo the asset_name_label, and both should have the same policy, is the plan to have several reference tokens in the same tx output but one datum?

2. In this case could you detail how much this saves if at all with examples? Because the additional reserved bool, 721 key, `policy_id` and `asset_name` take space that was not required before. So I'm not convinced by the gain compared to several tx outputs, each with its own datum. 

3. Using a different name label could be considered instead of adding a reserved bool and 721 key.

4. Additionally I think some details about the metadata update process should be added. Should the datum be split when a single token needs to be updated? 

5. Overall please detail the aim of this proposal. What issue does it solve? I'm not sure consistency has been an issue, and the new format is still not fully consistent with CIP25 anyway (bytes instead of hex+utf8, additional `extra` field, etc.). Moreover FT and RFT (333,444) have slightly different metadata format than NFT 222 and CIP25. Therefore if there are consistency issues, could you show real world examples and show how this would solve them?

6. Lastly a major benefit of CIP68 is that smart contracts (onchain validators) can read the metadata. Is this still possible with your proposal? Do scripts have access to all the data required to lookup the correct asset metadata in the datum? I think some plutus example would be needed to demonstrate it. Also the `extra` field, very often used by onchain scrips, is now the same for all reference tokens in this proposal as it is in the common part of the tx output, so how do this work?

thaddeusdiamond:
> 1. I'm not sure to understand the proposal. As the reference token asset_name must match the user token one modulo the asset_name_label, and both should have the same policy, is the plan to have several reference tokens in the same tx output but one datum?

Correct that's the plan.

> 2. In this case could you detail how much this saves if at all with examples? Because the additional reserved bool, 721 key, `policy_id` and `asset_name` take space that was not required before. So I'm not convinced by the gain compared to several tx outputs, each with its own datum.

I've calculated that several of my artists could get a roughly 75% reduction in minUTxO.  For a multi-thousand art piece collection this is several hundred to several thousand dollars.

> 3. Using a different name label could be considered instead of adding a reserved bool and 721 key.

The only problem with that is since the name is part of the reference token itself, you would have to "reprint" the reference tokens instead of making this at the datum level.  But I'm open to suggestions.

> 4. Additionally I think some details about the metadata update process should be added. Should the datum be split when a single token needs to be updated?

I would think this would be up to the dApp developer? So long as it conforms to the spec they can either "respend" the datum and rewrite every related token metadata, or they could split it overall.

> 5. Overall please detail the aim of this proposal. What issue does it solve? I'm not sure consistency has been an issue, and the new format is still not fully consistent with CIP25 anyway (bytes instead of hex+utf8, additional `extra` field, etc.). Moreover FT and RFT (333,444) have slightly different metadata format than NFT 222 and CIP25. Therefore if there are consistency issues, could you show real world examples and show how this would solve them?

Consistency issues? This aims to be backwards-compatible with the existing data formats so I'm not sure I'm following.  As for the rationale, I'll attach it to the bottom of this comment and you can let me know what/where you want it.

> 6. Lastly a major benefit of CIP68 is that smart contracts (onchain validators) can read the metadata. Is this still possible with your proposal? Do scripts have access to all the data required to lookup the correct asset metadata in the datum? I think some plutus example would be needed to demonstrate it. Also the `extra` field, very often used by onchain scrips, is now the same for all reference tokens in this proposal as it is in the common part of the tx output, so how do this work?

Yes there is no change to SC usage.  Since the datum is just "larger" the contract may be slightly more complex, but there is no structural change.  This is more on the dApp or SC developer to interpret the datum in a new way.  Not sure about `extra` field but existing dApps can continue as is IMO.  This is an optional optimization.

---

## **Rationale for Introducing Optional CIP-0025-Compatible Metadata Within CIP-0068**

The current design of CIP-0068 inadvertently creates a significant cost barrier for large NFT collections due to minimum-ADA requirements. As ADA approaches parity with USD, artists and platforms minting 5,000–10,000+ on-chain assets are increasingly impacted by the **minUTxO amplification effect**: the metadata-bearing reference scripts and datums mandated by CIP-0068 can cause **10–20% of project budgets** to be consumed purely by ADA locked in outputs, rather than by art production or ecosystem growth.

This cost profile wasn’t as visible when ADA was significantly cheaper, but it has now become an operational friction point across multiple creator workflows.

The proposal aims to **reduce the minUTxO footprint** for CIP-0068 assets by allowing an **optional alternate metadata shape** that mirrors CIP-0025 (721-style metadata) while preserving CIP-0068’s semantics. Concretely:

### **1. Dual-format Metadata Reduces Unnecessary ADA Locking**

Allowing a CIP-0025-shaped metadata payload means:

* The on-chain can consolidate across multiple reference tokens.
* The minUTxO requirement is correspondingly reduced.
* Large-scale mints return to economically sensible territory, especially for low-cost or promotional collections where the metadata weight provides no additional value to creators.

This addresses a real-world pain point reported by multiple production-level minting pipelines, not just isolated cases.

### **2. Improved Migration Path from CIP-0025 to CIP-0068**

A sizeable portion of the ecosystem still lives on CIP-0025, particularly older marketplaces and archived collections. When artists revisit or reissue older collections, the transition path to CIP-0068 is currently friction-heavy, requiring metadata restructuring and toolchain updates.

Supporting a CIP-0025-compatible metadata envelope inside CIP-0068:

* Enables forward migration without metadata rewrites.
* Preserves backward recognizability.
* Reduces engineering cost for wallets, indexers, and minting tools.

### **3. Strict Optionality Preserves Backward Compatibility**

This proposal does **not** modify or deprecate existing CIP-0068 patterns.
Instead, it introduces an explicitly flagged alternate metadata payload, using a reserved binary key or indicator to avoid collisions with existing fields.

This allows:

* Full backward compatibility.
* No disruptive changes for current indexers or marketplaces.
* Opt-in use where economic relief is necessary.

---

## **Why a CIP-0068 Version Bump May Be Preferable to a New CIP**

Although this change is cross-cutting, it:

* Modifies an existing standard rather than creating a new conceptual primitive.
* Maintains the same reference-NFT architecture, same minting semantics, same high-level goals.
* Represents an **evolution**, not a parallel track.

A versioned update (e.g., v5) avoids fracturing standards, while still being explicit enough for tooling authors and indexers to adopt cleanly.

SmaugPool:
Thanks. A lot of the confusion came from the fact that you previously didn't introduce the MinUTXO issue at all in your proposal. The introduction was:
> Add version 4 metadata support with 721-ERC-style mappings to NFT (222), FT (333) and RFT (444) standards, ensuring consistency across all asset classes.

Which is why I thought it was about consistency. So it looked like it was trying to solve some metadata issues, but it actually tries to solve some cost and locked ADA issues. This is therefore a work-around for the current value of MinUTXO, quite unrelated to metadata formats.

Overall I think comparing to CIP25 is very confusing, and even more "721-ERC". This has almost nothing to do with CIP25:
- it uses a user token and a reference token
- it uses name labels
- encoding is very different (bytes everywhere)
- it uses datum instead of metadata
- it has an extra field for scripts
- it supports 333 and 444 metadata format which is closer to cip26 token registry format than CIP25
- etc.

The only common thing with CIP25 that you introduce is the "policy_id -> asset_name" mapping. The 222/333/444 metadata is then not encoded like CIP25. So given all these differences, I don't think it particularly helps migrating to CIP68. This might instead introduce a lot of errors like in CIP25 v2, where users use strings instead of bytes or mix both versions. 

Therefore in my opinion this proposal should be introduced without talking about CIP25 at all, but instead just introduce the aim of supporting multiple assets per tx_output and datum.

I will add other note in changes comments.

CIP-0068/README.md
```diff
@@ -155,16 +155,20 @@ The `user token` represents an NFT (non-fungible token).
...
 `reference NFT`: `(100)Test123`
 
 ##### Metadata
 
-This is a low-level representation of the metadata, following closely the structure of CIP-0025. All UTF-8 encoded keys
-and values need to be converted into their respective byte's representation when creating the datum on-chain.
+This is either:
+
+1. A low-level direct representation of the metadata, following closely the structure of CIP-0025
+2. A 721-ERC-style token mapping exactly matching the structure of CIP-0025
```
SmaugPool: This is confusing. Why referencing 721-ERC-style here? What does this mean? This is an Ethereum standard and it is not referenced anywhere. The only link with CIP25 is the 721 metadata label.

Also "mapping exactly matching the structure of CIP-0025" is also confusing. I think you are only talking about the `policy_id -> asset_name` mapping? The rest is different (222/333/444, bytes instead of strings, etc.). Also only CIP25v2 use bytes as keys in the mapping.

Note that "following closely the structure of CIP-0025" was about the individual asset metadata (name, image, files, etc.). Whereas you seem to talk about its encapsulation in `policy_id -> asset_name` maps.

CIP-0068/README.md
```diff
@@ -188,6 +192,26 @@ metadata =
...
+      ; As many of the below as needed based on which reference tokens are added
+      "<policy_id>":
+      {
+        "<asset_name_1>" : metadata,
+        ; ...
+        "<asset_name_n>" : metadata,
+      }
+    }
+  }
+
```
SmaugPool, line 205: This is not some valid CDDL. It looks like you are confusing CDDL with JSON.
More importantly, this does not tell how the `policy_id` and `asset_name` keys are encoded (I guess byte encoding like CIP25v2 and all CIP68 fields?).

Overall we should avoid talking about JSON as metadata and datum are CBOR encoded and CBOR -> JSON is not directly bijective. This only leads to confusion. 

Note that JSON in CIP68 is only discussed in examples and even this is slightly questionable in my opinion.

CIP-0068/README.md
```diff
@@ -236,6 +260,45 @@ Example datum as JSON:
 }
 ```
 
+Example datum as JSON (721-style):
```
SmaugPool: Again what "721-style" means?

CIP-0068/README.md
```diff
@@ -188,6 +192,26 @@ metadata =
     ; ... Additional properties are allowed
   }
 
+metadata_map = 
+  {
+    __RESERVE_KEYWORD_721_V4__ : bool, ; Long name to avoid conflict
+
+    "721":
```
SmaugPool: What is the point of having `__RESERVE_KEYWORD_721_V4__` bool here? Wouldn't the presence of "721" key be enough?

CIP-0068/README.md
```diff
@@ -304,6 +372,26 @@ metadata =
...
+      ; As many of the below as needed based on which reference tokens are added
+      "<policy_id>":
+      {
+        "<asset_name_1>" : metadata,
+        ; ...
+        "<asset_name_n>" : metadata,
+      }
+    }
+  }
+
```
SmaugPool, line 380: Again not valid CDDL

CIP-0068/README.md
```diff
@@ -430,19 +523,39 @@ metadata =
...
```
SmaugPool, line 526: Again not valid CDDL

CIP-0068/README.md
```diff
@@ -578,6 +694,10 @@ versions of the affected tokens. `asset_name_labels` **MUST** only be marked obs
 
 - Added [* bounded_bytes] support to the image and src tags on the metadata 
 
+#### version 4
+
+- Add backwards-compatible support for multi-asset 721-ERC-style metadata mappings
```
SmaugPool: "Backwards-compatible" with what exactly? 
And again "721-ERC-style" is not clear. There is no ERC on Cardano.

SmaugPool:
> The only problem with that is since the name is part of the reference token itself, you would have to "reprint" the reference tokens instead of making this at the datum level. But I'm open to suggestions.

I meant introducing a new label instead of reusing 222/333/444. Without doing this, this not fully backward compatible because a current indexer will not expect a version 4. So some new CIP68 token minted as a single asset but with version 4 will be ignored or rejected by such tools.

I don't understand what you mean by "you would have to "reprint" the reference tokens".

SmaugPool:
> I would think this would be up to the dApp developer? So long as it conforms to the spec they can either "respend" the datum and rewrite every related token metadata, or they could split it overall.

I'm not sure there is any choice: assuming there are n tokens in a single tx output with their metadata in the output datum. To update only one of the tokens metadata, you will have to spend all of them, at least as change, but CIP68 metadata lookup is described as:
> Construct reference NFT from user token
Look up reference NFT and find the output it's locked in.
Get the datum from the output and lookup metadata by going into the first field of constructor 0.

Note that the lookup only looks at the unspent output (not a previous output). Therefore you will have to rewrite the metadata of all tokens in a datum, hence it makes no sense to split them, it would be even costlier.

thaddeusdiamond:
> > I would think this would be up to the dApp developer? So long as it conforms to the spec they can either "respend" the datum and rewrite every related token metadata, or they could split it overall.
> 
> I'm not sure there is any choice: assuming there are n tokens in a single tx output with their metadata in the output datum. To update only one of the tokens metadata, you will have to spend all of them, at least as change, but CIP68 metadata lookup is described as:
> 
> > Construct reference NFT from user token
> > Look up reference NFT and find the output it's locked in.
> > Get the datum from the output and lookup metadata by going into the first field of constructor 0.
> 
> Note that the lookup only looks at the unspent output (not a previous output). Therefore you will have to rewrite the metadata of all tokens in a datum, hence it makes no sense to split them, it would be even costlier.

Makes sense.  So rewrite it then makes sense to me.

SmaugPool:
> > > I would think this would be up to the dApp developer? So long as it conforms to the spec they can either "respend" the datum and rewrite every related token metadata, or they could split it overall.
> > 
> > 
> > I'm not sure there is any choice: assuming there are n tokens in a single tx output with their metadata in the output datum. To update only one of the tokens metadata, you will have to spend all of them, at least as change, but CIP68 metadata lookup is described as:
> > > Construct reference NFT from user token
> > > Look up reference NFT and find the output it's locked in.
> > > Get the datum from the output and lookup metadata by going into the first field of constructor 0.
> > 
> > 
> > Note that the lookup only looks at the unspent output (not a previous output). Therefore you will have to rewrite the metadata of all tokens in a datum, hence it makes no sense to split them, it would be even costlier.
> 
> Makes sense. So rewrite it then makes sense to me.

This is quite a significant drawback compared to current CIP68 and even CIP25 (as only one of the tokens can be minted again to update its metadata), so maybe it's worth thinking more about it to avoid introducing a new flaw?

thaddeusdiamond:
> > > > I would think this would be up to the dApp developer? So long as it conforms to the spec they can either "respend" the datum and rewrite every related token metadata, or they could split it overall.
> > > 
> > > 
> > > I'm not sure there is any choice: assuming there are n tokens in a single tx output with their metadata in the output datum. To update only one of the tokens metadata, you will have to spend all of them, at least as change, but CIP68 metadata lookup is described as:
> > > > Construct reference NFT from user token
> > > > Look up reference NFT and find the output it's locked in.
> > > > Get the datum from the output and lookup metadata by going into the first field of constructor 0.
> > > 
> > > 
> > > Note that the lookup only looks at the unspent output (not a previous output). Therefore you will have to rewrite the metadata of all tokens in a datum, hence it makes no sense to split them, it would be even costlier.
> > 
> > 
> > Makes sense. So rewrite it then makes sense to me.
> 
> This is quite a significant drawback compared to current CIP68 and even CIP25 (as only one of the tokens can be minted again to update its metadata), so maybe it's worth thinking more about it to avoid introducing a new flaw?

If it's an optional capability what's the drawback? I don't know how to solve the minUTxO problem without consolidating the NFT metadata.

To give a specific example: one of the artists I am minting for has taken roughly 10,000 ADA in project funds and the locked minUTxO is over 1,000 ADA.  So that's 10% of his total funds that are locked and can't contribute to his product development.

SmaugPool:
> Makes sense. So rewrite it then makes sense to me.

Actually, the datum does not have to be inline, in which case I think it can be later referenced just by its hash. I guess this would solve this issue assuming your standard recommends using non-inline datum, at least when the policy allows updates.

CIP-0068/README.md
thaddeusdiamond: You're absolutely right. I've removed the "721-ERC-style" terminology and clarified that:
- Option 1 refers to the direct metadata format with individual asset metadata fields (name, image, files, etc.) following CIP-0025's structure
- Option 2 refers to CIP-0025's nested map organization with policy_id -> asset_name keys, where the outer map contains a "721" key

The clarification now distinguishes between the individual metadata structure (which CIP-0025 defines) and the encapsulation structure (the policy_id -> asset_name mapping that CIP-0025 uses).

CIP-0068/README.md
thaddeusdiamond: Fixed. The CDDL now properly uses bounded_bytes for all keys:
- "721" key is encoded as bounded_bytes (UTF-8 bytes: 0x373231)
- policy_id keys are bounded_bytes (as in CIP-0025 version 2)
- asset_name keys are bounded_bytes (as in CIP-0025 version 2)

This matches CIP-0025 version 2 conventions where keys are byte-encoded rather than text strings.

CIP-0068/README.md
thaddeusdiamond: Changed to "nested map format" to be consistent with the terminology used throughout the document and avoid the ambiguous "721-style" reference.

CIP-0068/README.md
thaddeusdiamond: Removed the `__RESERVE_KEYWORD_721_V4__` marker. You're correct that the presence of the "721" key is sufficient to detect the nested map format. The format can be determined simply by checking if the metadata map contains a "721" key.

CIP-0068/README.md
thaddeusdiamond: All metadata_map definitions across 222, 333, and 444 standards have been updated with proper CDDL syntax using bounded_bytes for all keys, consistent with CIP-0025 version 2.

CIP-0068/README.md
thaddeusdiamond: All metadata_map definitions across 222, 333, and 444 standards have been updated with proper CDDL syntax using bounded_bytes for all keys, consistent with CIP-0025 version 2.

CIP-0068/README.md
thaddeusdiamond: Clarified:
- "Backwards-compatible" means compatible with versions 1-3, which only supported direct metadata format. Version 4 adds the nested map format as an additional option.
- Removed "721-ERC-style" terminology entirely, replacing it with "nested map format following CIP-0025's policy_id -> asset_name structure"

thaddeusdiamond:
I published a new change, rationale, and some of my responses.  By "reprint" I mean that I already have a ton of (222) NFTs that I want to optimize without having to airdrop everyone a token.  Since this is optional and backward-compatible, I can do this optimization inline for my artists now without requiring the creation of 555/666 or whatnot.

Please let me know the path to approval @rphair @SmaugPool

rphair (commented):
@thaddeusdiamond at the CIP meeting earlier today we spontaneously picked this one off the `Unconfirmed` list  and decided that we'd been looking for the response that you made a few hours later. 😅

In my opinion this satisfies all the demanding review points beginning with https://github.com/cardano-foundation/CIPs/pull/1112#issuecomment-3643025222 with corresponding changes in the text.

Personally I appreciate that https://github.com/cardano-foundation/CIPs/pull/1112#issuecomment-3642619340 was put literally into the _Rationale_ & I'd also like to see that happen for any remaining clarifications resulting from the debate — past, present or future — for the benefit of new implementors and others who may be interested in migrating their tokens.

@SmaugPool can you please go back & double-check the conversations beginning with your https://github.com/cardano-foundation/CIPs/pull/1112#discussion_r2611487924 and mark "resolved" the ones you believe have been properly addressed... or note otherwise if you think they need more debate or adjustment?

Otherwise I think we have a achieved a CIP candidate and so have put this on _Review_ for the next CIP meeting (https://hackmd.io/@cip-editors/126) & we'll double-check all pending review points again in 2 weeks.

@thaddeusdiamond @SmaugPool please also tag other associates who you think would be interested in confirming, disputing, or reviewing this update. 🙏

rphair:
@thaddeusdiamond due to a lot of proposals on the `Triage` list we had to postpone editors' & attendees' review until the next meeting https://hackmd.io/@cip-editors/127 - please (cc @SmaugPool) continue the review & response here online in the meantime if possible.

thaddeusdiamond:
Seems this review may have been deleted? But no, this is intended to be
fully backwards-compatible.

rphair (commented):
@thaddeusdiamond my guess is that @Ryun1 deleted that review (posted a few minutes before today's CIP meeting) after answering his own question about the backward compatibility (this had already been briefly settled at an earlier meeting).

We pressed this item through our Review agenda at that meeting and those present believed it to be technically in order, but with some concern coming from "whether wallet devs have asked for this".  The editors believe this could be an issue because of pathological cases like (hypothetically):
1. without interest, let alone buy-in, to this addition: this "next generation" of the CIP-0068 standard is merged in, but wallet implementations don't (or don't consistently) use it;
2. then some other conception of that "next generation" comes along and this update needs to be backed out somehow — or worse: somehow merged with another evolution that it might not be easily compatible with.

So our decision was to progress this if & when we can get some endorsement from tagged representatives from wallet development teams.  @Ryun1 and @perturbing said they would work on this and we're awaiting their tagging some of these for review.

In the meantime I think we can call this `Confirmed` (though not `Last Check` as it would be for an unquestioned update PR) as we would any technically complete update awaiting more detailed review.

thaddeusdiamond:
Sounds good.

Ryun1:
@thaddeusdiamond appreciate the efforts 

It would be beneficial to see further review from a range of implementors
to confirm the motivation for this new version and that it is reasonable to be adopted

Tagging some known CIP-68 implementors for visibility
@AlexDochioiu @Scitz0 @refi93 @ashisherc

AlexDochioiu:
I had a quick look and I do not clearly understand how this is intended to work.

Today, with CIP-68, we have:

```
{policy}.({type}){name}
```

To retrieve metadata, we replace the type with the reference type and then read the datum from the UTxO that contains the reference asset.

Each token has a clearly identifiable reference pattern:
```
{policy}.(222){name} -> {policy}.(100){name}
```

So we know exactly which reference token to look for in order to extract metadata.

What I do not understand is how multiple tokens such as:

```
{policy}.(222){name1}
{policy}.(222){name2}
{policy}.(222){name3}
```

are mapped to a single reference.

How do I determine which reference token to query?

Will there be a single reference shared across multiple tokens?  
Or is the idea that multiple reference tokens are placed in the same UTxO and share a single datum?

I also noticed mention of including the policy as a key in the datum. Is this because we expect multiple reference tokens from different policies to coexist within the same UTxO?

---

If this proposal assumes multiple reference tokens sharing a UTxO:

- The value for wallets is close to zero. It does not meaningfully improve speed or efficiency, since fetching UTxOs by transaction hash is already fast with most indexers.
- The impact on wallets is manageable. The metadata extraction logic becomes slightly more complex, but not significantly so.

One additional point:

**I am generally not a fan of CIP-68. Across nearly all Cardano indexers we use, both managed and self-hosted, retrieving the latest transaction for a given asset, which is required to locate the UTxO containing the datum, is consistently problematic. It is a slow operation and frequently times out or fails altogether.**

thaddeusdiamond:
Let me try to answer inline below:

> I had a quick look and I do not clearly understand how this is intended to work.
> 
> Today, with CIP-68, we have:
> 
> ```
> {policy}.({type}){name}
> ```
> 
> To retrieve metadata, we replace the type with the reference type and then read the datum from the UTxO that contains the reference asset.
> 
> Each token has a clearly identifiable reference pattern:
> 
> ```
> {policy}.(222){name} -> {policy}.(100){name}
> ```
> 
> So we know exactly which reference token to look for in order to extract metadata.
> 
> What I do not understand is how multiple tokens such as:
> 
> ```
> {policy}.(222){name1}
> {policy}.(222){name2}
> {policy}.(222){name3}
> ```
> 
> are mapped to a single reference.
> 
> How do I determine which reference token to query?

The name determines how you query.  So there are no changes to the token type (that lives in the actual asset name itself as a prefix).  All you are looking up is the metadata of {asset_id}{name} => CDDL

> 
> Will there be a single reference shared across multiple tokens? Or is the idea that multiple reference tokens are placed in the same UTxO and share a single datum?

The idea is to share a single datum for space and cost efficiency as all.

> 
> I also noticed mention of including the policy as a key in the datum. Is this because we expect multiple reference tokens from different policies to coexist within the same UTxO?

It allows for that (similar to how we did it in CIP-0025)

> 
> If this proposal assumes multiple reference tokens sharing a UTxO:
> 
> * The value for wallets is close to zero. It does not meaningfully improve speed or efficiency, since fetching UTxOs by transaction hash is already fast with most indexers.

Agree with this.

> * The impact on wallets is manageable. The metadata extraction logic becomes slightly more complex, but not significantly so.

Good to hear.


> 
> One additional point:
> 
> **I am generally not a fan of CIP-68. Across nearly all Cardano indexers we use, both managed and self-hosted, retrieving the latest transaction for a given asset, which is required to locate the UTxO containing the datum, is consistently problematic. It is a slow operation and frequently times out or fails altogether.**

Yeah CIP-0068 is controversial, but the rationale here is to save artists developing on the chain from ballooning minUTxO (which often reaches into the many thousands of ADA for mints).

AlexDochioiu:
@thaddeusdiamond Thank you for the reply! I feel like your message already mostly implies it, but I'll still ask the following question directly to be 100% I'm not mis-interpreting the purpose of this CIP.

**Today:** Every reference token sits in a dedicated UTxO and the UTxO datum only refers to this (single) token.

**The CIP proposal:** Multiple reference tokens (potentially from multiple policies) can coexist in the same UTxO and share a single datum (containing the metadata for all of them).

Am I correct in understanding correctly that this is the core (and only?) proposal made by this CIP?

thaddeusdiamond:
Yes exactly!

rphair:
@thaddeusdiamond really nice to see this proposal moving along into peer review. ⭐ If you're going to keep responding by email to GitHub, please pay attention to how you enter your text (separating by a newline is essential; maybe even a blank line) so the entire quoted email doesn't appear here in the GitHub review stream. 🙏 (otherwise editors have to fit it: e.g. https://github.com/cardano-foundation/CIPs/pull/1112#issuecomment-3920809772)

thaddeusdiamond:
> @thaddeusdiamond really nice to see this proposal moving along into peer review. ⭐ If you're going to keep responding by email to GitHub, please pay attention to how you enter your text (separating by a newline is essential; maybe even a blank line) so the entire quoted email doesn't appear here in the GitHub review stream. 🙏 (otherwise editors have to fit it: e.g. [#1112 (comment)](https://github.com/cardano-foundation/CIPs/pull/1112#issuecomment-3920809772))

haha TIL... Fixed.  Sorry about that.

Scitz0:
Hi, from Eternl's perspective, I don't see any direct issue with the proposed CIP but I haven't spent much time digging through it either. This will likely first happen at the implementation stage if there are any issues. If I read change and comments correctly, fetching the datum works just like before, its just that the content of the datum can now either be as before with direct metadata, or using the 721 style with policy and name in the data structure?

thaddeusdiamond:
> Hi, from Eternl's perspective, I don't see any direct issue with the proposed CIP but I haven't spent much time digging through it either. This will likely first happen at the implementation stage if there are any issues. If I read change and comments correctly, fetching the datum works just like before, its just that the content of the datum can now either be as before with direct metadata, or using the 721 style with policy and name in the data structure?

That is correct yes.

Scitz0:
Ok, then it should be fairly easy to implement I think.

thaddeusdiamond:
@Scitz0 @rphair any updates on this CIP? Thanks.

Scitz0:
> @Scitz0 @rphair any updates on this CIP? Thanks.

I haven't spent any time inplementing the change in Eternl but if its accepted and demand is there we will put it on the list of requested features.

CIP-0068/README.md
rphair: @thaddeusdiamond I'm resolving this as per your statement that you've 'removed the "721-ERC-style" terminology' which certainly seems to satisfy the review point.

CIP-0068/README.md
rphair: resolving for reasons noted & apparently fixed on one of the later commits.   @thaddeusdiamond please get in the habit of marking these "resolved" when you believe them to be fixed (reviewers e.g. @SmaugPool can/should also do this).

CIP-0068/README.md
rphair: (also apparently fixed on one of the latter commits)

CIP-0068/README.md
rphair: (continued: also resolving)

CIP-0068/README.md
rphair: Thanks @SmaugPool & @thaddeusdiamond for working out these details & making a stronger spec & better written proposal: more uniquely qualifying Cardano for this type of application.  Based on this rigorous review I feel confident moving to a last round of review before merge.

rphair (approved):
@thaddeusdiamond @Scitz0 (cc @SmaugPool @AlexDochioiu) I also agree that merging this will help confirm the demand for this method.  Please do whatever you can before & after an anticipated merge of this update to ensure developers know about the new specification.
* @Ryun1 @Crypto2099 please also post ASAP if you can think of any recommendations that should be made about compatibility issues (not likely I think, given the statement of "Full backward compatibility") or other developer announcements (and circulate these, since this mightn't be obvious from this PR's merge itself).

I think the last review about the details of nomenclature and CDDL have satisfied all significant community challenges so far & so I feel comfortable endorsing this and moving it on to `Last Check` for the upcoming meeting (https://hackmd.io/@cip-editors/134).
* Note we have many last-checks on this agenda due to a harmonic convergence of mature CIPs & CPSs... so although I'm happy to invite the author & all advocates of this proposal & we'll do our best to merge it on time, this can't be assured especially if other CIP editors still need to review it... but I'll ping @Ryun1 @perturbing to see if we can clear that requirement ahead of the meeting (with no rush of course).

thaddeusdiamond:
> @thaddeusdiamond @Scitz0 (cc @SmaugPool @AlexDochioiu) I also agree that merging this will help confirm the demand for this method. Please do whatever you can before & after an anticipated merge of this update to ensure developers know about the new specification.
> 
> * @Ryun1 @Crypto2099 please also post ASAP if you can think of any recommendations that should be made about compatibility issues (not likely I think, given the statement of "Full backward compatibility") or other developer announcements (and circulate these, since this mightn't be obvious from this PR's merge itself).
> 
> I think the last review about the details of nomenclature and CDDL have satisfied all significant community challenges so far & so I feel comfortable endorsing this and moving it on to `Last Check` for the upcoming meeting (https://hackmd.io/@cip-editors/134).
> 
> * Note we have many last-checks on this agenda due to a harmonic convergence of mature CIPs & CPSs... so although I'm happy to invite the author & all advocates of this proposal & we'll do our best to merge it on time, this can't be assured especially if other CIP editors still need to review it... but I'll ping @Ryun1 @perturbing to see if we can clear that requirement ahead of the meeting (with no rush of course).

Great! I will work on the implementation for my minting tools and look forward to seeing it live in Eternl.  In the meantime, since JPG.store is shutting down the only real implementers would be @SmaugPool, WayUp (I can reach out to Cash once it's merged) and the wallets.

thaddeusdiamond:
Sample consolidation transaction:
https://cardanoscan.io/transaction/6c900994bb705921cbcbdd40a404830b60997ea8015cca91bb4875d591a4f05c?tab=utxo

Sample consolidation UTxO datum:
```
d8799fa143373231a1581c8dcb1f87d74f3c973e6373aaa24f17ba9369e4af71bdc1973b37b732ae5157696c6454616e677a2053322036383032b0496d656469615479706549696d6167652f706e674770726f6a6563745357696c642054616e677a205365726965732032436675724843686172636f616c44626f64794744656661756c7444657965734653746f6e6564446e616d65581c57696c642054616e677a202d205365726965732032202d202339353145696d6167659f47697066733a2f2f583b6261667962656966737272376a706f3570333667356e3268676b66353273656472346f6f32743269736a7334793463347861646962373678767875ff456d6f7574684554656574684765796577656172474d6f6e6f636c6548636c6f7468696e674a562d4e65636b2054656548686561647765617257507572706c6520616e6420476f6c642042616c6c6361704a6261636b67726f756e644d466f756e7461696e20426c75654b6163636573736f726965734d47616d65646179205061696e744c757067726164656446726f6d4d57696c6454616e677a203935314e6f726967696e616c506f6c69637958383333353638616431316639336233653739616538646565356164393238646564373261646365613731396539323130386361663135323162466172746973744b406e6574616e656c63686e5157696c6454616e677a2053322035303835b0496d656469615479706549696d6167652f706e674770726f6a6563745357696c642054616e677a2053657269657320324366757244426c756544626f64794744656661756c7444657965734744656661756c74446e616d65581d57696c642054616e677a202d205365726965732032202d20233431323745696d6167659f47697066733a2f2f583b6261667962656968777362776c33337433793479687272686670706a367a7734366b697a61676269686e746b6163686b34326266633732786e7479ff456d6f75746845546565746847657965776561724841766961746f727348636c6f7468696e6752547572746c656e65636b2053776561746572486865616477656172464265616e69654a6261636b67726f756e644d41766f6361646f20477265656e4b6163636573736f7269657352536b656c65746f6e20466163657061696e744c757067726164656446726f6d4e57696c6454616e677a20343132374e6f726967696e616c506f6c69637958383333353638616431316639336233653739616538646565356164393238646564373261646365613731396539323130386361663135323162466172746973744b406e6574616e656c63686e5157696c6454616e677a2053322035303533b0496d656469615479706549696d6167652f706e674770726f6a6563745357696c642054616e677a205365726965732032436675724843686172636f616c44626f64794744656661756c74446579657349426c61636b20457965446e616d65581d57696c642054616e677a202d205365726965732032202d20233338333645696d6167659f47697066733a2f2f583b626166796265696364636e696f3377777477793372326c33657a787962353469726f63667171706b6a3635726d6b6d35746872766a787368773661ff456d6f75746845426c756e744765796577656172474d6f6e6f636c6548636c6f7468696e6752547572746c656e65636b20537765617465724868656164776561724c4d4154412042616c6c6361704a6261636b67726f756e644a4f72616e6779205265644b6163636573736f72696573444e6f6e654c757067726164656446726f6d4e57696c6454616e677a20333833364e6f726967696e616c506f6c69637958383333353638616431316639336233653739616538646565356164393238646564373261646365613731396539323130386361663135323162466172746973744b406e6574616e656c63686e5157696c6454616e677a2053322035303635b0496d656469615479706549696d6167652f706e674770726f6a6563745357696c642054616e677a205365726965732032436675724542726f776e44626f64794744656661756c74446579657345416e677279446e616d65581d57696c642054616e677a202d205365726965732032202d20233131383645696d6167659f47697066733a2f2f583b6261667962656963326336346d3672777868633475357a7235783635353371786435656976757469647170646175787462636570616f6665706761ff456d6f7574684b467275697420536e61636b47657965776561724b536b6920476f67676c657348636c6f7468696e674854616e6720546f704868656164776561724742616e64616e614a6261636b67726f756e644d41766f6361646f20477265656e4b6163636573736f72696573444e6f6e654c757067726164656446726f6d4e57696c6454616e677a20313138364e6f726967696e616c506f6c69637958383333353638616431316639336233653739616538646565356164393238646564373261646365613731396539323130386361663135323162466172746973744b406e6574616e656c63686e5157696c6454616e677a2053322034363030b0496d656469615479706549696d6167652f706e674770726f6a6563745357696c642054616e677a2053657269657320324366757245477265656e44626f64794744656661756c74446579657349537175696e74696e67446e616d65581d57696c642054616e677a202d205365726965732032202d20233433313445696d6167659f47697066733a2f2f583b626166796265696876333634326e366f7a6d356e673670366378367167717367726864356c6c676a77367765326a6c6f62633467696b6c6a64356dff456d6f75746845546565746847657965776561724a53756e676c617373657348636c6f7468696e674d427573696e657373205375697448686561647765617247506c756e6765724a6261636b67726f756e6448426c7565204b6f694b6163636573736f726965734b416e67656c2057696e67734c757067726164656446726f6d4e57696c6454616e677a20343331344e6f726967696e616c506f6c69637958383333353638616431316639336233653739616538646565356164393238646564373261646365613731396539323130386361663135323162466172746973744b406e6574616e656c63686e5157696c6454616e677a2053322034363534b0496d656469615479706549696d6167652f706e674770726f6a6563745357696c642054616e677a205365726965732032436675724542726f776e44626f64794744656661756c7444657965734653746f6e6564446e616d65581d57696c642054616e677a202d205365726965732032202d20233430333345696d6167659f47697066733a2f2f583b62616679626569643779356c35636f72786e763437796f347a787677696e79766c626b7678643636663267636f6f74356b726a6b616e7061376a75ff456d6f7574684a476f6c6420546f6f746847657965776561724c5377696d20476f67676c657348636c6f7468696e6745436c6f616b486865616477656172444e6f6e654a6261636b67726f756e644a4f72616e6779205265644b6163636573736f72696573494e6f73652052696e674c757067726164656446726f6d4e57696c6454616e677a20343033334e6f726967696e616c506f6c69637958383333353638616431316639336233653739616538646565356164393238646564373261646365613731396539323130386361663135323162466172746973744b406e6574616e656c63686e5157696c6454616e677a2053322034353039b0496d656469615479706549696d6167652f706e674770726f6a6563745357696c642054616e677a20536572696573203243667572464f72616e676544626f64794744656661756c74446579657349537175696e74696e67446e616d65581d57696c642054616e677a202d205365726965732032202d20233430303345696d6167659f47697066733a2f2f583b62616679626569646d6a376f617663366b73357671646e37653372646d6b796968656e646b6d6d33746c7072736a75676c787171616e376e6c3471ff456d6f75746845546565746847657965776561724841766961746f727348636c6f7468696e674e486177616969616e2053686972744868656164776561724842656572204861744a6261636b67726f756e644a4f72616e6779205265644b6163636573736f72696573444e6f6e654c757067726164656446726f6d4e57696c6454616e677a20343030334e6f726967696e616c506f6c69637958383333353638616431316639336233653739616538646565356164393238646564373261646365613731396539323130386361663135323162466172746973744b406e6574616e656c63686e5157696c6454616e677a2053322034383735b0496d656469615479706549696d6167652f706e674770726f6a6563745357696c642054616e677a205365726965732032436675724843686172636f616c44626f64794744656661756c74446579657345416e677279446e616d65581d57696c642054616e677a202d205365726965732032202d20233338393445696d6167659f47697066733a2f2f583b626166796265696779726171766b366769697078346366623472716e6d63787962333673767a366965706167737978653568703761776368706769ff456d6f75746845546565746847657965776561724a53756e676c617373657348636c6f7468696e674e486177616969616e205368697274486865616477656172464265616e69654a6261636b67726f756e644d546f626163636f2042726f776e4b6163636573736f72696573444e6f6e654c757067726164656446726f6d4e57696c6454616e677a20333839344e6f726967696e616c506f6c69637958383333353638616431316639336233653739616538646565356164393238646564373261646365613731396539323130386361663135323162466172746973744b406e6574616e656c63686e5157696c6454616e677a2053322034373831b0496d656469615479706549696d6167652f706e674770726f6a6563745357696c642054616e677a205365726965732032436675724843686172636f616c44626f64794744656661756c74446579657346437279696e67446e616d65581d57696c642054616e677a202d205365726965732032202d20233338303445696d6167659f47697066733a2f2f583b62616679626569633466366e767a326a61717135346d78723473326a366c71736272716c3273637578636f6c6d6e6b35796976376d366b77797375ff456d6f7574684642616e616e6147657965776561724a53756e676c617373657348636c6f7468696e674d427573696e6573732053756974486865616477656172484865616462616e644a6261636b67726f756e644f416d65746879737420507572706c654b6163636573736f72696573444e6f6e654c757067726164656446726f6d4e57696c6454616e677a20333830344e6f726967696e616c506f6c69637958383333353638616431316639336233653739616538646565356164393238646564373261646365613731396539323130386361663135323162466172746973744b406e6574616e656c63686e5157696c6454616e677a2053322034373433b0496d656469615479706549696d6167652f706e674770726f6a6563745357696c642054616e677a20536572696573203243667572464f72616e676544626f64794744656661756c7444657965734744656661756c74446e616d65581d57696c642054616e677a202d205365726965732032202d20233338303345696d6167659f47697066733a2f2f583b62616679626569636b6166676e78787a70676d3661676f6769326669773337716632327a677467346e736973347073706c327675646e6f75667575ff456d6f7574684943696761726574746547657965776561724c5377696d20476f67676c657348636c6f7468696e6752547572746c656e65636b205377656174657248686561647765617257507572706c6520616e6420476f6c642042616c6c6361704a6261636b67726f756e644d546f626163636f2042726f776e4b6163636573736f72696573444e6f6e654c757067726164656446726f6d4e57696c6454616e677a20333830334e6f726967696e616c506f6c69637958383333353638616431316639336233653739616538646565356164393238646564373261646365613731396539323130386361663135323162466172746973744b406e6574616e656c63686e5157696c6454616e677a2053322034353438b0496d656469615479706549696d6167652f706e674770726f6a6563745357696c642054616e677a205365726965732032436675724542726f776e44626f64794744656661756c74446579657345416e677279446e616d65581d57696c642054616e677a202d205365726965732032202d20233337373045696d6167659f47697066733a2f2f583b62616679626569633433327672616f6969326b377574656364646e7732656a36707964627a64766f76647a76626665676e6e716c6879646f327a34ff456d6f7574684553747261774765796577656172444e6f6e6548636c6f7468696e674c437265776e65636b205465654868656164776561724842656572204861744a6261636b67726f756e644c4361726d696e652050696e6b4b6163636573736f726965734b416e67656c2057696e67734c757067726164656446726f6d4e57696c6454616e677a20333737304e6f726967696e616c506f6c69637958383333353638616431316639336233653739616538646565356164393238646564373261646365613731396539323130386361663135323162466172746973744b406e6574616e656c63686e5157696c6454616e677a2053322034363634b0496d656469615479706549696d6167652f706e674770726f6a6563745357696c642054616e677a2053657269657320324366757245477265656e44626f64794744656661756c74446579657345416e677279446e616d65581d57696c642054616e677a202d205365726965732032202d20233337313845696d6167659f47697066733a2f2f583b626166796265696466743468783671746261717972683373616d736a6364777562796d6474727463706d797632697a78626f766a7673767a757669ff456d6f7574684b467275697420536e61636b4765796577656172474d6f6e6f636c6548636c6f7468696e674d426f6d626572204a61636b6574486865616477656172484865616462616e644a6261636b67726f756e644d546f626163636f2042726f776e4b6163636573736f72696573494e6f73652052696e674c757067726164656446726f6d4e57696c6454616e677a20333731384e6f726967696e616c506f6c69637958383333353638616431316639336233653739616538646565356164393238646564373261646365613731396539323130386361663135323162466172746973744b406e6574616e656c63686e5157696c6454616e677a2053322034353830b0496d656469615479706549696d6167652f706e674770726f6a6563745357696c642054616e677a2053657269657320324366757244426c756544626f64794744656661756c74446579657349426c61636b20457965446e616d65581d57696c642054616e677a202d205365726965732032202d20233336393145696d6167659f47697066733a2f2f583b626166796265696132793565346a7363746477786a77706c7632717266356c666968776f34773778666d766a3775626e6e7173706279646e6b7034ff456d6f75746845546565746847657965776561724a53756e676c617373657348636c6f7468696e674d426f6d626572204a61636b65744868656164776561724e436c6f7665722042616c6c6361704a6261636b67726f756e644d546f626163636f2042726f776e4b6163636573736f726965734d47616d65646179205061696e744c757067726164656446726f6d4e57696c6454616e677a20333639314e6f726967696e616c506f6c69637958383333353638616431316639336233653739616538646565356164393238646564373261646365613731396539323130386361663135323162466172746973744b406e6574616e656c63686e5157696c6454616e677a2053322034373335b0496d656469615479706549696d6167652f706e674770726f6a6563745357696c642054616e677a2053657269657320324366757244426c756544626f64794744656661756c74446579657349537175696e74696e67446e616d65581d57696c642054616e677a202d205365726965732032202d20233335393745696d6167659f47697066733a2f2f583b62616679626569686270376363347966746d366f75796e71756e6e65616c327a786c72717a786d6564766e716f676d68656d3735366b646a777a75ff456d6f7574684943696761726574746547657965776561724a53756e676c617373657348636c6f7468696e674e486177616969616e2053686972744868656164776561724a4865616470686f6e65734a6261636b67726f756e6448426c7565204b6f694b6163636573736f72696573444e6f6e654c757067726164656446726f6d4e57696c6454616e677a20333539374e6f726967696e616c506f6c69637958383333353638616431316639336233653739616538646565356164393238646564373261646365613731396539323130386361663135323162466172746973744b406e6574616e656c63686e04d8799f581c2d2b94c39535a96f74cedf908b704eda9605c7d4e788a0e9f0a133c7581de19ae47a5fee22019c8d426ad9638dbff902fbbd7422966b4361c56829ffff
```

Please use a datum inspector to validate the output JSON (e.g., https://cardanoscan.io/datumInspector)

The asset `asset143k4tan37qqx53u36an5smssd9yumvwc58zfxv` should render with this JSON from the CBOR into a well-formed NFT:

```
{
   "map_11":{
      "mapKey":"WildTangz S2 4664",
      "mapValue":{
         "map_0":{
            "mapKey":"mediaType",
            "mapValue":"image/png"
         },
         "map_1":{
            "mapKey":"project",
            "mapValue":"Wild Tangz Series 2"
         },
         "map_2":{
            "mapKey":"fur",
            "mapValue":"Green"
         },
         "map_3":{
            "mapKey":"body",
            "mapValue":"Default"
         },
         "map_4":{
            "mapKey":"eyes",
            "mapValue":"Angry"
         },
         "map_5":{
            "mapKey":"name",
            "mapValue":"Wild Tangz - Series 2 - #3718"
         },
         "map_6":{
            "mapKey":"image",
            "mapValue":[
               "ipfs://",
               "bafybeidft4hx6qtbaqyrh3samsjcdwubymdtrtcpmyv2izxbovjvsvzuvi"
            ]
         },
         "map_7":{
            "mapKey":"mouth",
            "mapValue":"Fruit Snack"
         },
         "map_8":{
            "mapKey":"eyewear",
            "mapValue":"Monocle"
         },
         "map_9":{
            "mapKey":"clothing",
            "mapValue":"Bomber Jacket"
         },
         "map_10":{
            "mapKey":"headwear",
            "mapValue":"Headband"
         },
         "map_11":{
            "mapKey":"background",
            "mapValue":"Tobacco Brown"
         },
         "map_12":{
            "mapKey":"accessories",
            "mapValue":"Nose Ring"
         },
         "map_13":{
            "mapKey":"upgradedFrom",
            "mapValue":"WildTangz 3718"
         },
         "map_14":{
            "mapKey":"originalPolicy",
            "mapValue":"33568ad11f93b3e79ae8dee5ad928ded72adcea719e92108caf1521b"
         },
         "map_15":{
            "mapKey":"artist",
            "mapValue":"@netanelchn"
         }
      }
   }
}
```

FYI @SmaugPool @Scitz0

Ryun1 (approved):

thaddeusdiamond:
@SmaugPool please let me know the plan for implementation when you can I know it's more work but I don't want to consolidate and rug everyone before they render.

PR #1115, opened 2025-11-26
CIP-0088, CIP-0151 | update to Active

Crypto2099:
Both CIP-88 and CIP-151 have seen heavy and active development and main network usage throughout 2025. With the rise of Ekklesia voting for CIP-151 and VeriGlyph publication and validation for CIP-88 I would argue that both of these standards have risen to the level of being "active" on chain and should be adopted as such.

### DIFF
--- a/CIP-0088/README.md
+++ b/CIP-0088/README.md
@@ -2,12 +2,15 @@
 CIP: 88
 Title: Token Policy Registration
 Category: Tokens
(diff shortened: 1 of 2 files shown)


rphair:
OK @Ryun1 we effectively have quorum so I'm putting this through (cc @perturbing so any objection might be corrected later)

PR #1117, opened 2025-11-29
CIP-???? | Governance Participant Compensation

Thomas-nada:
This  proposal introduces Governance Participant Compensation to provide automated, deterministic compensation for Cardano governance actors (CC, DReps, SPOs) who cast a valid vote.

It achieves this by:

Funding: Consuming a portion of the proposal's deposit (not the Treasury) upon finalization.

Distribution: Using new ledger rules to split the funds into pools and distributing them via a hybrid model (equal share + stake-weighted) only to active voters.

Mechanism: Operating entirely within the ledger rules (like staking rewards) to ensure reliability and trustlessness.

Goal: Ensure sustainable, decentralized governance by making participation economically feasible.

---

([rendered latest proposal](https://github.com/Thomas-nada/CIPs/blob/Compensatory-Ledger-Mechanism/CIP-%3F%3F%3F%3F/README.md))

### DIFF
--- a/CIP-????/README.md
+++ b/CIP-????/README.md
@@ -6,8 +6,8 @@ Status: Proposed
 Authors:
     - Thomas Lindseth <thomas.lindseth@intersectmbo.org>
 Implementors: []
-Discussions:
-    - 
+Discussions: 
+    - https://github.com/cardano-foundation/CIPs/pull/1117
 Created: 2025-11-29
 License: CC-BY-4.0
 ---
--- a/CIP-????/README.md
+++ b/CIP-????/README.md
@@ -221,10 +221,10 @@ By providing guaranteed compensation:
 
 This CIP becomes **Active** when:
 
-* All new protocol parameters are implemented in the ledger.  
-* Distribution logic executes correctly on testnet.  
-* A governance action on mainnet distributes compensation correctly.  
-* Sufficient community and technical review confirms correctness and consensus.
+* [ ] All new protocol parameters are implemented in the ledger.  
+* [ ] Distribution logic executes correctly on testnet.  
+* [ ] A governance action on mainnet distributes compensation correctly.  
+* [ ] Sufficient community and technical review confirms correctness and consensus.
 
 ---
 
--- a/CIP-????/README.md
+++ b/CIP-????/README.md
@@ -241,4 +241,4 @@ This CIP becomes **Active** when:
 
 ## Copyright
 
-This CIP is licensed under **CC-BY-4.0**.
+This CIP is licensed under [CC-BY-4.0](https://creativecommons.org/licenses/by/4.0/legalcode).
--- a/CIP-????/README.md
+++ b/CIP-????/README.md
@@ -221,22 +221,21 @@ By providing guaranteed compensation:
 
 This CIP becomes **Active** when:
 
-1. All new protocol parameters are implemented in the ledger.  
-2. Distribution logic executes correctly on testnet.  
-3. A governance action on mainnet distributes compensation correctly.  
-4. Ledger events expose compensation in a standardized, wallet-readable format.  
-5. Sufficient community and technical review confirms correctness and consensus.
+* All new protocol parameters are implemented in the ledger.  
+* Distribution logic executes correctly on testnet.  
+* A governance action on mainnet distributes compensation correctly.  
+* Sufficient community and technical review confirms correctness and consensus.
 
 ---
 
 ### Implementation Plan
 
-1. **Community Discussion & Review**  
-2. **Ledger Engineering Review**  
-3. **Implementation**  
-4. **Testnet Deployment**  
-5. **Hard Fork Activation**  
-6. **Post-Deployment Monitoring**
+* **Community Discussion & Review**  
+* **Ledger Engineering Review**  
+* **Implementation**  
+* **Testnet Deployment**  
+* **Hard Fork Activation**  
+* **Post-Deployment Monitoring**
 
 ---
 
--- a/CIP-????/README.md
+++ b/CIP-????/README.md
@@ -102,7 +102,7 @@ The remainder is refund-eligible for the proposer if the action passes.
 
 ### Distribution Rules
 
-#### Compensation to Constitutional Committee (Equal Share)**
+#### Compensation to Constitutional Committee (Equal Share)
 
 Only voting CC members are compensated.
 
@@ -114,7 +114,7 @@ Non-voting CC members receive 0.
 
 ---
 
-#### DRep Compensation (Hybrid Model)**
+#### DRep Compensation (Hybrid Model)
 
 ##### Equal-Share Portion
 ```
@@ -138,7 +138,7 @@ Payout_DRep = Payout_Equal_DRep + Payout_Weighted_DRep
 
 ---
 
-#### **4.3 SPO Compensation (Hybrid Model)**
+#### SPO Compensation (Hybrid Model)
 
 ##### Equal-Share Portion
 ```
@@ -147,7 +147,7 @@ Payout_Equal_SPO = EqualShareAmount / numberOfVotingSPOs
 ```
 
 ##### Stake-Weighted Portion  
-Weighted by **block-producing active stake**.
+Weighted by **block-producing active stake.
 
 ```
 WeightedAmount = SPO_Pool × spoWeightFactor
@@ -175,7 +175,7 @@ This internalizes governance processing costs.
 
 ## Rationale: how does this CIP achieve its goals?
 
-### Ledger-Native Implementation Ensures Maximum Reliability**
+### Ledger-Native Implementation Ensures Maximum Reliability
 All compensation logic is implemented inside ledger rules.  
 This guarantees:
 
@@ -186,7 +186,7 @@ This guarantees:
 
 ---
 
-### Incentive Alignment Through Hybrid Distribution**
+### Incentive Alignment Through Hybrid Distribution
 The hybrid model balances:
 
 - **Equal-share rewards** → compensate time, labor, and research  
@@ -196,7 +196,7 @@ This prevents domination by large actors while recognizing the security contribu
 
 ---
 
-### Fair Cost Attribution via Proposer Deposits**
+### Fair Cost Attribution via Proposer Deposits
 Because proposer deposits fund compensation:
 
 - Governance costs are paid by those creating that cost  
@@ -206,7 +206,7 @@ Because proposer deposits fund compensation:
 
 ---
 
-### Promotes Broad Participation and Decentralization**
+### Promotes Broad Participation and Decentralization
 By providing guaranteed compensation:
 
 - Smaller actors can sustainably participate  
--- a/CIP-????/README.md
+++ b/CIP-????/README.md
@@ -1,7 +1,7 @@
 ---
 CIP: XXXX
 Title: Compensatory Ledger Mechanism (CLM)
-Category: Metadata
+Category: Ledger
 Status: Proposed
 Authors:
     - Thomas Lindseth <thomas.lindseth@intersectmbo.org>
--- a/CIP-????/README.md
+++ b/CIP-????/README.md
@@ -1,0 +1,245 @@
+---
+CIP: XXXX
+Title: Compensatory Ledger Mechanism (CLM)
+Category: Metadata
+Status: Proposed
+Authors:
+    - Thomas Lindseth <thomas.lindseth@intersectmbo.org>
+Implementors: []
+Discussions:
+    - 
+Created: 2025-11-29
+License: CC-BY-4.0
+---
+
+## Abstract
+
+This CIP introduces the **Compensatory Ledger Mechanism (CLM)**—a deterministic, protocol-level compensation system for governance actors participating in Cardano governance actions. A portion of each governance action’s deposit is automatically distributed to Constitutional Committee (CC) members, Delegated Representatives (DReps), and Stake Pool Operators (SPOs) who cast a valid vote. 
+
+This mechanism requires **no smart contracts**, **no Treasury withdrawals**, and **no manual triggers**, operating fully within Cardano’s ledger rules in a manner analogous to staking rewards. It creates an incentive-aligned, sustainable, and decentralized compensation model that reflects the real cost of governance participation.
+
+---
+
+## Motivation: why is this CIP necessary?
+
+Cardano’s decentralized governance framework requires persistent engagement from CC members, DReps, and SPOs. These actors must analyze governance actions, evaluate technical and economic implications, and cast informed votes. This work incurs ongoing time, expertise, and operational costs. Without a reliable compensation model, several risks emerge:
+
+### **1. Governance Fatigue**
+Uncompensated governance labor leads to declining participation over time.
+
+### **2. Centralization Risks**
+Only large, well-funded entities can consistently afford to participate, reducing diversity and representation.
+
+### **3. Low Governance Participation / Quorum Instability**
+Without compensation, new and smaller actors lack incentives to engage, harming governance quality.
+
+### **4. Misaligned Incentives**
+Proposers incur no cost commensurate with the burden placed on reviewers and voters.
+
+### **5. Need for a Deterministic, Automated Mechanism**
+Any off-chain or trust-based compensation scheme introduces friction and reduces security.
+
+The CLM addresses these issues by:
+- Making governance participation economically feasible  
+- Ensuring compensation flows only to active participants  
+- Funding compensation from proposer deposits, aligning costs with governance load  
+- Executing payments entirely through the ledger for maximum reliability  
+
+---
+
+## Specification
+
+### 1. New Protocol Parameters
+
+The following fields are added to `PParams`:
+
+| Parameter | Type | Description |
+|----------|------|-------------|
+| `ccCompensationRate` | UnitInterval | Portion of `govDeposit` allocated to CC. |
+| `drepCompensationRate` | UnitInterval | Portion allocated to DReps. |
+| `spoCompensationRate` | UnitInterval | Portion allocated to SPOs. |
+| `drepWeightFactor` | UnitInterval | Fraction of DRep pool distributed stake-weighted. |
+| `spoWeightFactor` | UnitInterval | Fraction of SPO pool distributed active-stake-weighted. |
+
+**Constraint:**  
+`ccCompensationRate + drepCompensationRate + spoCompensationRate ≤ MAX_COMP_RATE`  
+where `MAX_COMP_RATE` is hard-coded (e.g., 0.25).
+
+---
+
+### 2. Compensation Trigger
+
+Upon finalization of any governance action requiring a deposit—whether it **passes**, **fails**, or **expires**—the ledger automatically:
+
+1. Finalizes deposit accounting  
+2. Creates the compensation pools  
+3. Distributes funds to voting governance actors  
+
+This occurs atomically during governance-action completion.
+
+---
+
+### 3. Compensation Pot Creation
+
+Let `govDeposit` be the required deposit for the governance action.
+
+```
+TotalCompensation =
+  govDeposit × (ccCompensationRate + drepCompensationRate + spoCompensationRate)
+```
+
+Split into:
+
+```
+CC_Pool   = govDeposit × ccCompensationRate
+DRep_Pool = govDeposit × drepCompensationRate
+SPO_Pool  = govDeposit × spoCompensationRate
+```
+
+The remainder is refund-eligible for the proposer if the action passes.
+
+---
+
+### 4. Distribution Rules
+
+#### **4.1 Compensation to Constitutional Committee (Equal Share)**
+
+Only voting CC members are compensated.
+
+```
+Payout_CC = CC_Pool / numberOfVotingCCMembers
+```
+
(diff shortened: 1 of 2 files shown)


CIP-????/README.md
```diff
@@ -0,0 +1,244 @@
+---
+CIP: XXXX
+Title: Compensatory Ledger Mechanism (CLM)
+Category: Metadata
```
```suggestion
Category: Ledger
```
This substantially changes `Ledger` behaviour, as well as adding new protocol parameters, so this goes far beyond `Metadata`.

CIP-????/README.md
```diff
@@ -0,0 +1,244 @@
...
+
+This mechanism requires **no smart contracts**, **no Treasury withdrawals**, and **no manual triggers**, operating fully within Cardano’s ledger rules in a manner analogous to staking rewards. It creates an incentive-aligned, sustainable, and decentralized compensation model that reflects the real cost of governance participation.
+
+---
+
+## Motivation: why is this CIP necessary?
+
+Cardano’s decentralized governance framework requires persistent engagement from CC members, DReps, and SPOs. These actors must analyze governance actions, evaluate technical and economic implications, and cast informed votes. This work incurs ongoing time, expertise, and operational costs. Without a reliable compensation model, several risks emerge:
+
+### **1. Governance Fatigue**
```
```suggestion
### Governance Fatigue
```
(example; continuing below) These section headings are boldfaced automatically (so the Markdown boldfacing is redundant) and the heading titles themselves provide external references so the numbering, as one would for a printed document, is effectively obsolete.

CIP-????/README.md
```diff
@@ -0,0 +1,244 @@
...
+## Path to Active
+
+### Acceptance Criteria
+
+This CIP becomes **Active** when:
+
+* All new protocol parameters are implemented in the ledger.  
+* Distribution logic executes correctly on testnet.  
+* A governance action on mainnet distributes compensation correctly.  
+* Sufficient community and technical review confirms correctness and consensus.
```
```suggestion
* [ ] All new protocol parameters are implemented in the ledger.  
* [ ] Distribution logic executes correctly on testnet.  
* [ ] A governance action on mainnet distributes compensation correctly.  
* [ ] Sufficient community and technical review confirms correctness and consensus.
```
FYI the "tick-box" approach tends to work well on these lists in other CIP's (though it's at author & perhaps editor discretion to use them).

CIP-????/README.md
```diff
@@ -0,0 +1,244 @@
...
+* **Implementation**  
+* **Testnet Deployment**  
+* **Hard Fork Activation**  
+* **Post-Deployment Monitoring**
+
+---
+
+## Copyright
+
+This CIP is licensed under **CC-BY-4.0**.
```
```suggestion
This CIP is licensed under [CC-BY-4.0](https://creativecommons.org/licenses/by/4.0/legalcode).
```
It's more polite _and_ more useful to link the legal document rather than simply boldfacing its title. 😅

CIP-????/README.md
```diff
@@ -0,0 +1,244 @@
+---
+CIP: XXXX
+Title: Compensatory Ledger Mechanism (CLM)
```
```suggestion
Title: Compensatory Ledger Mechanism
```
It's neither necessary nor customary to provide an acronym when it's as obvious as this: especially when (as far as I know) no prior documentation is using that acronym for this purpose.

rphair (commented):
Marked for `Triage` in next CIP meeting https://hackmd.io/@cip-editors/124.

CIP-????/README.md
Thomas-nada: Updated

rphair: partially resolved with https://github.com/cardano-foundation/CIPs/pull/1117/commits/d47cb499ba2e70ebf73a1fbf463fb355d85f79ec (since the numbering continues through the _Specification_ and _Rationale_) but I'll leave it to the other editors to determine whether this is a review point to insist upon.

rphair: fully resolved with https://github.com/cardano-foundation/CIPs/pull/1117/commits/afc77ab9904eb583d023afe9777156338dc7e4c1 (though _Implementation Plan_ is still superfluously boldfaced and could also use the tick-boxes).

CIP-????/README.md
```diff
@@ -0,0 +1,311 @@
...
+This mechanism requires **no smart contracts**, **no Treasury withdrawals**, and **no manual triggers**, operating fully within Cardano’s ledger rules in a manner analogous to staking rewards. It creates an incentive-aligned, sustainable, and decentralized compensation model that reflects the real cost of governance participation.
+
+---
+
+## Motivation: why is this CIP necessary?
+
+Cardano’s decentralized governance framework requires persistent engagement from CC members, DReps, and SPOs. These actors must analyze governance actions, evaluate technical and economic implications, and cast informed votes. This work incurs ongoing time, expertise, and operational costs. Without a reliable compensation model, several risks emerge:
+
+### Governance Fatigue
+Uncompensated governance labor leads to declining participation over time.
```
Ryun1, line 28: have we seen this?
do we have numbers on this?

CIP-????/README.md
```diff
@@ -0,0 +1,311 @@
...
+
+### Centralization Risks
+Only large, well-funded entities can consistently afford to participate, reducing diversity and representation.
+
+### Low Governance Participation / Quorum Instability
```
Ryun1, line 33: Probably roll this one into Centralization Risks

CIP-????/README.md
```diff
@@ -0,0 +1,311 @@
...
+Without compensation, new and smaller actors lack incentives to engage, harming governance quality.
+
+### Misaligned Incentives
+Proposers incur no cost commensurate with the burden placed on reviewers and voters.
```
Ryun1, line 37: Proposers do incur the opportunity cost of lost staking rewards

CIP-????/README.md
```diff
@@ -0,0 +1,311 @@
...
+License: CC-BY-4.0
+---
+
+## Abstract
+
+This CIP introduces the **Compensatory Ledger Mechanism (CLM)**—a deterministic, protocol-level compensation system for governance actors participating in Cardano governance actions. A portion of each governance action’s deposit is automatically distributed to Constitutional Committee (CC) members, Delegated Representatives (DReps), and Stake Pool Operators (SPOs) who cast a valid vote.
+
+This mechanism requires **no smart contracts**, **no Treasury withdrawals**, and **no manual triggers**, operating fully within Cardano’s ledger rules in a manner analogous to staking rewards. It creates an incentive-aligned, sustainable, and decentralized compensation model that reflects the real cost of governance participation.
+
+---
```
Ryun1, line 21: please also consider addressing the points from [CPS-20 Governance Stakeholder Incentivization](https://github.com/cardano-foundation/CIPs/tree/master/CPS-0020)

CIP-????/README.md
```diff
@@ -0,0 +1,311 @@
...
+CIP: XXXX
+Title: Compensatory Ledger Mechanism
```
Ryun1:
```suggestion
Title: Governance Participant Compensation
```

I feel something referring to Governance Participants is probably more accurate and obvious at first glance

CIP-????/README.md
```diff
@@ -0,0 +1,311 @@
...
+Only large, well-funded entities can consistently afford to participate, reducing diversity and representation.
+
+### Low Governance Participation / Quorum Instability
+Without compensation, new and smaller actors lack incentives to engage, harming governance quality.
+
+### Misaligned Incentives
+Proposers incur no cost commensurate with the burden placed on reviewers and voters.
+
+### Need for a Deterministic, Automated Mechanism
+Any off-chain or trust-based compensation scheme introduces friction and reduces security.
```
Ryun1, line 40: maybe worth discussing a smart contract based solution

CIP-????/README.md
```diff
@@ -0,0 +1,311 @@
...
+|----------|------|-------------|
+| `ccCompensationRate` | UnitInterval | Portion of `govDeposit` allocated to CC. |
+| `drepCompensationRate` | UnitInterval | Portion allocated to DReps. |
+| `spoCompensationRate` | UnitInterval | Portion allocated to SPOs. |
+| `drepWeightFactor` | UnitInterval | Fraction of DRep pool distributed stake-weighted. |
+| `spoWeightFactor` | UnitInterval | Fraction of SPO pool distributed active-stake-weighted. |
+
+**Constraint:**  
+`ccCompensationRate + drepCompensationRate + spoCompensationRate ≤ MAX_COMP_RATE`  
+where `MAX_COMP_RATE` is hard-coded (e.g., 0.25).
```
Ryun1, line 66: probably not good to hard-code
best to have everything as an adjustable protocol parameter

CIP-????/README.md
```diff
@@ -0,0 +1,311 @@
...
+
+1. Finalizes deposit accounting  
+2. Creates the compensation pools  
+3. Distributes funds to voting governance actors  
+
+This occurs atomically during governance-action completion.
+
+---
+
+## Compensation Eligibility Based on Relevant Governance Actor Groups
```
Ryun1:
```suggestion
### Compensation Eligibility Based on Relevant Governance Actor Groups
```

CIP-????/README.md
rphair: yes... my first reading of the title also (before I got around to reading the document) was that this was some accounting capability and using "compensatory" in the mathematical sense (rather than "compensation").

CIP-????/README.md
```diff
@@ -0,0 +1,311 @@
...
+
+```
+Payout_CC = CC_Pool / numberOfVotingCCMembers
+```
+
+Non-voting CC members receive 0.
+
+---
+
+#### DRep Compensation (Hybrid Model)
```
Ryun1, line 180: Is there consideration of the IOR proposed DRep compensation scheme?

CIP-????/README.md
```diff
@@ -0,0 +1,311 @@
...
+
+---
+
+### Deposit Finality
+
+The compensation portion of the deposit corresponding to EffectiveRate(group) values greater than zero is consumed. Any portion corresponding to groups with EffectiveRate(group) = 0 (i.e., non-relevant governance actor groups) is not consumed and remains refundable.
+
+---
+
+## Rationale: how does this CIP achieve its goals?
```
Ryun1, line 243: worth discussing that a mechanism like this would greatly increase the cost to proposers, could this have knock-on negative effects?

CIP-????/README.md
```diff
@@ -0,0 +1,311 @@
...
+## Motivation: why is this CIP necessary?
+
+Cardano’s decentralized governance framework requires persistent engagement from CC members, DReps, and SPOs. These actors must analyze governance actions, evaluate technical and economic implications, and cast informed votes. This work incurs ongoing time, expertise, and operational costs. Without a reliable compensation model, several risks emerge:
+
+### Governance Fatigue
+Uncompensated governance labor leads to declining participation over time.
+
+### Centralization Risks
+Only large, well-funded entities can consistently afford to participate, reducing diversity and representation.
+
```
Ryun1, line 32: from CIPs call, another potential benefit for the proposed design
- no reliance from volunteering

CIP-????/README.md
```diff
@@ -0,0 +1,311 @@
...
+---
+
+### Deposit Finality
+
+The compensation portion of the deposit corresponding to EffectiveRate(group) values greater than zero is consumed. Any portion corresponding to groups with EffectiveRate(group) = 0 (i.e., non-relevant governance actor groups) is not consumed and remains refundable.
+
+---
+
+## Rationale: how does this CIP achieve its goals?
+
```
Ryun1, line 244: From @Cerkoryn 

With this CIP it solves the compensation issue for all three branches by adding an actual monetary cost to GA deposits.  This could give us the flexibility to dial back govActionDeposit gradually and allow more of the deposit to go toward paying the governance actors that are required to do work BECAUSE of said GA

Cerkoryn: To take it a bit further, I think this is the only compensation mechanism I've seen that is inherently scalable AND sustainable.

Since compensation comes from the proposers (instead of the treasury or reserve), we don't really have to worry about draining either of those things faster.  And since proposers pay per proposal, governance actors will inherently earn more if there are more proposals (more work to do) and visa versa.

CIP-????/README.md
Cerkoryn: I might even argue that we could do without `MAX_COMP_RATE` altogether.

In theory we could have 100% of `govActionDeposit` be usable for compensation and then lower it a bunch.  Right now giving up 6 epochs worth of staking rewards on 100k ADA costs about ~200 ADA.  If we instead had `govActionDeposit` at something like 1000 ADA, we could make governance more accessible, compensate governance actors fairly, AND mitigate spam proposals all at the same time.

Those are just hypothetical numbers and we definitely should make incremental changes instead of all at once.  But the idea is that with this CIP it would all be a simple matter of just fine-tuning a few parameters.

CIP-????/README.md
fallen-icarus: I'm not sure I agree. A small governance deposit (e.g., 1000 ADA) seems unlikely to cover the cost of dReps and CC member time considering the deposit will be split across all representatives. If it costs 10,000-100,000 ADA to cover everyone's time, now you have a significant barrier to proposing anything - even things Cardano desperately needs. Some people are likely only comfortable proposing things today *because* it is just a deposit. Weren't several proposal deposits crowdfunded? Would that crowdfunding have succeeded if the deposit was confiscated to pay representatives? I'm not so sure...

---
Right now we are all paying 20% into the Cardano treasury every transaction. I would honestly rather just take 0.1-0.2% of that and split it among the representatives each epoch. (I'm guessing on what meaningful proportions would be.)

fallen-icarus:
The idea I brought up in the CIP meeting was to use [CIP-159](https://github.com/cardano-foundation/CIPs/pull/1061) for micro-taxes:

_When a user delegates to a dRep, their wallet automatically subscribes them into paying 0.001 ADA to their dRep every time they make a tx._

The user can always opt-out, but opt-out still typically sees >80% participation rates. We can do the same sort of micro-taxes for funding other entities like the CC or stake pool operators running the testnets.

CIP-????/README.md
Thomas-nada: True, but is that considered cost? They are not paying for it. They incur potential losses, not cost.

CIP-????/README.md
Thomas-nada: Agreed. If a smart contract solution can do the same with less changes to the ledger I am all for it.

CIP-????/README.md
Thomas-nada: The intent of this was to eliminate the possibility of overcharging. 
I am all for fair compensation, but I don't want being an SPO, a DREP or a CC member to potentially become a "get rich quick" scheme.  So to protect that I intruduced a max cap. 
This could obviously be a variable parameter, and the 25% was just put in as an example that made somewhat sense when I was playing around with the parameters in various scenarios.

CIP-????/README.md
Thomas-nada: I have barely read that one. So not it has not been considered.

CIP-????/README.md
Thomas-nada: In the short run, maybe. In the long run, I believe that eventually we will hit a sweet spot where the govActionDeposit parameter could be lowered to a point where the governance actors are pleased with the compensation they recieve as well as the proposers are OK to pay this amount. 
A bonus to this solution is that it will make proposers think twice about what they submit as there is an actual cost attached to it. This would allow for lowering the govActionDeposit while still keeping spam to a minimum.

CIP-????/README.md
Thomas-nada: I like the idea of the treasury paying for compensation, but correct me if I am wrong, does not every withdrawal from the treasury currently require approval through a governace action?

Thomas-nada:
> The idea I brought up in the CIP meeting was to use [CIP-159](https://github.com/cardano-foundation/CIPs/pull/1061) for micro-taxes:
> 
> _When a user delegates to a dRep, their wallet automatically subscribes them into paying 0.001 ADA to their dRep every time they make a tx._
> 
> The user can always opt-out, but opt-out still typically sees >80% participation rates. We can do the same sort of micro-taxes for funding other entities like the CC or stake pool operators running the testnets.

With [CIP-149 ](https://github.com/cardano-foundation/CIPs/tree/master/CIP-0149) currently being rolled out by wallet providers, SDKs and explorers we will be able to see the adpotion rate of a similar approach and also how the distribution rate will be. 
If this is a success I am happy to go this route. But I am for the most against relying on voluntarism as the backbone of compensation. I draw similarities with waiters who are poorly paid because their employers know the waiters can rely on tips.

rphair:
Thanks @Ryun1 @Cerkoryn @fallen-icarus @Thomas-nada for getting all your feedback from today's CIP meeting into GitHub.  We've resolved at that meeting to leave this `Unconfirmed` but not for lack of interest in the potential of this proposal... only because the feedback could influence some key points of the specification.

Once these suggestions appear to be settled, we can get this back onto the meeting agenda for confirmation & assignment of a CIP number: perhaps at our next scheduled CIP meeting in the New Year (https://hackmd.io/@cip-editors/125).

fallen-icarus:
@Thomas-nada CIP-149 is not the same as what I'm suggesting. There is a huge difference between *opt-in* vs *opt-out* mechanisms.

- Opt-in usually gets about 15-20% participation.
- Out-out usually gets about 80-90% participation.

>  I draw similarities with waiters who are poorly paid because their employers know the waiters can rely on tips.

This example is *opt-in* - customers must opt-in to paying a tip (of any size) to their waiter.

As much as people complain about opt-out, *most people don't actually opt-out*. With this in mind, I actually expect very low adoption of CIP-149 precisely because it is *opt-in*. (Also, I think its UX will be much worse than what I'm suggesting with CIP-159.)

CIP-????/README.md
fallen-icarus: The way I'm thinking about it, it would not go *through* the treasury. Staking rewards don't go through the treasury; they are automatically calculated and distributed at each epoch boundary. We could do something similar for representative compensation. So each epoch boundary: 19.8% would go to the treasury and 0.2% would be distributed to representatives automatically.

CIP-????/README.md
Ryun1: I feel probably best to mention this

CIP-????/README.md
Ryun1: >  believe that eventually we will hit a sweet spot where the govActionDeposit parameter could be lowered to a point where the governance actors are pleased with the compensation they recieve as well as the proposers are OK to pay this amount.

> This would allow for lowering the govActionDeposit while still keeping spam to a minimum.

good points, please ensure theyre included in the CIP :)

CIP-????/README.md
Ryun1: might be worth reading :)

Cerkoryn: @Ryun1 are you by chance referring to the Threshold(k) mechanism proposed in this paper?

https://iohk.io/en/research/library/papers/reward-schemes-and-committee-sizes-in-proof-of-stake-governance

Cerkoryn: If so then I don't think they are exclusive.  In fact I think they complement each other.

As an example, we could set this Threshold(k) parameter to 100 to reward the top 100 dReps _equally_ (as explained in the paper).  But those rewards would come from taking a commission out of each `govActionDeposit` as a percent for dReps, SPOs, and the CC respectively (with exceptions for GAs that their votes aren't counted on) as explained by this CIP.

You could have it such that the dRep fee is 1% of a GovAction (so 1k ADA) and then set this other Threshold(k) to target the top 100 dReps.  Then each of those dReps would get 10 ADA per GA that they vote on.

It may not be much but it's sustainable and solves both the compensation issue and mitigates proposal spam by imposing a small cost on proposers.

rphair (commented):
@Thomas-nada as it says in CIP-0001: submissions to this repository often require an _advocate_ (usually the author) who posts regularly to keep bringing editorial & peer review together with community statements & ideas and providing responses to feedback... with requests for re-review every time consensus needs to be, or appears to be, achieved.

So far this thread has been like a posting board for ideas on the subject, without any emphatic confirmation that the proposed method would be acceptable to anyone.  But perhaps now that this has been merged (cc @maureenblack @daniellestanko-gov - with maybe @thenic95 also interested):
* [CPS-0033 | DRep Voting Power Concentration](https://github.com/cardano-foundation/CIPs/tree/master/CPS-0033)

... we might clarify these discussions, and you might offer this proposal back to the relevant communities — to see if they can provide some evidence that they would find this proposal acceptable — and to implement some changes if necessary to gain their support.
* Of course you would be welcome to use this thread to hold such discussions: as the CIP Editors generally hope for in the review process.
* If we again see some progress beyond the above inconclusive review threads, we can get this proposal back in the CIP Meeting review agenda for hopeful confirmation as a CIP candidate (with a number) & then consideration as a real technical proposition of interest to the governance community.

Conversely: without that progress, this proposal would proceed toward closure: not because it isn't a good idea, but because we can't use it as a CIP without the vital component of conclusive peer review.  cc @Ryun1 @perturbing + reviewers @Cerkoryn @fallen-icarus

rphair:
@Thomas-nada perhaps you could check with the Civics Committee — sharing with them the comments above, about why this proposal would have to be closed without any further response or advocacy — and see if you or any other advocate could respond to review enough to confirm & maybe someday merge this?  Otherwise we would have to close it due to apparent abandonment.

PR #1107, opened 2025-10-28
CIP-0179? | On-Chain Surveys and Polls

Thomas-nada:
This is a new **Metadata CIP** proposing a standard for **on-chain surveys and community sentiment polling**.

The goal is to provide a **single, standardized, transparent, and reliable way** for any entity (Foundations, DReps, projects, or community members) to conduct simple, structured consultations directly on-chain — without requiring deposits or governance actions.

This proposal defines two minimal **CIP-0068 (label 674)** metadata payloads:

* **Survey Definition Payload:** Specifies the survey question, available options, eligibility criteria, and vote aggregation method (`StakeBased` or `CredentialBased`). It may optionally reference a governance action by transaction ID and action index.
* **Survey Response Payload:** Enables any eligible stakeholder to submit their response, immutably linked to the originating survey transaction ID.

This standard provides a lightweight, interoperable foundation for ecosystem-wide sentiment gathering — allowing explorers, wallets, and dApps to discover, display, and aggregate survey results in a consistent, transparent way.

---

([rendered latest proposal version](https://github.com/Thomas-nada/CIPs/blob/poll-metadata-standard/CIP-0179/README.md))

PR #1119, opened 2025-12-09
Add metadata label 17 for on-chain surveys and polls

Thomas-nada:
This PR updates `CIP-0010/registry.json` to add a new metadata label:

**Label:** 17  
**Use case:** On-chain surveys and polls metadata  
**Author:** Thomas Lindseth  
**Purpose:** Defines a standardized metadata format for creating and responding to on-chain surveys and polls.

This label is referenced by the proposed CIP "On-Chain Surveys and Polls":
https://github.com/cardano-foundation/CIPs/pull/1107

### DIFF
--- a/CIP-0088/README.md
+++ b/CIP-0088/README.md
@@ -2,12 +2,15 @@
 CIP: 88
 Title: Token Policy Registration
 Category: Tokens
-Status: Proposed
+Status: Active
 Authors:
-- Adam Dean <adam@crypto2099.io>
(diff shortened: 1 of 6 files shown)


rphair:
Closing as per https://github.com/cardano-foundation/CIPs/pull/1107#issuecomment-3635072112 since the CIP-0010 update should be made in the CIP pull request which introduces the need for that metadata label (in cases that there _is_ such a CIP).

PR #1124, opened 2026-01-05
Do not specify per-transaction plutus limits

ch1bo:
See note in protocol parameters section why increasing the per-transaction limit poses a risk. This changes the specification to avoid this risk and deliberately pass on the potential upside.

This is based on a discussion from a past CIP editor meeting: https://github.com/cardano-foundation/CIPs/pull/1078#issuecomment-3353472016


ch1bo:
Was intended as PR onto the already existing PR, superseded by https://github.com/cardano-scaling/CIPs/pull/28

PR #1134, opened 2026-01-16
CIP-0153 | Tighten unValueData canonicality, add scaleValue builtin

colll78:
Specify unValueData canonicality checks: require strictly ascending, distinct keys for both outer (currency symbols) and inner (token names) maps; fail on empty inner maps and zero-quantity entries; keep existing bounds/size constraints.

Add scaleValue builtin to the operation list: multiply all quantities by a provided integer, drop zeros / remove empty inner maps to preserve invariants, and fail on out-of-bounds results.

This allows `unValueData` to simultaneously serve two purposes, the first of which is to enforce that a `BuiltinData` is a canonical multi-asset value in data encoding, the second is to convert that `BuiltinData` to `BuiltinValue`. 

Importantly, this should allow the ex-unit cost of `unValueData` to be linear instead of quadratic.

### DIFF
--- a/CIP-0153/README.md
+++ b/CIP-0153/README.md
@@ -106,8 +106,17 @@ We propose the following set of builtin functions to accompany the new builtin t
     - encodes a `BuiltinValue` as `BuiltinData`.
 6. `unValueData :: BuiltinData -> BuiltinValue`
     - decodes a `BuiltinData` into a `BuiltinValue`, or fails if it is not one.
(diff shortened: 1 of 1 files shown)


Ryun1:
Pending a Plutus team member review, Ill be happy to give this the green tick straight away

zliu41 (approved):

Ryun1 (approved):

PR #1135, opened 2026-01-19
CIP-0172? | Self-Usable Transaction Outputs

carlostome:
Allow transactions to use scripts and data from their own outputs during validation.

([rendered proposal](https://github.com/carlostome/CIPs/blob/carlos/usable-outputs/CIP-0172/README.md))

Notes:
- I've left a TODO in the Implementation plan, since there seem to be two viable alternatives: either implement it in Conway (via a hardfork) or make it part of the upcoming Dijkstra era.


rphair:
@carlostome I'm marking this `Triage`: to introduce at the CIP meeting in about 12 hours' time (https://hackmd.io/@cip-editors/126) if we can... though I would feel strange rushing to review this without a Ledger reviewer (cc @lehins @WhatisRT) first taking a quick look to confirm this is possible: maybe also confirming that it's consistent with other Ledger design goals.

Also @colll78 @polinavino @fallen-icarus especially from your interaction with the Nested Transactions CIP I think we would benefit from your insights about this proposal: including views on use cases, efficiency, safety, and other factors that have come up before.  Personally I can't yet understand how this proposal and CIP-0118 are intertwined & I hope to get a better view of that through upcoming review.

fallen-icarus:
@carlostome Do you know of any concrete use cases for this? As a dApp developer, I've never found myself wishing for this ability.

carlostome:
> @carlostome Do you know of any concrete use cases for this? As a dApp developer, I've never found myself wishing for this ability.

Hi @fallen-icarus, [in the CIP](https://github.com/carlostome/CIPs/blob/carlos/usable-outputs/CIP-self-usable-transaction-outputs/README.md#use-cases-nested-transactions), I have outlined the use case I have in mind. That is, making sharing scripts between top- and sub- level transactions (in the context of Nested transactions) independent on the order in which they are processed within a batch. I'm happy to add more explanation and/or clarifications if you find it is not clear enough.

I do not know of other use cases at the moment.

fallen-icarus:
> ... making sharing scripts between top- and sub- level transactions (in the context of Nested transactions) independent on the order in which they are processed within a batch.

This is an abstract use case. What real world product needs this ability? Do batchers need this for aggregating sub-txs? Why wouldn't they just have the sub-txs omit the required script/data from the sub-tx so they can just add it to the top-level?

CIP-0172/README.md
```diff
@@ -0,0 +1,90 @@
...
+The current ledger logic prohibits transactions from accessing their own
+outputs during validation. However, this restriction comes at the cost of
+redundancy when the same script or datum serves dual purposes: being included
+in an output for future use and being required for current validation.
+
+This CIP achieves its goal by relaxing this restriction so that validation may
+reference scripts and data committed in the transaction body’s outputs, which
+removes the need to carry duplicate copies in witnesses and reduces transaction
+size and fees.
+
```
Ryun1, line 104: Would/could there be downsides of adding this behaviour to the Ledger?
Is there significant complexity to implement?

CIP-0172/README.md
```diff
@@ -0,0 +1,90 @@
...
+This restriction creates a specific inefficiency: if a transaction produces an
+output containing a script or datum that is also required to validate that same
+transaction, the transaction cannot utilize the copy present in the output.
+Instead, the transaction must provide a redundant copy of the same script or
+data (e.g., in the transaction witnesses).
+
+This forced redundancy increases transaction size, and thus transaction fees,
+directly inflating the costs incurred by users.
+
+The key idea of this CIP is to remove this restriction, allowing transactions
```
Ryun1, line 43: for my understanding
in other words this proposal would remove a potentially annoying 'gotcha' that developers currently need to navigate and would need to navigate with nested transactions

rphair:
summary notes from today's CIP meeting:

➤ **Use cases**: as noted above, this question also came up at the meeting.  My understanding is that this demand has already been met: when we combine @fallen-icarus's enquiry about use cases with @carlostome's response that "Nested Transactions" is _itself_ the use case, it appears that "use cases of Nested Transactions" are _also_ generally use cases of this proposal.
* @carlostome therefore, since that statement is already in your CIP document at this time, if possible please further qualify that set of use cases by confirming which _particular_ patterns of nested transactions would benefit (most) from this proposal.

➤ Thanks mainly to @colll78, we highlighted a substantial benefit for this proposal (by avoiding an "ordering dependency") in that protocol creators wouldn't have to wait for the effects of something like a token minting to be processed by the chain in a previous transaction (whose Tx would be far too big to include as another Tx input) before submitting further transactions to define that protocol.                                                                   
* @carlostome - likewise, please consider adding this scenario (waiting for large transactions to complete first, if I've understood correctly) to your Motivation if you think it's relevant.

My own list of review questions that came up are:
* Can making such a change — with any implied requirements on whether grouped transactions are "ordered" or "atomic" — break anything currently under development that would take the existing Ledger rules for granted?
* Would implementing this have any performance implications for the Ledger and/or the efficiency of regular or Nested transactions?
* @lehins @WhatisRT what about the "complexity cost" of adding this?  Even if feasible to add (as asked already in [this preliminary review](https://github.com/cardano-foundation/CIPs/pull/1135#issuecomment-3770976461), and as @colll78 also enquired, would it be worth the technical burden to accommodate this proposal?                                                                                         

Mainly pending Ledger response about feasibility we chose to leave this `Unconfirmed` until at least the feasibility & acceptable cost of a Ledger implementation can be established.

carlostome:
> What real world product needs this ability?

I don't know.

> Do batchers need this for aggregating sub-txs? 

This allows batchers to reduce duplication (and thus fees) in the case of cyclic dependencies between data and/or scripts in the top- and sub-level transactions.

> Why wouldn't they just have the sub-txs omit the required script/data from the sub-tx so they can just add it to the top-level?

Because batchers cannot modify sub-txs since they are signed. Scripts are not signed but data is (via `scriptIntegrityHash`; see the Alonzo spec).

CIP-0172/README.md
carlostome: > Would/could there be downsides of adding this behaviour to the Ledger? 

A downside could be breaking backwards compatibility of scripts. If a script's functionality depends on the script itself not being part of the outputs of a transaction then this change can break it.

> Is there significant complexity to implement?

I don't expect so, since AFAICS the only affected part of the implementation is the one in charge of collecting scripts and data from a transaction, which will have to include looking at transaction outputs.

CIP-0172/README.md
carlostome: > for my understanding in other words this proposal would remove a potentially annoying 'gotcha' that developers currently need to navigate and would need to navigate with nested transactions

That is correct. The 'gotcha' is likely not that annoying in pre-Dijkstra (since it doesn't seem anybody has raised issues before). However in Dijkstra I expect this to become a real annoyance.

carlostome:
>  @carlostome therefore, since that statement is already in your CIP document at this time, if possible you should further qualify that set of use cases by confirming which _particular_ patterns of nested transactions would benefit (most) from this proposal.

I've changed the CIP to reflect the particular nature of the use case within nested transactions.

> My own list of review questions that came up are:
 
> Can making such a change — with any implied requirements on whether grouped transactions are "ordered" or "atomic" — break anything currently under development that would take the existing Ledger rules for granted?

I'm not sure what the scope of _anything currently under development_ is, but a priori yes, in the sense that this change modifies the behaviour of some part of the Ledger rules.
 
> Would implementing this have any performance implications for the Ledger and/or the efficiency of regular or Nested transactions?

I can't foresee any implications in terms of performance and/or efficiency. However, this change will definitely simplify the implementation of data and script sharing in Nested transactions.

carlostome:
> ➤ Thanks mainly to @colll78, we highlighted a substantial benefit for this proposal (by avoiding an "ordering dependency") in that protocol creators wouldn't have to wait for the effects of something like a token minting to be processed by the chain in a previous transaction (whose Tx would be far too big to include as another Tx input) before submitting further transactions to define that protocol.
> 
>  @carlostome - likewise, please consider adding this scenario (waiting for large transactions to complete first, if I've understood correctly) to your Motivation if you think it's relevant.

I don't fully understand this scenario and how it could benefit from script sharing. In particular I don't understand the part about submitting further transactions to define the protocol after minting the token.

WhatisRT:
> what about the "complexity cost" of adding this? Even if feasible to add (as asked already in https://github.com/cardano-foundation/CIPs/pull/1135#issuecomment-3770976461, and as @colll78 also enquired, would it be worth the technical burden to accommodate this proposal?

This feature was originally motivated by a simplification of the logic of the specification. So any implementation that implements the logic in a similar manner should also see a simplification in their code. Roughly speaking, checking that everything is ordered correctly is extra work, and this CIP recognises that this extra work is unnecessary and not doing it gives some extra flexibility to DApp authors.

rphair:
> > Can making such a change... break anything currently under development that would take the existing Ledger rules for granted?

> I'm not sure what the scope of _anything currently under development_ is, but a priori yes, in the sense that this change modifies the behaviour of some part of the Ledger rules.

apologies @carlostome: I didn't mean to introduce a potentially infinite scope. 😅  What I meant more practically was (if you might know of some answers): _What common design patterns in Cardano development would depend upon the unmodified Ledger rules and therefore be affected by this CIP?_

The editors (among others) would just be trying to estimate the trade impact of making this change, and so anything you (or other reviewers) can provide here would be helpful: especially if you add these scenarios to the CIP itself.

> I don't fully understand this scenario and how it could benefit from script sharing. In particular I don't understand the part about submitting further transactions to define the protocol after minting the token.

@colll78 can you please document here this example as you described it at yesterday's CIP meeting — a typical project launch, beginning with a token mint, with this CIP avoiding waiting periods for prior transactions to be verified before continuing to each new step — which was considered an advantage and use case for this CIP?  I think we need your precision here so the author knows what to do with that suggestion & perhaps include it as a use case.

---

@carlostome please note that my writing above is mainly feedback from a CIP meeting yesterday (held biweekly) in which I put your proposal on the agenda for introduction.  We can't rely on those present at the meeting to record their meeting statements back on GitHub... so during or after the meetings one of the CIP editors will generally try to document it back here.  We do our best to record these statements but often can't perfectly reproduce the same point of view or technical understanding.  You'd be very welcome at the meeting yourself & please find the schedule here: https://luma.com/cardano-cip

lehins:
I believe it should have been implemented this way right from the start when reference scripts where introduced as a feature, because I can't see a single reason why it shouldn't have been done that way, especially since implementation for this is very simple. Luckily we should be able to retroactively enable this feature for PlutusV1-V3 scripts as well, without much effort, since witnesses are not supplied in the script context, so scripts cannot know for sure whether script was provided as a witness or as a reference script.

In my opinion, this CIP is simply an optimization. All it does is allows a user to use a script in the same transaction that adds it to the UTxO for sub-subsequent transaction to reuse as reference script. In a common case, it does not really make something that was previously impossible possible, except when a script that a person is trying to use is pretty big in size. It only makes some certain use cases cheaper, simpler for users and reduce amount of data placed on chain. This point is quite simple to prove: any transaction that could benefit out of using a script that was supplied in one of the outputs that it is creating, could always also add it to the witness set. So, all this CIP will allow is to do is accommodate those cases without including such script in a transaction twice, which is also why scripts that are so large in size that two copies of it would not fit in a transaction, could really benefit from this CIP. 

Some benefits from this CIP that come to mind:
  * From @colll78 the example of minting a token, at least as far as I imagine it, maybe @colll78 has a different use case in mind. Today in order to be able mint and burn a token, while using a reference script you require at least 3 transactions: first transaction adds a script to UTxO, while subsequent two that mint and burn use that script. With this CIP the same transaction that does the minting also could add the script to UTxO for the transaction that does the burning to use. So less fees and less contention for ephemeral tokens.
  * A similar use case as for token minting will be applicable for [CIP-160 - Protected Adddresses](https://github.com/cardano-foundation/CIPs/pull/1063). Because then the script that needs to be executed for output creation can contain that said script, so later when it is being spent there will no longer going to be a need to even specify where the script is coming from
  * Simplifies usage of reference scripts in Nested Transactions. Also by its nature Nested Transactions will have many transactions in it with a size limit of a single transaction, so anything we can do to reduce binary overhead of a transaction will be beneficial for Nested Transaction especially
  * Unlike scripts that are supplied through reference inputs these scripts provided by the output do not carry the same dangers of not being priced by the regular fee calculatoin. Therefore, those scripts will not need to participate in the same costing function as reference scripts, since they will already be priced through the size of transaction, similarly to how scripts that are used through a witnesses are priced.


Best thing I like about this CIP is that implementation of this feature is very simple and, as far as I can see it, there are only benefits and no drawbacks.

Big :+1: from me.

lehins:
> This is an abstract use case. What real world product needs this ability? Do batchers need this for aggregating sub-txs? Why wouldn't they just have the sub-txs omit the required script/data from the sub-tx so they can just add it to the top-level?

I can try and make it concrete. Here is an example without this CIP:
* Some sub-transaction creators start requiring the same script in `topLevelRequiredGuards`, while supplying said script in the witness set, or not supplying it at all. (It can be any script type, it doesn't have to be a guard script, I've used guards for the sake of example)
* Most likely aggregator (batch creator) has two priorities in mind: quickest transaction acceptance on chain and lowering the cost. Without this CIP however, aggregator can only optimize for one of those two: 
  1. include that script in a witness set to be able to use it right away and submit another transaction later (or if the script fits also create an output int the same transaction) adding the script for further reuse.
  2. create another transaction, wait for it to be accepted on chain and then start using that script.

With this CIP, top-level transaction builder will have this ability to add a script to UTxO that any of the sub-transactions might need for immediate **and** sub-subsequent use. So, IMHO, this CIP is especially useful for the top-level transaction builder.

rphair (commented):
@carlostome @lehins the elaboration of Ledger impact in https://github.com/cardano-foundation/CIPs/pull/1135#issuecomment-3786577709 abundantly addresses the only blocker we had at our last CIP meeting — uncertainty over whether this could be implemented easily enough in the Ledger — and so I'd like to promptly confirm this as a CIP candidate: as we would have if we'd been sure of this on Tuesday's meeting.

I've added this back on Review for the next CIP meeting (https://hackmd.io/@cip-editors/127), where we could do that: but if both @Ryun1 and @perturbing agree here in the meantime we'd have unanimity & can re-tag this as `Confirmed`, assign a CIP number, and start tagging other devs for a deep (technical) and broad (ecosystem) review.

Ryun1:
@rphair 
happy to have a CIP number applied 💪

colll78:
> > ➤ Thanks mainly to @colll78, we highlighted a substantial benefit for this proposal (by avoiding an "ordering dependency") in that protocol creators wouldn't have to wait for the effects of something like a token minting to be processed by the chain in a previous transaction (whose Tx would be far too big to include as another Tx input) before submitting further transactions to define that protocol.
> > @carlostome - likewise, please consider adding this scenario (waiting for large transactions to complete first, if I've understood correctly) to your Motivation if you think it's relevant.
> 
> I don't fully understand this scenario and how it could benefit from script sharing. In particular I don't understand the part about submitting further transactions to define the protocol after minting the token.

Currently to deploy just about any DeFi protocol or dApp on Cardano requires multiple transactions, even for an extremely simple dApp that only has a single script. 

The minimal flow for a stateful dApp at the moment is roughly:
Transaction 1: Deploys reference script UTxOs
Transaction 2: Include the reference script UTxOs that were produced in the previous transaction as reference inputs, and use them to mint the relevant state tokens / dynamic parameters UTxO NFT. 

With this proposal you could deploy such applications in a single transaction that produces the reference script UTxOs (ie. that contain the minting policy of the state tokens) and mint the state tokens in the same transaction (ie. create a DEX pool in the same tx where you publish the DEX pool script).

CIP-0172/README.md
rphair: @Ryun1 - part 2 of the question has been answered with the Ledger response in https://github.com/cardano-foundation/CIPs/pull/1135#issuecomment-3786577709.

@carlostome - part 1 would also be resolved if you could added that same note about "backward compatibility" — the usual place to do this in CIPs is the _Rationale_.

CIP-0172/README.md
rphair: @carlostome based on what you said, the change coming with Dijkstra — both the default restriction and how this CIP would affect it — are an important part of the _Motivation_ for this proposal and therefore should be in that section.  Can you add it there?

It should help developers to be aware how a such a restriction & its removal would affect their development plans: even after the fact, to modify systems developed before Dijkstra.

CIP-self-usable-transaction-outputs/README.md
```diff
@@ -0,0 +1,92 @@
+---
+CIP: ?
```
rphair:
```suggestion
CIP: 172
```

CIP-self-usable-transaction-outputs/README.md
```diff
@@ -0,0 +1,92 @@
...
+Created: 2026-01-15
+License: CC-BY-4.0
+---
+
+## Abstract
+
+We propose to allow transactions to reference scripts and data present on their
+own outputs during validation.
+
+## Motivation: why is this CIP necessary?
```
rphair:
```suggestion
## Motivation: Why is this CIP necessary?
```
This is going to be capitalised as of #1125 (cc @Ryun1).

CIP-self-usable-transaction-outputs/README.md
```diff
@@ -0,0 +1,92 @@
...
+There is an [Agda
+specification](https://github.com/IntersectMBO/formal-ledger-specifications/tree/carlos/usable-outputs)
+prototype of the proposed changes (for Conway).
+
+### Changes to the ledger logic
+
+- The function `txscripts` needs to be modified to include scripts from `txouts`.
+- The function `getDatum` needs to be modified to include data from `txouts`.
+
+## Rationale: how does this CIP achieve its goals?
```
rphair:
```suggestion
## Rationale: How does this CIP achieve its goals?
```
This is going to be capitalised as of #1125 (cc @Ryun1).

rphair (commented):
Since all editors who left this `Unconfirmed` at last week's CIP meeting have now supported making this a candidate, we're promoting it without waiting for the next meeting.  @carlostome please rename the containing directory to `CIP-0172` and updated the "rendered" link in your original comment. 🎉

Also please consider adding those couple of things to the _Motivation_ and _Rationale_ to resolve the comments in https://github.com/cardano-foundation/CIPs/pull/1135#discussion_r2709184101 & https://github.com/cardano-foundation/CIPs/pull/1135#discussion_r2709430844 based on @Ryun1's & other editorial feedback.

Though this proposal is of general developer interest and we would welcome comments from anyone, I'm tagging some further reviewers with a good eye for overall design patterns — especially contributors to the Nested Transactions discussion & [roundtable](https://github.com/cardano-foundation/CIPs/pull/862#issuecomment-3102872506) — to review the CIP and maybe help us plan how to introduce this to the dev community:

@perturbing @Quantumplation @polinavino @ch1bo @sandtreader @AndrewWestberg @michele-nuzzi @disassembler @MicroProofs

CIP-0172/README.md
rphair: Fixed with https://github.com/cardano-foundation/CIPs/pull/1135/commits/cd2b8d32d80273dc18055d6c5f7f76e3734367fd.

CIP-0172/README.md
rphair: Fixed with https://github.com/cardano-foundation/CIPs/pull/1135/commits/405f1a7b88f2f7a22a924e0afc97cb2a5aa8c50e.

CIP-0172/README.md
```diff
@@ -0,0 +1,120 @@
...
+
+## Abstract
+
+We propose to allow transactions to reference scripts and data present on their
+own outputs during validation.
+
+## Motivation: Why is this CIP necessary?
+
+Transaction outputs carry scripts (via reference scripts) and/or data (via
+inline datum). However, the Cardano ledger currently enforces a restriction
```
rphair:
```suggestion
inline datum). However, the Cardano Ledger currently enforces a restriction
```
Cardano readers would expect Ledger to be capitalised, though technically not required since the Cardano term itself makes it a proper noun phrase.  I think it would look more consistent with other CIPs if capitalised but grammatically I think either one will do so it's up to you.

CIP-0172/README.md
```diff
@@ -0,0 +1,120 @@
...
+
+This restriction creates a specific inefficiency: if a transaction produces an
+output containing a script or datum that is also required to validate that same
+transaction, the transaction cannot utilize the copy present in the output.
+Instead, the transaction must provide a redundant copy of the same script or
+data (e.g., in the transaction witnesses). This forced redundancy increases
+transaction size, and thus transaction fees, directly inflating the costs
+incurred by users.
+
+In combination with the upcoming feature of Nested transactions, in the
```
rphair:
```suggestion
In combination with the upcoming feature of [nested transactions](https://github.com/cardano-foundation/CIPs/blob/master/CIP-0118/README.md), in the
```
Linking the first reference to this CIP will help those who haven't encountered it before.

(not suggesting capitalising this term since this would only be capitalised when referring to the CIP, not the concept... and anyway don't want to clutter up the review with capitalisation changes which I think different authors will want to do differently)

CIP-0172/README.md
```diff
@@ -0,0 +1,120 @@
...
+mints the state token. This eliminates the need for submitting multiple
+transactions, thereby reducing costs, while at the same time improving the
+developer's experience.
+
+#### Nested transactions: order-agnostic script sharing across a batch
+
+The nested transactions CIP specifies that scripts are to be shared among all
+sub- and top-level transactions within a batch (see [CIP, Changes to
+Transaction
+Validity](https://github.com/carlostome/CIPs/tree/master/CIP-0118#changes-to-transaction-validity)).
```
rphair:
```suggestion
sub- and top-level transactions within a batch (see [CIP-0118 > Changes to Transaction
Validity](https://github.com/carlostome/CIPs/tree/master/CIP-0118#changes-to-transaction-validity)).
```
(or equivalent: as long as readers know this is the same CIP as the "nested transactions" link above)

CIP-0172/README.md
```diff
@@ -0,0 +1,120 @@
...
+sub- or top-level transaction they appear in. This eliminates the dependency on
+processing order and enables more flexible script sharing within the batch.
+
+## Specification
+
+There is an [Agda
+specification](https://github.com/IntersectMBO/formal-ledger-specifications/tree/carlos/usable-outputs)
+prototype of the proposed changes (for Conway).
+
+### Changes to the ledger logic
```
rphair:
```suggestion
### Changes to the Ledger logic
```
(this use of "Ledger" is probably important to capitalise)

CIP-0172/README.md
```diff
@@ -0,0 +1,120 @@
...
+
+## Path to Active
+
+### Acceptance Criteria
+
+- Fully implemented in Cardano node
+
+### Implementation Plan
+
+- TODO
```
rphair:
```suggestion
### Acceptance Criteria
- [ ] Fully implemented in Cardano node

### Implementation Plan
- [ ] Update the formal Ledger specification with "changes to the Ledger logic" above
- [ ] Implement the outlined changes in the Cardano node
- [ ] Complete a hard fork enabling support for the changes outlined here
```
I used https://github.com/carlostome/CIPs/tree/master/CIP-0118#implementation-plan as a template:
1. @Ryun1 @lehins can please double check for consistency with similar Ledger proposals?  
1. @carlostome & others please add anything in developer relations that could be seen as a dependency (a common issue with Ledger changes so perhaps implicit & therefore no need to publicise it?)

rphair (commented):
@carlostome thanks for all your updates in response to previous reviews which I can see have addressed all the commentary from editor's review and our last meeting discussion.  Thanks especially for your force-push being done in a way that preserves the commit history from different sources. 🙏

Editors will keep an eye out for further technical review as hopefully launched with those tagged in https://github.com/cardano-foundation/CIPs/pull/1135#pullrequestreview-3702731009, and if no objection or further evolution will keep moving it forward in the CIP review cycle towards merge.

Once the suggestions below settle, plus hopefully some confirming review, I'd be ready to approve this CIP.

CIP-0172/README.md
rphair: p.s. I can also see (somehow missed before) in the original comment the current uncertainty about implementation in Conway vs. Dijkstra... understandably this would still be `TODO` while that decision is made.

CIP-0172/README.md
```diff
@@ -0,0 +1,119 @@
...
+dApps on Cardano requires multiple transactions. The minimal workflow for such
+a stateful dApp requires (1) deploying the dApp's script in the UTxO and (2)
+referencing this script (via reference inputs) to mint the relevant state
+tokens.
+
+With the proposed changes, such applications could be deployed in a single
+transaction, which simultaneously produces the reference script UTxO's and
+mints the state token. This eliminates the need for submitting multiple
+transactions, thereby reducing costs, while at the same time improving the
+developer's experience.
```
ch1bo: Why is a reference script at all needed in this use case?

This exact annoyance could guide a dApp developer to not use reference scripts, but just include the script in the witness set of the transaction that mints the tokens. Thus, _only_ incurring a transaction fee and not require the additional deposit for the reference script in the output (which grows the utxo state, that must be paid).

CIP-0172/README.md
```diff
@@ -0,0 +1,119 @@
...
+that prevents a transaction from satisfying its validation requirements using
+scripts and data contained within its own outputs.
+
+This restriction creates a specific inefficiency: if a transaction produces an
+output containing a script or datum that is also required to validate that same
+transaction, the transaction cannot utilize the copy present in the output.
+Instead, the transaction must provide a redundant copy of the same script or
+data (e.g., in the transaction witnesses). This forced redundancy increases
+transaction size, and thus transaction fees, directly inflating the costs
+incurred by users.
```
ch1bo: I don't follow the argument of re-using inlined data.

If a transaction wants to spend an output that carries inlined data, that data would already be in the utxo set and thus the creating tx paid for it. For "normal" datum use, there would only be a datumhash on the utxo and the spending transaction provides the preimage in the witness set. This data resolves across any spent inputs that are parameterized by the same data (hash). I don't know how this evolves for nested transactions but I would expect it to allow re-use of hashed data. For other validation purposes, there is no datum in the context.

So, what exactly motivates the sketched change of `getDatum` in the agda spec?

CIP-0172/README.md
```diff
@@ -0,0 +1,119 @@
...
+
+With the current restriction, this requirement can only be partially achieved
+by sharing scripts either (1) from transaction witnesses or (2) via reference
+inputs that point to outputs of sub-transactions in the same batch which have
+already been applied to the UTxO state.
+
+With the proposed change in this CIP, the requirement can be achieved allowing
+scripts to be shared directly from transaction outputs, regardless of which
+sub- or top-level transaction they appear in. This eliminates the dependency on
+processing order and enables more flexible script sharing within the batch.
```
ch1bo, line 81: I doubt that this will make anything easier at the point of validating transactions within a node because it will be presented with a completely ordered batch and just needs to validate it.

Any code that assembles nested transactions must be aware of the UTxO graph and how (sub-)transactions are to be balanced using inputs/outputs and correctly sequence them for these reasons.

In fact, to "enable more flexible script sharing within the batch" the (nested) transaction creating code would not use reference scripts at all and instead use script hashes and include the script once in the witness set. So maybe you meant here to improve script sharing _across_ batches?

ch1bo (commented):
Gave a look because I got notified through Github and I could not really follow the motivation for this.

CIP-0172/README.md
lehins: Comparing to just including the witness set, adding a script to the UTxO will require the user to lock up the funds in the form of minUtxO, so that is the motivating factor against it. Which means the only reason why a user would want to do this instead of just using a script in witnesses is for further reuse.

So, this statement is incorrect: "not require the additional deposit for the reference script" There is still deposit for including a script in the output.

CIP-0172/README.md
lehins: Sounds like you totally missed the point of the CIP. Please read my comment and hopefully it will clarify it for you and answer the question you asking: https://github.com/cardano-foundation/CIPs/pull/1135#issuecomment-3786577709

CIP-0172/README.md
lehins: I do agree with this point, since datums are not reusable from reference inputs either. This change alone would lead to strange semantic where datum from outputs created can be reused throughout a transaction, while ones specified in inputs and reference inputs cannot. If we are to make such a change it would require its own CIP. However, I vaguely remember discussing this point long time ago with some folks that actually write DApps and their argument against was that it is very rare that a script requires the same data and it is usually fairly small anyways.
I think it would be worthwhile analyzing the chain and see whether there could actually be some space savings if we were to add such feature of inline datums being shared throughout a transaction.

CIP-0172/README.md
colll78: The point is that the script that is used to deploy the initial state tokens will of-course be reused. IE. A DEX pool script (minting policy and spending validator multi-validator). Right now you need to deploy reference scripts, and create the pool in two separate txs. 

You could just mint the tokens providing pool script as witness and then deploy reference scripts after but that's still two txs.

ch1bo: Then, it was just not clear to me as a reader that this example is not an isolated usage, but the optimization of combined publishing scripts as reference + the first usage in the same transaction. The example was helpful and I suggest it should be added to the CIP / existing examples be made more concrete.

CIP-0172/README.md
ch1bo: Thanks for pointing me there. I should have read the whole discussion before reviewing. I was somehow assuming I would see all comments in the diff anyways.

CIP-0172/README.md
rphair: @carlostome I think adding such an example would help make the proposal more "user friendly" - which has been important for other CIPs to keep people aware of best-practice design patterns.

CIP-0172/README.md
rphair: @ch1bo @lehins resolving since this matter seems cleared up.

CIP-0172/README.md
rphair: Looks resolved by https://github.com/cardano-foundation/CIPs/pull/1135/changes/1b4d41eeb6a549bdb1301d21d2d8a9a4cf976b20.

rphair (approved):
There were enough people getting enthusiastically behind this proposal shortly after it was submitted — and repeatedly confirmed — that I think it needs to ride that momentum out into whatever discussion circles would lead to its implementation & then its usage by Cardano projects & protocol developers.

@carlostome the only remaining review point https://github.com/cardano-foundation/CIPs/pull/1135#discussion_r2762388201 would be nice to have an answer — in the CIP, if possible — but not necessary since I think the general case of this feature's usability has already been established without making such a detailed study.

Generally we would not merge this without other editors' opportunity to review, which often happens in anticipation of a `Last Check` discussion so I've tagged it as such for the next CIP meeting: https://hackmd.io/@cip-editors/129

CIP-0172/README.md
```diff
@@ -0,0 +1,123 @@
...
+## Specification
+
+There is an [Agda
+specification](https://github.com/IntersectMBO/formal-ledger-specifications/tree/carlos/usable-outputs)
+prototype of the proposed changes (for Conway).
+
+### Changes to the Ledger logic
+
+- The function `txscripts` needs to be modified to include scripts from `txouts`.
+- The function `getDatum` needs to be modified to include data from `txouts`.
```
lehins: This CIP should not affect datums
```suggestion
```

CIP-0172/README.md
```diff
@@ -0,0 +1,123 @@
...
+
+## Rationale: How does this CIP achieve its goals?
+
+The current ledger logic prohibits transactions from accessing their own
+outputs during validation. However, this restriction comes at the cost of
```
lehins: The wording isn't accurate. Ledger doesn't prohibit anything and there really isn't such a concept of transaction accessing anything. Also it kinda contradicts the fact that outputs are available in plutus context
If I understand correctly what you are trying to say I would suggest instead something like this:


```suggestion
The current logic in Ledger rules does not utilize reference scripts present in the outputs in any way. This restriction comes at the cost of
```

CIP-0172/README.md
```diff
@@ -0,0 +1,123 @@
...
+redundancy when the same script or datum serves dual purposes: being included
+in an output for future use and being required for current validation.
+
+This CIP achieves its goal by relaxing this restriction so that validation may
```
lehins:
```suggestion
This CIP achieves its goal by enhancing Ledger rules so that validation may
```

CIP-0172/README.md
```diff
@@ -0,0 +1,123 @@
...
+reference scripts and data committed in the transaction body’s outputs, which
```
lehins:
```suggestion
reference scripts committed in the transaction body’s outputs, which
```

CIP-0172/README.md
```diff
@@ -0,0 +1,123 @@
...
+removes the need to carry duplicate copies in witnesses and reduces transaction
+size and fees.
+
+### Backward compatibility
+
+The changes proposed in this CIP break backward compatibility of Plutus scripts
+whose logic depends on the script being part of the outputs of the transaction
+being validated.
```
lehins: Plutus scripts are not affected by this change, since script witnesses where never translated to plutus context.

```suggestion
The changes proposed in this CIP **do not break** backward compatibility of any of the older Plutus scripts.
```

Plutus script would have been affected if we also changed semantics of inline datums, but where scripts are being supplied from will not affect execution of the scripts. That is why, for example we were able to implement CIP-110 and make reference scripts usable for PlutusV1.

CIP-0172/README.md
```diff
@@ -0,0 +1,123 @@
...
+
+## Abstract
+
+We propose to allow transactions to reference scripts and data present on their
+own outputs during validation.
+
+## Motivation: Why is this CIP necessary?
+
+Transaction outputs carry scripts (via reference scripts) and/or data (via
+inline datum). However, the Cardano Ledger currently enforces a restriction
```
lehins:
```suggestion
Transaction outputs carry scripts (via reference scripts). However, the Cardano Ledger currently enforces a restriction
```

CIP-0172/README.md
```diff
@@ -0,0 +1,123 @@
...
+that prevents a transaction from satisfying its validation requirements using
+scripts and data contained within its own outputs.
```
lehins:
```suggestion
that prevents a transaction from satisfying its validation requirements using
scripts contained within its own outputs.
```

CIP-0172/README.md
```diff
@@ -0,0 +1,123 @@
...
+
+This restriction creates a specific inefficiency: if a transaction produces an
+output containing a script or datum that is also required to validate that same
+transaction, the transaction cannot utilize the copy present in the output.
+Instead, the transaction must provide a redundant copy of the same script or
+data (e.g., in the transaction witnesses). This forced redundancy increases
```
lehins:
```suggestion
output containing a script that is also required to validate that same
transaction, the transaction cannot utilize the copy present in the output.
Instead, the transaction must provide a redundant copy of the same script in the transaction witnesses. This forced redundancy increases
```

CIP-0172/README.md
```diff
@@ -0,0 +1,123 @@
...
+transaction size, and thus transaction fees, directly inflating the costs
+incurred by users.
+
+In combination with the upcoming feature of [nested transactions](https://github.com/cardano-foundation/CIPs/blob/master/CIP-0118/README.md), in the
+Dijkstra era, the current restriction imposes a strict constraint in how
+scripts and/or data can be shared within a batch. Specifically, without this
```
lehins:
```suggestion
scripts can be shared within a batch. Specifically, without this
```

CIP-0172/README.md
```diff
@@ -0,0 +1,123 @@
...
+proposal, scripts and/or data produced in transaction outputs are only
```
lehins:
```suggestion
proposal, scripts produced in transaction outputs are only
```

CIP-0172/README.md
```diff
@@ -0,0 +1,123 @@
...
+accessible to subsequent sub-transactions (via reference inputs). This forces
+transaction builders to artificially order transactions or to provide redundant
+copies of the same script or data (e.g., if two sub-transactions mutually
```
lehins:
```suggestion
copies of the same script (e.g., if two sub-transactions mutually
```

CIP-0172/README.md
```diff
@@ -0,0 +1,123 @@
...
+depend on each other's script and/or data).
```
lehins:
```suggestion
depend on each other's script).
```

lehins (changes requested):
Aside from inclusion of inline datums in this proposal and few other minor points this CIP looks great.

CIP-0172/README.md
rphair: data / datum removed with https://github.com/cardano-foundation/CIPs/pull/1135/changes/0f406a929d1d0f43bce2d0ee64830233803d9a56 (resolving)

rphair:
@carlostome I'm taking this back from the optimistically scheduled `Last Check` since we still have `TODO` in the _Path to Active_ (requested to be updated here: https://github.com/cardano-foundation/CIPs/pull/1135#discussion_r2728769940)) and other review points that the editors would generally need to you or the contributor to mark "resolved" with at least a quick comment about how it's been fixed.

If the Implementation Plan is still `TODO` based on uncertainty about the era of implementation, we can leave it in this state while you determine that: just please comment here when the document is complete with this information (or if reviewers can help with that decision) & when it's ready for another Ledger + editor review.

kylixafonso:
## Flash Loans Must Be an Explicit Design Consideration and Acceptance Criterion

For context, I am a builder at Liqwid Labs, the team behind Liqwid Finance, the premier lending protocol on Cardano.

One of the largest gaps between Cardano DeFi and the broader DeFi ecosystem is the lack of a robust flash loan primitive. Today, Liqwid and other Cardano lending protocols are unable to support flash loans in a manner comparable to Aave. This is not because flash loans are undesirable or unsafe, but because the current ledger architecture does not provide the atomic execution guarantees and transaction composition capabilities required to implement them efficiently.

This is why I believe [CIP-0118 (Nested Transactions)](https://github.com/cardano-foundation/CIPs/tree/master/CIP-0118), and any successor proposal such as CIP-0172 if implemented first, must explicitly evaluate and provide a path toward flash loan support.

### Flash Loans Are Not a Niche Feature

Flash loans are one of the defining innovations of modern DeFi and are a foundational component of Aave, the largest lending protocol in the industry.

On Ethereum, flash loans became possible because of the account-based execution model and later received standardization through [EIP-3156](https://eips.ethereum.org/EIPS/eip-3156), followed by further improvements in [EIP-7399](https://eips.ethereum.org/EIPS/eip-7399).

Over time, flash loans have become critical infrastructure supporting:
* Uncollateralized liquidations
* Refinancing and debt migration
* Arbitrage and price alignment across venues
* Automated leverage management
* Cross-protocol composability
* MEV and searcher infrastructure
* Capital-efficient market making

These use cases improve market efficiency, lending protocol solvency, and overall ecosystem health.

### Flash Loans Are Especially Important for Lending Protocols

For lending protocols, flash loans are not merely a convenience feature.

They enable liquidators to act without maintaining large idle capital reserves. This significantly lowers barriers to participation and increases liquidation competition.

More liquidators means:
* Faster liquidations
* Better protocol health
* Reduced bad debt risk
* More resilient markets during volatility

Without flash loans, liquidators must either hold substantial capital permanently or rely on external funding sources, both of which reduce participation and increase centralization pressures.

As a lending protocol team, we view this as a material limitation to achieving parity with protocols such as Aave.

### Nested Transactions Are Potentially the Missing Primitive

CIP-0118 is the first proposal that appears capable of introducing the transaction composition primitives necessary to make flash loans feasible at the ledger level.

However, the current proposal explicitly contains a restriction that all inputs of all transactions in a batch must already exist in the UTxO set before any transaction in the batch is applied.

The rationale given is to prevent temporary duplication of assets and to avoid "falsifying access to assets via flash loans."

This design decision deserves substantially more discussion because it may unintentionally eliminate one of the most powerful and economically important use cases of transaction batching.

If nested transactions are introduced without some mechanism for:
* Block-wide transaction context
* Atomic multi-step settlement guarantees
* Bidirectional transaction chaining
* Conditional inclusion semantics
* Intra-batch state transitions

then Cardano may gain nested transactions while simultaneously preventing the class of capital-efficient strategies that make them most valuable.

### Bidirectional Chaining and Atomic Settlement Should Be Considered

A flash loan fundamentally requires:

1. Temporary access to liquidity.
2. Arbitrary actions performed with that liquidity.
3. Guaranteed repayment.
4. Atomic failure if repayment does not occur.

In a UTxO system this likely requires some form of conditional transaction inclusion where multiple transactions are only considered valid if the entire chain settles successfully.

In other words, transaction A and transaction B should not merely coexist in a block; they should be mutually dependent such that neither can be included without the other.

Whether this is ultimately achieved through nested transactions, transaction dependency graphs, block-level settlement semantics, or another mechanism is less important than ensuring the capability exists.

### Avoiding a Strategic Missed Opportunity

The purpose of nested transactions is to expand transaction composability and unlock new classes of applications.

If Cardano introduces nested transactions but intentionally excludes flash loans, we risk solving a narrower set of use cases while leaving one of the most impactful DeFi primitives unavailable.

The result would be:

* Reduced capital efficiency
* Less competitive lending markets
* More difficult liquidations
* Inferior arbitrage infrastructure
* Reduced composability between protocols
* Continued feature disparity with Ethereum DeFi

From a lending protocol perspective, that would represent a significant missed opportunity.

### Proposed Addition

At minimum, I would like to see the CIP include an explicit design goal or acceptance criterion requiring evaluation of whether the proposed architecture can support atomic flash-loan-style execution.

If the current design cannot support flash loans, the CIP should clearly identify:

* Which constraints prevent them.
* Whether those constraints are fundamental or temporary.
* What future ledger changes would be required.
* Whether alternative transaction composition mechanisms are being considered.

Flash loans are not an edge case. They are a foundational DeFi primitive that underpins some of the largest and most successful lending protocols in the industry.

A transaction composition framework that does not account for them risks limiting Cardano's long-term competitiveness and utility as a DeFi platform.

fallen-icarus:
@kylixafonso Flash loans are already possible if dApps just compose _inside the same tx_ like I'm pushing for with the DeFi Kernel philosophy. You can only spend the UTxO with the liquidity if you return it in the same tx + a fee. No ledger changes are needed; just different dApp architectures.

kylixafonso:
> @kylixafonso Flash loans are already possible if dApps just compose _inside the same tx_ like I'm pushing for with the DeFi Kernel philosophy. You can only spend the UTxO with the liquidity if you return it in the same tx + a fee. No ledger changes are needed; just different dApp architectures.

I think you're conflating two fundamentally different concepts.

What you're describing is transaction-level atomic composition: liquidity can be accessed and must be returned within the same transaction. That's certainly a useful pattern, but it's not equivalent to what most people mean by a flash loan, which derives its properties from broader state transitions and block-level execution context.

More importantly, I don't think sequencer-based/shared-state architectures and fragmented UTxO architectures should be treated as mutually exclusive or ideologically opposed design choices. Different applications have different requirements, and the ecosystem benefits from supporting both approaches.

In practice, many of the lending and liquidity protocols that have achieved meaningful scale across the industry - including the dominant lending protocols on Cardano - Liqwid - rely on pooled liquidity and shared-state semantics. Supporting those designs requires access to richer block-level context and coordination mechanisms, rather than restricting everything to strict transaction atomicity.

So the question isn't whether transaction-level composition is possible - it clearly is. The question is whether the platform should also support the broader class of architectures that many large-scale DeFi systems depend on. I would **strongly argue** that it should.

CIP-0172/README.md
rphair: @carlostome @Ryun1 @lehins I'm committing this suggestion since it fixes the only glaring reason not to merge this proposal (assuming from the time lapse that it might be abandoned & if so we might want to merge it anyway).

CIP-0172/README.md
rphair: Resolving: fixed in https://github.com/cardano-foundation/CIPs/pull/1135/changes/75424f1ae805f3306315da328b6d8a8b02e59620.

CIP-0172/README.md
rphair: Resolving: the offending sentence has been removed (not sure what commit)

CIP-0172/README.md
```diff
@@ -0,0 +1,122 @@
+---
+CIP: 172
+Title: Self-Usable Transaction Outputs
+Status: Proposed
+Category: Ledger
+Authors:
+    - Carlos Tomé Cortiñas <carlos.tome-cortinas@iohk.io>
+Implementors:
```
rphair:
```suggestion
Implementors: []
```
We will have implementors but don't know who they are yet (hence the empty set).

rphair (approved):
@carlostome @lehins (as the two principal review & response parties) although this was stalled I can see that it has passed through two rounds of review while:
* attracting some support for the pathological cases that it would solve (better explained by @colll78 above & confirming what was noted in the prior CIP meeting);
* the unique confirmation from @lehins that "it should have originally been designed to work this way" (paraphrased from memory);
* answering questions from community experts (e.g. @ch1bo) who might not originally have understood the purpose for this but appear to have all questions answered here.

The one remaining item that pulled this back from `Last Check` before was an uncertainty about which hard fork this would be included in (as documented in the OP https://github.com/cardano-foundation/CIPs/pull/1135#issue-3828855495).
* I don't see why this should be a problem since we've had many CIPs over the years which defined future changes to the Ledger but only indicated that "a hard fork" would be required without divining any hard fork names or scope in advance.
* Therefore I've installed a simple `Path to Active` that resolves that `TODO` item — giving us a _Path to Active_ consistent with the Nested Transactions work scope & timeline — as explained already in [this comment & change](https://github.com/cardano-foundation/CIPs/pull/1135#discussion_r2728769940).

I think this puts us safely back at `Last Check` but please @lehins feel free to go over this again: especially if there are any Ledger considerations that should be added to the _Path to Active_ to make it plausible.
* I would hope to merge this with no further reservations but please post if so & we'll take it off the agenda for now: https://hackmd.io/@cip-editors/141

Note I'll be submitting some commits below to correct the [validation problems](https://github.com/cardano-foundation/CIPs/actions/runs/32097845431/job/95592532936?pr=1135) but these will just be formatting & not substantial changes.

Ryun1 (approved):

PR #1140, opened 2026-01-22
CIP-0175? | Stake Pool Hot Credentials

Cerkoryn:
After brainstorming this concept with the Technical Steering Committee earlier this week I've put together this draft CIP that will allow SPOs to vote on Conway-era Governance Actions by using their Calidus Keys as defined by @Crypto2099 and @gitmachtl  in [CIP-0151](https://github.com/Crypto2099/CIPs/tree/cip-88-v2-stake-pools/CIP-0151).  

This CIP aims to extend the ledger state to index CIP-151 registrations and also complies with the CIP-151 goal of not needing to change stake pool registration certificates.

I believe @gufmar also expressed interest in this CIP during the call yesterday, so feel free to review at your convenience. 
([rendered](https://github.com/Cerkoryn/CIPs/tree/calidus-for-ledger/CIP-0175))

### DIFF
--- a/CIP-0175/README.md
+++ b/CIP-0175/README.md
@@ -1,5 +1,5 @@
 ---
-CIP: ?
+CIP: 175
 Title: SPO Governance Voting with Calidus Keys
 Category: Ledger
 Status: Proposed
--- a/CIP-0175/README.md
+++ b/CIP-0175/README.md
@@ -1,6 +1,6 @@
 ---
 CIP: 175
-Title: Stake Pool Hot Credentials 
+Title: Stake Pool Hot Credentials
 Category: Ledger
 Status: Proposed
 Authors:
@@ -17,47 +17,31 @@ License: CC-BY-4.0
 
 Stake Pool Operators (SPOs) currently sign on-chain governance votes with pool
 cold keys. This CIP introduces new on-chain certificates that let a pool cold
-key authorize a governance hot credential for a stake pool. For
+key authorize a hot credential for a stake pool. For
 `StakePoolVoter poolId`, the ledger accepts either:
 
 - the pool cold key witness, or
 - a witness that satisfies the currently authorized hot credential.
 
-Hot credentials are defined as governance `Credential` values (key hash or
-script hash), so native and Plutus script voting paths are supported from day
-one. This CIP also introduces a failsafe precedence rule where a cold-authorized
-vote supersedes a hot-authorized vote for the same pool and governance action.
+Hot credentials are defined as `Credential` values (key hash or
+script hash), so native and Plutus script voting paths are supported.
 The proposal preserves the existing `StakePoolVoter` model from CIP-1694 and
 requires a future hard fork for activation.
 
 ## Motivation: Why is this CIP necessary?
 
-CIP-1694 gives SPOs an on-chain voting role, but cold-key-only operation is
+CIP-1694 gave SPOs an on-chain voting role, but cold-key-only operation is
 high-friction and increases operational risk because cold keys are meant to
 remain offline.
 
 Authorization for consensus-critical voting must be ledger-visible and
-ledger-validated. Transaction metadata is not the right substrate for this
-authorization path. A certificate-based design provides explicit state
-transitions, deterministic validation, and consistent tooling semantics.
+ledger-validated. Existing Calidus keys from CIP-0151 rely on transaction
+metadata, but is not the right substrate for on-chain voting.
+A certificate-based design provides explicit state transitions,
+deterministic validation, and consistent tooling semantics.
 
 This CIP enables day-to-day governance operation through authorized hot
-credentials while preserving cold-key voting and adding an explicit emergency
-override model.
-
-## Goals
-
-- Add on-chain, certificate-driven authorization of SPO hot credentials.
-- Support both key and script credentials from initial deployment.
-- Preserve CIP-1694 `StakePoolVoter` semantics and cold-key voting.
-- Define deterministic precedence behavior where cold votes supersede hot votes.
-
-## Non-goals
-
-- Replacing CIP-0151.
-- Introducing new SPO voter types.
-- Defining final CBOR constructor tags in this draft (semantic specification
-  only).
+credentials while preserving cold-key voting as it already works today.
 
 ## Specification
 
@@ -80,20 +64,14 @@ This CIP introduces two new stake-pool governance certificates:
 - `AuthStakePoolHotKey(poolId, hotCred)`
 - `ResignStakePoolHotKey(poolId)`
 
-The certificates are semantic definitions in this draft. Final constructor IDs
-and concrete CBOR encoding are out of scope for this revision and are assigned
-in implementation artifacts for the activating hard fork.
-
 ### Certificate Validation Rules
 
 For both certificates:
 
 1. `poolId` MUST identify a currently registered pool.
 2. The transaction MUST include a valid witness by that pool's cold key.
-3. No deposit or refund applies.
-4. No anchor field is defined.
 
-Additional transaction restrictions:
+Additional transaction restrictions to prevent ambiguity:
 
 1. A transaction MUST NOT include more than one of these certificates for the
    same `poolId`. If it does, the transaction is invalid.
@@ -149,10 +127,7 @@ All other vote semantics (vote options, anchors, role structure, timing windows,
 and tallying model) remain as defined by CIP-1694 unless explicitly modified by
 this CIP.
 
-### Cold-over-Hot Override (Standalone Rule)
-
-This section is intentionally isolated in case future governance discussion
-chooses to revise or remove this behavior.
+### Cold Key Override
 
 For each `(govActionId, poolId)` vote slot, the ledger records both vote value
 and `VoteSource` (`Cold` or `Hot`).
@@ -185,12 +160,6 @@ This CIP integrates at existing Conway rule boundaries:
 The proposal does not require new voter types and keeps `StakePoolVoter` as the
 canonical SPO governance voter identity.
 
-### Versioning
-
-This CIP is semantically specified and does not depend on CIP-0151 for ledger
-authorization. Changes to CIP-0151 versions do not alter this CIP's consensus
-rules.
-
 ## Rationale: How does this CIP achieve its goals?
 
 - **Certificates, not metadata**: Governance authorization must be explicit
@@ -213,16 +182,15 @@ rules.
   Operators MAY reuse the same underlying key material for both systems, but it
   is not required by consensus.
 
-## Security Considerations
+### Security Considerations
 
 - **Hot credential compromise near deadline**:
-  Cold-over-hot precedence provides an emergency override path.
+  Cold key override gives SPOs a failsafe in case their hot keys are suspected
+  to be compromised.  Additionally, large stake pool groups can mitigate this
+  risk further by choosing to use a multisig hot credential.
 - **Blast radius for shared hot credentials across multiple pools**:
   This is allowed by design. Operators should consider script credentials (for
   example multisig and timelock designs) to reduce single-key compromise risk.
-- **Script policy risk**:
-  Script credentials are opt-in; policy logic quality and tooling UX remain
-  operational concerns.
 
 ## Path to Active
 
--- a/CIP-0175/README.md
+++ b/CIP-0175/README.md
@@ -17,16 +17,19 @@ License: CC-BY-4.0
 
 Stake Pool Operators (SPOs) currently sign on-chain governance votes with pool
 cold keys. This CIP introduces a new on-chain certificate that lets a pool cold
-key authorize a hot credential for a stake pool. For
-`StakePoolVoter poolId`, the ledger accepts either:
+key authorize a hot credential for a stake pool. It also makes SPO vote
+authorization explicit in the voter representation:
 
-- the pool cold key witness, or
-- a witness that satisfies the currently authorized hot credential.
+- `StakePoolVoter poolId Nothing` requires the pool cold key witness.
+- `StakePoolVoter poolId (Just hotCred)` requires witnesses that satisfy
+  `hotCred`, and the ledger verifies that `hotCred` matches the pool's
+  currently authorized hot credential.
 
 Hot credentials are defined as `Credential` values (key hash or
-script hash), so native and Plutus script voting paths are supported.
-The proposal preserves the existing `StakePoolVoter` model from CIP-1694 and
-requires a future hard fork for activation.
+script hash), so native and Plutus script voting paths are supported. For each
+`(govActionId, poolId)`, the ledger only needs to track the latest SPO vote and
+whether it was cast as `Cold` or `Hot`. The proposal requires a future hard
+fork for activation.
 
 ## Motivation: Why is this CIP necessary?
 
@@ -36,12 +39,14 @@ remain offline.
 
 Authorization for consensus-critical voting must be ledger-visible and
 ledger-validated. Existing Calidus keys from CIP-0151 rely on transaction
-metadata, but is not the right substrate for on-chain voting.
+metadata, but metadata is not the right substrate for on-chain voting.
 A certificate-based design provides explicit state transitions,
 deterministic validation, and consistent tooling semantics.
 
 This CIP enables day-to-day governance operation through authorized hot
-credentials while preserving cold-key voting as it already works today.
+credentials while preserving cold-key voting. It does so with explicit voter
+contents rather than by inferring authorization mode indirectly during witness
+validation.
 
 ## Specification
 
@@ -52,6 +57,9 @@ credentials while preserving cold-key voting as it already works today.
 - **Hot credential**: A governance `Credential` authorized for a pool's
   governance voting role. It may be either a key hash credential or script hash
   credential.
+- **SPO voter**: `StakePoolVoter !(KeyHash 'StakePool) (Maybe (Credential Hot))`.
+- **Cold vote**: An SPO vote with `StakePoolVoter poolId Nothing`.
+- **Hot vote**: An SPO vote with `StakePoolVoter poolId (Just hotCred)`.
 - **Hot authorization map**: Ledger state map
   `poolGovHotCreds : Map PoolId Credential`.
 - **Vote source**: Authorization source for a recorded SPO vote:
@@ -73,7 +81,7 @@ This CIP introduces one new stake-pool governance certificate:
 ### Certificate CDDL
 
 ```cddl
-; Numeric certificate tag allocation is TBD in Conway ledger CDDL integration.
+; Numeric certificate tag allocation is TBD in Ledger CDDL integration.
 assign_stake_pool_hot_credential_tag = uint
 
 assign_stake_pool_hot_credential_cert = [
@@ -92,11 +100,6 @@ For `AssignStakePoolHotCredential(poolId, hotCredOpt)`:
 3. If multiple `AssignStakePoolHotCredential` certificates for the same
    `poolId` appear in one transaction, they are processed in transaction order;
    the last one determines the final hot credential state for that `poolId`.
-4. A transaction MAY include SPO governance vote(s) for the same `poolId` as
-   one or more `AssignStakePoolHotCredential` certificates.
-5. Same-transaction vote authorization for `poolId` MUST evaluate against the
-   hot credential state resulting from certificates that appear earlier in that
-   transaction's certificate list.
 
 ### Ledger State Extension
 
@@ -125,39 +128,47 @@ Pool retirement lifecycle:
 - When retirement is enacted, the ledger clears that pool's
   `poolGovHotCreds` entry.
 
-### Governance Vote Authorization Change
+### SPO Voter Representation and Authorization
 
-For each vote with `Voter = StakePoolVoter poolId`, authorization succeeds if
-either of the following holds:
+This CIP changes the SPO voter representation to:
 
-1. The transaction includes a valid witness by the pool cold key for `poolId`.
-2. `poolGovHotCreds[poolId] = hotCred` exists and transaction witnesses satisfy
-   `hotCred`.
+```haskell
+StakePoolVoter !(KeyHash 'StakePool) (Maybe (Credential Hot))
+```
+
+Authorization semantics:
+
+1. `StakePoolVoter poolId Nothing` is a cold-authorized SPO vote. The
+   transaction MUST include a valid witness by the pool cold key for `poolId`.
+2. `StakePoolVoter poolId (Just hotCred)` is a hot-authorized SPO vote. The
+   transaction MUST include witnesses that satisfy `hotCred`, and the ledger
+   MUST verify that `poolGovHotCreds[poolId] = hotCred`.
 
 Hot credential satisfaction rules:
 
 - If `hotCred` is a key credential, the corresponding vkey witness is required.
 - If `hotCred` is a script credential, a valid governance voting script witness
-  is required under existing Conway voting-script rules for
-  `StakePoolVoter poolId`.
+  is required under existing governance voting-script rules for
+  `StakePoolVoter poolId (Just hotCred)`.
 - Both native scripts and Plutus scripts are supported.
 
 All other vote semantics (vote options, anchors, role structure, timing windows,
 and tallying model) remain as defined by CIP-1694 unless explicitly modified by
 this CIP.
 
-### Vote Source and Overwrite Rules
+### Vote Type and Overwrite Rules
 
-For each `(govActionId, poolId)` vote slot, the ledger records both vote value
-and `VoteSource` (`Cold` or `Hot`).
+For each active `(govActionId, poolId)` vote slot, the ledger records only the
+latest SPO vote together with its `VoteSource` (`Cold` or `Hot`). Historical
+overwritten SPO votes do not need to be retained.
 
 Overwrite rules:
 
-1. A `Cold` vote MAY overwrite any existing vote (`Cold` or `Hot`) for the same
+1. A `Cold` vote MAY replace the current vote (`Cold` or `Hot`) for the same
    `(govActionId, poolId)`.
-2. A `Hot` vote MAY overwrite an existing `Hot` vote for the same
+2. A `Hot` vote MAY replace the current `Hot` vote for the same
    `(govActionId, poolId)`.
-3. A `Hot` vote MUST NOT overwrite an existing `Cold` vote for the same
+3. A `Hot` vote MUST NOT replace the current `Cold` vote for the same
    `(govActionId, poolId)`.
 
 Effectively, once any cold-authorized vote exists for a given
@@ -166,46 +177,59 @@ Effectively, once any cold-authorized vote exists for a given
 ### Authorization-Change Vote Invalidation
 
 If `AssignStakePoolHotCredential(poolId, hotCredOpt)` results in an actual
-authorization-state change for `poolId`, the ledger MUST clear all previously
-recorded SPO votes for that `poolId` across still-active governance actions.
+authorization-state change for `poolId`, the ledger MUST clear the current SPO
+votes for that `poolId` across still-active governance actions whose recorded
+`VoteSource` is `Hot`.
 
 If `AssignStakePoolHotCredential(poolId, hotCredOpt)` does not change the
 resulting authorization state for `poolId`, no vote invalidation occurs.
 
-### Ledger Rule Integration (Conway)
+Current `Cold` votes remain valid.
 
-This CIP integrates at existing Conway rule boundaries:
+### Ledger Rule Integration
+
+This CIP integrates at existing governance ledger rule boundaries:
 
 - **`UTXO`** applies certificate-driven updates to `poolGovHotCreds` and
   enforces transaction-level constraints for this certificate type.
-- **`UTXOW`** extends SPO vote witness authorization checks to allow either
-  pool cold or authorized hot credential satisfaction.
+- **`UTXOW`** validates the explicit SPO voter form, requiring either the pool
+  cold witness for `StakePoolVoter poolId Nothing` or witnesses satisfying
+  `hotCred` plus `poolGovHotCreds[poolId] = hotCred` for
+  `StakePoolVoter poolId (Just hotCred)`.
 - **`GOV`** (vote state handling) records `VoteSource` for stake-pool votes and
-  applies cold/hot overwrite rules plus vote invalidation on actual
+  stores only the latest current vote per `(govActionId, poolId)`, applying
+  cold/hot overwrite rules plus hot-vote invalidation on actual
   authorization-state changes.
 
-The proposal does not require new voter types and keeps `StakePoolVoter` as the
-canonical SPO governance voter identity.
+The proposal does not introduce a new governance role. It extends
+`StakePoolVoter` so the SPO vote explicitly carries whether it is using cold
+authorization or a specific hot credential.
 
 ## Rationale: How does this CIP achieve its goals?
 
 - **Certificates, not metadata**: Governance authorization must be explicit
   ledger state with deterministic rule evaluation.
+- **Deterministic vote validation**: The SPO vote itself states whether it uses
+  cold authorization or a specific hot credential, and the ledger checks that
+  claim directly.
 - **Operational security**: Day-to-day activity uses hot credentials while cold
   keys remain available for high-assurance override and recovery.
-- **Compatibility**: Cold-key voting and `StakePoolVoter` semantics remain
-  intact.
+- **Compatibility**: The SPO governance role remains `StakePoolVoter`, while
+  its payload is extended to carry hot-credential information explicitly.
 - **Future-proofing**: Credential-based payload supports key and script custody
   models without a second hard-fork change.
 - **MPO support**: Explicitly permitting hot credential reuse across pools
   supports multi-pool operational workflows.
 - **Simplicity**: A single certificate with nullable payload reduces complexity
-  in specification, validation, and tooling.
+  in specification, validation, and tooling, and the ledger only needs the
+  latest vote and its type.
 
 ### Backward Compatibility
 
-- Existing cold-key SPO voting remains valid and unchanged.
-- `StakePoolVoter poolId` remains the SPO voter representation.
+- Existing cold-key SPO voting remains valid through
+  `StakePoolVoter poolId Nothing`.
+- This CIP extends the SPO voter payload to include
+  `Maybe (Credential Hot)` explicitly.
 - This CIP introduces no dependency on transaction metadata.
 - CIP-0151 remains compatible as an off-chain identity/authentication mechanism.
   Operators MAY reuse the same underlying key material for both systems, but it
@@ -215,22 +239,24 @@ canonical SPO governance voter identity.
 
 - **Hot credential compromise near deadline**:
   If hot keys are suspected to be compromised, changing or revoking the
-  authorized hot credential immediately invalidates previously recorded SPO votes
-  for that pool on still-active governance actions. Operators can then recast
-  under current authorization (including with the cold key). Additionally, large
-  stake pool groups can mitigate this risk further by choosing to use a multisig
-  hot credential.
+  authorized hot credential immediately invalidates current hot-authorized SPO
+  votes for that pool on still-active governance actions, while current
+  cold-authorized votes remain valid. Operators can then recast under current
+  authorization (including with the cold key). Additionally, large stake pool
+  groups can mitigate this risk further by choosing to use a multisig hot
+  credential.
 - **Blast radius for shared hot credentials across multiple pools**:
   This is allowed by design. Operators should consider script credentials (for
   example multisig and timelock designs) to reduce single-key compromise risk.
 - **Intentional delegation via shared or third-party hot credentials**:
   A pool cold key can authorize any `Credential`, including one controlled by a
-  different operator or reused across multiple pools. Votes witnessed by that
-  credential then count for each `StakePoolVoter poolId` that has authorized it.
-  This representative behavior is an intentional capability, not a validation
-  flaw, but it can concentrate SPO voting influence. If that outcome is
-  undesirable, pool operators can revoke or replace the hot credential and, for
-  the current governance action, reassert intent with a cold-authorized vote.
+  different operator or reused across multiple pools. Votes cast as
+  `StakePoolVoter poolId (Just hotCred)` then count for each pool that has
+  authorized that `hotCred`. This representative behavior is an intentional
+  capability, not a validation flaw, but it can concentrate SPO voting
+  influence. If that outcome is undesirable, pool operators can revoke or
+  replace the hot credential and, for the current governance action, reassert
+  intent with a cold-authorized vote.
 
 ## Path to Active
 
@@ -239,10 +265,11 @@ canonical SPO governance voter identity.
 - [ ] Ledger implementation merged in at least one node client.
 - [ ] Ledger implementation includes:
       `AssignStakePoolHotCredential`,
-      `poolGovHotCreds` state management, SPO vote authorization extension, and
-      cold/hot vote overwrite and authorization-change invalidation behavior.
+      `poolGovHotCreds` state management, the explicit SPO voter
+      representation, latest-vote-plus-type tracking, and cold/hot overwrite
+      plus authorization-change invalidation behavior.
 - [ ] Compatible tooling available to create/submit new certificates and submit
-      SPO votes using key or script hot credentials.
+      SPO votes using either cold authorization or key/script hot credentials.
 - [ ] Integrated in a hard fork release.
 - [ ] Implementation present within block producing nodes used by 80%+ of stake.
 
@@ -250,11 +277,11 @@ canonical SPO governance voter identity.
 
 - Add certificate constructors and semantic validation for pool hot
   credential assignment/revocation via a nullable payload.
-- Add and maintain `poolGovHotCreds` in Conway ledger state.
-- Extend SPO vote witness verification for authorized hot credentials
-  (key and script forms).
-- Implement vote-source tracking, cold/hot overwrite behavior, and
-  authorization-change vote invalidation.
+- Add and maintain `poolGovHotCreds` in ledger state.
+- Extend the SPO voter representation to carry `Maybe (Credential Hot)` and
+  validate explicit cold/hot vote authorization.
+- Implement latest-vote-plus-type tracking, cold/hot overwrite behavior, and
+  hot-vote invalidation on authorization-state changes.
 - Update tooling and documentation for certificate flows and hot credential
   voting.
 - Deploy in a future hard fork.
--- a/CIP-0175/README.md
+++ b/CIP-0175/README.md
@@ -146,24 +146,31 @@ All other vote semantics (vote options, anchors, role structure, timing windows,
 and tallying model) remain as defined by CIP-1694 unless explicitly modified by
 this CIP.
 
-### Cold Key Override
+### Vote Source and Overwrite Rules
 
 For each `(govActionId, poolId)` vote slot, the ledger records both vote value
 and `VoteSource` (`Cold` or `Hot`).
 
 Overwrite rules:
 
-1. A `Cold` vote overwrites any existing `Hot` vote for the same
+1. A `Cold` vote MAY overwrite any existing vote (`Cold` or `Hot`) for the same
    `(govActionId, poolId)`.
-2. A `Hot` vote MUST NOT overwrite an existing `Cold` vote for the same
+2. A `Hot` vote MAY overwrite an existing `Hot` vote for the same
    `(govActionId, poolId)`.
-3. A later `Cold` vote MAY overwrite an earlier `Cold` vote.
-4. A later `Hot` vote MAY overwrite an earlier `Hot` vote, unless a `Cold` vote
-   has already been recorded for that pair.
+3. A `Hot` vote MUST NOT overwrite an existing `Cold` vote for the same
+   `(govActionId, poolId)`.
 
 Effectively, once any cold-authorized vote exists for a given
 `(govActionId, poolId)`, that pair is locked against hot-authorized overwrites.
-This applies regardless of transaction or block ordering history.
+
+### Authorization-Change Vote Invalidation
+
+If `AssignStakePoolHotCredential(poolId, hotCredOpt)` results in an actual
+authorization-state change for `poolId`, the ledger MUST clear all previously
+recorded SPO votes for that `poolId` across still-active governance actions.
+
+If `AssignStakePoolHotCredential(poolId, hotCredOpt)` does not change the
+resulting authorization state for `poolId`, no vote invalidation occurs.
 
 ### Ledger Rule Integration (Conway)
 
@@ -174,7 +181,8 @@ This CIP integrates at existing Conway rule boundaries:
 - **`UTXOW`** extends SPO vote witness authorization checks to allow either
   pool cold or authorized hot credential satisfaction.
 - **`GOV`** (vote state handling) records `VoteSource` for stake-pool votes and
-  applies the cold-over-hot overwrite rules.
+  applies cold/hot overwrite rules plus vote invalidation on actual
+  authorization-state changes.
 
 The proposal does not require new voter types and keeps `StakePoolVoter` as the
 canonical SPO governance voter identity.
@@ -206,12 +214,23 @@ canonical SPO governance voter identity.
 ### Security Considerations
 
 - **Hot credential compromise near deadline**:
-  Cold key override gives SPOs a failsafe in case their hot keys are suspected
-  to be compromised.  Additionally, large stake pool groups can mitigate this
-  risk further by choosing to use a multisig hot credential.
+  If hot keys are suspected to be compromised, changing or revoking the
+  authorized hot credential immediately invalidates previously recorded SPO votes
+  for that pool on still-active governance actions. Operators can then recast
+  under current authorization (including with the cold key). Additionally, large
+  stake pool groups can mitigate this risk further by choosing to use a multisig
+  hot credential.
 - **Blast radius for shared hot credentials across multiple pools**:
   This is allowed by design. Operators should consider script credentials (for
   example multisig and timelock designs) to reduce single-key compromise risk.
+- **Intentional delegation via shared or third-party hot credentials**:
+  A pool cold key can authorize any `Credential`, including one controlled by a
+  different operator or reused across multiple pools. Votes witnessed by that
+  credential then count for each `StakePoolVoter poolId` that has authorized it.
+  This representative behavior is an intentional capability, not a validation
+  flaw, but it can concentrate SPO voting influence. If that outcome is
+  undesirable, pool operators can revoke or replace the hot credential and, for
+  the current governance action, reassert intent with a cold-authorized vote.
 
 ## Path to Active
 
@@ -221,7 +240,7 @@ canonical SPO governance voter identity.
 - [ ] Ledger implementation includes:
       `AssignStakePoolHotCredential`,
       `poolGovHotCreds` state management, SPO vote authorization extension, and
-      cold-over-hot precedence behavior.
+      cold/hot vote overwrite and authorization-change invalidation behavior.
 - [ ] Compatible tooling available to create/submit new certificates and submit
       SPO votes using key or script hot credentials.
 - [ ] Integrated in a hard fork release.
@@ -234,7 +253,8 @@ canonical SPO governance voter identity.
 - Add and maintain `poolGovHotCreds` in Conway ledger state.
 - Extend SPO vote witness verification for authorized hot credentials
   (key and script forms).
-- Implement vote-source tracking and cold-over-hot overwrite behavior.
+- Implement vote-source tracking, cold/hot overwrite behavior, and
+  authorization-change vote invalidation.
 - Update tooling and documentation for certificate flows and hot credential
   voting.
 - Deploy in a future hard fork.
--- a/CIP-0175/README.md
+++ b/CIP-0175/README.md
@@ -17,7 +17,7 @@ License: CC-BY-4.0
 
 Stake Pool Operators (SPOs) must currently sign on-chain governance votes with
 pool cold keys. This CIP enables an alternate, authorized "hot" key for SPO
-voting by reusing the Calidus key registration defined in CIP-0151. The ledger
+voting by reusing the Calidus key registration defined in [CIP-0151 | On-Chain Registration - Stake Pools](https://github.com/cardano-foundation/CIPs/pull/999). The ledger
 will accept SPO votes signed by either the pool cold key (status quo) or the
 currently active Calidus key for that pool, as determined by the highest-nonce
 CIP-0151 registration metadata. No new certificate type is introduced. Instead,
--- a/CIP-0175/README.md
+++ b/CIP-0175/README.md
@@ -16,7 +16,7 @@ License: CC-BY-4.0
 ## Abstract
 
 Stake Pool Operators (SPOs) currently sign on-chain governance votes with pool
-cold keys. This CIP introduces new on-chain certificates that let a pool cold
+cold keys. This CIP introduces a new on-chain certificate that lets a pool cold
 key authorize a hot credential for a stake pool. For
 `StakePoolVoter poolId`, the ledger accepts either:
 
@@ -59,28 +59,44 @@ credentials while preserving cold-key voting as it already works today.
 
 ### Certificate Semantics
 
-This CIP introduces two new stake-pool governance certificates:
+This CIP introduces one new stake-pool governance certificate:
 
-- `AuthStakePoolHotKey(poolId, hotCred)`
-- `ResignStakePoolHotKey(poolId)`
+- `AssignStakePoolHotCredential(poolId, hotCredOpt)`
+
+`hotCredOpt` semantics:
+
+- If `hotCredOpt = hotCred` (a `Credential`), the pool's hot credential is
+  set to `hotCred` (overwrite allowed).
+- If `hotCredOpt = null`, any existing hot credential for `poolId` is revoked.
+  If none exists, this is a valid no-op.
+
+### Certificate CDDL
+
+```cddl
+; Numeric certificate tag allocation is TBD in Conway ledger CDDL integration.
+assign_stake_pool_hot_credential_tag = uint
+
+assign_stake_pool_hot_credential_cert = [
+  assign_stake_pool_hot_credential_tag,
+  pool_id : pool_keyhash,
+  hot_credential : null / credential
+]
+```
 
 ### Certificate Validation Rules
 
-For both certificates:
+For `AssignStakePoolHotCredential(poolId, hotCredOpt)`:
 
 1. `poolId` MUST identify a currently registered pool.
 2. The transaction MUST include a valid witness by that pool's cold key.
-
-Additional transaction restrictions to prevent ambiguity:
-
-1. A transaction MUST NOT include more than one of these certificates for the
-   same `poolId`. If it does, the transaction is invalid.
-2. A transaction MUST NOT include any SPO governance vote for `poolId` if it
-   also includes one of these certificates for `poolId`. If it does, the
-   transaction is invalid.
-
-`ResignStakePoolHotKey(poolId)` is valid even if no hot credential is currently
-authorized for `poolId`; it is a no-op in that case.
+3. If multiple `AssignStakePoolHotCredential` certificates for the same
+   `poolId` appear in one transaction, they are processed in transaction order;
+   the last one determines the final hot credential state for that `poolId`.
+4. A transaction MAY include SPO governance vote(s) for the same `poolId` as
+   one or more `AssignStakePoolHotCredential` certificates.
+5. Same-transaction vote authorization for `poolId` MUST evaluate against the
+   hot credential state resulting from certificates that appear earlier in that
+   transaction's certificate list.
 
 ### Ledger State Extension
 
@@ -92,9 +108,12 @@ poolGovHotCreds : Map PoolId Credential
 
 State transitions:
 
-- `AuthStakePoolHotKey(poolId, hotCred)` sets
-  `poolGovHotCreds[poolId] = hotCred` (overwrite allowed).
-- `ResignStakePoolHotKey(poolId)` removes `poolId` from the map if present.
+- `AssignStakePoolHotCredential(poolId, hotCredOpt)` with
+  `hotCredOpt = hotCred` sets `poolGovHotCreds[poolId] = hotCred`
+  (overwrite allowed).
+- `AssignStakePoolHotCredential(poolId, hotCredOpt)` with
+  `hotCredOpt = null` removes `poolId` from the map if present (otherwise
+  no-op).
 
 No uniqueness constraint is imposed on `hotCred`. The same hot credential MAY be
 authorized for multiple pools.
@@ -151,7 +170,7 @@ This applies regardless of transaction or block ordering history.
 This CIP integrates at existing Conway rule boundaries:
 
 - **`UTXO`** applies certificate-driven updates to `poolGovHotCreds` and
-  enforces transaction-level constraints for these certificates.
+  enforces transaction-level constraints for this certificate type.
 - **`UTXOW`** extends SPO vote witness authorization checks to allow either
   pool cold or authorized hot credential satisfaction.
 - **`GOV`** (vote state handling) records `VoteSource` for stake-pool votes and
@@ -172,6 +191,8 @@ canonical SPO governance voter identity.
   models without a second hard-fork change.
 - **MPO support**: Explicitly permitting hot credential reuse across pools
   supports multi-pool operational workflows.
+- **Simplicity**: A single certificate with nullable payload reduces complexity
+  in specification, validation, and tooling.
 
 ### Backward Compatibility
 
@@ -198,7 +219,7 @@ canonical SPO governance voter identity.
 
 - [ ] Ledger implementation merged in at least one node client.
 - [ ] Ledger implementation includes:
-      `AuthStakePoolHotKey`/`ResignStakePoolHotKey`,
+      `AssignStakePoolHotCredential`,
       `poolGovHotCreds` state management, SPO vote authorization extension, and
       cold-over-hot precedence behavior.
 - [ ] Compatible tooling available to create/submit new certificates and submit
@@ -209,7 +230,7 @@ canonical SPO governance voter identity.
 ### Implementation Plan
 
 - Add certificate constructors and semantic validation for pool hot
-  authorization/resignation.
+  credential assignment/revocation via a nullable payload.
 - Add and maintain `poolGovHotCreds` in Conway ledger state.
 - Extend SPO vote witness verification for authorized hot credentials
   (key and script forms).
--- a/CIP-0175/README.md
+++ b/CIP-0175/README.md
@@ -1,44 +1,63 @@
 ---
 CIP: 175
-Title: SPO Governance Voting with Calidus Keys
+Title: Stake Pool Hot Credentials 
 Category: Ledger
 Status: Proposed
 Authors:
     - Ryan Wiley <rian222@gmail.com>
 Implementors: []
 Discussions:
-    - CIP-0151 original pull request: https://github.com/cardano-foundation/CIPs/pull/999
-    - add SPO governance voting: https://github.com/cardano-foundation/CIPs/pull/1140
+    - https://github.com/cardano-foundation/CIPs/pull/999
+    - https://github.com/cardano-foundation/CIPs/pull/1140
 Created: 2026-01-22
 License: CC-BY-4.0
 ---
 
 ## Abstract
 
-Stake Pool Operators (SPOs) must currently sign on-chain governance votes with
-pool cold keys. This CIP enables an alternate, authorized "hot" key for SPO
-voting by reusing the Calidus key registration defined in [CIP-0151 | On-Chain Registration - Stake Pools](https://github.com/cardano-foundation/CIPs/pull/999). The ledger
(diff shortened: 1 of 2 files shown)


CIP-0175/README.md
```diff
@@ -0,0 +1,296 @@
...
+> Validation Method 1 (Beacon/Reference Token) is intentionally excluded from
+> on-chain voting authorization in this CIP. It relies on an additional token
+> reference and policy constraints that are not currently validated by the
+> ledger for stake pools. Supporting Method 1 would require new on-chain
+> constraints tying the beacon policy to the pool cold key, which is outside
+> the scope of this proposal.
+
+#### Signature Payload Derivation (Strict)
+
+Let `payload` be the CIP-0151 Registration Payload object (the map at key `1`
```
Ryun1: Maybe silly question
Can the Ledger read and interpret transactions metadata?

alternatively to transaction metadata
adding a `register_calidus` certificate?
similar to how Constitutional Committee Hot Credentials work
and then the same logic could be reused

although it wouldn't be aligned with current Calidus implementations

Cerkoryn: Good question.  TBH I'm not sure if the Ledger can read and interpret Tx metadata.  I had to learn a lot of new things while writing this CIP and that never crossed my mind 😅 

I'm not aware of how CC hot credentials work, but that's probably an alternative implementation that should be explored.

Ryun1:
@Cerkoryn 

Thank you for this proposal

I have marked it as triage for the next CIP editors call 💪
Definitely a cool one I look forward to it being developed further

CIP-calidus-for-ledger/README.md
```diff
@@ -0,0 +1,296 @@
+---
+CIP: ?
+Title: Enable SPO Governance Voting with Calidus Keys
```
rphair:
```suggestion
Title: SPO Governance Voting with Calidus Keys
```
Word `Enable` will generally be redundant in CIP titles: since they are generally "enabling" the thing that they are proposing.  (Also updating PR title in sync but please @Cerkoryn feel free to adjust them as needed further.)

CIP-calidus-for-ledger/README.md
```diff
@@ -0,0 +1,296 @@
...
+Category: Ledger
+Status: Proposed
+Authors:
+    - Ryan Wiley <rian222@gmail.com>
+Implementors: []
+Discussions:
+    - https://github.com/cardano-foundation/CIPs/pull/999
+    - https://github.com/cardano-foundation/CIPs/pull/1115
```
rphair:
```suggestion
Discussions:
    - CIP-0151 original PR: https://github.com/cardano-foundation/CIPs/pull/999
    - add SPO governance voting: https://github.com/cardano-foundation/CIPs/pull/1140
```
1. #1115 is a "metadata" update to reflect implementation status and providers: neither those changes nor that PR discussion say anything about CIP-0151 itself (nor its extensibility).
2. We do need to add `this` PR into the discussion, though.

@Ryun1 (with some help from me) is in the process of enabling descriptive keywords for the YAML lists and I think the multiplicity here would be a good use for them.  I've marked it up according to my understanding of the new "rules" in #1125 (soon to be added to CIP-0001 itself) and hopefully he will correct me if the format is wrong or non-ideal.
  * **NOTE** the key for the added line for `this` PR (`SPO governance voting`) might have its CIP number added (before merge: e.g. `CIP-01XX: add SPO governance voting`)
  * @Ryun1 @perturbing if this seems awkward during the editing process (a similar possibility to what I described in https://github.com/cardano-foundation/CIPs/pull/1128#issuecomment-3775662590) then maybe the editorial recommendation should be different, e.g. some "placeholder"... what do you think?)

rphair (commented):
Agreed with @Ryun1: this is a natural step forward into maturity for the overall governance platform.  If I recall correctly I think there's already been talk in the SPO & gov communities about how beneficial this would be.

CIP-calidus-for-ledger/README.md
rphair: FYI @Cerkoryn (cc @Ryun1) we're also considering abbreviating `pull request` to `PR` in these headers: stay tuned... https://github.com/cardano-foundation/CIPs/pull/1128#issuecomment-3787095779

CIP-0175/README.md
ch1bo: Yeah I wondered the same thing and I suspect it's either not possible or going to be VERY annoying. IMO the path of adding a new `PoolCert` would probably be the least resistance, for the pool-related ledger state would also be available in that "POOL" ledger rule.

However, that would mean that we cannot re-use the Calidus key registration, which is purely metadata based?

Not sure if this defeats the purpose of this proposal or maybe the actual motivation is to use "some hot key" to do SPO voting (and not strictly Calidus keys)?

ch1bo: BTW, what had me have a closer look to this CIP was a big similarity between this key registration and usage with the mechanisms that we are going to need for [Peras CIP-140](https://github.com/cardano-foundation/CIPs/blob/master/CIP-0140/README.md) and [Leios CIP-164](https://github.com/cardano-foundation/CIPs/blob/master/CIP-0164/README.md) (both will be registering BLS keys, but those we could use easily for authorizing SPO votes). CC @perturbing

CIP-0175/README.md
```diff
@@ -0,0 +1,296 @@
...
+under metadata label `867`), encoded as CBOR with map keys in **ascending
+numeric order**.
+
+Define:
+
+- `payload_cbor` = CBOR encoding of `payload` (byte string)
+- `payload_hex` = ASCII byte string of lowercase hex digits encoding
+  `payload_cbor`, with **no prefix**
+- `sig_payload` = `blake2b-256(payload_hex)`
```
ch1bo: This quite a funky signature scheme.. why not sign the original bytes (`payload_cbor`) directly?

Going the indirection via hex encoding, ascii encoding, and then ascii decoding (we can only hash bytes, not text) seems very odd.

CIP-0175/README.md
```diff
@@ -0,0 +1,296 @@
...
+## Rationale: how does this CIP achieve its goals?
+
+- **Security**: Allows routine voting with a hot key while keeping cold keys
+  offline, reducing exposure and operational risk.
+- **Compatibility**: Retains cold-key voting as a fallback and does not alter
+  the on-chain vote structure defined in CIP-1694.
+- **Reuse of existing standards**: Leverages CIP-0151 metadata to avoid
+  introducing a new certificate type.
+- **Operational simplicity**: Key rotation and revocation follow the established
+  CIP-0151 nonce and zero-key semantics.
```
ch1bo: I appreciate these two bullets about re-using standards and keys we already have. However, CIP-151 seems to be a standard that is foreign to the ledger and now we aim to introduce it to it. The other comment thread by @Ryun1 may already have identified a challenge with this.

For the SPO user-journey though, this does not matter and we could register the same key in two ways. Although the "scanning metadata" way of CIP-151 could be deprecated if the ledger will keep track of it and expose that part of the ledger state.

ch1bo (commented):
Very interesting proposal and potential synergy with other CIPs that specify new key registration mechanisms.

gitmachtl:
So calidus keys are ED25519_BIP32 based and were created to have a method to do pool owner votings offchain. Its not limited to votings, calidus keys can also be used for logins, sign. As they can be imported into light wallets via the 24 word mnemonics they can easily be used for all kind of online actions. Eternl-Wallet, Typhon and so have already implemented message-signing for calidus1xxx ids. This all is already working and in a few days also used on ekklesia for SPO voting.

Afaik, its not possible for the ledger to read metadata. But, if we can introduce a new Certificate - which is not a PoolRegistration one, SPOs are afraid of those - that would be a nice addition to the system. Calidus keys are hot keys and in there nature designed to be rotated at any point. They can also be revoked. So a new Certificate to Register and Deregister a CalidusKey - which should also be an extended ed25519 one like used right now - that is signed by the pool cold key would be the logical next step imo.

For governance on the ledger level we already have a cold/hot system with the cc keys. So, if the ledger can hold this new hot key data and depends the voting results on those instead of the cold keys, that would be a nice shift in the future.

There is no need to rush it imo, we can have a couple of spo votes on ekklesia before moving forward with this on the ledger level? I am not opposed to do it right now for sure.

CIP-0175/README.md
Cerkoryn: @ch1bo the thought of having a new `PoolCert` occurred to me as well, but the existing Calidus keys rely on token metadata and avoid making new certificates or changes to existing ones.  It may be better to create a new certificate, but my understanding is that it would basically be a new Calidus/hot key and would make the old ones obsolete.

I'm open to either option, but it's probably worth further discussion.  Also it sounds like if we will need new keys for Leios/Peras then maybe there's some opportunity to kill multiple birds with one stone.

CIP-0175/README.md
Cerkoryn: This was intended to match CIP-0151 v2's signing payload definition, which I believe was done this way to help with hardware wallet support.

CIP-calidus-for-ledger/README.md
```diff
@@ -0,0 +1,296 @@
...
+currently active Calidus key for that pool, as determined by the highest-nonce
+CIP-0151 registration metadata. No new certificate type is introduced. Instead,
+ledger state is extended to index validated CIP-0151 registrations (scope: stake
+pool), and the governance witness checks are updated to admit the Calidus key as
+an additional authorized signer. This change reduces operational risk by keeping
+cold keys offline while preserving backward compatibility and the existing vote
+format defined in CIP-1694. The proposal is ledger-only and requires a future
+hard fork for activation.
+
+## Motivation: why is this CIP necessary?
```
rphair:
```suggestion
## Motivation: Why is this CIP necessary?
```
(pending repo-wide corrections in #1125)

CIP-calidus-for-ledger/README.md
```diff
@@ -0,0 +1,296 @@
...
+All other vote semantics (vote format, anchors, role definitions, and tallying)
+remain unchanged.
+
+### Versioning
+
+This CIP depends on CIP-0151 version 2 and ignores registrations with other
+versions for ledger authorization. If CIP-0151 is revised with a new version,
+this CIP must be updated or superseded to recognize it.
+
+## Rationale: how does this CIP achieve its goals?
```
rphair:
```suggestion
## Rationale: How does this CIP achieve its goals?
```
(pending repo-wide corrections in #1125)

rphair (commented):
@Cerkoryn - our goal of continuing improvement of Cardano's governance system was a key factor in accepting this as a CIP candidate at the meeting today.  As you also observed there, it will be most important to get early & comprehensive review from `Ledger` architects to look over all the implications of adding these keys much more intimately than they've been used so far.

@lehins @WhatisRT I hope haven't jumped the gun in declaring this a CIP candidate without going through the customary process of pinging you for a quick confirming review first.  There were some other subject experts at the meeting today who agreed that were would surely be _some_ way of doing this even if the spec needs to be fine-tuned or even the whole approach changed... looking forward to your reviews in any case.

@Cerkoryn please rename the containing directory to `CIP-0175` and update the "rendered" link to your proposal accordingly 🎉

WhatisRT:
I never made the connection that the keys inside the `StakePoolParams` are actually the cold keys, even though I should have. So I think this is actually a pretty big design flaw in Conway, SPOs shouldn't have to dig out their cold keys from some undisclosed location in the woods just to vote!

That being said, while I generally like community solutions to problems like this, and while the Ledger could technically look inside the metadata, for a core Ledger feature like this we shouldn't be (ab)using metadata. A certificate works perfectly fine and is relatively simple to add.

gufmar:
> SPOs shouldn't have to dig out their cold keys

from what I can remember this was discussed - or at least requested by SPOs - several times, before it became a "cold shower" thing. 
I think @dcoutts had some arguments why the existing hot key was not considered as an option.

gitmachtl:
I think that idea came late in the design phase, and we were running out of release time back in the days.

gitmachtl:
> A certificate works perfectly fine and is relatively simple to add.

Yea, thats the way to go i think. Ledger decision are not based on metadata content. So we need a new registration/deregistration certificate for calidus spo keys. The thing is, with the current metadata solution, its also possible with pool cold keys on hardware wallets to generate the needed CIP8/30 signature and cose key. Such a new certificate would require to also add those functions into cardano-hw-cli and implementing it in ledger&co firmware too.
on the other hand, the governance action stuff for pool must also be changed to work with "hot" keys instead of cold keys.

rphair:
This came up _ad hoc_ at the CIP meeting today and the number of possibilities still on the table inclined @Cerkoryn to confirm that he will be doing some research & polling to choose which approach to take: including today's additional suggestion of using the BLS keys that must be created by SPOs to support Peras / Leios and will be added natively to the chain soon (cc @ch1bo).

Once the desired approach is selected — also considering what @WhatisRT has suggested above — we would then go forward by asking Ledger to re-review.

gitmachtl:
I would highly recommend to ask @lehins before doing any further ideas on how it could be implemented. 

BLS keys are not compatible with LightWallet Message-Signing on Cardano. And that was one of the major reasons the currently used Calidus Keys are based on normal ed25519 ones. So all standard tooling can work with those SPO hot keys.

Would be a bit strange if we have to run two different SPO hot keys at the same time imo.

We need a fast solution, not a complicated one. And nobody knows how long it will take until we see something about Peras/Leios really in the ledger.

CIP-0175/README.md
```diff
@@ -0,0 +1,296 @@
...
+boundaries:
+
+- **ConwayLEDGER** (module `Cardano.Ledger.Conway.Rules.Ledger`): unchanged
+  sequencing, but the subordinate `UTXOW`/`UTXO` behavior is extended as
+  described below.
+- **UTXOW** (rule `"UTXOW"` in `Cardano.Ledger.Conway.Rules`): extends witness
+  verification for SPO votes by expanding the required key hash set.
+- **UTXO** (rule `"UTXO"` in `Cardano.Ledger.Conway.Rules`): updates the
+  `calidusKeys` map from validated CIP-0151 metadata found in transaction
+  auxiliary data.
```
lehins: Metadata is called this way for a reason: Ledger will never look at its contents, so this approach is out of the question.

lehins (commented):
Please come to next Ledger Working Group and we'll be happy to have a proper discussion about this CIP: https://luma.com/xju4zn8t

CIP-0175/README.md
```diff
@@ -0,0 +1,296 @@
...
+    - add SPO governance voting: https://github.com/cardano-foundation/CIPs/pull/1140
+Created: 2026-01-22
+License: CC-BY-4.0
+---
+
+## Abstract
+
+Stake Pool Operators (SPOs) must currently sign on-chain governance votes with
+pool cold keys. This CIP enables an alternate, authorized "hot" key for SPO
+voting by reusing the Calidus key registration defined in CIP-0151. The ledger
```
lehins: It would be useful to provide a link and a title for the CIP-151, at least at its first occurrence
```suggestion
voting by reusing the Calidus key registration defined in [CIP-0151 | On-Chain Registration - Stake Pools](https://github.com/cardano-foundation/CIPs/pull/999). The ledger
```

CIP-0175/README.md
```diff
@@ -0,0 +1,296 @@
...
+preserving the ability to vote with cold keys as a fallback.
+
+## Specification
+
+### Terminology
+
+- **Pool cold key**: The Ed25519 key whose hash defines the Pool ID.
+- **Pool ID**: The blake2b-224 hash of the pool cold verification key.
+- **Calidus key**: The Ed25519 public key registered in CIP-0151 metadata for a
+  stake pool (registration payload field 7).
```
lehins: As mentioned in my other comment, Ledger will never inspect anything in the metadata, since that would introduce huge computational burden and open us up to DDoS attacks. Also this goes against the whole philosophy of why we chose Haskell in the first place, this solution untyped, while the whole point of using Haskell for Ledger implementation is to provide type safety.

One more problem with this is that it would make Ledger rules depend on a standard that is not enforced by the Ledger itself, which is totally unacceptable IMHO.

CIP-0175/README.md
```diff
@@ -0,0 +1,296 @@
...
+The optional Calidus Key (7) is used as the candidate key to authorize SPO
+voting.
+
+### Validation of CIP-0151 Registrations (Ledger Rule Additions)
+
+To prevent unauthorized key claims, the ledger MUST validate CIP-0151
+registrations before they can influence voting authorization. A registration is
+**valid** for ledger purposes when all of the following hold:
+
+1. The metadata conforms to CIP-0151 version 2 and scope = stake pool.
```
lehins: This point by itself means that Ledger must implement the whole of CIP-151 in Ledger! So, before we go with this CIP that simply uses Calidus keys we should create a CIP about how we would CIP-151 would look implemented in Ledger and whether it is feasible and desireable. One thing for sure it will not be done through parsing metadata.

gitmachtl:
Thank you for the Ledger-WG-Meeting today, here are a few points/toughts of mine:

The reason a new additional certificate is more preferable imo:
- Talking with SPOs, they are way more comfortable to submit a "standalone" hot key certificate, rather than be forced to do a pool-update. Pool-Updates are always a bit "scary", and many SPOs have done it only once, maybe twice.
- The registration is instant and not "delayed" like normal pool-updates.

About the certificate itself:
- The only things that are needed in there are the pool-id and the hash of the new hot-public key right? To make this more usable for other tooling and verification. Can we enforce that such a certificate must not only be signed by the pool cold key, but also with the new hot key as a witness? Therefore the public key would be included in the transaction and could be used in db-sync and co later on. EDIT: Had a little chat with Alexey about that and its hard to "enforce" things if not really needed by the ledger. So, i guess that can only be resolved by adding the witness anyway via tooling, or to attached the already existing calidus registration metadata structure to the transaction and query the public key "somewhere" else.
- Do we need something like an expire interval in there?

About the key generation:
- As i stated above, we should use a standard ed25519 key for this. "Easy" to implement in current ledger structure, and existing tooling can be used right away. This keeps the option to generate them via mnemonics on various platforms.
- Not a "must have", but introducing such keys officially to the ledger gives us the opportunity to define a derivation path in CIP5 for it too. The "next in line" would f.e. be `1852H/1815H/<acc>H/6/<idx>`. @Ryun1 is now used to do those updates 😄 And the term SPO Hot Key (Calidus) could be reused too.

Cerkoryn:
Just pushed an update to the CIP based on good feedback from @lehins and others during the Ledger meeting earlier today.  This version introduces two new certificates as a mirror of the existing CC implementation:

`AuthStakePoolHotKey(poolId, hotCred)`
`ResignStakePoolHotKey(poolId)`

Additionally it allows for a credential instead of just a `KeyHash` which also mirrors the CC implementation and was a point of discussion during the meeting.  It occurred to me while writing this that one of the best use cases for allowing a credential (`ScriptHash` or `KeyHash`) is the ability to use multisigs.  I think this is probably a great solution to the hot key compromise risk that @fallen-icarus mentioned since multisigs offer M of N redundancy.  It may be sufficient enough that the cold key override we discussed may not be needed, though I left it in as its own section if we think the juice is worth the squeeze.

Another major benefit to having a hot key with a multisig will be for large pool operations like exchanges to simply register all their pools to the same hot key.  This allows for voting that is far easier and far more secure at the same time.

@gitmachtl I had written most of this before I read your comment, but I'd like to think this draft covers most of your concerns.  If not I'm also happy to iterate on it further.  Also I changed the title to be `Stake Pool Hot Credentials` so that it is not tied specifically to the name "Calidus", but I do think it is possible to keep existing Calidus key functionality by re-using the public part of the ed25519 key in the new credential.  In which case I am open to keeping the name if the group wants to.

Crypto2099:
This CIP as written now is a substantial improvement over the initial proposal to simply parse CIP-0151 metadata. CIP-0151 was written as a stop-gap measure to give SPOs a meaningful and verifiable way to connect with dApps and other services which may require their authorization/proof of pool ownership for any number of reasons. This CIP, once merged and adopted into the ledger, will likely supercede and make CIP-0151 obsolete but is the better and more correct way to handle the stated goals.

CIP-0175/README.md
rphair: fixed around major rewrite https://github.com/cardano-foundation/CIPs/pull/1140/changes/fb494bd24d9312e865e29588d4603236d85b661a

CIP-0175/README.md
rphair: solved by changing approach as indicated in https://github.com/cardano-foundation/CIPs/pull/1140#issuecomment-3947800536

CIP-0175/README.md
rphair: solved by changing approach as indicated in https://github.com/cardano-foundation/CIPs/pull/1140#issuecomment-3947800536

CIP-0175/README.md
```diff
@@ -0,0 +1,229 @@
...
+CIP: 175
+Title: Stake Pool Hot Credentials
+Category: Ledger
+Status: Proposed
+Authors:
+    - Ryan Wiley <rian222@gmail.com>
+Implementors: []
+Discussions:
+    - https://github.com/cardano-foundation/CIPs/pull/999
+    - https://github.com/cardano-foundation/CIPs/pull/1140
```
rphair:
```suggestion
    - CIP-0151: https://github.com/cardano-foundation/CIPs/pull/999
    - Original PR: https://github.com/cardano-foundation/CIPs/pull/1140
```
Suiting the goal of providing labels for each CIP (not necessary but part of @Ryun1's initiative to provide more useful front matter data).

rphair (commented):
@Cerkoryn comparing your latest version https://github.com/cardano-foundation/CIPs/pull/1140/commits/44ffc1917f6cfdc9b3c3672745b59ab7c5e5d9d6 with the discussion before it, leading up to the resolutions at the Ledger Working Group meeting, I think the current document is a good reflection of what's been resolved so far & therefore a viable proposal.

While this is tagged `Last Check` to perhaps merge at the next CIP meeting — if agreed all around about the new approach, the details, and their documentation — @lehins please have another look to confirm it's consistent with best practice as you've all understood it.

I'll also tag the other CIP editors for review & I'm about to approve it myself (after some detail fixes + Ledger confirmation).  We can easily put a hold on it if needed, but it seems currently to me like this is a good solution that's been agreed all around by knowledgeable stakeholders.

CIP-0175/README.md
```diff
@@ -0,0 +1,229 @@
...
+Introduce:
+
+```
+poolGovHotCreds : Map PoolId Credential
+```
+
+State transitions:
+
+- `AuthStakePoolHotKey(poolId, hotCred)` sets
+  `poolGovHotCreds[poolId] = hotCred` (overwrite allowed).
```
ch1bo: Is the poolId checked before this entry is made?

I suppose it is (should mention it somewhere) and consequently, this means that `AuthStakePoolHotKey` must follow any `RegPool` in the same transaction. Not sure though if the ledger enforces an evaluation order or the transaction creator can pick it. The formal spec just says "runs on a sequence of signals" here https://intersectmbo.github.io/formal-ledger-specifications/site/Ledger.Conway.Specification.Certs.html#sec:the-certs-transition-system

gitmachtl:
> This CIP, once merged and adopted into the ledger, will likely supercede and make CIP-0151 obsolete but is the better and more correct way to handle the stated goals.

It would only make it truly usable if we can include a hotkey publickey somehow in the certificate/update process. For external usage - thats what the intention of calidus key was - we need a reliable way to verify against the hot public key, not the hash of it. So until we can "convince" the ledger team to also add the requirement that for a registration certificate we include both witnesses (cold key witness and hot key witness), there is no way to query the actual hot public key f.e. via db-sync. So metadata "calidus" and the "ledger" version could be out of sync and would require the spos to still use metadata to update there publickey.

gitmachtl:
> Another major benefit to having a hot key with a multisig will be for large pool operations like exchanges to simply register all their pools to the same hot key. This allows for voting that is far easier and far more secure at the same time.

Using the same hotkey for multiple pools is a feature of calidus yes. But it does not require a multisig/script per se imo. The question is if the ledger team would also add a multiassignment into the rules, or if its one hot key per one pool?

Another topic is also if this could be pulled of without introducing many breaking changes in the cardano-cli queries. Best outcome would be if voting with a pool cold key or pool hot key would result in the same voting queries as we have right now to keep tools working. So voting with a hot key would just show up as a normal pool voting like right now.
I guess @CarlosLopezDeLara would like to be involved.

lehins:
> So until we can "convince" the ledger team to also add the requirement that for a registration certificate we include both witnesses (cold key witness and hot key witness), there is no way to query the actual hot public key f.e. via db-sync. 

@gitmachtl Why is it so hard for people to believe me when I am saying something is a bad idea? It seems like again the burden goes on me to do the convincing why someone proposes an idea that is a bad. :cry: 

This is definitely an XY problem. You are suggesting a solution without explaining problem problem properly. So, please, I would appreciate if you could elaborate on the problem you are trying to solve:

What does it mean "It would only make it truly usable". Truly usable by who? Ledger doesn't need it, which means there is no requirement for ledger to enforce witness for hot credential in this new certificate. Usable offchain? Then why don't we consider trying to come up with an offchain solution that does not require unnecessarily complicated changes to the Ledger that is already far from trivial.

Here I am going to explaining why your suggestion is a bad idea, regardless of what your answer to the first point is

1. You suggest requiring a witness, which means also signing the transaction, i.e. proving ownership of the private key. But you only need the preimage of hash, i.e. public key. So, why this extra security requirement? Do you have a good reason for it?
2. Calidus Key is by definition just a key. While this CIP is proposing a hot **credential**. Which means that it will be usable as a plutus script or multisig. Is there going to be support for Calidus scripts? I doubt, since its not even unclear what does it mean to have a plutus script that suppose to work both on chain and off chain.
Also, why would we require a witness for plutus scripts and multisigs? Calidus keys don't need it, Ledger doesn't need it? So what's the point.

>  So metadata "calidus" and the "ledger" version could be out of sync and would require the spos to still use metadata to update there publickey.

Metadata and calidus will already be out of sync by definition, since calidus does not support scripts.

I recommend, instead of hacking into Ledger logic that it doesn't need, taking a step back and thinking of alternative solutions.

I personally would suggest either amending Calidus CIP or introducing a new version of it. Such CIP would require a transaction that contains authorization certificate to also include metadata with the public key of the hash, but only in the case when the hot credential is indeed a key hash:
* If such public of key is present, the it would be called a "Calidus" key
* If subsequent transaction authorizes a different hot key without public key in metadata or a script for the same stake pool, then previous Calidus key becomes invaldiated.

This solution would be easily verifiable offchain (just compare the hash of the public key and to the hot key hash) and could be easily used offchain. It doesn't need to be more complicated than that.

lehins:
> While this is tagged Last Check to perhaps merge at the next CIP meeting 

@rphair, please, let's not rush this. We still have another Ledger Working Group Meeting planned about this in two weeks. And we still need to resolve various aspect of how these new stake pool hot credentials will interplay with Cardano governance. We also still need to specify necessary changes to CDDL and probably many more other points. In other words, it is nowhere near to be considered ready, so please remove the tag Last Check.

gitmachtl:
@lehins 
My comment about "It would only make it truly usable" was about Adams statement that this ledger solution could supercede and make CIP-0151 obsolete. Which is not the case. Yes we're talking two different things here, onchain ledger logic and offchain utility usage. And my statement was that it only would make CIP-0151 obsolete if we can somehow query the hot public key from ledger other than parsing metadata in transactions.

I guess the most simple solution for now is that tooling (CNTools, SPO-Scripts, etc.) submit the metadata beside a new certificate to fulfil the utility usage.

lehins:
> So until we can "convince" the ledger team to also add the requirement that for a registration certificate we include both witnesses (cold key witness and hot key witness), there is no way to query the actual hot public key f.e. via db-sync.

@gitmachtl My response was to this statement in particular. I also provided a solution that would likely work with db-sync as well.

gitmachtl:
> 1. You suggest requiring a witness, which means also signing the transaction, i.e. proving ownership of the private key. But you only need the preimage of hash, i.e. public key. So, why this extra security requirement? Do you have a good reason for it?

Without a witness of the hot key itself, the pool owner can authenticate any hot-key-hash. Even one from a 3rd party pool. So if the 3rd party pool votes with his hot key, that would also be counted as a vote from the own pool, no?
If the signature of the hot key witness is required, you cannot use someone else hot key hash in your own pool.

I think we had the same discussion back about if it should be possible to registered the stake account of anyone as the rewards account of a pool f.e.

lehins:
> Using the same hotkey for multiple pools is a feature of calidus yes. But it does not require a multisig/script per se imo. The question is if the ledger team would also add a multiassignment into the rules, or if its one hot key per one pool?

This is a very good question. For CC members we allow for multiple cold credentials to be mapped to the same cold credential. I think we should follow the same for Stake Pools as well. Implementing for stake pools of this approach will be quite a bit more involved though, when comparing to CC, because for CC we can just iterate over all of the members when we tally up the votes, because CC is usually small.

I'll give it a bit more thought in the next few days, when I have a bit more time available for this and will make appropriate suggestions.

CIP-0175/README.md
lehins: > I suppose it is (should mention it somewhere) and consequently, this means that AuthStakePoolHotKey must follow any RegPool in the same transaction.

It doesn't need to be in the same transaction. It can be in a subsequent transaction. These Hot credentials will be optional, and will be available for addition to already registered stake pools.

> Not sure though if the ledger enforces an evaluation order or the transaction creator can pick it.

Yes, ledger preserves the order in which certificates where supplied over the wire

CIP-0175/README.md
```diff
@@ -0,0 +1,229 @@
...
+  `poolGovHotCreds : Map PoolId Credential`.
+- **Vote source**: Authorization source for a recorded SPO vote:
+  `Cold | Hot`.
+
+### Certificate Semantics
+
+This CIP introduces two new stake-pool governance certificates:
+
+- `AuthStakePoolHotKey(poolId, hotCred)`
+- `ResignStakePoolHotKey(poolId)`
```
lehins: We can get away with adding only one certificate instead:
```haskell
AssignStakePoolHotCredential (KeyHash StakePool) (Maybe (Credential StakePoolHot))
```
or in pseudocode:
```
- `AssignStakePoolHotCredential(poolId, hotCred)`
```

* `Nothing` case would remove hoot credential authorization if there was any
* while `Just` case would overwrite whatever was there before with a new hot credential

The main benefit of this approach would be reduction in complexity (one certificate vs two) and reduction in necessary validation and contention prevention. With two certificates we would probably have to add some validation against prior state. For example in case of resign we would probably have to check that there was a hot credential there before.

In any case. Simpler is better, IMHO.


We'll also need to describe this as necessary change to CDDL

CIP-0175/README.md
```diff
@@ -0,0 +1,229 @@
...
+
+For both certificates:
+
+1. `poolId` MUST identify a currently registered pool.
+2. The transaction MUST include a valid witness by that pool's cold key.
+
+Additional transaction restrictions to prevent ambiguity:
+
+1. A transaction MUST NOT include more than one of these certificates for the
+   same `poolId`. If it does, the transaction is invalid.
```
lehins: This is not necessary. It is ok to include multiple certificates with the same poolId`. It will be the last in the list that will be the final value. This is the same approach we use for all certificate, so no reason to deviate for this one.

CIP-0175/README.md
```diff
@@ -0,0 +1,229 @@
...
+2. A transaction MUST NOT include any SPO governance vote for `poolId` if it
+   also includes one of these certificates for `poolId`. If it does, the
+   transaction is invalid.
```
lehins: Why not?
It's ok for CC member to authorize hot credential and immediately use it int he same Tx

lehins:
> Without a witness of the hot key itself, the pool owner can authenticate any hot-key-hash. Even one from a 3rd party pool. So if the 3rd party pool votes with his hot key, that would also be counted as a vote from the own pool, no?
If the signature of the hot key witness is required, you cannot use someone else hot key hash in your own pool.


I don't see this as a problem. If that is the intention of the stake pool owner then so be it.

Even if we require a witness for hot credential a stake pool could authenticate a multi sig that doesn't require any witness at all, in which case anyone in the world would be able to vote for that pool. Or the owner of stake pool A could authenticate 1of2 multi sig native script where one is owned by pool A and the other by pool B, which would allow pool B to vote on behalf of pool A). So, you can't protect against that.

gitmachtl:
> I don't see this as a problem. If that is the intention of the stake pool owner then so be it.

hmm

lehins:
> hmm

hmmmmm

:rofl:

fallen-icarus:
>>  Without a witness of the hot key itself, the pool owner can authenticate any hot-key-hash. Even one from a 3rd party pool. So if the 3rd party pool votes with his hot key, that would also be counted as a vote from the own pool, no? If the signature of the hot key witness is required, you cannot use someone else hot key hash in your own pool.
>
> I don't see this as a problem. If that is the intention of the stake pool owner then so be it.

AFAIU this will affect Cardano's governance quite substantially by creating "stake pool dReps". Maybe this will not be an issue as long as the hot key can be revoked and a new vote immediately recast? That way if my stake pool dRep votes in a way I don't like, I can still vote directly on the current matter. But given how risk-averse stake pool operators are with their cold keys, we will likely end up with centralized voting power of stake pool dReps.

CIP-0175/README.md
Crypto2099: This is the same way we  handle it with current CIP-151, simply register with a blank "Calidus" key to revoke any previous authorization, cleaner and easier to maintain versus multiple certificate types.

Crypto2099:
> > > Without a witness of the hot key itself, the pool owner can authenticate any hot-key-hash. Even one from a 3rd party pool. So if the 3rd party pool votes with his hot key, that would also be counted as a vote from the own pool, no? If the signature of the hot key witness is required, you cannot use someone else hot key hash in your own pool.
> > 
> > 
> > I don't see this as a problem. If that is the intention of the stake pool owner then so be it.
> 
> AFAIU this will affect Cardano's governance quite substantially by creating "stake pool dReps". Maybe this will not be an issue as long as the hot key can be revoked and a new vote immediately recast? That way if my stake pool dRep votes in a way I don't like, I can still vote directly on the current matter. But given how risk-averse stake pool operators are with their cold keys, we will likely end up with centralized voting power of stake pool dReps.

I was going to say the same thing, we've "accidentally" created SPO DReps via this proposed CIP but I also don't think that's a bad thing. As written and following the same logic as a CC Hot Authorization you could immediately revoke/change your voting key if your SPO Rep votes in a way counter to your beliefs, otherwise, you can follow someone who you respect and feel has a better sense of the pulse on the right way to vote for those requiring SPO votes. IMO this is a cleaner/better solution than our existing "Always No/Abstain" options for SPOs.

gitmachtl:
@fallen-icarus 
I don't like this when it comes to signal hardfork readyness tbh. So i am ready with my setup to do the hardfork, and there are a punch of other SPOs that are using my hot key hash and they may not be ready. And as we have said in the meeting, making an extra excuse/processing again for the hardfork is confusing. I am on your pov that once authorized, a hotkey should be able to vote on all actions for a pool.

CIP-0175/README.md
```diff
@@ -0,0 +1,229 @@
...
+   `(govActionId, poolId)`.
+2. A `Hot` vote MUST NOT overwrite an existing `Cold` vote for the same
+   `(govActionId, poolId)`.
+3. A later `Cold` vote MAY overwrite an earlier `Cold` vote.
+4. A later `Hot` vote MAY overwrite an earlier `Hot` vote, unless a `Cold` vote
+   has already been recorded for that pair.
+
+Effectively, once any cold-authorized vote exists for a given
+`(govActionId, poolId)`, that pair is locked against hot-authorized overwrites.
+This applies regardless of transaction or block ordering history.
```
Crypto2099: I'm not sure if this level of complexity is truly needed here. If a pool has a hot credential authorized, they can change this at any time (reverting to null/none or matching their cold credential). So we can then simplify the governance process for SPOs by only allowing votes via an authorized hot credential. This reduces the surface area of attack where an "SPO DRep" (i.e. I delegate my pool's voting power to @gitmachtl) casts a bad vote with his hot key but override the votes for ATADA with his cold key. Likely this would only happen once as I would delegate away from Martin if he did this, but the point still stands.

However, that does raise another issue, if I have a pool and I have voted with a "hot key" that is enabled at the time and alter nullify or change that hot key prior to the next voting snapshot/tally, does that vote count or is changing the hot key also effectively nullifying any votes that have happened between _n_ and _n+1_ snapshots?

gitmachtl: Still letting allow ColdKey voting gives us an upgrade path without the need that every tool/wallet/hw-wallet is on the same page before we enable all of this.

lehins:
> this will affect Cardano's governance quite substantially by creating "stake pool dReps".

Not exactly. We will need to make sure every vote will contain not just the hot credential, but also the poolId. This means, unlike DReps, there would still need to be a vote per pool. This is necessary to prevent an attack where a pool would submit unlimited number of votes by simply rotating its hot credential. So we will still need to track one vote per pool, the only difference will be that if hot crential is supplied in the vote and it matches the authorized one then we require a witness for hot credential, instead of the cold key.

@fallen-icarus I am also leaning towards your suggestion where if a cold key was used for a vote then a subsequent vote by an authorized hot credential would get rejected.

> I don't like this when it comes to signal hardfork readyness tbh.

@gitmachtl I did bring up this as a question in the meeting yesterday and I tend to agree with you. SPOs need to be actively voting indicating their readiness. We could make HardForkInitiation special and disallow hot credential voting for this action period. Thoughts?

Crypto2099:
> > this will affect Cardano's governance quite substantially by creating "stake pool dReps".
> 
> Not exactly. We will need to make sure every vote will contain not just the hot credential, but also the poolId. This means, unlike DReps, there would still need to be a vote per pool. This is necessary to prevent an attack where a pool would submit unlimited number of votes by simply rotating its hot credential. So we will still need to track one vote per pool, the only difference will be that if hot crential is supplied in the vote and it matches the authorized one then we require a witness for hot credential, instead of the cold key.
> 
> @fallen-icarus I am also leaning towards your suggestion where if a cold key was used for a vote then a subsequent vote by an authorized hot credential would get rejected.
> 
> > I don't like this when it comes to signal hardfork readyness tbh.
> 
> @gitmachtl I did bring up this as a question in the meeting yesterday and I tend to agree with you. SPOs need to be actively voting indicating their readiness. We could make HardForkInitiation special and disallow hot credential voting for this action period. Thoughts?
>
> > I don't like this when it comes to signal hardfork readyness tbh.
> 
> @gitmachtl I did bring up this as a question in the meeting yesterday and I tend to agree with you. SPOs need to be actively voting indicating their readiness. We could make HardForkInitiation special and disallow hot credential voting for this action period. Thoughts?

Based on the above, it sounds like me (ADAM pool) cannot simply piggy-back on the votes of ATADA pool because I would never have access to Martin's hot credential to sign my votes, so this is a bit of a non-issue aside from maybe some troll SPO making it seem like they were part of a larger group although that pool would never cast a vote?

CIP-0175/README.md
Crypto2099: Given that a hardfork would be required before implementing this I don't think looking at tools/wallets/hw-wallet being ready for it is the end of the world.

Crypto2099: Also, the idea is that you would have a path to upgrade:

```
if (!authorizedHotKey) then: coldKeyVotingEnabled
if (authorizedHotKey) then: hotKeyVotingEnabled
if (HotKey === none) then: coldKeyVotingEnabled
```

But get rid of the complexity that both keys can be used simultaneously/interchangeably due to the complexity of calculating which key is valid/authorized at the time of a vote being cast.

gitmachtl:
> This means, unlike DReps, there would still need to be a vote per pool.

I guess that would open up the option to have the pool-id in the vote-certificate as an array instead of a single entry. So larger entities with multiple pools could still vote for there multiple pools with one hot key in one go?

> SPOs need to be actively voting indicating their readiness. We could make HardForkInitiation special and disallow hot credential voting for this action period. Thoughts?

Have to think about it, exceptions are always a bit confusing. Atm for me personally it would feel better to simply disallow using other ones hot hashes that simple. I am not sure if creating such as a SPO-Drep is a good or a bad thing tbh. This may need some input from other "governance" people?

CIP-0175/README.md
lehins: It can work either way, from the Ledger perspective, so I don't have a strong opinion on what we decide:
* Allow cold key to do the voting and give it higher priority
* or only allow hot credentials do the voting

I think the important point we for this discussion is that:
* poolId will need to be included with each vote, regardless of which approach we will take
* authorization of a new hot credential will immediately invalidate all old votes

Last point is important for the concern that was raised by the @fallen-icarus yesterday when a hot credential gets compromised, there should an immediate way to invalidate all votes made on behalf of the cold key.

lehins:
> Based on the above, it sounds like me (ADAM pool) cannot simply piggy-back on the votes of ATADA pool because I would never have access to Martin's hot credential to sign my votes, so this is a bit of a non-issue aside from maybe some troll SPO making it seem like they were part of a larger group although that pool would never cast a vote?

Correct ADAM pool using hot credential of ATADA pool would not automatically receive ATADAs votes. ATADA would have to actively cast a vote on behalf of ADAM pool.

So, in a sense it could be viewed as "SPO DReps" in a sence, except such DReps would have to actively vote for each of their delegators

lehins:
> I guess that would open up the option to have the pool-id in the vote-certificate as an array instead of a single entry. So larger entities with multiple pools could still vote for there multiple pools with one hot key in one go?

No need to complicate things. They can just cast multiple votes in one transaction

> Have to think about it, exceptions are always a bit confusing. Atm for me personally it would feel better to simply disallow using other ones hot hashes that simple. I am not sure if creating such as a SPO-Drep is a good or a bad thing tbh. This may need some input from other "governance" people?

I think you are missing my point. By allowing hot **credentials**, we allow scripts, which can be used to bypass this "uniqueness" protection that you are thinking of. As the simplest example, imagine 3 separate pools `A`, `B` and `C` that authorize different `Ah`, `Bh` and `Ch` respectively, where:
* `Ah = MultiSig_1of2[commonHotKey, dummyHotKey1]`
* `Bh = MultiSig_1of2[commonHotKey, dummyHotKey2]`
* `Ch = MultiSig_1of2[commonHotKey, dummyHotKey3]`

All three hot credentials will be different, but all of them can be satisfied by `commonHotKey`. Does this make sense?

gitmachtl:
It makes sense yes, i'am still not sure if - as i have said in the call - enabling scripts for this is a good thing or not. 👀 But i will not decide this. 😄

fallen-icarus:
> Correct ADAM pool using hot credential of ATADA pool would not automatically receive ATADAs votes. ATADA would have to actively cast a vote on behalf of ADAM pool.
> 
> So, in a sense it could be viewed as "SPO DReps" in a sence, except such DReps would have to actively vote for each of their delegators

This seems pretty easy though. Can't you fit ~150 votes in one tx? It would also be public which pool ids are delegated to your hot key so this seems easy to auto-generate the required tx(s). If ATADA actually wants to be an SPO DRep, they would likely do this.

gitmachtl:
> If ATADA actually wants to be an SPO DRep, they would likely do this.

I will not be a SPO Drep lol 😄 I am not voting in behalf of other SPOs... but thx. 🤣

Crypto2099:
> > Correct ADAM pool using hot credential of ATADA pool would not automatically receive ATADAs votes. ATADA would have to actively cast a vote on behalf of ADAM pool.
> > So, in a sense it could be viewed as "SPO DReps" in a sence, except such DReps would have to actively vote for each of their delegators
> 
> This seems pretty easy though. Can't you fit ~150 votes in one tx? It would also be public which pool ids are delegated to your hot key so this seems easy to auto-generate the required tx(s). If ATADA actually wants to be an SPO DRep, they would likely do this.



> > Correct ADAM pool using hot credential of ATADA pool would not automatically receive ATADAs votes. ATADA would have to actively cast a vote on behalf of ADAM pool.
> > So, in a sense it could be viewed as "SPO DReps" in a sence, except such DReps would have to actively vote for each of their delegators
> 
> This seems pretty easy though. Can't you fit ~150 votes in one tx? It would also be public which pool ids are delegated to your hot key so this seems easy to auto-generate the required tx(s). If ATADA actually wants to be an SPO DRep, they would likely do this.

This seems like a net benefit rather than a drawback (if any SPO is willing to serve in this capacity on behalf of stake pools that they do not control for whatever reason) and, it's _contingent_ on both parties agreeing to the relationship so it's not as though I can _force_ @gitmachtl to vote for me.

gufmar:
I'm not saying that I like or endorse such setups, but there will definitely be a group of stake pool constellations where the SPaaS guy will appreciate the option to save time and customer costs by casting everything at once instead of dealing with many pools and cold keys. 

Question is then if this would increase the voting participation and clarity on things like HF support

rphair:
Removing `Last Check` tag in response to the huge amount of constructive discussion (plus @lehins' https://github.com/cardano-foundation/CIPs/pull/1140#issuecomment-3950850767) following the first of the Ledger meetings where this proposal was discussed.

Cerkoryn:
@lehins I have on my calendar that there is a Ledger WG meeting today, presumably as a continuation of our discussion of this CIP.  However, I don't seem to have a link or invite to the meeting 😅 

If it is happening still, would anyone mind shooting me the link to join it?

lehins:
> If it is happening still, would anyone mind shooting me the link to join it?

@Cerkoryn Sorry, I was unavailable last week during the regular time of the event, hence it wasn't scheduled last week and will happen this week instead

Apologies for the short notice. We'll be holding a Ledger Working Group in 4 hours. If you could make it that would be awesome. We could discuss this CIP further.
https://luma.com/5e1698s5

CIP-0175/README.md
```diff
@@ -0,0 +1,270 @@
...
+
+## Abstract
+
+Stake Pool Operators (SPOs) currently sign on-chain governance votes with pool
+cold keys. This CIP introduces a new on-chain certificate that lets a pool cold
+key authorize a hot credential for a stake pool. For
+`StakePoolVoter poolId`, the ledger accepts either:
+
+- the pool cold key witness, or
+- a witness that satisfies the currently authorized hot credential.
```
lehins: I'd suggest doing the same thing as with CC members, only cold key can authorize new hot credentials. If we don't do that it will require a bit of work:
* Support for Plutus scripts will need to be implemented for this new certifying approach
* Certificate will need to be explicit which hot or cold is to be used for authorization, since "either or" will not be automagically inferred by the ledger

CIP-0175/README.md
```diff
@@ -0,0 +1,270 @@
...
+
+- If `hotCredOpt = hotCred` (a `Credential`), the pool's hot credential is
+  set to `hotCred` (overwrite allowed).
+- If `hotCredOpt = null`, any existing hot credential for `poolId` is revoked.
+  If none exists, this is a valid no-op.
+
+### Certificate CDDL
+
+```cddl
+; Numeric certificate tag allocation is TBD in Conway ledger CDDL integration.
```
lehins: Conway will definitely not have this feature implemented
```suggestion
; Numeric certificate tag allocation is TBD in Ledger CDDL integration.
```

CIP-0175/README.md
```diff
@@ -0,0 +1,270 @@
...
+
+For `AssignStakePoolHotCredential(poolId, hotCredOpt)`:
+
+1. `poolId` MUST identify a currently registered pool.
+2. The transaction MUST include a valid witness by that pool's cold key.
+3. If multiple `AssignStakePoolHotCredential` certificates for the same
+   `poolId` appear in one transaction, they are processed in transaction order;
+   the last one determines the final hot credential state for that `poolId`.
+4. A transaction MAY include SPO governance vote(s) for the same `poolId` as
+   one or more `AssignStakePoolHotCredential` certificates.
```
lehins: I am not sure what this point really means. Could you please elaborate. Maybe an example will help?

CIP-0175/README.md
```diff
@@ -0,0 +1,270 @@
...
+5. Same-transaction vote authorization for `poolId` MUST evaluate against the
+   hot credential state resulting from certificates that appear earlier in that
+   transaction's certificate list.
```
lehins: Submission of votes is processed after certificates are processed, so this property will automatically hold. Just pointing out that it might be worth mentioning it here

CIP-0175/README.md
```diff
@@ -0,0 +1,270 @@
...
+  `poolGovHotCreds` entry.
+
+### Governance Vote Authorization Change
+
+For each vote with `Voter = StakePoolVoter poolId`, authorization succeeds if
+either of the following holds:
+
+1. The transaction includes a valid witness by the pool cold key for `poolId`.
+2. `poolGovHotCreds[poolId] = hotCred` exists and transaction witnesses satisfy
+   `hotCred`.
```
lehins: That's not the right approach. We need to try to be deterministic in this.
A better way would be for Voter to be changed to:
```haskell
 | StakePoolVoter !(KeyHash 'StakePool) (Maybe (Credential Hot))
```
And semantics would be: 
* If hot credential is present then witness for it is required
* otherwise it is the cold key witness is required

In the first case Ledger would ensure that credential matches the one in the stake pool state

CIP-0175/README.md
Cerkoryn: I think I was thinking that you could vote and create a hot credential in the same Tx, but based on your comment for point #5 I think it's fine to let the ledger automatically inherit this property.  I will just remove both points for simplicity.

CIP-0175/README.md
Cerkoryn: Since this is an inherent property of the ledger I think it's simpler just to leave it out.

Cerkoryn:
@lehins I implemented the changes/clarifications you recommended in the latest update here a few weeks ago.  If it looks good to you then maybe we can check with @rphair or @Ryun1 about what else needs to be done to progress this CIP?

rphair:
re: latest comment above: if @lehins @Cerkoryn you are both satisfied with the latest changes (please resolve the individual review points to confirm), I'd be willing to approve it myself (pending one last check for formatting / readability) and then it can be moved to `Last Check` for the following CIP meeting.

If there are any other reservations expressed that need discussion, we can add it to the CIP meeting _Review_ section if they remain unsettled here.  So we will keep this moving along one way or another.

Cerkoryn:
@rphair @lehins Just checking back in on this CIP to see if there are any changes that might need to be made?

CIP-0175/README.md
```diff
@@ -0,0 +1,297 @@
...
+CIP: 175
+Title: Stake Pool Hot Credentials
+Category: Ledger
+Status: Proposed
+Authors:
+    - Ryan Wiley <rian222@gmail.com>
+Implementors: []
+Discussions:
+    - CIP-0151: https://github.com/cardano-foundation/CIPs/pull/999
+    - Original PR: https://github.com/cardano-foundation/CIPs/pull/1140
```
rphair:
```suggestion
    - Original PR: https://github.com/cardano-foundation/CIPs/pull/1140
    - CIP-0151 PR: https://github.com/cardano-foundation/CIPs/pull/999
```
Merged in the meantime; though PR discussion likely more relevant than CIP-0151 content itself; while still not more relevant than `this` PR (therefore appearing first).

rphair (approved):
@Cerkoryn without any apparent objection remaining after @lehins' series of comments beginning with https://github.com/cardano-foundation/CIPs/pull/1140#discussion_r2941069188 I do think we would be ready to merge this unless any remaining objections are emphasised or new objections posted.
* @lehins if we need to amend that please continue any of the discussion points that you raised above.
* Also it would be a help to mark the adequately resolved questions & issues "Resolved" in the UI itself.

Since this has been without dispute after the latest update for 3 months — certainly more time than necessary, mainly awaiting advocacy to press for merge and/or this season's governance preoccupations to mostly conclude — I would feel comfortable marking this `Last Check` for merge if possible at the next CIP meeting: https://hackmd.io/@cip-editors/139

... while asking all other prior reviewers to use the next week to post any objections or express approval for the spec as adjusted after the most recent feedback from the Ledger WG meeting (https://github.com/cardano-foundation/CIPs/pull/1140/commits/5c6b45a7d38444a9e14a68d35076b828de038b49): 

@Ryun1 @gitmachtl @WhatisRT @ch1bo @gufmar @fallen-icarus @Crypto2099

CIP-0175/README.md
rphair: @Cerkoryn I'm committing this to get an immediate confirmation that this passes CIP validation (or display why it doesn't).

CIP-0175/README.md
```diff
@@ -0,0 +1,297 @@
...
+```
+
+State transitions:
+
+- `AssignStakePoolHotCredential(poolId, hotCredOpt)` with
+  `hotCredOpt = hotCred` sets `poolGovHotCreds[poolId] = hotCred`
+  (overwrite allowed).
+- `AssignStakePoolHotCredential(poolId, hotCredOpt)` with
+  `hotCredOpt = null` removes `poolId` from the map if present (otherwise
+  no-op).
```
lehins, line 119: We need to explicitly list here that, unlike regular pool params update, this certificate will cause the credential to be updates for the stake pool immediately after transaction is applied on chain. Which is also the reason why it had to be implemented as a new certificate, instead of relying on the usual RegPool certificate that is capable of updating stake pool state.

CIP-0175/README.md
```diff
@@ -0,0 +1,297 @@
...
+
+No uniqueness constraint is imposed on `hotCred`. The same hot credential MAY be
+authorized for multiple pools.
+
+Pool retirement lifecycle:
+
+- If a pool has a retirement certificate scheduled but not yet enacted, its hot
+  credential remains valid.
+- When retirement is enacted, the ledger clears that pool's
+  `poolGovHotCreds` entry.
```
lehins: Technically speaking it is not needed, since upon retirement the whole stake pool state is cleared out, which will include the hot credential.
But, it's ok to keep it here, since what I mentioned above is just an implementation detail (i.e. it would be impossible to retain the hot credential for a retired pool)

CIP-0175/README.md
```diff
@@ -0,0 +1,297 @@
...
+   MUST verify that `poolGovHotCreds[poolId] = hotCred`.
+
+Hot credential satisfaction rules:
+
+- If `hotCred` is a key credential, the corresponding vkey witness is required.
+- If `hotCred` is a script credential, a valid governance voting script witness
+  is required under existing governance voting-script rules for
+  `StakePoolVoter poolId (Just hotCred)`.
+- Both native scripts and Plutus scripts are supported.
+
```
lehins, line 161:
```suggestion
- Existing `Voting` script purpose will be used for Plutus script execution, since it was unused for stake pools before

```

CIP-0175/README.md
```diff
@@ -0,0 +1,297 @@
...
+overwritten SPO votes do not need to be retained.
+
+Overwrite rules:
+
+1. A `Cold` vote MAY replace the current vote (`Cold` or `Hot`) for the same
+   `(govActionId, poolId)`.
+2. A `Hot` vote MAY replace the current `Hot` vote for the same
+   `(govActionId, poolId)`.
+3. A `Hot` vote MUST NOT replace the current `Cold` vote for the same
+   `(govActionId, poolId)`.
```
lehins: I am suggesting "SHOULD", because ignoring such a vote does not compromise the system, but would result in potentially dangerous bugs in tooling and confusing users.
```suggestion
3. A `Hot` vote MUST NOT replace the current `Cold` vote for the same
   `(govActionId, poolId)`. An attempt to do so SHOULD result in a predicate failure, instead of being silently ignored.
```

CIP-0175/README.md
```diff
@@ -0,0 +1,297 @@
...
+
+Effectively, once any cold-authorized vote exists for a given
+`(govActionId, poolId)`, that pair is locked against hot-authorized overwrites.
+
+### Authorization-Change Vote Invalidation
+
+If `AssignStakePoolHotCredential(poolId, hotCredOpt)` results in an actual
+authorization-state change for `poolId`, the ledger MUST clear the current SPO
+votes for that `poolId` across still-active governance actions whose recorded
+`VoteSource` is `Hot`.
```
lehins, line 189: It is not necessary to clear out the vote. In fact we do not want to do this, because that would be an expensive operation. It is much simpler to not count votes during ratification that were made by a hot credential that does not match the up to date vote in the stake pool state.
```suggestion
If `AssignStakePoolHotCredential(poolId, hotCredOpt)` results in an actual
authorization-state change for `poolId`, the ledger MUST not use this current SPO
votes for that `poolId` across still-active governance actions whose recorded
`VoteSource` is `Hot`.
```

CIP-0175/README.md
```diff
@@ -0,0 +1,297 @@
...
+  `hotCred` plus `poolGovHotCreds[poolId] = hotCred` for
+  `StakePoolVoter poolId (Just hotCred)`.
+- **`GOV`** (vote state handling) records `VoteSource` for stake-pool votes and
+  stores only the latest current vote per `(govActionId, poolId)`, applying
+  cold/hot overwrite rules plus hot-vote invalidation on actual
+  authorization-state changes.
+
+The proposal does not introduce a new governance role. It extends
+`StakePoolVoter` so the SPO vote explicitly carries whether it is using cold
+authorization or a specific hot credential.
```
lehins: This whole section goes into how this CIP would be implemented. I'd recommend avoiding these details, especially since they are inaccurate.
```suggestion
```

CIP-0175/README.md
```diff
@@ -0,0 +1,297 @@
...
+### Security Considerations
+
+- **Hot credential compromise near deadline**:
+  If hot keys are suspected to be compromised, changing or revoking the
+  authorized hot credential immediately invalidates current hot-authorized SPO
+  votes for that pool on still-active governance actions, while current
+  cold-authorized votes remain valid. Operators can then recast under current
+  authorization (including with the cold key). Additionally, large stake pool
+  groups can mitigate this risk further by choosing to use a multisig hot
+  credential.
```
lehins, line 242: It would probably be useful to note here that during one epoch delay before gov action enactment, neither hot credential authorization change, not a vote change can affect that enactment. This is also applicable to votes by DReps and CC members alike.

CIP-0175/README.md
```diff
@@ -0,0 +1,297 @@
...
+  example multisig and timelock designs) to reduce single-key compromise risk.
+- **Intentional delegation via shared or third-party hot credentials**:
+  A pool cold key can authorize any `Credential`, including one controlled by a
+  different operator or reused across multiple pools. Votes cast as
+  `StakePoolVoter poolId (Just hotCred)` then count for each pool that has
+  authorized that `hotCred`. This representative behavior is an intentional
+  capability, not a validation flaw, but it can concentrate SPO voting
+  influence. If that outcome is undesirable, pool operators can revoke or
+  replace the hot credential and, for the current governance action, reassert
+  intent with a cold-authorized vote.
```
lehins: This is incorrect. We should not apply automatically one vote made by one hot credential for all stake pools that authorized the same hot credential. In fact this is not what we discussed during the working group either.

There are technical reasons why we do not want to implement it this was. But also, we do not want one SPO A mimicking votes of another SPO B, without authorization, by the SPO B.

Therefore, I'd recommend changing this text to something like this.
```suggestion
- **Intentional delegation via shared or third-party hot credentials**:
  A pool cold key can authorize any `Credential`, including one controlled by a
  different operator or reused across multiple pools. Many votes with different
  `poolId`s, but the same `hotPoolCred` can be submitted like this: 
  `StakePoolVoter poolId (Just hotPoolCred)`. In fact as many as there are pools 
  that authorized that `hotPoolCred`. Note that there is no requirement for all 
  those votes to be the same, eg. for some pools it could be Yes, for others No.
  This representative behavior is an intentional capability, not a validation
  flaw, but it can concentrate SPO voting influence. If that outcome is
  undesirable, pool operators can revoke/replace the hot credential or replace
  the vote by hot credential for the current governance action with a 
  coldkey-authorized vote.
```

lehins (commented):
I still had a few minor suggestion and one important one that needs to be addressed before I would consider this CIP to be ready.

Cerkoryn:
Thanks for review @lehins.  I've applied you suggestions and added revisions for the sections that you mentioned.  Please let me know if there is anything remaining to be done before this CIP can be deemed ready.

rphair:
I came across this letter of endorsement which helped my own understanding of the eventual role of this CIP & does seem to suggest the importance of merging it soon, if that can be done without reservation on the Ledger side (@lehins please double-check soon as suggested in @Cerkoryn's https://github.com/cardano-foundation/CIPs/pull/1140#issuecomment-4981473292 so we can have a fully informed `Last Check` on Tuesday):
* https://docs.google.com/document/d/19NCm3wC2LQI6g3DcpbBm9g597uZcjZiMvwWziqSy_04

lehins (approved):
This CIP looks great 👍

lehins:
My response to the document: https://docs.google.com/document/d/19NCm3wC2LQI6g3DcpbBm9g597uZcjZiMvwWziqSy_04

> We request that the TSC prioritize CIP-0151 for inclusion in the Dijkstra-era hard fork CIP shortlist, and confirm the review timeline needed to keep this work on track for the 2027 release window. 

@rphair I am not sure whether this is just an oversight or is it me simply not understanding this document correctly, but CIP-151 has already been merged and is a totally of chain component, which does not require any Ledger changes as pointed out [in the CIP](https://github.com/cardano-foundation/CIPs/tree/23b89bd6c37ba9161e3f3172a57f9b2fa516bc18/CIP-0151#motivation-why-is-this-cip-necessary): "This method is preferred over a change to the in-ledger stake pool registration certificates"

That being said, if Civics Committee actually meant to formally endorse this CIP-175, then I personally have no objection, but I do have important points that we need to keep in mind:
* First and foremost, I'd highly recommend Civics Committee to modify the letter to specify CIP-175 instead of CIP-151, just so there are no ambiguities or misunderstandings
* According to the document, Dijkstra delivery is expected in 2027 by the Civics Committee. However, from what I know, there seems to be a desire at IOG to have Dijkstra era ready for hard fork in late 2026. So, if this CIP-175 is to be added as a feature to Dijkstra era, it would postpone it's delivery at the very least by 4 months. This CIP is not a trivial one that could be done in matter of a few days, because it touches many aspects of the Ledger, namely new certificates, Stake Pool update process, governance, etc. Besides implementation in Haskell it will also require formalization in Agda.
* Last, but not least. As far as I know, there was no Treasury Withdrawal proposal to implement any CIP that allows Calidus keys to be used in Cardano Governance, which means currently there are no resources allocated for development of such CIP. But, according to the governance on Cardano as we have it today, someone has to pay for its implementation and if it is the Treasury, then realistically it would postpone the Dijkstra deadline way more than 4 months, since that process takes quite a bit of time.

rphair:
thanks @lehins — I'd just noticed that document [posted](https://discord.com/channels/1136727663583698984/1262393293581324418/1527749103931359363) in the Intersect Discord's TSC Community Chat (cc @Cerkoryn) and thought it would be useful to those following the newer CIP's review.  I'm definitely not representing, nor am I supporting, the Civics Committee: but I'll link that posting to your feedback here in the interest of a better CIP evolution.

kenerik:
> My response to the document: https://docs.google.com/document/d/19NCm3wC2LQI6g3DcpbBm9g597uZcjZiMvwWziqSy_04
> 
> > We request that the TSC prioritize CIP-0151 for inclusion in the Dijkstra-era hard fork CIP shortlist, and confirm the review timeline needed to keep this work on track for the 2027 release window.
> 
> @rphair I am not sure whether this is just an oversight or is it me simply not understanding this document correctly, but CIP-151 has already been merged and is a totally of chain component, which does not require any Ledger changes as pointed out [in the CIP](https://github.com/cardano-foundation/CIPs/tree/23b89bd6c37ba9161e3f3172a57f9b2fa516bc18/CIP-0151#motivation-why-is-this-cip-necessary): "This method is preferred over a change to the in-ledger stake pool registration certificates"
> 
> That being said, if Civics Committee actually meant to formally endorse this CIP-175, then I personally have no objection, but I do have important points that we need to keep in mind:
> 
> * First and foremost, I'd highly recommend Civics Committee to modify the letter to specify CIP-175 instead of CIP-151, just so there are no ambiguities or misunderstandings
> * According to the document, Dijkstra delivery is expected in 2027 by the Civics Committee. However, from what I know, there seems to be a desire at IOG to have Dijkstra era ready for hard fork in late 2026. So, if this CIP-175 is to be added as a feature to Dijkstra era, it would postpone it's delivery at the very least by 4 months. This CIP is not a trivial one that could be done in matter of a few days, because it touches many aspects of the Ledger, namely new certificates, Stake Pool update process, governance, etc. Besides implementation in Haskell it will also require formalization in Agda.
> * Last, but not least. As far as I know, there was no Treasury Withdrawal proposal to implement any CIP that allows Calidus keys to be used in Cardano Governance, which means currently there are no resources allocated for development of such CIP. But, according to the governance on Cardano as we have it today, someone has to pay for its implementation and if it is the Treasury, then realistically it would postpone the Dijkstra deadline way more than 4 months, since that process takes quite a bit of time.

Thanks @lehins  and the rest of you working on this.
Tbh, I've spent hours on this topic as well.

Short background:
The Beyond MVG state of governance report has measured a steady decline in SPO participation in governance over the past year. 
The report collected data from the SPO community to find out why that is so. Using cold keys for voting was "the friction," rather than SPOs not wanting to participate, this is due to security concerns.
This report was presented to the CIVICS by the Beyond team and they agree to endorse reducing the SPO friction.
CIVICS did not examen the Technical approach on how to implement, rather just recognizing of the signature scheme in CIP 151 as a "tool" to get there.

As part of the Beyond MVG team, I started looking for CIPs in the ecosystem and other ideas back in February. I also had to climb a steep learning curve on the inner "nuts and bolts" of the Cardano ledger.

Beyond MVG wanted to find some "low hanging fruits", simple to implement, that could reduce voting power concentration and reduce friction for SPOs.
That's why the report, among many other things, wanted to bring CIP-151 to the ledger development team's attention.
With my limited knowledge, I understand that CIP-175 is not a "low hanging fruit," as it requires a much larger scope of resources to implement.

That's why I landed on CIP-151 and the idea of creating a "relation" between the Calidus signature and the pool ID, and expanding it into an on-chain ledger feature. With my limited knowledge, I was thinking along the lines of a new STS rule to fetch a Calidus signature (with a header) from the metadata, verify the signature, and update that index with a nonce and expiry. Then, in the witness requirement computation for SPO votes, add an OR rule so a valid hot-key signature satisfies the requirement in place of the cold key.

Sorry for my scribbling here with limited knowledge.

Looking at your rationale, the "mission" here is to enable SPO voting with hot credentials that can be kept in a Trezor/Ledger/wallet, so that SPOs can interact with governance. I see now that CIP-175 is a more "established and reasonable" approach.

I will bring this very good feedback to CIVICS' attention and get back to you. I will also inform the Beyond MVG team of your sharp insights.

Thanks @rphair for adding this here

For the record, the Dijkstra hard fork estimate was just a fair guess. 
I hope we can get the scope down to fit the timeline.
Understandably, I agree that SPO hot credentials are not feasible to implement on this short timeline if CIP-175 is the only way to go.

lehins:
> That's why I landed on CIP-151 and the idea of creating a "relation" between the Calidus signature and the pool ID, and expanding it into an on-chain ledger feature. With my limited knowledge, I was thinking along the lines of a new STS rule to fetch a Calidus signature (with a header) from the metadata, verify the signature, and update that index with a nonce and expiry. Then, in the witness requirement computation for SPO votes, add an OR rule so a valid hot-key signature satisfies the requirement in place of the cold key.

Yeah, something similar was also proposed by @Cerkoryn as the initial idea to solving this problem. However, as I pointed out in [Ledger Working Group 33](https://drive.google.com/file/d/1FBo9LFzt62PX12uRZ8MoXTMDgn5XXDcB/view) this is not a viable idea. Ledger cannot relay on contents of Metadata, which is how Calidus keys are really defined.
This CIP-175 is the next iteration on that initial idea that is actually sound.

gitmachtl:
Yea the outcome was that tooling will try to send metadata along a "new" SPO HotKey registration to keep OnChain and OffChain data in sync. 

@lehins We talked about to have a separate Certificate for this and not forcing the SPOs to do a Pool-ReRegistration every time they update there HotKey, whats the current state on this?

@kenerik Its not really needed to have a HW-Wallet Integration for a HotKey. The whole point of a HotKey is that it can be changed quickly in case of a compromised one. But ofc would be nice to have. Once we have settled on an integration, i can talk with the VL guys if it would be doable and in what TimeFrame. Depends also on some GovAction Proposals to keep the HW-Wallet maintained.

Thx!

lehins:
> @lehins We talked about to have a separate Certificate for this and not forcing the SPOs to do a Pool-ReRegistration every time they update there HotKey, whats the current state on this?

Yes, it is crucial to introduce a new certificate for this to be safe, since it can take up to a whole epoch for pool params to be updated using the usual mechanism, while hot credentials need to be updateable ASAP. If you look at the second sentence in the CIP you will see this is exactly the case:
> This CIP introduces a new on-chain certificate that lets a pool coldkey authorize a hot credential for a stake pool.

perturbing (approved):
Given that @lehins is okay with the CIP (and that it is sound), I approve.

That said; hope someone will fund this work 👍

Ryun1 (approved):

CIP-0175/README.md
ch1bo: @Cerkoryn I have sketched this out now in https://github.com/IntersectMBO/formal-ledger-specifications/pull/1309 and https://github.com/IntersectMBO/cardano-ledger/pull/6050 to see how hard this really is. The need for having a new witness type was not obvious to me before, but could still be a nice in-between version until we have the Calidus registration as described in this CIP? CC @lehins

Cerkoryn: I think a lot of SPOs, community members, and the Civics committee would love if we could figure out how to get this into production.

That being said I think I rewrote the CIP since our earlier conversation here several months ago.  I think we gave up on the idea of trying to make the Calidius keys work and just went with a new cert instead.

lehins: @ch1bo your sketch is absolutely not what this CIP is describing.

Issue #989, opened 2025-02-18
github action workflows


Ryun1:
ideas for workflows to automate things

- workflow to check pull requests that contain CIPs/CPSs include the correct yaml header
- conformance to CIP/CPS template
- pull request description linking to rendered version

Ryun1:
> workflow to check pull requests that contain CIPs/CPSs include the correct yaml header

a workflow like [validate-header.yml](https://github.com/Ryun1/template-test/blob/91c2929f39b3131cbb52af7286e8b645eae5160b/.github/workflows/validate-header.yml)

runs a script like [validate_markdown_header.py](https://github.com/Ryun1/template-test/blob/91c2929f39b3131cbb52af7286e8b645eae5160b/.github/scripts/validate_markdown_header.py)

rphair:
> a script like [validate_markdown_header.py](https://github.com/Ryun1/template-test/blob/91c2929f39b3131cbb52af7286e8b645eae5160b/.github/scripts/validate_markdown_header.py)

and also, as per https://github.com/cardano-foundation/CIPs/pull/925#discussion_r1959732496, impose a sort order on the YAML fields?  (though personally I think it's fine to leave to author's discretion, since some fields seem to have a different "significance" based on the subject of a CIP / CPS)

(this comment essentially duplicates https://github.com/cardano-foundation/CIPs/issues/863#issuecomment-2666022734)

0xi3gold:
We should create workflows to automatically validate PRs that include CIPs or CPSs. Specifically:
 
1. Validate YAML headerEnsure every CIP/CPS includes a valid YAML header.
Check presence of required fields.
Flag missing or malformed fields.

2.Conformance to CIP/CPS template
Compare the submitted file against the official CIP/CPS template.
Warn if sections are missing or in the wrong format.

3.Check PR description
Ensure the PR description contains a link to the rendered version of the document.
Possibly comment on the PR if the link is missing.

Ryun1:
idea; workflow to check that links work

Ryun1:
here is my working list of validations to apply
I feel this is a reasonable first go at it
the deeper you look the more there is that could be done

### CIP Validation

Validations applied to `/CIP-*/README.md`

File level validations:
- line endings are UNIX

`yaml` header validations:
- all required fields present and in order: CIP, Title, Category, Status, Authors, Implementors, Discussions, Created, License
- all required fields are expected types/structure (as defined by a JSON schema)
[...]

PR #1143, opened 2026-01-29
CPS-0026? | Approaches to Fully Homomorphic Encryption 

kozross:
This CPS discusses what benefits having full homomorphic encryption (specifically [TFHE][tfhe]) on Cardano would have, and what problems would have to be addressed to deal with it. 

[Rendered](https://github.com/mlabs-haskell/CIPs/blob/tfhe/CPS-0026/README.md)

[tfhe]: https://tfhe.github.io/tfhe/


CPS-tfhe/CPS.md
```diff
@@ -0,0 +1,187 @@
+---
+CPS: TBD
+Title: Fully homomorphic encryption with programmable bootstrapping
+Category: Plutus
+Status: Open
+Authors: Koz Ross <koz@mlabs.city>
+Proposed Solutions: []
+Discussions: https://github.com/cardano-foundation/CIPs/pull/?
```
rphair:
```suggestion
Discussions:
  - https://github.com/cardano-foundation/CIPs/pull/1143
```
`this`

CPS-0026/README.md
```diff
@@ -0,0 +1,187 @@
+---
+CPS: TBD
+Title: Fully homomorphic encryption with programmable bootstrapping
```
rphair: ... as indicated above: great title for a CIP but likely too specific & not "open" enough for a CPS.

CPS-0026/README.md
```diff
@@ -0,0 +1,187 @@
...
+* Implementing TFHE ourselves and integrating it into Plutus Core (the _native_)
+  approach.
+
+From a developer experience perspective, this choice wouldn't affect much, as
+the interface that would be provided would be very similar in either case.
+However, other considerations make these two choices quite different in
+practice, as discussed in the 'Open Questions' section. Thus, it is likely that
+a research step would be needed before any decision is made either way.
+
+## Open Questions
```
rphair, line 133: Eventually some of the (currently) big paragraphs below could use some breaking up.  Personally I think it would be more readable to interested parties with some subheadings for the background items like "state of the art" and then either headings or list items for the "open questions" themselves.

Right now it reads like a "wall of text" and breaking it up somehow would be more of an invitation to participate in the discussion & get started on solutions.

rphair (commented):
@kozross this is on agenda for `Triage` at next CIP meeting (https://hackmd.io/@cip-editors/127) where I hope we can confirm some title which is specific enough to cover Fully Homomorphic Encryption but general enough to accommodate any of the approaches under consideration here.  
* Ideally a CPS should be stated as a problem rather than a solution.  I've given it a "working title" and please feel free to refine it further & we'll also visit that issue at the meeting.

CPS-0026/README.md
kozross: I understand where you're coming from with this. I'm used to writing CIPs, rather than CPSes, which is the probable cause. Programmable bootstrapping is a rather key component of why I believe TFHE specifically to be so worthwhile: the question is less 'which scheme should we use', but 'how should we use it', and I'd like the title to reflect that. 

Definitely open to suggestions!

CPS-0026/README.md
kozross: Yeah, agreed, it is definitely a bit too unstructured. Will try to organize it better.

kozross:
@rphair - I renamed `CPS.md` to `README.md` to match your request [on my other CPS draft](https://github.com/cardano-foundation/CIPs/pull/1144). 

Edit: I've also reworded the open questions section to hopefully be less wordy and better organized.

CPS-0026/README.md
rphair:
```suggestion
Title: Approaches to Fully Homomorphic Encryption
```
OK, I will put in a vote for what is currently the PR title... pending any stronger suggestions that might come in.

CPS-tfhe/README.md
```diff
@@ -0,0 +1,226 @@
...
+having full control over the implementation would mean two additional benefits:
+
+* Only Plutus Core would require changing; and
+* The implementation of TFHE could potentially be used elsewhere, in ways not
+  related to the chain, by other Haskell developers.
+
+At the same time, such an implementation would require highly non-trivial
+amounts of work. Although two reference implementations of TFHE exist, neither
+are suitable for direct use: the C implementation is not of production quality,
+and `tfhe-rs` is written in Rust, which is difficult to use in Haskell via FFI.
```
perturbing: This is not per se a problem, given that `tfhe-rs` can export them to C.

A PoC example of such a Rust -> C -> Haskell pipeline can be found [here](https://github.com/cardano-scaling/rust-accumulator) which we bind to [here](https://github.com/cardano-scaling/haskell-accumulator).

perturbing (commented):
Just a quick review: nice proposal! I see the value, I do wonder if these operations will fit inside the Plutus budget. 🤔 Having a benchmark as an indication is paramount. (see for example the bls12-381 msm cip [here](https://cips.cardano.org/cip/CIP-0133))

CPS-tfhe/README.md
kozross: That is good to know - I will amend the text in light of this. Thank you!

kozross:
@perturbing - 'fitting inside the budget' is a problem, but I feel the _bigger_ issue is that programmable bootstrapping is like higher-order builtins on steroids. Consider something like this kind of builtin and its intended signature (in Plinth):

```MapList :: forall a b . (a -> b) -> BuiltinList a -> BuiltinList b```

Nothing semantically prohibits us from defining such a builtin. However, how on earth do you cost this? The only assurances we have are that the function argument would get called a number of times proportional to the length of the list argument, as the function argument could do more-or-less arbitrary amounts of work, which we wouldn't be able to figure out. Programmable bootstrapping is this, but even worse: Boolean circuits are a very general model of computation, which gives even less to go on. And that's _before_ we consider that they can be encrypted!

@kwxm explained the problem of higher-order builtin costing to me at length - I would love to hear his opinion on the problem of costing something like programmable bootstrapping and what possible solutions could exist.

In terms of having a benchmark, I believe this is a concern for a CIP, not a CSP. Indeed, I see this as a call to action towards _getting_ us to the point where we know what we need to even do such a measurement.

**Comment by perturbing** [comment]
> However, how on earth do you cost this? 

I see your point, and I do not have the answer (unfortunately). I know that costing is complex and the devil is in the details, and clearly your function `MapList` is too unconstrained to be costed (as the `f` that gets mapped can be arbitrarily complex). But is this also the case for FHE? Is the space of arguments of such a function not bound? If so, we just need to find the measure :thinking: 

I do not know yet how the interface of your linked library works, but if it is indeed similar to `MapList`, I wonder if it can be done.

**Comment by kozross** [comment]
@perturbing - that is exactly my point. Hopefully @kwxm might be able to chime in and clarify the situation a bit better, as I'm not a costing expert.

Edit: Also @perturbing - addressed your comment regarding Rust FFI.

File: CPS-tfhe/README.md [hunk 5]
```diff
@@ -0,0 +1,234 @@
...
+how to address the interaction of costing and programmable bootstrapping.
+Answers to these questions are related, but ultimately independent: whatever
+choice is made for the first will impact, but not eliminate, the second. We
+discuss these questions further below, and describe some of the implications of
+possible solutions.
+
+### Choice of approach
+
+Each of the three previously-described approaches to making TFHE capabilities
+available have advantages and drawbacks, which makes the choice between then
```
Unisay:
```suggestion
available have advantages and drawbacks, which makes the choice between them
```

kozross: Good catch, thanks!

rphair (commented):
@kozross this was considered a well researched production at the CIP meeting today and a good starting point for discussing how to include this in Cardano.  There were some views that maybe homomorphic encryption would be better provided from elsewhere: and then there was a counter-argument of "why pass that business off to other chains & L2's when we might not have to?".

As already pointed out the main question is cost and we noted that @kwxm was already tagged - although if this is coming from/through Mlabs we also noted that the primary responsibility of providing a refined costing model & estimate should be from the proposer.

@kozross please continue to tag others who may be interested in the general features of this - I am no subject matter expert myself but think @effectfully @michaelpj could be interested & insightful about this.

Please in any case change the containing directory to `CPS-0026` and update the "rendered" link in your original comment 🎉

CPS-tfhe/README.md
```diff
@@ -0,0 +1,234 @@
+---
+CPS: TBD
```
rphair:
```suggestion
CPS: 26
```

CPS-0026/README.md
rphair: @kozross no better titles came up @ the meeting today so I suppose the last recommendation still stands. 😅

CPS-tfhe/README.md
```diff
@@ -0,0 +1,234 @@
...
+Title: Fully homomorphic encryption with programmable bootstrapping
+Category: Plutus
+Status: Open
+Authors: Koz Ross <koz@mlabs.city>
+Proposed Solutions: []
+Discussions: https://github.com/cardano-foundation/CIPs/pull/1143
```
rphair:
```suggestion
Discussions:
  - Original PR: https://github.com/cardano-foundation/CIPs/pull/1143
```
Standard to use a list here... also following the new mode of labelling coming across the repo through #1125.

rphair (commented):
(continued)

kozross:
@rphair - thanks for that! The 'L2 or not L2' question is precisely the difference between the 'non-native' approach and the other two, so I believe the CPS provides for both possibilities. I happen to agree that 'owning' the capability is better for a lot of reasons, but we expect the necessary research phase before any implementation choices to consider this more thoroughly.

In terms of costing models and estimates - the standard process MLabs have experienced with all of the CIPs proposing builtins was that we specify the overall scaling (for example, 'linear in the first argument'), while the exact model is defined based on work once a PR of the new builtin(s) has been sent to Plutus Core. Obviously, if costing such a builtin is outright impossible, no amount of work on the proposer's part will help. This is one reason I want to hear what Kenneth has to say.

Edit: Updated as per your request.

kozross:
@rphair - maybe this is noted somewhere, so sorry if I'm asking for a reiteration, but what is required for this PR to merge? I guess a better question is 'once a CPS reaches Confirmed state, what needs to happen for it to merge'?

CPS-0026/README.md
rphair: Title has happily stabilised now, so resolving this.

CPS-0026/README.md
```diff
@@ -0,0 +1,235 @@
+---
+CPS: 26
+Title: Fully homomorphic encryption with programmable bootstrapping
+Category: Plutus
+Status: Open
+Authors: Koz Ross <koz@mlabs.city>
+Proposed Solutions: []
+Discussions: 
+    - Original discussion: https://github.com/cardano-foundation/CIPs/pull/1143
```
rphair:
```suggestion
    - Original PR: https://github.com/cardano-foundation/CIPs/pull/1143
```
It's implied that the "original PR" will include a discussion before merging it, so it's more meaningful to state it this way (and it may help [our new validator](https://github.com/cardano-foundation/CIPs/issues/989) if this language can be kept consistent: cc @Ryun1).

CPS-0026/README.md
```diff
@@ -0,0 +1,235 @@
...
+implemented as part of a script, doing so would be prohibitively difficult,
+assuming it is even feasible at all. These difficulties are similar to the
+difficulties of implementing _any_ cryptographic capability as a script, as
+evidenced by [the Grumplestiltskin project][grumplestiltskin], as well as the
+[SHA512][sha512-plutus] and [Ed25519 signature verification][ed25519-plutus]
+benchmarks in Plutus, which is why Cardano scripts have access to various
+cryptographic primitives as builtin operations. Something similar would need to
+exist for TFHE to be usable, but as current, it does not.
+
+## Use cases
```
rphair:
```suggestion
## Use Cases
```
According to the validation standard, main section headings are being treated as titles so all words are capitalised (cc'ing @Ryun1 to double check).

rphair (commented):
@kozross the overall process goes as in [this state diagram](https://github.com/cardano-foundation/CIPs/wiki/301.-State-tagging#state-flow-diagram) (terms defined in the section before it): generally with one biweekly interval required per state change on the shortest path to merge.

As I understand it, the only uncertainty about this CPS subject would apply to whatever CIPs develop from it, rather than the CPS itself: so I'm personally ready to sign off on it after some format nitpicks.  Once the first editor green-ticks it, it can be marked `Last Check` which I think would suit the next CIP meeting in a week's time: https://hackmd.io/@cip-editors/128

For anything else that definitely should be added to the CPS before merge, I would follow @perturbing's technical opinion.  Also I think we should wait until seeing at least a post from @kwxm about the efficiency issues as you yourself have requested.

kozross:
@rphair - thanks for the clarification! I'm certainly in no rush: Kenneth told me recently he won't have time to respond to this for a little while yet, and I'd definitely like to hear from him.

I'll make the requested edits.

Edit: Done.

CPS-0026/README.md
rphair: Fixed in https://github.com/cardano-foundation/CIPs/pull/1143/commits/014f0ccfb62b71bc765f2b800ce0ef37c8fe1ba1.

CPS-0026/README.md
rphair: Fixed in https://github.com/cardano-foundation/CIPs/pull/1143/commits/014f0ccfb62b71bc765f2b800ce0ef37c8fe1ba1.

rphair (commented):
@kozross as per https://github.com/cardano-foundation/CIPs/pull/1146#issuecomment-4322985429: I think this is ready to confirm & move on to `Last Check` (due to response from experts & a stable result) once this editorial issue is fixed:
* set title in CIP document to be the same as the current PR title (and consistently with your other concurrent CPSs): https://github.com/cardano-foundation/CIPs/pull/1143#discussion_r2744616310

kozross:
@rphair - Done.

CPS-0026/README.md
```diff
@@ -0,0 +1,236 @@
+---
+CPS: 26
+Title: Approaches to fully-homomorphic encryption
```
rphair:
```suggestion
Title: Approaches to fully homomorphic encryption 
```
(appears unhyphenated everywhere else in document)

rphair (approved):
Looks good; moving on to final review as per https://github.com/cardano-foundation/CIPs/pull/1143#pullrequestreview-4193600976.

Ryun1 (approved):

PR #1144, opened 2026-01-29
CPS-0027? | Approaches to Post-Quantum Signatures

kozross:
This CPS attempts to take a small bite out of the overall problem of making Cardano post-quantum secure. Specifically, it discusses making post-quantum digital signatures available to script developers by way of one (or more) builtins to verify such signatures.

[Rendered](https://github.com/mlabs-haskell/CIPs/blob/quantum-sig-cps/CPS-0027/README.md)


CPS-?/CPS.md
```diff
@@ -0,0 +1,184 @@
+---
+CPS: TBD
+Title: Post-quantum signatures
+Category: Plutus
+Status: Open
+Authors: Koz Ross <koz@mlabs.city>
+Proposed Solutions: []
+Discussions: https://github.com/cardano-foundation/CIPs/pull/?
```
rphair:
```suggestion
Discussions:
  - https://github.com/cardano-foundation/CIPs/pull/1144
```
`this`

CPS-?/CPS.md
```diff
@@ -0,0 +1,184 @@
...
+Created: 2026-01-39
```
rphair:
```suggestion
Created: 2026-01-29
```

CPS-?/README.md
```diff
@@ -0,0 +1,184 @@
+---
+CPS: TBD
```
rphair: @Ryun1 please post what you think about use of `TBD` as a value here, instead of the `?` called for by CIPs 1 & 9999, as you are going through https://github.com/cardano-foundation/CIPs/issues/989 & related formatting massage... since your validator will run on CIPs & CPSs before merge.

Especially since the `?` would have to be `\` escaped to be valid YAML (and look ugly) I would rather admit other terms that the author wishes to write here as a placeholder... and possibly add `TBD` to a short list of placeholders suggested in CIPs 1 & 9999.

@kozross I would be happy to admit the `TBD` here & please forgive addressing this question in this thread for simplicity's sake.

CPS-0027/README.md
```diff
@@ -0,0 +1,184 @@
...
+that a quantum computer with sufficient resources could break any existing
+cryptographic scheme that relies on integer factorization or discrete logarithms 
+for its hardness. As many currently-used public-key cryptography systems rely on
+the hardness of one of these problems, this is a looming concern. In the world
+of blockchain, concerns over quantum computing are also being taken seriously: 
+the Ethereum Foundation has elevated post-quantum
+security to a [top strategic priority][ethereum-top-priority] this year, and
+Algorand have [been taking proactive steps][algorand-steps] around post-quantum 
+security since at least 2022. In the Cardano ecosystem, The Intersect Product 
+Committee's [Cardano 2030 Strategic Framework][2030-strategic-framework]
```
rphair, line 38: Link needs to be added to endnote section to render properly (currently `2030-strategic-framework` is undefined).

rphair (commented):
Thanks @kozross for reopening this problem space which hasn't seen a word in the CIP repository since:
* https://github.com/cardano-foundation/CIPs/issues/413
* https://github.com/cardano-foundation/CIPs/pull/441

Tagged `Triage` for introduction at next week's Tuesday CIP meeting (authors always encouraged to attend & everyone is invited): https://hackmd.io/@cip-editors/127

CPS-?/README.md
kozross: The main reason I used `TBD` as such is exactly because `?` broke YAML formatting on its own. I'm not attached to it either way.

rphair:
@kozross before further review this file also needs to be renamed from `CPS.md` to `README.md`.

kozross:
@rphair - thanks for the feedback! I've addressed everything, including renaming the file. Also thank you for directing me to the older issue and PR related to this topic: I will have a read of them and see if they bring up anything I should mention.

rphair:
@kozross really looking forward to this PR & CPS bringing together past, present & future work on this 🙏

In that regard I also think this CPS title could be made more "open" or like a survey, rather than something suggesting a document _providing_ the signatures... a bit like I tried to point out here for your other submission: https://github.com/cardano-foundation/CIPs/pull/1143#pullrequestreview-3725313893 ... We can try to brainstorm some titles perhaps better for a CPS than a CIP on the subject, if possible without making the title too wordy, online & at the next CIP meeting.

kozross:
Some feedback on a better title would be great. I have no experience in CPSes, only in CIPs, and would be happy to accept any help there.

CPS-0027/README.md
```diff
@@ -0,0 +1,185 @@
+---
+CPS: TBD
+Title: Post-quantum signatures
```
rphair:
```suggestion
Title: Approaches to Post-Quantum Signatures
```
(other candidates welcome: here & at [upcoming CIP meeting](https://hackmd.io/@cip-editors/127))

@kozross I'm hoping it's OK to drop the word `Digital` here for the sake of generally required brevity because it's understood from the blockchain context (vs. "paper" signatures, "analog" signatures, etc.)... please correct me if wrong.

rphair (commented):
OK @kozross how about to kick-start the discussion we leave a mediocre title in place which invites the reader & reviewer to come up with something better... rather than the current `Post-quantum digital signatures` with the known defect that it suggests the document presents a solution.

A working title `Approaches to ...` would reflect the similar uncertainty to https://github.com/cardano-foundation/CIPs/pull/1143 and perhaps any better choice of words, or means of suggesting possible solutions, would apply to both during the concurrent review of these CPSs that will likely follow.  So I'll set that title on this PR in the meantime and we can continue accumulating feedback here...

perturbing:
@hjeljeli32 , could you be so kind as to review this and share your opinion on such a transition strategy? Loved your presentation in Porto about it; I think your view will be valuable here!

perturbing:
> Some feedback on a better title would be great. I have no experience in CPSes, only in CIPs, and would be happy to accept any help there.

I always approach a CPS with the mentality of a review paper; it tries to get an overview of what is and where things are heading (and how it is blocked).

A good CPS that I really like is [this](https://cips.cardano.org/cps/CPS-0021) one, credits to @nhenin  and @rrtoledo

CPS-0027/README.md
kozross: That seems fine to me. It's definitely clear from the context.

kozross:
@perturbing - that's a great suggestion, thank you! I'll have a read of the provided link.

hjeljeli32:
> @hjeljeli32 , could you be so kind as to review this and share your opinion on such a transition strategy? Loved your presentation in Porto about it; I think your view will be valuable here!

Thanks for tagging me, @perturbing. Much appreciated! I’m really glad to see this problem space being revisited, and excited about the direction toward quantum-ready signatures in Cardano. I’ll review the CPS in detail and share feedback later today.

CPS-0027/README.md
```diff
@@ -0,0 +1,185 @@
...
+that a quantum computer with sufficient resources could break any existing
+cryptographic scheme that relies on integer factorization or discrete logarithms 
+for its hardness. As many currently-used public-key cryptography systems rely on
+the hardness of one of these problems, this is a looming concern. In the world
+of blockchain, concerns over quantum computing are also being taken seriously: 
+the Ethereum Foundation has elevated post-quantum
+security to a [top strategic priority][ethereum-top-priority] this year, and
+Algorand have [been taking proactive steps][algorand-steps] around post-quantum 
+security since at least 2022. In the Cardano ecosystem, The Intersect Product 
+Committee's [Cardano 2030 Strategic Framework][2030-strategic-framework]
```
hjeljeli32, line 38: First of all, this is a really great initiative! Thank you for putting this together! Framing post-quantum readiness early and explicitly is very valuable, and this CPS feels like an important first concrete step in that direction.

One small suggestion for emphasis (not scope expansion): it could be helpful to explicitly mention that script-level post-quantum capabilities are one part of a broader, longer-term effort toward making the whole Cardano stack post-quantum ready. While this CPS is intentionally scoped to scripts and dApps, the choices made here will naturally relate to future work at other layers (ledger, wallets, key management, etc.).

We can show to the community that this CPS is an early building block in a larger ecosystem-wide journey.

CPS-0027/README.md
```diff
@@ -0,0 +1,185 @@
...
+such possibilities exist in a practically usable way.
+
+## Use cases
+
+The chief use case today for such capabilities are script or dApp developers
+whose work currently relies on digital signatures. A good example of this would
+be an auction system, where bids are signed by the bidders. Concretely, consider
+the following validator code snippet:
+
+```haskell
```
hjeljeli32, line 95: This use case is very clear and helpful! One additional aspect that might be worth briefly calling out here is performance and execution cost.

Coming a bit from an EVM background, my instinct is to think in terms of gas costs for signature verification. From my understanding, similar considerations apply to Plutus scripts in terms of execution units and resource usage. This makes builtin support especially valuable, since hand-rolled implementations would likely be much more expensive.

CPS-0027/README.md
```diff
@@ -0,0 +1,185 @@
...
+post-quantum smart-contract-driven wallets can secure Cardano without having to
+modify the ledger directly. While this is very much a worst-case scenario,
+having this possibility available would be an additional layer of defence
+against quantum computing compromising the chain. Post-quantum digital
+signatures would be required to make this possible, and having such builtins
+would allow it to be implemented easily if the need ever arose.
+
+## Goals
+
+Ultimately, the goal would be to make one (or possibly several) post-quantum
```
hjeljeli32, line 153: I really like how this section already highlights that different post-quantum schemes come with different trade-offs. One optional way to strengthen this section could be to group those trade-offs a bit more explicitly from a developer or engineering perspective.

For example, some factors that may matter differently depending on the use case could include:
- signature size (impact on transaction size and data handling),
- verification key size (wallet and storage considerations),
- signature verification cost (script execution resources),
- diversity of cryptographic assumptions (to hedge against future breakthroughs).

This kind of framing might help make it clearer why there may not be a single “best” scheme, and why different applications might reasonably prefer different trade-offs.

CPS-0027/README.md
```diff
@@ -0,0 +1,185 @@
...
+* The existence of a reference implementation that can be integrated into
+  `cardano-base` (a necessary requirement for builtins doing digital
+  signatures).
+* Size of signature and verification key representations.
+* Time required to verify a signature.
+* Existing adoption (including by other chains in case of interoperability
+  requirements in the future).
+
+## Open Questions
+
```
hjeljeli32, line 172: The way the open questions are framed here feels very appropriate for a CPS, especially the acknowledgement that real research and evaluation will be needed before any concrete choice is made.

As an optional clarification, we can mention a few well-known post-quantum signature families or schemes purely as illustrative examples, without implying any preference or decision. For instance, some schemes are often described as more “balanced,” others as having more compact signatures, and others as being more conservative but larger.

Even a high-level mention like this could help readers who are less familiar with the post-quantum landscape better understand what kinds of trade-offs the open questions are referring to.

CPS-0027/README.md
```diff
@@ -0,0 +1,185 @@
...
+evidence for the choice.
+
+A related question to the above is whether it might be worthwhile to have
+_multiple_ schemes available. Currently, there are builtins covering three
+different pre-quantum schemes (although two are variants on the same curve),
+which suggests that doing something similar for post-quantum signature
+verification may be a good idea. This could help 'cover our bases' regarding
+questions of performance and existing adoption, but would magnify all of the
+effort involved in getting usable builtins, starting with the research step and
+ending with costing and implementation. A related concern is that of educating
```
hjeljeli32, line 190: I agree with the point raised here about developer understanding. If multiple schemes are supported, the trade-offs are unlikely to be obvious to most script or dApp developers.

It could be very valuable to provide supporting materials such as simple guidance, example usage patterns, and possibly benchmarking results showing relative verification costs. This would help developers make informed choices without needing deep cryptography expertise.

hjeljeli32 (approved):
This CPS is very well framed and clearly motivated. The direction it takes feels both timely and thoughtful, and it reads as a solid first step toward post-quantum readiness for the Cardano ecosystem. I’ve left a few small, optional suggestions above. Thank you for taking the initiative on this.

CPS-0027/README.md
```diff
@@ -0,0 +1,185 @@
+---
+CPS: TBD
+Title: Post-quantum signatures
+Category: Plutus
```
perturbing, line 4: IMO; we need a CPS that covers and indexes all attack surfaces for when a QC hits the scene, so I do not think it is _just_ a plutus problem. As an overview

- For Praos we have a KES key, which under the hood uses a tree of keys on the ed25519 curve
- The cold key that ultimately binds a stake pool also uses keys from the ed25519 curve.
- The VRF that we use for Praos leadership selection is also using the ed25519 curve
- Then we have in plutus various built-ins that use curves that will break (ecds/schnorr over Secp256k1, and bls pairing crypto).
- the security of our hash functions will also be halved (Grover's algorithm), so we would need to go from sha256 -> sha512 for example, in all places (probably the easiest fix, though block/tx hashes will get bigger)

As you can see, if we get a quantum computer, our consensus and settlement layer will probably be attacked first, so the problem is wider.

But from this overview we also see what kind of quantum tools and construction we will need (not only signature schemes);  we will also need a VRF, KES construction, and most likely some quantum secure ZK scheme. 

A complex problem with many open questions. And from a presentation that @hjeljeli32 gave once, I think the best we can do now, is redesign our code base that we can swap in and out our underlying crypto primitives (which is also probably not trivial). This way, we can adapt ourselves to new schemes that pop up, but research needs to drive this solution space.

perturbing: Given this broader context, another open question is; given that QC has broken the above schemes, how can we ensure that bootstrapping nodes follow the right chain?

E.g, QC breaks the forward security of the KES keys that sign blocks

rrtoledo:
Thanks a lot for the CPS!

PQC is quite a tricky subject as there is not obvious, perfect solution. You detail here the reasons why we need to tackle this issue and finish by saying which digital signatures schemes would be broken and suggesting selecting one to many PQ resistant schemes.

I would go a bit farther, as @hjeljeli32  and @perturbing already mentioned, and add some other problematics such as transitionning, long term attacks, and potential impacts on the blockchain (necessity to have larger blocks, required hardware...). One, if not perhaps the most, contentious point will be the calendar: when should we start using PQC and depreciating current cryptography? Another one would be regulation: is it acceptable to use exotic cryptography or should we focus on standardized schemes (e.g. NIST candidates such as Crystal-Dilithium).

You are focusing in this CPS on digital signatures, but we may be using other assymetric cryptography algorithms, or require specific properties, that would be broken, and potentially whose PQ resistant solutions are not designed yet (VRF, multi-signatures, VDF perhaps...).  I am not sure however if having one generic CPS for all PQ questions would better than distinct ones for each subject.

perturbing:
> I am not sure however if having one generic CPS for all PQ questions would better than distinct ones for each subject.

I think a CPS for the initial problem space is wanted. Having a clear overview of what needs to be replaced and what constraints are current there (like time/space wise) are already of big value!

CPS-0027/README.md
kozross: That was very much my intent, so I'm happy to adjust the text as you suggested.

CPS-0027/README.md
kozross: Yes, 100%. I linked the examples of Grumplestiltskin and the two Plutus benchmarks (all of which were designed by MLabs) to show just _how_ painful anything cryptographical onchain can be if implemented 'from the primitives'. Even with CIP 121, 122 and 123! Perhaps I should make it a bit clearer.

CPS-0027/README.md
kozross: I believe I already do this, just below in fact. Do you feel it should be re-worded?

CPS-0027/README.md
kozross: I agree with that. I have a few specific ones in mind, and will add their descriptions.

CPS-0027/README.md
kozross: I don't disagree with any of this - in fact, I agree very strongly! However, the sheer magnitude of this problem means that no single CPS will cover it reasonably. In fact, the [last one that tried](https://github.com/cardano-foundation/CIPs/issues/413) got _severely_ bogged down in details and ended up not progressing as a result.

The reasons behind why we (meaning 'MLabs') wanted to focus on a specific case of post-quantum security was to prevent exactly this kind of scope blowout and death by details. Specifically, we tried to make for a CPS such that:

* It addresses post-quantum security with regard to Cardano somehow;
* The scope is well-defined, allowing a CIP or at least an action plan without death by detail;
* Any potential solution could be used to inform further work, whether it's by choice of implementation, research required, resources expended, etc.

It's not that we feel this is necessarily the only, or even the most important, task or set of questions, merely the one we feel most able to tackle in future work (and want feedback on), and the one residing in the area we understand best (as MLabs are responsible for the implementation of quite a few builtins, including CIPs and working with Plutus Core maintainers).

Does this help answer your questions?

kozross:
@rrtoledo - again, I disagree with none of this. The problem space is enormous, and potentially affects a great many things. However, we at MLabs feel that writing one huge CPS that tries to address everything will end up overly general and just get bogged down.

You can check my reply to @perturbing for further thoughts on this subject.

CPS-0027/README.md
kozross: Yes, but it is inherently tricky to do. Learning resources on Cardano, and Plutus, are spread quite widely, and on top of that, choosing between cryptographic primitives to perform a given task is often quite tricky. 

From my perspective, I feel that the research step is quite important here in terms of scheme choice. If the step turns up a single choice that happens to be 'the best', this problem vanishes, as there will be 'one right way to do it'. If not, there will be reasons for this, which should be used to document why you might want which scheme.

kozross:
@hjeljeli32 - I have addressed your feedback, and left some follow-up questions as well. Thanks for the review!

CPS-?/README.md
```diff
@@ -0,0 +1,248 @@
...
```
rphair:
```suggestion
CPS: 27
```

CPS-?/README.md
```diff
@@ -0,0 +1,248 @@
...
+Title: Post-quantum signatures
+Category: Plutus
+Status: Open
+Authors: Koz Ross <koz@mlabs.city>
+Proposed Solutions: []
+Discussions: https://github.com/cardano-foundation/CIPs/pull/1144
```
rphair:
```suggestion
Discussions:
  - Original PR: https://github.com/cardano-foundation/CIPs/pull/1144
```
Standard to use a list here... also following the new mode of labelling coming across the repo through https://github.com/cardano-foundation/CIPs/pull/1125.

rphair (commented):
@kozross this was considered a well researched production at the CIP meeting today and continued the immediate interest from online review.  This was a packed meeting and conversation didn't turn to signature algorithms or families so I hope can tag others who can provide review suitable for the long-term consideration of this CPS.

The technical reservations were somewhat the same as #1143 in the questions of costing.  As in your other PR I don't think costing is so much a CPS as a CIP requirement (cc @perturbing @colll78) and likely not as much an issue for signatures as for the concurrent homomorphic problem.

If you do think that costing is an issue that CIPs will face, perhaps you can do some preliminary research about the actual costing beyond just relative comparisons of computational expense.  In the meantime I hope general Plutus reviewers would be interested in commenting on this (I don't have a corresponding list of cryptography reviewers): cc @zliu41 @effectfully @kwxm @MicroProofs

Please change the containing directory to `CPS-0027` and update the "rendered" link in your original posting 🎉

CPS-0027/README.md
hjeljeli32: Thanks @kozross , that makes sense. I agree the criteria already capture the key trade-offs well. I don’t think any rewording is strictly necessary; my comment was mostly about framing rather than adding new requirements.

CPS-0027/README.md
hjeljeli32: @kozross That makes sense, thanks for clarifying. I agree that a research-first step can simplify things a lot, and that documentation naturally follows from the outcomes of that process.

kozross:
@rphair - Thanks for this! Costing this should be pretty easy in principle, as it'll follow the same model(s) as used by existing builtins. The challenge here is more minimizing what this cost is, both in terms of the builtin to verify the signed message, but also the size(s) of the verification key and message we would need.

Edit: Made your requested changes.

CPS-0027/README.md
kozross: OK, great!

CPS-0027/README.md
kozross: Yes. If anything, the outcomes of the research step should be _part_ of the documentation that follows.

CPS-0027/README.md
```diff
@@ -0,0 +1,249 @@
...
+
+## Copyright
+
+This CPS is licensed under [Apache-2.0](http://www.apache.org/licenses/LICENSE-2.0).
+
+[quantum-computing]: https://en.wikipedia.org/wiki/Quantum_computing
+[shors-algorithm]: https://en.wikipedia.org/wiki/Shor%27s_algorithm
+[post-quantum-cryptography]: https://en.wikipedia.org/wiki/Post-quantum_cryptography
+[ethereum-top-priority]: https://thequantuminsider.com/2026/01/26/ethereum-foundation-elevates-post-quantum-security-to-top-strategic-priority/
+[algorand-steps]: https://thequantuminsider.com/2026/01/26/ethereum-foundation-elevates-post-quantum-security-to-top-strategic-priority/
```
rphair: This creates the same document link for Algorand as for Ethereum: probably accidental? (because the Ethereum article doesn't address the Algorand effort)

rphair (commented):
@kozross as per https://github.com/cardano-foundation/CIPs/pull/1146#issuecomment-4322985429: I think this is ready to confirm & move on to `Last Check` (due to response from experts & a stable result) once this editorial issue is fixed (plus more below):
* set title in CIP document to be the same as the current PR title (and consistently with your other concurrent CPSs): https://github.com/cardano-foundation/CIPs/pull/1144#discussion_r2744271333

@mgajda you are credited with some earlier work on this subject so your comment on the CPS before merging would also be welcome.

CPS-0027/README.md
kozross: Yeah, that's an error. Will fix.

mgajda:
> @mgajda you are credited with some earlier work on this subject so your comment on the CPS before merging would also be welcome.

@rphair @kozross I support the call for a single builtin parameterized with signature algorithm. This is the best way forward when we expect algorithms to become broken and obsoleted with progress. It will be useful whether we have post-quantum algorithms from current NIST call, or even better ones.

kozross:
@mgajda - I have no objections to this per se.

kozross:
@rphair - everything should be fixed now.

CPS-0027/README.md
rphair: fixed with https://github.com/cardano-foundation/CIPs/pull/1144/changes/dbfd2518ef6530ab583f9998ff9016d2e57687d3

rphair (approved):
Looks good after https://github.com/cardano-foundation/CIPs/pull/1144/changes/dbfd2518ef6530ab583f9998ff9016d2e57687d3 & marking `Last Check` as per https://github.com/cardano-foundation/CIPs/pull/1144#pullrequestreview-4193623851.

Ryun1 (approved):

Issue #1138, opened 2026-01-20
Enhance Plutus CIP Category definitions


Ryun1:
based on discussion from CIP editors call # 126
Discussion arose from PR #1134, which was adding more precise definitions to a proposed Plutus builtin.
These further definitions were required as during implementation development time was wasted pursuing an implementation not matching the the author's original intent.

In the future it would be good to ensure all Plutus Builtin proposals have unambiguous definitions, ensuring implementors are able to match author's intensions.

Perhaps we can update [CIP-0035 | Plutus Core Evolution](https://github.com/cardano-foundation/CIPs/tree/master/CIP-0035) to include guidelines on such definitions?

rphair:
An example of what we would be looking for, as I recall @colll78 saying in the #1134 discussion today:

When different means of implementing a function are possible, order-of-magnitude estimates should also be included for possible implementations to help choose an implementation method.  This would have avoided the apparently incorrect choice which had to be retracted with #1134.

As @Ryun1 agreed, our goal is to have a list of criteria to add to CIP-0035 that helps the community pre-empt issues like this.

rphair:
I've retitled this to match what I believe @colll78 was explaining at that meeting.  We have a current need to keep the issue queue clear of the word "Category" since it draws attention away from our 2 new CIP category CIPs (`Governance` & `Node`) and I think this title is more specific about what we are looking for.

rphair:
@kwxm - in hindsight this issue was delayed due to my error in not tagging you here (cc @perturbing @Ryun1); likely because when I realised your input would be welcome on the https://github.com/cardano-foundation/CIPs/issues/1138#issuecomment-3775601258 @colll78 originally made, CIP review was currently focused on this very different aspect of performance:
* https://github.com/cardano-foundation/CIPs/pull/1146#issuecomment-3858149465

After a CIP implementation difficulty earlier (full story above), we were looking for a way that order-of-magnitude expectations can be requested for Plutus builtins via [CIP-0035](https://github.com/cardano-foundation/CIPs/blob/master/CIP-0035/README.md), however it might fit in with the Plutus team's profiling and benchmarking:
* If you could produce a PR to add this language to CIP-0035 it would be ideal.
* If not, you could write a summary and the editors will prepare a CIP-0035 update PR accordingly & circulate it for approval.
* If you or @zliu41 @SeungheonOh think it's not practical or possible to prepare such a requirement then please say so and we'll close this.

PR #1146, opened 2026-02-05
CPS-0029? | Approaches to Higher-Order Costing

kozross:
Something of a follow-up to #1143 . Given that costing higher-order (in fact or in deed) builtins is a concern that goes beyond the specific case described in #1143 , I believe it is worth both describing the status quo, and discussing what would need to improve and how.

I would like @kwxm to chime in on this in particular: I have tried my best to describe things accurately, but I may have missed something.

[Rendered](https://github.com/mlabs-haskell/CIPs/blob/higher-order-costing/CPS-0029/README.md)

### DIFF
--- a/CPS-hoc/README.md
+++ b/CPS-hoc/README.md
@@ -93,9 +93,9 @@ introduced three builtin operations:
 * `ListToArray`, which converts a builtin list into a builtin array with the
   same contents in the same order.
 
-The intent behind CIP-138 arrays is to act as a 'transient' type within scripts:
-builtin arrays have no `Data` encoding that doesn't involve going through a
-builtin list. This creates two significant limitations:
+Currently, CIP-138 arrays lack a `Data` encoding, and no existing ledger type
+makes use of them. Thus, they currently act as a 'transient' type within
+scripts. However, there are two significant limitations to this use case:
 
 * There is no way to programmatically create a builtin array without first going
   through a builtin list; and
@@ -107,7 +107,14 @@ Both of these mean that any non-trivial computation involving arrays onchain
 requires a large number of intermediate lists. Not only do these take up onchain
 resources that shouldn't really have to be spent, given that they require a lot
 of 'hand rolling', they increase the possibility of errors (such as the infamous
-'off by one').
+'off by one'). 
+
+These issues become an even bigger problem if CIP-138 arrays were to get a
+native `Data` encoding, or if their use in ledger types were to increase. Unless
+scripts only need to check lengths or access elements, working with arrays
+provided by datums or ledger types would be much less convenient than for lists,
+which would paradoxically make such improvements far less useful than if they
+kept using lists.
 
 One solution would be to provide a builtin similar to the following:
 
--- a/CPS-hoc/README.md
+++ b/CPS-hoc/README.md
@@ -6,7 +6,7 @@ Status: Open
 Authors: Koz Ross <koz@mlabs.city>
 Proposed Solutions: []
 Discussions:
-    - https://github.com/cardano-foundation/CIPs/pull/?
+    - Original pull request: https://github.com/cardano-foundation/CIPs/pull/1146
 Created: 2026-02-04
 License: Apache-2.0
 ---
--- a/CPS-hoc/README.md
+++ b/CPS-hoc/README.md
@@ -1,0 +1,282 @@
+---
+CPS: TBD
+Title: Approaches to Higher-Order Costing
+Category: Plutus
+Status: Open
+Authors: Koz Ross <koz@mlabs.city>
+Proposed Solutions: []
+Discussions:
+    - https://github.com/cardano-foundation/CIPs/pull/?
+Created: 2026-02-04
+License: Apache-2.0
+---
+
+## Abstract
+
+Currently, costing of builtins makes most higher-order builtins (namely those
+that take function arguments) impossible to add to Plutus Core. Such builtins
+could provide significant performance and functionality improvements, and there
+are no semantic reasons that we could not implement such builtins. Thus, an
+expansion of the costing of higher-order builtins would allow significant
+improvements in terms of expressiveness and performance.
+
+## Problem
+
+[Costing][costing], or the Plutus cost model, is an important feature of Cardano
+scripts and their execution. This assigns, to every script, a cost in both
+execution units (corresponding to time used) and memory. As part of this, every
+Plutus builtin operation is given a cost based on the arguments it is given: the
+exact method of calculation used for current builtins is given
+[here][plutus-cost-model-doc]. Determining how this cost is calculated is an
+important requirement for any new builtin to be added to Plutus Core.
+
+Currently, costs are calculated for builtins on the basis of either the _sizes_
+of arguments, or their _value_, the second of which being used only for integers
+and bytestring lengths. For the builtins that currently exist, this works fine,
+as they are all essentially _first-order_ (meaning, none of them take function
+arguments). However, if we ever wanted to add [higher-order][higher-order] builtins 
+(specifically those that takes function arguments), this approach becomes
+inadequate. More specifically:
+
+* A function's size is not a meaningful indicator of its evaluation costs. Even
+  if we have access to the function's code, a small function could require a lot
+  of resources to evaluate (for example, it may involve a fixed point finding
+  operation).
+* A function argument may be called an arbitrary number of times by a
+  higher-order builtin. While this may be predictable for some builtins based on
+  other arguments given to them, there is no guarantee this can be done in
+  general.
+
+Thus, higher-order builtins cannot, in general, be costed at the moment. There
+is one notable exception to this case, which used to exist in the `CaseList` and
+`CaseData` builtins, with parentheses added for clarity:
+
+```
+CaseList :: b -> (a -> [a] -> b) -> ([a] -> b)
+
+CaseData :: (Integer -> [Data] -> b) -> ([(Data, Data)] -> b) -> ([Data] -> b)
+-> (Integer -> b) -> (ByteString -> b) -> (Data -> b)
+```
+
+We can see that both `CaseList` and `CaseData` are actually 'doubly
+higher-order' in their intent: not only do they _take_ functions as arguments,
+they also return a function. These could be costed for the following reasons:
+
+* None of the function arguments are ever called: we instead construct a
+  'dispatch function'. This calls the appropriate function argument based on its
+  (first-order) argument, and this _can_ be costed as a regular lambda.
+* The only thing the 'dispatch function' ends up doing is a pattern match on a
+  data type, followed by a dispatch call, which introduces only a fixed and
+  known amount of overhead.
+
+While these builtins no longer exist, they represent the maximum capability for
+higher-order builtin costing available to us currently.
+
+## Use cases
+
+Higher-order functions are ubiquitous in functional programming, and thus, use
+cases abound. While UPLC (and any higher-level language targeting it) gives us
+the ability to define many higher-order functions, in some cases, this isn't
+efficient, or even sufficient. To demonstrate cases where higher-order builtins
+would give us significant benefits, we give a few examples below.
+
+### Builtin arrays and `GenerateArray`
+
+[CIP-138][cip-138] introduced builtin arrays to UPLC, and has been implemented
+into Plutus Core. In particular, in addition to the new array type, CIP-138
+introduced three builtin operations:
+
+* `LengthOfArray`, which returns the number of elements in its argument as an
+  `Integer`;
+* `IndexArray`, which given an array and an `Integer` index, produces the
+  element at that index in constant time; and
+* `ListToArray`, which converts a builtin list into a builtin array with the
+  same contents in the same order.
+
+The intent behind CIP-138 arrays is to act as a 'transient' type within scripts:
+builtin arrays have no `Data` encoding that doesn't involve going through a
+builtin list. This creates two significant limitations:
+
+* There is no way to programmatically create a builtin array without first going
+  through a builtin list; and
+* Given an existing array, there is no way to produce a modified version. The
+  only option is to first convert to a builtin list (using a hand-rolled loop
+  over indexes), then convert to an array again.
+
+Both of these mean that any non-trivial computation involving arrays onchain
+requires a large number of intermediate lists. Not only do these take up onchain
+resources that shouldn't really have to be spent, given that they require a lot
+of 'hand rolling', they increase the possibility of errors (such as the infamous
+'off by one').
+
+One solution would be to provide a builtin similar to the following:
+
+```
+GenerateArray :: Integer -> (Integer -> a) -> Array a
+```
+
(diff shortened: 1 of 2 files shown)


rphair:
@kozross + @kwxm & other reviewers, FYI  — although this may be about a different aspect of "costing" — we noted in an earlier CIP meeting that a recently added (or updated) builtin proposed by @colll78 (#1134) was implemented with a higher-order execution time than would have been possible if all alternatives were included in the CIP (or its update) so that an optimal implementation could be chosen.

The CIP editors aren't able to progress this problem on our own & have just recorded it here for now:
* https://github.com/cardano-foundation/CIPs/issues/1138

We'll keep that issue & this PR as separate as they need to be: although some commenters on this PR may wish to contribute to that issue.  Our goal for that issue would be a CIP-0035 update about how to document order-of-magnitude cost estimates as necessary to be beneficial to Plutus implementors (cc @Ryun1 @perturbing).

CPS-hoc/README.md
```diff
@@ -0,0 +1,286 @@
...
+
+* `LengthOfArray`, which returns the number of elements in its argument as an
+  `Integer`;
+* `IndexArray`, which given an array and an `Integer` index, produces the
+  element at that index in constant time; and
+* `ListToArray`, which converts a builtin list into a builtin array with the
+  same contents in the same order.
+
+The intent behind CIP-138 arrays is to act as a 'transient' type within scripts:
+builtin arrays have no `Data` encoding that doesn't involve going through a
```
ana-pantilie: `Data` will be extended to include `Array`, and it's possible some parts of the script context which are currently encoded as lists will be encoded as arrays in future language versions. Therefore, a large part this section's argument will be outdated in the future.

kozross: This is good to know, and I will amend the section in light of this.

colll78:
Particularly interesting to me is limited higher order costing approaches (as opposed to higher order costing which is a whole different beast of vastly higher complexity) I wonder if that warrants a different CPS or CIPs to limited higher order costing can also be proposed under this CPS.

kozross:
@colll78 - what do you see as the difference between 'limited higher-order costing' and 'higher-order costing'? My CPS is partly a call toward figuring out what limits to higher-order costing exist, as well as informing the community of what those are, and why. 

It'd be nice if, as a natural consequence of this, we'd also get costing capabilities _up to_ those limits, of course. These capabilities will be 'limited' in some sense almost by definition. Hence, I'm unsure what you mean.

colll78:
limited higher order costing refers to the costing of limited higher order builtins, which are higher order builtin function which can only apply compositions of existing builtin functions instead of any arbitrary UPLC code. 

ie. 
```haskell
`builtinMap :: (IsBuiltinComposition (a -> b)) => BuiltinList a -> (a -> b) -> BuiltinList b
```
In the above `(a -> b)` must be a composition of builtin functions ie. `(head . tail . tail . snd . unConstrData)`

kozross:
@colll78 - this is an interesting design space which hadn't occurred to me! One major concern I'd have is that a lot of function arguments to something like 'mapping' operations want matching, which isn't available by means of just builtins, or at least not efficiently. For example, matching on builtin `Bool` is significantly faster than using `IfThenElse`, and if you need both head _and_ tail of a known non-empty list, using a one-arm match is better than the combination of `HeadList` and `TailList`. 

Furthermore, composition can be a bit awkward, unless you mean 'composition' modulo arbitrary permutation of arguments? This is veering fairly close to my suggestion of 'no applications except to builtins', just without matching.

colll78:
> @colll78 - this is an interesting design space which hadn't occurred to me! One major concern I'd have is that a lot of function arguments to something like 'mapping' operations want matching, which isn't available by means of just builtins, or at least not efficiently. For example, matching on builtin `Bool` is significantly faster than using `IfThenElse`, and if you need both head _and_ tail of a known non-empty list, using a one-arm match is better than the combination of `HeadList` and `TailList`.
> 
> Furthermore, composition can be a bit awkward, unless you mean 'composition' modulo arbitrary permutation of arguments? This is veering fairly close to my suggestion of 'no applications except to builtins', just without matching.

It is awkward, but it is also significantly easier to cost, and it allows us to have phased release schedule where we can nail higher-order builtins with real use-case feedback and developer input instead of trying to do it perfectly in one go and missing the mark. I think even limited higher order builtins with no modulo arbitrary permutation of arguments will satisfy the vast majority of possible use-cases, certainly long enough to give us time to handle costing of true higher order built-ins with lots of care. The examples of why this would introduce some inefficiency compared to your proposed higher-order builtins are true but are complete micro-optimizations in the scope of what this would provide, limited higher order builtins would reduce the ex-cost of most common operations by 90%+, and the same with builtin matching would reduce these marginally more at the expense of a significantly longer costing / delivery timeline. 

I may be misevaluating the difficulty of costing this variant vs the variant with argument application and matching, and if that is the case then all the merrier, if the scope is feasible for your proposed version, then I'm happy.

CPS-hoc/README.md
```diff
@@ -0,0 +1,293 @@
+---
+CPS: TBD
+Title: Approaches to Higher-Order Costing
+Category: Plutus
+Status: Open
+Authors: Koz Ross <koz@mlabs.city>
+Proposed Solutions: []
+Discussions:
+    - https://github.com/cardano-foundation/CIPs/pull/?
```
Ryun1:
```suggestion
    - Original pull request: https://github.com/cardano-foundation/CIPs/pull/1146
```

rphair:
@kozross your newer pair of CPS PRs were triaged at the CIP meeting today and the consensus is that @kwxm's first order response, as you have requested, would be a simple condition for this CPS to be declared a candidate (though we could surely consider alternate sources of confirmation).  If he or another expert reviewer for this particular subject were to confirm feasibility of defining the problem space in this way, it would certainly be sufficient.

Another note: @kevinhammond [said in the meeting](https://discord.com/channels/971785110770831360/971785111030870087/1473349940867760320) he has a strong interest in this subject and has written a few papers on it.  I'll try to ping him somehow & invite him to the discussion.

kevinhammond:
Yes, I'll take a look.  I've published a POPL paper for example (that is dealing with very general HOF costing) and designed languages for more restricted settings - @kwxm was part of that project

kevinhammond:
Restricting to second-order functions might be sufficient?  That covers a lot of real-world uses.  Also you seem to be mainly interested in builtin HOFs (which can have specific cost models) rather than user-definable functions?  That's definitely easier to deal with.

For example

`map myfun [1..10]`

versus

`map map`

or

`myMap myfun [1..10]`

or similar.  The work that Martin Hofmann and Steffen Jost did exploited linearity, but the analysis was quite complicated IIRC. It was restricted to a strict setting but that would be OK for UPLC

kevinhammond:
In terms of the cost model, you would need to reflect the base costs into a cost function, and apply these in the right place.  If you are restricting to second order functions, that should be easier.

So

`mapArray f`
 
should give a cost model such as
 
`\ a -> length a * cost_f`

which could then be applied in a concrete application of `mapArray` provided the length of the argument is known at compile time.  It gets harder if you need to construct the cost model from first principles, e.g.

`cost_mapArray = \a -> 1 * cost_f + cost_mapArray (next a) + cost_next`

Transforming the latter to the former used to require an external solver in the cost analysis.  You might be able to build a simple solver within the cost model instead, or perhaps use type-level operations.

One complication with costing HOFs is that you generally can't tell a-priori whether a higher-order function can be costed.   It's necessary to try it and see whether the solver accepts it.  Sometimes, counter-intuitively, making a function more complex can make it easier to cost (the function becomes restricted).

kevinhammond:
Final thoughts: I don't know how the UPLC cost analysis is constructed, but this could well involve a major rewrite, of course, if e.g. there now need to be cost-level lambdas.  That could be a fun project.  There could also be increases in compile time.  I don't know how significant that would be (there's already some metaprogramming in the UPLC compiler, AFAIK, so maybe performance isn't a major issue).

kozross:
@kevinhammond - thank you for the insights! I agree that a restriction to second-order builtins would allow most of what people would want from 'higher order builtins'. Furthermore, we suggested two restrictions to the kinds of function arguments such a higher-order builtin would accept:

* My writeup suggests forbidding applications of anything that isn't a builtin; and
* @colll78 suggested forbidding anything that _isn't_ an application of a builtin.

I'm not sure if this makes the task easier or more difficult. 

Would you be able to give me a link to your paper? It might help if we're on the same page.

CPS-hoc/README.md
kozross: Thanks for catching that, done.

kevinhammond:
There is another approach that might work: defunctionalisation transforms HOFs into first order functions.  So you could write a defunctionalising Plutus-Plutus transpiler (or Aiken-Aiken transpiler etc).  As long as the first order functions are standard Plutus (Aiken etc), then you don't need a new cost analysis

https://en.wikipedia.org/wiki/Defunctionalization

kevinhammond:
@kozross Here's the POPL paper, but that really deals with lazy evaluation as well as HOFs.  It describes a type-level analysis.

https://dl.acm.org/doi/10.1145/1706299.1706327

This is open access, but the work is technically pretty complex (it took 4-5 years and about 20 failed attempts before we could work it out!).  It relies on an external solver (we used one that could handle convex hulls).


This was the paper that deal with higher-order functions in a strict setting, IIRC (might be behind a paywall - let me know if you can't get it easily - there may be a preprint on Arxiv for example).

https://link.springer.com/chapter/10.1007/978-3-642-05089-3_23


This is the paper by Martin Hofmann and Steffen Jost - first order, heap allocation and linear but describes the basic principles quite nicely

https://dl.acm.org/doi/10.1145/640128.604148

rphair (commented):
@kozross I think this ended up in a stall because were were expecting some broad review like we had in https://github.com/cardano-foundation/CIPs/pull/1150 (leading to its belated confirmation today) but we wouldn't want to discriminate against this CPS just because the subject is more esoteric.

Given that we might have to relax our criteria in that way, I think @kevinhammond's own comments above serve as confirmation that this problem statement is of general interest and usefully formulated; so putting it on the agenda here to confirm it as a numbered CPS if editors & potential reviewers agree: https://hackmd.io/@cip-editors/132

@kozross we would still hope for more participants in the discussion (please tag & invite out of band if you can) and we might also assign a CPS number early if @Ryun1 @perturbing agree before that meeting.

kozross:
@rphair - I believe I've already tagged and out-of-banded everyone I can. I'd still love to hear from @kwxm, primarily as a lot of the techniques in @kevinhammond's linked papers are quite involved. A sense of 'what would be seen as acceptable by the Plutus Core team' in this context would help future CIPs addressing this problem not get bogged down.

rphair:
@kozross after over 5 years of trying to gauge what's expected of the CIP process, I believe authors & editors are only expected to show "due diligence" (rather than a tangible result) in getting responses from those considered to be stakeholders.  It sounds like that's been done in this case so I would be happy to confirm & then merge this — and recommend this to other editors — in due course (i.e. the next 2 CIP meetings: since a quorum of editors is required for both).

CPS-hoc/README.md
```diff
@@ -0,0 +1,293 @@
+---
+CPS: TBD
```
rphair:
```suggestion
CPS: 29
```

rphair (commented):
@kozross - CIP meeting discussion today confirmed my previous comment about when this would be ready to merge.  Lack of dispute or alternative recommendations should still keep this moving along & in the meantime please change the containing directory to `CPS-0029` and update the "Rendered" link in your OP 🎉

CPS-hoc/README.md
kozross: Done.

kozross:
@rphair - just did this now. Thanks!

rphair (approved):
@kozross for reasons I've already mentioned it seems like your statement of this problem is not going to be disputed but also that we might wait indefinitely for a confirmation: which is arguably grounds to keep this moving forward.

I believe this makes sense (without any subject matter expertise of my own) because your document is readable and sensibly ordered in its explanations: so at the very least it will be a useful reference as more contributors become involved in this subject.

So please first post here if it's OK to mark this `Last Check` to see if other CIP editors agree about merging this at the next CIP meeting: https://hackmd.io/@cip-editors/133 — and also @colll78 @kevinhammond @ana-pantilie please post if there are any hard objections or changes you'd like to recommend first.

Also @kozross we may similarly need your advice on whether you feel these concurrent CPSs have had enough review & are ready to move forward for the same reason (i.e. neither dispute nor confirmation from a very small set of subject matter experts):
* https://github.com/cardano-foundation/CIPs/pull/1143
* https://github.com/cardano-foundation/CIPs/pull/1144
* https://github.com/cardano-foundation/CIPs/pull/1150

kozross:
@rphair - I feel that `Last Check` is appropriate for this CPS. Until we hear from @kwxm, there isn't much more to add. While between myself, @kevinhammond and @colll78, a lot of possibilities for higher-order costing have been offered, without feedback from the Plutus project itself (in terms of what would be seen as acceptable or useful), there's not much we can do from here.

In terms of the rest:

* TFHE (homomorphic encryption) can definitely move forward. It's been largely set for a while.
* Post-quantum signatures, likewise, should be good to move forward with.
* Approaches to call-by-need require me to address some feedback from @zliu41 , but once that's been done, I think we can also proceed.

zliu41 (approved):
This is indeed an open question, and one obviously worth pursuing further.

While possible solutions are out of the scope of the CPS, one possibility that comes to mind is to not try to cost the entire execution of the higher-order builtin at once. Instead, when the function argument is called, somehow track its cost separately.

kozross:
@zliu41 - I feel a really important question here is this: what tradeoff between 'being able to cost things' and 'costing complexity' would the Plutus project be OK with? @kevinhammond's linked papers provide descriptions of schemes able to cost _significantly_ more complex cases than what either I or @colll78 suggested. However, these require non-trivial machinery, which would have to be implemented (and debugged!) by someone familiar enough with the background of the papers, which is a big ask, and might not be something the Plutus team are interested in.

While obviously @kwxm is the expert, given your position, out of the options listed in this discussion, which one(s) do you feel are the most likely to be accepted or considered as viable?

zliu41:
I haven't given any detailed thought to the solution. I simply think that the problem is a meaningful one, so the CPS should be accepted, since a CPS isn't required to propose solutions.

The first possible solution that came to my mind, which is what I mentioned above, probably requires defunctionalization, so it may be the same thing as what Kevin suggested. I think this option should at least be seriously considered before settling for a simpler but more restrictive solution. My sense is that it's feasible.

rphair:
thanks @kozross & since the comments from @zliu41 seem about the already documented open question(s) then we'll consider (with @perturbing) merging this at the meeting today (i.e. now) with perhaps an impromptu green-check from editors present (or soon after): [https://hackmd.io/@cip-editors/~~133~~**134**](https://hackmd.io/@cip-editors/134)

(will update other review threads accordingly as per your https://github.com/cardano-foundation/CIPs/pull/1146#issuecomment-4322985429)

Ryun1 (approved):

PR #1164, opened 2026-03-19
CIP-0182? | Optimistic Constitutionality

ptrdsh:
This CIP introduces **Optimistic Constitutionality**, changing how governance actions are approved under CIP-1694.

# Overview
Learning from the last nearly 100 Governance Actions onchain, no proposal that DReps wanted approved got rejected by the CC. The CCs effort however was insurmountable despite this perfect agreement. 

It turned out that DReps are excellent in filtering wrongful/bad proposals naturally by voting on their personal preference, regardless of constitutionality. This fact can be used to the network's advantage and reduce the cost of constitutional review drastically by borrowing from Zero Knowledge applications, where Optimistic Rollups (TRUE until someone says otherwise) is an effective expensive-verification-mitigation technique.

Optimistic Constitutionality shifts the CC from constant involvement to on-demand oversight, dramatically reducing workload and, among others, providing constructive aid to the ongoing debates around CC compensation.

# Changes proposed

Instead of requiring the Constitutional Committee (CC) to actively approve every action, proposals are assumed constitutional by default and move forward unless one CC member raises an objection by voting "No"/"Unconstitutional" - a "constitutional challenge" or veto. If such a challenge occurs, the original CIP-1694 approval process resumes, requiring full CC voting and the other CC members then either confirm the veto-ing CC's challenge by voting no/unconstitutional as well, or overrule it by voting yes/constitutional with the defined majority threshold. 

The required change is happening in the ledger logic for ratification, which is changed to: 

```math
$$\text{RATIFIED} \iff \left(V_{\text{CC}}^{\text{no}} = 0 \;\lor\; \frac{V_{\text{CC}}^{\text{yes}}}{|CC|} \geq T_{\text{CC}}\right) \wedge \left(\frac{S_{\text{DRep}}^{\text{yes}}}{S_{\text{DRep}}^{\text{active}}} \geq T_{\text{DRep}}\right) \wedge \left(\frac{S_{\text{SPO}}^{\text{yes}}}{S_{\text{SPO}}^{\text{delegated}}} \geq T_{\text{SPO}}\right)$$
```
`RATIFIED = (zero CC no votes exist **OR** (CC Yes votes ≥ CC threshold)) **AND** (DRep Yes stake ≥ DRep threshold) **AND** (SPO Yes stake ≥ SPO threshold)`

And, the introduction of a new  protocol parameter `ccOptimisticEnabled :: Bool` to control whether Optimistic Constitutionality is active or not, is helpful. 

# Path to Active 
This CIP requires additional code in the ledger, and can only be activated with a hardfork. Due to this, it will be beneficial to let the ecosystem vote on this via an info action, as this change will be packaged within a fork containing countless changes.  

> Possible Action NOW:
In order to not wait for the hard fork after the next hard fork to activate this and alleviate the CCs current pain, there is nice "soft activation" path forward, which could double as testing/confirmation period:
An info action, in which the CC announces it will from now insta-vote yes on everything by default without commentary, and only vote no once they are called or see issues themselves.

# RISKS
- **sleeping at the wheel**
Given that CC members are, and in fact must, be active community members, "sleeping at the wheel" risks are diminishingly low, as CC members (are assumed) to already follow Cardano Governance naturally anyway. This is additionally mitigated by now massively increased redundancy: Only 1 of (currently) 7 members must vote no. 

- **power abuse**
CC members could single-handedly block all approved, near-ratified Governance Actions by voting No/ calling out a "constitutional challenge" up until the last block of an epoch. This risk is real, and currently known possible mitigations for it, such as introducing approval window extensions are only increasing complexity a lot, introduce other issues, and overall provide no actual benefit. 
I postulate that the best mitigation for such power abuse would be to acknowledge the risk and rely on state of no confidence posteriori. 

# OPEN QUESTIONS
- I am not a ledger developer, so the currently contained suggested code changes under "Ledger Rule Changes" critically need to be reviewed. I used Opus 4.6 to generate them.

---

([rendered latest version from branch](https://github.com/ptrdsh/CIPs/blob/patch-2/CIP-0182/README.md))


ptrdsh:
An X user commented: 
> There is also a 3rd risk which is a hybrid of the 2 you have mentioned. If a constitutional issue is realised too late there won’t be time to stop the GA.


This is my response, and with it, i make a note to myself to need to incorporate such arguments in the CIP, and underline them with quantitative figures, sources where possible. 


> Fortunately, by now (after ~100 GAs) we know that its not realistic that CCs, DReps and all other community members are sleeping at the wheel simultaneously. "Given enough eyeballs, all bugs are shallow"-kinda thing happening in realtime. 
> Ask any GA proposer, and you'll hear that they are being "made aware" of constitutional "issues" every. single. time. - not by the CC and in nearly all cases unfounded, but the eyeballs are there. 
> Read enough DRep rationales, and you'll see that many quote issues with the constitution and reject the proposal because of it. 
> Its not a valid argument, as AIs cant replace humans in that matter, but loads of people use AI to assess/monitor constitutionality for their own additional personal opinion-building (also publicly available here [http://proposalexaminer.cardanofoundation.org](https://t.co/i3Usc2hAUB) btw). Sure, those AIs wont spot intricate detailed violations, as we would hope the CC to be able to spot, but it underlines that "is it constitutional" isnt isolated to the CC at all - automatically. 
> 
> Point being: 
> There is plenty of evidence that constitutionality review is more-dimensional than just the handful of people who are overburdened today to formalize a solid review, which in the majority of proposals wouldnt be required at all anyway, since the majority of proposals are constitutional.
> 
> OC would lift some/most of that burden, thanks to the ecosystem's collective review powers, which they will do in any case. Collective Intelligence at play and people cant turn off seeing and calling out issues. Its human. 
> 
> In the end, i think optimistic constitutionality even enhances the quality of the CC and the protection they provides, as it will free them to focus on being the last line of defense, not formality warriors.

CIP-Optimistic-Constitutionality/readme.md
```diff
@@ -0,0 +1,269 @@
+---
+CIP: XXXX
+Title: Optimistic Constitutionality
+Category: Ledger, Governance
```
rphair:
```suggestion
Category: Ledger
```
We don't actually have a `Governance` category yet and might not even have one (due to attempted submissions having generally been "social" or organisational rather than technical) and we would have to define it first to prevent it from being overrun with social / organisational proposals.

In any case there is only a single category assigned (also an ongoing vocabulary question, settled on a single classification since there hasn't been enough community interest expressed in "tags" that would extend the field(s) of interest somehow) and in any case this is abundantly concerned with Ledger for both review and implementation.

CIP-Optimistic-Constitutionality/readme.md
```diff
@@ -0,0 +1,269 @@
...
+
+Authors:
+  - Alex Moser <alexander.moser@cardanofoundation.org>
+    
+Implementors:
+  - N/A
+    
+Discussions:
+  - https://github.com/cardano-foundation/CIPs/pull/XXXX
+
```
rphair:
```suggestion
Authors:
  - Alex Moser <alexander.moser@cardanofoundation.org>
Implementors:
  - N/A
Discussions:
  - https://github.com/cardano-foundation/CIPs/pull/XXXX
```
Seems like blank lines entered here to suggest extensibility of these lists (?)... but since no other CIP has them, there would be some reader disorientation & perhaps parsing problems upstream.

CIP-Optimistic-Constitutionality/readme.md
```diff
@@ -0,0 +1,269 @@
...
+This change reduces the CC's operational burden from active interaction with every governance action to reactive oversight "on demand". The approach aligns with the empirical reality that DReps already filter harmful proposals effectively through their votes of preference. As a second-order effect, Optimistic Constitutionality substantially changes the cost of constitutionality verification for the network, and provides relief to the contentious question of CC compensation, which could still be enabled separately - but under different assumptions for effort spent. 
+
+## Motivation
```
rphair:
```suggestion
## Motivation: Why is this CIP necessary?
```
Extended title is part of our CIP template (awaiting formalisation by @Ryun1 to update CIP-0001 and the template files) and other CIPs have been modified (waiting in the wings in related PRs plus a validator script) to match.  1 of 2 such changes...

CIP-Optimistic-Constitutionality/readme.md
```diff
@@ -0,0 +1,269 @@
...
+### Interaction with CC Expiry and No-Confidence
+
+- **Expired CC members**: An expired CC member's prior votes (including `No` votes) are discarded at expiry, consistent with current behavior. An action that was challenged only by a now-expired member reverts to unchallenged status.
+- **No-confidence state**: When the CC is in a state of no-confidence, Optimistic Constitutionality does not apply — no governance actions can be ratified in this state, consistent with CIP-1694.
+
+### Versioning
+
+This specification is versioned by the CIP document itself. Changes to the specification require an amendment to this CIP or a superseding CIP.
+
+## Rationale
```
rphair:
```suggestion
## Rationale: How does this CIP achieve its goals?
```
... change 2 of 2.

CIP-Optimistic-Constitutionality/readme.md
```diff
@@ -0,0 +1,269 @@
...
+
+### Backward Compatibility
+
+This proposal changes the semantics of CC non-voting from "implicit No" to "implicit approval." This is a **breaking change** to the ratification logic and requires a hard fork.
+
+Governance actions submitted before the activation epoch follow the pre-existing rules. Actions whose voting period spans the activation boundary should be evaluated under the new rules from the activation epoch onward.
+
+All existing CC credentials, hot/cold key infrastructure (CIP-0105), governance metadata standards (CIP-0100, CIP-0108, CIP-0136), and wallet integrations (CIP-0030, CIP-0095) remain compatible. The change is entirely within the ledger's `RATIFY` rule and does not affect transaction formats, certificate types, or on-chain data structures.
+
+## Remaining Risk
```
rphair:
```suggestion
### Remaining Risk
```
Though it may have been intended for this to be a major section, we would blow our CIP template if we allowed authors to do this for emphasis (which may also have been intended).  However, the title itself is clear and its position within the _Rationale_ (since related to implementation choices: something _not_ to try to fix) is logical, as is the subsection below it.

CIP-0182/README.md
```diff
@@ -0,0 +1,269 @@
...
+5. The local-state-query protocol exposes challenge status for governance actions.
+
+### Info Action
+A community vote is most likely beneficial or even required for such a change, as it cannot individually be voted on within the hardfork GA, with which it would get activated. 
+
+### Implementation Plan
+
+This CIP requires a **hard fork** to activate, as it changes the ledger's ratification semantics.
+
+#### Phase 1: Specification & Review
```
rphair, line 252: If the "phases" in this section could be converted into a nested bulleted lists, it would be much easier for reviewers, editors, and implementors to work with.

CIP-Optimistic-Constitutionality/readme.md
```diff
@@ -0,0 +1,269 @@
...
+### Acceptance Criteria
+
+1. The formal Conway ledger specification (`cardano-ledger`) is updated to reflect the modified `RATIFY` rule as described in this CIP.
+2. The implementation is deployed on the `preview` testnet and exercised with both challenged and unchallenged governance actions.
+3. The implementation is deployed on the `preprod` testnet with a functioning Constitutional Committee.
+4. A hard fork governance action activating the new ledger era is ratified and enacted on Cardano mainnet.
+5. The local-state-query protocol exposes challenge status for governance actions.
+
+### Info Action
+A community vote is most likely beneficial or even required for such a change, as it cannot individually be voted on within the hardfork GA, with which it would get activated.
```
rphair: As with the adjacent item, if the list above could be tick-boxed as it many other CIPs (YMMV [by browser, I hear] for getting this to work with ordered lists) then it could be visualised & processed more easily.  Then this particular item can be added to it (maybe nested as a prerequisite of one of the main acceptance criteria).

CIP-Optimistic-Constitutionality/readme.md
```diff
@@ -0,0 +1,269 @@
...
+- Deployment on `preview` testnet.
+- Integration testing with governance tooling (GovTool, cardano-cli, Eternl, Lace).
+- Deployment on `preprod` testnet.
+
+#### Phase 3: Mainnet Activation
+
+- Hard fork governance action submitted, requiring approval from the CC, DReps, and SPOs under the *current* (pre-Optimistic) rules.
+- Upon ratification and enactment at the subsequent epoch boundary, Optimistic Constitutionality becomes active with `ccOptimisticEnabled :: Bool` set to `TRUE`.
+
+#### Optional: Off-Chain Convention (Pre-Hard-Fork)
```
rphair: Since the goal of these is to keep the _Path to Active_ lean, if this is a complex consideration it can be defined in one of the earlier sections and then the "decision" & potential task(s) entered into one of the concise lists in this section.

What the editors / CIP process are trying to avoid are new ideas about the CIP substance being introduced at the "last minute" in the final part of the document.

rphair (commented):
Thanks @ptrdsh - this looks like an intelligent & beneficial submission.  I understand it will be admissible regardless of implications to "social governance" (see #937) because it's simply a Ledger change.  On for `Triage` at the next CIP meeting — https://hackmd.io/@cip-editors/131 — and just some formatting considerations before general review begins here:

fallen-icarus:
Should a minority of dReps and/or SPOs be able to flag a proposal as *potentially* unconstitutional? If they do, the CC would be required to weigh in to over-rule their concern. The flagging thresholds could be ~30%.

Thomas-nada:
To mitigate the risk of a single CC member sabotaging a proposal at the final block of an epoch, I propose the following modification to the ratification logic:

#### **The Challenge Window ($X$ Epochs)**
A **Constitutionality Challenge** can only be triggered during the first $X$ epochs of a Governance Action’s lifetime. 
* **Inside the Window:** If a CC member votes `No`, the "Optimistic" presumption is voided. The action immediately transitions to **Challenged** status.
* **Outside the Window:** A CC member can still manually change their vote to `No` at any time for the remainder of the action's life (for transparency/dissent), but doing so **will not** trigger a challenge if one hasn't already been started.

CIP-Optimistic-Constitutionality/readme.md
```diff
@@ -0,0 +1,269 @@
...
+Discussions:
+  - https://github.com/cardano-foundation/CIPs/pull/XXXX
+    
+Created: 2026-03-18
+License: CC-BY-4.0
+---
+
+## Abstract
+
+This CIP proposes **Optimistic Constitutionality**, a modification to the governance ratification rules introduced by CIP-1694. Under the current model, every governance action requires the Constitutional Committee (CC) to reach a positive vote threshold before it can be ratified. This proposal inverts that requirement: governance actions are **presumed constitutional by default** and proceed to ratification without dedicated CC approval, unless one or more CC members explicitly file an objection (a "constitutionality challenge"). When a CC member votes 'No' (when such a "constitutionality challenge" is triggered), the current model for CC approval, as originally designed by CIP-1694, re-activates and all other CC members should vote too, to either support the "constitutionality challenge" or overwrite it with approvals.
```
thenic95:
```suggestion
This CIP proposes **Optimistic Constitutionality**, a modification to the governance ratification rules introduced by CIP-1694. Under the current model, all governance actions besides the "Motion of no-confidence," "Update committee/threshold," and "Info Action" require the Constitutional Committee (CC) to meet a positive vote threshold before they can be ratified. This proposal inverts that requirement: governance actions are **presumed constitutional by default** and proceed to ratification without dedicated CC approval, unless one or more CC members explicitly file an objection (a "constitutionality challenge"). When a CC member votes 'No' (when such a "constitutionality challenge" is triggered), the current model for CC approval, as originally designed by CIP-1694, re-activates and all other CC members should vote too, to either support the "constitutionality challenge" or overwrite it with approvals.
```

The ledger does not allow CC members to vote on the "Motion of no-confidence" and "Update committee/threshold" actions and "Info Action" cannot be enacted onchain.

CIP-0182/README.md
```diff
@@ -0,0 +1,269 @@
...
+
+This change reduces the CC's operational burden from active interaction with every governance action to reactive oversight "on demand". The approach aligns with the empirical reality that DReps already filter harmful proposals effectively through their votes of preference. As a second-order effect, Optimistic Constitutionality substantially changes the cost of constitutionality verification for the network, and provides relief to the contentious question of CC compensation, which could still be enabled separately - but under different assumptions for effort spent. 
+
+## Motivation
+
+**A structural observation:**
+
+Across the first ~100 governance actions on mainnet, no proposal was rejected by the CC where DReps would otherwise have approved it. Every proposal the CC found unconstitutional was independently rejected by DRep vote. DReps, acting in rational self-interest to protect their stake and the infrastructure their businesses rely on, have proven effective at identifying bad proposals without CC guidance.
```
thenic95, line 25: The argument could also be made that the DReps are closely monitoring the constitutionality assessments by the CC and naturally would not approve of an action that fails the constitutionality assessment.

CIP-0182/README.md
```diff
@@ -0,0 +1,269 @@
...
+2. **DRep discouragement.** When CC rejections arrive late, DReps who already spent time evaluating proposals see their work invalidated. Over time, this suppresses DRep participation.
+3. **False urgency.** The requirement that CC members evaluate every proposal creates a baseline workload that scales linearly with governance activity, regardless of whether the proposals are contentious.
+4. **Compensation pressure.** The active-approval model makes the "CC members must be compensated" argument seem self-evident, because their workload is indeed real and ongoing. Regardless, since the activation of Conway-era governance, the community has been unable to reach consensus on whether and how Constitutional Committee members should be compensated, despite mentions of this in (all versions so far of the) constitution. Optimistic Constitutionality aids to dissolve this by dramatically reducing the expected workload, and introduces new, more transparent possibilities for CC compensation, should that remain a desire.
```
thenic95, line 36: > Optimistic Constitutionality aids to dissolve this by dramatically reducing the expected workload

I think the 'reduced workload' is an illusion, as the responsibility for 'catching' unconstitutional governance actions still rests with the Constitutional Committee (CC). CC members will likely continue to review every single action to ensure nothing slips through the cracks; otherwise, they risk being held liable if the treasury is raided. Relying on 'Optimistic' oversight risks a dangerous diffusion of responsibility, where every member assumes someone else is doing the vetting, leaving the door wide open for malicious actors.

Furthermore, I don't think this approach accounts for specific attack vectors, such as draining the treasury or implementing malicious parameter changes. For example, someone could submit a governance action in the final slots of an epoch and collude with the top 10 DReps to vote "yes." If no CC member catches this mischievous behavior in time, the action could meet the ratification threshold without the CC being able to stop it.

In my view, the risk is simply too high to consider this approach.

CIP-Optimistic-Constitutionality/readme.md
ptrdsh: Nice idea. fully agreed

CIP-Optimistic-Constitutionality/readme.md
ptrdsh: Understood and Agreed!

CIP-0182/README.md
ptrdsh: Thank you, I agree. That section is anyway rather weak and would need some edits. Will change the formatting while doing so.

CIP-Optimistic-Constitutionality/readme.md
ptrdsh: Yes you're fully right. Thanks for spotting and flagging! Will change

rphair:
@ptrdsh looks like your last commit https://github.com/cardano-foundation/CIPs/pull/1164/commits/0e38ae349c17c167286c43079fdddbedec04ed11 takes care of at least half the review points in https://github.com/cardano-foundation/CIPs/pull/1164#pullrequestreview-3976475992, still visible above.  These would be resolved & abbreviated automatically if committing through the GitHub UI: which is generally easier (only requiring a `git pull` back into the local fork) & shows useful evidence of peer review... as well as credit which is sometimes significant.

So if you need to make these changes outside the UI (and for the corrections above), can you "Resolve" each of the conversations to help us keep track of the review points that remain? 🙏

thenic95:
To really simplify this proposal, it essentially changes how the ledger treats non-voting CC members on active governance actions.

Instead of treating non-voting from CC members as a NO, this change defaults the CC to a YES, as long as no CC member actively casts a NO vote. If even one NO vote is cast, the ledger reverts to the current rules, where not voting is treated as a NO.

I see two main risks:

1. Governance Action proposers could submit GAs close to the end of an epoch and collude with DReps to push a change through before any CC member notices it.
2. CC members hold an ultimate veto and could abuse their power by casting a NO vote right before an action would be approved. For example, let's assume a parameter change only obtains the necessary DRep or SPO votes in the last epoch of its lifetime. Because no CC member has voted yet, everyone assumes the action will pass. Then, in the final slots of the epoch, a single CC member votes NO to effectively veto the action, since the other CC members cannot quickly enough cast their YES votes. (Shout out to Megan for highlighting this!)

These are the main attack vectors I currently see.

CIP-Optimistic-Constitutionality/readme.md
```diff
@@ -0,0 +1,269 @@
+---
+CIP: XXXX
```
rphair:
```suggestion
CIP: 182
```

rphair (commented):
@ptrdsh we agreed at the CIP meeting today that this should be confirmed a CIP candidate because of the quality & positive direction of the comments made by reviewers so far: on both the technical proposal as well as any implications for the overall governance process.

Next we would need you to respond to all the review points above before progressing this (marking the resolved things `Resolved` with a note about any change that was made).  The comments you've already responded affirmatively to would make a great first update & when editors see that's happened we can pass this around for a second review & maybe put it explicitly on the CIP agenda then.

In the meantime please rename the containing directory to `CIP-0182` and update the "latest version" link in the original post accordingly 🎉

CIP-Optimistic-Constitutionality/readme.md
rphair: (no contest to the long tradition of having only a single category for each CIP)

CIP-0182/README.md
rphair: @ptrdsh I've committed the formatting changes so editors can focus on the rewrite issues... just please `git pull` these in to your working branch so a further `git push` (without force) will keep everything well documented here 🤓

brouwerQ:
I don't think challenging the constitutionality should be limited to the first X epochs. I would just prolong the gov action's lifetime by 1 epoch if the first NO vote is casted during the last epoch. This is somewhat similar to what we have now if a new constitutions gets enacted in the last epoch of a gov action. If 1 epochs is too short, it could be 2.

Also, while still in optimistic mode (with no legacy CC threshold met), the gov action should only be able to get ratified if it's live for at least 2 of 3 epochs or so, so CC has enough time to look at a gov action and can't be overtaken by DReps voting early. Either we need a new parameter for this, or we keep it simply and take half of the current Gov Action Validity parameter. Earlier gov action ratification is still possible if the required number of YES votes according to legacy rules are casted.

CIP-0182/README.md
ptrdsh: Yep, could be made, but from memory, it feels like DRep voting trends already approached the outcome before any CC assessments, as they vote rather late.

CIP-0182/README.md
ptrdsh: reduced workload: i believe optimistic constitutionality could have not been implemented from the start, as the CC and ecosystem at large would have for sure ran into the issue you are describing, and noone would have been educated about best practices on how to be constitutional. I also believe we are allowed to grow out of this phase and assume a certain level of maturity. The risk of reliance on others is mentioned in the risk section "sleeping at the wheel", and is real, but acknowledgeable, imo, given the status quo. 


Collusion: ill add that as new risk. Definitely an attack vector that needs to be considered.

CIP-0182/README.md
ptrdsh: lets see if i do this right :D

ptrdsh:
@brouwerQ I added a new parameter to protect against early ratification, also based on many other suggestions received. 

@fallen-icarus i really like this idea, but its a LARGE change, which requires implementation work across all layers, and re-educating the community. I believe its out of scope, and for what it "costs", its not worth the mitigation it provides to the risks of this CIP. 

@Thomas-nada agreed. and mentioned in the CIP as risk, but a state of no confidence should be enough safety net, imo. 

@thenic95 thank you for all your valuable feedback and direct contributions! much appreciated <3 

@rphair thank you for all your editing and feedback! much appreciated as well <3

brouwerQ:
> @brouwerQ I added a new parameter to protect against early ratification, also based on many other suggestions received.

I do think you should still allow for a gov action to pass earlier than this IF the threshold has been met by explicit YES votes though, as this is the case in the current system too.

rphair:
@ptrdsh @brouwerQ we have been waiting for comments for a few weeks on this: so as soon as you agree about the latest nuance, editors might (depending on the confidence expressed here by both of you) either move this on to `Last Check` (if highly confident) or put it explicitly on the next meeting's `Review` agenda (if not).

rphair:
p.s. @ptrdsh this is not fit for `Last Check` until the containing directory is changed from `CIP-Optimistic-Constitutionality` to `CIP-0182`.  Leaving this undone (after the [original request](https://github.com/cardano-foundation/CIPs/pull/1164#pullrequestreview-4041446823), which always comes when a CIP is confirmed) is the most common cause of "face-palm" by editors, so please take care of this immediately. 😅

CIP-0182/README.md
```diff
@@ -0,0 +1,282 @@
+---
+CIP: 182
+Title: Optimistic Constitutionality
+Category: Ledger
+Status: Proposed
+Authors:
+  - Alex Moser <alexander.moser@cardanofoundation.org>
+Implementors:
+  - N/A
```
rphair:
```suggestion
Implementors: N/A
```
We want a null set of implementors (can also put `[]` instead of `N/A`): rather than a null set _as_ an implementor 🤓

CIP-0182/README.md
```diff
@@ -0,0 +1,282 @@
...
+Discussions:
+  - https://github.com/cardano-foundation/CIPs/pull/XXXX
```
rphair:
```suggestion
Discussions:
  - https://github.com/cardano-foundation/CIPs/pull/1164
```
`this`

rphair (commented):
thanks @ptrdsh — some some bit(s) more before I approve & mark `Last Check` (if no objection in the meantime)...

rphair (dismissed):

CIP-0182/README.md
```diff
@@ -0,0 +1,281 @@
...
+4. **Compensation pressure.** The active-approval model makes the "CC members must be compensated" argument seem self-evident, because their workload is indeed real and ongoing. Regardless, since the activation of Conway-era governance, the community has been unable to reach consensus on whether and how Constitutional Committee members should be compensated, despite mentions of this in (all versions so far of the) constitution. Optimistic Constitutionality aids to dissolve this by dramatically reducing the expected workload, and introduces new, more transparent possibilities for CC compensation, should that remain a desire.
+5. **CC discouragement.** The CC is already overburdened by its current workload, and rejections of compensation doesnt improve the situation. New CC members are discouraged to put themselves up for vote. 
+
+## Specification
+
+### Overview
+
+The core change is to the ratification logic in the Conway ledger rules. `TODO: Describe the changes required for the extenstion of the live epochs for CCs to react.`
```
Ryun1, line 43: please addresses TODO :)

kenerik:
This proposal risks eroding the 'checks and balances' essential to our governance. The CC's current role is to provide proactive assurance, not just a final line of defense. By assuming constitutionality by default, we eliminate the necessary deliberation window that serves as a vital security layer. I believe this move compromises the institutional integrity that CIP-1694 was designed to protect.

rphair (commented):
@ptrdsh - the explicit reason for not merging this at the CIP meeting today was the still `TODO` status of one of your proposal points that @Ryun1 thankfully flagged here: https://github.com/cardano-foundation/CIPs/pull/1164#discussion_r3382344234

By chance this also creates an opportunity for both observers & editors to express their reservations about this proposal:
* there is no prohibition of CIPs characterising unpopular ideas — so I won't imply that this can't be merged if people object to the apparent "things are constitutional by default" interpretation (cc @Crypto2099); but
* the editors _would_ demand that these objections be addressed in the proposal somehow: likely elaborating your table _Addressing Community Concerns_ into a more detailed document subsection with further subheadings for each point.

Once this is enough to satisfy everyone, please notify editors if we appear to miss that any objections have at least gone on record: so we might tag this `Last Check` again.  **p.s.** we are expecting some critical reviews of this proposal to come in based on comments at the meeting today.

rphair (commented):
@ptrdsh I had approved this as well on track to merge before, but have noted (after some time originally waiting for an update from you) that the still-`TODO` item noted in https://github.com/cardano-foundation/CIPs/pull/1164#discussion_r3382344234 contains a _missing part of the Ledger rules_.
* This in turn creates a contingency because we can't ask @lehins @WhatisRT to review this — essential for a proposal with Ledger impact — while the Ledger-facing specification is still incomplete.

This idea seems unpopular but editorially we wouldn't disqualify it for that alone.  The editors' main reservation would be against merging a half-completed or incompletely-reviewed proposal & so we have to tag this `Waiting for Author` until we have some progress toward:
* _either_ replacing the `TODO` sentence with the missing part of the spec _or_ removing that sentence if in fact this is already done;
* some response to a Ledger review about the possibility of making this change: without necessarily addressing any other criticism that the change might not be desirable.

@thenic95 @daniellestanko-gov since approval from Ledger is orthogonal to review of Governance impact, I still hope you & any other governance advocates besides @kenerik (noting his objection https://github.com/cardano-foundation/CIPs/pull/1164#issuecomment-4662140981) will now have some time to also post a review based on impact to the governance process.

ptrdsh:
@rphair ToDos resolved, and I made some edits to reflect the concerns of commenters. Thanks for bumping me. 

@kenerik DReps proved perfectly capable of providing checks and balances, and as evident from various other cases, the honourable intention of CIP-1694 for the CC to be a vital security layer is overstated. Further, the CC still has the deliberation window just the same as now, OC just substantially lowers the effort to act, helping CCs do their job better - not not all.

CIP-0182/README.md
```diff
@@ -0,0 +1,286 @@
...
+[ ] - 1. The formal Conway ledger specification (`cardano-ledger`) is updated to reflect the modified `RATIFY` rule as described in this CIP.
+[ ] - 2. The implementation is deployed on the `preview` testnet and exercised with both challenged and unchallenged governance actions.
+[ ] - 3. The implementation is deployed on the `preprod` testnet with a functioning Constitutional Committee.
+[ ] - 4. A hard fork governance action activating the new ledger era is ratified and enacted on Cardano mainnet.
+[ ] - 5. The local-state-query protocol exposes challenge status for governance actions.
+
+### Acceptance Criteria 
+(Off-chain)
+
+[ ] - 1. one or more CC members adopt the model of Optimistic Constitutionality on their own terms, enforced by either promise, or smart contracts.
```
rphair:
```suggestion
## Path to Active

### Acceptance Criteria 

(On-chain)
- [ ] 0. A community vote is most likely beneficial or even required for such a change, as it cannot individually be voted on within the hardfork GA, with which it would get activated. 
- [ ] 1. The formal Conway ledger specification (`cardano-ledger`) is updated to reflect the modified `RATIFY` rule as described in this CIP.
- [ ] 2. The implementation is deployed on the `preview` testnet and exercised with both challenged and unchallenged governance actions.
- [ ] 3. The implementation is deployed on the `preprod` testnet with a functioning Constitutional Committee.
- [ ] 4. A hard fork governance action activating the new ledger era is ratified and enacted on Cardano mainnet.
- [ ] 5. The local-state-query protocol exposes challenge status for governance actions.

(Off-chain)
- [ ] One or more CC members adopt the model of Optimistic Constitutionality on their own terms, enforced by either promise, or smart contracts. 
```
You have probably figured out that natively ordered lists are incompatible with tickboxes.  However: the tickbox needs to be **after** the bullet character to render correctly ([currently not even rendered as lists](https://github.com/ptrdsh/CIPs/blob/5639fd9a030dfeb4cd6cf08cb561d1ec4d18594d/CIP-0182/README.md#acceptance-criteria)).

Also we may have validation problems (and probably _should_ have a validation problem... @Ryun1?), if not reader confusion, if the duplicate subheading __Acceptance Criteria__ is not resolved into one.

rphair (commented):
@ptrdsh thanks for cleaning this up.  To me it seems logical in every way, and the points upon which other governance advocates would disagree are beyond my experience: although I would interject (as I already said above) that this can still be merged as a proper CIP for current & future consideration even if the overall sentiment is against implementing this.

@lehins (cc @carlostome @williamdemeo) can you please check the feasibility of the [Ledger Rule Changes](https://github.com/ptrdsh/CIPs/blob/patch-2/CIP-0182/README.md#ledger-rule-changes)?  I would approve this myself if this would create no Ledger problem or conflict (pending the format fixes below).

Also @ptrdsh I still don't know who to ask or tag about the [proposed protocol parameter changes](https://github.com/ptrdsh/CIPs/blob/patch-2/CIP-0182/README.md#new-governance-protocol-parameter-optional) so perhaps you would tag anyone who can give an assessment about that.

ptrdsh:
@rphair thanks for the formatting fixes. 
Regarding the protocol parameters.. I have posted this CIP in the TSC and PC with the intention to get feedback, specifically on the protocol parameters, I think since none of them are essential, a deepdive wouldnt be necessary at this stage. One is fully optional (its just a toggle for OC on or off), and the other is a(one) mitigation for a risk that Nico flagged.

thenic95:
I checked the on-chain record myself: across 121 ratifiable governance actions, I could not find a single case where the CC was the sole binding blocker on something DReps and SPOs had approved. The closest candidate, the 2025 liquidity budget withdrawal, had unanimous CC rejection but sat at 63.6% DRep approval against a 67% threshold. It would have failed anyway. So yes, the CC has not overturned the community's will.

That finding is the reason I oppose this proposal.

If the CC has never changed an outcome, then this CIP is a hard fork that changes no outcomes. It only changes workload. That is a poor trade, because the two failure modes are not symmetric. Under active approval, the failure mode is a false negative: a legitimate proposal stalls, the proposer resubmits, and the cost is time. Under optimistic constitutionality, the failure mode is a false positive: an unconstitutional action ratifies because seven CC members all happened to look away at the same time. For a treasury withdrawal, that is not recoverable. You do not accept an irreversible tail risk to save reviewer hours.

My second concern is precedent. This changes the constitution from a precondition into an exception. Nothing in the ledger rules would stop that, but the argument it enables is obvious: once CC members demonstrably do not need to interact with ~80% of actions, someone will argue the role itself is redundant and should be removed. I would expect that argument to come from this direction, and I think it belongs in the Unintended Consequences section rather than being left implicit. Motivation # 4 makes this harder to dismiss, because it frames the mechanism partly as a way to weaken the case for CC compensation. That is designing a constitutional safeguard around a budget dispute. Compensation should be settled on its own merits.

I also want to flag an assumption. DReps assessing constitutionality is an emergent practice, not a designed redundancy. It is not written into anyone's mandate and it can decay, particularly if the protocol itself signals that constitutionality is now optional. Hardening a ledger rule around a norm that could reverse is fragile.

The problems here split into two, and both have better solutions.

Wasting DReps' time. This is the legitimate one, and the data supports it: 43.4% of all DRep votes are cast before the first CC vote on that action even exists, and the median first CC vote arrives 2.21 epochs into a roughly six-epoch window. So sequence the vote instead of removing the requirement. Let the CC vote first as an approval gate, then open DRep and SPO voting. That fixes the actual harm, which is DReps doing full reviews on proposals ruled unconstitutional at the last minute. It creates urgency for the CC rather than removing their obligation. And it preserves the default.

Compensation justification. Decide it directly. Do not engineer the workload away to win the argument.

Finally, the CIP already proposes an off-chain social convention as a pre-hard-fork step. Do that first. Have the CC adopt optimistic behavior voluntarily for a defined period, publish the results, and see whether the workload saving and the risk profile look the way this proposal predicts. A convention can be withdrawn in a week. A hard fork cannot.

rphair:
@ptrdsh my apologies for not reapplying the `Confirmed` tag when this emerged from `Waiting for Author` — likely the consequence of a dodgy mouse (remediating now & double checking everything).

To discuss how to resolve @thenic95's https://github.com/cardano-foundation/CIPs/pull/1164#issuecomment-5620639479 we could review this live at a near-term CIP meeting: please let me know & I'll schedule it, although it could be better to log the feedback that comes from that same debate in governance circles.  (FYI the CIP meeting 5 days from now already has a packed agenda so it would not be the day for an unhurried discussion about it.)

CIP-0182/README.md
```diff
@@ -0,0 +1,284 @@
...
+
+Across the first ~100 governance actions on mainnet, no proposal was rejected by the CC where DReps would otherwise have approved it. Every proposal the CC found unconstitutional was independently rejected by DRep vote. DReps, acting in rational self-interest to protect their stake and the infrastructure their businesses rely on, have proven effective at identifying bad proposals without CC guidance.
+
+This observation is the empirical foundation for Optimistic Constitutionality: if the CC almost never needs to override DRep consensus, why require the CC to actively approve every action?
+
+### The Current Model's Inefficiencies
+
+Under CIP-1694, the CC must reach its approval threshold for every governance action type that requires CC consent (all types except Info actions and No Confidence motions against themselves). This creates several problems:
+
+1. **Late voting.** CC members empirically tend to vote in the final 20% of a governance action's lifetime, after DReps have already cast votes. This wastes DRep effort on proposals that the CC ultimately rejects.
```
lehins, line 33: This CIP does not address this point. In fact it makes things even worse.
With this CIP CC members will always rely on DReps voting first, before any one of them will chime in, because this CIP removes responsibility to review proposals. Casting a vote is not the hard part, it's only a matter of submission a single transaction. Reviewing proposals and making decision on how vote is the CC members' job.

CIP-0182/README.md
```diff
@@ -0,0 +1,284 @@
...
+2. **DRep discouragement.** When CC rejections arrive late, DReps who already spent time evaluating proposals see their work invalidated. Over time, this suppresses DRep participation.
```
lehins, line 34: With this CIP, CC rejections will still arrive late, probably even later.

CIP-0182/README.md
```diff
@@ -0,0 +1,284 @@
...
+3. **False urgency.** The requirement that CC members evaluate every proposal creates a baseline workload that scales linearly with governance activity, regardless of whether the proposals are contentious.
```
lehins, line 35: I believe this is the reason why CC members vote late, because they want to see if there is interest in the community on the proposal before they waste their time reviewing a proposal. Which is totally sensible. So, I don't believe there is this "False urgency". Most importantly, this CIP make this so much worse, because CC members will not have any urgency to evaluate proposals, until one CC member votes "No", at which point every CC member will be under real sense of urgency!

CIP-0182/README.md
```diff
@@ -0,0 +1,284 @@
...
+4. **Compensation pressure.** The active-approval model makes the "CC members must be compensated" argument seem self-evident, because their workload is indeed real and ongoing. Regardless, since the activation of Conway-era governance, the community has been unable to reach consensus on whether and how Constitutional Committee members should be compensated, despite mentions of this in (all versions so far of the) constitution. Optimistic Constitutionality aids to dissolve this by dramatically reducing the expected workload, and introduces new, more transparent possibilities for CC compensation, should that remain a desire.
```
lehins, line 36: > The active-approval model makes the "CC members must be compensated" argument seem self-evident, because their workload is indeed real and ongoing. 

Same argument can be made for DReps. So, if we apply this "optimistic" solution to DReps as well, then what is the point of having governance?

CIP-0182/README.md
```diff
@@ -0,0 +1,284 @@
...
+5. **CC discouragement.** The CC is already overburdened by its current workload, and rejections of compensation doesnt improve the situation. New CC members are discouraged to put themselves up for vote.
```
lehins, line 37: Presumably CC members have another stream if income that stems from Cardano. So, it is in their interest to make sure sustainability of Cardano.
If CC members do not have any income from Cardano, then they must be doing it for fun, because I can't see any other reasons why someone would volunteer to become a member of any committee.

CIP-0182/README.md
```diff
@@ -0,0 +1,284 @@
...
+Thus, ratification of GAs within its first epoch should become disallowed, which creates a minimum life time for all Governance Actions of 2 epochs. 
+
+Further, every CC can single-handedly block and drop Governance Actions just before ratification of a DREP-,and SPO-approved Governance Action by voting `No` at the last moment, voiding the effort of the ecosystem to get to an approved state. 
+Thus, a CC `No` Vote of an GovAction that achieved its threshold and would transition at the epoch boundary from approved to `in ratification`, should stay in its state to allow for the other CC members to confirm/overwrite the No vote that put the GA in status `challenged` at the last moment.
+
+An alternative, easier mitigation for such power abuse would be to just acknowledge this power abuse risk and solely rely on submitting a state of no confidence, as the CCs block could be seen a malicious action, of which the voters could be the judge.
```
lehins, line 67: How do you distinguish such action between malicious and a legitimate last minute vote? If a CC member is in danger of being viewed as malicious then such last minute vote will not happen, despite that it might be necessary.
So, this is not a reasonable alternative, in my opinion.

CIP-0182/README.md
```diff
@@ -0,0 +1,284 @@
...
+- If **zero** CC members have cast a `No` vote, the CC approval condition is satisfied automatically.
+- CC members MAY still vote `Yes` or `Abstain` on unchallenged actions, but these votes are not required for ratification.
+- The DRep and SPO thresholds remain unchanged and must still be met for their respective Governance Action types.
+
+#### Challenged Actions (one or more CC `No` votes)
+
+- If **one or more** CC members cast a `No` vote for their respective allowed Governance Action types, the governance action transitions to "challenged" status. 
+- Once challenged, the **existing CIP-1694 CC threshold logic applies in full**: the action requires CC Yes votes ≥ CC threshold to satisfy the CC approval condition.
+- The threshold is evaluated as today: each current committee member has one vote, and the threshold is the governance parameter `committeeMinSize` / `committeeThreshold`.
+- The DRep and SPO thresholds remain unchanged.
```
lehins, line 91: All of the current and future CC members will need to be 100% onboard with this change to governance, because if any of the CC members disagrees with this CIP, then they can deliberately always vote No first, which will take us back to the system we have today, but puts a negative view on all proposals, because a CC member voted No.

CIP-0182/README.md
```diff
@@ -0,0 +1,284 @@
...
+
+This parameter provides a safety valve: if the community determines that Optimistic Constitutionality is being exploited or producing poor outcomes, it can be disabled through normal governance without requiring a hard fork.
+
+### New Governance Protocol Parameter (Recommended)
+
+This CIP proposes a new governance protocol parameter:
+
+- `minEpochsLiveGA :: Int` — Controls how many epochs must pass at mininum before a Governance Action is approved, irregardless of voting thresholds. 
+ 
+This parameter directly addresses the risk "Collusion" in the Remaining Risk chapter below, and provides mitigation for a collusion attack, where a proposer and the Top x DReps submit a malicious Governance Action last minute, which could fully drain the treasury within a single epoch, or introduce contentious parameters (within guardrails).
```
lehins, line 174: This solution does not completely address the collusion behavior.

Either disguised or clearly malicious proposal can sit there for `minEpochsLiveGA` number of epochs, while colluded DReps wait before voting on it right before the end of the last epoch in this perios, putting into a ratified state last moment, hence guaranteeing its enactment one epoch later. Since colluded votes were submitted right at the end of the epoch, it would give no time to CC members to act.

The reason why I say it does not completely address, because it is based on the assumption that CC members will **review all proposals**. But, if there are no votes by DReps throughout the `minEpochsLiveGA` period, then CC members might not even look at such proposal, until it is already too late.
On the other hand if we expect CC members to **review all proposals** as they come in, then what is the point of this proposal? Voting is the easy part, it is reviewing that takes time.

CIP-0182/README.md
```diff
@@ -0,0 +1,284 @@
...
+
+### Versioning
+
+This specification is versioned by the CIP document itself. Changes to the specification require an amendment to this CIP or a superseding CIP.
+
+## Rationale: How does this CIP achieve its goals?
+
+### Why Not Just Remove the CC?
+
+The CC serves a necessary constitutional function. Even if DReps empirically catch bad proposals, the CC provides a formalized, accountable body that acts as last line of defense and upholds the standard of the Constitution. Optimistic Constitutionality preserves this role while making it reactive rather than proactive. The CC remains essential as the body authorized to formally declare "this violates the Constitution" — a declaration that carries specific meaning and consequences.
```
lehins, line 190: > Optimistic Constitutionality preserves this role while making it reactive rather than proactive.

I think this is the fundamental flaw of the whole proposal.

Both reactive and proactive expect CC members to review proposals. However, if we do not require CC members to vote, we do not get the guarantee that the proposal was reviewed.

Today if a CC member voted "yes" on a proposal, which is clearly unconstitutional, it would give solid grounds for the community to evict such CC member. But, with reactive approach, if no CC member acts on such proposal then who is at fault? Do we boot the whole Constitutional Committee for negligence?

Most importantly, this proposal reduces the duty of CC members, which is exactly the goal of this CIP, but that is, in my opinion, what will make the whole Cardano Governance worse.

lehins (commented):
As a developer I try to stay neutral to this kind of CIPs and only comment on the feasibility of implementation. So, implementation wise this CIP is fairly straightforward. But, personally, I am in total agreement with @thenic95's https://github.com/cardano-foundation/CIPs/pull/1164#issuecomment-5620639479

One issue that is technical, instead of social, that I see in this CIP is that normally ratification happens as soon as all thresholds have been reached and if most votes came in last block before the end of the epoch, it would not give an opportunity for CC members to submit a vote that challenges the proposal. I made a longer comment about this in the section that introduces `minEpochsLiveGA` protocol parameter as a solution to the problem.

Overall, I personally do not agree with the issues listed in the CIP, consequently I believe that the changes proposed in this CIP would make Cardano governance much weaker. That is the whole point of a CC member: is to review and act on proposals that are submitted on chain. If CC is no longer required to vote, that  would effectively delegate their power to DReps, since CC members would no longer even be forced to review any of the submitted proposals. With such change in responsibilities I really do not see a point in having a Constitutional Committee at all. 
IMHO, being a CC member is not only a privilege, because they were elected by the community, but also a very important responsibility. Most importantly, that position is taken up voluntarily with fully disclosed responsibilities up front, same as various other [committees at Intersect](https://docs.intersectmbo.org/intersect-knowledge-base/intersect-overview/intersect-committees). If a CC member is too tired or simply not interested to vote, then a different one should be elected.

Moreover, I do not for a second believe that an empirical evidence from less than two years of governance is a sufficient argument for making such a drastic change to Cardano Governance.

With all that said, this is just my opinion. So, if the community and all CC members feel differently, Ledger team can certainly implement changes proposed in this CIP.

kenerik:
@ptrdsh Sorry for my late reply to this debate.

This proposal treats constitutional safeguards like a Layer 2 scaling solutions
A dangerous category error. Optimistic rollups work in cryptography because fraud proofs are deterministic.

Democratic checks and balances, however, require friction, active human judgment, and proactive deliberation.
As in the foundational premise of CIP-1694 that was to establish a body with an explicit duty to review.

By making CC consent passive you eliminate the guarantee of oversight.
It equals to the Cardano Constitution becomes an exception rather than the foundation.

This doesn't streamline democratic values.
It degrades Cardano into a federated state where accountability can be bypassed by mere apathy.
We cannot sacrifice institutional integrity for convenience.

@lehins Arguments are spot on.
"changes proposed in this CIP would make Cardano governance much weaker"

To add on @thenic95 arguments.

A crucial point regarding human psychology.
Democratic checks and balances are not designed to be efficient.
They are designed to be secure.
An optimistic model inevitably triggers the bystander effect.
If every CC member assumes someone else is actively monitoring the proposals, no one actually does.

rphair (commented):
1 - Because I'm now certain the CIP editors will eventually face such a decision regarding this proposal (and others in the future), I need to ask the reviewers here (@lehins @thenic95 @kenerik @Thomas-nada @brouwerQ @Ryun1 + you too @ptrdsh; cc @perturbing @ch1bo) a basic question on behalf of the community — to incorporate the expectations of your own governance _and_ technical circles:
<ol type="a">
  <li>Should well-defined but unpopular CIPs be merged anyway — assuming they also include (either as a limiting factor on <i>Motivation</i> or a summary in the <i>Rationale</i>) the adverse reactions received in peer review — OR</li>
  <li>Should these CIPs be <s>blocked at the source</s> merged with <code>Status: Inactive (...reason...)</code>, because they might otherwise support a false perception of "standardisation" for the unpopular idea or create undue pressure to implement it?</li>
</ol>

2 - It is also a good time to emphasise the simple axiom that <ins>CIPs are not governance</ins> and only create propositions that can be used in that governance process.
* For this reason, the duly noted substantial opposition to this proposal actually supports our primary requirement that a proposition have a good enough technical definition to be universally understood and unambiguously implemented.

However @ptrdsh, before were to consider merging this someday, we would have to confirm that all the arguments above are fairly included in the section [Addressing Community Concerns](https://github.com/ptrdsh/CIPs/blob/patch-2/CIP-0182/README.md#addressing-community-concerns) — even (and especially) if it means adding a table column or new section like "How This CIP **Does Not** Address" them.
* Even as it stands now, most readers & certainly those considering implementing this CIP would click the original PR `Discussions:` link from which the argument against implementation should be easy to follow.

**p.s.** I thought after submitting this comment we might also want to include, in both the poll & general review, the 2 who have been CIP editors _and_ served on the Constitutional Committee: @KtorZ @Crypto2099

lehins:
> Should these CIPs be blocked at the source, because they might support a false perception of "standardisation" for the unpopular idea or create undue pressure to implement it?

@rphair I personally prefer this direction. Otherwise it is not clear which CIPs should be implemented and which ones should not. Normally in software development if something landed into the `master` branch, that means it has passed all the review and CI and is considered either a good solution or is on the way to become a good solution. So, it seems to me counter productive merging CIPs that only a handful of people think is a good idea. Also, if it is not a popular proposal, does it really constitute an "improvement", as in the "Cardano *Improvement* Proposal"?

In other words, I believe that if there is no agreement between the overwhelming majority of participants that a solution proposed is adequate for the problem stated, then it should be turned into a CPS instead, so exploration of a better solution can happen.

On the topic of this CIP in particular, I believe there is still a technical issue unresolved, which I mentioned here: https://github.com/cardano-foundation/CIPs/pull/1164#discussion_r3996775788
Because of that, I would recommend not merging this CIP, until that issue in the design is resolved. FTR, I do not have a suggestion on how to address it, I just know that it is a problem.

kenerik:
Your question, @rphair  goes to the core of what the CIP repository represents.
As CIP-0001 states in its abstract, CIPs are a mechanism for "collecting community input" and their history serves as "the historical record."

I believe well defined but unpopular CIPs should be merged precisely to serve as this historical record.
I know many have the perception that a CIP number implies an endorsed standard ready for adoption, but in the broader blockchain industry, that is not the case.

For example, controversial BIPs and EIPs are recorded but explicitly tagged as "Rejected", "Stagnant" or "Withdrawn". 
Archiving a flawed idea and permanently documenting why the community rejected it.
This prevents the exact same idea from resurfacing two years later.

If we block this PR at the source, we erase the valuable labor this review group just did in highlighting the governance dangers, and we doom ourselves to repeat this exact debate in the future.

If merged properly with the opposing arguments included, this CIP serves as an excellent educational tombstone.
It proves the idea was formalized, robustly debated, and ultimately "Rejected" by community review.

On a side note, i know CIP 001 was writen way back ago.
There is a tread here from 2022 on the topic on tags i mentioned that we might need to revisit :-)
[CIP-0001 | Rework to reflect reality](https://github.com/cardano-foundation/CIPs/pull/331#top)](https://github.com/cardano-foundation/CIPs/pull/331#discussion_r978597992)

gitmachtl:
Just found this CIP...

First, i find it a bit of disturbing, that such drastically changes to Cardano governance behavior are discussed here in a Github pull request comment section, away from most of Cardano related people. Flying way under the radar, this should be brought to the public on other platforms, not discussed within here alone imo.

Second, we have a voting period of ~6 epochs for a reason. People need time to review governance actions. It can't be the normal case that someone of the CC members has to block an action by submitting a fast early NO vote, just to save the review time for other CC members.

Third, i dislike CIPs that are on the surface trying to solve a problem, that is not even a problem. There is not a single governance action where the CC voting came in too late. Yes, sometimes it takes a while, but thats the nature. I personally dislike solutions were something can go sideways just because an important entity did not see it in time.

Fourth, we have a critical DRep voting power distribution imo. Not many DReps can push true an action super quick if they organize and do it.

Sixth, take the HardForkInitiation as an example. Everytime when i asked about that we must take care about all integrations, exchanges etc..  the answer was always "The CC takes care of it and will only greenlight it when we have reached a certain threshold on exchanges and co". So should the default behavior of a CC member be to quickly vote NO just to make room and time for other CC members? Does not look good if there is always a NO vote at first on actions.

rphair:
@kenerik re: https://github.com/cardano-foundation/CIPs/pull/1164#issuecomment-5667995454 — we've merged CIPs with the `Inactive` tag before — not often, but for the reasons you mention & when the author was cooperative with that idea — and [CIP-0001 has provided for that](https://github.com/cardano-foundation/CIPs/tree/master/CIP-0001#status-inactive) ever since the pull request that you highlighted.
* Therefore I've updated my discussion question above so it suggests merging with `Status: Inactive (...reason...)` rather than the ambiguous "block at source" — which might suggest just cancelling the PR after failing to get consensus about merge (which BTW is generally something the authors themselves do: explicitly closing when they meet too much resistance, or silently abandoning the PR).
* @ptrdsh I think we have a good candidate for merging as `Inactive` with a term to be chosen that honestly (and cooperatively) reflects how the community feels about it.  If you can start thinking about that option then we would have a more ready way out of any deadlock in this discussion.

@gitmachtl re: https://github.com/cardano-foundation/CIPs/pull/1164#issuecomment-5668243233 — CIP-0001 itself suggests ([1.b. Authors seek feedback](https://github.com/cardano-foundation/CIPs/tree/master/CIP-0001#1b-authors-seek-feedback)) using other media besides GitHub for discussions & then providing links to those discussions back here in the PR.  Indeed this would have been a good case for the author & reviewers to do that: though note the editors don't have the social network presence to do it themselves.  But editors also noted that:
* in March the CF governance lead participated in the review (https://github.com/cardano-foundation/CIPs/pull/1164#issuecomment-4097188357) 
* in June, a representative from the Civics Committee: in response to a public discussion at the CIP meeting (https://github.com/cardano-foundation/CIPs/pull/1164#issuecomment-4662140981).

If you can provide any support that would push this CIP candidate even more "into the radar" then please use your social network influence to do so & link it back here so we can include that feedback.  We could really use your help in getting any such contended CIPs out into the open early & often.

katomm:
Agree with @gitmachtl on all points, especially the first one. A change to how the CC works should not be decided in a PR thread.

Two numbers on the premise, from mainnet data (all 157 governance actions through epoch 654):

The premise holds. 80 actions crossed their DRep threshold and the CC blocked none of them. The one that was not enacted failed on the SPO vote, with the CC 5 to 0 in favour.

The saving does not: 39 of the 151 actions with CC votes carry at least one No, and 9 of the 81 enacted actions were approved with a No in the mix, usually 6 to 1. A single dissenting member is the normal case. Under this rule every one of those would have reopened the full CIP-1694 process anyway, so the committee still does the complete review for about a quarter of all actions, now against the clock. What changes is the default for the rest: an (for whatever reason) absent committee currently blocks, under this rule it ratifies, with no rationale on record.

@rphair posted on DRepTalk for better visibility: https://dreptalk.com/t/optimistic-constitutionality-a-cip-that-would-make-cc-silenc-xgwovq/

lehins:
> If merged properly with the opposing arguments included, this CIP serves as an excellent educational tombstone.
It proves the idea was formalized, robustly debated, and ultimately "Rejected"/"Inactive" by community review.

I was not aware that this was an option. In this case I fully agree with @kenerik It is best to merge a CIP as rejected, instead of just closing a PR

rphair:
@ptrdsh @lehins — I resolved to wait around the clock after the last suggestion — to give anyone else, especially the author, time to comment.

@ptrdsh would your goals for submitting this proposal still be met if simply publishing the idea along with these community disclaimers vs. implementation?  (With an `Inactive` status, those comments resting here in the PR would be enough & wouldn't need to be added to the CIP text.)

If so, since the `Status:` would be set to `Inactive (TERM)`, what would you feel comfortable with for `TERM` (preferably 1 or 2 words that other CIPs might use, or maybe 3 at the most for a unique circumstance)?

PR #1173, opened 2026-03-31
Update CPS workflow to manual dispatch run

Ryun1:
- Fix workflow to ignore case sensitive directory names
- Update workflow to allow editors to manually run the workflow on specific PRs
  - by default the workflow runs on commits which contain a `/cps*/` directory
  - a manual workflow dispatch allows runs on existing PRs without new commits

### DIFF
--- a/.github/workflows/cps-validation.yaml
+++ b/.github/workflows/cps-validation.yaml
@@ -3,7 +3,8 @@ name: CPS Validation
 on:
   pull_request:
     paths:
-      - 'CPS-*/README.md'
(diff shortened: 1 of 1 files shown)


rphair (approved):
@Ryun1 I trust you in this matter (and only have little practical experience with GitHub CI: not enough to correct anything you've set up) so will be signing off categorically on all of these... in the interest of getting them into production ASAP so they can be verified or corrected further.

perturbing (approved):

PR #1175, opened 2026-04-02
CPS-0030? | Quantum-secure settlement layer

perturbing:
This CPS discusses the technical details of making the Cardano layer one settlement layer quantum secure. The scope of this work is limited by the consensus side of the PQC discussion.

[Rendered version](https://github.com/perturbing/CIPs/blob/perturbing/CPS-PQC/CPS-0030/README.md)


CPS-0030/README.md
```diff
@@ -0,0 +1,46 @@
...
+## Abstract
+<!-- A short (\~200 word) description of the target goals and the technical obstacles to those goals. -->
+
+## Problem
+<!-- A more elaborate description of the problem and its context. This section should explain what motivates the writing of the CPS document. -->
+
+## Use Cases
+<!-- A concrete set of examples written from a user's perspective, describing what and why they are trying to do. When they exist, this section should give a sense of the current alternatives and highlight why they are not suitable. -->
+
+## Goals
```
hjeljeli32, line 252: **Goals (proposal)**

Following the threat model above, we could define the following high-level goals:

- **Ensure post-quantum security of the root of trust**
  - Cover signature schemes, VRF, and delegation mechanisms  
  - Favor mature, well-studied, and standardized constructions (e.g. NIST PQC)

- **Optimize for practical constraints (latency, bandwidth, storage)**
  - Carefully evaluate trade-offs between signature size, public key size, and verification cost  
  - Prefer schemes that minimize impact on block size, propagation, and validation  

- **Maintain acceptable network performance**
  - Ensure block production, propagation, and validation remain within acceptable limits  
  - Allow for potential adaptation of protocol parameters if needed  

- **Support crypto-agility**
  - Enable smooth and incremental migration of cryptographic primitives  
  - Minimize disruption to existing infrastructure and avoid costly future migrations

Do you think it makes sense?

hjeljeli32 (commented):
Thanks @perturbing for kicking this off 🙏

I added a couple of comments on the *Problem* and *Goals* sections to try to structure the discussion around threat model and priorities.

For the next steps, I think it could be interesting to expand on:
- **Use cases** (especially from a consensus / protocol perspective)
- You and Gamze have a very strong view on the use cases and system aspects.

Of course, very open to feedback on my suggestions above 👍

hjeljeli32 (approved):

rphair:
@hjeljeli32 thanks for joining the discussion; I look forward to seeing your & the co-authors' continuing work on this.

Just so the CIP editors (@perturbing among them) can track when this will be ready to introduce into review (by default, only when taken out of `Draft` status): when you approved this in https://github.com/cardano-foundation/CIPs/pull/1175#pullrequestreview-4087714561, what were you approving? (since there is currently no content in the CIP)

hjeljeli32:
> @hjeljeli32 thanks for joining the discussion; I look forward to seeing your & the co-authors' continuing work on this.
> 
> Just so the CIP editors (@perturbing among them) can track when this will be ready to introduce into review (by default, only when taken out of `Draft` status): when you approved this in [#1175 (review)](https://github.com/cardano-foundation/CIPs/pull/1175#pullrequestreview-4087714561), what were you approving? (since there is currently no content in the CIP)

Thanks for pointing this out @rphair. I approved by mistake and did not mean to signal that the draft is ready for review.

At this stage I only intended to leave feedback on structure and content. Please disregard the approval; @perturbing or another editor can dismiss it if needed.

rphair:
thanks @perturbing - FYI other followers of this thread: will be introduced at CIP meeting starting in a few minutes: https://hackmd.io/@cip-editors/133

CPS-0000/README.md
```diff
@@ -0,0 +1,282 @@
+---
+CPS: ?
```
rphair:
```suggestion
CPS: 30
```

CPS-0030/README.md
```diff
@@ -0,0 +1,282 @@
...
+1. Ensure that signatures in transaction, governance, staking, and block forging
+   cannot be forged.
+2. Ensure the blockchain remains forward and adaptive secure.
+
+Again, such a CIP does not have in scope the Plutus language, as it is
+constraint by other design principles and timings.
+
+## Open Questions
+
+<!-- A set of questions to which any proposed solution should find an answer. Questions should help guide solutions design by highlighting some foreseen vulnerabilities or design flaws. Solutions in the form of CIP should thereby include these questions as part of their 'Rationale' section and provide an argued answer to each. -->
```
rphair, line 267: @perturbing I'm assuming this is left here in case co-authors need it as a guide to expand upon the current list of "open questions" somehow (either in number or detail).  When you feel this section is complete, please remove this comment. 🤓

rphair (commented):
@perturbing @hjeljeli32 this is a useful collection of applicable ideas that (from a non-expert perspective) I believe will be a proper guide to effective discussion, development and standardisation on this subject.

CIP meeting discussion has also acknowledged this as a candidate & so please update the directory name to `CPS-0030` and fix up the "Rendered version" link in the OP 🎉

As with other esoteric CIPs and CPSs we may rely upon yourselves to indicate when sufficient expert review has been obtained... and hopefully the CPS number will help qualify this as ready for more demanding review and circulation.  @perturbing of course you can also move this on to `Last Check` yourself and/or ask CIP co-editors for a last-stage review when that appears to the the case.

CPS-0030/README.md
perturbing: Indeed, I left it there to just hint at what could be added later (following the discussions)

ptrdsh:
Thank you for starting this CPS! 
Its obviously relevant... but also highly timely - perhaps more so than we think. 

I strongly believe we would benefit greatly as ecosystem from adding PQ _mitigation_ (not final solutions) already in Dijkstra, or at least into Euler (afaik thats the name for the next-next HF). Those mitigations could be kill switches, or expensive PQ solutions (which we I THINK already have in a way through Plutus scripts with hash-based spending rules, right?), but sending the signal of partial or optional readiness over just a roadmap is extremely valuable. 

___
Regardless of perception, I would like to share an approach for (partial) mitigation by the Bitcoin community, which I found insightful, and might be interesting for Cardano, as it addresses all three key types: 

Their work includes a kill switch (basically BIP 361) for user funds to block spending outputs that are based on simple signatures... 
... but then also to unlock those funds by proving ownership through ZK : 
1. utxos effectively get censored by SPOs if they rely on just an ECDSA sig to spend them. (or spending gets disallowed after epoch x on protocol level) 
2. Now fortunately, the majority of funds are sitting on addresses that are derived from a seed using BIP32/CIP-1852/ 1853 and 1854. Also fortunately, BIP32 derivation includes running the seed through a hash function before deriving child keys for usage in addresses. As @perturbing wrote in the CPS, current Quantum Computing (predominantly relying on Shor's algorithm) does not crack common hashing functions (there is Grover's algorithm, but it reduces the hash bits from 256 to 128, but that is high enough still). Since QC only cracks elliptic curve cryptography, not hashes, it is impossible to derive the seed from an exposed pubkey.
3. With that "lucky circumstance", there now exists a neat way to unlock those censored ECDSA utxos after a kill switch is activated: The owner of that utxo can create a ZK proof that they derived the exposed public key from their seed using BIP32 and can thus prove they are the rightful owner and spend that utxo. 
4. Problem is only: where should they spend it to? Once it is spent, the utxo would be exposed to the same situation as before.2 solutions:
- the pubkey is exposed only with the spending signature, and an address is the result of a hash, which is not vulnerable, so always using an address exactly once would already work as mitigation. 
- Or: make use of Plutus and spend "ECDSA utxos" that got unlocked with ZK-HD-proofs into script-addresses that require some hash to spend.

they also have this pretty dashboard: https://www.projecteleven.com/bitcoin-risq-list/metrics
________
but so much so only idea sharing...


In my mind, next steps should be defining or agreeing on what the goals and non-goals are, or at least separate them. 
- Are we trying to find a solution for coins that are sitting on vulnerable addresses? 
  - for those that re-use addresses deliberately? 
  - for those that inadvertadly ignore every and all migration advice (or simply those coins that are lost)?
- the aspect about VRF and KES being vulnerable prob separates that question about catering to vulnerability further than just end-user issues, as those two are systemic risks and not just a matter of personal security.  
  
- Or are we just trying to find [signatures, address-formats, logic, solutions] that are PQ resistant?

perturbing:
Thank you, @ptrdsh.

The bitcoin approach is indeed a known way to "upgrade", but sadly it only works/is secure for BIP32 (hash-based) derived keys. The keys that are generated via the CLI (or do not follow BIP32) cannot produce a seed phrase (this is really impossible). A few other hurdles with this approach, though a quantum-secure proof system exists, are that these are also not battle-tested. Here the risk is that a ZKP is underconstrained or even the soundness of the proving method is broken (so agility is here also a must). Then there is also the problem that different hardware wallets derive keys differently (ever tried to load a key from a Trezor to a Ledger? :smirk: ). There is no one-size-fits-all.

> In my mind, next steps should be defining or agreeing on what the goals and non-goals are,

I think this is why we have written this CPS, and just to reiterate our position. The weakest link is consensus; if that breaks, all bets are off. 

That is why plutus is out of scope (it has entirely different design principles and security models). Also, the first thing at risk is the consensus model. For plutus, I do have an idea for a quick fix, which is adding a SHA512 primitive. This way one can encode the public key hash via that and rotate it for every witness in the plutus script.  This would work for all signature schemes Plutus currently supports but does mean a more impactful rewrite of scripts.

As to the goals on consensus, I think the problem is threefold: the signatures for spending/governance/spo operation, etc. (all ed25519 based), the VRF leader schedule primitive (breaks adaptive security if QC attacks it); and lastly, the KES key, which ensures forward security to the chain. Note that the latter is also impactful, as we cannot retrospectively use different keys from past forged blocks; that is, we have to accept that a QC adversary can break this forward security, meaning another approach for safe bootstrapping a new node has to be found.

Also note, a CPS is not there to give a solution but to describe the problem :+1:  

So as an overview

>Are we trying to find a solution for coins that are sitting on vulnerable addresses?

Yes, that is in scope in this problem statement (this CPS does not describe how to solve it).

> for those that re-use addresses deliberately?

Yes

>for those that inadvertadly ignore every and all migration advice (or simply those coins that are lost)?

Good question; currently the CPS does not describe the details of a transition. I'll extend the CPS to scope out more about the details of a transition (and what questions will pop up like yours).

> the aspect about VRF and KES being vulnerable prob separates that question about catering to vulnerability further than just end-user issues, as those two are systemic risks and not just a matter of personal security.
>
>Or are we just trying to find [signatures, address-formats, logic, solutions] that are PQ resistant?

Note that the problem is equal in parts; both the spending issue and consensus break major protocol guarantees. A user is not secure if only one is solved. Thus, this CPS addresses both! Note that 3 CIPs that address each of the 3 parts (Ed25519 / VRF / KES) individually can happen.

CPS-0030/README.md
```diff
@@ -0,0 +1,285 @@
+---
+CPS: 30
+Title: Quantum-Secure Cardano settlement layer
```
rphair:
```suggestion
Title: Quantum-secure settlement layer
```
... if you concur, can you also update the PR title accordingly?  From what I see around the Internet, `quantum-secure` generally carries a hyphen & also matches other terms in your document like `quantum-capable`.

rphair (commented):
@perturbing it occurs to me belatedly that we've been advising contributors to leave off the term `Cardano` from CIP & CPS titles, especially when it's unambiguous and easily inferred:

CPS-0030/README.md
perturbing: Good point, thanks!

ptrdsh:
Thanks @perturbing for the detailed reply.

Agreed on the unpredictability of ZK ownership proofs with the BTC method (albeit we can heavily rely on research done there to mitigate some of the unconstrained risks, etc.), but indeed, the method has its flaws, which includes complicated derivation differences, as well as "raw" HD-less keys and other subsets of all assets that are not included.

The reason I brought it up in a CPS and in connection with the goal- and non-goals question is that I believe specifically the transitioning part from ed25519-sig protected addresses to hash- or lattice-based-sig protected keys/addresses is a non-trivial, if perhaps not impossible task, because we simply cant force the transition for users. (this is why i asked Are we trying to find a solution for coins that are sitting on vulnerable addresses?).
Note that I do not mean the providing of a transition recommendation, or the providing of transition targets (such as addresses on a new address/key-format), which are obviously part of the goals, but specifically whether it is our (us being "people working towards Cardano PQ") task to protect vulnerable assets at rest, or those that are lost/eternally abandoned, as this type of problem is in a different category to upgrading to PQ, and should perhaps even get a different CPS.

If we, as a result of a potential impossibility are forced to exclude the transition for end-users from the goals of the making Cardano quantum-secure (not unreasonable in my mind), then especially it should go into a different CPS, as the metric for solutions/CIPs built out of such a problem statement would be % of unprotected assets protected, and include different discussions and considerations - especially also on social consensus levels.

I think differentiating this frees us from looking forward unconstrained.

CPS-0030/README.md
```diff
@@ -0,0 +1,285 @@
...
+
+The prospect of a quantum-capable adversary is becoming increasingly credible,
+and with that, the Cardano blockchain should prepare for it. This CPS describes
+the problems this QC adversary brings to Cardano. It has in scope all the
+cryptographic primitives that are used in the base settlement layer of Cardano.
+The goal of this problem statement is to describe the technical weaknesses in
+detail and pose a non-exhaustive overview of some replacement primitives to show
+the impact and give a sense of the design space.
+
+This CPS does not have the on-chain smart contract language Plutus in scope, as
```
perturbing, line 28: After an interesting discussion with @rrtoledo, we noted that the two are somewhat intertwined.

That is, from the Plutus context, we can view who signed the transaction or which keys are signing certificates, like Drep votes / CC operations, etc... 

Note that transactions are validated in two phases, and in phase 1 these signatures are validated. This means that a swap of signature schemes in ledger/consensus does not impact the run time of phase 2 significantly. The most notable impact here for plutus would be that the public keys that are in the `TxInfo`, for example, `txInfoSignatories`, will increase in size.

A comment on how the two are intertwined is in place. 👍

CPS-0030/README.md
```diff
@@ -0,0 +1,285 @@
...
+2. Ensure the blockchain remains forward and adaptive secure.
+3. Any migration should explicitly address how the old crypto should transition.
+   This includes how credentials backed by different derived keys are handled.
+
+Again, such a CIP does not have in scope the Plutus language, as it is
+constraint by other design principles and timings.
+
+## Open Questions
+
+<!-- A set of questions to which any proposed solution should find an answer. Questions should help guide solutions design by highlighting some foreseen vulnerabilities or design flaws. Solutions in the form of CIP should thereby include these questions as part of their 'Rationale' section and provide an argued answer to each. -->
```
rphair:
```suggestion
```
I assume we are ready to execute the suggestion in https://github.com/cardano-foundation/CIPs/pull/1175#discussion_r3158084511 now that the original ideation & developmental review have concluded.

rphair (approved):
@perturbing it definitely seems to me that the review on this has stabilised — with no new comment in 3 months — and with much of the discussion & suggested improvements coming from the co-authors & all of it incorporated for some time now.

I think you would agree it's better for CPSs to be pushed out into the community to seed the work on conceiving and developing particular solutions.  As Consensus mechanisms continue discussion in the limelight, I would much prefer to refer to "CPS-0030" on the post-quantum front as a proper CPS rather than a "candidate".

So I would much rather merge this than incubating it indefinitely: and if you still consider it a work in progress then please @perturbing @hjeljeli32 @ptrdsh and other subject matter experts feel free to submit any update PRs as necessary: after `Last Check` and hopefully editorial agreement to merge tomorrow (or whenever any proposed changes are executed): https://hackmd.io/@cip-editors/141

CPS-0030/README.md
rphair: Resolved as https://github.com/cardano-foundation/CIPs/pull/1175#discussion_r3800585708 with https://github.com/cardano-foundation/CIPs/pull/1175/commits/b3421ac50b9caca05d0186335acfa8d2628983b0.

hjeljeli32 (approved):
LGTM 🔒 🚀

CPS-0030/README.md
perturbing: closed by https://github.com/cardano-foundation/CIPs/pull/1175/changes/f90820aa5c6262305095c21452201d473b7b73b3

CPS-0030/README.md
perturbing: thhanks for removing this!

rphair:
@Ryun1 FYI at the CIP meeting today we confirmed that @perturbing:
* has finished his own updates as of https://github.com/cardano-foundation/CIPs/commit/f90820aa5c6262305095c21452201d473b7b73b3 resolving [this 1 remaining point](https://github.com/cardano-foundation/CIPs/pull/1175#discussion_r3257185895);
* is happy with the material as-is & so this can be merged any time you sign off (no rush).

Ryun1 (approved):

PR #1177, opened 2026-04-08
CIP-0102 | Version 2 Standard & CIP-0088 Extension

SamDelaney:
This pull request introduces version 2 of CIP-102, including support for multiple royalty policies (intrapolicy royalties) & optional extension of CIP-88. It clarifies how royalty tokens are constructed, referenced, and discovered, and provide detailed examples for implementors.

### CIP-0102 Standard Specification:

- Updated `README.md` for CIP-0102, expanding the royalty datum format with support for multiple royalty policies with numeric postfixes and integration with CIP-88 via a new extension. Includes detailed field descriptions, rationale, example calculations, and discovery workflows for both off-chain and on-chain contexts.

### CIP-88 Extension:

- Extended the CIP-88 master CDDL schemas (`CIP88_Master_v1.cddl`, `CIP88_Master_v2.cddl`) to include the new `cip102-details` field, enabling collections to advertise their royalty tokens directly in their registration certificates.

- Added a CIP-88 extension specification for CIP-0102 (`CIP-0088/CIPs/0102/README.md`), detailing how royalty token asset names are to be listed in CIP-88 registrations and providing usage examples for both single and multiple royalty policies.

- Added a new CDDL schema (`CIP102_v1.cddl`) and example JSON (`CIP102_v1.json`) defining the structure and serialization of the royalty datum metadata, including support for multiple royalty token asset names.


SamDelaney:
Tagging @Crypto2099 because CIP-88

CIP-0102/README.md
```diff
@@ -19,21 +20,62 @@ License: CC-BY-4.0
 
 This proposal makes use of the onchain metadata pattern established in [CIP-0068][] to provide a way to store royalties with greater assurance and customizability.
 
+## Table of Contents
```
rphair:
```suggestion
<details>
  <summary><h2>Table of contents</h2></summary>
```
(also would need closing `</details>` after the TOC end)

Since this is not a main CIP section, what we agreed with long CIP authors so far has been:
* "long" CIPs for which a Table of Contents is justified are quite a bit longer than CIP-0102: whose headings are close enough to read in a few scrolls & a visual scan;
* a Table of Contents should be collapsible (on GitHub) and in a standard form that `cips.cardano.org` can easily recognise and omit from the markup: since it's already generating a TOC from the headers of every CIP.

If you still want to include the TOC, please update it to the Markdown + HTML that you can see here, for instance: https://github.com/cardano-foundation/CIPs/tree/master/CIP-0164#table-of-contents

BTW if this is not in CIP-0001 or CIP-9999 yet it's because editors are collectively working on standardising some other aspects of the template & format documentation related to @Ryun1's ongoing standardisation work.

rphair (commented):
@SamDelaney this is on the agenda for `Triage` at the next CIP meeting: https://hackmd.io/@cip-editors/132 — besides @Crypto2099's review I would also hope to get from the meeting a list of representative developers to tag from projects that might have live or in-progress implementations: to ensure we get a comprehensive stakeholder review & properly announce that a new version is on the way.

Also, if you could identify key parts of CIP-0088 and CIP-0102 that you've added or substantially changed, it might facilitate our meeting preliminary review if you added links to those parts in your OP above.

CIP-0102/README.md
SamDelaney: All reasonable. This is a short enough CIP I don't think a TOC is necessary - dropping it.

rphair (commented):
@SamDelaney thanks for introducing this at the CIP meeting today where it was confirmed all around as a valid update.

I also noted at the meeting today that NFT royalties — though NFTs themselves are no longer so popular — might also suit more progressive use cases like interest or dividends from RWAs... and though CIP-0102 might not need to mention this latter use case (though you could feel free to add it) it does mean we should still value & work on the maintenance of these protocols.

Meeting also confirmed the request for more tagging from industry projects & appreciates that @SamDelaney will be circulating such a request.  Since backward compatibility issues are not likely, this should proceed toward merge if there is no dispute from project stakeholders or relevant developers.

@SamDelaney please remind the editors once there's been such a review period so we can proceed toward final review and "last check".

SamDelaney:
Tagging relevant cnft devs to make sure everyone has a chance to comment if they want:

@solidsnakedev @BrockCruess @yHSJ @colll78 @PatrickTobler @alessandrokonrad @Dodilanne

rphair (approved):
@SamDelaney without any dispute from the CNFT developers tagged above, and given your intimate involvement with the CIP-0102 standard itself as represented on 3 different projects, I would consider this a valid update.

We can merge this at or near the next CIP meeting (https://hackmd.io/@cip-editors/134) if there is still no dispute from those devs (@solidsnakedev @BrockCruess @yHSJ @colll78 @PatrickTobler @alessandrokonrad @Dodilanne) and assuming editors @Ryun1 @perturbing are also in agreement about the changes.

Also I would still hope for any comment from @Crypto2099 here to confirm the extension to CIP-0088 is all in order.

BrockCruess:
No specific issues stand out to me 👍

CIP-0102/README.md
```diff
@@ -81,60 +110,90 @@ extra = plutus_data
...
-- If the field is present and > 1 the validators must require a royalty input.
-- If the field is present and set to 0 the validators don't need to search for a royalty input.
-- If the field is not present, validators should accept a royalty input, but not require one.
+- If the field is present and > 0, validators must require a royalty input. The value itself is the numeric postfix identifying which Royalty Token applies to this NFT (e.g., `royalty_included = 2` → look for `(500)Royalty2`).
+- If the field is present and set to 0, validators don't need to search for a royalty input.
+- If the field is not present, validators should accept a royalty input but not require one (compatible with v1 collections using `(500)Royalty` with no postfix).
+
+The correct datum may then be reliably found by searching for a Royalty Token with a matching prefix (& postfix). Examples of how to do so are provided below.
+
+In the case of ambiguity, users/dapps may choose whichever policy they wish to honor.
```
yHSJ: I think a standard should define the standard behavior here. If a dApp correctly implements this CIP, it should be clear what the behavior would be in the "case of ambiguity".

I also think the "In the case of ambiguity" phrasing should be dropped, since "ambiguity" is a pretty vague term.  I would prefer more explicit terminology.

What situations can lead to ambiguity? The fallback should probably be to the first policy (since that's easier to find than the "latest" policy).

CIP-0102/README.md
```diff
@@ -23,17 +24,39 @@ This proposal makes use of the onchain metadata pattern established in [CIP-0068
 
 The inability to create trustless onchain royalty validation with [CIP-0027][] is a major drawback to Cardano NFTs. The pattern defined in CIP-68 represents an opportunity to upgrade the standard to support onchain validation. This CIP aims to eliminate that drawback and demonstrate better support for developers, NFT creators, and NFT collectors, ultimately attracting dapps & NFT projects that would otherwise have taken their talents to another blockchain.
 
-In addition, this standard allows royalties to be split between multiple addresses, another limitation of the CIP-27 royalty schema. Future versions of this standard could  also easily support multiple royalty policies defined for a single collection, applied at the level of individual tokens.
+In addition, this standard allows royalties to be split between multiple addresses, another limitation of the CIP-27 royalty schema.
+
+Version 2 this standard also supports:
```
yHSJ: Version 2 "of" this standard, grammatical error

CIP-0102/README.md
```diff
@@ -23,17 +24,39 @@ This proposal makes use of the onchain metadata pattern established in [CIP-0068
 
 The inability to create trustless onchain royalty validation with [CIP-0027][] is a major drawback to Cardano NFTs. The pattern defined in CIP-68 represents an opportunity to upgrade the standard to support onchain validation. This CIP aims to eliminate that drawback and demonstrate better support for developers, NFT creators, and NFT collectors, ultimately attracting dapps & NFT projects that would otherwise have taken their talents to another blockchain.
 
-In addition, this standard allows royalties to be split between multiple addresses, another limitation of the CIP-27 royalty schema. Future versions of this standard could  also easily support multiple royalty policies defined for a single collection, applied at the level of individual tokens.
+In addition, this standard allows royalties to be split between multiple addresses, another limitation of the CIP-27 royalty schema.
```
yHSJ, line 27: Is there any policy around *how* to split royalties between multiple addresses? Or just the sum of royalties paid to all the addresses must equal the expected amount.

yHSJ (commented):
Thanks for the tag! I wouldn't consider myself a "cnft developer", even though I once worked at JPG Store. That said, I am always happy to provide my thoughts!

First of all, I think it's useful to make it clear what the language means. From what I remember of my time at JPG Store, deciding what is valid given the royalty CIPs was very challenging, since the language was not explicit.

I think it would be very helpful to make it clear you're following [RFC-2119](https://datatracker.ietf.org/doc/html/rfc2119), if that is true.

Other than that, I left a few other comments that I had from a quick scan of the changes.

CIP-0102/README.md
SamDelaney: How royalties are to be split is defined by the datum itself. In the definition for the datum metadata:

```
royalty_recipient = #6.121([
              address,                    ; definition can be derived from:
                                          ; https://github.com/input-output-hk/plutus/blob/master/plutus-ledger-api/src/PlutusLedgerApi/V1/Address.hs#L31
              int,                        ; variable fee ( calculation: ⌊1 / (fee / 10)⌋ ); integer division with precision 10
              optional_big_int,           ; min fee (absolute value in lovelace)
              optional_big_int,           ; max fee (absolute value in lovelace)
            ])

royalty_recipients = [ * royalty_recipient ]
```

[ * royalty_recipient ] indicates a list of royalty_recipient objects, each of which defines the fee they expect.

It should probably be [ 1* royalty_recipient ] though, indicating a minimum of 1 recipient. Will update that shortly.

CIP-0102/README.md
SamDelaney: Changed to `If multiple outputs match this pattern, implementors may choose whichever policy they wish to honor from the valid datums.`

You're right about the 'case of ambiguity' being itself ambiguous - hopefully this makes it clearer.

First policy does make sense, but I'd like to leave the option of evolving a policy which has had its datum accidentally (or short-sightedly) locked, but still has an open minting policy. I think it's still better to allow the implementor the choice instead of forcing them into the easier option.

Still open to being convinced otherwise if you feel strongly about this though.

SamDelaney:
Thanks for the feedback @yHSJ, all has been addressed in my latest commit.

CIP-0102/README.md
rphair: resolving: fixed in https://github.com/cardano-foundation/CIPs/pull/1177/commits/81027edc177b07ad4a0be1dc07c6c65555fc63a6

yHSJ: Ah, yes this does make sense. Sorry, I clearly did a quick skim 😅

CIP-0102/README.md
yHSJ: My biggest concern is just unexpected behavior for the end user(s).

I think it's a reasonable decision to allow implementors to choose whichever policy they wish to honor. That is often the case with "unspecified behavior". In this case though, I think there is more value in consistency across implementations, so that creators and users get some consistency.

SamDelaney: After thinking about this for awhile, I think you're right.

Leaving it optional could allow projects to bypass the validation logic of their locking script, which is intended to be a promise of behavior to the community.

Updated to `If multiple outputs match this pattern, the first-published output should be used.`

Ryun1 (approved):
Seems like a reasonable backwards compatible extension to the existing standards. Thank you @SamDelaney.

PR #1178, opened 2026-04-09
CIP-0183? | Conflict-Based Fee Priority in Mempool

fallen-icarus:
This CIP proposes UTxO conflict-based fee priority as the default mempool policy — when two transactions conflict over the same UTxO, keep the one paying a higher fee. **No hardforks are required.**

This helps alleviate UTxO contention while benefiting Cardano's long-term financial sustainability at the same time. Critically, this introduces a **targeted fee market**; normal users are unaffected.

---
[Rendered README](https://github.com/fallen-icarus/CIPs/tree/mempool-fee-market/CIP-0183)


fallen-icarus:
Mass tagging people who might be interested :slightly_smiling_face: 

@lehins @WhatisRT @colll78 @Quantumplation @ch1bo @matteocoppola @theeldermillenial @perturbing @coot @will-break-it @lupowilk @Crypto2099 @Cerkoryn

CIP-????/README.md
```diff
@@ -0,0 +1,407 @@
...
+
+The correct mechanism for allocating a scarce resource to the party who values it most is an
+auction. This is not a novel insight — it is backed by thousands of years of economic practice and
+centuries of formal theory.
+
+When two transactions conflict over the same UTxO, the mempool runs a trivial auction: keep the
+higher-fee transaction, drop the lower-fee one. The winner pays their bid to the network via
+transaction fees. The loser's transaction fails cleanly — on eUTxO, a transaction that references a
+spent UTxO simply fails validation at no cost to the submitter. It does not execute with a worse
+outcome.
```
Cerkoryn: How does this work when a user's Tx passes validation, but then later is surpassed by a conflicting transaction with a higher fee after it has been accepted and propagated to many nodes?

CIP-????/README.md
```diff
@@ -0,0 +1,407 @@
...
+"child" transaction can reference an output that exists only in the mempool, creating dependency
+chains among pending transactions. An adversary can attach a large, low-feerate child to a parent
+transaction, making the parent artificially expensive to replace. Mitigating this required years of
+incremental design: descendant package tracking, eviction-weight calculations, cluster mempool
+redesigns, and topology restrictions like TRUC/V3 transactions.
+
+Cardano's mempool does not allow spending unconfirmed outputs. Every transaction input must reference
+a UTxO that is already confirmed on-chain. There are no parent-child chains in the mempool, and
+therefore nothing to pin. The entire class of complexity that dominates Bitcoin's RBF design simply
+does not exist on Cardano.
```
Cerkoryn: What about transaction chaining or nested transactions?  Don't both of these contain a similar parent-child dependency in different ways?

CIP-????/README.md
```diff
@@ -0,0 +1,407 @@
...
+higher-fee one, it switches. The higher-fee transaction spreads across the network, displacing its
+competitor node by node. By the time a block producer selects transactions for the next block, the
+higher-fee transaction is more likely to be in its mempool — but not certain to be.
+
+This probabilistic nature is a feature, not a limitation. A deterministic guarantee would require a
+single global ordering authority — a centralizing force. Probabilistic resolution still enables
+rational decision-making: a participant who consistently pays more than their competitors will
+consistently win more contests, and can model their expected success rate based on network topology
+and competitor behavior. Over many contests, the law of large numbers ensures that the fee signal
+is reliable even though any individual outcome is uncertain.
```
Cerkoryn: Earlier you mentioned...

> Transactions are fully deterministic: a user knows exactly what inputs will be consumed and what outputs will be produced before they sign.

Which I assume is in the sense of you either get exactly what you pay for or your transaction never happens.  It is all or nothing.  However, this seems to add a new way for transactions to "fail" in the event that their transaction passes validation and is accepted, but later is surpassed by a higher-paying transaction and subsequently dropped.

I think in practice this would look more like a rollback than non-determinism, but I still wonder how it would affect the user experience.

CIP-????/README.md
```diff
@@ -0,0 +1,407 @@
...
+produced before they sign. There is no shared global state that an adversary's transaction could
+modify to change the outcome of yours.
+
+This CIP does not introduce any new attack vectors. It does facilitate competitive extraction of
+existing value — when bots bid against each other to capture a mispriced limit order, the surplus
+between the order price and the market price is a form of MEV. But this extraction is non-adversarial:
+it cannot worsen any user's execution, the limit order creator receives exactly the price they
+specified, and the bidding war converts extractable value into fee revenue for the network rather
+than leaving it as pure profit for whichever bot happened to propagate fastest. This is MEV
+redistribution, not MEV harm.
```
Cerkoryn: I'm envisioning a scenario where there is some kind of token launch or NFT mint where many users are scrambling to quickly get their transactions included.  Could either a malicious actor or a greedy whale bypass the entire line by setting extremely high fees on their own transactions many times in quick succession?

The fees are of course a limiting factor here, but they would be paid to the SPO reward pool.  The users themselves could be left with a bad experience.  Perhaps this could be mitigated at the dApp/application level?

I could also see this as a feature instead of a bug though, but it probably requires careful design at the dApp level.

Quantumplation:
I don't have time to read the proposal right now, but happy to share some initial thoughts just based on your PR summary;

To me, this doesn't materially alleviate the "bad" parts of contention; indeed, I believe your proposal makes both worse. Whether it's not this is actually true, or whether it is worth it on other grounds, I will reserve judgement until I get time to read your proposal.

To me, there are two closely intertwined "bad" properties of contention:

- As a UX consideration, it forces a user to re-sign their transaction; in particular, this is only made worse by your proposal. In the current model, at the very least if I get my transaction into the mempool early, I can be fairly confident it'll settle; and me and others can build transactions that chain off of it. In your proposal, any transaction spending a contentious UTxO can be pre-empted, making it more common that I have to re-submit, and it breaks entire chains of transactions.
- Second, contention creates fairness problems; in classical finance terms, it ensures the market provides different spreads to different classes of users. Users with access to more sophisticated technology (automated bots, hot-keys for resigning, better mempool monitoring, better connection within the network, get better access to those UTxOs, and thus get an economic advantage. Automated trading bots get dibs on the best opportunities, while average users are pushed towards worse prices. In an orderbook, for example, the UX burden pushes users towards placing limit orders, rather than matching against the best bid/ask, and they get traded against by those sophisticated bots. Your proposal doesn't ameliorate this. On the one hand, it does make this advantage more expensive to exploit, but that benefit shifts to the network, not the "average" user.

Thus, if protocols (like Sundae) want to maintain a structural fairness and nice UX for all users, they're pushed towards protocols that avoid contention (perhaps even more strongly!)

Now, maybe that's fine; those protocols continue to remain unchanged, and other protocols accept this structural imbalance, and at least the network benefits from it, but it doesn't materially lesson the "problems" of contention.

And, at the very least (and maybe you mention this in your writeup), it should be sophisticated enough for the mempool to consider the *chain* of transactions it's replacing, not just an individual transaction, so that earning 0.1 extra ADA on a transaction doesn't cause the network to lose 20 ada worth of fees from the whole chain :sweat_smile:

AndrewWestberg:
I really like this concept overall.

The mempool currently decides whether an incoming transaction is valid by looking at its own bespoke ledger state. This mempool ledger state is not the same as the ledger state of the blockchain tip itself, but rather the blockchain tip ledger state + mempool txns. This ledger state for the mempool exists as if all mempool transactions have already been applied. This is what allows us to do transaction chaining in the mempool for many protocols today.

As such, if a new transaction arrives that gets rejected, how will we tell the difference between a malformed transaction, a truly "already spent utxo" transaction, and one where the only utxo conflict is another tx already in the mempool? Is there an algorithm we can add to this CIP to clarify the process a tx goes through today (rejected due to conflict), vs what process it will go through in the future (kicking out the existing conflicting tx if we paid a higher fee)?

Great work on this CIP!

CIP-????/README.md
```diff
@@ -0,0 +1,407 @@
...
+The metrics agree when competing transactions are roughly the same size — the transaction paying
+more in total also pays more per byte. They diverge when transaction sizes differ significantly:
+
+| Scenario | Tx A (incumbent) | Tx B (challenger) | Absolute fee winner | Fee rate winner |
+|:---------|:-----------------|:------------------|:--------------------|:---------------|
+| Similar size | 500 bytes, 0.5 ADA | 520 bytes, 0.8 ADA | B | B |
+| Different size | 2000 bytes, 1.0 ADA | 500 bytes, 0.6 ADA | A | B |
+
+In the second scenario, the two metrics give opposite answers. Absolute fee keeps the 1.0 ADA
+transaction. Fee rate keeps the 0.6 ADA transaction because its per-byte efficiency is higher.
```
Cerkoryn: At first I thought this would create additional denial-of-service attack vectors, but on second thought that should be already handled by the existing fee calculation, `Fee = minFeeB + minFeeA * size(TX)`

If we consider this to be an additional fee on top of that, then I think we can set all those concerns aside.  However, instead of using an absolute fee winner (flat fee), or a fee rate winner (percent) for the replacement delta I wonder if it could be better served by using a _multiplier_ of the conflicting transactions it would need to replace.

Consider your earlier example of...

> If `B` conflicts with multiple mempool transactions `A₁, A₂, ..., Aₙ`, then `B` must pay strictly
more than the **sum** of all conflicting transactions' fees, plus the replacement delta:
> fee(B) > fee(A₁) + fee(A₂) + ... + fee(Aₙ) + delta

Instead of being just +1 lovelace above the _sum_ of all conflicting transactions, what if it had to be double?  Or 1.5x?  Or some other number?

That would make the fee ramp up much quicker, which would also likely make the contention conflicts (and thus the UX issues of dropped transactions) less common, which could be a good thing.

(I had another reason that led me to this idea in the first place, but it slipped out of my head while I was trying to write it out 😅 .  Will post again if I think of it)

CIP-????/README.md
Cerkoryn: On second thought, I think this is largely solved by intents.  Have a user sign an intent once and then it will eventually go through.

However the tradeoff here is that it may force some dApp designs to have to use intents (and thus the dreaded intermediary 😅 ).  Maybe it would be possible for this CIP to be optional at the dApp-level somehow?

Cerkoryn (commented):
Largely a fan of this CIP, especially the economics of it.  Left some comments on some specific sections mostly about the UX implications.

On a semi-related note, I am reminded of the determinism tradeoffs of the full version of Ouroboros Leios to which @Quantumplation gave multiple possible solutions here:
https://www.314pool.com/post/leios

It is largely a physics/causality problem, but I wonder if having a UTxO contention fee market like this could be another possible solution.  I.e., if multiple people try to spend the same UTxO from different "light cones", then the one with the highest fee wins and the other is rolled back.

I'll need to give that another read though, it's been a minute since I worked through that.

CIP-????/README.md
```diff
@@ -0,0 +1,407 @@
+---
+CIP: ????
+Title: Conflict-Based Fee Priority in Mempool Policy
```
rphair: I'll update the PR title according to this CIP title which I think avoids the assumption that this would create a "fee market" — likely, but not certain — and also avoid the [practically nonexistent term `fee-market`](https://trends.google.com/explore?q=fee%2520market%2Cfee-market&date=today%201-y&geo=Worldwide).

CIP-????/README.md
```diff
@@ -0,0 +1,407 @@
...
+transactions explicitly declare their inputs, contention can be attributed to *specific UTxOs*
+rather than to the network as a whole. This makes it possible to build a fee market that raises fees
+only for the resources under contention, leaving everyone else unaffected.
+
+This CIP implements that capability. When two transactions in a node's mempool attempt to spend an
+overlapping UTxO, the node keeps the one paying a higher fee. No ledger rules, fee formulas, or
+transaction formats change — only the mempool's admission heuristic for handling conflicting
+transactions.
+
+## Motivation
```
```suggestion
## Motivation: Why is this CIP necessary?
```
Full title (to help reviewers confirm content is arranged like other CIPs) as per CIP template (note capitalisation will be consistent with template change under works via @Ryun1).

CIP-????/README.md
```diff
@@ -0,0 +1,407 @@
...
+  eviction if a conflicting transaction paying higher fees arrives — which is the intended behavior.
+- **No transaction format changes.** Fee overpayment is already supported by the ledger.
+- **No consensus changes.** Block validity rules are unchanged. This is purely a mempool policy.
+
+Nodes that do not adopt the new policy will continue to operate correctly. They will resolve
+conflicts via FIFO, foregoing the fee-priority optimization. Each node benefits unilaterally from
+adopting the policy, creating a natural incentive for organic adoption without requiring coordinated
+activation.
+
+## Rationale
```
```suggestion
## Rationale: How does this CIP achieve its goals?
```
(as per previous comment)

rphair (commented):
@fallen-icarus this is a well expressed proposal that should be assessed from all quarters of the ecosystem.  The writing is perfect except for some minor format things; any technical question I might have raised has been better addressed already by the previous reviewers.

On track for introduction via `Triage` at next CIP meeting (https://hackmd.io/@cip-editors/132) and it would also help get the word out for a community-wide review if you can make it to the meeting to say a few words about it.

Tagging @jpraynaud in the meantime to confirm that what is proposed here would mainly impact the `Network` system (as currently categorised) and not have greater implications for `Ledger` (cc @lehins @WhatisRT already tagged: but again regarding this particular question).

CIP-????/README.md
rphair:
```suggestion
Title: Conflict-Based Fee Priority in Mempool
```
p.s. I'm testing the truncation of the term `Policy` since on a second reading the phrase _Priority in Mempool Policy_ would seem (at my level) to achieve what people would think of as _Priority in Mempool_ and therefore be more concise with better impact.

If there's a subtle difference then @fallen-icarus @Quantumplation @AndrewWestberg @Cerkoryn please advise & in any case we'll confirm a best title at the upcoming meeting.

Quantumplation: The purpose of the word "policy" here is to emphasize that mempool behavior is not an enforceable property of the protocol, and so if a single node chooses different behavior, it will still follow the same tip as other nodes. What is being described here is really a convention or property that @fallen-icarus is recommending for the mempool implementations, rather than a protocol change.

That being said, that is true without the word "policy" as well, so to me it just reads as rhetorical emphasis, and doesn't *actually* change the meaning, so... :shrug:

CIP-????/README.md
berewt: Yes, this is an issue. Building on top of mempool transaction exists on Cardano, it is partially covered later in the CIP with the consideration of multi-input conflict (it's a different problem, but may be solved with the same solution). I would suggest rewriting both parts to make it clearer.

CIP-????/README.md
```diff
@@ -0,0 +1,407 @@
...
+Two transactions that both *reference* the same UTxO without spending it do not conflict — they can
+coexist in the same block regardless of ordering. A conflict arises only when at least one
+transaction *spends* a UTxO that the other spends or references.
+
+### Replacement rule
+
+When a node receives a new transaction `B`, and its mempool already contains a transaction `A` such
+that `A` and `B` conflict:
+
+- **If `fee(B) > fee(A) + delta`:** Evict `A` from the mempool. Admit `B`.
```
berewt: The problem I see here is that depending on delta, it introduces a new subtle FIFO problem. If B is higher than A but lower than A + delta, the order matters again.
We can consider it as a marginal tie breaker though.

berewt: It puts a lot of pressure on the definition of delta. Being to low, you're open to tx spam, too high, you would revert to the previous mechanism.

CIP-????/README.md
berewt: It would be a bad UX with the current model as well. It's something to handle at the dApp level (sharding, address tracking). The auction may actually add an extra guarantee to it, as you mentioned.

fallen-icarus:
@Quantumplation Both of your concerns are directly addressed by the mechanics of a localized fee market, and I think the fear of endless bidding wars might be a bit overstated. Consider the following:

1. The retail user is in full control of how much they overpay upfront.
2. Outside of high-value arbitrage/liquidations, bots tend to "penny-pinch" to preserve their margins. Because they can easily retry transactions, they are mathematically better off keeping their bids small.

A retail user can take advantage of this characteristic for scenarios where they desperately need a contentious UTxO. If the user overpays by 5x-10x (so a standard 0.5 ADA fee becomes 2.5-5.0 ADA), they are very likely to win the "auction." It is like a real-world auction where everyone is incrementing by $1, and suddenly someone increments by $100. At that point, the room usually folds. 

The eUTxO situation actually makes this even better because many of these contested resources are "continuing UTxOs" where the resource is immediately made available again after the current tx. When a bot sees a 5x retail bid, the rational, margin-optimizing choice isn't to start a bidding war—it's to let that UTxO go and just grab the next one in the subsequent block for a fraction of the cost. 

A retail user choosing to pay 2.5-5.0 ADA for guaranteed priority does not mean "only the rich can play." A huge reason bidding wars get so toxic on other chains is collateral damage: normal transactions get caught in the crossfire and have to bid for global block space. This CIP's targeted fee market confines the bidding *only* to the specific resource in question, meaning there is no extraneous bidding pushing up the price for the rest of the network. Instead of one giant auction, you have many small auctions:  one UTxO can have a 15 ADA bid while another one could have a 1.5 ADA bid.

CIP-????/README.md
```diff
@@ -0,0 +1,407 @@
...
+network's opportunity cost of admitting `B` is the total fee revenue from all of them, not just the
+most expensive one. The sum rule captures this correctly.
+
+### Evicted transaction behavior
+
+When a transaction is evicted due to replacement, it is simply dropped. It is not re-propagated to
+peers. The evicted transaction may still exist in other nodes' mempools across the network; whether
+it survives depends on whether the replacement reaches those nodes before the next block.
+
+From the user's perspective, detecting eviction is straightforward: monitor the UTxOs that the
```
berewt: I disagree with this. It's straightforward, yes, but way worse than having a direct feedback from the node you submitted to.  We increase the odds of having a tx included in the mempool on submission and not included on-chain, which is generally a bad UX.
More importantly, it means that you can't know (without running your own mempool) whether your transaction has been evicted by a higher-fee one.
Ideally, node implementing it should have an off-chain way to communicate about transactions that have been evicted.

berewt (commented):
First review round. There's some approximations and minor challenges.
My fear is that in the current state, it may lead to an as arbitrary result than today. But maybe with a small fee increase, which isn't that bad.

fallen-icarus:
@AndrewWestberg @Quantumplation @berewt I did not know Cardano supports tx chaining in the mempool. Thanks for pointing this out! 

However, I don't think it changes much because Cardano is structurally different than Bitcoin so many of the downsides shouldn't apply:
1. A Cardano mempool is ~2 blocks vs. 300 MB on Bitcoin
2. Cardano block production is ~20 seconds vs. 10 minutes on Bitcoin

By taking advantage of these differences, we can come up with a simpler solution than on Bitcoin.
- There is no need to artificially cap the chain depth because the small mempool already limits how deep chains can grow.
- The short block time means chains are confirmed quickly, giving an attacker a narrow window to build a pinning chain before the contest is resolved.

With this in mind, it should be enough to just include descendant tx fees in the replacement sum — i.e., B must pay more than the total fees of all transactions it would evict (the conflicting transaction plus its descendants).

---
Please lmk if I'm misunderstanding something else about the mempool. Ideally, someone who understands the low-level details of the mempool can weigh in about the required algorithm.

CIP-????/README.md
fallen-icarus: The new tx still needs to be propagated to the rest of the nodes. It is possible the original tx still makes it into a block if the new tx doesn't propagate fast enough.

CIP-????/README.md
fallen-icarus: Nested txs does not support cross-tx chaining. But for tx chaining, see this comment: https://github.com/cardano-foundation/CIPs/pull/1178#issuecomment-4222798167

CIP-????/README.md
fallen-icarus: It is only in the mempool that the tx can be dropped. Once an SPO adds it to a block, it is too late to replace the tx.

> On second thought, I think this is largely solved by intents. Have a user sign an intent once and then it will eventually go through.

You don't need to use an intent. See this comment: https://github.com/cardano-foundation/CIPs/pull/1178#issuecomment-4222553852

CIP-????/README.md
fallen-icarus: A token launch should rely on CIP-159 where the tokens are held in the account address. This way users can take from the address concurrently. This CIP is more for use cases that can't use another architecture to handle contention (e.g., an order book). Developers should use the right tool for the job.

CIP-????/README.md
fallen-icarus: The delta needs to be small enough to be useful but large enough to deter spam. The multiplier might be too large. Block production is 20 seconds so each auction should be fairly short lived which means there shouldn't be much of a ddos chance here.

CIP-????/README.md
fallen-icarus: Agreed we need to find a balance. I erred on the side of caution by having delta be the full propagation cost of the new tx. IMHO this should be good enough. It means the minimum new bid is approximately +0.5 ADA over the current bid. At current prices, that is ~$0.15. I'd hardly call that an expensive jump :slightly_smiling_face:

CIP-????/README.md
fallen-icarus: Having feedback directly from a node would be ideal, but that would add more work to deliver this CIP. Off-chain tooling is already capable of tracking this. So perhaps this CIP could be phased? Implement the policy now and update the node to give feedback for dropped mempool txs later?

berewt: It's more work yes, but I do prefer more work than making the user experience work. Offchain tooling will notice rejection too late if you don't have access to the node. It creates a mix of blind auction (for users without access to the mempool) with insiders (solutions with access to the mempool) that can use standard auction.
I agree that this UX issue already exists in the current model, when we have divergent mempools, but it's already something people complain about, and we're making it worse.

AndrewWestberg:
@fallen-icarus A cardano mempool "default" is 2 blocks. This is configurable in cardano-node. Many protocols such as dripdropz or Eternl wallet run with a much higher mempool. In ddz case, we use a 1gb mempool. In the early days when we did the Sundae token drop this was absolutely required and we had about a 4-hour backlog of valid chained transactions in our mempool that slowly confirmed on chain. In this scenario, you also have to handle rollbacks and re-inflation of the mempool as the node doesn't handle that automatically. Any rollback kills the mempool so it needs to have transactions replayed.

All this to say, we can't assume small mempools so an O(n^2) change still needs to be considered carefully.

berewt:
Add to this that Leios will increase the size requirement. It would still be a few blocks but higher chance of conflict and arbitrage.

fallen-icarus:
I looked into Bitcoin's RBF approach to see how they handle tx chain eviction, but its mempool is structurally different than Cardano's. Efficiently handling these evictions on Cardano seems to require additional data structures alongside the current mempool. And we also need to consider CIP-159. This is turning out to be a much bigger CIP than I was hoping for :face_with_diagonal_mouth: 

Also, this may be a controversial take, but IMHO tx chain fees **should not** be considered when deciding the auction winner. Auction theory seems pretty clear about this.

**Bundling auctions reduces revenue.** The whole point of this CIP is that each contested UTxO gets its own independent auction. Counting descendant fees undermines that by merging separate auctions into one.

Consider a continuing UTxO that cycles four times. Without chaining in the fee comparison, each cycle gets its own auction. If each attracts a 15 ADA bid, the network earns 60 ADA. With chaining, all four cycles are bundled under a single replacement threshold. A chain paying 20 ADA total across all four beats out individual 15 ADA bids on each — the network earns 20 ADA instead of 60. Separate auctions over independent resources yield at least as much total revenue as bundled ones. This is a standard result in auction theory, and continuing UTxOs are the dominant case for DeFi contention.

**It also misprices demand.** This fee market prices UTxO contention *right now*. A transaction chain spreads its fees across future blocks. Requiring a challenger to outbid the next few hours of activity is not pricing current demand, it is creating a market inefficiency by ignoring the time-value of money.

If we don't consider tx chain fees, then we need a way to protect against ddos since these txs are fully validated but dropped before their fees are collected. Bitcoin caps the max chain depth but it doesn't have plutus script validation. So my current thought is a compromise: If Tx B conflicts directly with Tx A and Tx A has descendants, use the absolute fee for Tx A but only the required minimum fees for each descendant.

```
Replacement Rule:
fee(B) > fee(A) + minFee(B) + totalMinFee(descendants(A))
```

The fee of Tx B only needs to cover the lost validation work of the descendants, not outbid them. This separates the auctions as much as possible while still protecting against ddos.

Quantumplation:
Transactions chained in the mempool are also "right now"; the mempool will include as many transactions as it can in the block, it doesn't force chained txs into consecutive blocks.

And I don't think your hypothetical is correct;

Let's suppose I have A -> B -> C -> D, that pay 1 ADA in fees each;

Let's say there's someone who wants to replace A, and is willing to pay 2 ADA for it.

Replacing A invalidates B, C, and D, meaning the network goes from earning 4 ADA in fees to 2 ADA in fees, you've worsened the UX for users B, C, and D; if B, C, and D are still valuable, then you *might* get them back to the table, and if you do, you've imposed additional computational load on the network.

B, C, and D have already been validated by anyone who puts them in the mempool; that is computational work they should be paid, and it only makes sense to accept an A' if the collected fee is greater than the cost to validate A, A', B, C, and D.

On the other hand, if the minimum fee to replace A' is `4+epsilon`, then you increase the revenue of the network, and if B, C, and D also get replacements as still valuable, then you end up getting paid `(4+a)+(3+b)+(2+c)+(1+d)` in fees, where a,b,c, and d are the respective utilities of ensuring your transaction supplants the alternatives.

I think, as a DDOS prevention mechanism, we should instead impose a minimum increment on the fee; that is, don't bother replacing for 1 lovelace more of income each time.

Also, all of this becomes that much more sensitive in a Leios world, because the mempool determines what suffix gets attached in a Leios endorsement block, so mempools are going to get much much larger, and the data structures needed to maintain them are going to be that much more sensitive to re-evaluating.

fallen-icarus:
> ... you've worsened the UX for users B, C, and D; ...

Are you assuming retail will be building tx chains? Isn't that already a bad UX? Wouldn't using nested txs be better for retail for these scenarios?

I'm expecting bots to be the dominant chain builders and I want to avoid the situation where bots quickly build a tx chain and make it prohibitively expensive for retail to bid for the very first resource usage. Retail needs to easily be able to cut the bot line; bots can easily make B', C', and D' off the retail user's new continuing output. So no revenue is actually lost in your example (if the "users" are bots).

> B, C, and D have already been validated by anyone who puts them in the mempool; that is computational work they should be paid, ...

Agreed! The replacement formula is already doing this by factoring in the _minimum required fee_ for each descendant tx in the chain. The replacement tx's fee needs to cover the work already done for the entire chain it is displacing. By using the minimum fee instead of the actual fee for the descendants, bots can't easily price out retail for Tx A just by quickly building a tx chain on Tx A. @Quantumplation I thought you would like this feature 🙂

> I think, as a DDOS prevention mechanism, we should ... impose a minimum increment on the fee; ...

Agreed! And this is already covered by the minFee(B) term. The minimum increment is the minimum required fee for the replacement tx.

> ... all of this becomes that much more sensitive in a Leios world, ... the data structures needed to maintain them are going to be that much more sensitive to re-evaluating.

When a mempool needs to be re-evaluated after evicting a chain, only Phase 1 needs to be repeated; Phase 2 can be skipped thanks to Cardano's determinism. Why would Leios need to redo Phase 2?

---
So the TLDR:
1. The replacement tx fee needs to cover the work already done for the displaced tx chain + the work required to propagate the replacement tx itself. The latter means +1 lovelace bids aren't valid.
2. By using min fee for the descendants in the formula, retail will be more competitive against fast bots.
3. When a tx chain is evicted, only Phase 1 needs to be repeated in the mempool.

Quantumplation:
My point was about evaluating the new chain of txs that flood in, not the same set of txs not already there.

I'm not up on all the details / haven't thought it through carefully enough, I just know that one of the things the Leios team is very focused on is the design of the mempool to support Leios, because Leios makes it very tricky to use the existing mempool design. Might be worth having a conversation with @ch1bo et al.

I think your proposal is workable from the fee perspective, but by your arguments, if tx builders are happy to pay elevated prices for every step of the chain, why would we turn away that fee revenue? If someone is willing to pay 15*4=60 ada for ABCD (each with a minFee of 1), why would we let someone replace A (and *potentially* forgo BCD) for just 15+4 ada?

What if we have a TX that consumes *two* UTxOs in the mempool, so:

```
A <- B---\
           Z <- W
X <- Y---/
```

And Z has already been bid-up 15 ada, and someone wants to replace X?

fallen-icarus:
> If someone is willing to pay 15*4=60 ada for ABCD (each with a minFee of 1), why would we let someone replace A (and potentially forgo BCD) for just 15+4 ada?

Slight correction: the formula works differently than this. To displace this chain, the replacement tx fee must be at least:

```
15 + 1*3 + minFee(newTx)
```

Then the idea is the auctions will quickly be replayed by bots to recreate the chain on the new root which means the actual revenue is:

```
15 + 1*3 + minFee(newTx) + 15*3
```

I'm assuming here the Cardano tx fee formula already has a profit margin baked in; otherwise, what is the point of being an SPO?

> What if we have a TX that consumes two UTxOs in the mempool, ...

Z would be counted as a descendant if either one gets displaced.

Cerkoryn:
@fallen-icarus I'm curious to know if it would be possible to implement this CIP at different layers, since there are some design patterns that probably wouldn't work as well with it.  Maybe it is possible to have some mempools configured to work this way, and others configured in the typical way.  And then dApps that want this functionality can choose to use it or not?

Or alternatively, maybe it could be handled at the smart contract layer?

I like the idea a lot, but I would hesitate to support a change that made this method a requirement.

fallen-icarus:
@Cerkoryn Can you elaborate on what these "design patterns" are? Maybe there are better ways of implementing those patterns. This is not something that can be solely addressed at the smart contract level because smart contracts cannot compare two txs. It is possible for some mempools to behave differently, but for this to work, most need to respect the auction; otherwise, the outcome won't be predictable enough.

CIP-????/README.md
rphair: CIP meeting today agreed that @Quantumplation's explanation is helpful & that dropping the word `Policy` is OK... it mightn't fix "redundancy" but the extra word isn't necessary either (so committing this: cc @fallen-icarus).

CIP-????/README.md
```diff
@@ -0,0 +1,482 @@
+---
+CIP: ????
```
```suggestion
CIP: 183
```

rphair (commented):
CIP meeting today noted that this policy would suit the progressive _DeFi Kernel_ model of markets with tentative & composable transactions... without any regression in performance for the incumbent "batcher" (relatively centralised) paradigm or any other current design pattern... unless @Quantumplation @Cerkoryn we have collectively missed something in prior discussion.

Therefore it's essential to a well rounded & progressive ecosystem that we consider this a CIP candidate; @fallen-icarus please rename the containing directory to `CIP-0183` and update the README link in your OP 🎉

Cerkoryn:
@fallen-icarus I can't give specific examples and I am unfortunately not dApp developer, so I can't speculate too much.  I just didn't want to jump to the conclusion that all existing dApps would continue working just fine if we change the mempool to use conflict-based fee priority chain-wide.

fallen-icarus:
This is a pretty big update so here are the core changes:
1. Rewrote *Motivation* section to highlight the core problem is that FIFO is not actually fair. It favors well-connected infrastructure. It also leaves money on the table.
2. Simplified the replacement fee formula (just the equation, it is still functionally the same).
3. Switched the descendant tx fees from absolute to minimum required to cover their network costs.
4. Reworked *Rationale* section to clarify things like how existing dApps will be unaffected by this change. And MEV is a dApp level problem for dApps that are affected by this CIP.
5. Added a section to the *Rationale* to explain why this won't result in bursts of bids right before each new block. Waiting until the last minute is a losing strategy.

berewt:
We are working on dynamic pricing at the moment which is orthogonal to this but target a similar problem. We realised that moving transaction in and out of the mempool would actually need a thorough analysis of the impact on mempool.
It isn't a complete blocker but it's a significant roadblock that needs to be considered carefully.
The nice thing though is that as it's mentioned in the CIP, the solution could be opt-in at the node level, and doesn't need to work for every node.
Although, the impact on the mempool design may be a huge challenge. So I second what @Quantumplation said the other day on that matter. It's a smaller problem for Praos than Leios, but Praos shouldn't be the target.

In our case, we had to be careful about the timing of the operation, and to rule out some variant that required too important reorganisation. It's definitely something to consider here as well.

Maybe we can summon @nfrisby who is working intensively on the mempool design of Leios to have his view on it.

fallen-icarus:
@berewt TBH I'm not married to this implementation. My end goal is a UTxO contention fee market because mass-adoption will require one, but we don't need it yet. This CIP was meant to just spark a community discussion. If you guys can come up with a better implementation, I would gladly sunset this CIP in favor of that one!

berewt:
Unfortunately, as I said, we work on another problem: prioritisation in time of congestion. It has some similar constraints, but aren't totally identical. I don't know if we would have time to assess if the mechanism we'll come with will be applicable to contention.
In the worst case, the outcome should be useful to assess feasibility of this one.

polinavino:
Finally getting around to commenting on this - Im one of the people working on tiered pricing, which is, indeed, similar but somewhat orthogonal to this proposal, as @berewt said. A couple of things come to mind :

(1) I am not totally clear on whether is this pre-supposing nested transactions? If not, by what mechanism is the fee for an auctioned-off UTxO supposed to be paid out to the user who put the tx into the block? A priori, it is not possible to know who will make the next block. If this proposal assumes nested transactions are implemented, this makes sense to me: it was sort of designed for auctioning stuff off, UTxOs included. This assumes that tx submitters can build nested txs and block producers are using nexted txs aggregators (which I hope is ok). 

(2) The way I understand the mempool change is that in an incoming transaction is paying more than a conflicting transaction in the mempool, instead of rejecting the incoming one (current policy), the conflicting one is rejected, and then something like the following protocol is executed : start with a Tx list `L` containing only the conflicting transaction paying less for the same UTxO, and go down the mempool list `M`, checking for conflicts pairwise between each pair of `(l , m)` transactions in `L x M`. Move `m` from M to L if there is a conflict. This (or a faster version) could have vastly varying running times in practice, and much analysis would be needed. Perhaps it would make sense to have a provision for abandoning the process in favour of existing queue if there is no time to finish it.

(3) Is transaction bidding on multiple UTxOs rejected automatically? Or only in the case that it conflicts with another one spending the same UTxO (regardless of who pays more)? Automatic rejection would make sense to me because of the complexity of dealing with a network of bids that would ensue

fallen-icarus:
@polinavino 

1. I think you are misunderstanding how this CIP would work. It is just the normal fee mechanism which goes into the epoch's pot to be distributed with the next rewards. *It does not go directly to the producer who includes the tx.* This makes it agnostic to nested txs.
2. The mempool implementation is left out of the CIP, but the tx which pays more must cover all costs for all txs it replaces. So if a tx chain built up in the mempool on top of the dropped tx, the chain needs to be evicted too and the new tx needs to cover the total cost of the chain to prevent a DDoS scenario.
3. No, it is not automatically rejected. Are you sure you aren't over-complicating the "network of bids"? If tx A conflicts with *both* tx b and tx C, tx A must pay more than the sum of tx b and tx c. In practice, this means it will be hard to win auctions when you try taking multiple competitive UTxOs in one tx. If you challenge them one at a time, you might be fine with a small bid, but if they are all in the same mempool together, you will lose the auction. So you need to create a disproportionately large bid just in case. As a result, the revenue from this feature should more than compensate for the complexity of having to track evicting multiple tx chains at once.

polinavino:
(1) Right, i see. Then, id like to point out that this auction will be possible to implement with nested txs regardless whether it exists also via the regular fee mechanism. The nested txs version will be such that a transaction with extra consumed x lovelace will serve as a bet on whatever UTxO (could be indicated via metadata), and is much more tempting for a block producer to engage with over any tx that is paying extra *fees* to spend that UTxO. Paying an extra "tip" to the block producer on top of a competitive fee will likely be the more successful strategy. So i guess what I am saying is that this auction proposal could benefit from a nested-txs aggregator section

(2) Yeah, I am saying that this analysis is needed before people would want to accepts this as a mempool update. Best, average, and worst case.

(3) Right, perhaps I overstated this concern. The revenue from any two transactions spending sets `a` and `b` of UTxOS can always compared, and the one making less money can be kicked out.

fallen-icarus:
1. AFAIU the nested tx approach requires propagating sub-txs which is not always desirable/practical. We need a solution that works for normal txs too. I agree the nested tx approach is complementary and worth pursuing in addition to this one, but not sure it needs to be mentioned in this CIP since it is slightly tangential.

2. Agreed! But I'm not the right person to do that analysis :sweat_smile:

polinavino:
@fallen-icarus for (1) what I am saying is that the mechanism you propose would have to compete with one implemented via nested txs, which may not be as practical because of having to deal with sub-txs, but would make the block producer significantly more money (since the revenue would not be shared across all stake-holders, as it would be for the case of paying fees directly)

also, for (3), just to make sure I understand the principle
`totalBidValue tx1 > totalBidValue tx2` always means that tx1 gets kicked out, right? it doesnt matter that tx2 could be bidding a small amount on many UTxOs, and tx1 bid almost as much as tx2 in total, but on a single UTxO?

fallen-icarus:
1. Fair point. But taking that further, the nested tx approach also competes with Linear Leios. As a block producer, would I rather include an EB or create an RB with the sub-tx fulfilled? Since EB revenue is shared network-wide, the block producer could rationally choose to fulfill the sub-tx directly instead of all the txs in the EB.

3. Do you mean `tx2` gets evicted? If `totalBidValue tx1 > totalBidValue tx2`, then `tx2` gets evicted. It doesn't matter how many UTxOs are being bid for in each tx. All that matters is how many txs are concurrently competing.

polinavino:
1. Yeaaaah
2. Yeah, i got it backwards sorry. The point is that the total amount of "overpayment" above the min fee (i.e. `totalBidValue` , the total amount the tx is bidding on all UTxOs it's participating in an auction for) is what gets the transaction in. It would make more sense if the deciding factor was, instead, based on `totalBidValue / minfee`, since that way, it would be proportional to the amount of block space the transaction is occupying.

rphair (approved):
@fallen-icarus as I was re-reading through the entire review & response history here (since it's almost 3 months since review bandwidth has fallen to `0`) I noted:
* your "major rewrite" https://github.com/cardano-foundation/CIPs/pull/1178/changes/dd2ec66069650567f443306857211c2bb89391e9 seems to have addressed all feedback that came before it;
* all comments since — mainly from developers of intents & tiered pricing — have been inquisitive rather than critical & you appear to have addressed them all & answered all questions;
* those latter reviewers i.e. @polinavino @berewt haven't identified anything that would supersede or challenge this proposal: as stated, "mostly orthogonal" in practice though maybe parallel in motivation.

So this has met the requirements for peer review in my opinion and I would be happy to see this merged (cc @perturbing @Ryun1):
* We have a huge `Last Check` meeting coming up (mostly from my going through all the stalled proposals): https://hackmd.io/@cip-editors/141
* — so although we may not finish merging everything (and we will start with proposals a lot older than this one) please all reviewers post any last-minute reservations if there is any reason this should not be merged tomorrow: cc @Quantumplation @AndrewWestberg @Cerkoryn @colll78

CIP-0183/README.md
```diff
@@ -0,0 +1,423 @@
+---
+CIP: 183
+Title: Conflict-Based Fee Priority in Mempool
+Category: Network
+Status: Proposed
+Authors:
+  - fallen-icarus <modern.daidalos@gmail.com>
+Implementors: []
+Discussions: []
```
Ryun1:
```suggestion
Discussions:
 - https://github.com/cardano-foundation/CIPs/pull/1178
```

Ryun1 (approved):

CIP-0183/README.md
rphair: whoops; @Ryun1 I guess we have come to depend on the validator for this: though this is one of the few remaining boundary cases because it had no commits since the validator was installed (and will check now when I commit this) 😅

perturbing (approved):

PR #1181, opened 2026-04-22
CIP-0108 | Improve JSON schema

Ryun1:
- Adding maxLength restriction to schema, to match the CIP document's restrictions. `Title`s limited to 80 chars and `Abstract`s limited to 2500 chars.
- Enforce required properties of `witness`s in JSON schema
- Enforce uniqueness of `reference` entries in JSON schema
- Clarify `@type` of references in CIP

### DIFF
--- a/CIP-0108/README.md
+++ b/CIP-0108/README.md
@@ -89,6 +89,9 @@ The following properties extend the potential vocabulary of [CIP-100](https://gi
 - We extend CIP-100's references field.
 - This SHOULD NOT support markdown text styling.
 - To be an OPTIONAL set of objects, using the `@set` property.
+- Each object MUST have an `@type` field indicating the kind of reference. This MUST be one of:
+  - `GovernanceMetadata` - when the reference points to another CIP-100 compliant governance metadata document.
+  - `Other` - for any other kind of supporting document (e.g. a blog post, research paper, or community discussion).
 - Each object MUST have a `label` field to describe the reference, such as; "blog - Why we must continue to fund Catalyst".
 - Each object MUST have a `uri` field.
 - Each object MAY have a OPTIONAL `referenceHash` object.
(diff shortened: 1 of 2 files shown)


rphair (commented):
@Ryun1 these updates look like sensible evolution & we should be able to confirm this easily in `Triage` at the next meeting (https://hackmd.io/@cip-editors/133): including a review of general concerns that accompany enforcing fixed length strings (vs. pathological or future scenarios).

Inviting review from CIP-0108 reviewers so far: @gitmachtl @Quantumplation @kderme @disassembler @Crypto2099

gitmachtl:
@Ryun1 requested a feature a while ago so that we can have the schema stuff linked via an external https link in the json file.
This was implemented and is working since new year i think. Maybe we should mention this also in some CIP?

That way, we can host a common "big" schema json on github f.e., and just reference it in the governance action metadata json. Because so far, the biggest mistakes users make is to add fields to the content, but they do not use a proper schema section for that. This relates to information loss during the canonization to calculate the hash for the signatures. A bunch of governance actions in the past and currently live have this problem. Most explorers do not do that check, but cardano-signer and f.e. govtools show an invalid signature.

rphair (approved):
@Ryun1 this was well substantiated at the CIP meeting today, especially through @perturbing's explanations of where these updates were coming from & how they reflect what the tooling is already doing (i.e. enforcing the max length of these fields: so this schema update would be essential to avoid related user error).

Therefore this is ready to merge after @perturbing's confirmation & we can do that out-of-band (effectively with all 3 editors unanimous) or revisit merging @ the next CIP meeting: https://hackmd.io/@cip-editors/134

perturbing (approved):
Thank you for this @Ryun1!

I think it is ready to merge, but just to check, do you want additional feedback from others on this @Ryun1?

PR #1189, opened 2026-05-14
CIP-0186? | Cardano Wallet Deep-Link Signing

realdecimalist:
# Cardano Wallet Deep-Link Signing

([rendered latest document](https://github.com/Flux-Point-Studios/CIPs/blob/feat/cip-deeplink-signing-draft/CIP-0186/README.md))

## Summary

This CIP defines a deep-link wire protocol that lets a native mobile dApp ask an installed Cardano wallet to perform CIP-30 operations (chiefly `signTx` and `signData`) over the OS's URL-handling mechanism — no relay server, no embedded WebView, no QR, no WebSocket. It closes the gap CIP-30's `window.cardano` injection contract cannot reach: mobile-native dApps running outside a wallet-controlled browser context.

## Solution-To: CPS-0010

[CPS-0010 (Wallet Connectors)](https://github.com/cardano-foundation/CIPs/tree/master/CPS-0010) documents that CIP-30 is a JavaScript-injection contract over a shared browser `window` object, and explicitly enumerates "mobile-native dApps" as an unaddressed connector shape. This proposal addresses that gap by specifying a transport profile for CIP-30's existing method surface — `signTx`, `signData`, `getUsedAddresses`, `getRewardAddresses`, `getNetworkId`, with optional `submitTx` and a reserved CIP-95 `signGovernance` profile — over OS deep-links (Universal Links on iOS, App Links on Android), with an X25519-encrypted payload and a BLAKE2b-256 commit binding so the wallet can prove which transaction it actually signed. The CIP does not replace CIP-30; a wallet that implements both serves browser-extension, in-app WebView, and mobile-native dApps with one signing engine.

## Reference Implementations

Two reference implementations are in active development under Flux Point Studios, alongside a companion repo for context:

- **[`cip30-deeplink-client`](https://github.com/Flux-Point-Studios/cip30-deeplink-client)** — dApp-side TypeScript SDK (Apache-2.0). Public repository created at draft submission; README and anticipated public surface are published, with source landing in Wave 2. The SDK is consumed by the Aegis iOS app (Aegis itself is closed-source; only the on-chain validator suite is public — see below).
- **[`yuti`](https://github.com/Flux-Point-Studios/yuti)** — wallet-side reference implementation (Flutter, iOS + Android + PWA, Apache-2.0). Yuti will implement the URL handler in Swift (iOS) and Kotlin (Android), plus a `lib/services/cip30_deeplink_service.dart` adapter that decrypts the payload and routes to Yuti's signing engine. The receiver patch is estimated at ~290 lines across 8 files; this is the patch surface we propose Lace / Eternl / Vespr / Typhon / Yoroi / SecondFi could adapt against their existing signing engines.
- **[`aegis-contracts`](https://github.com/Flux-Point-Studios/aegis-contracts)** — companion repo only (not part of this CIP). The on-chain Aiken validator suite for Aegis, V12.2 live on Cardano preprod. Listed for completeness so reviewers can see the production context the dApp-side SDK is consumed from. The CIP itself is a wallet/dApp transport, with no on-chain surface.

Reference implementation status (transparent — both impls are pre-shipping at the time of this PR open):

- dApp-side SDK (`cip30-deeplink-client`): repo public, README populated; source targeted for the next Aegis mobile wave (4-6 weeks).
- Yuti receiver: repo public, license + security policy + CIP cross-references in place; receiver patch targeted in parallel (4-6 weeks for receiver, +2-4 weeks for signing-engine integration).
- First end-to-end `signTx` round trip on preprod: targeted within 8-12 weeks of this PR opening.

These dates are well inside the 6-month median Cardano-Foundation review window for new CIPs, so the spec will have empirical preprod evidence by the time editors are ready to merge. We chose to open the Draft PR early to surface the design to editors and other wallet teams as soon as possible, not to claim premature implementation.

## Prior art surveyed

- **EIP-831 (Ethereum) — URI envelopes**: <https://eips.ethereum.org/EIPS/eip-831>. One-way URIs only; no encryption, no callback. Useful as a URI-syntax precedent.
- **Solana MWA 2.0 (Mobile Wallet Adapter)**: <https://github.com/solana-mobile/mobile-wallet-adapter>. Uses P-256 ECDH + WebSocket + reflector. Optimised for high-throughput in-app gaming (Saga phone); their 2024 retrospective recommended Safari Web Extensions for low-throughput dApps — which is exactly the shape this CIP targets, with the Phantom/Petra deep-link pattern instead.
- **Phantom Deep-Links (Solana)**: <https://docs.phantom.com/phantom-deeplinks/provider-methods>. Universal Links plus NaCl box (X25519 + XSalsa20-Poly1305). Mainnet evidence across years of production. The encryption envelope and `dappKey`/`walletKey`/`nonce`/`payload` URL shape proposed here mirrors theirs.
- **Petra Deep-Links (Aptos)**: <https://petra.app/docs/deep-link-api>. Same construction as Phantom, custom-scheme transport. Confirms the design generalises beyond Solana.
- **BIP-21 (Bitcoin URIs)**: <https://github.com/bitcoin/bips/blob/master/bip-0021.mediawiki>. Bare-URI payment intents, no signing roundtrip — included as a "what NOT to do" precedent for signing flows.
- **CIP-45 (Cardano WebRTC)**: <https://github.com/cardano-foundation/CIPs/tree/master/CIP-0045>. The closest existing Cardano connector for mobile, but assumes a WebTorrent tracker as relay. This CIP is complementary — WebRTC remains the right answer for desktop-browser-to-mobile-wallet; deep-link is the right answer for mobile-to-mobile on one device.
- **CIP-0162 (Cardano URIs - DRep Links)**: <https://github.com/cardano-foundation/CIPs/tree/master/CIP-0162>. Recently merged. Establishes precedent for the `web+cardano` URI family. This CIP intentionally does NOT use that family because the deep-link target is a wallet vendor's Universal Link domain (`https://lace.io/...`), not a Cardano-namespace URI.

1. **`submitTx` MUST vs MAY?** The current draft marks `submitTx` as OPTIONAL on the wallet side because dApps with their own submission backend (the Aegis case) do not need it. Should we instead require it so dApps without infrastructure can rely on the wallet? Tradeoff: required `submitTx` adds an attack surface (wallet now relays arbitrary signed tx) but eases dApp authors.

2. **CIP-95 governance in v1 or v1.1?** The draft reserves the method name `signGovernance` and references the CIP-95 surface, but does not fully specify the URL envelope for vote delegations, DRep registrations, or governance proposal data signatures. Should v1 cover these too, or is a follow-up profile (v1.1) once CIP-95 itself reaches Active the right call? Editor preference here would help us scope.

3. **URL length budget — 12 KB the right ceiling?** The draft mandates wallets accept URLs up to 12 KB and MAY refuse larger. A typical Aegis insurance buy CBOR is 1.2-3 KB pre-encoding; multi-asset txns can push 6 KB. iOS Safari and Android Chrome both handle 32 KB, but Universal Link previews truncate around 8 KB. Is 12 KB the right pragmatic ceiling, or should the spec mandate 16 KB / 24 KB to accommodate complex Plutus transactions? (We can pre-publish reference scripts via CIP-31 to shrink, but smaller dApps may not.)

4. **Wallet registry — central vs decentralised?** Appendix B of the draft proposes a central `wallet-registry.json` maintained in this CIP's folder, listing each wallet's `id`, `scheme`, and `https` prefix. An alternative is to require each wallet to publish its own `apple-app-site-association` / `assetlinks.json` at a CIP-defined well-known path and have dApps probe at runtime. The central registry is faster to iterate but adds CIP-editor maintenance burden. Editor preference?

5. **iOS custom-scheme fallback — keep or drop?** The draft retains `cip30dl:` as a fallback for the install-but-no-AASA-yet edge case. Custom schemes have a well-documented hijacking risk (any app can register the scheme after install). The fallback is rate-limited by the `redirect` host binding, but a strict-Universal-Links-only profile is also viable. We lean keep-as-defence-in-depth; reviewers may disagree.

## Test plan

The spec will be validated empirically by the two reference implementations end-to-end. Test cases (all to be exercised on Cardano preprod before any Active-status push):

- **Aegis ↔ Yuti happy path** (`connect` → `signTx` → `getUsedAddresses` → `disconnect`) — to be captured in a published Aegis ↔ Yuti integration test once both impls land.
- **Replay rejection** — same `(dappKey, nonce)` tuple is replayed inside `ttl + 600s`; wallet MUST return `errorCode -5 NonceReplay`.
- **Commit-mismatch rejection** — dApp builds CBOR, wallet receives a different CBOR with the original `commit` URL parameter; wallet MUST return `errorCode -2 CommitMismatch` BEFORE rendering signing UI.
- **Redirect host-binding rejection** — second app re-uses captured `session` token with a different `redirect` host; wallet MUST return `errorCode -4 RedirectHostMismatch`.
- **TTL-expired rejection** — `ttl < now`; wallet MUST return `errorCode -6 TtlExpired`.
- **Conformance suite** — to be published under `cardano-foundation/cip-deeplink-conformance` once draft merges. Suite is the Aegis ↔ Yuti integration harness, generalised to drive any wallet under test.

Target Path-to-Active metrics, repeated from the CIP body:
- 1 production wallet ships on iOS App Store + Google Play.
- 1 production dApp ships a mobile-native client with ≥ 1,000 mainnet transactions signed via this protocol over a 30-day window.
- Conformance suite published and passed by all implementing wallets.
- A second wallet interoperates with the same dApp without dApp-side code changes.

## Co-author opportunity

The draft is intentionally aligned with the Phantom / Petra protocol shape so the wallet-side patch surface inside an existing signing engine is small (~400 lines around a URL handler). We explicitly invite the **Lace, Eternl, Vespr, Typhon, and Nami** teams to be listed as co-authors in v1.1 once they review the spec and confirm the patch surface is workable inside their codebases. Discussions are open at `contact@fluxpointstudios.com` and on this PR thread. Our goal is to ship v1.1 with all five wallets as named co-authors and at least two of them as named Implementors in the preamble.

## Out-of-band

- The Aegis team will summarise this draft at the next CIP editors' biweekly meeting (HackMD agenda link to be added once the meeting page is open).
- Cardano Forum thread for pre-PR discussion (already open): <https://forum.cardano.org/t/cip-proposal-mobile-deep-link-signing-for-native-dapps-cip-30-extension/154561>. The same URL is set on the CIP's `Discussions:` front-matter field.
- Reference impl repos and the conformance suite repo will move from `Flux-Point-Studios/*` to a vendor-neutral home under `cardano-foundation` if editors prefer; today they live where the maintainers do.

Thank you for your time on review.


rphair:
@realdecimalist I was just reviewing this & then your forced push came in.  I'll save those review points for later & assume this is a "work in progress" until you can confirm you've done the last of those forced pushes.

(Generally we recommend forced pushing since we've got potentially any GitHub user submitting change suggestions, often with linked commits created on GitHub that would get overwritten if pushing a branch without the GitHub commit history.)

Looking forward to a quick "format" review (you can double check [CIP-0001](https://github.com/cardano-foundation/CIPs/tree/master/CIP-0001) (as updated a few times since the CIP-0030 early days) for proper headings in the meantime).  We'll mark this for `Triage` and tag some potentially interested reviewers after we can be sure your content has initially stabilised.

realdecimalist:
oh, yes. my bad. was just cleaning up the two commits into a single clean commit. finalized at this time.

CIP-deeplink-signing/examples/.gitkeep
```diff
@@ -0,0 +1,6 @@
+# Reference implementations will be linked here in Wave 2.
+# Planned contents:
+#   - dapp-side TypeScript client snippet (extracted from aegis frontend)
+#   - wallet-side Swift URL handler (iOS reference)
+#   - wallet-side Kotlin URL handler (Android reference)
+#   - conformance test runner README
```
rphair: It sounds like you are already aware that code assets in CIPs would be samples, examples, snippets, tests, etc. while the bulk of any implementation would live in a repository belonging to the author / maintainer / implementor (since CIP editors can't be any of these things).  So one review point will be to double-check that the level of code detail included here is only informative, rather than the complete body of code.

CIP-deeplink-signing/README.md
```diff
@@ -0,0 +1,1251 @@
...
+CIP: "?"
+Title: Cardano Wallet Deep-Link Signing (CIP-30-DeepLink)
+Category: Wallets
+Status: Proposed
+Authors:
+    - "Nathaniel \"decimalist\" Minton (Flux Point Studios) <nathanielminton@fluxpointstudios.com>"
+Implementors: []
+Discussions:
+    - https://forum.cardano.org/t/cip-proposal-mobile-deep-link-signing-for-native-dapps-cip-30-extension/154561
+    - https://github.com/cardano-foundation/CIPs/pull/1189
```
rphair:
```suggestion
Discussions:
    - https://forum.cardano.org/t/cip-proposal-mobile-deep-link-signing-for-native-dapps-cip-30-extension/154561
    - Original PR: https://github.com/cardano-foundation/CIPs/pull/1189
```
Note you can take out the Forum link if nobody else responds after [my initial comment](https://forum.cardano.org/t/cip-proposal-mobile-deep-link-signing-for-native-dapps-cip-30-extension/154561/2).  However I would be happy to see some public commentary on it & if so then the Forum link would be welcome: especially if it includes suggestions that are incorporated into the CIP.

A CIP process improvement in the meantime is that GitHub & downstream sites are confirmed to support labels for these links... the only "standard" one is `Original PR:` (which all CIPs & CPSs have) but you are welcome to provide short labels for other links as well (documentation in CIP-0001 forthcoming when this convention is soon finalised).

CIP-deeplink-signing/README.md
```diff
@@ -0,0 +1,1251 @@
...
+over the operating system's URL-handling mechanism, with no relay server, embedded WebView, QR code, or
+WebSocket. The protocol layers a small set of request-response endpoints onto a single namespaced URI
+prefix (`https://<wallet-domain>/cip30dl/v1/...`, with `cip30dl:` as a platform fallback), pairs the dApp
+and wallet with an ephemeral X25519 key exchange, binds every signing request to a BLAKE2b-256 commit
+over the canonical transaction body so the dApp can verify *which* transaction was signed, and returns
+the witness set via a universal-link callback. The spec is profiled for iOS (Universal Links preferred,
+custom scheme fallback) and Android (App Links plus Intent disambiguation). It is a strict subset of
+CIP-30's method surface so wallets can reuse their existing signing UI.
+
+## Motivation
```
rphair:
```suggestion
## Motivation: Why is this CIP necessary?
```
(1 of 2) Long titles for 2 sections have been required for some years now: since otherwise content often ends up on the wrong side of the boundary.

CIP-deeplink-signing/README.md
```diff
@@ -0,0 +1,1251 @@
...
+  UI like `"highlightedFeeBreakdown": [...]`) that wallets MAY display but MUST NOT treat as
+  load-bearing.
+
+This split is the load-bearing reason a wallet can refuse a URL with an unrecognised key and still
+support forward-compatible application-layer evolution. It mirrors how TLS handles unknown
+extensions (ignore unless critical-bit set), how HSTS handles unknown directives (ignore), and how
+HTTP/3 frames handle unknown frame types (drop without close). The CIP rejects the unqualified
+Postel rule because every URL key here is in the critical set.
+
+## Rationale
```
rphair:
```suggestion
## Rationale: How does this CIP achieve its goals?
```
(2 of 2)

CIP-deeplink-signing/README.md
```diff
@@ -0,0 +1,1251 @@
...
+     |        aegis://cb?response=approved                      |
+     |          &nonce=...&payload=enc(witnessSet) )            |
+     <-----------------------------------------------------------+
+     |                                                          |
+     |  5. dApp decrypts, re-verifies commit, assembles signed  |
+     |     tx, POSTs to its own /api/tx/submit                  |
+     |                                                          |
+```
+
+## Versioning
```
rphair:
```suggestion
### Versioning
```
This approach to "versioning" definitely reads like part of the specification and so naturally would appear normally here within the _Specification_ section.  An H2 / `##` would be a major heading and therefore orphan the `### Unknown-key policy` section immediately below it.

rphair (commented):
thanks @realdecimalist for presenting what looks like a promising architecture.  Though we have CIP editor @Ryun1 with years of experience evaluating CIP-0030 extensions, bridges, performance considerations, etc., we should also have a chorus of commercial wallet developers reviewing this proposal.

I'm glad to hear [on the Forum](https://forum.cardano.org/t/cip-proposal-mobile-deep-link-signing-for-native-dapps-cip-30-extension/154561/3) that you intend to come & help introduce this at the next CIP meeting (https://hackmd.io/@cip-editors/135).  We would start by confirming that the approach is sound... then formulate some tagging / outreach targets for circulation & review of the proposal.

These are just format corrections & organisational points in the meantime:

CIP-deeplink-signing/README.md
realdecimalist: addressed in 78f79a6. added the `Original PR:` and `Forum thread:` labels per the forthcoming CIP-0001 convention. keeping the Forum link in since we'd like the public commentary surface too.

CIP-deeplink-signing/examples/.gitkeep
realdecimalist: addressed in 78f79a6 `.gitkeep` now explicitly states the folder is informative snippets only, with bulk impls in author repos per CIP-0001. All three reference-impl repos linked from the `.gitkeep` so reviewers can navigate directly.

rphair:
@realdecimalist there's been a strong expression of support from wallet developers about this idea [here on Discord](https://discord.com/channels/971785110770831360/992011119872970762/1504795397468717137) (see [CIP Wiki front page](https://github.com/cardano-foundation/CIPs/wiki) for Discord server invitation) (cc @aljosa) and I've followed up there by suggesting the CIP Discord [#wallets-working-group](https://discord.com/channels/971785110770831360/1176612265894613133) for high-bandwidth conversations: if devs want to verbosely discuss the particulars of this scheme before bringing back resolutions here to GitHub review (cc @Ryun1).

CIP-0186/README.md
```diff
@@ -0,0 +1,1264 @@
...
+`Content-Type: application/json` and `Cache-Control: max-age=86400`). The manifest is the
+authoritative source for the wallet's CIP-30-DeepLink commitments and supersedes the human-readable
+registry entry in Appendix B for runtime validation. Schema:
+
+```json
+{
+  "$schema": "https://github.com/cardano-foundation/CIPs/blob/master/CIP-deeplink-signing/cip30dl-attestation.schema.json",
+  "wallet_id": "<lowercase ASCII id matching Appendix B>",
+  "protocol_versions": ["1"],
+  "methods": ["connect", "signTx", "signData", "getUsedAddresses", "getRewardAddresses", "getNetworkId"],
```
Ryun1, line 717: Have you considered support for CIP-30 extensions too?

i.e. [CIP-0095](https://github.com/cardano-foundation/CIPs/tree/master/CIP-0095)

CIP-deeplink-signing/README.md
```diff
@@ -0,0 +1,1264 @@
+---
+CIP: "?"
```
rphair:
```suggestion
CIP: 186
```

rphair (commented):
@realdecimalist this was accepted (before you arrived) as candidate at the CIP meeting today based on a brief review: including appreciation that the link resolution into CIP-0030 — especially in mobile platforms where "favourite" application launchers are generally designated — is something long awaited by the wallet dev community & will open the door to a much better UX.

Associations with CIP-0013 / CPS-0016 were also immediate & so I think it will help if you keep explaining (as you already have in the CIP) how this proposal will be relevant to applicability & performance of Cardano URI Scheme links.

We also expressed the hope that @Ryun1 and other standards advocates in the wallet field will be able to poll for review & feedback in both public & private groups on the subject (it's already been introduced in the CIP Discord's [`#wallets-working-group`](https://discord.com/channels/971785110770831360/1176612265894613133) and interested parties are welcome for lightweight review there).

In the meantime please update the CIP number to `186` in any extension names where it may be applicable and rename the containing directory to `CIP-0186` in both the branch and in link pathnames in the OP above. 🎉

realdecimalist:
Two reference deliverables landed today, both anchored on this PR:

**1. Conformance test suite — added in commit [\`6311533\`](https://github.com/Flux-Point-Studios/CIPs/commit/6311533)**

Just pushed to this PR. `CIP-0186/tests/` now ships:
- 57 vectors / 34 negative across 13 spec sections (deep-link decode, base64url strict, attestation manifest, wallet registry, CBOR commit/extract, witness splice, canonical-subject, replay/nonce, error responses, response envelope, sign-data, network, addresses)
- `verify.ts` — TypeScript verifier any wallet author can plug their adapter into (`npx tsx tests/verify.ts`)
- `reference-impl-stub.ts` — minimal correct reference stub (57/57 passing)
- `tests/README.md` — adapter contract for wallet authors

**2. First wallet reference implementation — Yuti**

Open at https://github.com/Flux-Point-Studios/yuti/pull/24. Flutter / `cardano_flutter_sdk` based. 49 tests / 21 negative, CI green:
- Native deep-link receiver (iOS `cip30dl-yuti://` + Universal Link / Android intent filter + autoVerify App Link)
- ABNF-strict URI parser, all 16 error codes
- Attestation response builder with canonical-subject hashing
- Signing engine over `CardanoWallet.signTransaction` (real signing, no stubs — WitnessSet CBOR round-trips through `cardano_dart_types`)

**Spec ambiguities surfaced (Yuti + test-vectors converge on most of these):**

1. **Appendix C.1 nonce example is wrong** — shown as `AgICAgICAgICAgICAgICAgICAgICAgICAgIC` (36 chars → 27 bytes); correct value for 24 bytes of `0x02` is `AgICAgICAgICAgICAgICAgICAgICAgIC` (32 chars). Caught by both impls independently — Yuti's CI flagged it via test-vector decode; conformance suite's `base64url_001_valid_no_padding` uses the corrected form.
2. **`signature` / `walletKey` / `response` / `errorCode` / `errorMessage` not in ABNF `key` production** — §Response signing requires these on every response URL, but the URI-format ABNF rejects them as unknown keys under strict-decode. ABNF should be extended to enumerate response-only keys.
3. **`aead` not in ABNF `key` production** — §Methods preamble describes `aead=xsalsa20poly1305` / `aead=chacha20poly1305-ietf` URL params on `connect` but ABNF doesn't list `aead`. Same issue class as (2).
4. **§Callback host binding `-4` destination ambiguity** — spec doesn't say which host receives the rejection envelope when `redirect` host differs from `connect`-time binding.
5. **§Methods `result-json` `commit` echo invariant** — spec shows the field but doesn't state that the dApp MUST verify the echo. Suggest promoting to MUST as defence-in-depth.
6. **Witness-set merge for `partialSign=true`** — spec inherits CIP-30's flag verbatim but doesn't specify array-append semantics on map key 0 (vkey_witnesses). Suggest one paragraph.
7. **`extractTxBody` literal vs re-serialise** — spec mandates "canonical re-serialise" of element 0, but real Conway bodies must be the **literal** input bytes (BLAKE2b-256 over them == chain's `tx_id`). Suggest restating as "extract the literal tx_body bytes" with a note that the source CBOR is expected to already be canonical.
8. **§Response signing port inclusion** — spec doesn't say whether explicit port is part of canonical authority. Yuti includes it per RFC 3986 §3.2; suggest making this explicit.
9. **§Conway tx body extraction reference encoder** — spec says "canonical re-serialise" but doesn't pin a reference encoder; libraries vary on deterministic-only vs strictly-deterministic CBOR.

Happy to open separate spec-edit PRs for (1)–(9) once you've triaged. The conformance suite + Yuti PR both encode the implementation choice we landed on for each, so editors can see the test fixtures alongside the suggested wording.

rphair (commented):
thanks @realdecimalist I think this is showing the validation that we would need to merge this readily (once changes are stabiliised) and maybe even mark as `Active` if other wallets have implemented this by then.  Regarding the last comment
> Happy to open separate spec-edit PRs for (1)–(9) once you've triaged

... if you're talking about separate pull requests _on the CIPs repository_ that would be too much detail for the CIP process to accommodate here: that work would generally be done upstream & then only changes that affect the documented specification would go into CIP changes reviewed & committed here in `this` PR.
* Since you've moved some of the reference implementation detail out into your own repository then hopefully these updates would be minimal at this stage.
* If you're changing too much in the CIP over changes in the reference implementation that could continue to change, it may be a sign that more detail needs to be the external location rather than here (since we'd like to avoid the document becoming unmaintainable if constant CIP updates are required beyond regular "version" updates).
* I am trusting that co-editor @Ryun1 has a good level of experience in how development in this field affects the body of CIP material so (again) I would trust his discretion on this partitioning between repo & CIP material & would look forward to his own review of this issue.

**p.s.** so TL;DR separate _commits_ here would be fine for editors & reviewers to assess each of the 9 additions.

realdecimalist:
9-ambiguity bundle patch landed in commit [\`f76b5bf\`](https://github.com/Flux-Point-Studios/CIPs/commit/f76b5bf). 77 insertions / 25 deletions, all surgical, no semantic re-litigation outside the 9 listed items. Summary per ambiguity:

1. **Appendix C.1 nonce** — 36-char value → 32-char `AgICAgICAgICAgICAgICAgICAgICAgIC` (24 bytes). Fixed in both the constant listing and the request-URL example.
2. **ABNF response-side keys** — split `key` into `request-key` and `response-key` productions. Adds `response`, `walletKey`, `nonce`, `payload`, `signature`, `errorCode`, `errorMessage` to the response side.
3. **ABNF `aead`** — added to `request-key` production (was missing despite being referenced in §Methods AEAD envelope).
4. **§Unknown-key policy** — now references both ABNF productions so the strict-reject rule applies symmetrically to wallets parsing requests and dApps parsing responses.
5. **§Callback host binding `-4` destination** — `errorCode=-4 RedirectHostMismatch` envelope MUST be delivered to the `connect`-time persisted `redirect`, NOT the offending request's `redirect`. Closes the cross-host phishing defeat where -4 would otherwise sink into the attacker's domain.
6. **§signTx `result-json.commit` echo** — promoted to MUST: dApp verifies echo equals request commit; mismatch is `errorCode=-2`. Defence-in-depth complement to the wallet's pre-render commit check.
7. **§signTx `partialSign=true` witness merge** — added paragraph on array-append semantics on map key 0 of `transaction_witness_set` (plus keys 1, 3, 4, 5, 6, 7 for native scripts / Plutus / datums / redeemers). Replace-semantics silently drops prior co-signer witnesses.
8. **§Conway tx body extraction** — literal-bytes rule. Wallet MUST slice the source CBOR's element-index-0 byte range; MUST NOT re-encode. Chain's `tx_id` is BLAKE2b-256 over source bytes; wallet re-canonicalisers that differ from the source encoder break every downstream hash-pin.
9. **§Conway tx body / §Response signing port** — pinned reference encoders (cardano-ledger Haskell, pallas-codec Rust, cardano-multiplatform-lib JS/Rust). Added explicit port-inclusion rule to §Response signing canonical-subject step 2: include port iff literal in URL authority; default ports MUST NOT be emitted.

Each patch matches an Issue Agent G's conformance suite (#6311533) or Yuti's reference impl (Flux-Point-Studios/yuti#24) independently surfaced during implementation. The conformance vectors and Yuti's test fixtures already encode the post-patch semantics, so existing vectors validate as-is.

Ready for editorial review.

rphair:
@realdecimalist - editorially (my opinion) the CIP after https://github.com/cardano-foundation/CIPs/pull/1189/commits/f76b5bf0418dbd04bc5beb794660ebfa3f46f175 seems of equally great value & validity with those "ambiguities" addressed: although still the same technical challenge to review as before (i.e. we need developer reviews more than editors at this stage).

I certainly think this commit should stay in the branch, and it's just as well devs & architects + dApp implementors weren't tagged all around yet (as suggested in https://github.com/cardano-foundation/CIPs/pull/1189#pullrequestreview-4368082074) before this latest change; so maybe you & @Ryun1 can wait on that tagging until after it's posted here that these commits are finished for now.

realdecimalist:
@rphair apologies for the single bundle commit — I should have split it per-ambiguity from the start. Done now: the previous `f76b5bf` is force-replaced with 9 commits (`c2cce5c` through `a0f6106`), one per ambiguity, byte-identical final tree. Each commit message states the spec gap and the rationale so you and @Ryun1 can ack or kick back individual issues without blocking the others.

Going forward, taking on board what you flagged:

- **CIP changes stay in this PR**, not separate PRs upstream. Understood.
- **No further CIP changes** unless they're genuine spec gaps (not Yuti-impl detail). The 9 patches were the spec-level set we found during the first reference implementation; subsequent Yuti work will live in `Flux-Point-Studios/yuti` and only surface here if it reveals a true ambiguity in the spec text.
- **Patches stable as of `a0f6106`** — no more incoming. Whenever you and @Ryun1 are ready to tag broader dev/architect/dApp-implementor reviewers per [your prior review](https://github.com/cardano-foundation/CIPs/pull/1189#pullrequestreview-4368082074), this is the stable point.
- **Deferring to @Ryun1's discretion** on what stays in the CIP vs what belongs in the reference impl repo. Happy to move material out if anything in the current spec reads more like implementation detail than normative protocol.

On the wallet-implementation gate you mentioned for Active: outreach is in flight on the side via Discord [`#wallets-working-group`](https://discord.com/channels/971785110770831360/1176612265894613133) and direct DMs (already engaged with Alex Dochioiu / VESPR, since Yuti uses VESPR's `cardano_flutter_sdk`). Will report back if any wallet team commits to implementing.

Thanks for the patience and the editorial steer.

rphair:
@realdecimalist we briefly reviewed this at the CIP meeting today:
* @Ryun1 is still preparing to tag some specialist reviewers in the wallet space (both here & in working groups) to see if they can provide deeper review with regard to their own experience & projects.
* Without any further comment from @Ryun1 about the level of detail here vs. the reference implementation, there have been no objections to it: though it should still be a review point for the devs (confirming the CIP _text itself_ is workable).
* @Crypto2099 mentioned a similar ~~workaround~~process for link handling might be in use by one of the CF projects: Veridian, if I remember correctly; @Kammerlo would you know anything about such a thing?  (cc @mpizenberg)

Ryun1:
tagging a few people who may be interested in this proposal for review
@fabianbormann @Scitz0 @AlexDochioiu @refi93

mpizenberg:
> a similar workaround for link handling might be in use by one of the CF projects

@rphair I’m not sure about a "workaround" but CF closest involvement on this topic is in CIP-158, which is mentioned in this CIP I believe, though haven’t had the time to read.
https://x.com/YetAnotherAiGuy/status/2056345144769593800
https://cips.cardano.org/cip/CIP-0158

Kammerlo:
> * @Crypto2099 mentioned a similar workaround for link handling might be in use by one of the CF projects: Veridian, if I remember correctly; @Kammerlo would you know anything about such a thing?  (cc @mpizenberg)

@rphair Sorry, forgot to answer here. I checked with the Veridian team and there isn't something similar.

realdecimalist:
@rphair @Ryun1 — following your note to flag when the commits are settled: the spec changes I had queued are in, and I consider the text stable for review again.

The one substantive change since the ambiguity-resolution commits is the **`connect` handshake**. It was previously delivered unsigned — the dApp adopted the session on trust — and is now a signed, replay-bound first-contact handshake: the response carries `method=connect` (domain separation), an `echo` of the request nonce, and the session-key Ed25519 `signature` over the canonical subject, all verified before a session is seated. That's consistent with the "every response is signed" rule already in §Response signing and it closes the only unauthenticated leg. It's in commit `5adc79c`.

Per your guidance, only this spec-affecting change went into the PR; the reference-implementation detail (the dApp SDK, the conformance vector suite, and the wallet-side reference) stays in the external repos. I don't have further CIP edits queued, so the document is stable for the tagged reviewers to assess.

On the interop front, in case it's useful for review: a second, independent wallet (Gero) has implemented the signed `connect` and it round-trips end-to-end against the reference dApp SDK; Yuti remains the wallet-side reference. So the handshake now has two independent implementations agreeing on the wire format, and there's a conformance vector suite reviewers can test against.

CIP-0186/README.md
realdecimalist: Yes — carrying CIP-30 extensions is a core design goal, and CIP-95 specifically is covered in the **§CIP-95 governance profile (RESERVED)** section.

The principle is that CIP-186 is a *transport* for CIP-30, so extensions ride on the existing envelope rather than needing the transport re-specified per extension. Concretely for CIP-95:
- `api.cip95.signData` (accepting a `DRepID`) and the Conway certificate types in `api.signTx` need **no new wire method** — a CIP-95 wallet accepts the `DRepID` in the `signData` request's `addr` field (CIP-95:421) and recognises Conway certs inside the existing `signTx` encrypted CBOR (CIP-95:328).
- The read-only additions (`getPubDRepKey`, `getRegisteredPubStakeKeys`, `getUnregisteredPubStakeKeys`) are reserved as backwards-compatible new methods under the same `v=1` envelope, gated by a capabilities advertisement in the `connect` session JSON.

More generally, the method ABNF is open and the unknown-key policy — strict at the URL layer, lenient inside the encrypted JSON payload — is what lets CIP-30 extensions slot in without changing the transport.

I've kept the concrete CIP-95 read-method definitions as RESERVED rather than fully specifying them here, on the principle that this CIP shouldn't re-spec each extension it carries. Happy to make them normative in this PR if you'd prefer — your call on the right level of detail.

realdecimalist:
@mpizenberg you're right that it's referenced — CIP-158 is covered in the **§Relationship to CIP-13 `web+cardano:` URI scheme** section. The short version is that the two are complementary rather than overlapping:

- **CIP-158 (`browse` authority)** is navigation-only — by its own Rationale it's explicitly out of scope for signing flows and callback paths. It navigates to a dApp; it doesn't carry a signing request or response.
- **CIP-186** is the signing transport: it carries the CIP-30 `signTx` / `signData` request to the wallet and the encrypted, signed response back over a deterministic deep link.

So a dApp could be *reached* via CIP-158's `browse` and then *sign* via CIP-186 — they sit at different layers. That's probably why "workaround" doesn't quite fit: it's more that no existing `web+cardano://` authority covers the signing round-trip.

The CIP also leans on CIP-158 as a precedent: the same section notes a future minor revision MAY define a CIP-13-grammar-compliant alternate form `web+cardano://cip30dl/v1/<wallet-id>/<method>?<params>`, mirroring CIP-158's `browse` grammar as the established CIP-13 extension shape and registered through the CPS-16 authority registry to avoid collisions. The *primary* URI deliberately routes to one named wallet on the wallet's own domain instead of a shared `web+cardano:` authority — specifically to avoid the OS wallet-chooser on every link — with the rationale in that same section.

No rush on reading at all — but if you get to it, your eyes on how the CIP frames the CIP-13 / CPS-16 relationship would be very welcome, given how close you've been to that area.

CIP-0186/README.md
```diff
@@ -0,0 +1,1366 @@
+---
+CIP: 186
+Title: Cardano Wallet Deep-Link Signing (CIP-30-DeepLink)
```
rphair:
```suggestion
Title: Cardano Wallet Deep-Link Signing
```
The extra title does look like a useful familiar term, but keeping it also in the CIP title is too much detail for this generally brief proposal name... especially because the title is already unique without it & should be findable from any external `CIP-30-DeepLink` reference.

CIP-0186/README.md
```diff
@@ -0,0 +1,1366 @@
...
+Category: Wallets
+Status: Proposed
+Authors:
+    - "Nathaniel \"decimalist\" Minton (Flux Point Studios) <nathanielminton@fluxpointstudios.com>"
```
rphair:
```suggestion
    - Nathaniel "decimalist" Minton (Flux Point Studios) <nathanielminton@fluxpointstudios.com>
```
@realdecimalist I pointed this out in an earlier review and still insist the single quoted string will be a problem vs. the general email/URL contact format expected downstream.

@Ryun1 do you know why the CIP Validator isn't invoked on this PR (not workflow railing to run... it's not even attempting to run it)?

CIP-0186/README.md
```diff
@@ -0,0 +1,1366 @@
...
+  external wallet target because (a) Alex Dochioiu's team already publishes an open-source SDK,
+  (b) Vespr's signing engine separation matches the patch shape we recommend, and (c) the team is
+  responsive to CIP-aligned work.
+- Stage 3 (T+12 weeks): Lace, Eternl, Typhon, Yoroi, and SecondFi review the draft against their
+  signing engines; feedback merged as a v1.1 revision.
+- Stage 4 (T+20 weeks): At least two production wallets ship the protocol. dApp ecosystem onboarded
+  via a `CIP-30-DeepLink for dApp builders` guide.
+- Stage 5 (T+24 weeks): Active.
+
+## Reference Implementation
```
rphair:
```suggestion
## Appendices

### Reference Implementation
```
So you don't have to shuffle your document around, let me propose a fix here for the 2 ways your CIP diverges from the standard format & would fail validation of that format if the GitHub workflow were running here (currently not):
* the CIP standard has no top level section _Reference Implementation_ (generally it's part of the Specification)
* a top level section containing appendixes is either called _Appendix_ or _Appendices_ since it would be problematic to validate, or link later to, those section names with arbitrary text afterward.

CIP-0186/README.md
```diff
@@ -0,0 +1,1366 @@
...
+  plus the negative cases from Appendix C below. Will be published under
+  [`Flux-Point-Studios/cip30-deeplink-conformance`](https://github.com/Flux-Point-Studios/cip30-deeplink-conformance)
+  (repository to be created alongside the first conformance pass); if editors prefer a
+  vendor-neutral home under `cardano-foundation/` we will migrate.
+
+The spec authors will surface real preprod transaction hashes (Aegis &harr; Yuti) and the conformance
+test pass status to the Discussions thread on this PR as those milestones land, so editors can track
+progress empirically rather than on promises.
+
+## Appendix A &mdash; Error codes
```
rphair:
```suggestion
### Appendix A &mdash; Error codes
```
(continued)

CIP-0186/README.md
```diff
@@ -0,0 +1,1366 @@
...
+| -10  | ResponseSignatureInvalid | dApp-side: Ed25519 verification of `signature` failed against `session.signingPublicKey` (for `connect`, against the `signingPublicKey` inside the just-decrypted payload), or a `connect` response is missing `method=connect`/`signature`/`echo` |
+| -11  | AuxiliaryDataHashMismatch | `auxiliary_data_hash` in `transaction_body` did not match the supplied `auxiliary_data`, OR a hash is present without data, OR data is present without a hash |
+| -12  | SessionEntropyReuse      | Wallet detected `session_entropy` reuse across `connect` events for the same `dapp_host` |
+| -13  | SourceAppUnverified      | Source app's bundle ID could not be verified against the `redirect` host's AASA / `assetlinks.json` |
+| -14  | UseInjectedAPI           | Request originated from a WebView hosted by the wallet; dApp MUST use the injected CIP-30 surface instead |
+| -32  | PayloadTooLarge          | Request URL exceeded the wallet's accepted size               |
+| -100 | InternalError            | Wallet bug or unrecoverable state                             |
+
+## Appendix B &mdash; Wallet registry (v1 draft, advisory)
```
rphair:
```suggestion
### Appendix B &mdash; Wallet registry (v1 draft, advisory)
```
(continued)

CIP-0186/README.md
```diff
@@ -0,0 +1,1366 @@
...
+enforces lowercase ASCII `id`, per-wallet `scheme` matching `^cip30dl-[a-z0-9][a-z0-9_-]{0,30}[a-z0-9]$`,
+and HTTPS-only `https` value with the trailing slash. The schema file is shipped in the same PR
+that introduces the registry. New wallet entries are added by PR against this folder.
+
+Per-wallet schemes (`cip30dl-lace:`, `cip30dl-eternl:`, ...) prevent the "shared scheme hijack"
+class of attack where any installed app can register the bare `cip30dl:` scheme and intercept
+fallback URIs intended for a specific wallet. The HTTPS form uses each wallet's own domain so
+Universal Link claims do not collide.
+
+## Appendix C &mdash; Test vectors
```
rphair:
```suggestion
### Appendix C &mdash; Test vectors
```
(continued... ***and*** also demote the `C.` subsections below)

rphair (commented):
@realdecimalist that's great to hear & I've put it on the next CIP meeting's agenda specifically for _Review_ (https://hackmd.io/@cip-editors/138) to hopefully issue a broader confirmation, flag any action items for fixes or additional peer review, and keep moving toward `Last Check` if unobstructed.

We should be able to accommodate a detailed, interactive proposal review at the meeting this time and so I'll re-invite any wallet developers & ecosystem advocates already tagged: @fabianbormann @Scitz0 @AlexDochioiu @refi93 @aljosa @gufmar @francisluz @Crypto2099 @MarcelKlammer @zuzunker

@realdecimalist I don't know if this OP offer still stands:
> **Co-author opportunity**
> ... We explicitly invite the Lace, Eternl, Vespr, Typhon, and Nami teams to be listed as co-authors in v1.1 once they review the spec and confirm the patch surface is workable inside their codebases.

... but even though you've gotten this into a very refined state I would not pressure you to push forward merging the existing proposal if someone offered to take this on a slightly different track leading to a ready mainstream implementation... so please feel free to re-ask this question the meeting if still appropriate.

@Ryun1 you are our most likely CIP editor to certify this so I also hope you can assess the current state in the next week &½ sometime (_after_ the format cleanup in this review; to avoid duplication of effort) and perhaps give us a shortlist of topics we should try to address in a live review at the meeting.

Note I am removing the phrase "— draft for editorial review" from the OP title since now we are expecting _everybody_ to review it. 😉

marcuspuchalla:
Hi @realdecimalist, @rphair, @Ryun1 — **Eternl** here.

We've built a working reference implementation of CIP-0186 inside Eternl and round-tripped the full method surface on real Android and iOS devices, including the signed `connect` handshake. We tested not just the happy path but the security-critical rejection cases (tampered commit, replay, redirect hijack, forged response signature), and our implementation independently converged on the same resolutions you landed on for the nine ambiguities, so the wire format now has a third implementation agreeing with Yuti and Gero.

The headline: **the protocol holds up well in practice**, implementable end-to-end with no relay, WebView, QR, or socket. We plan to publish our reference implementation (a demo dApp plus the supporting libraries) once we have the bandwidth, and the wallet-side support will ship in one of the next Eternl releases.

### Answers to the five open questions

1. **`submitTx` MUST vs MAY → keep MAY (optional).** Agreed. We implemented it, but dApps with their own backend don't need it, and the relay attack surface you flagged is real. Optional is the right default.

2. **CIP-95 governance → yes, include it.** We'd support specifying it in v1. The transport carries it with no changes , extensions ride the existing envelope and Conway certs already sit in the `signTx` CBOR, so it's just a matter of spelling out the governance read-methods and DRep `signData` envelope.

3. **12 KB ceiling → fine as-is.** Our `signTx` URLs ran ~2–4 KB, never near the ceiling, so 12 KB already has plenty of headroom in practice. No need to raise it.

4. **Wallet registry → decentralised (origin-anchored).** We'd favour each wallet publishing its association and attestation documents under a well-known path on its own domain, resolved at runtime, over a central `wallet-registry.json`. We've implemented this and it works cleanly, keeping every wallet authoritative over its own entry. Suggest making the well-known path normative and any central registry advisory-only.

5. **iOS custom-scheme fallback → keep.** Agreed, as defence-in-depth. In our testing the custom scheme routed deterministically on both platforms, and the redirect-host binding plus the signed connect bound the hijack risk you noted. Worth stating plainly in the spec that it's the fallback and Universal/App Links are preferred.

### Patch surface

For the record on @realdecimalist's ~290–400-line estimate: we can **confirm the patch surface is workable inside Eternl's own codebase**. CIP-0186 dropped into our existing signing engine as a self-contained transport layer (URL handler → decrypt → existing CIP-30 method dispatch) without disturbing the engine itself, in line with that estimate. So the "small patch around an existing signing engine" claim holds up in a second, independent production wallet.

Thanks for driving this, it's a good additional piece for mobile-native Cardano UX.

rphair:
@realdecimalist we still have this scheduled for a status review at the next CIP meeting (https://hackmd.io/@cip-editors/138) but we can postpone it to the following meeting if you still haven't responded to the last editorial review (https://github.com/cardano-foundation/CIPs/pull/1189#pullrequestreview-4574458890) or would like more time to respond to @marcuspuchalla's Eternl feedback from https://github.com/cardano-foundation/CIPs/pull/1189#issuecomment-4844269977.

rphair:
@realdecimalist after having made some space at the CIP meeting today for any active discussion, we agreed there that the last reviews posted & emphasised in the last couple of comments — some format fixes, but much more valuably the Eternl feedback in https://github.com/cardano-foundation/CIPs/pull/1189#issuecomment-4844269977 — are detailed enough to make plenty of progress on GitHub.

When you can respond to those items we'll observe this immediately, and may put this back on the Review meeting agenda when this is resolved and/or more discussion questions come up.  Please also feel free to use the [Discord channel](https://discord.com/channels/971785110770831360/1176612265894613133) in lieu of a more formal working group whenever it might suit you.

Looking forward to further online review including continued guidance from co-editor @Ryun1 who has been a significant contributor himself to Cardano's wallet API.

realdecimalist:
> Hi @realdecimalist, @rphair, @Ryun1 — **Eternl** here.
> 
> We've built a working reference implementation of CIP-0186 inside Eternl and round-tripped the full method surface on real Android and iOS devices, including the signed `connect` handshake. We tested not just the happy path but the security-critical rejection cases (tampered commit, replay, redirect hijack, forged response signature), and our implementation independently converged on the same resolutions you landed on for the nine ambiguities, so the wire format now has a third implementation agreeing with Yuti and Gero.
> 
> The headline: **the protocol holds up well in practice**, implementable end-to-end with no relay, WebView, QR, or socket. We plan to publish our reference implementation (a demo dApp plus the supporting libraries) once we have the bandwidth, and the wallet-side support will ship in one of the next Eternl releases.
> 
> ### Answers to the five open questions
> 1. **`submitTx` MUST vs MAY → keep MAY (optional).** Agreed. We implemented it, but dApps with their own backend don't need it, and the relay attack surface you flagged is real. Optional is the right default.
> 2. **CIP-95 governance → yes, include it.** We'd support specifying it in v1. The transport carries it with no changes , extensions ride the existing envelope and Conway certs already sit in the `signTx` CBOR, so it's just a matter of spelling out the governance read-methods and DRep `signData` envelope.
> 3. **12 KB ceiling → fine as-is.** Our `signTx` URLs ran ~2–4 KB, never near the ceiling, so 12 KB already has plenty of headroom in practice. No need to raise it.
> 4. **Wallet registry → decentralised (origin-anchored).** We'd favour each wallet publishing its association and attestation documents under a well-known path on its own domain, resolved at runtime, over a central `wallet-registry.json`. We've implemented this and it works cleanly, keeping every wallet authoritative over its own entry. Suggest making the well-known path normative and any central registry advisory-only.
> 5. **iOS custom-scheme fallback → keep.** Agreed, as defence-in-depth. In our testing the custom scheme routed deterministically on both platforms, and the redirect-host binding plus the signed connect bound the hijack risk you noted. Worth stating plainly in the spec that it's the fallback and Universal/App Links are preferred.
> 
> ### Patch surface
> For the record on @realdecimalist's ~290–400-line estimate: we can **confirm the patch surface is workable inside Eternl's own codebase**. CIP-0186 dropped into our existing signing engine as a self-contained transport layer (URL handler → decrypt → existing CIP-30 method dispatch) without disturbing the engine itself, in line with that estimate. So the "small patch around an existing signing engine" claim holds up in a second, independent production wallet.
> 
> Thanks for driving this, it's a good additional piece for mobile-native Cardano UX.

@marcuspuchalla thank you! please do link the reference implementation when it's published as it slots directly into the Path to Active evidence.

adopting your answers on the five open questions:

1. submitTx: stays OPTIONAL (MAY). No spec change.

2. CIP-95: agreed, v1. I'll promote the §CIP-95 governance profile from RESERVED to normative, spelling out the governance read-methods and the DRep signData envelope. (@Ryun1 this also answers your earlier review question: normative in v1 it is.)

3. 12 KB ceiling: stands, given your 2–4 KB real-world sizes.

4. Registry: agreed. I'll make the origin-anchored well-known path normative and demote the central wallet-registry.json to advisory-only, per your running implementation.

5. Custom-scheme fallback: retained as defence-in-depth, with explicit language that Universal/App Links are the preferred transport and the custom scheme is fallback only.

spec commits for 2, 4, and 5 to follow as separate commits (same one-commit-per-change pattern as the ambiguity fixes).


also, the co-author offer from my proposal still stands, and you're the first team to complete exactly what it asked here on the thread. happy to list Eternl as co-author in v1.1 if you're up for it. same invitation goes to the Gero team, whose implementation validated the wire format even earlier (will reach out to them directly).

realdecimalist:
@rphair all six format suggestions accepted and committed. title trimmed, authors line unquoted, Reference Implementation moved under a top-level ## Appendices, and Appendix A/B/C demoted accordingly (C's subsections demoted in a follow-up commit, per your note).

on the co-author offer: yes, it still stands @marcuspuchalla's Eternl team is the first to complete it, and I'm extending the same invitation to Gero, who implemented even earlier. details in my reply above.

the Eternl feedback is now resolved on all five open questions; spec commits for CIP-95 v1 normativity, the registry normativity flip, and explicit custom-scheme-fallback language to follow as separate commits.

happy for this to go back on the review agenda whenever it fits.

rphair (commented):
@realdecimalist thanks for moving ahead so definitively with both the mundane and technical review responses.  It seems all pending issues are settled at this point except the 2 possible co-authors: editors will keep an eye out that anyone claiming the credit you offer will have their preferred name/email included before merging this.

I've also officially postponed your promised meeting review from our last meeting ([#138](https://hackmd.io/@cip-editors/138)) to our upcoming meeting: https://hackmd.io/@cip-editors/139 ... so we can cover any remaining review goals & any more devs/teams/products to include in consensus before merge & also confirming editor @Ryun1 is happy with it so far.

If no other significant call for revision or waiting for further cooperation, the goal at next week's meeting would be to put this at `Last Check` for the meeting after that.  This would put this on track to merge in about 3 weeks so please start thinking about anyone else we really need to (and are able to) get feedback from before merging. 🙏

CIP-0186/README.md
rphair: Resolved with https://github.com/cardano-foundation/CIPs/pull/1189/changes/01d1485f18cd5c12f52c82a89046656e8111a2ef.

CIP-0186/README.md
rphair: Resolved with https://github.com/cardano-foundation/CIPs/pull/1189/changes/7c99f3b0c0d6b1c038519a5f3059ddca56c825d6 although @Ryun1 the validator is still not running here — even though there's been some commits — but I've noted that in the [container issue](https://github.com/cardano-foundation/CIPs/issues/1216) instead of continuing these observations here.

CIP-0186/README.md
rphair: Resolved with https://github.com/cardano-foundation/CIPs/pull/1189/changes/107c9d4bda6bf4293b1c5e7783fcd70ce7df356e.

CIP-0186/README.md
rphair: Resolved with https://github.com/cardano-foundation/CIPs/pull/1189/changes/627ec41d2f46fe15f52b511cd8f782695a6aba26.

CIP-0186/README.md
rphair: Resolved with https://github.com/cardano-foundation/CIPs/pull/1189/changes/627ec41d2f46fe15f52b511cd8f782695a6aba26.

CIP-0186/README.md
rphair: Resolved with https://github.com/cardano-foundation/CIPs/pull/1189/changes/627ec41d2f46fe15f52b511cd8f782695a6aba26 & https://github.com/cardano-foundation/CIPs/pull/1189/changes/197b5b6f940768ed272e57977bbb6ef08cd9be4a,

rphair (commented):
@realdecimalist thanks for attending the CIP meeting today to provide a general status of this effort; as expected there's been broad support from wallet devs (3 projects so far have indicated progress) and without any sign of trouble we're free to mark this `Last Check` for the next CIP meeting: https://hackmd.io/@cip-editors/140

perturbing (approved):
I am not a wallet expert, but given that @marcuspuchalla closed the open gaps and that those are in the CIP, I approve.

Awesome to see this implemented by more wallets than Eternl, and congratulations on a well-designed, impactful design for the broader ecosystem (native mobile support is a big gap, imo).

Ryun1 (approved):
this is a very cool standard
im looking forward to trying it out

thank you for the contribution @realdecimalist

rphair (approved):
@realdecimalist generally we would wait until the CIP meeting to merge a proposal as profound and beneficial as this, so I'm not hitting the Merge button yet despite the unanimous approval.  Thanks for setting such a great example for the standards community and all the development efforts that come into contact with it.

PR #1187, opened 2026-05-12
CIP-???? | Address handles in hierarchical namespace

kingslavcho:
This PR proposes a new CIP defining the .id handle namespace on Cardano.

GetMyID (https://getmyid.today) is a live Cardano mainnet platform where 
users mint personal identity handles like $john.smith.id as NFTs under a 
single Policy ID. The resolver API is already public and operational at 
https://getmyid.today/api/resolve/

Both testnet (preprod) and mainnet versions are running simultaneously,
making GetMyID the only Cardano naming solution with a live testnet 
environment for developer testing.

Requesting community review and CIP number assignment.

PR #1192, opened 2026-05-15
CPS-????: Cardano Handle Namespace Governance Framework

kingslavcho:
This CPS documents the absence of an open governance framework for 
human-readable handle namespaces on Cardano and proposes requirements 
for a solution.

This is submitted as a follow-up to the discussion in CIP PR #1187 
where the CIP editors and community identified the need for a broader 
problem statement before a technical solution can be proposed.

Key problems documented:
- No open standard exists for Cardano handle namespaces
- Well-funded incumbents can claim any namespace at any time
- No protection for users of non-incumbent platforms
- The CIP process provides no enforcement mechanism

The CPS also documents GetMyID's reference implementation of the .did 
namespace as Cardano's implementation of the W3C DID standard.

Related PR: https://github.com/cardano-foundation/CIPs/pull/1187

---

([rendered document](https://github.com/kingslavcho/CIPs/blob/master/CPS-0%3F%3F%3F%3F/README.md))

### DIFF
--- a/CPS-0????/README.md
+++ b/CPS-0????/README.md
@@ -1,0 +1,84 @@
+CPS: ???? Title: Cardano Handle Namespace Governance Framework Status: Open Category: Tokens Authors:
+
+    Slavcho King Andreevski support@getmyid.today Created: 2026-05-15 License: CC-BY-4.0
+
+Abstract
+
+This CPS (Cardano Problem Statement) documents the absence of an open, governed framework for human-readable handle namespaces on Cardano. It identifies the problems created by the current unregulated situation and proposes the requirements that a solution must satisfy — including the establishment of the .did (Decentralized Identifier) namespace under the GetMyID reference implementation as a starting point for community governance.
+Problem
+
+    No open standard exists for Cardano handle namespaces
+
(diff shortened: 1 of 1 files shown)


rphair:
@kingslavcho I have to close this because, even after guidance in https://github.com/cardano-foundation/CIPs/pull/1187#issuecomment-4455993695, it has again been submitted from your `master` branch.  For proper collaborative review we need you to use a different branch than the one your CIP will/should be from; so please follow the documentation in this section: https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/proposing-changes-to-your-work-with-pull-requests

Also this isn't formatted in Markdown as a CPS (or CIP) would be: likely the result of improper format conversion (or maybe only a copy/paste) from a web page where it was rendered.  (I'm putting a "rendered proposal" link in your OP above just so you can see how the actual content appears: i.e. without proper Markdown headings, links & other formatting.)

When ready to resubmit your PR, please also you will need to title it like a CPS ("**problem statement**") would be: synonymous _with a **problem**_ rather than a solution:
* I believe we agreed in #1187 that the problem was how to technically divide ownership of hierarchical handle namespace, so you could choose a title like this:
* `Coordination of Hierarchical Handle Namespace`

PR #1194, opened 2026-05-18
CPS-0031? | Prioritising Urgent Transactions

willjgould:
This PR adds CPS-prioritising-urgent-transactions, which frames the problem of a lack of protocol-enforced urgency signalling on Cardano. It's been written in response to interviews with community members, hosted by @CarlosLopezDeLara at Buidler Fest 3, on the topic. The appetite exists for some mechanism whereby priority inclusion can be granted to transactions.

Examples of potential solutions could include things like tiered pricing or certain forms of dynamic pricing.

Any potential solutions must target linear-Leios.

[Rendered](https://github.com/willjgould/CIPs/blob/urgency-signalling/CPS-0031/README.md)


CPS-0031/README.md
```diff
@@ -0,0 +1,132 @@
+---
+CPS: 31
```
rphair, line 2:
```suggestion
CPS: "?"
```
Any metasyntactic `null` will do (`?` was common but GitHub has been giving YAML errors when it's not quoted).

CPS-0031/README.md
```diff
@@ -0,0 +1,132 @@
...
+Title: Urgency Signalling
```
rphair:
```suggestion
Title: Prioritising Urgent Transactions
```
`Urgency Signaling` would be a good title for one of the "candidate solutions" (i.e. a CIP).  This alternative is from your Abstract & we might also test it as a PR title to see if it covers all the possible solutions without bias toward any in particular.  (I believe the qualitative `urgent` would cover the quantitative "high urgency" as a CIP specification might define it.)

CPS-0031/README.md
```diff
@@ -0,0 +1,132 @@
...
+
+How can we retain fee quote validity across repricing intervals?
+
+What information is leaked when a transaction signals urgency?
+
+What MEV opportunities are created or amplified by public urgency signals?
+
+## Copyright
+
+This CPS is licensed under CC-BY-4.0.
```
rphair:
```suggestion
## Copyright

This CPS is licensed under [CC-BY-4.0](https://creativecommons.org/licenses/by/4.0/legalcode).
```
Linking to the full license text is recommended by both CC and Apache & is also done in all othe CIPs and CPSs (often from [our template](https://raw.githubusercontent.com/cardano-foundation/CIPs/refs/heads/master/.github/CPS-TEMPLATE.md)).

rphair (commented):
Thanks @willjgould @polinavino — marking `Triage` for next CIP meeting: https://hackmd.io/@cip-editors/135

Please rename the containing directory ASAP to something without a number, like `CPS-urgency-signalling`.  There is never any guarantee in advance a CIP will be given a particular number: in general it might not be given a number at all.
* The reasons why not to do this are [explained in CIP-0001](https://github.com/cardano-foundation/CIPs/tree/master/CIP-0001#1a-authors-open-a-pull-request) (also referenced in CIP-9999).
* Since this has already has some community discussion, please also avoid any other references to this as "CPS 31" until if & when the CIP meeting can give this a sequentially assigned CPS number.

Please also @willjgould — now that your content appears sorted — I would ask that this be the last of your forced pushes ([as we also ask here](https://github.com/cardano-foundation/CIPs/tree/master/CIP-0001#follow-a-reviewer--and-editor-friendly-review-process)): since authors don't always include prior commit history, which is often assembled by interactive review committed through GitHub.
* You might force further pushes if you take care to include all these commits: but generally it's easier to pull them back from the PR before pushing your latest changes out again (merging locally as necessary). 🙏

CPS-0031/README.md
CarlosLopezDeLara: Both can work for the title. IMHO the suggestion just changes the perspective _"Urgency Singalling"_ is from the user's perspective, while _"Prioritising Urgent Transactions"_ seems to be more from the system's point of view. Don't know if we have a convention here, but if we are stating problems (CPS), I believe it's better to frame it from the user's POV.    

I wont offer too much resistance, as I said, I believe both options work.

rphair: @CarlosLopezDeLara yes I see your point & have noted to bring up this question at the next CIP meeting, where we try to get these questions resolved without too much abstract discussion.  We've never specifically considered "user vs. system" point of view before for naming.

When in doubt we would revert to the author's original suggestion: so we'll include any further comments in this thread from reviewers & select something then based on consensus about which point of view works best.

CPS-0031/README.md
rphair: Resolving: fixed by https://github.com/cardano-foundation/CIPs/pull/1194/commits/4fa7f9f95aee19507901aa48d0e6173e785e3a7e.

CPS-0031/README.md
rphair: Resolving: fixed by https://github.com/cardano-foundation/CIPs/pull/1194/commits/4fa7f9f95aee19507901aa48d0e6173e785e3a7e.

CPS-0031/README.md
```diff
@@ -0,0 +1,132 @@
...
+
+From stakeholder interviews during Buidler Fest #3, hosted by Carlos Lopez De Lara:
+
+2. **Permissionless access.** Priority must be available to anyone willing to fulfil the necessary prerequisites, not negotiated through relationships or in private arrangements.
+
+3. **Predictability over raw speed.** The signal predictably improves access to timely inclusion, rather than only modestly improving odds. This includes a reduction in wait time variance for high-urgency transactions.
+
+### Constraints
+
+1. **Automatable semantics.** The urgency signal must be encodable in smart contract logic and readable by automated systems, not just manually configured.
```
fallen-icarus: Why *must* smart contracts see it? The solution needs to be networking level. I don't see why smart contracts *need* to see the signal.

CPS-0031/README.md
```diff
@@ -0,0 +1,132 @@
...
+
+During congestion, these transactions compete for block space on equal terms with traffic that has no particular time sensitivity. The protocol treats all valid transactions alike with respect to urgency: there is no way for a transaction to express that it is urgent, and no mechanism for block producers to commit to honouring such a signal. Urgent and non-urgent transactions queue together, and inclusion order is determined by factors opaque to the submitter.
+
+Users and protocols lose value to avoidable delay. Block producers leave value uncaptured because users cannot express how much timely inclusion is worth to them. Additionally, the absence of a legitimate priority channel creates pressure toward off-chain arrangements that undermine the permissionless properties of the network.
+
+### Explored Alternatives
```
Ryun1, line 33: Presumably, different node implementations could add this a differentiating feature?

edit; dependent on the proposed solution, I believe

CPS-0031/README.md
Ryun1: what about `Transaction Prioritisation`?
I feel that this captures the problem space well
whereas `Urgency Signaling` feels more like the solution for prioritisation

CPS-0031/README.md
fallen-icarus: I'm not sure this is true. If not enough nodes support it, then users can't plan around it. It needs to be **predictably** reliable.

CPS-0031/README.md
polinavino: Yeah, good point. The constraint should have been two entirely separate ones :
- the urgency signal (and any related changes to on-chain transaction processing) should not compromise ledger determinism (to prevent attacks making use of Phase 2 validation)
- the costs related to urgency signalling should be documented on-chain in ledger logic (this is to ensure everyone has equal access to current prices for urgency signalling, and that the price updates are computed correctly)

polinavino:
Things to also potentially discuss moving into goals :
- No information beyond information XYZ (tbd) is leaked when a transaction signals urgency
- ensuring that low-urgency users are not starved of block space (alternatively, this should be in non-goals)

CPS-0031/README.md
rphair: @Ryun1 @CarlosLopezDeLara @willjgould as somewhat expected, the CIP meeting today was in agreement that `Urgency Signaling` sounded more like the/a CIP & `Prioritising ...` like the CPS.

The suggestion immediately above (`Transaction Prioritisation`) also seems to accommodate @polinavino's important point in https://github.com/cardano-foundation/CIPs/pull/1194#issuecomment-4547629357 that a priority implementation might also affect _non-_ urgent transactions.

In any case I think the editors can now follow the lead from the authors & subject matter experts: if either the PR or the CIP title is updated further, we'll just make sure the other follows.

CPS-prioritising-urgent-transactions/README.md
```diff
@@ -0,0 +1,132 @@
+---
+CPS: "?"
```
rphair:
```suggestion
CPS: 31
```

rphair (commented):
@willjgould this was confirmed at the CIP meeting today... please assign the number `31` which now I feel bad about having you retract: although in fact there was one CPS which came in ahead of this one & so that number was not assured. 😅  In any case please rename the containing directory to `CPS-0031` and update the "Rendered" link in your OP. 🎉

I also had a pending question about whether the Tx focus could characterise this as a Ledger category proposal (rather than `Consensus`) which @perturbing answered neatly: "Ledger is about what the blocks look like: Consensus is about how & when the blocks are made."

Editors will do their best in monitoring the peer review that's already begun over the technical detail & scope... so please authors, advocates and/or critics post here when you think review has concluded so we can schedule either a more intensive meeting review or make a path towards merging this.

polinavino:
@rphair , glad to hear this was confirmed! 
@willjgould will be back next week
also, this proposal will definitely affect consensus (which included the mempool I believe), and very likely also the ledger (to ensure price transparency)

CPS-0031/README.md
willjgould: Very good points both! Updated.

rphair: Resolved with https://github.com/cardano-foundation/CIPs/pull/1194/commits/95e8b0e60b8674bef189069f408a2c0d91602c8d.

CPS-0031/README.md
rphair: Resolving since review & update is proceeding well with current PR title `Prioritising Urgent Transactions`.

colll78:
The "urgency signal" should be setting the transaction fee in the tx body higher than the minimum requirement. With respect to "updating a signal", this should be a non-goal. It would be better to simply rebuild the same tx with a higher tx fee and submit that.

We should reframe around the problem being that currently overpaying tx-fees has no impact on transaction prioritization. The question then becomes how we get the network to respect some form of tx-fee ordered transaction processing, and what that ordering mechanism looks like.

willjgould:
> The "urgency signal" should be setting the transaction fee in the tx body higher than the minimum requirement. With respect to "updating a signal", this should be a non-goal. It would be better to simply rebuild the same tx with a higher tx fee and submit that.

I wanted to avoid loading the CPS with any solution-shaped statements, but this is indeed along the lines of a solution we're working on. Regarding the signal update, this is something that's largely avoided by construction in the solution. But in those situations where an update is required, rebuilding the tx is necessary.

> We should reframe around the problem being that currently overpaying tx-fees has no impact on transaction prioritization.

This is where I'd start to disagree. There could be a connection here, but it's not necessary to make one in the CPS. The problem of currently overpaying fees is a separate problem, and it doesn't need to be resolved by urgency signalling. [The fee-change CIP](https://github.com/cardano-foundation/CIPs/pull/1218) can be used to address this independently of prioritisation.

> The question then becomes how we get the network to respect some form of tx-fee ordered transaction processing, and what that ordering mechanism looks like.

As you say, this is one answer to what is essentially the core of the problem, but I'd avoid framing it in terms of fees in a CPS since that's solution-shaped.

colll78:
>This is where I'd start to disagree. There could be a connection here, but it's not necessary to make one in the CPS. The problem of currently overpaying fees is a separate problem, and it doesn't need to be resolved by urgency signalling. https://github.com/cardano-foundation/CIPs/pull/1218 can be used to address this independently of prioritisation.

The problem is not overpaying fees; the problem is that paying more than base fees buys you nothing. Transaction prioritization and fee escalation are a tightly coupled problem. Any prioritization mechanism must have some sort of economic cost, in the form of increased fees. 

[CIP-0192? | Transaction Fee-Change Field](https://github.com/cardano-foundation/CIPs/pull/1218)

Does not solve or address this problem in any way. The problem is concretely that right now, a transaction with higher than required fees is not prioritized (by vanilla nodes) over a transaction with minimum required fees. Fee change is strictly unrelated, and in my opinion, a very bad feature to introduce.

willjgould:
> The problem is not overpaying fees; the problem is that paying more than base fees buys you nothing. 

I understand the point you're making, but the way this is framed already assumes a solution. The problem is that there is no protocol-recognised mechanism to signal transaction urgency, which is what this CPS is for.

> Transaction prioritization and fee escalation are a tightly coupled problem.

They can be coupled by a particular solution, but whether prioritisation should be purchased through higher fees is a design question for a CIP addressing this CPS, not an assumption the CPS should make.

> Any prioritization mechanism must have some sort of economic cost, in the form of increased fees.

This doesn't follow. An economic cost doesn't necessarily need to be in the form of increased fees.

> Does not solve or address this problem in any way. The problem is concretely that right now, a transaction with higher than required fees is not prioritized (by vanilla nodes) over a transaction with minimum required fees. Fee change is strictly unrelated, and in my opinion, a very bad feature to introduce.

Fee change is desirable in a system which posts a specific price for entry to a specific tier. As @polinavino mentioned in that CIP's comments, there's an (additional) upcoming CIP that depends upon fee change.

drospa:
Dear CIP editors @Ryun1 @rphair, in behalf of the project team working on this proposal: what it is missing to be able to publish the CPS? All comments have been addressed

CPS-0031/README.md
rphair: resolving: question appears to be addressed.

rphair (approved):
@drospa we were only tentatively waiting for a conclusion of the latest discussion between @willjgould & @colll78... though this most often happens with a comment like yours, because people usually don't comment to accept the other person's point of view & explicitly confirm a consensus.

In any case @colll78 please post again if that issue was not resolved to your satisfaction with https://github.com/cardano-foundation/CIPs/pull/1194#issuecomment-5394065401.  Also:
* @polinavino please post with any issues we might be waiting for if you don't think this is ready for merge.
* @fallen-icarus please also take one more look & post any reservations you may have, especially in the economic scenarios in _Use Cases_.
* @CarlosLopezDeLara you are still welcome to post any points to include in the CPS that came from your event described in the OP.

With no amendment or dispute that would hold it back, this will be up for `Last Check` for merge at next week's CIP meeting: https://hackmd.io/@cip-editors/143

Personally I think the content forms an excellent preface to the work expected in support of Cardano's transaction flexibility required for new & expanding markets.

fallen-icarus:
This CPS seems biased towards latency as the configurable dial when mempool propagation is another option. There are two types of urgency: temporal and causal.

- **Temporal:** My loan is about to expire and I need my tx processed before that expiration slot passes.
- **Causal:** I'm racing to arbitrage a trade before another arbitrager gets to it.

The latter group is more economically important since it currently has no workaround. Meanwhile temporal urgency can be managed by just planning ahead. I'd appreciate it if this CPS could be more unbiased about the two types.

Especially with Peras, I don't think latency is the main concern; it is mempool propagation. If I want to execute a trade before someone else, I need the network to selectively propagate my tx faster than my competitors. With causal urgency, I don't need my tx processed within the next 2 minutes; I just need my tx to get processed before another tx which touches the same state.

---

~This is directly relevant to the upcoming Leios Urgency CIP which is choosing to allow txs in RBs for the urgency lane even though that requires sacrificing Linear Leios throughput. By focusing on mempool propagation instead, we can support urgency without hurting Cardano's throughput.~

TIL that Linear Leios is only expected to produce ~50% of RBs that contain EBs due to the slot delay required for EBs. So the other ~50% of RBs could be used for the urgency lane without impacting throughput. Though focusing on mempool propagation should still be a valid alternative implementation mentioned by this CPS.

willjgould:
There are two kinds of urgency, and this CPS currently targets the temporal kind.

We could, if it's desirable, edit this CPS to also include the causal kind. It seems implied to be allowed, [according to CIP-9999](https://cips.cardano.org/cip/CIP-9999#statuses), for a single CPS to cover multiple problems, which can be addressed by multiple CIPs.

@rphair @fallen-icarus would this be an acceptable solution?

fallen-icarus:
> We could, if it's desirable, edit this CPS to also include the causal kind. ... would this be an acceptable solution?

Yes. Mentioning causal urgency too would be enough for me.

rphair:
@willjgould yes of course that would be desirable: putting in the parallel but intimately related kind of problem here would minimise bureaucracy, not just for related CIP authorship & editing but perhaps also for implementations.  Specifically we try to avoid the condition of "one CPS per CIP" since it's always been the intention that single problem definitions can have more than one possible solution.  (Maybe some currently unanticipated solutions will be both causal _and_ temporal and so a single problem definition would be even more useful in that case.)

willjgould:
The CPS now describes both aspects of the problem, and CIP-0183 references (what will be) the link to the CPS as a `Solution To`.

rphair:
thanks @willjgould re: https://github.com/cardano-foundation/CIPs/pull/1194/changes/16751d281cc4857141747814d42240b7586f0e10 — yes, we generally suggest that, to save an extra PR afterward 🙏

CPS-0031/README.md
```diff
@@ -0,0 +1,164 @@
...
+License: CC-BY-4.0
+---
+
+## Abstract
+
+During periods of congestion, high-urgency transactions lose value when they cannot obtain timely inclusion. A protocol-recognised urgency signal could help preserve more transaction value during congestion, especially for transactions whose value is highly delay-sensitive. Urgency may also be causal rather than temporal: a transaction may need inclusion before a conflicting transaction consumes a UTxO it depends on, which can happen even when block space is plentiful.
+
+Candidate solutions should be evaluated by how they handle prioritising high-urgency transactions, and by how they affect ordinary and low-urgency users during sustained congestion. Candidate solutions may address temporal urgency, causal urgency, or both.
+
+
```
perturbing:
```suggestion
```

CPS-0031/README.md
```diff
@@ -0,0 +1,164 @@
...
+* Multi-relay submission: Where the node is connected to multiple SPO relays to increase the likelihood that the transaction reaches the next block producer quickly
+
+Deployed as standard infrastructure. Improves latency-to-mempool, not confirmation ordering. Once in the queue, the transaction competes equally with everything else.
+
+* Private SPO arrangements: 
+
+Explored and rejected. Even agreements with major SPOs yield next-block probability insufficient for liquidations. More importantly, this produces a worse outcome than a formal mechanism: an opaque, permissioned, off-chain priority market accessible only to well-capitalised incumbents.
+
+
```
perturbing:
```suggestion
```

CPS-0031/README.md
```diff
@@ -0,0 +1,164 @@
...
+7. **Registration of unique names or handles**
+
+   **Scenario:** Users compete to register the same available name in an application that permits direct registration and enforces uniqueness on-chain.
+
+   **Example:** Two users submit conflicting claims for the same handle. Once one registration succeeds, the other cannot obtain that handle. The urgency comes from competing for an exclusive opportunity.
+
+   **Who loses today:** Users who cannot express the value of precedence through a public, permissionless mechanism and whose access instead depends on submission speed or application-level intermediaries.
+
+
```
perturbing:
```suggestion
```

CPS-0031/README.md
```diff
@@ -0,0 +1,164 @@
...
+Guaranteed precedence over every competing transaction
+
+Guaranteed retention of value for urgent transactions
+
+Elimination of congestion
+
+Any specific pricing mechanism
+
+
```
perturbing:
```suggestion
```

perturbing (approved):
Nitpicked the double extra empty lines 👍

perturbing (approved):
On the CIP meeting @colll78 confimed it okay, but noted that a tier one security audit might still be good. But IMO that is not a blocker for merging

rphair:
@perturbing @colll78 as I recall that comment was made about CIP-0113 / #444 — I'm not sure if we'd get a response from the first-tier auditors if we only have a _plan_ to develop something 😅

PR #1196, opened 2026-05-19
CIP-0164 | Replace wFA^LS committee with stake-based committee selection

ch1bo:
This PR proposes replacing the **wFA^LS (weighted Fait Accompli with Local Sortition)** voting-committee scheme currently specified in CIP-164 with a simpler **stake-based committee** scheme. The committee for an epoch is fixed at the epoch boundary by ordering pools by stake (descending) and selecting them, in order, until a configured cumulative-stake target is met. There are no per-EB elections, no non-persistent voters, and no per-vote sortition proofs. Every committee member is eligible to vote on every EB within the epoch.

The motivation is **certificate size and verification cost on the critical path**. Under wFA^LS, the bulk of certificate size and verification cost comes from the non-persistent voters' eligibility proofs. Removing local sortition eliminates that overhead. The certificate collapses to a bitfield over a known committee plus a single aggregated BLS signature. The following comparison uses the [`leios-wfa-ls-demo`](https://github.com/cardano-scaling/leios-wfa-ls-demo) benchmarks against mainnet stake data, with the wFA^LS configuration of 481 PVs + 94 NPVs as the reference point:

|                                       | wFA^LS (current spec)   | Stake-based, equivalent coverage (this PR) |
| ------------------------------------- | ----------------------- | ------------------------------------------ |
| Committee size                        | 481 PV + 94 NPV         | 916 voters (top-stake at 99% cumulative, mainnet epoch 612) |
| Certificate size                      | ~6.8 kB                 | ~203 bytes (bitfield ⌈916/8⌉ + agg. sig. + overhead) |
| Certificate verification (worst case) | ~10 ms                  | ~2 ms                                      |
| Vote size                             | 94 B (PV) / 171 B (NPV) | 94 B (uniform)                             |
| Vote verification                     | 427 μs / 1750 μs        | 427 μs (uniform)                           |

**On equivalent coverage.** Mainnet has roughly 3,000 SPOs with a Gini coefficient of ~0.84. Stake distribution data across 51 recent epochs (562–612, summarized in [Paul Clark's stake-distribution analysis](https://docs.google.com/document/d/1CmSC8BRDZTH32fk_bZ4db6baYGodZMB-71b-S18B2PQ)) shows highly stable coverage thresholds:

| Cumulative stake | Pools required (epoch 612) | Range across epochs 562–612 |
| ---------------- | -------------------------- | --------------------------- |
| 90%              | 471                        | 471–503                     |
| 95%              | 601                        | 601–636                     |
| 99%              | 916                        | 911–956                     |

The trend is gently *decreasing* (consolidation toward larger pools), which tightens the argument over time rather than weakening it. wFA^LS's configuration of 481 PVs + 94 NPVs gives expected coverage near 100% per round, but only as a *sample* of the residual — NPVs are stake-weighted Poisson draws, not a census. **Stake-based truncation at 99% cumulative coverage (916 voters at epoch 612) therefore matches wFA^LS's effective security regime closely**: both schemes capture the dominant top-471 (the 90% cut, essentially the same set as the 481 PVs), and the 99% cut additionally captures most of the same tail that wFA^LS's NPV draws sample from. The remaining ~1% of stake — fragmented across ~2,000 very small pools with a long Pareto tail — is too dispersed to threaten the 75% quorum under any of the schemes. Doubling the committee from ~480 (90% cut) to ~916 (99% cut) increases certificate size only by the bitfield delta of ~55 bytes; the cert-size argument is essentially invariant across the 90%–99% coverage range. The size estimate above (~203 B) is back-of-the-envelope and will be replaced with a measured value from the wfa-ls-demo before merge.

Certificates are perpetual on-chain overhead; certificate verification is on the critical path between receiving a certifying RB and being ready to extend the chain; votes diffuse under a tight $L_{\text{vote}}$ budget. The reductions compound: at coverage parity with wFA^LS, the cert-size reduction is over an order of magnitude (~6.8 kB → ~203 B, ~33×) and certificate verification time drops ~5×.

### Scope

The change is a refinement of the certificate format and committee-selection rules, not a redesign. CIP-164 Appendix A already states that *"alternative schemes satisfying these requirements could be substituted"*; this PR exercises that explicit extension point.

**Unchanged:**

- BLS over BLS12-381
- The 75% quorum value
- Certificate inclusion timing constraints
- The LeiosNotify / LeiosFetch mini-protocols
- The threat model

**Changed:**

- **Committee structure** — dual PV/NPV → single stake-truncated set
- **Vote structure** — two shapes → one uniform shape
- **Certificate format** — per-NPV eligibility proofs → bitfield + single aggregated signature
- **Epoch-boundary computation** — Fait Accompli precomputation → stake-ordered truncation
- **Quorum definition** — head-count fraction of committee → minimum fraction of total active stake
- **Appendix A requirements** — updated to reflect public deterministic committee, key rotation as a should-have, and loosened liveness bound

**Explicitly out of scope:**

- **BLS key rotation mechanism.** The adaptive-security claim applies equally to *all three* schemes considered (All-vote, stake-based truncation, wFA^LS) only under the assumption of regular rotation; without rotation, the static-vs-adaptive distinction collapses regardless of committee scheme. The on-chain mechanism (pool registration certificate path, epoch-boundary activation, prevention of within-epoch grinding) is a meaningful spec change in its own right and is left to a follow-up PR amending this CIP.
- **VRF key rotation.** Related, but Praos-layer and outside CIP-164.
- **A fixed committee size.** The parameter is a cumulative-stake target (or maximum-error equivalent); realized committee size tracks the stake distribution.

### Evidence

1. **[`cardano-scaling/leios-wfa-ls-demo`](https://github.com/cardano-scaling/leios-wfa-ls-demo)** — Haskell implementation of wFA^LS on mainnet stake data, with the vote/cert size and verification benchmarks cited above.
2. **[ARC voting crypto review](https://github.com/input-output-hk/ouroboros-leios/blob/41cefc99a/docs/arc-voting-crypto-review.pdf)** — cryptographic analysis comparing All-vote, Truncation (the scheme in this PR), and wFA^LS on adaptive-security and overhead grounds. With BLS key rotation, all three schemes provide comparable adaptive security in the practically-relevant regime; the one feature wFA^LS uniquely retains is non-targetability of non-persistent voters, but the persistent voters under wFA^LS are precisely the largest pools an adaptive attacker would already target. Also provides formal pen-and-paper algorithms for the voting and certificate scheme.
3. **[Network simulations](https://github.com/input-output-hk/ouroboros-leios/tree/main/analysis/sims/2026w18)** — results on `mini-mainnet` (750 nodes) and a derived `midi-mainnet` (1,500 nodes) show throughput, EB certification rate, and below-threshold rates essentially identical across the three schemes at all tested loads (0.150–0.350 TxMB/s), and stake-based truncation within the $L_{\text{vote}}$ budget. To stress committee size directly on the same topologies, an "everyone votes" mode was also simulated — literally every node including relays, yielding committee sizes of 750 and 1,500 respectively — and the protocol still performed as expected. The signal so far is that committee size does not dominate performance at these scales, which significantly reduces (but does not fully eliminate) the concern that vote diffusion could be a limiting factor. **Caveat:** these runs are on topologies smaller than full mainnet (~3,000 SPOs); the `midi-mainnet` is similarly connected to mainnet but not the same scale, so a full-scale mainnet-equivalent run remains outstanding. On current evidence, the wFA^IID variant is retained in the CIP's [Design Decisions](CIP-0164/README.md#voting-committee-selection) as a forward-looking option rather than a fallback the data calls for; rationale and remaining caveats are recorded there.

### To be discussed

1. **Naming.** Is "stake-based committee" the right name? Internal discussion has used "Truncate", "top-stake-fraction", and "stake-based cutoff" interchangeably. Proposal: **stake-based committee** as the formal spec name, with **stake truncation** as an acceptable synonym informally.
2. **Parameter shape.** *Cumulative stake covered* (e.g. 99%) vs *maximum truncation error* (e.g. ≤1%). The two are equivalent. Slight preference for the error formulation because the same parameter shape generalizes across all three schemes — useful for the staged-fallback story in *Alternatives & Extensions* — but cumulative-stake is more intuitive.
3. **Quorum as a stake threshold.** The current spec defines $\tau$ as a "minimum fraction of committee votes required for certification" — a head-count fraction over committee members. This PR redefines $\tau$ as a **minimum fraction of total active stake** whose votes must be present in the certificate. $\tau = 0.75$ means exactly "votes from pools holding ≥75% of total active stake have signed." The security argument (certified EBs are known to >25% of honest stake even under 50% adversarial stake) is then stated directly in terms of stake, not voter count. Under head-count quorum, a coalition of many small committee members could satisfy the threshold while representing far less stake than a smaller set of large pools — the quorum would not measure what the security property actually requires. With a stake threshold, the certificate directly proves the predicate the diffusion-period safety argument relies on. This also gives a natural constraint between the two parameters: $\sigma_c > \tau$, i.e. the committee must cover more stake than the quorum demands, or certification is impossible. At $\sigma_c = 0.99$ and $\tau = 0.75$ there is a 24-percentage-point margin between what the committee covers and what the quorum requires.


perturbing (approved):
As I was involved in this work, I am familiar with the changes and approve already 👍

CarlosLopezDeLara:
@rphair what's the process to get this merged?

CIP-0164/README.md
rphair, line 20:
```suggestion
https://github.com/cardano-foundation/CIPs/pull/1196
```
The change justification is very well explained in the CIP, but the discussion behind this is also important & at least would serve as a historical note + helpful to confirm peer review.

(GitHub does not allow specifying an exact change here: we just need 1 more line \[`this` PR\] at the end of this section)

rphair (commented):
@CarlosLopezDeLara (cc @ch1bo) since we already have editor @perturbing's endorsement, confirming where this change is coming from — _and_ since the change itself is abundantly justified both in the OP & the new material — I would feel comfortable merging this after all the factors marked above **To be discussed** have been settled & documented here on GitHub. So we will:
* mark this `Triage` to ensure these points are raised for peer review _no later than_ the next CIP meeting (https://hackmd.io/@cip-editors/135) — though discussion will hopefully proceed here on GitHub immediately;
* once participants confirm settlement of those questions, with no other objections, I'd green-tick it myself and move it on to `Last Check` (@Ryun1 can also do this).
* (FYI these tags are [explained here](https://github.com/cardano-foundation/CIPs/wiki/301.-State-tagging))

Tagging for review (selected from Leios PR commentary & prior review, including this one in-progress:
* #1167

@lehins @WhatisRT @bwbush @will-break-it @dcoutts @nhenin @rrtoledo @berewt @wolf31o2 @Quantumplation @rkuhn @coot @nfrisby @sandtreader @colll78

One additional item to review: We have also had issues before with **images** for CIPs needing consensus that they were consistent with the objectives & statements in the CIP... so reviewers should please also double-check the numerous image changes.

**p.s.** and @ch1bo please no more force-push (though assuming the last one was a final stroke for consistency with your back end work) unless the branch being forced includes any commits assembled here in GitHub 🤓

Ryun1 (approved):
appreciate the maintenance of this CIP as development moves forward 
thank you authors 🤝

CIP-0164/README.md
```diff
@@ -12,11 +12,12 @@ Authors:
   - Andre Knipsel <andre.knispel@iohk.io>
   - Yves Hauser <yves.hauser@iohk.io>
   - Simon Gellis <simon@sundae.fi>
+  - Paul Clark <paul.clark@iohk.io>
 Implementors:
   - Input Output Engineering
 Discussions:
   - https://github.com/input-output-hk/ouroboros-leios/discussions
-  - https://github.com/cardano-foundation/CIPs/pull/1078
+https://github.com/cardano-foundation/CIPs/pull/1196
```
Ryun1:
```suggestion
  - https://github.com/cardano-foundation/CIPs/pull/1196
```

rphair: and add this back in:
`- https://github.com/cardano-foundation/CIPs/pull/1078`

CIP-0164/README.md
```diff
@@ -12,11 +12,12 @@ Authors:
   - Andre Knipsel <andre.knispel@iohk.io>
   - Yves Hauser <yves.hauser@iohk.io>
   - Simon Gellis <simon@sundae.fi>
+  - Paul Clark <paul.clark@iohk.io>
 Implementors:
   - Input Output Engineering
 Discussions:
   - https://github.com/input-output-hk/ouroboros-leios/discussions
-  - https://github.com/cardano-foundation/CIPs/pull/1078
+  - https://github.com/cardano-foundation/CIPs/pull/1196
```
rphair, line 21:
```suggestion
  - https://github.com/cardano-foundation/CIPs/pull/1078
  - https://github.com/cardano-foundation/CIPs/pull/1196
```
Anyone able to make this change?  Confirmed ready to merge at today's CIP meeting... as soon as the wayward change to add `this` PR link can be properly executed 😅 cc @perturbing @ch1bo

rphair (approved):

bwbush:
This looks fine to me, especially since there is a protocol parameter that can limit the total number of voters.

CIP-0164/README.md
rphair: Reintroduced required change as https://github.com/cardano-foundation/CIPs/pull/1196#discussion_r3305949083.  However the authors/collaborators choose to clean this up (simply adding `this` PR as the one & only change to the `Discussions:` list), it _does_ need to be cleaned up before we merge - @ch1bo, cc @perturbing @Ryun1

ch1bo:
@KtorZ @wolf31o2 despite this PR now being merged, feel free to chime in still on this! I was kind of assuming this stays open for longer, but your and your team's feedback would be much appreciated still .. in fact, any time :heart_hands:

PR #1202, opened 2026-05-24
Add CIP Validation Workflow and documentation

Ryun1:
### Changes

- Adds `.github/scripts/validate-cip.py` (driven by `.github/schemas/cip-header.schema.json` and wired into  `.github/workflows/cip-validation.yaml` workflow that runs on PRs touching `CIP*/README.md`)
- Adds `.github/CIP-VALIDATION.md` for documentation of validation
- Updates to `.github/CIP-TEMPLATe.md` to match changes.
- Small update on CIP-0001 to match changes.
- Added `Open Questions` to optional sections.

### Motivation

- Create parity with CPS validation workflow for CIPs
- Automate parts of CIP editor role
- Improve standardisation across CIPs, making downstream tooling easier to maintain

test cases can be seen passing via my fork [here](https://github.com/Ryun1/CIPs/actions/runs/26361301401/job/77597143515).


Ryun1:
Should we standardise the `Implementors` header field? its handled differently across proposals

Could do to match `Authors`

```yaml
Implementors:
  - Mr Implementor <https://implementors-site.com>
```

or to match to `Discussions` and `Solutions To`

```yaml
Implementors:
  - Mrs Implementor: https://implementors-site.com
```

rphair:
@Ryun1 this is brilliant, especially on a Sunday 😇
> Should we standardise the Implementors header field? its handled differently across proposals
> ```
>  - Mr Implementor <https://implementors-site.com>
>  - Mrs Implementor: https://implementors-site.com
> ```

I think this difference is expressed in proposals whenever authors (often self-) identify implementors as individuals or as organisations.  It also depends on the "scale" of the CIP & whether, as in the minority of cases, authors were offering their own methods for possible standardisation, or trying to extend their reference implementation: something I think we should still encourage unaffiliated developers to do.

Therefore I can see the advantage in supporting _both_ syntaxes; i.e.:
* YES I think we can standardise the header field, but
* NO we don't have to choose one or the other if the validator can fork to accept proper YAML compliance...

e.g. the same YAML labels that work on `Discussions` and `Solutions To` could also be applied to fully qualified `Implementors`:
```
Implementors:
  - Reference implementation: Handle Hierarchy <https://github.com/foobardev/hhandle>
```
I think that is the most future proof but @Ryun1 I'm willing to support whichever extension or restriction you think is best.  cc @perturbing

Ryun1:
@rphair

The flexibility is nice for the author
but not so nice for the downstream tooling (or validation efforts 😆).

for me, I think the `Authors` style will probably be good enough
and straight forward for downstream tools etc.

Ryun1:
another question

Currently `Solution To` is an optional entry in the header

I feel it would be better form and easier for maintenance and downstream tools
for it to be compulsory

Shall we make this change too?

rphair:
@Ryun1 that's a good suggestion on symbolic principle — and would be on general principle, if CIPs were read primarily by machines and secondarily by humans — but I believe it would be burden on authors and also give the customary critics of the CIP process more fuel for that fire (i.e. accusations of bureaucracy, difficulty, and barriers of entry):
* Most CIPs don't have any associated CPS so it would institute a mandatory parameter that most authors won't — and perhaps shouldn't — think of including.
* It would cater to the hypothetical criticism that CIP editors will require that every novel CIP has a corresponding CPS... a stigma from 2 or 3 years ago, and one that took some time & effort to dispel.
* It would defeat the labour-saving purpose of automation: by introducing errors when not including an unnecessary field in their own naturally-conceived document... which must then be corrected manually by authors & sometimes identified by editors manually.

Ryun1:
@rphair 
you always have such well reasoned responses -- thank you!
and I agree, lets keep it simple for authors

rphair (approved):
@Ryun1 this is fantastic work, well precedented both conceptually & operationally by the corresponding CPS PRs and updates.  I believe we should merge it ASAP (what do you think, @perturbing?) at roughly the same time as
* https://github.com/cardano-foundation/CIPs/pull/1125

... once its merge conflicts can be cleaned up... waiting too long could lead to a snowballing of merge conflicts that will often get pushed out to authors.

I believe we have altogether a responsibly designed validator & data set & now it's time to lock it all in: then all of us will stay vigilant over the results to clean up any divergence after merging everything. 🚀

perturbing (approved):
> Automate parts of CIP editor role

❤️ 

Thanks Ryan!

PR #1203, opened 2026-05-29
CIP-0057 | Standardize parametric (generic) type naming

matiwinnetou:
## Summary

Adds a normative section to CIP-0057 that specifies how parametric (generic) type instantiations are named in the `definitions` registry, how `$ref` strings target them under RFC 6901, and how `title` is to be populated.

The amendment is **purely additive** — no existing fields, keywords, or examples are changed. The new section codifies a single angle-bracket syntax for generic instantiations and acknowledges an earlier dollar-separator form as legacy that new producers MUST NOT emit and consumers MAY (but need not) accept.

The text is written so that a future reader of the CIP will read the parametric-types section as part of the original specification — no "earlier revisions", "we chose", or amendment-narrative voice in the CIP body itself. Specific producers are also not named in the normative or rationale sections.

### What the section specifies

**Interop-critical (MUST):**

- **Definition keys**: angle-bracket grammar — `Option<Int>`, `List<ByteArray>`, `Pair<Int,ByteArray>`, `Map<K,V>`, nested arbitrarily (`List<Option<types/order/Action>>`). Module separators stay as forward slashes inside the qualified name.
- **Tuples**: doubled angle brackets — `Tuple<<A,B>>`, `Tuple<<A,B,C>>` — to disambiguate from a hypothetical single-argument generic `Tuple<A>`.
- **`$ref` escaping**: standard JSON Pointer (RFC 6901) — `/` → `~1`, `~` → `~0`, angle brackets and commas pass through verbatim.
- **Type identity** for parametric instantiations is carried by the definition key / `$ref`; consumers MUST NOT use `title` to distinguish entries.

**Hint, not interop-load (SHOULD / MAY):**

- **`title`** SHOULD hold only the bare constructor name (`"Option"`, not `"Option<Int>"`) so it works directly as a code-generation hint, but producers MAY emit the parameterized form. Consumers MUST tolerate either form; if they derive identifiers from `title`, they strip the angle-bracket suffix.

### Legacy syntax handling

- **New producers MUST NOT** emit the dollar-separator form (`Option$Int`, `List$ByteArray`) or any other historical separator.
- **Consumers MAY support** legacy forms for backwards compatibility with older blueprints, but are NOT required to. A consumer that rejects legacy-form blueprints outright is still conformant with this spec.
- When a legacy-supporting consumer produces keys itself (e.g. synthesised auxiliary definitions during code generation), it MUST emit the modern angle-bracket form regardless of what it consumed.

### What this amendment intentionally does NOT do

- It does not deprecate or remove anything that existing tools depend on. Existing blueprints remain valid; only new producers are constrained on the MUST rules.

## Motivation (PR-side; not in the CIP text)

Today CIP-57 is silent on parametric-type naming. As blueprints have appeared from a widening set of producers, conventions diverged for separating a type constructor from its arguments (`Option$Int`, `Option<Int>`, and bespoke separators in between). Without a normative rule, every new producer is free to invent a third or fourth.

Concretely: Bloxbean's cardano-client-lib annotation processor has carried dual `<>` / `$` parsing for two years because it has to consume both. We're cleaning that out on the library side now and this PR is the upstream half of that work — pinning the convention so future producers don't reintroduce the divergence. SHOULD wouldn't actually prevent that for definition keys / `$ref` / encoding, so those rules are MUST. `title`, by contrast, doesn't carry type identity (consumers reach identity through the key), so a MUST there would constrain producers without unlocking interop and is SHOULD instead.

## Findings from real-world verification

The first draft of this amendment put a MUST on `title` requiring the unparameterized form. After authoring a batch of integration tests that compile real Aiken stdlib v3.1 contracts and feed them through code generation, I observed that one blueprint emits **both** forms in `title`:

- `Option<X>` → title `"Option"` (unparameterized — matches the rule)
- `Tuple<<A,B>>` → title `"Tuple"` (unparameterized)
- `Pairs<A,B>` → title `"Pairs<A, B>"` (parameterized)

A MUST that says "title is the unparameterized name only" would therefore declare real Aiken output non-conformant for one specific type while adding no interop benefit. Identity flows through the definition key and `$ref`, both of which remain MUST and unambiguous. This is why the title rule is SHOULD with explicit consumer-side tolerance for the parameterized form.

## Rationale (also expanded inline in the CIP)

- **Familiarity**: angle brackets are the dominant generics syntax in target languages (Java, TypeScript, Rust, C#, Scala). Code generators can use the key string nearly verbatim.
- **Compositionality**: nesting is unambiguous (`List<Option<Foo>>`); identifier-character separators collide with target-language identifiers and need extra escaping for nesting.
- **Ecosystem alignment**: current major producers already emit the angle-bracket form. Standardising on what new producers already emit minimises churn.

JSON Pointer's `~1` escape is reused rather than introducing a new mechanism, because RFC 6901 is already the substrate `$ref` is built on.

## Versioning / change-tracking

CIP-1 § Versioning treats CIP-text revisions as git-history-tracked rather than version-tagged ("this does not apply to the CIP text, for which annotated change logs are automatically generated... as a history of CIP files and directories"). CIP-57 has no Specification-versioning section today, and this change is purely additive — it does not introduce a Specification version concept or alter any existing schema.

The `Discussions:` list is updated to add [aiken-lang/aiken#1270](https://github.com/aiken-lang/aiken/issues/1270) — the upstream implementor discussion that motivated this change — alongside the existing aiken-lang/aiken#972 link. Editors who prefer the meta-PR (#1203) to also be appended per CIP-1 § Header Preamble ("must include a link to the pull request that created the CIP and any pull request that modifies it"), please say so and I'll add it.

## Implementors list change

Added `Bloxbean cardano-client-lib` (https://github.com/bloxbean/cardano-client-lib) to the Implementors list — it's been a blueprint consumer for ~2 years, just wasn't listed previously.

- **Consistency with existing producer output**: the new section is intended to match what current major producers already emit. Producer maintainers — please confirm (or flag divergence in) the angle-bracket grammar described here for your toolchain.
- **Tuple syntax**: the `Tuple<<A,B>>` (doubled angle bracket) convention is chosen to disambiguate from a single-argument `Tuple<A>`. If reviewers prefer the cleaner `Tuple<A,B>` and accept the ambiguity-with-single-arg edge case, happy to adjust.
- **`title` for parametric types**: should we go further and drop normative content on `title` entirely, leaving only "it's a human-readable hint"? Current draft says SHOULD prefer the unparameterized form to make it useful for code generation, but tolerates the parameterized form. Editor preference welcome.

---

(new text sections added, rendered:)
- [Parametric (generic) types](https://github.com/matiwinnetou/CIPs/tree/cip-0057-generics-syntax/CIP-0057#parametric-generic-types)
- [Parametric type naming](https://github.com/matiwinnetou/CIPs/tree/cip-0057-generics-syntax/CIP-0057#parametric-type-naming)

### DIFF
--- a/CIP-0057/README.md
+++ b/CIP-0057/README.md
@@ -17,7 +17,7 @@ Discussions:
   - https://github.com/cardano-foundation/CIPs/pull/258
   - https://discord.gg/yUkkhqBnyV
   - https://github.com/aiken-lang/aiken/issues/972
-  - https://github.com/cardano-foundation/CIPs/pull/1203
+  - https://github.com/aiken-lang/aiken/issues/1270
 Created: 2022-05-15
 License: CC-BY-4.0
 ---
--- a/CIP-0057/README.md
+++ b/CIP-0057/README.md
@@ -283,7 +283,7 @@ This keyword's value must be an array of valid _Plutus Data Schema_; possibly em
 
 Many high-level languages targeting Plutus support parametric polymorphism — type constructors that take other types as arguments (e.g. `Option<a>`, `List<a>`, `Pair<k, v>`, `Map<k, v>`, or user-defined ADTs such as `Wrapped<a>`). Once a compiler instantiates such a constructor with concrete arguments, the resulting type is a distinct on-chain type that may appear in `datum`, `redeemer`, `parameters`, or `fields`. Each distinct instantiation needs a distinct entry in the [`definitions`](#definitions) registry so that `$ref` strings can target it unambiguously.
 
-Earlier revisions of this CIP did not prescribe a syntax for naming such instantiations, and implementations have diverged: some emit `Option$Int` (legacy Aiken `< v1.1`), others emit `Option<Int>` (Aiken `>= v1.1`, Scalus), and bespoke tools have at times used `_` or `-` as separators. This divergence prevents code-generators from interoperating across producers.
+Earlier revisions of this CIP did not prescribe a syntax for naming such instantiations, and implementations have diverged: some emit `Option$Int`, others emit `Option<Int>`, and bespoke tools have at times used `_` or `-` as separators. This divergence prevents code-generators from interoperating across producers.
 
 This section defines the normative naming and encoding rules. Implementations producing `definitions` keys and `$ref` strings MUST follow them so that one blueprint can be consumed by tools across the ecosystem. Consumers MUST accept the modern angle-bracket form defined here; consumer behaviour for legacy forms is described in [Legacy syntaxes](#legacy-syntaxes) below.
 
@@ -306,7 +306,7 @@ In prose:
 3. Each type argument is itself a key matching this grammar — angle brackets nest arbitrarily deep for higher-order instantiations.
 4. Names of built-in / shared types (e.g. `Int`, `ByteArray`) MAY be unqualified (i.e. have no `/`-path prefix), at the producer's discretion.
 
-Examples of well-formed definition keys (from real-world Aiken stdlib v3 blueprints):
+Examples of well-formed definition keys:
 
 ```
 Option<Int>
@@ -316,12 +316,12 @@ List<Option<types/order/Action>>
 Option<List<Option<types/order/Action>>>
 Pair<Int,ByteArray>
 Map<ByteArray,Int>
-aiken/interval/IntervalBound<Int>
+my_contract/range/Bound<Int>
 ```
 
 ##### Tuple keys
 
-For ordered fixed-arity products (n-tuples) that some languages model as a distinct kind from ADTs, producers MUST use the reserved name `Tuple` with a single angle-bracket argument list whose contents are themselves wrapped in angle brackets — i.e. `Tuple<<A,B>>`, `Tuple<<A,B,C>>`. The doubled brackets distinguish tuples from a hypothetical single-argument `Tuple<A>` and match what current Aiken implementations emit.
+For ordered fixed-arity products (n-tuples) that some languages model as a distinct kind from ADTs, producers MUST use the reserved name `Tuple` with a single angle-bracket argument list whose contents are themselves wrapped in angle brackets — i.e. `Tuple<<A,B>>`, `Tuple<<A,B,C>>`. The doubled brackets disambiguate the tuple kind from a hypothetical single-argument generic `Tuple<A>`, which under this grammar would otherwise be indistinguishable.
 
 #### `$ref` strings
 
@@ -336,7 +336,7 @@ Examples:
 ```
 "$ref": "#/definitions/Option<Int>"
 "$ref": "#/definitions/Option<cardano~1address~1StakeCredential>"
-"$ref": "#/definitions/aiken~1interval~1IntervalBoundType<Int>"
+"$ref": "#/definitions/my_contract~1range~1BoundType<Int>"
 "$ref": "#/definitions/Pair<Int,ByteArray>"
 ```
 
@@ -348,7 +348,7 @@ When a parametric instantiation is registered, its `title` field MUST hold only 
 
 The following historical forms have appeared in blueprints emitted by older toolchains:
 
-- **Dollar-separator form**: `Option$Int`, `List$ByteArray` — used by Aiken `< v1.1` (stdlib `v1` era). Equivalent to the modern `Option<Int>`, `List<ByteArray>`.
+- **Dollar-separator form**: `Option$Int`, `List$ByteArray` — emitted by some early producers. Equivalent to the modern `Option<Int>`, `List<ByteArray>`.
 
 New producers MUST NOT emit any of these legacy forms; they MUST emit the modern angle-bracket form defined above.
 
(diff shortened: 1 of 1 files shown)


CIP-0057/README.md
```diff
@@ -12,10 +12,12 @@ Implementors:
   - OpShin <https://github.com/OpShin>
   - Lucid <https://lucid.spacebudz.io/>
   - Mesh.js <https://martify.io/>
+  - Bloxbean cardano-client-lib <https://github.com/bloxbean/cardano-client-lib>
 Discussions:
   - https://github.com/cardano-foundation/CIPs/pull/258
   - https://discord.gg/yUkkhqBnyV
```
rphair:
```suggestion
```
No longer a valid Discord link.

CIP-0057/README.md
```diff
@@ -12,10 +12,12 @@ Implementors:
...
   - https://github.com/aiken-lang/aiken/issues/972
+  - https://github.com/aiken-lang/aiken/issues/1270
```
rphair:
```suggestion
  - https://github.com/aiken-lang/aiken/issues/1270
  - https://github.com/cardano-foundation/CIPs/pull/1203
```
`this`... especially important because of your statement (currently in the OP):
> Motivation (PR-side; not in the CIP text)

... to ensure we have a record of when & why the CIP was updated.

rphair (commented):
@matiwinnetou looks good to me based on the provenance but please also see if @KtorZ @scarmuega & any others interested in this effort can provide additional confirmation.  Marking `Last Check` in expectation of that (or at least no objection to these changes) in the time before review at next CIP meeting: https://hackmd.io/@cip-editors/136

I'll approve this myself after some simple edits below, but please keep in mind some of these will produce (or avoid) a merge conflict; if so, it'll be with this pending PR which updates most CIPs' front matter (cc @Ryun1):
* https://github.com/cardano-foundation/CIPs/pull/1125

Ryun1:
Are these changes backwards compatible? with existing blueprints?
or could this break tools?

maybe these changes make more sense as a version 2.0?

perturbing:
@KtorZ , could you confirm this is an "okay" change? Or perhaps @Riley-Kilgore?

rphair:
@matiwinnetou my confirmation of this update was based on the presumption that you had practically taken over this project after having worked on it with @KtorZ (this is based on memory, going back years from things that were said in CIP meetings long ago).

If this is true and/or you have some means of confirming that this update is coming from a consensus of those currently working on <ins>and _with_</ins> the Plutus Contract Blueprint, the editors would like to see this before we continue toward merge.

In the meantime I believe I can clean up the merge conflict here (**p.s.**: https://github.com/cardano-foundation/CIPs/pull/1203/changes/b7902280b2a37dd13558e0e48a2218326f8fee2b), which was a result of the `master` base change after the editors' https://github.com/cardano-foundation/CIPs/pull/1125 was merged.

CIP-0057/README.md
rphair: Fixed in https://github.com/cardano-foundation/CIPs/pull/1203/changes/b7902280b2a37dd13558e0e48a2218326f8fee2b (while resolving merge conflict).

CIP-0057/README.md
rphair: Fixed in https://github.com/cardano-foundation/CIPs/pull/1203/changes/b7902280b2a37dd13558e0e48a2218326f8fee2b (while resolving merge conflict).

perturbing:
@Unisay, just to check, are these changes okay?

KtorZ:
@perturbing @rphair This is indeed conformant with what we "recently" changed in the Aiken blueprint generation 🫡

rphair (approved):

Unisay:
Thanks for pinning this down. The angle-bracket direction is the right call. One concern on the `$ref` encoding rule, which I think is technically incorrect as written.

The amendment says:

> Angle brackets (`<`, `>`) and commas (`,`) MUST NOT be percent-encoded; they appear verbatim. JSON Pointer does not reserve them.

with the example `"$ref": "#/definitions/Option<Int>"`. The amendment frames this as conformant (its Rationale says "no new escape semantics are introduced"), but the verbatim-`<>` rule contradicts both JSON Schema's requirement that `$ref` be a URI-Reference and RFC 6901 §6, the very section it cites. The mistake is treating `$ref` as a bare JSON Pointer when it is actually the URI form, and for `<`/`>` that changes the answer. Here it is step by step.

**Step 1: `$ref` is a URI-Reference, not a bare JSON Pointer.** JSON Schema 2020-12, [§8.2.3.1 "Direct References with $ref"](https://json-schema.org/draft/2020-12/json-schema-core#name-direct-references-with-ref):

> The value of the "$ref" keyword MUST be a string which is a URI-Reference. Resolved against the current URI base, it produces the URI of the schema to apply.

So `#/definitions/Option<Int>` is parsed as a URI-Reference whose fragment is `/definitions/Option<Int>`. That fragment has to satisfy the URI grammar in [RFC 3986, *Uniform Resource Identifier (URI): Generic Syntax*](https://datatracker.ietf.org/doc/html/rfc3986), not only the JSON Pointer escaping rules.

**Step 2: what RFC 3986 allows in a fragment.** RFC 3986 [§3.5 "Fragment"](https://datatracker.ietf.org/doc/html/rfc3986#section-3.5):

> ```
> fragment    = *( pchar / "/" / "?" )
> ```

with `pchar` from [§3.3 "Path"](https://datatracker.ietf.org/doc/html/rfc3986#section-3.3):

> ```
> pchar         = unreserved / pct-encoded / sub-delims / ":" / "@"
> ```

`unreserved` ([§2.3 "Unreserved Characters"](https://datatracker.ietf.org/doc/html/rfc3986#section-2.3)) is `ALPHA / DIGIT / "-" / "." / "_" / "~"`, and `sub-delims` ([§2.2 "Reserved Characters"](https://datatracker.ietf.org/doc/html/rfc3986#section-2.2)) is `"!" / "$" / "&" / "'" / "(" / ")" / "*" / "+" / "," / ";" / "="`. `<` and `>` are in none of these sets: not unreserved, not sub-delims, not `:`/`@`/`/`/`?`. The only way they can appear in a fragment is as `pct-encoded`.

**Step 3: so the angle brackets need encoding, but the commas are fine.** `Option<Int>` inside a fragment has to be written `Option%3CInt%3E`. The comma rule in the amendment is correct (`,` is a sub-delim, valid verbatim), but the same paragraph is wrong about `<`/`>`. They are not in the same boat.

**Step 4: RFC 6901 says this too, in the section that applies here.** The amendment leans on [RFC 6901, *JavaScript Object Notation (JSON) Pointer*](https://datatracker.ietf.org/doc/html/rfc6901), to justify pass-through, but it has two representations. [§5 "JSON String Representation"](https://datatracker.ietf.org/doc/html/rfc6901#section-5) defines only the `~0`/`~1` escapes, which is the form for a JSON Pointer used on its own:

> Because the characters '~' (%x7E) and '/' (%x2F) have special meanings in JSON Pointer, '~' needs to be encoded as '~0' and '/' needs to be encoded as '~1'.

[§6 "URI Fragment Identifier Representation"](https://datatracker.ietf.org/doc/html/rfc6901#section-6) is the one that applies inside `$ref`, and it adds the RFC 3986 requirement explicitly:

> A JSON Pointer can be represented in a URI fragment identifier by encoding it into octets using UTF-8 [RFC3629], while percent-encoding those characters not allowed by the fragment rule in [RFC3986].

`<` and `>` are exactly such characters, so RFC 6901 §6 already requires encoding them. "JSON Pointer does not reserve them" holds for the §5 string form but not for the §6 fragment form that `$ref` uses.

**Step 5: practical impact.** A lenient consumer that resolves `$ref` by string-matching will not notice. A consumer that hands the value to a conformant URI parser will. `java.net.URI("#/definitions/Option<Int>")` throws `URISyntaxException: Illegal character in fragment`. That matters here because cardano-client-lib, added to the Implementors list in this PR, is Java, so a `MUST NOT percent-encode` rule risks breaking the consumer the change is meant to help, depending on how it resolves refs.

rphair:
@Unisay we'll make sure your point above is addressed before merging this (cc @matiwinnetou @KtorZ).

rphair:
@matiwinnetou we confirmed that this is all in proper order & coming from a source that satisfied everyone at the CIP meeting today.  However there still appears to be an unaddressed point above from @Unisay in https://github.com/cardano-foundation/CIPs/pull/1203#issuecomment-4732484833 that I promised we would conclude before merging this.

This may not be solvable, or may be trivial, but we have to be sure.  Here is how I described it at the meeting, rightly or wrongly: "This is kind of an upstream standardisation issue presenting different more-or-less standard ways of naming something canonically: both of which can be considered right & therefore both are also wrong.  We just need someone to make an authoritative choice about how we are handling this on Cardano."

So @matiwinnetou this will still be at `Last Check` for the next meeting but if you & @Unisay can confirm resolution in the meantime we can merge this with editor consensus before then.  cc @KtorZ

Ryun1:
@matiwinnetou is this good to go?
or does there need a response to Unisay's comment? 🙏

rphair (commented):
@Ryun1 @perturbing I believe it was fair enough to mark this `Waiting for Author` a few weeks ago, requesting that we get confirmation from any devs/users of any tooling that may exist for the Plutus Contract Blueprint (as per CIP-0057 itself).
* However, even without that confirmation, this update actually reflects the latest working version that the CIP repository is able to obtain.
* Therefore I'm putting this back at `Last Check` where it was before we made that CIP meeting decision (in early July) to wait for the all-around confirmation that has not come.

Upon re-review I feel confident doing this after noting @matiwinnetou's initial https://github.com/cardano-foundation/CIPs/pull/1203#issue-4546766817 that these changes are "purely additive" and therefore on general principle incapable of producing the "breaking changes" the editors were worried about.

So @matiwinnetou @scarmuega @KtorZ @Unisay I'm going to see if we can merge this at the next CIP meeting (https://hackmd.io/@cip-editors/141).
* If my explanation above is not accepted — i.e. vetoed by another editor — or if no other editor accepts my rationale and approves this, I will close this as abandoned: since at that point we would have no way of progressing this without answers to the editors' and peers' review questions above.

Ryun1 (approved):

PR #1204, opened 2026-05-31
CIP-0189? | Variable Deposits as Pledge

Cerkoryn:
Been brainstorming this one for a bit.  I think it has some potential interesting use cases down the road for dReps while solving a few pain points for SPOs at the same time.

([Rendered](https://github.com/Cerkoryn/CIPs/tree/variable-deposits-as-pledge/CIP-0189))

## Open questions for discussion:

- Should this CIP allow the same locked pledge to support multiple roles, such as SPO and dRep, simultaneously?
- Should this CIP introduce some sort of delay when updating a pledge deposit to withdraw funds to prevent gaming of the system?


CIP-XXXX/README.md
```diff
@@ -0,0 +1,322 @@
...
+Implementors: []
+Discussions:
+    - https://github.com/cardano-foundation/CIPs/pull/?
+Created: 2026-05-31
+License: CC-BY-4.0
+---
+
+## Abstract
+
+Cardano currently treats stake pool pledge as a declared value rather than ADA that is directly locked as pledge. This creates operational confusion between "declared pledge" and "live pledge", requires a separate owner-stake satisfaction check, and allows mangled-address constructions where apparent pledge can be inflated through stake credential "frankenwallets". Meanwhile, dReps have only a flat registration deposit and no equivalent pledge primitive, limiting future designs that may want dReps to have visible skin in the game.
```
```suggestion
Cardano currently treats stake pool pledge as a declared value rather than ADA that is directly locked as pledge. This creates operational confusion between "declared pledge" and "live pledge", requires a separate owner-stake satisfaction check, and allows mangled-address constructions where apparent pledge can be inflated through stake credential "mangled addresses". Meanwhile, dReps have only a flat registration deposit and no equivalent pledge primitive, limiting future designs that may want dReps to have visible skin in the game.
```
(or _franken addresses_ if you prefer)

This is the main term to correct: TL;DR replacing "Frankenwallet" with "mangled" or "franken address" which I am sure you mean.
* <ins>**_The_**</ins> [Frankenwallet](https://frankenwallet.com) is a procedure for building an air-gap or sandbox OS on a drive that handles only secure & crypto workflow: its name comes from patching together mutually supportive workflow tools & sometimes bits of hardware.
* <ins>**_A_**</ins> [franken address](https://cardano.stackexchange.com/questions/9784/guarantee-not-dispensing-native-assets-to-a-franken-address) is another name for [what's more generally called a _mangled address_](https://cardano.stackexchange.com/questions/tagged/mangled-address) — suggested, as you say, in CIP-0019 (cc @KtorZ) but not defined there.

I coined the Frankenwallet term & published its original outline before any of these dated references to "franken address" but — perhaps because of little overlap between the target audiences — the terms have failed to differentiate perfectly.  This is first time I've seen one term being used in place of the other: but not the first time someone assumed a relationship between the two (which there isn't).

Maybe because of that tendency it might help your CIP if you used the equivalent but distinct term "mangled address" which might also help the two "franken" terms to be used less ambiguously in the long run... although you have author's license to use whichever you think makes the best connection with your proposal's own target audience.

CIP-XXXX/README.md
```diff
@@ -0,0 +1,322 @@
+---
+CIP: ?
+Title: Variable Deposits as Pledge
+Category: Ledger
+Status: Proposed
+Authors:
+    - Ryan Wiley <rian222@gmail.com>
+Implementors: []
+Discussions:
+    - https://github.com/cardano-foundation/CIPs/pull/?
```
```suggestion
    - Original PR: https://github.com/cardano-foundation/CIPs/pull/1204
```
`this` as usual.

CIP-XXXX/README.md
```diff
@@ -0,0 +1,322 @@
...
+
+The same mechanism is introduced for dReps. The existing dRep deposit parameter is renamed to `minDrepDeposit` and becomes the minimum allowed dRep pledge deposit. A dRep may lock any amount at or above that minimum as dRep pledge. This CIP does not change dRep voting power formulas or define a dRep compensation scheme. It creates a simple, queryable, on-chain pledge primitive that future CIPs may use for dRep incentives, compensation eligibility, concentration mitigation, etc.
+
+Pledged stake remains economically active. Pool pledge deposits continue to earn staking rewards through the pool they pledge to and continue to count toward the dRep chosen by the pool operator's voting delegation. dRep pledge deposits continue to count toward that dRep's voting power and continue to earn staking rewards through the stake pool chosen by the dRep.
+
+## Motivation: why is this CIP necessary?
```
```suggestion
## Motivation: Why is this CIP necessary?
```
(1 of 2) All these are capitalised now — pending modifications under way to the template & its nearly-complete validation workflow (cc @Ryun1)

CIP-XXXX/README.md
```diff
@@ -0,0 +1,322 @@
...
+
+Wallets, CLIs, explorers, and governance tools SHOULD display pool pledge and dRep pledge as locked ADA. Tools SHOULD distinguish:
+
+- pledge deposit currently locked,
+- pledge effective for the current reward or governance snapshot,
+- pending change that will become effective at a future snapshot, if applicable.
+
+Tools SHOULD stop presenting pool pledge as both "declared pledge" and "live pledge". Under this CIP, the pledge deposit is the pledge.
+
+## Rationale: how does this CIP achieve its goals?
```
```suggestion
## Rationale: How does this CIP achieve its goals?
```
(2 of 2)

CIP-XXXX/README.md
```diff
@@ -0,0 +1,322 @@
...
+
+### dRep pledge is deliberately limited in this CIP
+
+dRep compensation and dRep voting-power concentration are important topics, but they are not settled by this proposal. This CIP introduces dRep pledge as an on-chain primitive and includes it in ordinary delegated stake accounting, but does not add a special dRep compensation or voting-weight formula.
+
+Future CIPs can use dRep pledge as one input while preserving backward compatibility with the primitive introduced here.
+
+### Relationship to CIP-50 and other pledge proposals
+
+CIP-50 and related reward-sharing proposals change how pledge affects rewards. This CIP changes what pledge is.
```
```suggestion
### Relationship to CIP-0050 and other pledge proposals

CIP-0050 and related reward-sharing proposals change how pledge affects rewards. This CIP changes what pledge is.
```
Matching both your use of `CIP-0019` above and the general editorial recommendation (for the benefit of search engines, AI learning, text search, etc.) that writers always use the 4-digit, `0`- padded CIP numbers in `CIP-` tags.

CIP-XXXX/README.md
```diff
@@ -0,0 +1,322 @@
...
+
+The two ideas are complementary. A reward formula can still include pledge influence, leverage caps, or other pledge-based behavior. Under this CIP, those formulas read pledge from locked pool deposits rather than from declared pool parameters and owner stake checks.
+
+## Path to Active
+
+### Acceptance Criteria
+
+- [ ] CIP editors confirm that the ledger semantics are complete, unambiguous, and consistent with the CIP process.
+- [ ] Ledger maintainers confirm that pool and dRep deposits can be exposed to the reward calculation, registration rules, and state query interfaces required by this CIP.
```
```suggestion
- [x] Ledger reviewers confirm the Ledger part of the specification is complete, unambiguous, and consistent.
- [x] Ledger maintainers confirm that pool and dRep deposits can be exposed to the reward calculation, registration rules, and state query interfaces required by this CIP.
```
The "CIP editors" item here is redundant because:
* the editors would do this with respect to whatever `Category:` a CIP candidate is in, whether Ledger or not... and the CIP process applies to every CIP so such a review is already implicit for all documents merged into the repo;
* completeness & specificity are established _much_ better & more routinely by Ledger reviewers than they are by CIP editors.

If you choose to keep these two items here, they would be a prerequisite for merge: so I've pre-ticked these boxes.  In my opinion, these 2 statements aren't necessary: though we'll tag the usual Ledger reviewers later if you leave these lines in.

CIP-XXXX/README.md
```diff
@@ -0,0 +1,322 @@
...
+
+4. Testing and rollout
+   - Add ledger unit and property tests for registration, updates, refunds, minimum deposit enforcement, reward snapshots, governance snapshots, retirement, and deregistration.
+   - Run a dedicated testnet or preview-era test deployment with mainnet-like parameters.
+   - File any necessary constitutional or governance actions for parameter guardrail changes.
+   - Activate through a hard-fork governance action after ecosystem readiness is confirmed.
+
+## Open questions
+
+- Should we allow the same locked pledge to support multiple roles, such as SPO and dRep, simultaneously?
```
rphair: This is <ins>a CPS feature</ins> which is out of place as a top level heading in a CIP.  The question itself (and why it's important, even if unanswered) could be posed somehow in your _Rationale_ and maybe some potential answers elaborated based on peer review that develops in this PR.

CIP-XXXX/README.md
```diff
@@ -0,0 +1,322 @@
...
+
+### Pledged ADA remains economically active
+
+This CIP does not remove pledged ADA from staking rewards or governance voting power. It changes how the ADA is locked and attributed as pledged stake.
+
+Pool pledge deposits are required to support the pool they pledge to for stake-pool accounting, but their governance voting power follows the pledge owner credential's dRep delegation. dRep pledge deposits are required to support the dRep they pledge to for governance accounting, but their staking rewards and stake-pool weight follow the pledge owner credential's stake-pool delegation.
+
+This preserves the expected behavior that locked pledge is still ADA participating in the protocol. Because pledged ADA is held in ledger deposit accounting rather than ordinary UTxOs, the ledger MUST include pledged stake in the relevant stake-pool and governance distributions by explicit deterministic rule.
+
+### The frankenwallet path is removed
```
```suggestion
### The mangled address path is removed
```
Similarly to the first comment above (https://github.com/cardano-foundation/CIPs/pull/1204#discussion_r3331338790): only asking that the term you choose here (`mangled` or `franken`) matches the one in your Abstract.

rphair (commented):
thanks @Cerkoryn for another contribution along similar lines as recently.  At first glance it seems an idea well presented enough to be considered along with similar incentive-altering suggestions: though I wouldn't underestimate what a drastic change this would be for SPO expectations and perhaps the Ledger itself.

Once some questions of basic terminology are cleared up we can mark this for `Triage` & confirm introduction at the next CIP meeting (and tag Ledger for review in the meantime):

Straightpool:
I like the idea of the variable pool deposit to be the defining setting for pool pledge (no more owner wallets). The SPO controls the pledge and could run with the pledge, so the pledge provider(s) would need to have high trust in the SPO with a proper trust relationship.

This would be a strong fix for this long standing issue https://github.com/IntersectMBO/cardano-node/issues/2109 

However, just making pool deposit variable won't be sufficient for pledge IMHO. All of the following will be needed IMHO:
- The pool deposit would need to fully count as pledge and also earn delegation rewards for the SPO. Right now the deposit serves a very different purpose, right now it only needs to ward of pool registration spamming. By this we also set a minimum pledge of 500 ADA, governed by the existing pool deposit parameter.
- The new pool deposit which is now pledge should be fully composable with Defi as defined in Pogun. If Pledge has actual meaning like with CIP-7 or CIP-50 then it might not need to be composable. Otherwise it would have to IMHO or we will see less and less pledged pools, once Pogun is life.
- After this change is live, changing the reward address of the pool should require the signature of both the pools cold key and the reward address private key, to avoid risking the pledge on the cold key alone, which is file based for legacy pools (with no way to migrate but setup a new pool)

CIP-XXXX/README.md
rphair: @Cerkoryn I'll commit some of these routine things now that we are getting technically oriented review, to keep the PR thread unobstructed...

CIP-XXXX/README.md
Cerkoryn: I'll remove it from the CIP and add it to the OP of the discussion section instead.  I think I had some other open questions that are worth discussing too.

CIP-XXXX/README.md
```diff
@@ -0,0 +1,324 @@
...
+
+### The mangled address path is removed
+
+CIP-0019 explicitly allows the payment and delegation parts of a Shelley address to be controlled by different entities. This is useful and valid, but it means current pledge satisfaction follows delegation rights rather than payment ownership.
+
+Under this CIP, a pool's pledge is not inferred from payment credentials, stake credentials, or address composition. It is the ADA locked under that pool's `PoolDeposit` purpose. A hybrid address can still exist, and stake can still be delegated normally, but it cannot inflate a pool's pledge unless ADA is actually locked as that pool's pledge deposit.
+
+This also addresses the pledge-as-a-service concern described in cardano-node [issue 2109](https://github.com/IntersectMBO/cardano-node/issues/2109). A third party may still fund or provide ADA for a pool pledge deposit, but doing so creates a real locked deposit and therefore a materially different trust relationship from merely sharing or registering stake credentials.
```
Ryun1, line 236: do we know the impact of this on the network? is this widespread?

CIP-XXXX/README.md
```diff
@@ -0,0 +1,324 @@
+---
+CIP: ?
```
```suggestion
CIP: 189
```

rphair (commented):
@Cerkoryn thanks for helping kick this off at the CIP meeting today in which this was acknowledged as a candidate mainly due to @Straightpool's https://github.com/cardano-foundation/CIPs/pull/1204#issuecomment-4594051847 implying (to those present) _both_ that this idea would somehow be workable _and_ some version of it would be acceptable to SPOs.

Please rename the containing directory to `CIP-0189` and update the "Rendered" link in your OP. 🎉

@lehins @WhatisRT we are now ready for you to please give this a rigorous review from the Ledger side.  If & when this approach is confirmed possible, we would then progress to some broader review to help clarify whether this would be worth the "trouble" that it might take to implement.

There will also be process & philosophical implications for SPOs and so, according our unanimous common sense, they & any SPO advocacy groups should also be included in the discussion as soon as feasibility is established: maybe gradually, if that happens by degrees, to progressively gauge the "worth the trouble" factor (I recommended contacting the SPO Marketing team for a monthly SPO blast if & when that time comes).

CIP-XXXX/README.md
Cerkoryn: @Ryun1 I tried to figure out how that could be researched, but unfortunately I don't know if it can be definitively shown when a pool operator is using a mangled address for pledge.  I think the best effort require a lot of probabilities and guess work.

CIP-0189/README.md
```diff
@@ -0,0 +1,324 @@
...
+- `minPoolDeposit`: minimum allowed `poolPledgeDeposit`.
+- `minDrepDeposit`: minimum allowed `drepPledgeDeposit`.
+
+These parameters no longer define fixed registration deposits. They define the minimum amount of ADA that must remain locked for the corresponding role while that role is registered.
+
+Changing either parameter by governance changes the minimum required pledge for future registration and update certificates. Existing registered pools and dReps with deposits below a newly increased minimum SHOULD remain valid until their next registration update, unless a future hard fork explicitly defines a migration rule that requires top-ups.
+
+### Stake pool registration and updates
+
+Stake pool registration and update certificates MUST include an explicit `poolPledgeDeposit` amount and a pledge owner credential.
```
lehins, line 79: Stake pool registration and update already do include pledge owner key hashes.

We've pondered on adding the actual deposit to the certificate, since it could be useful in other circumstances.
One issue is that such change will have to span two eras, since we cannot make a backwards incompatible change like this on a single HF. We'll have to add new certificates and deprecate the old ones. Similarly to how we handled addition of deposit amount to Account (un)registration certificates.

So, such change would take at least a couple of years from the start of implementation, before it can be relied on

CIP-0189/README.md
```diff
@@ -0,0 +1,324 @@
...
+
+For a newly registered pool:
+
+- The transaction MUST lock a `poolPledgeDeposit` for the pool.
+- `poolPledgeDeposit` MUST be greater than or equal to `pp.minPoolDeposit`.
+- `poolPledgeDeposit` is recorded in the ledger's deposit accounting under the pool's `PoolDeposit` deposit purpose.
+- `poolPledgeDeposit` is the pool's pledge.
+- The pledge owner credential identifies the reward account that receives staking rewards attributable to this pledged ADA and whose voting delegation determines the dRep voting power for this pledged ADA.
```
lehins, line 87: Who is the pledge owner actually?
Whenever a deposit is being left upon registration of the pool it is the stake pool operator that pays it. Those funds are not associated with any account, thus it cannot be used for DReps voting power.
However, as I've mentioned in another comment, today there is only one account address that is actually associated with the stake pool, namely where stake pool rewards are deposited, which means that is the only one that can be potentially considered as a pledge owner that has mechanism for delegation to a DRep. If so, then it needs to be explicitly stated.

CIP-0189/README.md
```diff
@@ -0,0 +1,324 @@
...
+Implementors: []
+Discussions:
+    - Original PR: https://github.com/cardano-foundation/CIPs/pull/1204
+Created: 2026-05-31
+License: CC-BY-4.0
+---
+
+## Abstract
+
+Cardano currently treats stake pool pledge as a declared value rather than ADA that is directly locked as pledge. This creates operational confusion between "declared pledge" and "live pledge", requires a separate owner-stake satisfaction check, and allows mangled-address constructions where apparent pledge can be inflated through stake credential "mangled addresses". Meanwhile, dReps have only a flat registration deposit and no equivalent pledge primitive, limiting future designs that may want dReps to have visible skin in the game.
```
lehins, line 17: Could you please define what a "mangled address" actually means?

CIP-0189/README.md
```diff
@@ -0,0 +1,324 @@
...
+
+This CIP replaces declared stake pool pledge with actual locked ADA by making the pool pledge deposit variable and using that deposit as the pool's pledge. Instead of declaring pledge in pool parameters and then checking whether owner stake credentials collectively control enough delegated stake, a pool operator locks the intended pledge amount in the ledger deposit pot. The existing pool deposit parameter is renamed to `minPoolDeposit` and becomes the minimum allowed pool pledge deposit.
```
lehins, line 19: We can't rename parameters. It must be a new parameter that is introduced

CIP-0189/README.md
```diff
@@ -0,0 +1,324 @@
...
+
+The same mechanism is introduced for dReps. The existing dRep deposit parameter is renamed to `minDrepDeposit` and becomes the minimum allowed dRep pledge deposit. A dRep may lock any amount at or above that minimum as dRep pledge. This CIP does not change dRep voting power formulas or define a dRep compensation scheme. It creates a simple, queryable, on-chain pledge primitive that future CIPs may use for dRep incentives, compensation eligibility, concentration mitigation, etc.
```
lehins: Ditto
```suggestion
The same mechanism is introduced for dReps. The existing dRep deposit parameter is deprecated and a new one named `minDRepDeposit` is introduced, which becomes the minimum allowed dRep pledge deposit. A dRep may lock any amount at or above that minimum as dRep pledge. This CIP does not change dRep voting power formulas or define a dRep compensation scheme. It creates a simple, queryable, on-chain pledge primitive that future CIPs may use for dRep incentives, compensation eligibility, concentration mitigation, etc.
```

CIP-0189/README.md
```diff
@@ -0,0 +1,324 @@
...
+
+Pledged stake remains economically active. Pool pledge deposits continue to earn staking rewards through the pool they pledge to and continue to count toward the dRep chosen by the pool operator's voting delegation. dRep pledge deposits continue to count toward that dRep's voting power and continue to earn staking rewards through the stake pool chosen by the dRep.
```
lehins, line 23: This is a massive change to semantics that we have today. Today stake pool deposit does NOT earn staking rewards or participate in governance. Also, because such deposit is not associated with any account address, it cannot be used for delegating to any DRep.
It would be a fairly complex change to stake distribution computation to make the deposit participate in earning staking rewards, and it is even harder task to make it participate in the governance, assuming we would use AccountAddress  associated with the stake pool where stake pool receives rewards as the source of truth for delegation to a DRep.

CIP-0189/README.md
```diff
@@ -0,0 +1,324 @@
...
+
+## Motivation: Why is this CIP necessary?
+
+### Current pool pledge is declared, not locked
+
+Stake pool pledge is currently a value declared in pool registration parameters. At reward calculation time, the ledger checks whether the pool owners collectively control enough stake to satisfy the declared pledge. If the pledge is met, the declared pledge value is used in the reward calculation. If it is not met, the pool receives no rewards for that epoch.
+
+This mechanism has two drawbacks.
+
+First, pledge is operationally more complex than it needs to be. A pool operator declares a value, ensures enough stake is delegated through pool owner stake credentials, and relies on ledger accounting to classify that stake as owner stake. The deposit required to register the pool is a separate fixed value and does not itself represent the pool's pledge. This also creates common confusion between declared pledge and live pledge.
```
lehins, line 33: It is not only operationally complex, but it is also complex from the implementation point of view. However, any time there is a suggestion to replace a complex part of Ledger with something simple we need to remember that we are not replacing, we are adding new logic, because we have to preserve old logic forever. So, from perspective of implementation this becomes even more complex, not simpler.

A simpler solution would be to provide better documentation and tooling that brings clarity to the current system.

CIP-0189/README.md
```diff
@@ -0,0 +1,324 @@
...
+
+Second, declared pledge is vulnerable to confusing address constructions. CIP-0019 describes Shelley addresses as having a payment part and a delegation part. These parts are usually controlled by the same party, but can be controlled separately in "mangled addresses" or "hybrid addresses". Since current pledge satisfaction follows stake credentials, not payment credentials, ADA controlled by one party's payment credential may still count as another party's pool owner stake if the address uses that pool owner's stake credential. This does not create stake from nowhere and does not double-count stake, but it can inflate apparent pledge without requiring the pool operator to actually lock the claimed amount.
```
lehins, line 35: >  it can inflate apparent pledge without requiring the pool operator to actually lock the claimed amount.

A stake pool owner would not risk sending funds to someone else while using a stake credential that belongs to such owner, because such output can be spent by the third party, bringing down the total amount below the declared pledge amount, thus causing stake pool loose all rewards. So, the only sensible thing an SPO can do is to retain control of "Owners" accounts. Note also that "Onwers" cannot be script addressed, which means SPOs cannot use a smart contract to lend those funds to anyone or participate in any other on chain financial processes.

So, IMHO this is a moot point.

CIP-0189/README.md
```diff
@@ -0,0 +1,324 @@
...
+
+A related long-standing node issue describes "pledge as a service", where pool operators may collect stake keys from third parties and register those credentials as pool owners. This can make pledge collaboration resemble ordinary delegation more than a high-trust owner relationship. This CIP addresses that issue by making pledge depend on ADA locked as a pool pledge deposit rather than on registered owner stake credentials.
```
lehins, line 37: I personally see this as a feature that multiple people can collaborate on satisfying pledge for a particular pool. Important point it cannot be done as a service, since actual stake pool owner has to trust those collaborators not to pull the carpet from underneath and reduce the pledge below the declared amount, which again would force the SPO to forfeit the rewards.

It seems like a totally viable approach where N people decide to collaborate on operating a pool. They can create a multi-sig account that uses all of those "owners" credentials for the account where rewards will be deposited to.
Which forces mutual trust: if any one of them misbehaves, none of the collaborators get the rewards.

CIP-0189/README.md
```diff
@@ -0,0 +1,324 @@
...
+
+### Deposits already express locked ADA
+
+The ledger already tracks deposits for stake credentials, stake pools, dReps, and governance actions. Deposits are returned only through the relevant deregistration, retirement, or finalization path. This CIP uses that existing accounting concept for pledge: a pool's pledge is the ADA locked under its pool pledge deposit, and a dRep's pledge is the ADA locked under its dRep pledge deposit.
```
lehins, line 41: One more time I'd like to drive the point of the existing mechanism that locks ADA in the deposit, thus making that ADA unable to participate in any other on chain mechanism. That was done by design and is essential for the security of the system as a whole.

CIP-0189/README.md
```diff
@@ -0,0 +1,324 @@
...
+
+- The transaction MUST lock a `poolPledgeDeposit` for the pool.
+- `poolPledgeDeposit` MUST be greater than or equal to `pp.minPoolDeposit`.
+- `poolPledgeDeposit` is recorded in the ledger's deposit accounting under the pool's `PoolDeposit` deposit purpose.
+- `poolPledgeDeposit` is the pool's pledge.
+- The pledge owner credential identifies the reward account that receives staking rewards attributable to this pledged ADA and whose voting delegation determines the dRep voting power for this pledged ADA.
+
+For an already registered pool:
+
+- Submitting a pool update certificate MUST NOT require pool retirement or deregistration.
```
lehins: This is already the case, so I don't see a point mentioning it
```suggestion
```

CIP-0189/README.md
```diff
@@ -0,0 +1,324 @@
...
+
+For transition compatibility, an implementation era MAY retain, deprecate, or ignore the existing serialized declared pledge field before removing it completely. Regardless of the transition encoding, ledger semantics after activation MUST treat `poolPledgeDeposit` as the sole source of pool pledge.
+
+### Pool pledge rewards and voting power
+
+A `poolPledgeDeposit` is pledged stake.
+
+For stake-pool accounting:
+
+- `poolPledgeDeposit` MUST count as stake delegated to the pool that the pledge deposit applies to.
```
lehins, line 107: As mentioned before, this is a serious change to how we handle deposits today. Deposit today is explicitly excluded from participating in earning staking rewards or voting, because that is exactly what is meant by a deposit: value is locked for the purpose of securing the system and is unusable for anything else.

lehins (changes requested):
In my personal opinion the significant complexity that this CIP would introduce to the Ledger implementation is not worth the effort. Moreover, current staking system had an immense research backing it up, while changes proposed in this CIP are based on an opinion and has no mathematical backing to it. For example, what is the impact of making deposit earn rewards and participate in governance? We would need to involve R&D and game theorists before we introduce changes like this to such a critical part of Cardano as its Proof of Stake mechanism.

While reading this CIP I have more questions than answers, eg:

* What happens when minPoolDeposit is increased? Do all pools that do not satisfy this minimal requirement stop receiving rewards?
* How does one reduce the pledge amount? Is the difference refunded on the epoch boundary when update is applied, since that is when normally the deposit is refunded? What happens when minPoolDeposit go updated at the same epoch boudary?
* What is the migration path from the system that we have today to the proposed one. We can't just all of a sudden start using locked deposit as pledge, since that will affect all stake pools at the hard fork boundary!

I am very much against this CIP, because:
1. I do not see problems that are being brought forth as actual problems.
2. Complexity in implementation that it would introduce is pretty significant.
3. Solution proposed has still too many problems to be considered a viable solution.

My recommendation is: before we start changing ledger we need to properly identify the problem and put some effort into trying and solve the pain points with tooling. I am more than confident this is the case here. So, I highly recommend first writing up a proper CPS and then we can use it as a starting point for coming up with a solution.

CIP-0189/README.md
Cerkoryn: This is just a Shelley address whose payment and delegation credentials are controlled by different parties.  Also called a "hybrid address" according to [CIP-0019](https://github.com/cardano-foundation/CIPs/tree/master/CIP-0019#delegation-part).

CIP-0189/README.md
Cerkoryn: Fair point.  My goal was to simplify pledge in a way that could better facilitate “pool alliances,” where multiple parties jointly fund and control a pool’s pledge through an appropriate multisig or script arrangement.  But in hindsight, trying to do this via deposit probably makes this worse, not better.

Cerkoryn:
Thank you for your review @lehins.  I also had some concerns about the complexities of this approach, but after reading your review it is clear that I'll need to take this idea back to the drawing board.  In particular trying to use a deposit to participate in staking/voting introduced more risks than I had previously considered.

> My recommendation is: before we start changing ledger we need to properly identify the problem and put some effort into trying and solve the pain points with tooling.

The problems I was primarily trying to address are perhaps best explained in this [comment](https://x.com/Hornan7/status/2051356261766078470?s=20) by @Hornan7 on X.  Additionally it would introduce a pledge mechanism that could be used for dReps in various ways in the future and hopefully clear up some of the community confusion between live vs. declared pledge.

@rphair feel free to close this PR at your earliest convenience 🙏

rphair:
Thanks @Cerkoryn for conceding these points and for making a sincere attempt to resolve some issues raised by @Straightpool @Hornan7 and others over the years.  The raw material is informative and thought-provoking at least, and especially including @lehins' https://github.com/cardano-foundation/CIPs/pull/1204#pullrequestreview-4509243064 this PR gets enough of the ideas onto the "drawing board" for a CPS to get a productive definition of the problem with some more attractive solution(s) next time. ⭐

PR #1207, opened 2026-06-06
CIP-0191? | Ekklesia - Hydra L2 Voting Protocol

Crypto2099:
This CIP describes the Ekklesia Voting Protocol: A Hydra-Powered voting and polling system leveraging CIP-67 tokens for ballot and result discovery as well as cryptographic proofs and auditability.

The CIP defines two new tokens under the CIP-67 schema: 600 + 601

- 600 Token: Defines and anchors the ballot definition to the L1 via token + inline datum. This provides an on-chain record of what was voted on. I have attempted to keep the standard at least spiritually aligned with CIP-179 | On-Chain Polling to attempt to promote tool reusability instead of needing to create entirely new tooling.
- 601 Token: This token is minted at the same time as the 600 Ballot token and is committed into the Hydra head where votes are recorded. Once voting is concluded, a results datum is attached to the 601 token and it returns to the L1 upon head closure, containing the evidence of voting results.

The CIP further describes the datum schema and auditing format for independent validation.

[Link to Rendered Version](https://github.com/Crypto2099/CIPs/blob/ekklesia/CIP-XXXX/README.md)


CIP-XXXX/README.md
```diff
@@ -0,0 +1,3496 @@
+---
+CIP: XXXX
+Title: Ekklesia - Hydra-Powered L2 Voting Protocol
```
```suggestion
Title: Ekklesia - Hydra L2 Voting Protocol
```
As submitted it's an effective title for service introductions and blog articles, but `Powered` may be too much of a marketing term for a CIP title.  You know CIPs are all about interrelated systems and it would be burdensome to objectively say every time which provides the "power" behind the others.

Currently the term `Hydra` is distinctive enough vs. other CIPs: and I think its existence alone provides an incentive for similar integrations to come... though we wouldn't want to have to say `-Powered` for all of them.

From my general perspective the `Hydra` designation means the same thing as `Hydra-Powered` and if more technical reviewers can confirm there's no ambiguity then the brevity would indeed produce a better title.

CIP-XXXX/README.md
```diff
@@ -0,0 +1,3496 @@
...
+merkle-proof-based inclusion-verification system, and a
+[CIP-179][CIP-0179]-compatible question/answer format that supports seven voting
+methods (binary, single-choice, multi-choice, range, ranked, weighted, and
+likert).
+
+CDDL schemas are provided for the two on-chain datums (600, 601) so an
+independent implementer can reproduce the on-chain bytes from this document
+alone.
+
+## Motivation: why is this CIP necessary?
```
```suggestion
## Motivation: Why is this CIP necessary?
```
(1 of 2) update for new CIP template

CIP-XXXX/README.md
```diff
@@ -0,0 +1,3496 @@
...
+  └─ Burn all voter tokens
+
+Phase 5: CLOSE (Hydra)
+  └─ Close Hydra head (no further updates possible)
+
+Phase 6: SETTLE (L1)
+  └─ (601) token returns to L1 via fanout with final datum
+```
+
+## Rationale: how does this CIP achieve its goals?
```
```suggestion
## Rationale: How does this CIP achieve its goals?
```
(2 of 2)

CIP-XXXX/README.md
```diff
@@ -0,0 +1,3496 @@
...
+
+The §7.8 auditor verification algorithm reflects this scope: it validates the
+integrity, completeness, and authenticity of the recorded vote set against the
+on-chain commitments. It does **not** opine on whether a ballot "passed."
+
+## Specification
+
+### 1. Token architecture
+
+#### 1.1. [CIP-67][CIP-0067] label allocation
```
```suggestion
### Token architecture

#### [CIP-67][CIP-0067] label allocation
```
(and so on; note the scope of this change is broken [by recent bugs on GitHub] so don't take it literally)

Someday we, or those who come after us, would have problems with updating this spec; either having to reorder it (changing the numbers) or to remove the numbers... both of which would break the GitHub Markdown anchors to every heading in the document.

The tables here (e.g. _Question and answer format_, _Vote methods_) will be usable (maybe vital) as references & therefore very likely to be linked externally in implementations & discussions and maybe for ballot / proposal writers or voters.

Maybe the from the authors' point of view it's unlikely that this spec will never be updated in a way that adds, removes, or recorders sections.  If so, can we imagine the next 10-year span when we've all moved on and some upgrade in the future needs (e.g.) to put in a more significant paragraph or table before the others in its section?

Removing the section numbers is even more important if the schemas themselves remain embedded in the text document rather than being pulled out into separate files (see next point): since these are much more likely to be referred to individually from external sources.

(links to other discussions on GitHub about this same subject available on request: though not yet producing any desire to "enforce" this in CIP-0001 or CPS-9999)

CIP-XXXX/README.md
```diff
@@ -0,0 +1,3496 @@
...
+pair sort), whose leaves are `blake2b_256(JSON.stringify(question))` over each
+entire question object using the compact serialization of §7.7. That leaf
+computation binds `contentHash` (and every other question field) to the on-chain
+commitment. The same value appears as `content_hash` in the (600) on-chain
+datum.
+
+##### 2.2.2. JSON Schema
+
+The following JSON Schema (Draft 2020-12) is normative. A document that fails
+this schema is not a conforming ballot definition.
```
rphair: These schemas are big enough to use independently as documentation (maybe even using in source code); regardless of size, wouldn't these be easier to work with (e.g. "validate against" the schema as often said here) if they were in separate files?  
- 2.2.2. JSON Schema
- 6.3. Vote-evidence JSON Schema
- 7.5. Full-results and merkle-proof JSON Schema
- 7.6. Auditor-companion artifacts

If that idea is acceptable, could we set a boundary: e.g. so that all schemas & code of individual, practical use be kept in separate files... and all those for illustrative properties remain embedded?  As you know this has worked well for many other CIPs (including your own).

CIP-XXXX/README.md
```diff
@@ -0,0 +1,3496 @@
...
+requiring the auditor to read the script source, trust a reproducible build, and
+reason about the script's behavior under all inputs.
+
+A future revision of this standard **MAY** define a Plutus-policy variant by
+specifying the normative properties any policy **MUST** enforce (decoupled from
+the script type) and the §7.8 augmentations needed to verify those properties
+against an arbitrary Plutus script. Until then, conforming ballots
+**MUST** use the native script defined above.
+
+#### 1.4. Why not [CIP-68][CIP-0068]?
```
rphair: Thanks for answering this question pre-emptively because it was the first question I had myself about the implementation after CIP-0067 was mentioned.  Strictly it would be a question for the _Rationale_ but I agree with your decision to move it up while the spec is being considered: otherwise that question would be kicking around in the reader's head all the time in between.

For review's sake we should check with some other implementors to make sure this design choice is documented so that everyone would make the same choice confidently... copying authors & reviewers of CIP-0068 extensions to see if they have any questions or reservations: @thaddeusdiamond @SamDelaney @yHSJ @fallen-icarus @SmaugPool @AndrewWestberg

rphair (commented):
@Crypto2099 @MadOrkestra this is an amazing contribution.  I wondered if something like this would be coming out as this other candidate has been discussed:
* https://github.com/cardano-foundation/CIPs/pull/1107
* (cc @Thomas-nada @thenic95 @gitmachtl @mpizenberg @Ryun1)

As @Cerkoryn replied here (https://github.com/cardano-foundation/CIPs/pull/1107#issuecomment-3994250798), the scope of the Ekklesia proposal may better suit larger implementations while other methods like candidate CIP-0179 might be set up more easily for smaller requirements.
* — **p.s.** with another crucial difference emphasised later (https://github.com/cardano-foundation/CIPs/pull/1107#issuecomment-4675366948) in that CIP-0179 surveys would be kept entirely on-chain)

I hope the authors & anyone with an interest in this subject can come for the introduction of this PR at the next CIP meeting: https://hackmd.io/@cip-editors/136

I'll note some review points below as detail comments (though they cover the whole document in scope) so that discussion can continue on each point in its own thread.
* About presentation choices (e.g. section numbering, embedded files) the author would have license to do things as they prefer: though perhaps after some practical discussion first.

CIP-XXXX/README.md
Crypto2099: I'm good with this :)

CIP-XXXX/README.md
Crypto2099: Gotcha, I'll go through and remove the section numbers then.

CIP-XXXX/README.md
Crypto2099: I was on the fence already whether to remove them to external files or to include them in the body. I'm happy to move them to external files within the repository.

CIP-XXXX/README.md
Crypto2099: This should now be complete, also identified one ambiguous heading that would have relied on ordering so deduplicated that and removed all inline MD syntax from headings as well to ensure downstream parsers and displays do not choke on weird syntax.

rphair: thanks @Crypto2099 - that all looks great now.

kenerik:
This is a very good proposal.
Ton's of usecases and supplement the utily of Hydra and the ability to quickly scale Cardano as L1 where it leave a proof on L1 after the hydra head close.
If i understand the math here correctly

rphair (commented):
@Crypto2099 @MadOrkestra - as expected this was confirmed a candidate at the CIP meeting today; please change the directory name & the above readable proposal link accordingly. 🎉

Also expected, a possible alignment with this concurrent proposal was mentioned:
* https://github.com/cardano-foundation/CIPs/pull/1107

... although what "alignment" might mean we can't yet anticipate.  It could just be that "good things" could happen for both proposals based on mutual review: even if their applicable domains were to be confirmed as, or remain, separate. - cc @Thomas-nada @mpizenberg @Cerkoryn @thenic95 @gitmachtl @Hornan7

CIP-XXXX/README.md
```diff
@@ -0,0 +1,3558 @@
+---
+CIP: XXXX
```
```suggestion
CIP: 191
```

CIP-XXXX/README.md
```diff
@@ -0,0 +1,3558 @@
...
+Authors:
+    - Adam Dean <adam@crypto2099.io>
+    - Mad Orkestra <mad@madorkestra.com>
+Implementors:
+    - Ekklesia <https://app.ekklesia.vote>
+    - Intersect <https://hydra-voting.intersectmbo.org>
+    - Civitas <https://www.civitasexplorer.com/>
+    - Fetch <https://cardano.fetchswap.io/governance>
+Discussions:
+    - https://github.com/cardano-foundation/CIPs/pull/1207
```
```suggestion
    - Original PR: https://github.com/cardano-foundation/CIPs/pull/1207
```

rphair (commented):
@Crypto2099 the change suggestion below is to fix 1 of the 3 validation problems reported at the bottom of the PR.  The other two are:
- We are still waiting for a branch owner (cc @MadOrkestra ?) to rename the containing directory to `CIP-0191`.
- `- Body references 'CIP-0179' but no such folder exists in the repository (use 'CIP-0179?' if it is still in PR)` could be "fixed" immediately but is still waiting editorial agreement that there is no "mandatory" syntax for referring to a candidate CIP/CPS and therefore this condition cannot be imposed by the validator (cc @Ryun1).

MadOrkestra:
not me ;)

PR #1208, opened 2026-06-07
Network parameter adjustment for staking competitiveness

nikkelj:
Network parameter adjustment for staking competitiveness
[README](https://github.com/cardano-foundation/CIPs/pull/1208/changes/d7d410c2bcdafc02b2d92c2d5d9b4f2868f714d6#diff-f1100127bca15d640544f2d4d387a728b86d6fadb67c71c7a1aa4e84a6b5119d)

### DIFF
--- a/CIP-0182/README.md
+++ b/CIP-0182/README.md
@@ -1,0 +1,138 @@
+---
+CIP: 42069
+Title: Network Parameter Adjustment for Staking Competitiveness and Adoption
+Category: Ledger
+Status: Proposed
+Authors:
+    - Jonathan Nikkel <nikkelj@gmail.com>
+        - dREP ID: drep1yfqkrlpz60efhuhz7atuy60m8ygqtef8ylfw96munn8vxqgj0pq92
+        - Bio: https://gov.tools/drep_directory/drep1g9slcgkn72dl9chh2lpxn7eezqz72fe86t3wklyuempszrm80p9
+Implementors: []
+Discussions:
+    - Original PR: https://github.com/cardano-foundation/CIPs/pull/?
+Created: 2026-06-07
+License: CC-BY-4.0
+---
+
+## Abstract
+
+This proposal recommends adjusting two Cardano protocol parameters — the monetary expansion rate (`rho`) and the treasury tax rate (`tau`) — to materially increase staking returns and improve Cardano's competitiveness as a proof-of-stake network. Specifically, it proposes raising `rho` from `0.003` to `0.0042` and reducing `tau` from `0.2` to `0.02`. Together, these changes are projected to raise the annualised staking yield toward approximately 8%, up from the current range of 2–4%, making Cardano staking meaningfully more attractive relative to competing networks and to traditional fixed-income alternatives. This is a pure protocol-parameter update and does not require a hard fork.
+
+## Motivation: Why is this CIP necessary?
+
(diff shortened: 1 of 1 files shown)


rphair:
@nikkelj before we can review this please:
* remove the out-of-range, self-assigned CIP number everywhere;
* remove the promotional material from the CIP header; also keeping in mind `Authors:` is not a nested list of contact information;
* put a link to the README file in your OP above as a courtesy to readers & reviewers of your proposal.

Also please note that the current governance conditions (having changed a lot in 4 years since some simple parameter change CIPs were accepted) suggest to me that CIPs which only change protocol parameters are better off presented as parameter change proposals (with their own governance path) instead... @ptrdsh @Cerkoryn @thenic95 can you please help us confirm or correct this notion?

nikkelj:
Updated, thanks for the feedback. I adjusted the CIP# to the apparent next in sequence. There were some that did not follow sequentially, so I thought I could just pick a number for the meme.

nikkelj:
If it turns out folks prefer a parameter change proposal instead, let me know and I will make one of those instead. I could use a pointer to where to make that though.

rphair:
@nikkelj you would have to read [CIP-0001](https://github.com/cardano-foundation/CIPs/blob/master/CIP-0001/README.md) and comply with the procedure there if you intend to submit a CIP.  It states there clearly (a couple times) that nobody chooses their own CIP numbers.

And yes, I would recommend finding out how & where you can submit a parameter change proposal: rather than trying to argue for a CIP to be accepted on this subject (I wouldn't be in favour of it myself, unless mandated somehow between a governance proposition & a hard fork):
1. one of those tagged above in my last comment might be able to recommend a next step
1. you can check the [Cardano Forum Parameter Change Proposals (PCP) category](https://forum.cardano.org/c/governance/parameter-change-proposals-pcp/257), where there's a pinned post & also an older posting "Instructions: how to use the PCP process"

rphair:
@nikkelj attendees at today's CIP meeting (coming from diverse places across the Cardano ecosystem) confirmed that parameter change proposals — while eventually intended to face on-chain voting as a tagged `PCP-` proposition (_not_ a CIP)  — should first be discussed in social networks and/or specialist groups: probably those focused in SPO advocacy in this case.

So posting your request on the Cardano Forum's PCP category above would likely be one of those 2 good places to start... the participants already there can likely indicate what their requirements would be to progress your proposal.

We all asked each other what "real person" might be able to provide guidance about this & no one was suggested at the meeting.  Whatever you find in your own investigation of this process, feel free to post here so CIP editors can inform any future requests accordingly.

Ryun1:
hey @nikkelj 

as per the constitution
parameter change governance actions require off-chain socialisation before being proposed on-chain as a Parameter Change governance action

The CIP process isn't the place for this off-chain socialisation.
I would suggest raising a discussion via the [Cardano Forum](https://forum.cardano.org).

There is a potential route Parameter Change process via the Intersect Parameter's Committee based on [Parameter Change Requests](https://forum.cardano.org/c/governance/parameter-change-proposals-pcp/257).
That is a potential route.

nikkelj:
Thanks all, I will start a thread in parameter change requests to discuss further

nikkelj:
Posted here: https://forum.cardano.org/t/pcp-rho-tau-jonathannikkel/155202

Issue #1216, opened 2026-06-23
Pending validator fixes + maybe link checking


rphair:
Thanks to @Ryun1 for producing the CIP and CPS validators that are well on their way to saving labour for editors & contributors alike: as confirmed at the CIP meeting today. 🎉

We have solved the big problem earlier reported — no permission for unmerged authors to run GitHub Workflow — thanks to an emeritus editor at the CF changing repository settings.
* FYI (in case users ask again, as in https://github.com/cardano-foundation/CIPs/pull/1199#issuecomment-4774561949) any part of this problem that remains is unavoidable...GitHub still considers certain accounts "new" even if they're been created a while ago but not had enough cooperative activity.

The following problems still remain (cc @perturbing); I can quote the linked parts if to help discussion if if there's any uncertainty:
* [ ] **Candidate name formatting** - requires exact syntax `CIP-XXXX?` for unmerged CIP tags (and similarly for CPS) even though other language like "candidate CIP-XXXX` should be acceptable.
  * Since there is no "universal" way to say that a CIP candidate isn't merged yet, and the `?` suffix goes out of date immediately after it's merged, as I said in the review threads for the affected CIPs I believe we have no way to impose this requirement.
  * Live examples:
    * Currently blocking validation of high profile CIP [here](https://github.com/cardano-foundation/CIPs/actions/runs/27882549081/job/82512493787?pr=1207) & I don't want to ask @Crypto2099 to "fix" it when there's nothing wrong with his CIP: https://github.com/cardano-foundation/CIPs/pull/1207#pullrequestreview-4538769930
    * Author had to fix this manually here, pending editors' deliberation & fix: https://github.com/cardano-foundation/CIPs/pull/1107#issuecomment-4736032806
  * Reasons: reported on GitHub already in
    * https://github.com/cardano-foundation/CIPs/pull/1107#discussion_r3408645295 ([point # 3 here](https://github.com/cardano-foundation/CIPs/pull/1107#pullrequestreview-4492046079))
* [ ] Doesn't catch **directory names with right CIP number but wrong 0-padding**.
  * in https://github.com/cardano-foundation/CIPs/pull/1199#pullrequestreview-4549383899 — editor had to spot problem manually & author had to fix manually.
* [x] (**p.s.**) Requirement that CIP and CPS header uses **`?` marks for an unmerged CIP number** needs to be broadened.
[...]

PR #1218, opened 2026-06-26
CIP-0192? | Transaction Fee-Change Field

polinavino:
This is a proposal to implement a mechanism for transactions to receive change to a specified reward address in case of fee overpayment. That is, only the minimum fee is collected, and anything over the `txfee - minfee` is returned.

The purpose of this is not to use the mechanism as-is, but to facilitate future development of a dynamic pricing mechanism. The actual change calculation of the amount of change given will depend on the dynamic pricing mechanism, but will be executed using the change mechanism we describe here.

---

([rendered proposal latest version](https://github.com/polinavino/CIPs/blob/fee-change/CIP-0192/README.md))


CIP-????/README.md
```diff
@@ -0,0 +1,185 @@
...
+Title: Transaction Fee-Change Field
+Category: Ledger
+Status: Proposed
+Authors:
+    - Polina Vinogradova <polina.vinogradova@iohk.io>
+    - Will Gould <will.gould@iohk.io>
+Implementors:
+  - Input Output Engineering <https://iohk.io>
+Discussions:
+  - Tiered Pricing repo internal discussion: https://github.com/input-output-hk/tiered-pricing/pull/36
```
rphair, line 12: The repo `input-output-hk/tiered-pricing` appears taken down (I looked but can't find where it may have been moved to); maybe when the old Tiered Pricing project was cancelled? https://www.iog.io/news/change-of-course-acropolis-tiered-pricing

It would be good to see that these discussions had been archived anywhere & please update the link if so.

CIP-????/README.md
```diff
@@ -0,0 +1,185 @@
+---
+CIP: ?
+Title: Transaction Fee-Change Field
```
rphair, line 3: I'll update the PR title to this CIP title because it's much more clear & specific: especially since the term "fee change" is itself ambiguous.  If you have other ideas for a technically correct _and_ inspirational title please let's share some ideas at the meeting.

CIP-????/README.md
```diff
@@ -0,0 +1,185 @@
...
+   validation of Plutus scripts
+- **Fixed portion of fee refunded**: rejected because the correct amount depends
+  on fee dynamics that are not yet fully specified.
+
+## Path to Active
+
+### Acceptance criteria
+
+- [ ]  This CIP has been communicated to, reviewed by, and agreed upon by a representative group of community members
+- [ ]  All major concerns have been addressed
```
```suggestion
## Path to Active

### Acceptance criteria

- [ ]  This CIP has been agreed upon by a representative group of community members
- [ ]  All major concerns have been addressed
```
@polinavino (cc @lehins @WhatisRT) I think we need more detail for what may be required for this proposal to become `Active` (the _Implementation plan_ below seems straightforward enough).

We'll also have to perform due diligence to ensure that the CIP has been "communicated to" _and_ "reviewed by" Ledger representatives, multi-node developers, and wallet architects _**before** this CIP is merged_ 🤓 and so therefore no need to list this explicitly.

After that stakeholder & community review, whatever "concerns" may remain _specifically about the determination of `Active` status_ would go here in place of the final list item.

rphair (commented):
thanks @polinavino & looking forward to `Triage` this submission at next CIP meeting & hope you can join us there: https://hackmd.io/@cip-editors/138

Consideration as a candidate and assignment of a CIP number would be based on confirmation of basic feasibility from someone behind the Ledger implementation.  @lehins @WhatisRT: to that end, can you give us a quick opinion on this?

**p.s.** cc @willjgould

CIP-????/README.md
polinavino: the repo will become public shortly and this link will work

CIP-????/README.md
polinavino: i am not sure what details it would make sense to add here. Do you mean like, what hard fork this would go into? Or make it dependent on the exact tiered/dynamic pricing strategy?

CIP-????/README.md
```diff
@@ -0,0 +1,185 @@
...
+This proposal discusses a mechanism for users to receive change for fee overpayment. 
+The amount of change given is the difference between the minimum fee calculated for that 
+transaction and the fee specified by the transaction. The change is sent to the 
+reward address specified by that transaction. 
+
+## Motivation: Why is this CIP necessary?
+
+Cardano currently collects the exact fee a transaction declares; there is no
+mechanism for the protocol to return any portion of it to the submitter. Future fee
+schemes may require such a mechanism.
```
fallen-icarus, line 29: > ... *may* require ...

"may require" and "will require" are two different things. I understand the need for making phased changes, but it seems too early to say this is the right direction to go in. We don't have a clear picture of the destination yet.

I see it is *opt-in* but is that really good enough to warrant a change before we know the destination? I can't evaluate if this is a good idea or a bad idea until I know the end goal. What if we use neither dynamic or tiered pricing?

polinavino: We have an upcoming CIP about the pricing mechanism in the next couple of weeks that should hopefully clear things up. "Will" require is the current status

CIP-????/README.md
rphair: @polinavino I was hoping to get some idea about how to make this section less vague from the first round of review by the Ledger reps and I apologise if my first reply was confusing on its own.  On second look: unless Ledger adds anything, I think you already have most of the major points in your Implementation Plan.

Inclusion in a hard fork would definitely be appropriate in the _Acceptance Criteria_ since then it would be `Active` without any doubt.  The other items currently there aren't required for `Active` status and apply more to the Implementation Plan (except those parts implicit in CIP review).

I'll suggest another reorganisation accordingly & resolve this point for now...

CIP-????/README.md
```diff
@@ -0,0 +1,185 @@
...
+- [ ]  This CIP has been agreed upon by a representative group of community members
+- [ ]  All major concerns have been addressed
+
+### Implementation plan
+
+1. Update the formal ledger specification (`TxBody`, UTXO rule)
+2. Update node ledger implementation/CDDL once the formal spec is stable.
+3. Update wallet SDKs and tooling.
+4. Test net deployment 
+5. Main net deployment
```
rphair: Removing redundant points from _Acceptance criteria_, anticipating that "major concerns" should be addressed immediately after any deployment stage, assuming a hard fork will be required, and improvising some community tests to demonstrate `Active` status....

Please let me know if you think anything should be changed before committing (note the first box is already ticked because we'll be doing it in the CIP review itself):
```suggestion
### Acceptance criteria

- [ ] Ledger changes live on Mainnet and included in a hard fork.
- [ ] Latest version of `cardano-cli` supports fee change specification by default.
- [ ] At least 2 mainstream Cardano wallets support fee change specification.

### Implementation plan

- [x] Agreement upon this CIP by a representative group of community members
- [ ] Update the formal ledger specification (`TxBody`, UTXO rule).
- [ ] Update node ledger implementation/CDDL once the formal spec is stable.
- [ ] Update wallet SDKs and tooling.
- [ ] Testnet deployment: addressing all major concerns
- [ ] Mainnet deployment: addressing all major concerns
```

rphair (commented):
... continued from https://github.com/cardano-foundation/CIPs/pull/1218#discussion_r3517224254 - needed to change scope of this suggestion:

CIP-????/README.md
```diff
@@ -0,0 +1,187 @@
+---
+CIP: ?
```
```suggestion
CIP: TBD
```
... or `X` or some other metasyntactic null.  This has long caused a problem in front matter rendering on GitHub & so editors are lately trying to change this convention after the initial years of requiring it. 😖

rphair (commented):
To fix the [validation error](https://github.com/cardano-foundation/CIPs/actions/runs/28713663476/job/85151249079?pr=1218) currently reported below:

CIP-????/README.md
```diff
@@ -0,0 +1,187 @@
+---
+CIP: TBD
```
```suggestion
CIP: \?
```
We (w/@Ryun1)  are working on the particulars in discussion of https://github.com/cardano-foundation/CIPs/issues/1216 - likely to conclude soon & I'll update CIP-0001 as soon as all the details are agreed upon & implemented in the validator.  In the meantime escaping the `?` should keep it happy (the alternative `TBD` that I suggested hasn't yet been implemented in practice).

rphair (commented):
@polinavino my apologies over my own uncertainty regarding our CIP validator which is in production the first time this week (current [validation failure](https://github.com/cardano-foundation/CIPs/actions/runs/28819773127/job/85468239751?pr=1218)):

CIP-????/README.md
rphair: p.s. apparently this too is incorrect & the CIP validator doesn't recognise escaping the `?`... it [must be double-quoted](https://github.com/cardano-foundation/CIPs/actions/runs/28713663476/job/85151249079?pr=1218) even though other syntaxes would resolve the YAML error (cc @Ryun1).  I'll commit this to avoid bothering anyone again & should get green-light from validator after that.

CIP-????/README.md
```diff
@@ -0,0 +1,187 @@
...
+
+There is precedent for a fee change mechanism in 
+[Ethereum](https://eips.ethereum.org/EIPS/eip-1559), which has been, to 
+some extent, the inspiration for the mechanism discussed in this CIP.
+
+## Specification
+
+### New transaction body field
+
+A new optional field `feeChangeAddr` is added to the transaction body `TxBody` :
```
lehins: We no longer call them RwdAddr, due to CIP-159, which promotes reward addresses to proper accounts
```suggestion
A new optional field `feeChangeAccount` is added to the transaction body `TxBody` 
```

CIP-????/README.md
```diff
@@ -0,0 +1,187 @@
...
+
+```
+feeChangeAddr : Maybe RwdAddr
```
lehins:
```suggestion
feeChangeAddr : Maybe AccountAddress
```

CIP-????/README.md
```diff
@@ -0,0 +1,187 @@
...
+```
+
+Where the reward (staking) address `feeChangeAddr` is the address to which the
```
lehins:
```suggestion
Where the `AccountAddress` in `feeChangeAccount` is the address of an account to which the
```

CIP-????/README.md
```diff
@@ -0,0 +1,187 @@
...
+compensation is credited. Using a reward address rather than a payment address
```
lehins:
```suggestion
compensation is credited. Using an account address rather than a payment address
```

CIP-????/README.md
```diff
@@ -0,0 +1,187 @@
...
+ensures the credit is handled through the existing rewards-withdrawal mechanism
```
lehins:
```suggestion
ensures the credit is handled through the account based mechanism introduced in CIP-159
```

CIP-????/README.md
```diff
@@ -0,0 +1,187 @@
...
+address optional allows for backwards compatibility. 
+
+### Ledger rule changes
+
+#### Transaction validity
+
+No additional signing requirements are imposed, as it is not mandatory for a 
+user to output fee change into an address they own. 
+
+**`feeChangeAddr` provided.** When a transaction provides
```
lehins:
```suggestion
**`feeChangeAccount` provided.** When a transaction provides 
```

CIP-????/README.md
```diff
@@ -0,0 +1,187 @@
...
+
+- a summand is added `feeChange pp utxo txb`, where we define 
+
+`feeChange pp utxo txb = txb .txFee - minfee pp utxo tx`
+
+(2) In the `UTXO` rule, we add a check 
+
+- `0 ≤ feeChange pp utxo txb`
+
+(3) We update the reward on the ledger state with this calculation,
```
lehins:
```suggestion
(3) We update the account balance on the ledger state with this calculation, 
```

CIP-????/README.md
```diff
@@ -0,0 +1,187 @@
...
+#### CBOR encoding
+
+The `feeChange` field is encoded in the transaction body map under a new key `x`.
+The key `x` is the next available field number in the era into which it is added.
+The value will be encoded as `bytes` representing the reward address.
+
+```
+transaction_body =
+  { ..existing fields..
+  , ? x : reward_account   ; feeChange
```
lehins:
```suggestion
  , ? x : account_address   ; feeChangeAccount
```

CIP-????/README.md
```diff
@@ -0,0 +1,187 @@
...
+  }
+```
+
+
+## Rationale: How does this CIP achieve its goals?
+
+### Choice of reward address over payment address
+
+Reward addresses are already used for staking rewards and pool
+deposits, making them a natural fit.
```
lehins:
```suggestion
Account addresses are already used for staking rewards and pool
deposits, and crucially for CIP-159 making them a natural fit.
```

CIP-????/README.md
```diff
@@ -0,0 +1,187 @@
...
+Crediting a reward address ensures the UTxO set changes remain predictable 
+locally.
```
lehins:
```suggestion
Crediting a account ensures the UTxO set changes remain predictable 
locally. Most importantly accounts do not have any minimal deposit requirement, like minUTxO, thus change returned can be as low as 1 lovelace.
```

CIP-????/README.md
```diff
@@ -0,0 +1,187 @@
...
+
+The reason for this is that all Plutus scripts
+should be blind to the fee change provided. This value is not fixed at the time of 
+transaction construction and therefore cannot be used to predict the outcome of 
+script validation at the time of construction. Note here that this change may 
+in the future make it 
+impossible for a script to itself check the produced/consumed balance, as 
+the `feeChange` calculation in future eras may require data not included 
+in `TxInfo` (e.g. if it depends on how long the transaction was waiting in 
+the mempool).
```
lehins: Safest thing to do is to fail transaction validation if an older plutus script is present in a transaction together with `feeChangeAccount` in the same transaction.

There could be an implicit assumption by a Plutus Script that an account that was fully drained by the transaction has zero ADA, while this mechanism can violate that assumption.

That being said, I don't think this restriction is a must have, if we need ability to use `feeChangeAccount` in presence of older scripts then I don't think it would be dangerous to simply ignore translation of this field in the TxInfo for those older scripts.

CIP-????/README.md
```diff
@@ -0,0 +1,187 @@
...
+
+#### CBOR encoding
+
+The `feeChange` field is encoded in the transaction body map under a new key `x`.
```
lehins:
```suggestion
The `feeChangeAccount` field is encoded in the transaction body map under a new key `x`.
```

CIP-????/README.md
```diff
@@ -0,0 +1,187 @@
...
+### Choice of reward address over payment address
+
+Reward addresses are already used for staking rewards and pool
+deposits, making them a natural fit.
+Crediting a reward address ensures the UTxO set changes remain predictable 
+locally. 
+
+### Opt-in via optional field
+
+By making `feeChange` optional, this CIP imposes zero overhead on transactions that
```
lehins:
```suggestion
By making `feeChangeAccount` optional, this CIP imposes zero overhead on transactions that
```

lehins (commented):
Thus far I can't think of any dangers in introducing a change like this to the Ledger, from the perspective of the Ledger safety.

However, it does bring a level of complexity from the tooling perspective.

For example I can see how this could be a challenge for various indexers, which will no longer be able to look at the transaction and easily say what was the fee collected by the transaction. They will either have to implement the fee calculation logic that is identical to Ledger's or inspect the whole transaction looking through all uses of the account address in withdrawals and direct deposits, including sub-transactions, while also getting the original balance of that account just so they can figure out the fee that was actually collected.

On a different note, I believe it is important to mention in this CIP, that the collateral amount will not be affected by this new field and the amount specified in the `fee` field will continue being used for deriving the required collateral amount.

CIP-????/README.md
```diff
@@ -0,0 +1,187 @@
...
+
+```
+transaction_body =
+  { ..existing fields..
+  , ? x : reward_account   ; feeChange
+  }
+```
+
+
+## Rationale: How does this CIP achieve its goals?
```
Ryun1, line 153: Including some discussion around, the potential impact of this change on ecosystem tooling would be appreciated in Rationale

CIP-????/README.md
```diff
@@ -0,0 +1,187 @@
+---
+CIP: "?"
```
```suggestion
CIP: 192
```

rphair (commented):
thanks again @polinavino ... based on @lehins's positive review and acknowledgement of the potential of this CIP (as well as other forthcoming work on the pricing system) at the CIP meeting today, this has been declared a candidate.

As @Ryun1 concurrently noted above (https://github.com/cardano-foundation/CIPs/pull/1218#discussion_r3537944718), the impact on tooling would be substantial enough to justify including some expectations (and maybe some planned responses or countermeasures) in the _Rationale_ to help prepare the developer community to accommodate this.

Looking forward to response to the reviews above & further review; please rename the containing directory to `CIP-0192` and update the "rendered proposal" link accordingly in your OP 🎉

lehins:
>  I can see how this could be a challenge for various indexers, which will no longer be able to look at the transaction and easily say what was the fee collected by the transaction.

@polinavino I was pondering on this problem for a bit on my walks and realized that there could be a very nice simple solution to this. We could add a new field `changeAmount :: Coin` to the transaction, instead of a transaction body, similar to the `isValid` flag. Where this field would be populated by the Ledger, instead of the user that submitted a transaction.
This would allow indexers and other tools to simply rely on the value in the field instead of figuring out the value on their own

polinavino:
> We could add a new field changeAmount :: Coin to the transaction, instead of a transaction body, similar to the isValid flag.

good idea, i will write that as an option or the main suggestion shortly

CIP-????/README.md
polinavino: will do!

CIP-????/README.md
polinavino: I am going to look at the Accounts CIP now, and edit this accordingly after I learn more about the proposal

polinavino:
Also, I will discuss this CIP's possible effect on tooling, and add that the collateral mechanism will not be affected

rphair:
@polinavino it's nice to see this moving along (all the issues I can recall so far look resolved) & please post when you think it's ready for another round of review.

colll78:
I would prefer we don't do this. If the user wants to overpay the fee intentionally, we should allow this. There is a huge amount of value in the fact that the outcome of a successful transaction can be entirely determined from the transaction body. Sacrificing this property, especially unnecessarily, doesn't seem desirable. 

For a healthy fee market or dynamic pricing mechanism, the ledger should not return arbitrary change depending on the network conditions, the fee spent should always be deterministic and be exactly what was committed to in the transaction body. If the user sets the tx fee as 5,000 Ada, then they should pay exactly 5,000 Ada in fees. 

I cannot understand why we would want to introduce such complexity specifically to refund users who overpay fees. The point of a dynamic pricing mechanism is that it empowers the user to overpay to prioritize their transaction. There is no need to give change for this. If it turns out there was capacity anyway and the prioritization was unnecessary, they should of-course still pay for it.

fallen-icarus:
@colll78 We want *worst-case determinism*. Full determinism like you are describing is unnecessary and economically wasteful since it means users frequently *overpay* the block's true price. That's not good.

*Worst-case determinism* gives actors the predictability they want in a way that doesn't unnecessarily waste economic resources (which includes the actor's ADA).

colll78:
> @colll78 We want *worst-case determinism*. Full determinism like you are describing is unnecessary and economically wasteful since it means users frequently *overpay* the block's true price. That's not good.
> 
> *Worst-case determinism* gives actors the predictability they want in a way that doesn't unnecessarily waste economic resources (which includes the actor's ADA).


If we want to sacrifice determinism for *worst-case determinism*, then there are significantly larger improvements we can make, we make massive sacrifices specifically to achieve complete determinism. Full determinism is an extremely valuable property though, and I disagree with the assessment that sacrificing it without an extremely thorough analysis of the benefits and consequences is the right choice. 

I am extremely confident that if we only plan to leverage this sacrifice in determinism to allow the return of excess fees to users, then it is not worth it whatsoever. The complexity such a mechanism would introduce, calculating fee change for priority fee txs based on capacity is absolutely insane, and the value is comparatively abysmal. There is a reason that no other chain does this. If a user wants to prioritize their transaction, then they have to pay for that prioritization regardless of actual network conditions. If they chose to pay for prioritization in a near-empty block, there is still work that said prioritization incurred on the nodes, and they must be charged accordingly. 

There are many cases where this sacrifice does not give the predictability actors want at all. I want to know my exact balance after a transaction, the total amount of assets I will possess if the transaction is included in the ledger. Worst case determinism doesn't provide that. One of the key value propositions of Cardano is that your signature approves an exact balance delta. You know exactly what will leave your wallet and what will enter it, providing you with perfect accounting. With the proposed approach, this is no longer the case, you go from knowing what your balance is to knowing "my balance is somewhere between X and Y", that's a really crappy accounting system, and fundamentally it's no stronger than Ethereum's, which already includes *worst-case determinism* via approvals (at most a TX can decrease your balance by what you approved). So we go from a system that has strictly more accounting predictability than Ethereum to one that has equivalent predictability.

fallen-icarus:
> ... Ethereum's, which already includes worst-case determinism via approvals (at most a TX can decrease your balance by what you approved).

Ethereum does not have *worst-case determinism*. On Ethereum, I pay for failed txs and I also can't lock in a specific "state snapshot" like I can by specifically encoding certain UTxOs into a Cardano tx (ie, my tx fails for free automatically if the state has changed).

> I want to know my **exact balance after a transaction**, the total amount of assets I will possess if the transaction is included in the ledger.

Knowing your exact balance at the *time of signing* is not as useful as you seem to assume. We are talking about a variance of a few ADA and I will know the actual exact balance as soon as I see the block which confirmed my tx.

Even with tx chaining, you can account for the variance since you actually know the worst-case *liability* of each tx.

---

Please give a concrete use case where worst-case determinism is not enough. As an example where full determinism is worse: consider arbitrage bots. They are racing against each other and need to outbid each other. If they constantly end up paying way more than necessary, then they can't arbitrage as much which means price discovery suffers.

EDIT: I want to make my stance clear. People generally want bounded downside with uncapped upside. So the debate right now is centered on *exact-outcome determinism* versus *worst-case determinism*. My stance is we need the latter, not the former. We are only talking about the fee paid here. Everything else in the tx is still deterministic.

colll78:
>If they constantly end up paying way more than necessary, then they can't arbitrage as much which means price discovery suffers.

The market reality disagrees with you. Both Solana and Ethereum do not refund gas for prioritization based on network utilization and both of those networks have extremely liquid markets due to massive arbitrage. The impact of introducing such a mechanism (to dynamically determine a congestion fee post tx submission based on dynamic network conditions) into a blockchain would be immense amounts of complexity and inefficiency.  

>If they constantly end up paying way more than necessary
Importantly what defines "way more than necessary"? If you charge based on some arbitrarily constructed formula around the value of prioritization based on congestion state, then by construction you rob the market of the ability to actually define what the actual economic value of prioritization is. 

A priority fee purchases an ex-ante entitlement to priority treatment. Its value cannot be inferred solely from whether the entitlement happened to become pivotal after the block was constructed. A person buying priority is purchasing protection against contention. The fact that contention did not ultimately materialize does not mean the service had no ex-ante value. Otherwise, the protocol is effectively providing a priority option and collecting its premium only in the states where the option is exercised.

Calling every amount above the realized controller quote an “overpayment” assumes the conclusion. A history-dependent utilization controller produces a protocol-administered posted price. It does not discover a unique, objective “true price” for a specific user’s urgency. 

If the priority fee charged is exactly what the user signs for, then the market itself will determine the economic value of prioritization. Arbitrage bots will set priority fees according to the economic value that ex-ante entitlement to prioritization offers. This makes price discovery natural based on real value, as opposed to derived from an arbitrarily imposed formula. If a user is at risk of liquidation and is willing to pay 1000 Ada in tx fees to maximally increase their entitlement priority, and as a result their position gets reinforced, and they avoid liquidation and save thousands of dollars and the network makes 1000 Ada in fee revenue, why would it be beneficial to refund them? The transaction was mutually beneficial for both parties.

fallen-icarus:
> Both Solana and Ethereum do not refund gas for prioritization based on network utilization ...

That's not true. [EIP-1559](https://eips.ethereum.org/EIPS/eip-1559) actually does refund for over payment.

colll78:
>That's not true. [EIP-1559](https://eips.ethereum.org/EIPS/eip-1559) actually does refund for over payment. 

It doesn't refund over payment of priority fee, which you can see labeled as `max_priority_fee_per_gas` in the EIP you linked. 

If you read the EIP you'll see it explicitly states that senders
>“will pay the priority fee per gas set in the transaction”

The fee calculation is:
```
gas_used * priority_fee_per_gas
```

What gets refunded is unused `maxFeePerGas` headroom. 

So for example if:
```
maxFeePerGas          = 100 gwei
maxPriorityFeePerGas  =   5 gwei
baseFee               =  20 gwei
```

The user pays:
```
20 + 5 = 25 gwei/gas
```

Note that priority fees are always paid as specified without any refund.

polinavino:
> I cannot understand why we would want to introduce such complexity specifically to refund users who overpay fees. 

The decision of whether or not to refund overpayment is a question of incentives. It is very likely that some fee overpayment will have to be refunded. In an upcoming CIP we will address this. In this CIP, we are focusing on establishing the mechanism by which this can take place. So, the question is whether this would serve as a suitable mechanism for providing change in the case that the goal of "giving change for fee payment" is a valid one.

colll78:
> It is very likely that some fee overpayment will have to be refunded.

I strongly disagree. I don't see why this would be the case when the base fee is deterministic. The goal of "giving change for fee payment" is not valid. Without a concrete definition of what constitutes "overpayment", we cannot discuss the mechanism for providing change.

polinavino:
This mechanism is optional so you do not need to provide a change address, but any overpayment will be going into the treasury. This strict change mechanism is to ensure that the two-tier pricing system does not turn into an arbitrary-number-of-tiers pricing system, which will (1) make it really hard for the mempool to handle and (2) allow higher-paying txs to incentivize block producers to evict tx paying less (by making them invalid rather than pushing them further back in the queue) - this i not the intention of tiered pricing

colll78:
This change mechanism is unnecessary for tiered pricing. Just add a price tier field (uint8) to the tx body, and ledger rule just validates in phase 1 tx fee >= fee_required_by_tier

rphair (commented):
@colll78 I hear you — and would assume your https://github.com/cardano-foundation/CIPs/pull/1218#issuecomment-5430650568 is both correct & relevant until proven otherwise — but there is a substantial bit of work coming along here that refers to some definitions in the CIP-0192 candidate:
* https://github.com/cardano-foundation/CIPs/pull/1265

... and without _something_ to call `CIP-0192` these references will be orphaned.

I would doubt the merged product can be perfect in any case, because even this new CIP draft coming in says itself that some things about its [Ledger Rule Changes](https://github.com/willjgould/CIPs/blob/cip-urgency-signalling/CIP-urgency-signalling/README.md#ledger-rule-changes) (cc @lehins) for urgency signalling in linear Leios should diverge from CIP-0192 (cc @polinavino @willjgould in case this requires further explanation).

So I am approving this & proposing that this be merged (cc @perturbing @Ryun1) if only to prevent dead ends in the documentation, **although**:
* Please we will need both Leios (cc @ch1bo) and Tiered Pricing advocates (including @fallen-icarus for external / impartial review) to please identify any reason why this _cannot_ be merged — or might need any disclaimers upon how / whether to actually implement this — ideally before this comes up for `Last Check` at the next CIP meeting: https://hackmd.io/@cip-editors/143
* Of course we will demote it from that status if anyone identifies an issue here that needs amendment before merging it: or also if there is any issue that would disqualify it completely... and in the latter case @polinavino @willjgould I suppose you would need to amend your references in the "Urgency Signalling" CIP draft accordingly.

rphair (dismissed):

rphair (commented):
@polinavino @willjgould  in a tentative `Last Check` at CIP meeting review today (not as lengthy a review as we would have needed before actually merging this) @colll78 summarised this CIP as catastrophic to Cardano due to its effects on transaction determinism.
* Without enough time that would have allowed us to record more opinions, the quick consensus was to take this down from `Last Check` until this matter could be fully clarified & settled.
* I'll record what I've noted & remembered but please I hope the principals in the discussion will correct whatever I may have wrong & keep providing more detail in further comment.

1 - @colll78 noted the fixed inputs and outputs (including collateral) that Cardano implementations have come to depend upon, as well as highlighting some worst case scenarios about astonishing fees being offered in transactions with no accounting of where the protocol would put them.
* The specific word used was "evil" which I personally have no way to assess but to consider this issue to be paramount for reconciliation in the review that follows.

2 - @Crypto2099 and @perturbing noted together that this CIP is not justifiable on its own (https://github.com/cardano-foundation/CIPs/pull/1218#discussion_r3516342967), requiring a follow-up CIP to be created and implemented.
* This would be a condition we have so far avoided in the CIP process: since it makes either CIP impossible to review on its own.
* @polinavino I'm not sure the best workaround for this: possibly to merge this proposition with the "future" CIP draft?
  * This would solve the impossibility of a traditional review & perhaps also allow the proposed adverse effects to be considered along with what might be gained from the _entire_ solution (since commentary above suggests this CIP is not a great value on its own).

3 - Earlier above, @fallen-icarus remarked that indeterminism may be OK in the fee as long the rest of the transaction is deterministic (https://github.com/cardano-foundation/CIPs/pull/1218#issuecomment-5232905280).
* Therefore this is another review point to confirm or reject: hopefully qualifying exactly "how much" of the overall determinism is lost & what consequences would be expected.
* My own understanding from the way Cardano has been marketed over the years is that people's expecation of Tx determinism should be <ins>absolute</ins> unless there is something undeniably irresistible to be gained from sacrificing it.

I hope that makes a concrete list of review points: if not, the editors can attempt to make a review checklist later.  In the meantime we would have to rely on tentative agreement here — mainly about the suggested adverse effects of this CIP — before we could continue considering this for merge.

polinavino:
Thank you very much for the feedback! It seems to me that the debate here is focused on whether this "get change sent to your account" idea is reasonable in general, rather than weather its technical implementation is acceptable. The proposal itself is decontextualized from the discussion of why it's needed. For this reason, we are actually going to combine this PR with the Urgency Signalling [PR](https://github.com/cardano-foundation/CIPs/pull/1265) . That PR already pre-supposes the fee change field, and it would be better you consider the full proposal as one CIP. Are there any objections to this?

rphair:
@polinavino that sounds like a good plan to me, as it would address the need to evaluate all inseparable propositions together as well as maybe provide some qualifications on how the fee-change mechanism is applied & help alleviate the concerns expressed yesterday.  Thanks for your response to that feedback and please feel at leisure to continue the community discussion here and/or discuss any plan to combine the PRs.

PR #1221, opened 2026-07-06
CIP-0118 | Enable Pay-Per-Use Message Passing with Sub-txs

fallen-icarus:
This PR looks to remove the utxo input requirement from sub-txs to enable using sub-txs for pay-per-use message passing.

Consider an oracle which wants to broadcast a price of 1 ADA/DJED which should be valid for 5 minutes. Today, they either:
- **Post it on-chain as a UTxO:** This ties updates to blockchain throughput and also makes it hard to monetize the service since referencing a UTxO can be done for free.
- **Pass around a redeemer with the price:** This enables faster updates but now payments and witness verification must happen during plutus execution which is extremely resource limited.

The sub-tx makes for the perfect abstraction of a pay-per-use message because it can tie use of a redeemer to a payment and rely on the ledger to efficiently check the witness and payment. The plutus runtime can be focused exclusively on the message.

But the way CIP-118 is currently designed prevents this. Currently, all sub-txs must be unique which means using a message is first-come-first-serve. *That completely kills this entire category of use cases.*

In order for this to work, sub-txs must be replayable which can be achieved by dropping the utxo input requirement for individual sub-txs. **But this has serious risks since it means a sub-tx could be replayed forever if no cancellation rules are included.** I don't believe this is any different than using an always succeeding plutus script so my stance is that this is not the responsibility of the ledger.

If pay-per-use message passing is indeed useful, then we need to drop this ledger requirement for sub-txs. We should not drop the utxo input requirement for the overall transaction because it is needed for other reasons!

It may not be too late to update the ledger rules before the Dijkstra hardfork, but the community needs to start considering it now.

---

([Rendered Update](https://github.com/fallen-icarus/CIPs/tree/reusable-sub-txs/CIP-0118)) > added sections:
* [Efficient pay-per-use messages](https://github.com/fallen-icarus/CIPs/tree/reusable-sub-txs/CIP-0118#new-use-cases)
* [9. Sub-transactions *may* have an empty input set](https://github.com/fallen-icarus/CIPs/tree/reusable-sub-txs/CIP-0118#changes-to-transaction-validity)
* [Replayable sub-transactions](https://github.com/fallen-icarus/CIPs/tree/reusable-sub-txs/CIP-0118#replayable-sub-transactions)

### DIFF
--- a/CIP-0118/README.md
+++ b/CIP-0118/README.md
@@ -121,6 +121,18 @@ purposes of encouraging users to interact with it. Implementing nested transacti
 will allow building transaction batches that subsidize
 DApp-using sub-transactions in this way.
 
+**Efficient pay-per-use messages.** Consider an oracle: it currently posts data on-chain as a UTxO
+for smart contracts to reference, but this ties oracle updates to blockchain throughput, and since
+referencing a UTxO is free, the oracle can't get paid a microfee for its service (with CIP-159). An
+alternative is to put the message in a redeemer to share off-chain for use by smart contracts. This
+works today, but how would the oracle trustlessly enforce payment per use of the message? It would
+have to enforce ledger rules using scarce plutus execution units. The ideal solution is to use a
+sub-tx to pass the message because it can enforce payments directly *without consuming plutus
+budget*. The oracle is one use-case but this can be generalized to any off-chain pay-per-use message
+service. Critically, **this requires replayability of sub-txs**. Otherwise, it is first come first
+serve for these messages. See [Replayable sub-transactions](#replayable-sub-transactions) for
+further discussion.
+
 ## Specification
 
 For the Agda specification prototype built on top of the current ledger spec,
@@ -184,6 +196,21 @@ quantities not for individual transactions, but for the entire batch, including 
 
 8. The total size of the top-level transaction (including all its sub-transactions) must be less than the `maxTxSize`.
 This constraint is necessary to ensure efficient network operation since batches will be transmitted wholesale across the Cardano network.
+
+9. Sub-transactions *may* have an empty input set. The rule that every transaction must consume at
+   least one UTxO entry (introduced in the Shelley design specification, [Section
+3.4.2][replay-protection], to give certificates replay protection) applies only to the overall (aka
+aggregate) transaction. A sub-transaction with no inputs is therefore *replayable*: an identical
+copy of it may appear in any number of batches until its cancellation conditions are met (e.g. its
+validity interval expires, or a reference input it names has been spent). *This is intentional*; see
+[Replayable sub-transactions](#replayable-sub-transactions).
+
+10. The outputs of every transaction in a batch get their output references from the hash of the
+    *overall transaction* that carried them. Sub-txs are applied in order of the tx body list and
+before the top-level transaction, and output indices are deterministically derived from that same
+order. **This means that the overall tx must still have at least one UTxO input!** It is not safe
+for the overall tx to be replayable too, because two inclusions would carry the same hash and cause
+UTxO naming collisions.
 
 ### Using older PlutusV1 - PlutusV3 scripts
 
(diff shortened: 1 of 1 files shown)


fallen-icarus:
Mass tagging people (sorry in advance :sweat_smile:).

@lehins @polinavino @willjgould @mpizenberg @nfrisby @ch1bo @colll78 @Quantumplation @GeorgeFlerovsky @WhatisRT @AndrewWestberg @berewt @coot @MicroProofs @michele-nuzzi

rphair (commented):
@fallen-icarus this is coming just in time to `Triage` concurrently with #1218 at the next CIP meeting (https://hackmd.io/@cip-editors/138) and in advance of a [forthcoming](https://github.com/cardano-foundation/CIPs/pull/1218#discussion_r3516342967) CIP for Tiered Pricing.
* There are likely some elements in common between all of these, although we don't have time at these next meetings for more than a "confirmation" review: leaving detailed review to GitHub.
* Still we will try to keep in mind the timeliness of this consideration in case it is in fact possible to include this with Dijkstra: we should air that question at tomorrow's meeting.

GeorgeFlerovsky:
Seems to me that replay of sub-txs is a pretty big risk... Punting it out of the ledger's responsibility for someone else (undetermined) isn't satisfying.

Also, I don't see how it's "just the same as an always-succeeds script." Scripts are re-useable, transactions are single-use (before this CIP).

berewt:
The risk of subTx being signed forever and valid forever makes me think that it's a very bad idea. It isn't. similar to an always true spending script, as the associated utxo would be consumed.

I would also suggest to move the discussion to a CPS first. It looks we are jumping on a specific solution here, and alternatives are barely considered. For example, while an "oracle" script purpose that triggers when you reference datum would take some UTXO budget, but it sounds to me a better alternative, as it's more versatile and doesn't introduce a potential vulnerability at the ledger level.

fallen-icarus:
> It isn't. similar to an always true spending script, as the associated utxo would be consumed.

If you create a new UTxO at that address, it can get drained too. That is similar to a sub-tx which withdrawals 5 ADA and doesn't expire; if you deposit more into that address, it can get drained.

It is also similar to message signing and off-chain intents which people use.

> I would also suggest to move the discussion to a CPS first.

If no resolution is found, I might. But I'm not convinced there isn't a simple solution here. Like what if we require a sub-tx without a UTxO to have an invalid-hereafter bound which fits within the ledger's `SafeHorizon` (I think it's 36 hours)? That would enable pay-per-use message passing without "forever sub-txs".

fallen-icarus:
Thinking about it more, this effectively requires a heavily restricted sub-tx to satisfy the concerns. The key use case is *paying* to use a message so balances should only be allowed to increase. If no UTxO input is present, then minting, certificates, voting, and withdrawing would all be disabled.

But that seems different enough to warrant a separate CIP or CPS like suggested. I'll close this for now. I rushed in hopes it would be simple enough to implement for the hardfork... :face_with_diagonal_mouth:

lehins:
> what if we require a sub-tx without a UTxO to have an invalid-hereafter bound

This doesn't solve anything. The replay issue is immediate. As soon as a transaction with sub-transaction with no inputs added to the mempool, anyone can flood the mempool with other transactions that include the same sub-transaction. This is exactly what is desired for your use case of oracles, but is the last thing we want for a normal sub-transaction. 

> If no UTxO input is present, then minting, certificates, voting, and withdrawing would all be disabled.

Yes, that is exactly what I was saying during the Ledger Working Group meeting yesterday.


I am glad that everyone is on the same page that this is too risky, I was worrying a bit yesterday that it would only be me that saw potential dangers in this suggestion.



@fallen-icarus With respect to the oracle use case. Could it be solved in this way (once CIP-159 is implemented):

1. Oracle mints say 10K number of tokens that have their symbol contain the price of an asset
2. All of those tokens are deposited into an account X
3. Account X is backed by a Plutus script that anyone can withdraw from one token at a time into a predefined account Y as long as some small amount of ada is also deposited into that same oracle's predefined account Y
4. Redeemer for such script could also include a public key of an oracle and a signature of a token hash, which would validate the authenticity that the token was indeed minted by the said oracle.
5. All those tokens from Accounts X and Y are burned by the oracle after 5 minutes have passed, while new ones are minted with updated price and deposited into X account. Repeat.

It would seem like this approach could provide access to a price of an asset as a payed service. Thoughts?

fallen-icarus:
The CIP-159 approach is bottlenecked by on-chain throughput; you have to submit a burn tx at the 5 minute mark and hope it gets included quickly. Stale quotes are a problem. Doing the message passing off-chain means there are no stale quotes anymore.

There may be other ways to do this with non-replayable sub-txs. I'm not sure it will be as efficient, but it might be the best approach given the security concerns mentioned here. I'll play around with the sub-txs when they become available and circle back to this idea if needed.

GeorgeFlerovsky:
> The CIP-159 approach is bottlenecked by on-chain throughput; you have to submit a burn tx at the 5 minute mark and hope it gets included quickly. Stale quotes are a problem. Doing the message passing off-chain means there are no stale quotes anymore.
> 
> There may be other ways to do this with non-replayable sub-txs. I'm not sure it will be as efficient, but it might be the best approach given the security concerns mentioned here. I'll play around with the sub-txs when they become available and circle back to this idea if needed.

Just chain your txs.

As you may recall from our full-featured-accounts CIP, if an account has sufficient liquidity to satisfy all txs currently trying to withdraw from it, then those txs have zero contention with each other (relative to that account). Thus, a transaction that spends a user's exclusively controlled utxo and withdraws a token from the oracle account will have zero contention in the typical case.

Therefore, the user is free to chain his follow-up tx (i.e. the one that relies on the oracle data) immediately afterward.

fallen-icarus:
Tx chaining is not guaranteed atomic which limits its utility. Using sub-txs would be guaranteed atomic. And chaining requires coordination which has limited practicality on a decentralized scale. Batcher-based dApps may be able to easily plan ahead and create tx chains, but p2p protocols cannot.

PR #1223, opened 2026-07-08
CIP-0116 | fix incorrect required property in Conway schema

Ryun1:
For SPO `Voter` objects, `Credential` was required
`Credential`s aren't used to represent SPOs
instead `pubkey_hash` is correct,
matching the exiting propertises.


rphair (approved):
Looks reasonable & would be happy to merge once @perturbing confirms... though I'll leave the merge to you @Ryun1 in case you want to wait for any other governance metadata advocates to confirm: @Quantumplation @thenic95 @Hornan7 @gitmachtl

perturbing (approved):

PR #1224, opened 2026-07-09
CIP-0179 | Revision 5: corrections and tightenings

mpizenberg:
- Generalize governance linkage to any action type (was Info Actions only); mechanism-B vote binding available only while the action is votable.
- Carry the link inside the CIP-108 body as a namespaced cip179 object with required @context terms; shorten kind to "survey-link".
- Add a mandatory rating require_all flag; when set, a present rating answer must rate every option.
- Tighten sealed-response serialization: canonical-CBOR plaintext padded to padding_size, raw de-armored ciphertext (not AGE/PEM armored text).
- Bump spec_version to 5 across README, schemas, examples, and test vectors.


mpizenberg:
## Revision 5

This bundles the spec refinements I've hit while building Tessera and testing Civitas against v4 into one release (`spec_version` 4 → 5). These are corrections and tightenings, not new features. The bigger items (multi-language, message-signing / fee-sponsored responses, delegation) are deliberately deferred to v6+ so this stays reviewable.

**Breaking is fine now:** there are no production deployments widely used yet (both tools are in development; the Preview/Mainnet surveys are experimental), so we can make the one breaking change cleanly instead of carrying compatibility cruft. Civitas and Tessera move to v5 together.

## What changed

**1. Linkage generalized to any action type** (was Info Actions only). Any governance action can now link a survey. Since a non-Info action can resolve (ratify/enact/expire) before its expiry epoch, the mechanism-B vote binding is available only while the action is still votable; afterwards a linked response proves its credential via mechanism A, exactly like a standalone one. The `end_epoch` rule generalizes to "the action's expiry epoch," and a survey may be linked by more than one action.

**2. Link moved into the CIP-108 `body`.** The old bare top-level anchor object sat outside `body`, so it wasn't covered by the author witness and a JSON-LD processor could silently drop it. The link now lives at `body.cip179` with required `@context` terms (new `#### CIP-179 @context terms` subsection), so it survives canonicalization and stays witnessed. `kind` shortens to `"survey-link"` (the `cip179` key already namespaces it).

**3. Rating `require_all`** — the only breaking wire change. `rating_question` gains a mandatory `bool` before the optional `required`: `true` forces every option to be rated (the "answer every row" toggle for a rating matrix, which we couldn't express before), `false` keeps today's subset behavior. It's mandatory and fixed-position rather than a second `? flag` so the trailing `required` stays unambiguous to decode. Answer encoding is unchanged.

**4. Sealed serialization tightened** — clarification, no wire-shape change. Two underspecified points that broke cross-tool decoding of Civitas's sealed votes: the tlock plaintext MUST be canonical CBOR of `[+ answer_item]` (not JSON), right-padded to `padding_size`; the ciphertext MUST be raw de-armored bytes (not AGE/PEM "armored" text, which is ~40% larger). New `### Sealed responses` section spells out encode/decode (read the first CBOR item; don't strip trailing `0x00`).

## Diff scope

- `README.md` — the four changes above + `spec_version` 4→5 (CDDL header, key comments, definition table, CBOR examples, Version 5 changelog).
- `schemas/` — `common.schema.json` (rating_question + two spec_version pins), plus the linkage-anchor and presentation schemas (`const` / `kind`).
- `examples/` — 17 payloads bumped to v5; the two off-chain docs updated (anchor rewritten to the `survey-link` object).
- `test-vector.md` — linkage vector regeneralized; version note.

No fixture uses a rating question, so the new mandatory `require_all` invalidates none of them.

Let me know what you think. I’ve mostly bundled things we have already discussed, and I think agreed on. But if you might not have had enough time in earlier discussions, here is you opportunity to see the concrete changes @Cerkoryn @Thomas-nada

rphair (commented):
@mpizenberg this is a brilliant continuation to https://github.com/cardano-foundation/CIPs/pull/1107#issuecomment-4779844107 (as expected) and I would be happy to merge this as soon as @Thomas-nada @Cerkoryn are in agreement over their own implementations & expectations for the specification.

We also are hoping this will make the rounds among governance tooling developers & advocates so I would also hope that the review baton might be passed around a little bit more than last time: cc @thenic95 @gitmachtl @Hornan7 @gufmar @ptrdsh

We'll likely confirm the update at the next meeting but please if consensus is reached among the 3 co-authors about merging this as-is, please state that here & we'll begin editorial review and likely change from `Triage` to `Last Check` even if before the next CIP meeting: https://hackmd.io/@cip-editors/139

Also @mpizenberg I'm retitling this PR so that the community might find it less intimidating (and for brevity's sake) — without dependency on the technical terms (which might also change in the course of review) according to your own words above.

Thomas-nada:
I have been too busy to properly contribute lately, but I have been following the discussion and suggestions and have no objections to this.

Cerkoryn:
No objections here either.

rphair (approved):
Then there is nobody left to say they want to update this differently 😅 unless coming from the "outside" and we've already acknowledged at the CIP meeting that the only implementations at this time are under direct supervision by the co-authors.

So it seems @Ryun1 @perturbing we can safely push this through on the heels of CIP-0179 itself & therefore promoting to `Last Check` for meeting https://hackmd.io/@cip-editors/139 — we can belay that if there is a review suggestion that would require modifying this update (or perhaps the original CIP).

mpizenberg:
I’ve implemented the V5 upgrade in Tessera. Here it is for anyone wanting to play with it: https://tessera-preview.matthieu-pizenberg.workers.dev/

@Thomas-nada The new code organisation of the Tessera app has split the CIP-179 core from the rest of the app, into a separate package that is now also published completely independently. CF docs here: https://github.com/mpizenberg/Tessera/tree/main/packages/cip179

And the package is here: https://www.npmjs.com/package/cip-179

mpizenberg:
## Follow-up commit on Revision 5

A follow-up on top of d0820bd — no new features. The `Owner` → `Keyholder` rename is the one we'd [already discussed](https://github.com/cardano-foundation/CIPs/pull/1107#issuecomment-4876129343); everything else is correctness and tightening.

These changes had slipped through my review of the changes. I forgot to port them from the Tessera project.

- **Role 4 `Owner` → `Keyholder`** (prose only; tag and encodings unchanged) — the
  rename we already discussed; removes the clash with the definition's `owner` (the
  cancellation authority).
- **`bool` → `flag` (correctness).** Cardano metadata has no boolean type, so
  `required`/`require_all` can't be a literal `bool`. Now defined as
  `flag = uint .le 1` (`1` = true, `0`/omitted = false), with the JSON Schema updated
  to match. No practical wire change (tools already emit these as uints).
- **Mechanism-B binding is now purely additive — the one behavioral change.** A
  `voting_procedures` entry that doesn't match (credential, one of the linked actions,
  role) is simply not a binding: it no longer invalidates the response, which falls
  back to mechanism A. Previously a present-but-failing binding invalidated it — too
  strict now that a survey can be linked by several actions.
- Editorial: reconciled `linked_action_id` wording; normalized non-table Markdown
  formatting (most of the diff, no content change).

Files touched: `README.md`, `schemas/common.schema.json`, `test-vector.md`.

rphair (approved):
@mpizenberg this all looks excellent to me, though it might be even better (and generally is, when submitting updates) if your Markdown editor can be made agnostic about the number of spaces to indent bulleted lists, using `_` instead of `*` for italics, etc.

As of https://github.com/cardano-foundation/CIPs/pull/1224/changes/2848f21b2f3bcecce80b73472d41118ab2a90acc there are a lot of code changes that are not content changes & so a more agnostic editor would help readers, reviewers & implementors target the functional changes without the superfluously modified lines.

However since you appear to be practically maintaining this, I'm happy to leave this editing question to you... and if the Markdown conforms to anyone's editor it should probably be _your_ editor.

mpizenberg:
Ah sorry, I had hope it was not too disturbing. I did format the readme
markdown. Then realized the formatted tables would be too much churn,
especially if new lines were added, so I un-formatted the tables but left
the rest.

Say the word if you'd prefer I keep the original format completely and I'll
revert it.

rphair:
It's OK with me as long as it's OK with @Cerkoryn @Thomas-nada = those whose editors (and attention on apparent changes) it would be churning with 😅

mpizenberg:
Tessera is _mostly_ done for the CIP v5: https://tessera-preview.matthieu-pizenberg.workers.dev/. It’s still at a personal deployment while in testing. Now what’s left is mostly getting this out to people who might be interested in using it, and helping integration of the CIP in other platforms.

Meanwhile, I need to find my next prioritizations, so I’d welcome your input and your suggestions on the following directions.

1. Trustless verification from auditors
2. Support for the "State of the developer ecosystem" survey
3. Innovative features exploration
4. Something else?

## Trustless verification from auditors:

Currently, the backend is based on API provider (Koios in my case but you could do with Blockfrost). For trustless verification by auditors, one needs to be able to recompute past registrations, stake distributions, voting powers, directly from onchain.

Currently, the backend produces snapshots in a standardized artifact format, described here: https://github.com/mpizenberg/Tessera/blob/main/backend/ARCHITECTURE.md

A trustless backend could produce exactly the same snapshot artifacts, so both approaches only trade convenience for trust of an API provider. I wrote a bit some thoughts about potential approaches for that in here: https://github.com/mpizenberg/Tessera/blob/main/backend/RESEARCH.md

Finally, the actual jackpot feature would be cheap trustless auditing based on merkle trees of ledger state snapshots. An exploration of my collegue Paolo suggests this is doable! https://gist.github.com/paolino/7cb040b176d7b0a339134147c2aef90c

## Support for the "State of the developer ecosystem" survey

CIP-179 mostly support all features already needed for it. Though after a call with Matthias, it seems there 2 more features that would be needed for full support of that survey needs:
- Multi-language support. I put my thoughts here: https://gist.github.com/mpizenberg/accc2d4e27cb8d33c8e21b504c61197c
- Adding ability to have an optional free-text answer to single-choice or multi-choice questions.
- Questions dependency graphs: some answers (un)lock other questions

## Innovative features exploration

At least 2 new features would be interesting to explore:

- A third method via message signing for vote authentication. That would also enable vote aggregation for efficient 3rd party sponsoring of votes. I put some thoughts here: https://gist.github.com/mpizenberg/87dd4f27b0bfbcccc186f383909210be

- A new liquid topic-scoped delegation model. That would enable some really cool exploration about improved delegation model, which could serve as a testbed for future improvements of Cardano governance itself. I put some thoughts here: https://gist.github.com/mpizenberg/d71c6553f14b628359d95ea71a8335d7

- Exploration towards private voting support with ZKP. No thoughts on this yet.

## Anything else?

Thoughts? Also, of course I’m available to help integrate the current V5, and future iterations, into other platforms.

Ryun1 (approved):

katomm:
Feedback from implementing the governance linkage side of CIP-179 in DRepTalk (currently on https://preprod.dreptalk.com): the linkage rule turns on a term whose mapping to an actual epoch number the document leaves open, and the two readings available to a new implementor differ by one.

**The rule**

> The survey `end_epoch` MUST equal the action's expiry epoch.

and in the implementation guide:

> check that the survey's `end_epoch` equals the action's expiry epoch

The surrounding text does give "expiry epoch" some shape: it is independent of early resolution, since an action that ratifies or expires sooner leaves the survey running, and `end_epoch` is itself an inclusive cutoff. What is missing is the mapping to a concrete epoch number, and that is where the two readings part:

1. the epoch value an implementor reads out of an indexer under the name `expiration`, for instance Koios `proposal_list.expiration` or db-sync `gov_action_proposal.expiration`
2. the last epoch of the action's scheduled voting window, which is one lower

The reference implementation documents the intended convention as 2. `cip-179@0.4.0`, `src/domain/records.ts:220`, on `UnresolvedGovAction`:

```
/** The action's expiry epoch (`expiration - 1`); a survey it could link ends here. */
readonly endEpoch: number;
```

`GovLink.endEpoch` (`records.ts:198`) is documented as the same quantity, and it is the one compared against the survey in `src/domain/survey.ts:249`. The package carries the convention as documentation only, the arithmetic itself happens in whatever indexer feeds it.

Reading 1 is the one a new implementor is likely to arrive at, because `expiration` is the field name the common indexers expose and the CIP's phrase points straight at it. In my own check of mainnet actions through Koios on 2026-06-22, every expired action I looked at had `expired_epoch == expiration`, which suggests `expiration` is the epoch in which the decision runs rather than the last epoch of the voting window. I am offering that as an indication, not as a statement about ledger internals.

**Why this one is worth pinning**

The failure mode is quiet. The epoch comparison in `survey.ts:249` drops a mismatched link without a rejection reason, so the survey stays standalone, and the CIP asks no tool to surface the mismatch anywhere. A survey creator or an action author looking at the result cannot tell a rejected link from a link that was never made. The recovery is not cheap either: the survey's `end_epoch` is fixed on chain at creation and has to be chosen before the linking action exists, so an off-by-one means republishing the survey, the action, or both.

**Suggested change**

Two parts, since the semantic definition and the field mapping are separate claims:

> The survey `end_epoch` MUST equal the action's expiry epoch, that is, the last epoch of the action's scheduled voting window, fixed when the action is proposed and unaffected by an earlier ratification or expiry.

and, as a note:

> Implementors should check this against their data source rather than assume it. Indexers commonly expose an `expiry` value one epoch higher, the epoch in which the ratify-or-expire decision runs, for instance Koios `proposal_list.expiration` and db-sync `gov_action_proposal.expiration`.

Since the mismatch is otherwise invisible, one addition next to "If linkage validation fails, tooling MUST NOT attach the survey":

> Tooling SHOULD make a rejected link and the reason for the rejection visible rather than presenting the survey as unlinked.

Happy to open a PR with this if it is useful, and to correct the field mapping above if it is off for a source I have not checked.

rphair:
@katomm that would indeed be welcome as a pull request.  I'm sure @mpizenberg @Cerkoryn @Thomas-nada could also discuss the details in a live issue but this repository has always gotten ***much*** better feedback on PR's than issues: something about seeing specific changes draws them in & keeps them engaged.  Thanks for your attention & looking forward to seeing how this feedback develops.

mpizenberg:
My understanding of the confusion here is that cip-179 mentions an action expiry epoch without enough precision. The correct interpretation is the one that makes sense. We stop voting for the survey at the same moment we stop voting for the gov action.

Part of the confusion comes from the fact that Koios reports the epoch at the start of which the action expires. While the CIP always mentions with `end_epoch` the end of an epoch. Leading to an off-by-one difference between the raw koios and this cip number.

IMO, stating that the end_epoch must equal the actions expiry epoch - 1 is wrong as it convey in words that the action and the survey do not end at the same time, while in reality, this is just start/end boundary interpretation difference.

I personally don’t think additional words like "the last epoch of the action’s ..." will save us. There will always be people/agents that will just look at the name of API X and think that "expiration" is a good name and just use that.

I agree with the note that link rejection would be good to display instead of just saying that a survey is unlinked.

In sum:
- ok for a temporary fix by adding some more words to convey the different boundary interpretation, because this is non-breaking for rev-5
- but we should change end_epoch to expiry_epoch and adopt the same meaning that other common tools for the next rev-6 of the CIP.
- ok to add a clause for tooling authors to suggest making rejection reason visible.

rphair:
@mpizenberg @Cerkoryn @Thomas-nada the Cardano Ambassador call just brought up (by @KtorZ) _event registration_ as a potential use case for CIP-0179 🤔

PR #1211, opened 2026-06-16
CPS-0033? | DRep Voting Power Concentration

maureenblack:
Stake-weighted voting models have a structural tendency to concentrate power over time, and the standard mitigations in other ecosystems rely on identity-based mechanisms that Cardano does not use and may never adopt. This CPS opens an investigation into whether Cardano's unique governance structure offers novel mechanisms for managing DRep voting power concentration without identity.
The problem is already being observed in practice: the Nakamoto coefficient for DRep voting power has been declining, and community concern over concentration is consistently surfaced in forum polls, governance roundtables, and the Beyond Minimum Viable Governance project's community research (2025–2026).
This CPS is connected to but distinct from CPS-020 (DRep incentives). The two efforts must be coordinated so that incentive design does not work against concentration mitigation, or vice versa.

---

([latest document rendered from branch](https://github.com/maureenblack/CIPs/blob/cps-drep-power-concentration/CPS-0033/README.md))

PR #1225, opened 2026-07-09
CIP-???? | DRep Pledge

Cerkoryn:
This CIP is a simplification of the earlier CIP draft, [Variable Deposits as Pledge](https://github.com/cardano-foundation/CIPs/pull/1204) that was closed.  The intent of this CIP is to simply add DRep pledge as a ledger field in a way that mirrors pool pledge as much as reasonably possible.

Adding this CIP and then using it for anything incentives-related will likely require two hard forks.  The first to create the mechanism for DReps to add pledge and then allow for explorers, indexers, and governance tools time to catch up.  The second hard fork would be for any use case that actually uses pledge, which is left out of scope for this CIP.

One lingering question that I left out of the CIP is that pool pledge is enforced via the Mark, Set, Go snapshots, which doesn't exist for DReps.  It may be necessary to add some sort of bonding period where DReps have to meet their pledge for X amount of time before anything takes effect.  Otherwise DReps can simply add pledge, vote, and then remove it to game the system.  It's possible that such a bonding period could be defined in this CIP, but IMHO it is likely better defined in the second part.  In any case, I think it is worthy of discussion.

([rendered](https://github.com/Cerkoryn/CIPs/tree/DRep-Pledge/CIP-XXXX))

### DIFF
--- a/CIP-XXXX/README.md
+++ b/CIP-XXXX/README.md
@@ -56,7 +56,7 @@ DRep registration certificates MUST include `drepPledge`.
 In CDDL-like form:
 
 ```cddl
-reg_drep_cert = (16, drep_credential, coin, anchor / nil, drep_pledge : coin)
+reg_drep_cert = (16, drep_credential, coin, anchor / nil, coin / nil)
 ```
 
 where:
--- a/CIP-XXXX/README.md
+++ b/CIP-XXXX/README.md
@@ -1,0 +1,231 @@
+---
+CIP: ????
+Title: DRep Pledge
+Category: Ledger
+Status: Proposed
+Authors:
+    - Ryan Wiley <cerkoryn@gmail.com>
+Implementors: []
+Discussions: []
+Created: 2026-07-09
+License: CC-BY-4.0
+---
+
+## Abstract
+
+Cardano's on-chain governance model allows Ada holders to delegate voting power to registered DReps. However, the ledger currently has no DRep pledge field. A DRep may register, publish metadata, receive delegated voting power, and vote on governance actions, but there is no ledger-native value by which that DRep can declare an amount of Ada as personal stake in the role.
+
+This CIP adds `drepPledge` as a declared `Coin` field to DRep registration and update certificates. The value is stored in DRep ledger state and exposed to ledger queries. It is analogous to stake pool declared pledge in that it is a declared amount recorded by the ledger, but this CIP does not enforce the declaration and defers it for use in any voting-power, reward, registration, activity, or compensation calculation.
+
+The purpose of this CIP is to introduce the mechanism for DRep pledge first. Future CIPs may use DRep pledge for voting-power limits, compensation eligibility, Sybil resistance, reputation signaling, or other incentive mechanisms. Those future mechanisms should not activate before DReps have had adequate time to declare pledge and for tooling to add support once this capability exists on-chain.
+
+## Motivation: Why is this CIP necessary?
+
+### DReps have delegated voting power but no pledge field
+
+[CIP-1694](../CIP-1694/README.md) introduced DReps as one of Cardano's governance bodies. Any Ada holder may register as a DRep, and other Ada holders may delegate voting rights to that DRep. DRep voting power is based on delegated Lovelace.
+
+This establishes the delegation side of the system, but it does not establish an on-chain pledge field for DReps. A DRep may have large delegated voting power without any ledger-native way to declare personal economic commitment. Today, tools can show metadata, payment addresses, voting history, and delegated stake, but not a ledger-defined DRep pledge.
+
+### Metadata is useful but insufficient for future ledger rules
+
+[CIP-0119](../CIP-0119/README.md) defines useful DRep metadata. That metadata can describe a DRep's identity, motivations, qualifications, payment address, and other public information. It is intentionally off-chain and flexible.
+
+Pledge is different. If future ledger rules are expected to use pledge as an input, the value should be represented directly in ledger state. A metadata-only pledge field would be useful for social signaling, but it would not provide a deterministic, consensus-visible value that later ledger rules can reference without depending on off-chain metadata interpretation.
+
+### Pledge should be introduced before incentives
+
+If a future hard fork were to introduce pledge incentives and the pledge field at the same time, DReps could face a race to update their registrations before pledge-dependent rules affected voting, rewards, or eligibility. That creates unnecessary operational and security risk.
+
+This CIP avoids that problem by adding only the declaration field. DReps can begin declaring pledge as part of normal registration and update workflows, while the community continues researching whether and how pledge should later affect governance incentives or constraints.
+
+## Specification
+
+### Terminology
+
+- `drepPledge`: the declared amount of Ada, represented as a `Coin`, associated with a registered DRep.
+- `drep_pledge`: the CDDL field name corresponding to `drepPledge`.
+- "Declared DRep pledge": DRep pledge recorded in ledger state.
+
+### DRep registration certificates
+
+DRep registration certificates MUST include `drepPledge`.
+
+In CDDL-like form:
+
+```cddl
+reg_drep_cert = (16, drep_credential, coin, anchor / nil, drep_pledge : coin)
+```
+
+where:
+
+- `drep_credential` is the DRep credential.
+- `coin` remains the DRep registration deposit supplied by the certificate.
+- `anchor / nil` remains the optional DRep metadata anchor.
+- `drep_pledge` is the declared DRep pledge.
+
+When a DRep is registered, the ledger MUST store `drepPledge` in the DRep's ledger state.
+
+### DRep update certificates
+
+DRep update certificates MUST include `drepPledge`.
+
+In CDDL-like form:
+
+```cddl
+update_drep_cert = (18, drep_credential, anchor / nil, drep_pledge : coin)
+```
+
+where:
+
+- `drep_credential` is the DRep credential.
+- `anchor / nil` remains the optional updated DRep metadata anchor.
+- `drep_pledge` is the updated declared DRep pledge.
+
+When a DRep update certificate is accepted, the ledger MUST replace the previous `drepPledge` value for that DRep with the value supplied in the update certificate.
+
+An update that changes only the DRep metadata anchor still MUST include `drepPledge`. If the DRep does not intend to change pledge, the update certificate should resubmit the current pledge value.
+
+### DRep ledger state
+
+DRep ledger state MUST include `drepPledge`.
+
+Conceptually:
+
+```haskell
+data DRepState = DRepState
+  { drepExpiry  :: EpochNo
+  , drepAnchor  :: StrictMaybe Anchor
+  , drepDeposit :: CompactForm Coin
+  , drepPledge  :: Coin
+  , drepDelegs  :: Set (Credential Staking)
+  }
+```
+
+The exact implementation type MAY use a compact representation if ledger maintainers prefer consistency with other stored `Coin` values.
+
+### Accepted pledge values
+
+`drepPledge` MAY be zero.
+
+`drepPledge` MAY be any non-negative `Coin` value that can be represented by the ledger's `Coin` type.
+
+This CIP does not introduce a minimum DRep pledge, maximum DRep pledge, pledge ratio, pledge leverage parameter, or pledge-specific protocol parameter.
+
+### Authorization
+
+DRep registration, update, and deregistration authorization remains unchanged from CIP-1694.
+
+The DRep credential or script that authorizes the DRep registration or update also authorizes the declared `drepPledge` value in that certificate. This CIP does not introduce any additional witness, payment credential, stake credential, owner credential, or pledge-specific authorization requirement.
+
+### Hard-fork activation and migration
+
+This CIP changes ledger certificate serialization and DRep ledger state. It therefore requires a hard fork into a new ledger era.
+
+At the hard-fork boundary:
+
+- all already-registered DReps MUST be migrated with `drepPledge = 0`;
+- the existing DRep deposit values MUST be preserved;
+- the existing DRep metadata anchors MUST be preserved;
+- the existing DRep expiry/activity state MUST be preserved;
+- the existing DRep delegation state MUST be preserved.
+
+After activation, new-era DRep registration and update certificates MUST include `drepPledge`. Old-shape post-activation DRep registration and update certificates that omit `drepPledge` MUST be rejected as invalid for the new era.
+
+Historical pre-activation transactions and certificates remain valid historical ledger data under their original era rules.
+
+### Ledger queries and tooling
+
+Ledger state queries SHOULD expose `drepPledge` for registered DReps.
+
+Wallets, CLIs, explorers, indexers, governance tools, and DRep directories SHOULD display this value as declared DRep pledge.
+
+## Rationale: How does this CIP achieve its goals?
+
+### A ledger field is the smallest useful primitive
+
+This proposal intentionally does not define a DRep compensation scheme, voting-power cap, pledge leverage mechanism, or proof-of-ownership rule. Those mechanisms require separate design work and likely separate community debate.
+
+However, many such mechanisms need the same primitive: a ledger-native DRep pledge value. Adding `drepPledge` first allows the ecosystem to build tooling, index the value, display it to delegators, and gather real-world usage data before pledge is used in higher-stakes calculations.
+
+### DRep pledge mirrors stake pool pledge as a declaration
+
+Stake pool pledge is already familiar to Cardano users as a declared value in pool parameters. This CIP uses the same general idea for DReps, but without the current stake pool pledge satisfaction check or any reward effect (for now).
+
+That limitation is deliberate. DRep pledge should not immediately change governance power or incentives before the community has agreed on the right incentive model.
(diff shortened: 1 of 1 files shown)


rphair (commented):
@Cerkoryn good to see you're continuing the applicable ideas from https://github.com/cardano-foundation/CIPs/pull/1204.  This very well suits concurrent discussion, especially:
* https://github.com/cardano-foundation/CIPs/pull/1211

... which I think could also be added as a `Solution To:` in this CIP header after the CPS is merged (with also the matching `Proposed Solutions:` line added to that CPS in this branch)... assuming it's merged before yours.

Please tag a governance advocate or two to start off review for this: I've personally been tagging them too often recently over the influx of governance-related proposals over the last few weeks. 😅

@lehins I think it also wouldn't be too early for you to give us the Ledger point of view, especially over the possibility of a "latency" period — described in the OP — after which the pledge becomes effective.

In any case will be happy to kick this off in `Triage` at the next CIP meeting: https://hackmd.io/@cip-editors/139

Cerkoryn:
> Please tag a governance advocate or two to start off review for this...

Paging fellow governance advocates @adamrusch @Crypto2099 @Hornan7 @thenic95 @maureenblack @Seomon4u @PedroLucasnet 😅

maureenblack:
Hey Ryan, thanks for the tag and for putting this together. You clearly put a lot of work into it. The spec is clean, the writing is clear and I've read it twice because I was convinced I must have missed the bit where the problem shows up.
...I still can't find it. :(
A few genuine questions :
What problem is this actually solving? The motivation tells me DReps don't have a pledge field. Right... but that's just how the system works today? No? Saying "there is no pledge field" is like saying "my kitchen doesn't have a chandelier." Okay... but why does it need one?
If the pledge isn't enforced, what's stopping me from pledging 10 million ADA while my wallet is surviving on 12 ADA? The ledger will faithfully record my imaginary fortune and everyone gets to admire it. Doesn't that create more opportunities for misleading people instead of fewer?
Isn't the strongest flex already on-chain? Voting participation rate, rationale quality, DRep expiry status. And If I want to show _I have skin in the game_,(which isn't necessary imo), I can delegate my own ADA to myself. That's actual stake, verifiable. An unenforced declared number doesn't measure commitment, it measures willingness to type a number.  Why are we hard forking the ledger to store the blockchain equivalent of "trust me, bro"?

Also... aren't DReps literally Delegated Representatives? Their whole job is to represent other people's stake. If I wanted someone to represent *their own* stake, I'd just delegate to... myself. 🤷🏽‍♀️ Shouldn't we be selecting DReps because they're thoughtful, active, and understand governance rather than because they're rich?

The "let's add it now and figure out what it's does later" argument is where you lose me. Future CIPs can absolutely introduce this field if they genuinely need it. They can even include a grace period. Where's this dramatic race we're trying to avoid? 
Then there's the cost. New ledger era. Certificate changes. Wallets, explorers, indexers, SDKs, governance tools... everybody gets homework. For... a number that currently does absolutely nothing. That's a lot of engineering effort.

Don't get me wrong, I appreciate the amount of work that went into this proposal. My issue isn't the implementation. It's that I'm struggling to see the motivation. Right now it feels less like "here's a problem and here's the solution" and more like "here's a really nice solution... now let's see if we can find it a problem." Happy to be convinced otherwise. I'm genuinely curious what governance issue this solves that existing delegation or metadata can't already address.

rphair:
(my last comment removed since I'd read the last comment as if it were for the CPS review thread)

maureenblack:
haha okay  thanks.. i was super lost.

Cerkoryn:
Thanks for the review @maureenblack 

> What problem is this actually solving? The motivation tells me DReps don't have a pledge field. Right... but that's just how the system works today? No? Saying "there is no pledge field" is like saying "my kitchen doesn't have a chandelier." Okay... but why does it need one?

This is an excellent point and something that I tried not to be overly prescriptive about (perhaps too hard 😅 ).  Some examples that have come to mind from discussions I've had involve ideas such as...

-Using pledge to limit voting power based on a multiple of that pledge (hard cap)
-Using pledge as an incentive mechanism to reward DReps who have more skin in the game (soft cap)
-Using pledge as an incentive mechanism to reward \*delegators\* to move to DReps who have more skin in the game (soft cap)
-Perhaps a combination of any of the above?

Regarding voting power concentration mentioned in [CPS-33](https://github.com/maureenblack/CIPs/blob/cps-drep-power-concentration/CPS-0033/README.md), my hypothesis is that this will help due to the nature of some of the largest DReps.  Wallets gain delegation simply by steering users toward a default option, not because they have a lot of their own ADA.  Influencers are in a similar boat due to the amount of views/impressions they receive that drive delegations, as opposed to their own ADA.  Custodians and CEXes also have a complex web of legal/technical hurdles they have to overcome if they want to use customer ADA as pledge.

I'm personally working on a couple different research streams in this direction as I mentioned in your [CPS discussion](https://github.com/cardano-foundation/CIPs/pull/1211#issuecomment-4905032013).  I'm hoping to find one or more feasible solutions that receive strong community support that would fulfill the second part of this CIP (i.e., actually using pledge for something).  So far I've received a lot of positive feedback from this direction.

> And If I want to show I have skin in the game,(which isn't necessary imo), I can delegate my own ADA to myself. That's actual stake, verifiable.

This is in essence what pledge is.  For SPOs it represents their own ADA that they are delegating to their own pool.  However, if an SPO's own ADA dips below what they have declared as their pledge, then their pool loses eligibility for rewards.  This is similar to what this CIP is trying to do with DReps, but in order to implement securely it it needs to be separated into two hard forks.  One to allow DReps to start pledging and tools to track it, and then the second to actually incentivize/enforce it.

maureenblack:
Thanks for the detailed reply! 😊
I think where we're still talking past each other is the distinction between a problem and a possible future solution. The examples you gave (hard caps, soft caps, incentive mechanisms, etc.) are all interesting ideas, but they're mechanisms that might use a pledge field. They're not, by themselves, evidence that the absence of a pledge field is currently causing harm. Take the wallet example. If a wallet becomes a very large DRep because thousands of people voluntarily delegate to it, is the problem that the wallet has little personal ADA, or is the problem that too many people chose the same DRep? Those feel like very different problems with potentially very different solutions.
The same goes for influencers. If people choose to delegate to someone because they trust them (or because they're famous 😅), is the lack of a pledge field really what's driving concentration?
I guess what I'm looking for is a statement along the lines of: "Today, X undesirable thing happens because DReps cannot declare a pledge."
Once that problem is clearly established, then we can debate whether pledge is the best solution.
The other thing I'm still wrestling with is your comparison to SPOs. Pool pledge makes sense because SPOs are operators securing the network with infrastructure and the pledge is part of an existing economic model.
DReps are different. They're representatives. Their legitimacy comes from the people who delegate to them, not from the amount of ADA they personally own. So I worry that we're importing an economic concept from stake pools into governance without first asking whether governance actually benefits from it. If the end result is that wealth becomes a prerequisite for being seen as a "good" DRep, I'm not convinced that's a direction we should encourage.
I'm definitely interested in the research, though. I just think I'd be much more comfortable seeing the research identify the governance problem first, and then introducing the ledger primitive if it's actually needed, rather than introducing the primitive and hoping the right use case follows.

rphair:
@Cerkoryn in spite of this PR's validation failing after https://github.com/cardano-foundation/CIPs/pull/1225/changes/e9fc91a516f55fd4caab1e417a59da42339e9628, I believe it's appropriate to keep the header item declaring this as a solution to `CPS-0033` — and therefore I won't suggest you change anything even though our validator reports it as a nonexistent document.
* I do think it's likely https://github.com/cardano-foundation/CIPs/pull/1211 will be merged before this PR and so in that case the current validation failure would be resolved before it would prevent this from being merged.

@Ryun1 I think it's reasonable that the validator continues to operate this way & I don't see an advantage in providing special circumstances that will allow the document with a not-yet-valid link to validate.  In fact if the destination document isn't merged first, it would lead to a broken link on the repo.
* If you can think of any recommendations for authors that would avoid this uncertainty, I would be happy to talk about it over in https://github.com/cardano-foundation/CIPs/issues/1216.

TL;DR in this case as long as CPS-0033 remains unmerged, we will have a validation failure here... which may be better than leaving that addition until the end & maybe forgetting it (in the worst case, requiring an extra update PR later).

Cerkoryn:
@maureenblack 
> I think where we're still talking past each other is the distinction between a problem and a possible future solution.

Perhaps that is part of it.  I have this written as a CIP, not a CPS, so I was writing it to be (part of) a solution.

> I guess what I'm looking for is a statement along the lines of: "Today, X undesirable thing happens because DReps cannot declare a pledge."

In theory I could write a separate CPS about this, but it would be very hard to quantify.  We could envision it qualitatively in a contrived solution like,

"Today, Vitalik Buterin is the number one DRep on Cardano with billions of ADA in voting power, despite owning zero ADA himself.  He has garnered a lot of delegation through his popularity and uses that influence to downvote any proposal that would give Cardano a competitive edge over Ethereum."

The example is of course contrived, but it shows how a conflict of interest could cause powerful DReps to work AGAINST the ecosystem because they have nothing at stake within it.  This is contrary to the design of Proof-of-Stake itself.

> So I worry that we're importing an economic concept from stake pools into governance without first asking whether governance actually benefits from it. If the end result is that wealth becomes a prerequisite for being seen as a "good" DRep, I'm not convinced that's a direction we should encourage.

I think DRep compensation would necessarily introduce economic concepts in any implementation, which is something this CIP aims to precede (though not necessarily a requirement).  However I empathize with your concern that if implemented carelessly it could cause gatekeeping via wealth.  I think that is primarily a matter of tuning though.  Consider:

- DRep A with 0 ADA owned and 1 Billion ADA delegated
- DRep B with 100,000 ADA owned and 1 Billion ADA delegated
- DRep C with 1 Million ADA owned and 1 Billion ADA delegated
- DRep D with 10 Million ADA owned and 1 Billion ADA delegated

I would hypothesize that most Cardano stakeholders would have little problem with DRep D (100x leverage) but discomforted by the misplaced interests of DRep A (infinite leverage).  The question becomes a matter of deciding what level of leverage is \*sufficient\* and then any DRep who meets that minimum threshold relative to their voting power wouldn't have to worry.

jennycita:
Even if the CIP says incentives are out of scope, adding "DRep pledge" as a formal field creates the political foundation for later saying: "well, now that pledge exists, we should use it."

The part that bothers me most is this:

"in a way that mirrors pool pledge as much as reasonably possible"

This is politically dangerous. It imports stake pool logic into representation. 

When addressing a problem involves introducing a new one, the tradeoffs need to be worth it.

Once a pledge field exists, it becomes very hard to prevent people from treating it as a legitimacy score. Especially when we have advocates out there intorucing GAs trying to make SPO pledge matter.

This is very concerning to me.

CIP-XXXX/README.md
```diff
@@ -0,0 +1,233 @@
...
+
+The purpose of this CIP is to introduce the mechanism for DRep pledge first. Future CIPs may use DRep pledge for voting-power limits, compensation eligibility, Sybil resistance, reputation signaling, or other incentive mechanisms. Those future mechanisms should not activate before DReps have had adequate time to declare pledge and for tooling to add support once this capability exists on-chain.
+
+## Motivation: Why is this CIP necessary?
+
+### DReps have delegated voting power but no pledge field
+
+[CIP-1694](../CIP-1694/README.md) introduced DReps as one of Cardano's governance bodies. Any Ada holder may register as a DRep, and other Ada holders may delegate voting rights to that DRep. DRep voting power is based on delegated Lovelace.
+
+This establishes the delegation side of the system, but it does not establish an on-chain pledge field for DReps. A DRep may have large delegated voting power without any ledger-native way to declare personal economic commitment. Today, tools can show metadata, payment addresses, voting history, and delegated stake, but not a ledger-defined DRep pledge.
```
Ryun1, line 30: To me, this doesn't feel like a problem
this is just the current design

Perhaps the problem is: delegators want to know how much 'skin in the game' each DRep has?

CIP-XXXX/README.md
```diff
@@ -0,0 +1,233 @@
...
+
+### Metadata is useful but insufficient for future ledger rules
+
+[CIP-0119](../CIP-0119/README.md) defines useful DRep metadata. That metadata can describe a DRep's identity, motivations, qualifications, payment address, and other public information. It is intentionally off-chain and flexible.
+
+Pledge is different. If future ledger rules are expected to use pledge as an input, the value should be represented directly in ledger state. A metadata-only pledge field would be useful for social signaling, but it would not provide a deterministic, consensus-visible value that later ledger rules can reference without depending on off-chain metadata interpretation.
```
Ryun1, line 36: as the primary motivation for this proposal seems to be  this fields use in future proposals, I feel it may be better for this change to be bundled into those potential future proposals

because as it stands I don't feel this motivation is sufficient

(please correct me if I have misinterpreted this motivation)

CIP-XXXX/README.md
```diff
@@ -0,0 +1,233 @@
...
+- "Declared DRep pledge": DRep pledge recorded in ledger state.
+
+### DRep registration certificates
+
+DRep registration certificates MUST include `drepPledge`.
+
+In CDDL-like form:
+
+```cddl
+reg_drep_cert = (16, drep_credential, coin, anchor / nil, drep_pledge : coin)
```
```suggestion
reg_drep_cert = (16, drep_credential, coin, anchor / nil, coin / nil)
```
maybe it would be nicer to keep it as backwards compatible
and allow pledge to be optional

CIP-XXXX/README.md
```diff
@@ -0,0 +1,233 @@
...
+
+`drepPledge` MAY be any non-negative `Coin` value that can be represented by the ledger's `Coin` type.
+
+This CIP does not introduce a minimum DRep pledge, maximum DRep pledge, pledge ratio, pledge leverage parameter, or pledge-specific protocol parameter.
+
+### Authorization
+
+DRep registration, update, and deregistration authorization remains unchanged from CIP-1694.
+
+The DRep credential or script that authorizes the DRep registration or update also authorizes the declared `drepPledge` value in that certificate. This CIP does not introduce any additional witness, payment credential, stake credential, owner credential, or pledge-specific authorization requirement.
```
Ryun1, line 123: Would existing script-based DReps be able to interact with these updated certificates?

or would there need to be a new Plutus version and thus deployment for existing scripts?

CIP-XXXX/README.md
```diff
@@ -0,0 +1,233 @@
...
+```
+
+where:
+
+- `drep_credential` is the DRep credential.
+- `coin` remains the DRep registration deposit supplied by the certificate.
+- `anchor / nil` remains the optional DRep metadata anchor.
+- `drep_pledge` is the declared DRep pledge.
+
+When a DRep is registered, the ledger MUST store `drepPledge` in the DRep's ledger state.
```
Ryun1, line 69:
```suggestion
When a DRep is registered, the ledger MUST store `drepPledge` in the DRep's ledger state.

The `coin` specified for pledge must be covered by the inputs of the transaction.
```

maybe I missed it?
but I feel it would be good to explicitly say that the pledge amount have to be passed into the transcation

CIP-XXXX/README.md
```diff
@@ -0,0 +1,233 @@
...
+
+- `drep_credential` is the DRep credential.
+- `anchor / nil` remains the optional updated DRep metadata anchor.
+- `drep_pledge` is the updated declared DRep pledge.
+
+When a DRep update certificate is accepted, the ledger MUST replace the previous `drepPledge` value for that DRep with the value supplied in the update certificate.
+
+An update that changes only the DRep metadata anchor still MUST include `drepPledge`. If the DRep does not intend to change pledge, the update certificate should resubmit the current pledge value.
+
+### DRep ledger state
```
Ryun1, line 93: it might be my ignorance to how SPO pledge currently works

but does pledge still get counted towards voting power? or is it more like the DRep deposit, where it is not counted?

CIP-XXXX/README.md
Cerkoryn: Agreed, this is probably an elusive and maybe difficult to quantify problem that warrants a CPS.  I will explore that option and see how it comes together.

CIP-XXXX/README.md
Cerkoryn: Agreed, and with the similar concerns expressed by @maureenblack and @jennycita.  I think I'll try and do some research to come up with a more holistic solution that is self-sufficient, rather than open-ended as this draft currently is.  Potentially with a CPS as well.

CIP-XXXX/README.md
Cerkoryn: Excellent question.  It would likely have to reflect a design similar to [CIP-175](https://github.com/cardano-foundation/CIPs/pull/1140) that allows for multisig hot credentials for stake pools.

CIP-XXXX/README.md
Cerkoryn: It is liquid ADA in a pool owner's pledge wallet, so it counts for both pledge and voting power.  But "declared pledge" is a threshold that the wallet must not dip below, otherwise pool's rewards will be forfeited on the epoch boundary.

Cerkoryn:
@rphair @Ryun1 @perturbing I've replied to the outstanding comments in this discussion.  Current plan is to do more research and come back with either a CPS, a holistic CIP with a complete solution, or a combination of both.

Feel free to close this PR at your convenience 🫡

PR #1205, opened 2026-06-03
CIP-0190? | Proof of Existence Transaction Metadata

ingvarson:
This PR proposes a new CIP specifying the schema, algorithm registries, canonical-CBOR encoding rules, and verification procedure for a **Proof of Existence (PoE)** record carried in Cardano transaction metadata under **metadata label 309**.

### Summary

A PoE record commits one or more cryptographic content hashes to the chain; the block time of the transaction is the authoritative witness that the content existed no later than that moment. A conformant verifier needs only the transaction metadata, optionally the content bytes, and a public blockchain explorer, with no issuer server, certificate authority, or trusted intermediary at any step. The record is content-first, issuer-agnostic, storage-agnostic, standalone-verifiable, and algorithm-agile (including a registered post-quantum hybrid KEM for the optional sealed-PoE key path).

### On the CIP number

This proposal makes no claim to a CIP number; the editors assign the number and Category at triage. The `309` in this document is the CIP-10 metadata label (`transaction_metadatum_label`), which is the on-chain identifier that implementations cite, not a proposed CIP number. The folder is named `CIP-0190`.

### Relationship to CIP-10

Metadata **label 309** is already reserved as `"Proof of Existence record"` in the [CIP-10 registry](https://github.com/cardano-foundation/CIPs/blob/master/CIP-0010/registry.json), and this CIP formalizes its schema. Once a CIP number is assigned, I'll add a `CIP-0010/registry.json` change to this same PR to attribute that entry and cross-link it to the number.

### Contents of this folder

- `README.md`: the proposal (drafted against the CIP-0001 template; `Status: Proposed`, `Category: Metadata`).
- `label-309.cddl`: the normative CDDL grammar for the record body.

The inline registry and error-code tables in the proposal are complete and normative. The wider machine-readable corpus, meaning the full algorithm registries, JSON Schemas, and pinned conformance vectors, lives in the public reference repository ([cardanowall/label-309](https://github.com/cardanowall/label-309)) and is linked from the proposal rather than duplicated here, to keep the folder lean.

### Reference implementations

Byte-parity reference implementations against shared canonical-CBOR test vectors are published open-source:

- TypeScript: https://github.com/cardanowall/label-309-ts
- Python: https://github.com/cardanowall/label-309-py
- Rust: https://github.com/cardanowall/label-309-rs
- CLI: https://github.com/cardanowall/label-309-cli

Feedback from the editors and community is very welcome.

[...]

PR #1226, opened 2026-07-10
CIP-0010 | Label 309: add CIP-0190

ingvarson:
Metadata label 309 is already reserved in this registry as "Proof of Existence record", but the entry does not name the specification that defines it. This PR adds that pointer.

Label 309 is specified by [CIP-0190](https://github.com/cardano-foundation/CIPs/tree/master/CIP-0190) (Proof of Existence Transaction Metadata), merged into this repository. The registry schema has no `url` or owner field, so the attribution belongs in the description text, the same way labels 721, 777, 867 and 3692 name CIP-0025, CIP-0027, CIP-0088 and CIP-0149 respectively.

The change is one line:

```diff
   {
     "transaction_metadatum_label": 309,
-    "description": "Proof of Existence record"
+    "description": "CIP-0190 - Proof of Existence record"
   },
```

No other entry is touched, the entry count stays at 47, and `registry.json` still validates against `registry.schema.json`. The new description is 36 characters, well inside the 256 limit.

Nothing changes on-chain. The load-bearing identifier is the integer 309; the description is human-readable documentation for wallets, explorers and indexers, which is exactly the audience that currently has no path from the label to the spec.

### DIFF
--- a/CIP-0010/registry.json
+++ b/CIP-0010/registry.json
@@ -29,7 +29,7 @@
   },
   {
     "transaction_metadatum_label": 309,
-    "description": "Proof of Existence record"
+    "description": "CIP-0190 - Proof of Existence record"
   },
   {
     "transaction_metadatum_label": 544,
--- a/README.md
+++ b/README.md
@@ -178,7 +178,7 @@ CIP editors facilitate discussions and progress submissions on GitHub, reviewing
 
 The following link lists "candidate" CIPs still under discussion with the community; these are assigned numbers to avoid later clashes and to facilitate community discussion (see further below for stalled proposals):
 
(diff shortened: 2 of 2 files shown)


rphair (approved):
Already suggested in the course of https://github.com/cardano-foundation/CIPs/pull/1205 — thanks @ingvarson for attending to this.

Note in renaming the GitHub label so it accommodates PR's like this (which don't create a "new label"), I had to make some changes to the PR label-based queries in the repo `README` (and the Wiki on [this page](https://github.com/cardano-foundation/CIPs/wiki/301.-State-tagging)).

@Ryun1 @Godspeed-exe why do you think the CIP-0010 registry validator doesn't run on this PR?

Ryun1 (approved):

PR #1227, opened 2026-07-10
CIP-0164 | Leios Mini-Protocols

coot:
* Edited the paragraph on protocol pipelining
* Added three messages:
  - `MsgLeiosNoBlockAnnouncement` (`LeiosAnnounce` mini-protocol)
  - `MsgLeiosNoVotes` (`LeiosVotes` mini-protocol)
  - `MsgLeiosBlockNoOffers` (`LeiosBlockNotify` mini-protocol)
* Split the mini-protocols IER table into four parts, one per mini-protocol.

### DIFF
--- a/CIP-0164/README.md
+++ b/CIP-0164/README.md
@@ -116,6 +116,7 @@ technical resources, visit the Leios Innovation R&D site at
 - [Figure 6: LeiosVotes mini-protocol state machine](#figure-6b)
 - [Figure 6: LeiosBlockNotify mini-protocol state machine](#figure-6c)
 - [Figure 7: LeiosFetch mini-protocol state machine](#figure-7)
+- [Figure:   Protocol Pipelining](#figure-protocol-pipelining)
 - [Figure 8: SPO profitability forecast under Leios](#figure-8)
 - [Figure 9: Time for transaction to reach the ledger](#figure-9)
 - [Figure 10: Transactions reaching the ledger](#figure-10)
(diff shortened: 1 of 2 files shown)


CIP-0164/README.md
```diff
@@ -1283,6 +1294,54 @@ after each 100 received). Sending the next request before the previous one has
...
+###### Clean termination
+
+We need `MsgLeiosNoBlockAnnouncement` message in order to retain the ability for
+the client to terminate a mini-protocol within a reasonable time limit.
+Without such a message, the server after receiving `MsgLeiosVotesRequestNext`
+will block until data is available, which might or might not happen quickly
+enough.  To avoid unclean mini-protocol shutdown, `MsgLeiosNoBlockAnnouncement`
+can be used, which gives back the agency to the client, which can decide
+whether to continue the mini-protocol (e.g. further await for data
+availability) or terminate the mini-protocol with `MsgDone`.  The
```
nfrisby, line 1326: > The `cardano-node` implementation deactivation timeout is set to `300s`.

Based on where that sentence occurs and the fact that I'm not familiar with the phrase "The `cardano-node` implementation deactivation timeout", my first interpretation was that you were suggesting 300s for the idle timer to cause the server to send `MsgLeiosNoBlockAnnouncement`.

I just CTRL-Fed and confirmed that "deactivation" only occurs in this paragraph.

Would you elaborate what the meaning of "The `cardano-node` implementation deactivation timeout" is, in order for the reader to understand that side of tradeoff this paragraph is highlighting? Maybe just bring the second-to-last sentence up to before this reference to it?

CIP-0164/README.md
```diff
@@ -1283,6 +1294,54 @@ after each 100 received). Sending the next request before the previous one has
...
+`cardano-node` implementation deactivation timeout is set to `300s`.  Note that
+this timeout must still be preserved if the mini-protocol is pipelined, e.g. if
+we pipeline five `MsgLeiosAnnounceRequestNext` messages, and we send
+`MsgLeiosNoBlockAnnouncement` after `10s`, it will take the client exactly `50s`
```
nfrisby, line 1330: > it will take the client exactly `50s` to await for all the pipelined `MsgLeiosAnnounceRequestNext` replies to arrive before it will be able to send `MsgDone`

The `typed-protocol` EDSL forces the pipelined peer to have collected a response to all of its requests before it can terminate with the `Done` action.

But... what's stopping it from sending a _pipelined_ `MsgDone`?

- It sends N pipelined requests.
- It receives signal to terminate.
- It can't send `MsgDone` because the pipeline is full (max depth already reached).
- As soon as _one_ `MsgLeiosNoBlockAnnouncement` arrives, it can send `MsgDone`.
- If the server has a LookAhead (see https://github.com/IntersectMBO/typed-protocols/pull/93), then the server can _immediately_ react to that `MsgDone` by sending N-1 `MsgLeiosNoBlockAnnouncement`s. (IE, basically cancelling the pipelined requests.)
- Then both server and client can gracefully terminate---after exactly one server-side timeout (+ ~1 RTT).

What do you think?

nfrisby: > What do you think?

In particular, does this disarm the tradeoff the paragraph mentions?

CIP-0164/README.md
coot: If the outbound governor decides to demote the peer, then it signals each mini-protocol and if any of them is not finished within deactivation timeout (e.g. `300s`), then the connection will be closed abruptly (non-clean shutdown).  This gives a different outcome for `Hot -> Warm` demotion, since in the non-clean demotion the peer will end up as `Cold` rather than `Warm`.

CIP-0164/README.md
coot: Even if you pipeline `MsgDone`, you still need to collect all the results, including the result of sending `MsgDone` before you'll be able to use the `Done` constructor - so it's not allowed by construction.

coot: I don't think we can pipeline `MsgDone` and terminate the protocol, while the other side is still sending requested responses.  We still need to receive all the data from the sending peer, so that the mini-protocol can be restarted if the peer is re-promoted: demotions / promotions can happen multiple time within lifetime of a connection.

PR #1228, opened 2026-07-12
CIP-0010 | Add BROCK Bot label

BrockCruess:
BROCK Bot is a Discord moderation bot. Servers pay for it in Cardano stablecoins (USDA, USDM, USDCx, DJED), and label 6607 records which Discord server a payment is for - otherwise a plain transfer to our address gives no way to tell.

Payload is one object with two keys:
`{ "6607": { "server_id": "837215135999197246", "months": 2 } }`

`server_id` is the Discord guild ID; `months` is how many 30-day increments were purchased (1–12).

It's posted only when someone buys or tops up a subscription from our payment page - one transaction per payment, signed by the user in their own wallet. Nothing automated, nothing periodic, no other bot activity touches the chain. The bot then reads the transaction back and verifies the recipient and amount itself before crediting the server; the metadata is just the reference.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,2 @@
+# Auto detect text files and perform LF normalization
+* text=auto
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,4 +1,2 @@
-# Auto detect text files and perform LF normalization
-* text=auto
 * linguist-vendored
 *.md linguist-vendored=false
--- a/CIP-0010/registry.json
+++ b/CIP-0010/registry.json
@@ -144,6 +144,10 @@
     "description": "CIP-0149 - Optional DRep compensation"
   },
   {
+    "transaction_metadatum_label": 6607,
+    "description": "BROCK Bot - Discord moderation bot subscription payment"
+  },
+  {
     "transaction_metadatum_label": 6676,
     "description": "Maya Protocol - Cross-chain swap memo (bifrost observer)"
   },


.gitattributes
```diff
@@ -1,2 +1,4 @@
+# Auto detect text files and perform LF normalization
+* text=auto
```
rphair:
```suggestion
```
This is far out of scope for a metadata registry change.  I wouldn't suggest or agree with it since it makes no sense on a repository that contains no code other than examples assumed to be part of text files.  In fact `.gitattributes` is only here to _avoid_ interpretation of any file on the CIP repository as code:
* https://github.com/cardano-foundation/CIPs/pull/1154

If you have some other reason why you'd like to suggest this, open up another PR or an issue to discuss the matter please.

rphair (commented):
@BrockCruess can please you provide some documentation — for readers' sake, maybe adding it into the OP above (currently empty) — about what this bot does and/or what you intend to do with it: including how & when it posts metadata to the chain?

.gitattributes
BrockCruess: Sorry I have no idea why that was changed or how it got in there lol, looks like it was from years ago. Totally fine to remove.

BrockCruess:
> @BrockCruess can please you provide some documentation — for readers' sake, maybe adding it into the OP above (currently empty) — about what this bot does and/or what you intend to do with it: including how & when it posts metadata to the chain?

Done!

rphair (approved):

Ryun1 (approved):

PR #1231, opened 2026-07-21
CIP & CPS Validators improvements and consolidate shared logic

Ryun1:
To address #1216

### Changes

- All logic that was duplicated between `validate-cip.py` and `validate-cps.py` into a new `.github/scripts/validation_common.py` (frontmatter parsing, number/Authors/Created/License checks, label–URL entry validation, directory-name checks, etc.).
- Refreshed error wording throughout so failures state the rule, show the offending value, and say how to fix it.
- Moved most field-level value rules out of the JSON schemas and into the scripts, so authors can get a targeted message instead of a raw schema-pattern failure.
- Unassigned numbers: alongside the quoted `"?"`, the word placeholders `unassigned` / `pending` / `TBD` are now accepted
- The `?` suffix on `CIP-NNNN?` / `CPS-NNNN?` references is now optional when linking to a pull request rather than mandatory, it marks a candidate still in PR, and the plain reference is also accepted. It remains invalid on merged documents.
- Removed the body cross-reference check (requiring CIP-NNNN/CPS-NNNN mentions in the document body to resolve to existing folders), since it can't distinguish intentional references to not-yet-merged proposals.


rphair (approved):
@Ryun1 this all looks very promising: and seems like a more maintainable design.  Looking forward to testing everything in practice as soon as we can merge this 😎 (cc @perturbing)

perturbing (approved):

Ryun1:
> Unassigned numbers: alongside the quoted "?", the word placeholders unassigned / pending / TBD are now accepted

should we also allow multiple `?`s? i.e. `???`
as this doesn't seem to cause invalid yaml

@rphair

rphair:
Technically we can allow anything inside the quotes (though no reason to quote them unless `?` or other metacharacters).  But I don't see an advantage for e.g. using `"???"` when `"?"` also serves that purpose: though since I think we've been telling people they can use any meaningful placeholder (particularly **`TBD`**: don't know if we're currently allowing that, but I think it would be helpful) then I think the more things we allow the better.

In any case @Ryun1 please merge this if you think it's ready... after at your discretion making any changes about what is allowed here. 😎

PR #1233, opened 2026-07-22
CIP-???? | Native Confidential Transfers

matiwinnetou:
This PR proposes a new **Ledger**-category CIP introducing **native confidential transfers**: transaction amounts (ADA and native-token quantities) hidden on chain as additively — partially — homomorphic Pedersen commitments, verified by zero-knowledge proofs of range and per-asset value conservation. The ledger never decrypts anything; validators check proofs only.

**Motivation.** Some entities need to transact **without the rest of the world learning the details of their financial activity — above all, how much they transact and hold**. Businesses running payroll, paying suppliers, settling invoices, or managing treasuries (and individuals alike) cannot reasonably do so on a ledger where every amount is public — yet they still need a verifiable, non-custodial ledger and the ability to satisfy auditors and tax authorities. This proposal therefore aims for **confidentiality from the general public, not from oversight**: amounts are hidden from everyone except the parties the account owner deliberately authorises. The owner can, of their own volition, designate **one or more auditors** — a tax authority, an accountant, tax-reporting software — by sharing the account's single viewing key (one key per stake address). Disclosure grants read-only access, never spend authority.

**Pure privacy / anonymity transfers are explicitly out of scope.** Designs that hide identities or break the transaction graph (mixers, shielded pools, and similar) address a different need and are better served by other CIPs or solutions. This proposal targets use cases where confidentiality and auditability must coexist.

**Highlights**

- Purely **native** (ledger-level): no smart contracts, escrow, or mixer — and the Rationale now argues the native-vs-application-layer choice explicitly rather than asserting it.
- Minimal primitive set, **no trusted setup**, and **no SNARKs, pairings, or circuits**: ristretto255 (prime-order group), Pedersen commitments, aggregated Bulletproofs range proofs, Diffie–Hellman amount transport, per-asset Schnorr balancing signatures, Fiat–Shamir with namespaced domain-separation tags.
- **Partially homomorphic by design (PHE class, addition only) — deliberately *not* fully homomorphic encryption (FHE)**: balance accounting needs nothing beyond addition, subtraction, and range checks, which keeps every ledger operation at microsecond scale.
- Addresses, the transaction graph, fees, and asset identities stay **public**; only quantities are hidden. Receiving is **opt-in per account, enforced by consensus**: a viewing-key **registration certificate** (with a refundable `viewingKeyDeposit`) binds the account's `P_view` to its stake credential; key-hash and **script-hash credentials register identically — multisig/timelock treasuries are first-class citizens** of v1.
- **Bidirectional auditability from a single key, at zero extra bytes on chain**: incoming amounts by direct decryption; outgoing amounts via deterministic ephemeral scalars, recomputable from the sender's viewing key against the public registry (a structural dividend of keeping addresses public — recipient-hiding designs must pay an extra encrypted record per output for the same property).
- On-chain structures are **versioned the ledger's way** (tagged unions + numbered map keys, no in-band version fields), with a normative **guarantees-to-future-proposals** section fixing what companion CIPs may rely on, and named verification-budget protocol parameters (`maxConfidentialCommitmentsPerTx` / `PerBlock`, `viewingKeyDeposit`).
- Confidential ADA does not contribute to stake, rewards, or governance voting power while hidden; unshielding restores participation. Trade-offs documented honestly, including ~3–5× transaction size (and fees), ~10–20× verification CPU, the smart-contract exclusion while hidden, and no post-quantum security.
- Worked example, illustrative CDDL (aggregated as a draft schema in `cddl/`), and a companion **[extensions.md](https://github.com/matiwinnetou/CIPs/blob/confidential-transfers/CIP-confidential-transfers/extensions.md)** scoring 12 anticipated extensions by the problem each solves, implementation effort, impact, new-cryptography burden, and compliance posture.

**Feedback especially welcome on:**

- the proposed **CIP-1852 role `6`** derivation path for viewing keys (pending wallet-ecosystem confirmation);
- the **scope of the initial hard fork** — which low-effort extensions should be merged into this CIP or activated alongside it (viewing-key **rotation** is the lead candidate; see extensions.md);
- confirmation of the balancing-proof form (per-asset Schnorr excess, constrained to remain aggregatable);
- Plutus script addresses (deferred by design; validation rule 10).

Number pending assignment by CIP editors (directory currently `CIP-confidential-transfers`).

📄 **[Rendered proposal](https://github.com/matiwinnetou/CIPs/blob/confidential-transfers/CIP-confidential-transfers/README.md)** · 🧩 **[Extensions analysis](https://github.com/matiwinnetou/CIPs/blob/confidential-transfers/CIP-confidential-transfers/extensions.md)**


Ryun1:
very cool and interesting proposal!
keen to see it develop

CIP-confidential-transfers/README.md
```diff
@@ -0,0 +1,553 @@
+---
+CIP: "?"
+Title: Native Confidential Transfers for ADA and Native Tokens
```
Ryun1:
```suggestion
Title: Native Confidential Transfers
```
a little snappier?

matiwinnetou: Agreed — applied in da5a2c1. Thanks!

fallen-icarus:
Isn't this the point of Midnight? Cardano users can utilize Midnight as a VPN for capital: 
- When you bridge to Midnight into a shielded address, you disappear.
- When you bridge back to Cardano, no one can see where the funds came from.

I wrote about this use case in my blog post [Public Markets, Private Participants](https://github.com/fallen-icarus/meditations-blog/blob/main/public-markets-private-participants/README.md).

---
I'm not opposed to this idea. But why make such a large change to Cardano when the niche can be satisfied by working with a partnerchain? Specialized (aka modular) chains working together is more maintainable and scalable.

matiwinnetou:
@fallen-icarus - yes it is somewhat a point of Midnight but Midnight goes much deeper than that, it is ZK first. There is no intention of this here and Cardano cannot become in the short term future a ZK first blockchain.

In fact even some people from Midnight agree that in the longer term Cardano can get some goodies from Midnight like: Starstream VM, confidential transfers. Its a common understandable knowledge, the question is not technical feasibility but rather community decision and financing support for those objectives.

CIP-confidential-transfers/README.md
```diff
@@ -0,0 +1,762 @@
...
+entirely unaffected.
+
+## Specification
+
+> **Scope.** This proposal specifies amount confidentiality for ADA and native tokens with the
+> asset *type* (policy id and asset name) remaining public. Hiding the asset type, hiding
+> sender/recipient identity, confidential outputs at script addresses, and interaction with
+> programmable tokens are **out of scope** and discussed under [Open Questions](#open-questions).
+
+### 1. Overview
```
rphair:
```suggestion
### Overview
```
When these enumerations inevitably get resequenced as the material gets reshaped & reorganised (especially in the current `Draft` state) — both during editing & perhaps after any post-merge updates to this CIP (likely, given its interdisciplinary scope) — incoming links to these sections will get broken.

Fixed section numbering is mainly applicable to static material determined in advance never to be changing.  I can't imagine how that would always be true for this specification & so the currently applied numbers might not be useful for _any_ purpose: even communication without URLs.

If you need to keep the section numbering for some specific implementation goal then please let us know: otherwise editorially it's considered important to keep all numbering out of section headings.

rphair (commented):
thanks @matiwinnetou @paweljakubas & associates: an exciting idea and I would love to see what use cases emerge for this.

Please post for the editors' attention If interested in a CIP meeting review (while still in `Draft` state; otherwise we delay `Triage` until you mark it _Ready for review_) or a CIP Discord channel for any related higher-bandwidth discussion.

@kozross: re: https://github.com/cardano-foundation/CIPs/pull/1143 you may be interested in an early look at this based on the application of homomorphism.

Once line-by-line reviews start coming, please keep in mind: 🙏
* it's easier for everyone to follow collective review when you commit any change suggestions through GitHub UI & then pull them back to your working branch;
* consequently: whenever forcing a push, please ensure these commits are included (since forced push of an empty branch will overwrite the commit history).

CIP-confidential-transfers/README.md
matiwinnetou: thanks, done

matiwinnetou:
@rphair thanks! This is still a draft and not ready for review yet - we are still shaping it. Once it is ready, we will mark it as Ready For Review and ask the CIP editors to take a look.

paweljakubas:
Several questions that might be good to answer.

**Q1. On the BLS12-381 alternative**

The CIP chose `ristretto255` for `Ed25519` field reuse. Please clarify why `BLS12-381 G1` was not considered as the commitment group — it is already prime-order, already consensus-critical, and has native Plutus builtins (`G1_add`, `G1_scalarMul`, `G1_equal`, `G1_hashToGroup`) that support exactly the operations `Pedersen` commitments need! Was this evaluated and rejected? If so, what were the specific reasons? If not, any chance to give it a brief analysis? This is the single question on which the entire "new primitives vs. existing infrastructure" disagreement hinges IMHO. 

Please be noticed that **Twisted ElGamal feasible over BLS12-381 G1** !!!
**No new Plutus built-ins are needed.**

The CIP and Mysten both solve this the same way: split the amount into small limbs (e.g., u16) and encrypt each limb separately. Each limb's discrete log is brute-forceable from a precomputed table of ~2^16 entries. This is why the CIP's range proofs are necessary — to prove each limb is within the valid range without revealing it.
Over BLS12-381 G1, the same approach works. The only difference is the field size (381 bits vs 255 bits for ristretto255), which affects:
- Ciphertext size: 48 bytes per point (compressed) on BLS12-381 G1 vs 32 bytes on ristretto255 — a 50% size increase per commitment
- Scalar multiplication cost: Slightly more expensive on the larger field — but this is prover-side only; the verifier uses Plutus builtins which are constant-time

paweljakubas:
**2. On the CPU multiplier and block validation deadline**

The CIP's own appendix documents a ~10–20× CPU multiplier for confidential transaction verification (~2–3 ms unbatched vs 0.1–0.2 ms transparent). For a block with 60 fully-confidential transactions, that is 0.1–0.2 s of additional verification within the slot budget. The CIP proposes `maxConfidentialCommitmentsPerBlock` as a tuning knob. Sub-questions  to understand:
- What is the estimated slot budget consumed at 100% confidential adoption (every tx in a block confidential)?
- Has the 80% stake activation threshold in Path to Active been validated against this budget, or is it aspirational?
- Is the ~2–3 ms figure a single-threaded measurement? Cardano nodes run verification in parallel where possible — does batch verification actually translate to wall-clock improvement on current node architecture?

paweljakubas:
**3. On cross-curve ZK composability**

The CIP's future extensions mention ZK predicates over commitments for governance (homomorphic tallying) and programmable tokens (amount-dependent policies). These assume `ristretto255`-native proving tools. But the Cardano ecosystem's ZK tooling is built on `BLS12-381` (Groth16, and potentially PLONK in the future). The CIP does not address the gap: if a developer wants to prove "my hidden balance exceeds X" or "my hidden amount is in range a,b" as part of a Groth16 circuit that also checks credential predicates, the `ristretto255` commitment cannot be efficiently represented in a `BLS12-381` R1CS. Of course, to my best knwoledge. Is this cross-curve boundary a conscious trade-off the author accept, or is it intended to be addressed by `ristretto255`-native proving tools that do not yet exist on Cardano? If the latter, what is the ecosystem dependency?

paweljakubas:
**4. On script-address confidentiality timeline**

Validation rule 10 prohibits Plutus script execution alongside confidential components. The CIP defers script-address confidentiality to future companion CIPs. The question is about timeline and dependency: the CIP's Path to Active requires 80%+ stake activation. If a significant use case for confidential transfers involves smart contracts (e.g., a DEX that needs to hide trade amounts), the key-locked restriction could limit adoption before the companion CIPs arrive. Is there an estimated timeline for the script-address companion? Is it a dependency for Path to Active, or is v1 intended to ship without it?

paweljakubas:
**5. On the viewing key registration deposit**

The CIP introduces `viewingKeyDeposit` (initially equal to `stakeAddressDeposit`, currently 2 ADA). This is a refundable deposit locked when registering a viewing key. The CIP also introduces a deregistration certificate that refunds the deposit. Question: does deregistration immediately revoke the ability to receive new confidential outputs (validation rule 12), or does it also affect the ability to spend existing confidential outputs? The CIP states "existing confidential outputs remain readable and spendable" after deregistration — but if the viewing key is derived from the wallet seed (not from the certificate), then deregistration is purely a directory entry removal, not a key revocation. Is this correct? If so, what prevents a sender from continuing to send confidential outputs to an address whose registration has been deregistered? (Presumably validation rule 12 catches this at the ledger level — but to confirm the mental model.)

paweljakubas:
**6. On the no-Plutus-execution rule (validation rule 10)**

Rule 10 says: "A transaction that contains confidential inputs or outputs must not execute Plutus scripts." This means a confidential transfer cannot also pay a DEX fee, interact with a lending protocol, or do anything else that involves Plutus — even if the Plutus script does not inspect amounts. The CIP acknowledges this and says the path is "unshield → interact transparently → optionally shield back." Each crossing reveals the bridged amount. So is this transfer-only-while-hidden restriction considered a feature (simpler security model) or an accepted limitation? If the latter, what is the anticipated cost of the companion CIP that would relax it (script-context extension + ZK predicate over hidden quantities)?

paweljakubas:
**7. Bulletproof - why not more modern system?**

What about https://eprint.iacr.org/2022/510 ? Why not this approach?
Please be noted that **Bulletproofs++**  is a transparent, drop-in improvement over Bulletproofs — same security model, same features, **~3–5× faster verification**, **~38% smaller proofs** — and would significantly reduce the CIP's CPU multiplier while preserving the no-trusted-setup property.

paweljakubas:
**8. The CIP creates a dual-system complexity burden on the entire ecosystem.** 

The CIP does not replace transparent transfers — it adds 12 new validation rules, new certificate types (viewing-key registration), new cryptographic operations (DH transport, Bulletproofs verification), and new wallet logic (key derivation, amount recovery, shield/unshield flows) that must coexist with the existing transparent system. Every component — ledger, wallets, block explorers, audit tools, developer tooling — must now handle two parallel transfer modes. Every future Cardano upgrade must consider its interaction with both. 

We can envisage an application-layer approach which adds zero complexity to the ledger; the Gate Script is self-contained, and if it has a bug, it affects one application, not every confidential transaction on the network.

How would you kindly answer to this line of reasoning and still advocate for CIP's approach?

matiwinnetou:
> **5. On the viewing key registration deposit**

Correct on all counts — your mental model matches the intended design exactly:

1. **Deregistration is directory-entry removal, not key revocation.** The keypair derives from the wallet seed and never depends on the certificate, so nothing about existing outputs changes: they remain readable and spendable by their owner (and by anyone previously given `sk_view`).
2. **Receiving stops at the ledger level.** Validation rule 12 requires every confidential output to be addressed to a stake credential with a *live* registration, so after deregistration a transaction attempting to send confidentially to that account is simply invalid — the sender's wallet cannot even construct it against current chain state, and if it tried, validators reject it. Nothing relies on sender politeness.
3. Spending needs no registration at all — rule 12 constrains *outputs being created*, not inputs being consumed.

This is also why the spec calls the certificate "a directory entry, not a key store".

matiwinnetou:
> **4. On script-address confidentiality timeline**

v1 is intended to ship without it, and it is deliberately **not** a Path-to-Active dependency. The reasoning, briefly:

1. **The compliance-first target audience transacts, first and foremost** — payroll, B2B settlement, treasury operations. Those work day one. Note also that as of the latest revision, **native-script addresses (multisig, timelocks) are already admitted in v1** — so corporate treasuries are covered in the base proposal; only *Plutus* addresses remain excluded (the "Plutus script addresses" open question).
2. **No timeline is promised, and that is intentional**: the script track has a real dependency ladder (ristretto builtins → script-context extension → amount predicates — items 4/6/7 in [extensions.md](https://github.com/matiwinnetou/CIPs/blob/confidential-transfers/CIP-confidential-transfers/extensions.md)) and pretending to a date would be theater. What the proposal *does* guarantee is that the door is structurally open: rule 10 is explicitly protocol-version-relaxable, and the "Guarantees to future proposals" section fixes what companions may rely on.
3. Worth stating plainly: a DEX that *reads* trade amounts cannot work over hidden amounts on any chain without bespoke ZK protocol design — that limitation is not specific to this proposal's sequencing (see the smart-contracts trade-off bullet).

And the broader point your question really touches: **the v1 feature list is not finalised — it is exactly what this review is meant to settle.** The current draft is deliberately a bare minimum (native-script/multisig support being the one extension already absorbed, on precisely this reasoning), and the authors are aware that an MVP cut too narrow could foreclose adoption-critical use cases. That is why the "Scope of the initial hard fork" open question exists: every anticipated extension can land in one of three ways — **merged into this CIP**, **a separate CIP activated in the same hard fork**, or **a later CIP** — and which items deserve which treatment is an open discussion we explicitly invite. If you believe a use case (DEX-class or otherwise) is important enough to pull parts of the script ladder forward, that is exactly the input we are asking for.

matiwinnetou:
> **6. On the no-Plutus-execution rule (validation rule 10)**

Honestly: both, in layers.

1. **The rule itself is a feature** — not merely simplification. Today's script context represents every output with a cleartext amount; there is no way to *represent* a hidden quantity in `TxInfo` at all. So "a script that doesn't inspect amounts" is not actually expressible: the ledger must either give scripts a complete view of the transaction or refuse to run them, and a script validating blind to parts of the transaction it checks is a known attack surface (double-satisfaction-style). Rule 10 picks the only sound option available *at this Plutus version*. This is also why the DEX-fee example fails today independent of policy: the context for that script literally cannot be constructed.
2. **The resulting user-facing restriction is an accepted, documented limitation** — the Trade-offs section states it bluntly (transfer-only while hidden; not a one-way door; the timing-correlation caveat), and wallets are told to surface it before users shield.
3. **The anticipated relaxation cost is itemised in [extensions.md](https://github.com/matiwinnetou/CIPs/blob/confidential-transfers/CIP-confidential-transfers/extensions.md)** — which scores every anticipated extension on **implementation effort, impact, new-cryptography burden, and compliance posture** precisely so this kind of question has a concrete referent: ristretto builtins (medium effort; engineering + cost-modelling, no new crypto), script-context extension via a new Plutus ledger-language version (high effort; no new crypto), and only *amount-reading* policies need the ZK-predicate layer (existing proof classes — shifted-commitment range statements, sigma protocols — composed into new statements). Notably, identity/credential-based script logic works unchanged once the context can carry commitments, because addresses and datums stay public.

So the sequencing claim is: the restriction is sound today, honest about its cost, and priced for removal — not an architectural dead end.

matiwinnetou:
> **7. Bulletproof - why not more modern system?**

Bulletproofs++ was considered — and it is precisely why the proposal's proof carriers are **versioned tagged unions**. `range_proof = [0, bytes]` exists so that a successor system arrives as alternative tag `1` under a later protocol version, with old transactions verifying under tag `0` forever. BP++ is literally item 3 in [extensions.md](https://github.com/matiwinnetou/CIPs/blob/confidential-transfers/CIP-confidential-transfers/extensions.md), listed with exactly the profile you describe.

Why not tag 0, day one? Deployment maturity, not the math:

1. **Audit and implementation surface.** The dalek Bulletproofs implementation has been hardened by years of production use securing real value (Monero migrated to Bulletproofs in 2018; the dalek lineage is widely deployed and independently audited). BP++ is a younger construction whose early external reviews of draft versions surfaced real issues that required revision — normal for a young proof system, but consensus-critical verification is the single worst place on a blockchain to be an early adopter.
2. **The asymmetry of regret.** Shipping conservative-then-upgrading costs a hard fork we can schedule; shipping fast-then-discovering-a-soundness-gap in the proof system that guards value conservation is catastrophic and unschedulable. Given the versioning machinery makes the upgrade genuinely drop-in (same statements, same commitments — your own words: "drop-in improvement"), the option value of waiting costs us only the interim CPU/size delta.
3. **The economics stay honest either way**: the verification-cost trade-offs are documented against tag-0 Bulletproofs, so the appendix numbers are worst-case relative to any successor.

If BP++ (or a further successor) accumulates the audit trail and deployment history by the time this proposal reaches implementation, proposing it as tag 1 — or even arguing it should be tag 0 — is a perfectly good review outcome. The structure was built for exactly this conversation.

matiwinnetou:
> **2. On the CPU multiplier and block validation deadline**

Good questions — answers in order:

1. **100% adoption is the appendix's own worst case**: a block filled entirely with confidential transactions (~60 at current block sizes) adds on the order of **0.1–0.2 s** of verification unbatched, roughly half with cross-transaction batch verification — against a block-validation budget well under the slot interval, and with verification parallelising trivially across transactions (independent proofs). So the estimated worst-case consumption is a modest fraction of the budget, not a squeeze.
2. **The 80% threshold is activation governance, not a performance claim** — it is the standard readiness bar for any hard-fork feature (implementations deployed across supermajority stake). Performance validation is a *separate* acceptance criterion in Path to Active: published, measured benchmarks per transaction and per block, batched and unbatched, **including full-sync replay impact**, demonstrating block-validation budgets are met. The appendix numbers are explicitly labelled order-of-magnitude estimates to be *replaced* by those measurements.
3. **Yes — single-core estimates**, derived from published benchmarks of the underlying primitives; that is exactly why they are estimates and why the benchmark criterion exists. Whether batch verification translates to wall-clock improvement on real node architecture is precisely what those measurements must show; batching here is plain multi-scalar-multiplication aggregation (no cross-transaction data dependencies), so it composes with per-transaction parallelism rather than competing with it — but the burden of proof is on measurements, not this reply.

One more safety valve worth noting: `maxConfidentialCommitmentsPerBlock` is an *updatable protocol parameter*, so the per-block verification budget can be tightened or relaxed by governance without a hard fork — initial values come from the benchmarks, and mis-estimation is correctable in production.

matiwinnetou:
> **Q1. On the BLS12-381 alternative**

Great question — and you're right that the proposal should argue this choice explicitly rather than assert it. I am adding a **"BLS12-381 G1 as the commitment group"** entry to *Alternatives considered* elaborating the analysis below. Before that, two clarifications about the design as written, because parts of the question describe a different architecture than this proposal's:

1. **The verifier here is the ledger, not Plutus.** All proofs in this proposal are checked by ledger validation rules (node code, like signature verification) — no Plutus script ever runs on the validation path, so Plutus builtins buy nothing for the base protocol *regardless of curve*. "The verifier uses Plutus builtins" would hold for an application-layer implementation, not a native one. Where builtins genuinely matter is the future *script track* (contracts verifying predicates over commitments) — and there this choice does cost new ristretto255 builtins, which [extensions.md](https://github.com/matiwinnetou/CIPs/blob/confidential-transfers/CIP-confidential-transfers/extensions.md) item 4 prices explicitly. That is the honest cost of the ristretto choice, and it is a cost of a future extension, not of base validation.
2. **This proposal has no limbs and no discrete-log tables.** "The CIP and Mysten both solve this the same way: split the amount into u16 limbs… brute-forceable from a ~2^16 table" — that is Sui's Twisted-ElGamal architecture, which this proposal explicitly considered and **rejected** (see *Alternatives considered*, first entry): the Diffie–Hellman transport delivers the full 64-bit amount and its blinding directly (8-byte masked amount), with no bounded message space and no lookup tables. Relatedly, the range proofs are not there to keep limbs brute-forceable — they prevent negative/overflowing committed values (a scalar wrapping mod ℓ would mint money), which is required in any commitment-based design regardless of amount transport.

With the layers separated, the actual curve comparison for **native** validation:

- **No pairings are used anywhere** — BLS12-381's entire raison d'être. What remains of it is the bill: a 381-bit field vs 255-bit, i.e. slower group arithmetic on the hottest path in the system (every node, every block — in a native design the verifier is the whole network, not a prover-side concern).
- **+50% on every element** (48 vs 32 bytes): every commitment, every ephemeral key, every proof element; on top of an already 3–5× transaction-size trade-off.
- **Cofactor and canonicity**: G1 has a large cofactor (~2^125), so every deserialized point needs an explicit subgroup check — fast with endomorphism tricks, but a per-point cost and a classic implementation footgun. ristretto255 decoding yields canonical, prime-order elements *by construction*, which is what makes validation rule 1 trivially sound and transcript hashing unambiguous.
- **Proof-system ecosystem**: the hardened Bulletproofs lineage (dalek; Monero production since 2018) is ristretto-native. Bulletproofs over BLS G1 means a bespoke implementation without that audit history — the same conservatism argument as the Bulletproofs++ discussion (#7).
- **Ed25519 synergy**: ristretto255 shares its field with the signatures every node already computes at scale — same implementation techniques, same audit base, in node code where this proposal actually lives.

On your closing framing ("new primitives vs. existing infrastructure"): worth saying plainly that **this proposal is scoped to the question "what would confidential transfers look like done natively?"** — it does not claim the native path is the only valid one for Cardano. Smart-contract, escrow-based, and application-layer approaches are legitimate alternatives with their own trade-off profiles, and deserve their own analysis (some of that discussion belongs to your #8, where I'll respond separately). Within the native scope, though, the curve analysis above is why ristretto255 wins on its own merits.

perturbing:
Disclaimer: I did not read the CIP in full.

Blocks on Cardano need to propagate within a certain time delta across the network for liveness.

Do you have an indication, or even benchmarks, on the time consumed by the extra work of verifying these transactions?

I think, based on my intuition and experience, that it might be the case that this system can never be verified on the current mainnet average hardware. Verifying ZK is expensive compared to plain ed25519 signatures.

Perhaps you can add some indication of this to the CIP :)

matiwinnetou:
> **3. On cross-curve ZK composability**

Technically accurate — representing ristretto255 group arithmetic inside a BLS12-381 R1CS is prohibitively expensive (non-native field arithmetic), and no efficiency trick changes that. But the premise worth examining is that amount-predicates would be SNARK circuits at all. The anticipated design ([extensions.md](https://github.com/matiwinnetou/CIPs/blob/confidential-transfers/CIP-confidential-transfers/extensions.md) item 7) is different:

1. **Predicates are sigma protocols, not circuits.** "Hidden amount ≥ X" is a shifted-commitment range statement (`C − X·H`, the same technique the base proposal already uses for minimum-ADA); proportional relations are Schnorr-style linear proofs. These operate *directly in the commitment group* — no circuit, no R1CS, no cross-curve representation problem. They are verified by future ristretto255 builtins, which is the ecosystem dependency — named and priced as extensions.md item 4 (medium effort, no new cryptography; the CIP-0381 playbook for a different curve).
2. **So: a conscious trade-off, with the dependency explicit** — not a bet on tooling that doesn't exist. Sigma-protocol verification over ristretto is a few scalar multiplications; it needs builtins, not a proving framework.
3. **Where the boundary genuinely bites**: composing an amount-predicate with a *large* BLS-side circuit (say, a Groth16 credential proof) into a **single** proof. There the honest answer is: don't — issue two proofs sharing public inputs (the commitment as a public input to both sides), which is the standard pattern, cheap, and loses nothing except single-proof aesthetics. Anyone insisting on one proof pays the curve-bridging cost, and that is a real limitation worth acknowledging.
4. Also worth noting the governance case specifically doesn't touch this: homomorphic tally opening is aggregate Pedersen openings and (in committee variants) threshold decryption — group arithmetic again, no circuits anywhere.

The "why ristretto255" rationale now states all of this explicitly — the cross-curve boundary, the sigma-protocol resolution, the builtins dependency, and the single-proof composition limitation (commit c1fe9ce).

matiwinnetou:
Thanks @perturbing — a cryptographer's eyes on the verification-cost question is exactly the review this proposal needs, and the block-propagation deadline is precisely the right constraint to push on. I'd genuinely value your read of the full document — in particular the Appendix section **"Estimated node resource impact"**, which exists to answer exactly this question, and I'd welcome your scrutiny of its assumptions.

The short version, and why I believe the intuition transfers from a different ZK regime than the one used here:

1. **There are no SNARKs, no pairings, and no circuits anywhere in this proposal.** The verification workload is sigma-protocol-class: an aggregated **Bulletproofs** range proof (the dominant cost — ~2–3 ms per typical 2-in/2-out transaction on one modern core, ~1 ms amortised with cross-transaction batching) plus one **Schnorr** excess signature per asset (~0.2 ms) — plain group arithmetic over ristretto255, the same class of work as signature checking, at roughly **10–20× a transparent transaction**. Heavy-ZK intuitions (pairing-based verification, circuit provers) don't map onto this profile.
2. **Worst case against the propagation budget**: a block filled entirely with confidential transactions (~60 at current sizes) adds on the order of **0.1–0.2 s** of verification unbatched — about half with batch verification — and the proofs are independent, so verification parallelises trivially across transactions. That is a modest fraction of the diffusion budget, not a new bottleneck class.
3. **Production precedent on weaker hardware than SPO spec**: every Monero transaction has carried Bulletproofs verified by every node — largely consumer hardware — since 2018. "Cannot be verified on average mainnet hardware" would have to explain seven years of that network's operation.
4. **The honest caveat, agreed in advance**: the appendix figures are order-of-magnitude, single-core estimates derived from published benchmarks of the primitives. That is why Path to Active makes **measured benchmarks** — per transaction, per block, batched and unbatched, including full-sync replay impact — an explicit acceptance criterion, and why `maxConfidentialCommitmentsPerBlock` is an updatable protocol parameter: the per-block verification budget is governance-tunable without a hard fork if measurements demand it.

Separately, worth knowing: there is a **distinct workstream exploring confidential transfers purely at the smart-contract layer** — no ledger changes and no new cryptographic primitives, using existing Plutus builtins. The two tracks are complementary rather than competing: this CIP answers "what would this look like done natively, with the ledger itself as verifier", and the comparison between the approaches is one the community should make explicitly.

If you can spare the time for a deeper pass, the primitives section and the appendix are the two places where your expertise would sharpen this most — including telling me where the estimates are wrong.

matiwinnetou:
> **8. The CIP creates a dual-system complexity burden on the entire ecosystem.**

This is the most important question of the eight, and it deserved a better answer than an assertion in the Abstract — so the proposal now argues it explicitly: see the new first Rationale subsection, [**"Why native (ledger-level) rather than application-layer"**](https://github.com/matiwinnetou/CIPs/blob/confidential-transfers/CIP-confidential-transfers/README.md#why-native-ledger-level-rather-than-application-layer) (commit fa93fb6). The short form of the position:

1. **The framing I'd push back on is "vs."** This CIP is scoped to the question *"what would confidential transfers look like done natively, with the ledger itself as the verifier?"* — it does not claim the native path is the only valid one for Cardano. An application-layer track (no ledger changes, existing builtins) is a legitimate parallel workstream, and the choice between the approaches — or the decision to pursue both — belongs to the community. What the new subsection does is make the comparison explicit instead of implied.
2. **On the substance, what native buys**: one chain-wide standard instead of per-application fragmentation (one commitment format, one key model, one auditor integration, full interoperability); consensus-grade trust with no contract administrator or upgrade key; confidentiality of **ADA and native tokens as such** rather than of wrapped representations inside one contract; and the uniform, homomorphically-summable commitment substrate that the anticipated extensions (staking, governance, script predicates) structurally require — per-application formats cannot be summed or attributed chain-wide.
3. **On the Gate-Script blast-radius argument** — "a bug affects one application, not every confidential transaction": true, and it cuts both ways. The application-layer path doesn't remove the complexity, it *distributes* it: every application builds its own confidentiality with its own audit surface, its own key management, its own disclosure tooling — N systems to trust instead of one specification with test vectors and an independent audit as hard acceptance criteria. Empirically, amount confidentiality shipped at the application layer on other chains has remained a niche wrapper rather than becoming an ecosystem standard.
4. **The dual-system burden is real and the proposal says so** — it is the honest cost of the native path (bounded by opt-in design, versioned structures, and the normative guarantees section). Which cost profile Cardano prefers — one specification everyone implements, or per-application systems each carrying their own — is exactly the judgement this review process exists to make.

matiwinnetou:
@fallen-icarus coming back to this with a fuller answer, because your modularity argument deserves one — and the proposal has evolved since (registration certificates, native-script/multisig support, an explicit extensions analysis).

I'd now put the distinction more precisely than my first reply. There are two different privacy products here:

1. **Transfer privacy** — hiding *how much* moves and is held, while the ledger still verifies conservation. That is this CIP, and nothing more.
2. **Computation privacy** — hiding contract *state and logic inputs*, with ZK proofs that execution was correct. That is Midnight's actual core (ZK-first VM, private smart contracts), and this CIP — including every anticipated extension — structurally cannot do it: datums, script logic, and contract state stay public here forever; our ZK only ever speaks about committed amounts.

Where I'd push back on the "VPN for capital" pattern specifically: bridge out → disappear → bridge back is **graph-breaking** — it deliberately severs the link between source and destination of funds. That is a different (and regulatorily much harder) product than what this CIP builds, and the difference is the whole point: here the transaction graph, addresses, and asset identities stay public **by design**, amounts are recoverable by anyone the owner hands the viewing key to (tax authority, auditor), and disclosure is verifiable against an on-chain registration. Confidentiality from the public, not from oversight. A round-trip through any external system also reveals amounts at both boundary crossings, adds bridge trust and a wrapped representation, and puts an ordinary compliance-minded business — payroll, B2B invoicing, treasury — in the position of explaining why its funds route through an anonymity layer. For that audience, "confidential in place, auditable on demand" is not a niche the partner chain already serves; it's a different product the base ledger is the right place for.

On modularity as an architectural principle — I agree with it, and I think this split *is* the modular answer: the layer that owns the value handles *value* privacy (amounts are a property of UTxOs; hiding them belongs to the thing that validates UTxOs), and the specialized chain keeps what genuinely needs a different execution model — private computation. The two compose rather than compete: confidential base-layer value can flow to wherever deeper privacy programming lives, and my understanding is that this division of labour is broadly how people close to Midnight see the long run as well, per my earlier comment.

The proposal's Rationale now argues the native-vs-application-layer choice explicitly, and [extensions.md](https://github.com/matiwinnetou/CIPs/blob/confidential-transfers/CIP-confidential-transfers/extensions.md) records what the native path can and cannot ever grow into — including marking full graph-hiding as an explicit non-goal. If the modularity concern is "don't burden the base ledger with what a partner chain does better," I believe the document now shows the boundary is drawn in exactly that spirit.

PR #1238, opened 2026-08-03
CIP-0195? | Data Encoding of Plutus V4 Ledger Types

zliu41:
This CIP specifies how the Plutus V4 ledger API types, including script context and the types it mentions, are to be encoded as `Data` objects. New UPLC primitives and language features, together with lessons learned from Plutus V1-V3, create an opportunity to adopt a simpler and more efficient encoding scheme.

---

([latest version rendered from branch](https://github.com/zliu41/CIPs/blob/v4/CIP-0195/README.md))

### DIFF
--- a/CIP-0195/README.md
+++ b/CIP-0195/README.md
@@ -28,7 +28,7 @@ Array-based encoding may sound like the automatic winner, but that's not the cas
 
 __Addressing known inefficiencies.__
 For example, in Plutus V1-V3 every product type is encoded with a redundant `Constr 0` tag, where a plain `List` could have been used.
-Small nested sum types such as `Maybe StakingCredential` require multiple layers of `Constr`, even though they could have been encoded more compactly using one `Constr`.
+Small nested types also require more layers of `Constr` than necessary: reaching a bound of an `Interval` requires peeling off two layers of `Constr`, and reaching the credential inside a `Maybe StakingCredential` also requires two, even though one layer would suffice in each case.
 
 __Providing a better specification.__
 In Plutus V1-V3 the `Data` encoding of ledger API types was not formally specified.
@@ -88,16 +88,18 @@ Doing so avoids the overhead of pair-destruction.
 
 __Nested sum types.__
 Nested sum types can sometimes be flattened to reduce encoding and decoding overhead.
-For example, `Maybe StakingCredential` combines two two-constructor sum types, and its `Data` representation requires two layers of `Constr`.
-Replacing it with a single three-constructor sum type removes one layer and improves efficiency.
+For example, the V1-V3 `Address` type has a `Maybe StakingCredential` field, which combines two two-constructor sum types, so its `Data` representation requires two layers of `Constr`.
+Replacing it with a single three-constructor sum type would remove one layer and improve efficiency.
 
 This should, however, not be done blindly since there are drawbacks:
 - it could cause constructor proliferation if we flatten a large nested sum type;
 - it could also cause datatype proliferation if we need to keep the original nested type because it is used independently.
   For example, if `StakingCredential` is used independently (not under `Maybe`), then we may need to keep both `StakingCredential`, as well as the flattened `Maybe StakingCredential`.
 
-Given these considerations, it appears that the only suitable candidate for flattening in the [originally proposed script context definition](https://github.com/IntersectMBO/plutus/issues/7342#issuecomment-4799132923) is `Maybe StakingCredential`.
-Since `StakingCredential` always appears under `Maybe`, we can replace `Maybe StakingCredential` with a flattened datatype without also retaining the original `StakingCredential` type.
+_We did not identify any nested sum type in the Plutus V4 types that is worth flattening._
+`Maybe StakingCredential` was the one clear candidate in the script context of Plutus V1-V3, because `StakingCredential` only ever appear under `Maybe`, and so could be flattened without retaining the original type.
+But `StakingCredential` no longer exists in Plutus V4, whose `Address` type now carries a `Maybe AccountId` instead of `Maybe StakingCredential` (see the type definitions below).
+That candidate is therefore gone, and every remaining nested sum type in the V4 types runs into one of the two drawbacks above.
 
 __Nested product types.__
 Nested product types can likewise be flattened into a single larger product type.
@@ -200,6 +202,54 @@ data TopTxInfo = TopTxInfo
   -- top level transaction all concatenated together with loss of some information
   }
 
+data TopTxInfoSimplified = TopTxInfoSimplified
+  { ttisIds :: [V3.TxId]
+  -- ^ List of all `TxId`s fro the whole transaction, including the top-level transaction, which is
+  -- always going to be the last one in the list.
+  , ttisInputs :: [TxInInfo]
+  -- ^ Concatenated list of all `txInfoInputs`'
+  , ttisReferenceInputs :: [TxInInfo]
+  -- ^ Concatenated list of all `txInfoReferenceInputs`'
+  , ttisOutputs :: [V2.TxOut]
+  -- ^ Concatenated list of all `txInfoOutputs`'
+  , ttisMints :: V3.MintValue
+  -- ^ `MintValue`s from all `txInfoMint` with positive amounts
+  , ttisBurns :: V3.MintValue
+  -- ^ `MintValue`s from all `txInfoMint` with negative amounts
+  , ttisTxCerts :: [TxCert]
+  -- ^ Concatenated list of all `ttisTxCerts`'. Note, that unlike individual lists in each
+  -- `txInfoTxCerts`, this one can contain duplicates.
+  , ttisWithdrawals :: Map V2.Credential V2.Lovelace
+  -- ^ Union of all `txInfoWithdrawals` with a sum on the range for duplicate credentials
+  , ttisDirectDeposits :: Map V2.Credential V2.Lovelace
+  -- ^ Union of all `txInfoWithdrawals` with a sum on the range for duplicate credentials
+  , ttisValidRange :: V2.POSIXTimeRange
+  -- ^ Intersection of all validity intervals from within the whole transaction
+  , ttisGuards :: [V2.Credential]
+  -- ^ Concatenated list of all `txInfoGuards`'
+  , ttisRequiredTopLevelGuards :: Map V2.Credential ()
+  -- ^ Deduplicated set of required top level guards. It is impossible to keep the range of the Map
+  -- due to potential presence of duplicates in the domain between different sub-transactions,
+  -- therefore the range is eliminated.
+  , ttisScriptPurposes :: Map ScriptPurpose ()
+  -- ^ Union of all of the `Redeemer`s. Note that it is not possible to preserve actual `Redeemer`s
+  -- upon `union` operation due to potential duplicates in the domain. Therefore it is collapsed to
+  -- a Set of `ScriptPurpose`s only with duplicates removed.
+  , ttisData :: Map V2.DatumHash V2.Datum
+  -- ^ Union of all `txInfoData`. Duplicates are simply removed, since domain and range are
+  -- a one-to-one mapping.
+  , ttisVotes :: Map Voter (Map GovernanceActionId Vote)
+  -- ^ Union of all of the votes. Note that a vote in a sub-sequent sub-transaction or a top level
+  -- transaction can replace a vote from a prior sub-transaction.
+  , ttisProposalProcedures :: [ProposalProcedure]
+  -- ^ Concatenated list of all `ProposalPrecedure`s.
+  , ttisCurrentTreasuryAmount :: Haskell.Maybe V2.Lovelace
+  -- ^ Value of the treasury, which will be present if any of sub-transactions or top level
+  -- transaction included such value
+  , ttisTreasuryDonations :: V2.Lovelace
+  -- ^ Sum of all `txInfoTreasuryDonation`s
+  }
+
 data Address = Address
   { addressCredential :: Credential
   -- ^ the payment credential
@@ -207,24 +257,19 @@ data Address = Address
   -- ^ the staking credential (flattened)
   }
 
--- flattend from `Maybe StakingCredential`
-data StakingCredential
-  = {-| The staking hash is the `Credential` required to unlock a transaction output. Either
-    a public key credential (`Crypto.PubKeyHash`) or
-    a script credential (`ScriptHash`). Both are hashed with /BLAKE2b-244/. 28 byte. -}
-    StakingHash Credential
-  | {-| The certificate pointer, constructed by the given
-    slot number, transaction and certificate indices.
-    NB: The fields should really be all `Word64`, as they are implemented in `Word64`,
-    but 'Integer' is our only integral type so we need to use it instead. -}
-    StakingPtr
-      Integer
-      -- ^ the slot number
-      Integer
-      -- ^ the transaction index (within the block)
-      Integer
-      -- ^ the certificate index (within the transaction)
-  | NoStakingCredential
+data TxInInfo = TxInInfo
+  { txInInfoOutRef   :: V3.TxOutRef
+  , txInInfoResolved :: TxOut
+  }
+
+data TxOut = TxOut
+  { txOutAddress         :: Address
+  , txOutValue           :: Value -- now a built-in type
+  , txOutDatum           :: V2.OutputDatum
+  , txOutReferenceScript :: Maybe V2.ScriptHash
+  }
+
+newtype AccountId = AccountId V2.Credential
 
 -- flattened from the `Interval` in V1-V3
 data Interval a = Interval
--- a/CIP-0195/README.md
+++ b/CIP-0195/README.md
@@ -48,7 +48,7 @@ __Keeping the `Data` type simple__.
 If the Plutus V4 types are encoded using lists, no change to the `Data` datatype will be required.
 If they are encoded using arrays, then we'll need two additional constructors - `Array` and `ConstrArray` - together with four additional built-ins for wrapping and unwrapping.
 Adding new constructors to `Data` would be a quite invasive change for the Plutus repository alone.
-It would also complicates other languages, compilers andy tooling, and requires substantial ledger integration work.
+It would also complicates other languages, compilers and tooling, and requires substantial ledger integration work.
 The overall cost is significant, and it is a change best avoided.
 
 __Lists are usually faster than arrays for working with datatypes__:
--- a/CIP-0195/README.md
+++ b/CIP-0195/README.md
@@ -289,7 +289,7 @@ The encoding rules are:
 2. __Newtypes.__
    Newtypes are transparent: a newtype encodes the same as its underlying type.
 3. __Product types__.
-   Product types encode as `List [x1...xn]`, where `xi` is the encoding of the `i`-th field.
+   Product types encode as `List [x1, ..., xn]`, where `xi` is the encoding of the `i`-th field.
    This differs from V1-V3, which encodes product types using `Constr 0`.
 4. __Sum types.__
    Sum types encode as `Constr t [x1, ..., xn]`, where `t` is the constructor's 0-based index, and `xi` is the encoding of the `i`th constructor field.
--- a/CIP-0195/README.md
+++ b/CIP-0195/README.md
@@ -97,7 +97,7 @@ This should, however, not be done blindly since there are drawbacks:
   For example, if `StakingCredential` is used independently (not under `Maybe`), then we may need to keep both `StakingCredential`, as well as the flattened `Maybe StakingCredential`.
 
 _We did not identify any nested sum type in the Plutus V4 types that is worth flattening._
-`Maybe StakingCredential` was the one clear candidate in the script context of Plutus V1-V3, because `StakingCredential` only ever appear under `Maybe`, and so could be flattened without retaining the original type.
+`Maybe StakingCredential` was the one clear candidate in the script context of Plutus V1-V3, because `StakingCredential` only ever appears under `Maybe`, and so could be flattened without retaining the original type.
 But `StakingCredential` no longer exists in Plutus V4, whose `Address` type now carries a `Maybe AccountId` instead of `Maybe StakingCredential` (see the type definitions below).
 That candidate is therefore gone, and every remaining nested sum type in the V4 types runs into one of the two drawbacks above.
 
--- a/CIP-0195/README.md
+++ b/CIP-0195/README.md
@@ -271,15 +271,14 @@ data TxOut = TxOut
 
 newtype AccountId = AccountId V2.Credential
 
--- flattened from the `Interval` in V1-V3
-data Interval a = Interval
-  { ivFrom        :: Extended a
-  , ivFromClosure :: Closure
-  , ivTo          :: Extended a
-  , ivToClosure   :: Closure
-  }
+type Interval = POSIXTimeRange
 
-type POSIXTimeRange = Interval POSIXTime
+data POSIXTimeRange = POSIXTimeRange
+  { fromInclusive :: Maybe POSIXTime
+    -- ^ Nothing means negative infinity
+  , untilExclusive :: Maybe POSIXTime
+    -- ^ Nothing means positive infinity
+  }
 ```
 
 The encoding rules are:
--- a/CIP-0195/README.md
+++ b/CIP-0195/README.md
@@ -48,7 +48,7 @@ __Keeping the `Data` type simple__.
 If the Plutus V4 types are encoded using lists, no change to the `Data` datatype will be required.
 If they are encoded using arrays, then we'll need two additional constructors - `Array` and `ConstrArray` - together with four additional built-ins for wrapping and unwrapping.
 Adding new constructors to `Data` would be a quite invasive change for the Plutus repository alone.
-It would also complicates other languages, compilers and tooling, and requires substantial ledger integration work.
+It would also complicate other languages, compilers, and tooling, and require substantial ledger integration work.
 The overall cost is significant, and it is a change best avoided.
 
 __Lists are usually faster than arrays for working with datatypes__:
--- a/CIP-0195/README.md
+++ b/CIP-0195/README.md
@@ -292,7 +292,7 @@ The encoding rules are:
    Product types encode as `List [x1...xn]`, where `xi` is the encoding of the `i`-th field.
    This differs from V1-V3, which encodes product types using `Constr 0`.
 4. __Sum types.__
-   Sum types encode as `Constr t [x1...xn]`, where `t` is the constructor's 0-based index, and `xi` is the encoding of the `i`th constructor field.
+   Sum types encode as `Constr t [x1, ..., xn]`, where `t` is the constructor's 0-based index, and `xi` is the encoding of the `i`th constructor field.
    Constructors are indexed in the order in which they appear in the above Haskell datatypes.
 
 ## Rationale: How does this CIP achieve its goals?
--- a/CIP-0195/README.md
+++ b/CIP-0195/README.md
@@ -44,12 +44,12 @@ We therefore propose to use `List`, not `Array`, to encode the Plutus V4 types.
 Here are the reasons.
 
 __Keeping the `Data` type simple__.
-Plutus V4 will introduce a new definition of `Data` with an additional constructor, `Value`, containing a builtin `Value` (CIP-0153).
-Correspondingly, `valueData` and `unValueData` built-ins will be added for wrapping and unwrapping.
 
-If the Plutus V4 types are encoded using lists, no further addition to `Data` will be required.
+If the Plutus V4 types are encoded using lists, no change to the `Data` datatype will be required.
 If they are encoded using arrays, then we'll need two additional constructors - `Array` and `ConstrArray` - together with four additional built-ins for wrapping and unwrapping.
-Adding more constructors to `Data` would also make the `chooseData` built-in more complex.
+Adding new constructors to `Data` would be a quite invasive change for the Plutus repository alone.
+It would also complicates other languages, compilers andy tooling, and requires substantial ledger integration work.
+The overall cost is significant, and it is a change best avoided.
 
 __Lists are usually faster than arrays for working with datatypes__:
 
@@ -109,20 +109,20 @@ As with sum types, this should be done selectively, especially if we use list-en
   For example, suppose type `A` contains a field of type `B`.
   If we flatten `B` into `A`, producing a flattened type `A'`, reconstructing `B` from `A'` becomes much more difficult and costly than extracting `B` directly from `A`.
 
-Considering these trade-offs, we identified only one case where flattening a nested product type is clearly worthwhile: the `Interval` type.
-It is currently defined as:
+Considering these trade-offs, we found only one clear case where flattening a nested product type is beneficial: the unnecessarily nested `Interval` type in Plutus V1-V3.
+
+Plutus V4, however, will no longer use `Interval`; it will use the following type instead:
 
 ```haskell
-data Interval a = Interval {ivFrom :: LowerBound a, ivTo :: UpperBound a}
-data LowerBound a = LowerBound (Extended a) Closure
-data UpperBound a = UpperBound (Extended a) Closure
+data POSIXTimeRange = POSIXTimeRange
+  { fromInclusive :: Maybe POSIXTime
+    -- ^ Nothing means negative infinity
+  , untilExclusive :: Maybe POSIXTime
+    -- ^ Nothing means positive infinity
+  }
 ```
 
-This is unnecessarily nested, so we propose to flatten it to:
-
-```haskell
-data Interval a = Interval (Extended a) Closure (Extended a) Closure
-```
+We do not believe any other nested product type in the Plutus V4 script context warrants flattening.
 
 ### Plutus V4 script context and encoding rules
 
--- a/CIP-plutus-v4-encoding/README.md
+++ b/CIP-plutus-v4-encoding/README.md
@@ -6,8 +6,7 @@ Status: Proposed
 Authors:
     - Ziyang Liu <ziyang.liu@iohk.io>
 Implementors:
-    - Plutus team
-    - Developers of languages targeting UPLC
+    - IOG Plutus Team <https://iohk.io>
 Discussions:
     - https://github.com/IntersectMBO/plutus/issues/7342
 Created: 2026-08-03
--- a/CIP-plutus-v4-encoding/README.md
+++ b/CIP-plutus-v4-encoding/README.md
@@ -285,7 +285,9 @@ Also, under this approach, a product type and a sum type could have the same `Da
 
 ### Implementation Plan
 
-This should be implemented by the Plutus team as well as author of other languages targeting UPLC, and related tooling.
+- [ ] Implementation by the Plutus team
+- [ ] Implementation by any authors of other languages targeting UPLC
+- [ ] Implementation in related tooling
 
 ## Copyright
 
--- a/CIP-plutus-v4-encoding/README.md
+++ b/CIP-plutus-v4-encoding/README.md
@@ -250,7 +250,7 @@ The encoding rules are:
    Sum types encode as `Constr t [x1...xn]`, where `t` is the constructor's 0-based index, and `xi` is the encoding of the `i`th constructor field.
    Constructors are indexed in the order in which they appear in the above Haskell datatypes.
 
-### Alternatives considered
+## Rationale: How does this CIP achieve its goals?
 
 The previous section already explains why we prefer list-encoding to array-encoding.
 We also considered the following alternatives along other dimensions.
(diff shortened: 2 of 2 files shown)


zliu41:
cc @colll78 @KtorZ @michele-nuzzi @lehins @yHSJ

CIP-plutus-v4-encoding/README.md
```diff
@@ -0,0 +1,292 @@
+---
+CIP: "?"
+Title: Data Encoding of Plutus V4 Ledger Types
+Category: Plutus
+Status: Proposed
+Authors:
+    - Ziyang Liu <ziyang.liu@iohk.io>
+Implementors:
+    - Plutus team
+    - Developers of languages targeting UPLC
```
rphair:
```suggestion
Implementors:
    - IOG Plutus Team <https://iohk.io>
```
This is the first of the sections that caused our CIP validator to fail: https://github.com/cardano-foundation/CIPs/actions/runs/30840445337/job/91775891099?pr=1238 — on entries unqualified with either a personal/project email address or a domain name.
1. `https://iohk.io` may not be very specific either, but it's been used in the most recently published Plutus CIP that has anything other than `N/A` in this field ([CIP-0133](https://github.com/cardano-foundation/CIPs/tree/master/CIP-0133)).
2. If the downstream language developers are not known yet, there is therefore no way of listing them here.

CIP-plutus-v4-encoding/README.md
```diff
@@ -0,0 +1,292 @@
...
+Discussions:
+    - https://github.com/IntersectMBO/plutus/issues/7342
```
rphair:
```suggestion
Discussions:
    - Original PR: https://github.com/cardano-foundation/CIPs/pull/1238
    - Plutus repo discussion: https://github.com/IntersectMBO/plutus/issues/7342
```
The validator also requires that these fields be labelled (they generally appear horizontally so labels should be as short as possible).

CIP-plutus-v4-encoding/README.md
```diff
@@ -0,0 +1,292 @@
...
+Created: 2026-08-03
+License: CC-BY-4.0
+---
+
+## Abstract
+
+This CIP specifies how the Plutus V4 ledger API types, including script context and the types it mentions, are to be encoded as `Data` objects.
+New UPLC primitives and language features, together with lessons learned from Plutus V1-V3, create an opportunity to adopt a simpler and more efficient encoding scheme.
+
+## Motivation: why is this CIP necessary?
```
rphair:
```suggestion
## Motivation: Why is this CIP necessary?
```
Also a validation requirement (we have fixed the unsightly institutionalised capitalisation error everywhere else).

CIP-plutus-v4-encoding/README.md
```diff
@@ -0,0 +1,292 @@
...
+2. __Newtypes.__
+   Newtypes are transparent: a newtype encodes the same as its underlying type.
+3. __Product types__.
+   Product types encode as `List [x1...xn]`, where `xi` is the encoding of the `i`-th field.
+   This differs from V1-V3, which encodes product types using `Constr 0`.
+4. __Sum types.__
+   Sum types encode as `Constr t [x1...xn]`, where `t` is the constructor's 0-based index, and `xi` is the encoding of the `i`th constructor field.
+   Constructors are indexed in the order in which they appear in the above Haskell datatypes.
+
+### Alternatives considered
```
rphair:
```suggestion
## Rationale: How does this CIP achieve its goals?
```
Finally (as per the standard format) there was no literal _Rationale_ here but I think this section has everything we generally look for there.

CIP-plutus-v4-encoding/README.md
```diff
@@ -0,0 +1,292 @@
...
+## Path to Active
+
+### Acceptance Criteria
+
+- [ ] The encoding rules and revised types are implemented in `plutus-ledger-api`.
+- [ ] A new `Data` type is created in `plutus-ledger-api` for Plutus V4, which should contain the `Value` constructor.
+
+### Implementation Plan
+
+This should be implemented by the Plutus team as well as author of other languages targeting UPLC, and related tooling.
```
rphair:
```suggestion
- [ ] Implementation by the Plutus team
- [ ] Implementation by any authors of other languages targeting UPLC
- [ ] Implementation in related tooling
```
A broken-up list would give a chance to document these milestones in case the implementation happens progressively.

rphair (commented):
thanks @zliu41; we will `Triage` this (consider for CIP candidacy) at the next CIP meeting tomorrow: https://hackmd.io/@cip-editors/140

At this point for a CIP on the border somewhere between Plutus and Ledger, we would also confirm that the bulk of the work & effects would be in the `Plutus` domain as opposed to `Ledger` (re: categorisation).

cc @SeungheonOh @michaelpj

colll78:
The proposed type for `txInfoValidityRange` should instead be:
```haskell
data POSIXTimeRange = POSIXTimeRange
  { fromInclusive :: Maybe Integer
  , untilExclusive :: Maybe Integer
  }
  ```
  where
  ```haskell
Nothing in fromInclusive = negative infinity
Nothing in untilExclusive = positive infinity
```
  
The current type is not accurate with respect to the actual invariants. The ledger guarantees the lower bound is inclusive and the upper bound is exclusive, the type `Closure` for both lower and upper-bound implies incorrectly that the ledger can produce a validity range with a lower bound that is exclusive etc.

CIP-plutus-v4-encoding/README.md
```diff
@@ -0,0 +1,294 @@
...
+    slot number, transaction and certificate indices.
+    NB: The fields should really be all `Word64`, as they are implemented in `Word64`,
+    but 'Integer' is our only integral type so we need to use it instead. -}
+    StakingPtr
+      Integer
+      -- ^ the slot number
+      Integer
+      -- ^ the transaction index (within the block)
+      Integer
+      -- ^ the certificate index (within the transaction)
```
lehins: Ptr is gone from `PlutusV4`.
Which means we should simply get rid of  `StakingCredential` in favor of `Maybe AccountId` in `Address`

```haskell
data Address = Address
  { addressCredential :: Credential
  , addressStakingAccountId :: Maybe AccountId
  }
```

```suggestion
```

CIP-plutus-v4-encoding/README.md
```diff
@@ -0,0 +1,294 @@
+---
+CIP: "?"
```
rphair:
```suggestion
CIP: 195
```

rphair (commented):
@zliu41 the CIP meeting today declared this as a candidate: mainly due to its apparent completeness & clarity and the fact that it immediately proceeded to expert review about the particulars.

Thankfully @colll78 was there to explain the importance of the definitions in this CIP (it wasn't clear to me on first reading at all).  We confirmed the `Plutus` category based on the point of view while agreeing neither the Ledger nor Plutus sides of the implementation would be "hard"... the greatest difficulty & care would be "in the CIP itself" to ensure this component will be secure and efficient and meet all the requirements @lehins would have for the Ledger implementation and its invariants.

@colll78 also mentioned that user, cross-team, and dApp developer feedback will be important to confirm they can make their designs compatible with these definitions... @perturbing @Ryun1 we will have to look for opportunities to bring these parties into the discussion at appropriate stages in the review process.

In the meantime please @zliu41 rename the containing folder to `CIP-0195` and update the "latest version" link in your OP to reflect the new pathname. 🎉

CIP-plutus-v4-encoding/README.md
zliu41: Updated the datatypes in https://github.com/IntersectMBO/plutus/pull/7876

CIP-plutus-v4-encoding/README.md
```diff
@@ -0,0 +1,294 @@
...
+__Lists are usually faster than arrays for working with datatypes__:
+
+- For single-field access, `List` is almost always faster in practice.
+  Although `dropList` is linear and `indexArray` is constant-time, `dropList`'s cost model has a low intercept and slope, making list encoding faster than array encoding even when accessing the 20th field of a constructor.
+
+  This has taken into account that `indexArray` directly returns the requested field, while list-encoding requires a call to `headList` after `dropList`.
+  Even with this overhead, list-encoding remains faster for the 20th field.
+
+  It is true that if `k` is sufficiently large, accessing the `k`th field of a list can be arbitrarily more expensive than accessing the `k`th field of an array.
+  However, this is irrelevant since among the Plutus V4 types, the constructor with the most fields is `TxInfo`, which as 20 fields.
```
kwxm:
```suggestion
  However, this is irrelevant since among the Plutus V4 types, the constructor with the most fields is `TxInfo`, which has 20 fields.
```

CIP-plutus-v4-encoding/README.md
```diff
@@ -0,0 +1,294 @@
...
+Given these considerations, it appears that the only suitable candidate for flattening in the [originally proposed script context definition](https://github.com/IntersectMBO/plutus/issues/7342#issuecomment-4799132923) is `Maybe StakingCredential`.
+Since `StakingCredential` always appears under `Maybe`, we can replace `Maybe StakingCredential` with a flattened datatype without also retaining the original `StakingCredential` type.
+
+__Nested product types.__
+Nested product types can likewise be flattened into a single larger product type.
+As with sum types, this should be done selectively, especially if we use list-encoding rather than array-encoding, because
+- Flattening produces a longer list, making `dropList` more expensive
+- Flattening makes constructing the nested type harder.
+  For example, suppose type `A` contains a field of type `B`.
+  If we flatten `B` into `A`, producing a flattend type `A'`, reconstructing `B` from `A'` becomes much more difficult and costly than extracting `B` directly from `A`.
```
kwxm:
```suggestion
  If we flatten `B` into `A`, producing a flattened type `A'`, reconstructing `B` from `A'` becomes much more difficult and costly than extracting `B` directly from `A`.
```

kwxm (commented):
Corrections for a couple of typos.

CIP-plutus-v4-encoding/README.md
SeungheonOh: Most offchain libraries use the names `addressCredential` and `addressStakingCredential`. Now that we got rid of StakingPtr(finally!), I propose we remove `AccountId` wrapper type and use `Credential` directly like so:

```hs
data Address = Address
  { addressCredential :: Credential
  , addressStakingCredential :: Maybe Credential
  }
```

lehins: It is best to distinguish AccountId from other Credentials
```haskell
data Address = Address
  { addressCredential :: Credential
  , addressStakingAccountId :: Maybe AccountId
  }
```

zliu41: Updated.

CIP-0195/README.md
```diff
@@ -0,0 +1,339 @@
...
+Discussions:
+    - Original PR: https://github.com/cardano-foundation/CIPs/pull/1238
+    - Plutus repo discussion: https://github.com/IntersectMBO/plutus/issues/7342
+Created: 2026-08-03
+License: CC-BY-4.0
+---
+
+## Abstract
+
+This CIP specifies how the Plutus V4 ledger API types, including script context and the types it mentions, are to be encoded as `Data` objects.
```
kwxm, line 19: I felt that the CIP started abruptly, assuming that the reader already knows about the Plutus V4 types.  It might help to have an introductory sentence that says something like  "Plutus V4 will introduce new ledger API types, as discussed at <link to discussion> ...".  There's already a link to the discussion in the header, but that's easy to miss.

CIP-0195/README.md
```diff
@@ -0,0 +1,339 @@
...
+New UPLC primitives and language features, together with lessons learned from Plutus V1-V3, create an opportunity to adopt a simpler and more efficient encoding scheme.
+
+## Motivation: Why is this CIP necessary?
+
+__Choosing between lists and arrays.__
+UPLC now includes `Array` as a built-in type, with support for constant-time indexing.
```
kwxm, line 25: Maybe add a link to the PR or CIP for arrays?

CIP-0195/README.md
```diff
@@ -0,0 +1,339 @@
...
+## Specification
+
+### List vs. Array
+
+It may sound counter-intuitive, but we believe `List` is a clearly better choice than `Array` for encoding datatypes.
+We therefore propose to use `List`, not `Array`, to encode the Plutus V4 types.
+Here are the reasons.
+
+__Keeping the `Data` type simple__.
+Plutus V4 will introduce a new definition of `Data` with an additional constructor, `Value`, containing a builtin `Value` (CIP-0153).
```
kwxm, line 47: Arguably it's extending the current definition, not introducing a new one.

CIP-0195/README.md
```diff
@@ -0,0 +1,339 @@
...
+- For single-field access, `List` is almost always faster in practice.
+  Although `dropList` is linear and `indexArray` is constant-time, `dropList`'s cost model has a low intercept and slope, making list encoding faster than array encoding even when accessing the 20th field of a constructor.
+
+  This has taken into account that `indexArray` directly returns the requested field, while list-encoding requires a call to `headList` after `dropList`.
+  Even with this overhead, list-encoding remains faster for the 20th field.
+
+  It is true that if `k` is sufficiently large, accessing the `k`th field of a list can be arbitrarily more expensive than accessing the `k`th field of an array.
+  However, this is irrelevant since among the Plutus V4 types, the constructor with the most fields is `TxInfo`, which has 20 fields.
+
+- For multi-field access, array encoding can benefit from the `multiIndexArray` builtin.
```
kwxm, line 65: `multiIndexArray` isn't available on the chain yet: maybe link to the CIP?

CIP-0195/README.md
```diff
@@ -0,0 +1,339 @@
...
+  However, our testing indicates that array-encoding is still slower than list-encoding most of the time, and is faster only in corner cases (such as accessing a few widely-separated fields from a constructor with many fields).
```
kwxm, line 66: Is any of the testing publicly visible?  If we want to be strict about transparency then people should be able to check these claims.  I'm not going to demand that though.

kwxm: From the user's point of view, what matters is the budget, not the speed (although we'd hope that these are strongly correlated; also, the budget may change if we change the cost model).  Maybe we should spell out exactly what metric we're using here.

CIP-0195/README.md
```diff
@@ -0,0 +1,339 @@
...
+```haskell
+data Interval a = Interval {ivFrom :: LowerBound a, ivTo :: UpperBound a}
+data LowerBound a = LowerBound (Extended a) Closure
+data UpperBound a = UpperBound (Extended a) Closure
+```
+
+This is unnecessarily nested, so we propose to flatten it to:
+
+```haskell
+data Interval a = Interval (Extended a) Closure (Extended a) Closure
```
kwxm, line 124: Is this going to be a new type (so we have two versions), or can we replace the existing one with this?

CIP-0195/README.md
```diff
@@ -0,0 +1,339 @@
...
+```
+
+### Plutus V4 script context and encoding rules
+
+We specify the Plutus V4 ledger API types as Haskell datatypes plus encoding rules.
+The Haskell datatypes are shown below.
+The definitions of types reused from V1-V3 are omitted.
```
kwxm, line 131: You could always add a link to them in the Plutus repository to make is easier for the reader to find them.

CIP-0195/README.md
```diff
@@ -0,0 +1,339 @@
...
+  -- ^ the purpose of the currently-executing script, along with information associated
+  -- with the purpose
+  , scriptContextScriptHash :: ScriptHash
+  -- ^ Hash of the script that is being executed
+  }
+
+data TxInfo = TxInfo
+  { txInfoId :: V3.TxId
+  , txInfoSubTxIx :: Maybe Integer
+  , txInfoInputs :: [TxInInfo]
```
kwxm, line 149: The version in the repository [has](https://github.com/IntersectMBO/plutus/blob/2334b4e98f21653e3cdaf7ec25878e781ce7d5dc/plutus-ledger-api/src/PlutusLedgerApi/V4/Contexts.hs#L196)
```
  , txInfoFee :: V2.Lovelace
```
which doesn't appear here.  Which is correct?

CIP-0195/README.md
```diff
@@ -0,0 +1,339 @@
...
+### Plutus V4 script context and encoding rules
+
+We specify the Plutus V4 ledger API types as Haskell datatypes plus encoding rules.
+The Haskell datatypes are shown below.
+The definitions of types reused from V1-V3 are omitted.
+
+```haskell
+data ScriptContext = ScriptContext
+  { scriptContextTxInfo :: TxInfo
+  -- ^ Information about the transaction the currently-executing script is included in
```
kwxm, line 136: These comments (and many others later) aren't included in the version in the repository.  Maybe they should be, so they appear in the Haddock.

CIP-0195/README.md
```diff
@@ -0,0 +1,339 @@
...
+  , txInfoCurrentTreasuryAmount :: Haskell.Maybe V2.Lovelace
+  , txInfoTreasuryDonation :: V2.Lovelace
+  }
+
+data ScriptInfo
+  = MintingScript V2.CurrencySymbol
+  | SpendingScript V3.TxOutRef (Haskell.Maybe V2.Datum)
+  | WithdrawingScript AccountId
+  | CertifyingScript
+      -- | 0-based index of the given `TxCert` in `txInfoTxCerts`
```
kwxm, line 173: More comments that aren't in the real source code.

CIP-0195/README.md
```diff
@@ -0,0 +1,339 @@
...
+
+data TopTxInfo = TopTxInfo
+  { topTxInfoSubTransactions :: [TxInfo]
+  -- ^ List of `TxInfo`s for all sub-transactions. Not that `TxInfo` for the top level transaction
+  -- itslef is not present in this list.
+  , topTxInfoDatums :: Map TxId V2.Datum
+  -- ^ Datums supplied in `requiredTopLevelGuards` for that script. Plutus scripts require a datum
+  -- to be supplied when listed `requiredTopLevelGuards`. That `Map` will be empty if none of the
+  -- transactions within the whole transaction require Plutus scripts to be present in `Guards`
+  , topTxInfoStartingAccountBalanceIntervals :: AccountBalanceIntervals
```
kwxm, line 197: In the repository this is [called](https://github.com/IntersectMBO/plutus/blob/2334b4e98f21653e3cdaf7ec25878e781ce7d5dc/plutus-ledger-api/src/PlutusLedgerApi/V4/Contexts.hs#L266) `topTxInfoStartingBalanceIntervals`.

Also, more comments missing from the source code.

CIP-0195/README.md
```diff
@@ -0,0 +1,339 @@
...
+  , ttisTxCerts :: [TxCert]
+  -- ^ Concatenated list of all `ttisTxCerts`'. Note, that unlike individual lists in each
+  -- `txInfoTxCerts`, this one can contain duplicates.
+  , ttisWithdrawals :: Map V2.Credential V2.Lovelace
+  -- ^ Union of all `txInfoWithdrawals` with a sum on the range for duplicate credentials
+  , ttisDirectDeposits :: Map V2.Credential V2.Lovelace
+  -- ^ Union of all `txInfoWithdrawals` with a sum on the range for duplicate credentials
+  , ttisValidRange :: V2.POSIXTimeRange
+  -- ^ Intersection of all validity intervals from within the whole transaction
+  , ttisGuards :: [V2.Credential]
```
kwxm, line 228: This is `  , ttisGuards :: Map V2.Credential ()` in the real source code.

CIP-0195/README.md
```diff
@@ -0,0 +1,339 @@
...
+  -- ^ Value of the treasury, which will be present if any of sub-transactions or top level
+  -- transaction included such value
+  , ttisTreasuryDonations :: V2.Lovelace
+  -- ^ Sum of all `txInfoTreasuryDonation`s
+  }
+
+data Address = Address
+  { addressCredential :: Credential
+  -- ^ the payment credential
+  , addressStakingCredential :: StakingCredential
```
kwxm, line 256: The [repository says](https://github.com/IntersectMBO/plutus/blob/2334b4e98f21653e3cdaf7ec25878e781ce7d5dc/plutus-ledger-api/src/PlutusLedgerApi/V4/Address.hs#L77) `addressStakingAccountId :: Maybe AccountId`

CIP-0195/README.md
```diff
@@ -0,0 +1,339 @@
...
+data TxOut = TxOut
+  { txOutAddress         :: Address
+  , txOutValue           :: Value -- now a built-in type
+  , txOutDatum           :: V2.OutputDatum
+  , txOutReferenceScript :: Maybe V2.ScriptHash
+  }
+
+newtype AccountId = AccountId V2.Credential
+
+-- flattened from the `Interval` in V1-V3
```
kwxm, line 274: ... but `Extended` and `Closure` are still the things from `V1.Interval` presumably?

CIP-0195/README.md
```diff
@@ -0,0 +1,339 @@
...
+```
+
+The encoding rules are:
+
+1. __Built-in types.__
+   `Integer` encodes as `I`; `ByteString` as `B`; `Data` encodes as itself; `[]` encodes as `List`; `Map` encodes as `Map`; `Value` encodes as the new `Value` constructor of `Data` (this differs from V1-V3 where `Value` isn't a built-in type).
+2. __Newtypes.__
+   Newtypes are transparent: a newtype encodes the same as its underlying type.
+3. __Product types__.
+   Product types encode as `List [x1...xn]`, where `xi` is the encoding of the `i`-th field.
```
kwxm, line 292:
```suggestion
   Product types encode as `List [x1, ..., xn]`, where `xi` is the encoding of the `i`-th field.
```

CIP-0195/README.md
```diff
@@ -0,0 +1,339 @@
...
+   This differs from V1-V3, which encodes product types using `Constr 0`.
+4. __Sum types.__
+   Sum types encode as `Constr t [x1...xn]`, where `t` is the constructor's 0-based index, and `xi` is the encoding of the `i`th constructor field.
```
kwxm, line 295:
```suggestion
   Sum types encode as `Constr t [x1, ..., xn]`, where `t` is the constructor's 0-based index, and `xi` is the encoding of the `i`th constructor field.
```

CIP-0195/README.md
```diff
@@ -0,0 +1,339 @@
...
+For example, the V1-V3 `Address` type has a `Maybe StakingCredential` field, which combines two two-constructor sum types, so its `Data` representation requires two layers of `Constr`.
+Replacing it with a single three-constructor sum type would remove one layer and improve efficiency.
+
+This should, however, not be done blindly since there are drawbacks:
+- it could cause constructor proliferation if we flatten a large nested sum type;
+- it could also cause datatype proliferation if we need to keep the original nested type because it is used independently.
+  For example, if `StakingCredential` is used independently (not under `Maybe`), then we may need to keep both `StakingCredential`, as well as the flattened `Maybe StakingCredential`.
+
+_We did not identify any nested sum type in the Plutus V4 types that is worth flattening._
+`Maybe StakingCredential` was the one clear candidate in the script context of Plutus V1-V3, because `StakingCredential` only ever appear under `Maybe`, and so could be flattened without retaining the original type.
```
kwxm:
```suggestion
`Maybe StakingCredential` was the one clear candidate in the script context of Plutus V1-V3, because `StakingCredential` only ever appears under `Maybe`, and so could be flattened without retaining the original type.
```

CIP-0195/README.md
```diff
@@ -0,0 +1,339 @@
...
+  There is no nice way to encode the standalone `StakingCredential` differently than the `StakingCredential` in `Maybe StakingCredential`.
+  The most obvious solution is adding newtype wrappers, but this would be quite annoying and inconvenient.
+
+__Using bare `List`/`Array` to encode sum types__.
+We previously proposed that product types should be encoded using `List`, rather than `Constr`.
+We could do the same for sum types, where the first element of the `List` would represent the constructor tag.
+
+If we use list-encoding, there would be little reason to do this, since it won't be cheaper than using `Constr`.
+If we use array-encoding, then it would be slightly more appealing because it would allow us to avoid adding the `ConstrArray` constructor to `Data`.
+However, accessing the constructor tag would then require an `indexArray 0` call which is much more expensive than extracting the tag from `ConstrArray`.
```
kwxm, line 321: Is it really that much more expensive?  `indexArray` costs 232,010 ExCPU according to the current cost model, whereas `fstPair . unConstrData` costs 141,895+24,588 = 166,483, possibly plus a few CEK steps.

CIP-0195/README.md
```diff
@@ -0,0 +1,339 @@
...
+Also, under this approach, a product type and a sum type could have the same `Data` representation; it is unclear whether this could lead to problems.
+
+## Path to Active
+
+### Acceptance Criteria
+
+- [ ] The encoding rules and revised types are implemented in `plutus-ledger-api`.
+- [ ] A new `Data` type is created in `plutus-ledger-api` for Plutus V4, which should contain the `Value` constructor.
```
kwxm, line 329: Really? I thought we were going to extend the definition of `Data` in `plutus-core` with an extra field.  Do we have to do something in `plutus-ledger-api` as well?

kwxm (approved):
I don't have a lot to say about the substance of the proposal: my comments were mostly small quibbles about presentation.  There seem to be  number of discrepancies between the types specified here and the types currently implemented in`PlutusLedgerApi.V4` though

CIP-0195/README.md
rphair: [CIP-0138](https://github.com/cardano-foundation/CIPs/blob/master/CIP-0138/README.md) ([discussion](https://github.com/cardano-foundation/CIPs/pull/915))

CIP-0195/README.md
rphair: [CIP-0156](https://github.com/cardano-foundation/CIPs/tree/master/CIP-0156) ([discussion](https://github.com/cardano-foundation/CIPs/pull/1050))

CIP-0195/README.md
kwxm: In fact, I see that the [first comment](https://github.com/IntersectMBO/plutus/blob/2334b4e98f21653e3cdaf7ec25878e781ce7d5dc/plutus-ledger-api/src/PlutusLedgerApi/V4/Address.hs#L17) in `PlutusLedgerAPI.V4.Address` says

> In V4 staking credentials no longer exist

CIP-0195/README.md
colll78: See:
https://github.com/cardano-foundation/CIPs/pull/1238#issuecomment-5175083975

I imagine the same is true for AccountBalanceIntervals, the ledger will provide invariants around closures for that type and thus closures do not need to exist in the data representation.

CIP-plutus-v4-encoding/README.md
colll78: What is `AccountId`? Here we are concerned around `Data` encoding, in which case `Maybe Credential` and `Maybe AccountId` are fungible.

CIP-0195/README.md
zliu41: You are right, we should reuse the existing Data type rather than creating a new one.

CIP-0195/README.md
zliu41: Yeah, it's not "much more", just "more".

CIP-0195/README.md
zliu41: I did test the budget, not the speed, though @Unisay's cost model of multiIndexArray is still WIP. Do you think it is accurate enough?

CIP-plutus-v4-encoding/README.md
zliu41: > Maybe Credential and Maybe AccountId are fungible.

Not exactly.

If `Credential` both appears under `Maybe` and standalone, then I wouldn't flatten `Maybe Credential`, since we'd need to keep two separate datatypes, `FlattenedMaybeCredential` and `Credential`, which is at the very least inconvenient.

If `AccountId` only appears under `Maybe`, then it's fine to flatten it, since we don't need to keep `AccountId` any more. However, I don't think this is the case - @lehins could you confirm?

CIP-0195/README.md
kwxm: There's a potential difficulty with extending the existing `data` type because we currently have
```
 chooseDataDenotation :: Data -> a -> a -> a -> a -> a -> a
```
and we'll need an extra branch to handle `value`, and we'll also need to keep the old version for old scripts.  However, I think I experimented with this a while ago and it turned out that we could actually have different arities depending on the semantic variant, and everything worked OK, including the costing (although we maybe need to add a new cost model parameter at the end).

CIP-0195/README.md
kwxm: > @Unisay's cost model of multiIndexArray is still WIP. Do you think it is accurate enough?

It looks pretty good.

CIP-0195/README.md
zliu41: Yeah, I'm now leaning against adding the Value constructor to Data. See https://github.com/cardano-foundation/CIPs/pull/1090#issuecomment-5295762077

zliu41:
> The proposed type for txInfoValidityRange should instead be:

@colll78 updated. I used `Maybe POSIXTime` instead of `Maybe Integer`. I also removed the claim that the Value constructor will be added to the Data datatype.

colll78:
> > The proposed type for txInfoValidityRange should instead be:
> 
> @colll78 updated. I used `Maybe POSIXTime` instead of `Maybe Integer`. I also removed the claim that the Value constructor will be added to the Data datatype.

I'm highly confident that Leios will be delayed. I think it's premature to remove `Value` from Data.

CIP-0195/README.md
```diff
@@ -0,0 +1,338 @@
...
+  , txInfoReferenceInputs :: [TxInInfo]
+  , txInfoOutputs :: [V2.TxOut]
+  , txInfoMint :: Value -- now a built-in type
+  , txInfoTxCerts :: [TxCert]
+  , txInfoWithdrawals :: Map V2.Credential V2.Lovelace
+  , txInfoDirectDeposits :: Map V2.Credential V2.Lovelace
+  , txInfoAccountBalanceIntervals :: AccountBalanceIntervals
+  , txInfoValidRange :: V2.POSIXTimeRange
+  , txInfoGuards :: [V2.Credential]
+  , txInfoRequiredTopLevelGuards :: Map V2.Credential (Maybe Datum)
```
colll78, line 159: I don't understand what `(Maybe Datum)` is doing here. Shouldn't it just be:
```
txInfoRequiredTopLevelGuards :: [V2.Credential]
```

>-- ^ Datums supplied in `requiredTopLevelGuards` for that script. Plutus scripts require a datum
  -- to be supplied when listed `requiredTopLevelGuards`. That `Map` will be empty if none of the
  -- transactions within the whole transaction require Plutus scripts to be present in `Guards`

I don't understand the meaning of this. What datum? Where does this data live? If it is intended that top level guards convey messages / invariants to sub transactions via builtindata using the `requiredTopLevelGuards` field then the type should be `Redeemer` not `Datum`.

CIP-0195/README.md
colll78: Then the real source code is strictly worse and the relevant `toPlutusContext` instance should be corrected to translate to builtin list as defined here.

CIP-0195/README.md
```diff
@@ -0,0 +1,338 @@
...
+  -- `txInfoTxCerts`, this one can contain duplicates.
+  , ttisWithdrawals :: Map V2.Credential V2.Lovelace
+  -- ^ Union of all `txInfoWithdrawals` with a sum on the range for duplicate credentials
+  , ttisDirectDeposits :: Map V2.Credential V2.Lovelace
+  -- ^ Union of all `txInfoWithdrawals` with a sum on the range for duplicate credentials
+  , ttisValidRange :: V2.POSIXTimeRange
+  -- ^ Intersection of all validity intervals from within the whole transaction
+  , ttisGuards :: [V2.Credential]
+  -- ^ Concatenated list of all `txInfoGuards`'
+  , ttisRequiredTopLevelGuards :: Map V2.Credential ()
```
colll78, line 230: Why map?

This should be `[Credential]`. The map definition introduces tons of inefficiency for no benefit.

CIP-0195/README.md
```diff
@@ -0,0 +1,338 @@
...
+  -- ^ Deduplicated set of required top level guards. It is impossible to keep the range of the Map
+  -- due to potential presence of duplicates in the domain between different sub-transactions,
+  -- therefore the range is eliminated.
+  , ttisScriptPurposes :: Map ScriptPurpose ()
```
colll78, line 234: This should just be `[ScriptPurpose]`....

CIP-0195/README.md
lehins: This is morally a Set, that is why I defined it using a Map. However, I forgot that Plutus Map preserves duplicates anyways, so there is no point in using a Map

CIP-0195/README.md
zliu41: @lehins what do you think?

lehins: >  ... it is intended that top level guards convey messages / invariants to sub transactions via builtindata ...

Yes, that is precisely the purpose, to provide the ability for sub-tx builders to supply an argument to a top level guard. 

Redeemer is not suitable name, because redeemers contains execution units as well as data. 
Datum is used, because it us just an arbitrary plutus Data.

lehins: FTR, it is absolutely essential that this Datum is present there. Without it, same script listed in multiple sub transactions would look exactly the same at the top level. So, no, this would be totally wrong:
```
txInfoRequiredTopLevelGuards :: [V2.Credential]
```

zliu41:
@lehins just to confirm, there's no more `txInfoFee` in V4?

zliu41:
Regarding adding `Value` to `Data`, see https://github.com/IntersectMBO/plutus/pull/7910

CIP-0195/README.md
colll78: >Yes, that is precisely the purpose, to provide the ability for sub-tx builders to supply an argument to a top level guard.

> Redeemer is not suitable name, because redeemers contains execution units as well as data.
Datum is used, because it us just an arbitrary plutus Data.

Then the name used should be `BuiltinData` (a type which represents arbitrary builtin data) not `Datum` which represents `BuiltinData` stored in UTxOs. In Plutus, both `Redeemer` and `Datum` are just newtypes around `BuiltinData` no ex-units on either. 

From the perspective of Plutus contract developers, `Redeemer` has no ex-units, and just represents transient user provided arbitrary data in the script context, where-as `Datum` represents persistent user provided arbitrary data.

CIP-0195/README.md
```diff
@@ -0,0 +1,338 @@
...
+It may sound counter-intuitive, but we believe `List` is a clearly better choice than `Array` for encoding datatypes.
+We therefore propose to use `List`, not `Array`, to encode the Plutus V4 types.
+Here are the reasons.
+
+__Keeping the `Data` type simple__.
+
+If the Plutus V4 types are encoded using lists, no change to the `Data` datatype will be required.
+If they are encoded using arrays, then we'll need two additional constructors - `Array` and `ConstrArray` - together with four additional built-ins for wrapping and unwrapping.
+Adding new constructors to `Data` would be a quite invasive change for the Plutus repository alone.
+It would also complicates other languages, compilers andy tooling, and requires substantial ledger integration work.
```
lehins:
```suggestion
It would also complicates other languages, compilers and tooling, and requires substantial ledger integration work.
```

CIP-0195/README.md
```diff
@@ -0,0 +1,338 @@
...
+It may sound counter-intuitive, but we believe `List` is a clearly better choice than `Array` for encoding datatypes.
+We therefore propose to use `List`, not `Array`, to encode the Plutus V4 types.
+Here are the reasons.
+
+__Keeping the `Data` type simple__.
+
+If the Plutus V4 types are encoded using lists, no change to the `Data` datatype will be required.
+If they are encoded using arrays, then we'll need two additional constructors - `Array` and `ConstrArray` - together with four additional built-ins for wrapping and unwrapping.
+Adding new constructors to `Data` would be a quite invasive change for the Plutus repository alone.
+It would also complicates other languages, compilers and tooling, and requires substantial ledger integration work.
```
Unisay:
```suggestion
It would also complicate other languages, compilers, and tooling, and require substantial ledger integration work.
```

CIP-0195/README.md
colll78: I find this surprising, normally the more work that can be done within a builtin the faster it is by a huge margin even if the builtin is not micro-optimized for the specific usecase (ie. `dropList 26 xs` builtin is more efficient than tail inlined 26 times). I guess in this case it is because the result of this builtin is a list, and so to actually use these fields the user still has to perform a ton of non-builtin work to bind the elements of a list to variables they can interact with. 

If `multiIndexArray` worked as:
```
let indices = [1, 5, 7, 9] 
in multiIndexArray indices ourArray (\a b c d -> ...) 
```

(ie. the third argument was a continuation function which the indexed elements get applied to) then it most certainly would be a huge improvement over manual extraction.

zliu41: > I guess in this case it is because the result of this builtin is a list

Yes, that's exactly the reason.

PR #1242, opened 2026-08-07
CIP-0197? | Post-Quantum ZK Signatures for HD Wallets

paweljakubas:
CIP: Post-Quantum Zero-Knowledge Signatures for Cardano HD Wallets

## Summary

This PR adds a new CIP proposing an optional, additive post-quantum signing layer for Cardano HD wallets built on CIP-1852's BIP32-Ed25519 derivation, based on the ZKPoSP construction (ePrint 2026/1508), with a concrete proof system already validated end-to-end on Cardano.

## Problem

Cardano keys are Ed25519, broken by a quantum adversary via Shor's algorithm. A CIP-1852 wallet derives its entire key hierarchy from one root seed — recovering a single signing scalar lets an adversary forge transactions. Replacing the derivation scheme would invalidate every existing address and break hardware wallets, CIP-1854 multisig, and governance keys.

## Proposal

Keep keys, addresses, and derivation unchanged; accompany the classical Ed25519 signature with a zero-knowledge proof of knowledge of the seed witness behind the existing public key. Security reduces to the conjectured quantum hardness of SHA-256/SHA-512 and the soundness of the proof system — no migration, no re-registration, no address change.

## Why STARK is not viable

ZK-STARK (RISC Zero) is transparent and post-quantum, but produces proofs of **≈219 KB** — far above Cardano's transaction size limits and impractical for bandwidth-constrained wallets. Verification is ≈9–10 ms and signing ≈12.5 s per transaction, with a one-time derivation cost of ≈156 s. These numbers make STARK a useful research baseline, but **not a deployable solution** for a UTxO chain where every byte in the redeemer counts.

## The viable path: NovaSlim (transparent folding + sumcheck)

[NovaSlim](https://github.com/cardano-foundation/nova-slim) is a transparent, trusted-setup-free proof system that NIFS-folds step-circuit witnesses into a Relaxed-R1CS accumulator and compresses it with a sumcheck argument. It is already working end-to-end on Cardano with an Aiken/Plutus V3 on-chain verifier.

For a Cardano key-ownership step circuit (~7.7K constraints, 255 steps, BLS12-381):

| Metric | NovaSlim | STARK (RISC Zero) |
|---|---|---|
| Proof size | **~1.0 KiB** slim | ~219 KB |
| Verify time | **~0.7 ms** | ~9–10 ms |
| Fold/prove time | ~138 s total (~543 ms/step) | ~12.5 s (signing) / ~156 s (derivation) |
| Trusted setup | **None** | None |
| On-chain verifier | **Aiken/Plutus V3 (validated)** | Requires new native primitive |
| Fits in `maxTxSize` | **Yes** | No |

The slim proof strips commitment openings, keeping only the sumcheck protocol data. It is commitment-scheme-agnostic (Pedersen, SIS, or Hash) and verifies with native field operations only — no pairings, no curve cycles, no EVM precompiles.

## Deployment phases

**Phase 1 — off-chain, adoptable immediately (recommended):** the classical Ed25519 signature stays on-chain; NovaSlim proofs are verified out-of-band by wallets, exchanges, and indexers. No ledger change — quantum readiness now.

**Phase 2 — native sumcheck verifier in the node (future):** NovaSlim proofs become the on-chain witness, verified by a native (non-Plutus) sumcheck primitive. Requires a consensus change. The **~1.0 KiB proof size is already within** current transaction limits, so Phase 2 is gated only by ledger development, not by fundamental proof-size constraints.

## Key design points

- Proofs are anchored at the **hardened account node** — required because non-hardened leaves are forgeable from public keys (paper Prop. 6.3/7.2), and every Cardano leaf is non-hardened.
- **Split proofs:** a one-time derivation certificate per account plus a per-transaction signing proof, bound by a shared hash (`hkR`).
- **Backend-agnostic payload:** the CIP defines the witness relation (seed → BIP32-Ed25519 derivation → public key); the proof format is versioned so that the specific argument system (currently NovaSlim sumcheck) can evolve.

1. Should the CIP mandate NovaSlim as the reference implementation, or keep the payload format generic enough to admit future transparent schemes?
2. Native sumcheck verifier in the ledger — should the specification include a concrete Plutus V3 cost model, or defer to a future CIP?
3. Proof-format versioning and parameter handling — how should wallets negotiate backend versions?

---

([latest rendered version from branch](https://github.com/paweljakubas/CIPs/blob/paweljakubas/post-quantum-hierarchical-deterministic-wallets/CIP-0197/README.md))

### DIFF
--- a/CIP-0197/README.md
+++ b/CIP-0197/README.md
@@ -222,18 +222,21 @@ zkVM. The comparison and selection of these options is discussed in
    maturity — a first-of-its-kind Nova implementation, unproven at Cardano scale — so it
    is a candidate for Phase 1 evaluation alongside the STARK route, not the default
    baseline.
-6. **NovaSlim [NOVASLIM] — practical Phase 2 stepping stone.** A Nova-family IVC
-   system that produces **sub-kilobyte transparent proofs** (~0.4–2.5 KiB, independent
-   of commitment scheme) with **sub-millisecond verification** (~0.2 ms) and **no
-   trusted setup**. It is implemented in Rust with support for six curves (including
-   BLS12-381, Cardano's native curve) and three commitment schemes (Pedersen, SIS,
-   Hash) — of which **SIS** and **Hash** are post-quantum (conjectured). Critically, it already has a **working Aiken eUTXO verifier** running in
-   Plutus V3 on BLS12-381 scalar arithmetic. Its protocol design, security proofs, and
-   benchmark analysis are specified in the NovaSlim whitepaper [NOVASLIM-WP].
-   Reference figures (VRF circuit, BLS12-381, Pedersen, 254 steps): fold ~3.5 s,
-   compress ~0.04 s, verify slim ~0.2 ms, proof ~0.4 KiB. With SIS commitment
-   (m=128): fold is faster than Pedersen (matrix–vector products replace MSMs)
-   and the slim proof size stays ~0.4 KiB.
+6. **NovaSlim [NOVASLIM] — experimental Phase 2 stepping stone.** A Nova-family IVC
+    system that produces **sub-kilobyte transparent proofs** (~0.4–2.5 KiB, independent
+    of commitment scheme) with **sub-millisecond verification** (~0.2 ms) and **no
+    trusted setup**. It is implemented in Rust with support for six curves (including
+    BLS12-381, Cardano's native curve) and three commitment schemes (Pedersen, SIS,
+    Hash).  **Hash is operationally SIS/Ajtai in disguise** (Blake2b-derived matrix
+    entries), so the same norm-enforcement caveat applies to both; neither gives a
+    proven post-quantum guarantee without additional range proofs.  Critically, it
+    already has a **working Aiken eUTXO verifier** running in Plutus V3 on BLS12-381
+    scalar arithmetic. Its protocol design, security proofs, and benchmark analysis are
+    specified in the NovaSlim whitepaper [NOVASLIM-WP].
+    Reference figures (VRF circuit, BLS12-381, Pedersen, 254 steps): fold ~3.5 s,
+    compress ~0.04 s, verify slim ~0.2 ms, proof ~0.4 KiB.  **Honest sizing:**
+    a fully sound system would be ~5–12 KiB (classical) and ~30–100+ KiB
+    (post-quantum) once PCS openings, fold verification, and norm proofs are included.
 
 ### Deployment: two phases
 
@@ -280,10 +283,14 @@ primary deployment (Conway-era limits; updatable via governance):
 | Plutus `maxTxExecutionUnits` | 14M memory / 10G steps per tx | No STARK verifier exists in Plutus; verification is orders of magnitude beyond this budget |
 | Native scripts (V1–V3) | threshold multisig only | Cannot express arbitrary NIZK verification |
 
-**However, NovaSlim [NOVASLIM] already bypasses these constraints:** its slim proofs
-are **~0.4–2.5 KiB** (well under `maxTxSize`) and its Aiken verifier runs in **Plutus
-V3** on BLS12-381 scalar arithmetic. This makes NovaSlim the only backend with a
-working Phase 2 path today; see [Proving backend options](#proving-backend-options).
+**However, NovaSlim [NOVASLIM] already bypasses the size constraints:** its slim
+proofs are **~0.4–2.5 KiB** (well under `maxTxSize`) and its Aiken verifier runs in
+**Plutus V3** on BLS12-381 scalar arithmetic. This makes NovaSlim the only backend
+with a working Phase 2 *infrastructure* path today — the verifier checks
+sumcheck consistency and algebraic correctness, but does not yet evaluate the
+circuit MLEs or open commitments, so it is suitable for **fund-recovery**
+(trusted prover) rather than day-to-day untrusted signing; see [Proving backend
+options](#proving-backend-options).
 
 Raising `maxTxSize` alone would not suffice for STARKs (the receipt also exceeds
 `maxBlockBodySize`), and the post-Q-day variant does not reduce proof size — it only
@@ -300,10 +307,13 @@ only if the previous one is closed for some reason:
 2. **One proof across several txs.** If no single tx can hold the receipt, split it
    over multiple mutually-referencing transactions, so the fragment proofs together
    form the full receipt.
-3. **NovaSlim for Phase 2 now.** If neither STARK option fits, use NovaSlim's
-   sub-kilobyte proofs with the existing Aiken verifier. This is the only backend
-   that already works end-to-end on Cardano today, although it remains experimental
-   and unaudited.
+3. **NovaSlim for Phase 2 infrastructure now.** If neither STARK option fits, use
+    NovaSlim's sub-kilobyte proofs with the existing Aiken verifier for the
+    **fund-recovery** use case (trusted prover, verifier checks algebraic
+    consistency). This is the only backend that already works end-to-end on
+    Cardano today, although it remains experimental and unaudited and its
+    verifier does not yet establish full knowledge soundness under an untrusted
+    prover.
 4. **Out-of-band proofs + node-side cache.** If none of the above, proofs travel
    out-of-band; nodes keep a cache of proofs and only adopt blocks of signatures for
    which they have seen the corresponding zk proof. This needs additional handling of
@@ -390,7 +400,9 @@ is a research prototype lacking production maturity (Lova):
 | Plonky3 | 🔜 Requires implementation | ❌ No (~10–50 KB) | Early maturity |
 | Nova / groth16-prover | ❌ No (DLOG-based) | — | Classical only |
 | Lova | ✅ Conjectured | 🔜 Future (~5–10 KB) | Research prototype |
-| **NovaSlim** | ✅/🔜 Conjectured (SIS & Hash) | ✅ **Yes** (~0.4–2.5 KiB) | **Implemented (experimental, unaudited)** |
+| **NovaSlim** | ✅/🔜 Conjectured (SIS & Hash) | ✅ **Yes** (~0.4–2.5 KiB slim) | **Implemented (experimental, unaudited)** |
+
+Hash is operationally SIS/Ajtai in disguise; neither gives a proven PQ guarantee without norm enforcement. NovaSlim's honest sound size is ~5–12 KiB (classical) / ~30–100+ KiB (PQ).
 
 That narrows the field to **two usable ways forward**, which is why the comparison
 serves only a small number of decisions:
@@ -401,25 +413,27 @@ serves only a small number of decisions:
   `maxTxSize` budget), so it is confined to the Phase 1 off-chain path until a
   proof-size-reduction programme (ZK-friendly hashing, smaller FRI configurations)
   brings it within budget. It is the low-risk default for Phase 1.
-- **Way 2 — NovaSlim, the experimental backend, demonstrated to work.** Initially
-  designed as a classical Nova-family IVC, its **SIS** and **Hash** commitments are
-  post-quantum (conjectured), giving a PQ path today, and it is the *only* option with
-  sub-kilobyte proofs (~0.4–2.5 KiB), sub-millisecond verification, **and** a working
-  on-chain Aiken verifier (Plutus V3). The author has demonstrated the end-to-end
-  pipeline — step-circuit folding, slim-proof generation, and on-chain verification —
-  at `cardano/cip197` [NOVASLIM-CIP197]. Its caveats are that it is experimental and
-  unaudited, and that its post-quantum guarantee is as strong as the hash/SIS
-  parameters, not as strong as Lova's lattice construction (the underlying sumcheck
-  folding is classical, random-oracle / Fiat–Shamir based).
+- **Way 2 — NovaSlim, the experimental backend, demonstrated to work.** Its
+  **SIS** and **Hash** commitments are post-quantum (conjectured), giving a PQ path
+  today, and it is the *only* option with sub-kilobyte proofs (~0.4–2.5 KiB),
+  sub-millisecond verification, **and** a working on-chain Aiken verifier (Plutus V3).
+  The author has demonstrated the end-to-end pipeline — step-circuit folding,
+  slim-proof generation, and on-chain verification — at `cardano/cip197`
+  [NOVASLIM-CIP197].  Its caveats are that (a) it is experimental and unaudited,
+  (b) its verifier does not yet evaluate circuit MLEs or open commitments, so
+  acceptance attests honest execution rather than knowledge of a valid witness,
+  (c) Hash is operationally SIS/Ajtai in disguise and inherits the same
+  missing-norm-enforcement caveat, and (d) the underlying sumcheck folding is
+  classical (random-oracle / Fiat–Shamir based), not lattice-hardened like Lova.
 
 The CIP leans on **Way 2 (NovaSlim)** for the Phase 2 infrastructure path, because it
 solves the exact problem that blocks every other option: proof size, verification cost,
-and on-chain verifier availability — and it is already shown working. The recommended
-posture is to deploy NovaSlim **now** for Phase 2 infrastructure (wallet integration,
-witness format, on-chain verifier, transport) with a PQ commitment, and to upgrade the
-proving backend to a lattice-hardened variant (Lova-style) when that cryptography
-matures, keeping the same proof format and Aiken verifier. This mirrors the CIP's
-Phase 1→2 philosophy: build the road before the car is finished.
+and on-chain verifier availability — and it is already shown working.  The recommended
+posture is to treat NovaSlim as a **fund-recovery mechanism** (verified by a Plutus
+contract once classical signatures are disabled) rather than a day-to-day signing
+scheme, and to upgrade the proving backend to a lattice-hardened variant (Lova-style)
+when that cryptography matures, keeping the same proof format and Aiken verifier.
+This mirrors the CIP's Phase 1→2 philosophy: build the road before the car is finished.
 
 ### Open Questions
 
@@ -427,8 +441,8 @@ Phase 1→2 philosophy: build the road before the car is finished.
   Phase 2 for STARKs is gated on reducing proofs to a few KB. **NovaSlim already
   achieves sub-kilobyte proofs (~0.4–2.5 KiB) and a working Aiken verifier,
   demonstrating that Phase 2 infrastructure is technically feasible today; the open
-  question is the PQ hardening of
-  its folding scheme.**
+  questions are (a) closing the verifier soundness gap (MLE evaluation, commitment
+  opening, fold verification) and (b) PQ hardening of its folding scheme.**
 - Native STARK verifier feasibility in the ledger: Plutus budgets are orders of
   magnitude too small; only a native (non-Plutus) primitive is credible. NovaSlim's
   Plutus V3 verifier sidesteps this by using BLS12-381 scalar arithmetic directly.
--- a/CIP-0197/README.md
+++ b/CIP-0197/README.md
@@ -227,7 +227,7 @@ zkVM. The comparison and selection of these options is discussed in
    of commitment scheme) with **sub-millisecond verification** (~0.2 ms) and **no
    trusted setup**. It is implemented in Rust with support for six curves (including
    BLS12-381, Cardano's native curve) and three commitment schemes (Pedersen, SIS,
-   Hash). Critically, it already has a **working Aiken eUTXO verifier** running in
+   Hash) — of which **SIS** and **Hash** are post-quantum (conjectured). Critically, it already has a **working Aiken eUTXO verifier** running in
    Plutus V3 on BLS12-381 scalar arithmetic. Its protocol design, security proofs, and
    benchmark analysis are specified in the NovaSlim whitepaper [NOVASLIM-WP].
    Reference figures (VRF circuit, BLS12-381, Pedersen, 254 steps): fold ~3.5 s,
@@ -376,13 +376,12 @@ server.
 ### Backend selection
 
 The proving backend options are specified in
-[Proving backend options](#proving-backend-options). Here we motivate the choice.
-The practical reality is stark: **only one option (STARKs) is proven post-quantum today,
-and its proof size (~219 KB) precludes on-chain settlement.** All other candidates are
-either not quantum-safe (Nova), require substantial re-engineering to become quantum-safe
-(Halo2, Plonky3), or are research prototypes lacking production maturity (Lova). This
-comparison ultimately serves a small number of decisions, and the CIP leans on NovaSlim
-for the Phase 2 infrastructure path; the reasoning is below.
+[Proving backend options](#proving-backend-options). Here we motivate the choice and
+distil the options down to the two that are actually deployable. The practical reality
+is stark: **only STARKs are proven post-quantum today, and their proof size (~219 KB)
+precludes on-chain settlement.** Every other candidate is either not quantum-safe
+(Nova), requires substantial re-engineering to become quantum-safe (Halo2, Plonky3), or
+is a research prototype lacking production maturity (Lova):
 
 | | PQ? | On-chain viable? | Status |
 |---|---|---|---|
@@ -391,16 +390,36 @@ for the Phase 2 infrastructure path; the reasoning is below.
 | Plonky3 | 🔜 Requires implementation | ❌ No (~10–50 KB) | Early maturity |
 | Nova / groth16-prover | ❌ No (DLOG-based) | — | Classical only |
 | Lova | ✅ Conjectured | 🔜 Future (~5–10 KB) | Research prototype |
-| **NovaSlim** | ❌ Classical | ✅ **Yes** (~0.4–2.5 KiB) | **Implemented (experimental, unaudited)** |
+| **NovaSlim** | ✅/🔜 Conjectured (SIS & Hash) | ✅ **Yes** (~0.4–2.5 KiB) | **Implemented (experimental, unaudited)** |
 
-Why NovaSlim is the backend we lean on: it is **not** a post-quantum backend in its
-current form — its security proof is classical — but it solves the exact infrastructure
-problem that blocks every other option: proof size, verification cost, and on-chain
-verifier availability. The recommended posture is to deploy NovaSlim **now** for Phase 2
-infrastructure (wallet integration, witness format, on-chain verifier, transport), and
-upgrade the proving backend to a lattice-hardened variant (Lova-style) when that
-cryptography matures, keeping the same proof format and Aiken verifier. This mirrors the
-CIP's Phase 1→2 philosophy: build the road before the car is finished.
+That narrows the field to **two usable ways forward**, which is why the comparison
+serves only a small number of decisions:
+
+- **Way 1 — STARK (RISC Zero), the established backend.** Proven post-quantum and
+  production-ready, with a reference implementation matching the [ZKPoSP] paper. Its
+  blockbuster problem is proof size: ~219 KB cannot settle on-chain (≈13× the entire
+  `maxTxSize` budget), so it is confined to the Phase 1 off-chain path until a
+  proof-size-reduction programme (ZK-friendly hashing, smaller FRI configurations)
+  brings it within budget. It is the low-risk default for Phase 1.
+- **Way 2 — NovaSlim, the experimental backend, demonstrated to work.** Initially
+  designed as a classical Nova-family IVC, its **SIS** and **Hash** commitments are
+  post-quantum (conjectured), giving a PQ path today, and it is the *only* option with
+  sub-kilobyte proofs (~0.4–2.5 KiB), sub-millisecond verification, **and** a working
+  on-chain Aiken verifier (Plutus V3). The author has demonstrated the end-to-end
+  pipeline — step-circuit folding, slim-proof generation, and on-chain verification —
+  at `cardano/cip197` [NOVASLIM-CIP197]. Its caveats are that it is experimental and
+  unaudited, and that its post-quantum guarantee is as strong as the hash/SIS
+  parameters, not as strong as Lova's lattice construction (the underlying sumcheck
+  folding is classical, random-oracle / Fiat–Shamir based).
+
+The CIP leans on **Way 2 (NovaSlim)** for the Phase 2 infrastructure path, because it
+solves the exact problem that blocks every other option: proof size, verification cost,
+and on-chain verifier availability — and it is already shown working. The recommended
+posture is to deploy NovaSlim **now** for Phase 2 infrastructure (wallet integration,
+witness format, on-chain verifier, transport) with a PQ commitment, and to upgrade the
+proving backend to a lattice-hardened variant (Lova-style) when that cryptography
+matures, keeping the same proof format and Aiken verifier. This mirrors the CIP's
+Phase 1→2 philosophy: build the road before the car is finished.
 
 ### Open Questions
 
@@ -451,7 +470,8 @@ selection](#backend-selection) for the motivation); nothing here is final.
 - NovaSlim proof-of-concept: end-to-end pipeline validation (VRF stand-in
   circuit; the BIP32-Ed25519 circuit is still under construction) + slim
   proof generation + Aiken on-chain verification, to validate the Phase 2
-  infrastructure path.
+  infrastructure path. A working end-to-end demonstration already exists at
+  `cardano/cip197` [NOVASLIM-CIP197], including the `E2E.md` runbook.
 - CDDL specification of the witness format for a future native STARK-verifier
   primitive.
 
@@ -479,6 +499,9 @@ selection](#backend-selection) for the motivation); nothing here is final.
   https://github.com/paweljakubas/nova-slim
 - [NOVASLIM-WP] NovaSlim technical whitepaper: protocol design, security proofs, and
   benchmark analysis. https://github.com/paweljakubas/nova-slim/raw/master/whitepaper.pdf
+- [NOVASLIM-CIP197] NovaSlim CIP-197 end-to-end demonstration: step-circuit folding,
+  slim proof generation, and on-chain Aiken verification (`E2E.md` runbook).
+  https://github.com/paweljakubas/nova-slim/tree/master/cardano/cip197
 - [LOVA] Fenzi, Knabenhans, Nguyen, Pham, "Lova: Lattice-Based Folding Scheme from
   Unstructured Lattices" (ASIACRYPT 2024), lattice-based and quantum-secure.
   https://eprint.iacr.org/2024/1964
@@ -501,5 +524,6 @@ This CIP is licensed under [CC-BY-4.0](https://creativecommons.org/licenses/by/4
 [NOVA]: https://github.com/cardano-foundation/bls/tree/main/groth16-prover
 [NOVASLIM]: https://github.com/paweljakubas/nova-slim
 [NOVASLIM-WP]: https://github.com/paweljakubas/nova-slim/raw/master/whitepaper.pdf
+[NOVASLIM-CIP197]: https://github.com/paweljakubas/nova-slim/tree/master/cardano/cip197
 [LOVA]: https://eprint.iacr.org/2024/1964
 [ZCASH-QR]: https://www.coindesk.com/research/building-the-zcash-machine-tachyon-and-quantum-readiness
--- a/CIP-0197/README.md
+++ b/CIP-0197/README.md
@@ -180,8 +180,60 @@ Two numbers shape the design. The ~219 KB proof is far too large to post on-chai
 today, which is why the recommended deployment (Phase 1) verifies proofs
 off-chain — see [On-chain constraints](#on-chain-constraints). And verification is
 fast and constant, so wallets, exchanges, and indexers can afford to check every
-transaction. The one-time derivation proof can be generated lazily or delegated to a
-proving server, so it does not slow down or complicate everyday signing.
+   transaction. The one-time derivation proof can be generated lazily or delegated to a
+   proving server, so it does not slow down or complicate everyday signing.
+
+### Proving backend options
+
+The proving system can be instantiated in several ways, each serving a different
+part of the specification — the Phase 1 off-chain path, the Phase 2 on-chain path, or
+a post-quantum guarantee. The [ZKPoSP] relations are backend-agnostic (any change to
+the backend re-opens the security argument), so the options below are the candidate
+instantiations under consideration; the reference implementation uses the RISC Zero
+zkVM. The comparison and selection of these options is discussed in
+[Backend selection](#backend-selection).
+
+1. **RISC Zero zkVM — the [ZKPoSP] paper's backend; the default baseline.** Proof relations are
+   written in ordinary Rust, reusing audited primitives (`curve25519-dalek`, `hmac`,
+   `sha2`). Lowest engineering risk (no hand-built constraints), but the zkVM's own
+   arithmetization, prover, and commitment scheme still require audit. Reference figures
+   (paper §9, AMD Ryzen 9 9950X3D): signing proof ~12.5 s, derivation proof ~156 s,
+   verification ~9–10 ms, proof ~219 KB. **The proof size is the blocking issue for
+   Phase 2.**
+2. **Halo2/midnight-zk with a FRI polynomial commitment scheme.** A heavily optimized
+   circuit for this relation already exists (≈114k rows @ k=17). The default KZG PCS is
+   pairing-based and therefore *not* quantum-safe; swapping it for FRI is real work — it
+   changes the commitment scheme, the FRI domain/blow-up, proof size, and soundness
+   parameters — and the security argument must be re-opened and re-audited. This is not
+   hypothetical: the Tachyon zkVM [TACHYON] already implements Halo2 with a FRI PCS, and
+   Zcash's own quantum-readiness roadmap plans hash-based or STARK-style hardening of its
+   Halo2 proofs [ZCASH-QR], a staged posture that mirrors this CIP's Phase 1 → 2.
+3. **Plonky3-style native FRI-STARK circuits.** The [ZKPoSP] paper's own forward plan for
+   proof-size reduction; removes the zkVM's fixed per-segment overhead, but requires an
+   audited circuit implementation that is at an earlier stage of maturity. The Tachyon
+   backend also tracks Plonky3 [TACHYON].
+4. **Nova / groth16-prover [NOVA] — efficiency comparison only.** Efficient and readily
+   available, but not quantum-proof and requires a trusted setup; useful as a performance
+   bound and for a possible hybrid, not as the primary mechanism.
+5. **Lova [LOVA] — lattice-based IVC.** A recursive-SNARK descendant of Nova that
+   replaces its pairing-based assumptions with a lattice-based construction, making it
+   quantum-proof. Its incrementally verifiable computation gives linear prover time and
+   constant verification, which fits the Phase 1 off-chain story; the caveat is early
+   maturity — a first-of-its-kind Nova implementation, unproven at Cardano scale — so it
+   is a candidate for Phase 1 evaluation alongside the STARK route, not the default
+   baseline.
+6. **NovaSlim [NOVASLIM] — practical Phase 2 stepping stone.** A Nova-family IVC
+   system that produces **sub-kilobyte transparent proofs** (~0.4–2.5 KiB, independent
+   of commitment scheme) with **sub-millisecond verification** (~0.2 ms) and **no
+   trusted setup**. It is implemented in Rust with support for six curves (including
+   BLS12-381, Cardano's native curve) and three commitment schemes (Pedersen, SIS,
+   Hash). Critically, it already has a **working Aiken eUTXO verifier** running in
+   Plutus V3 on BLS12-381 scalar arithmetic. Its protocol design, security proofs, and
+   benchmark analysis are specified in the NovaSlim whitepaper [NOVASLIM-WP].
+   Reference figures (VRF circuit, BLS12-381, Pedersen, 254 steps): fold ~3.5 s,
+   compress ~0.04 s, verify slim ~0.2 ms, proof ~0.4 KiB. With SIS commitment
+   (m=128): fold is faster than Pedersen (matrix–vector products replace MSMs)
+   and the slim proof size stays ~0.4 KiB.
 
 ### Deployment: two phases
 
@@ -231,7 +283,7 @@ primary deployment (Conway-era limits; updatable via governance):
 **However, NovaSlim [NOVASLIM] already bypasses these constraints:** its slim proofs
 are **~0.4–2.5 KiB** (well under `maxTxSize`) and its Aiken verifier runs in **Plutus
 V3** on BLS12-381 scalar arithmetic. This makes NovaSlim the only backend with a
-working Phase 2 path today; see [Proof of Concept](#proof-of-concept-novaslim--aiken-on-chain-verifier).
+working Phase 2 path today; see [Proving backend options](#proving-backend-options).
 
 Raising `maxTxSize` alone would not suffice for STARKs (the receipt also exceeds
 `maxBlockBodySize`), and the post-Q-day variant does not reduce proof size — it only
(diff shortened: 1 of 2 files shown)


CIP-POST-QUANTUM-WALLETS/README.md
```diff
@@ -0,0 +1,303 @@
...
+low.
+
+**The hashes.** Binding hashes use SHA-256; the key-derivation MAC remains
+HMAC-SHA512, unchanged from BIP32-Ed25519 — so the derivation itself is not touched.
+
+**What it costs.** Reference figures from the paper, on a 16-core box:
+
+| Operation | When | Cost |
+|---|---|---|
+| Signing proof | every transaction | ~12.5 s to prove |
```
perturbing, line 143: I think these numbers need to be specified more. What machine and more context of these timings are needed?

Especially I am interested in

1. Peak memory consumption (proving generally has one peak at the end)
2. Can you try running this via WASM in the browser?

The latter is interesting because it is generally the worst-case environment.

CIP-POST-QUANTUM-WALLETS/README.md
```diff
@@ -0,0 +1,303 @@
...
+phases; only the verification location and the on-chain witness differ. The prover
+needs private anchor material and may run on-device or be delegated to a proving
+server.
+
+## Rationale: How does this CIP achieve its goals?
+
+- **Why an additive signature layer:** public keys are byte-identical to standard
+  BIP32-Ed25519 output, so there is no migration, no re-registration, and no hardware
+  wallet or multisig (CIP-1854) breakage.
+- **Why anchor at `account'`:** proofs over non-hardened leaves are unsound (the tweaks
```
perturbing, line 248: maybe add here that this decision is also made because it saves two hard rounds, each having 2 HMAC (4 sha512 hashes) that are expensive in a circuit. So this decreases proving time while maintaining the soundness of our goal.

CIP-POST-QUANTUM-WALLETS/README.md
```diff
@@ -0,0 +1,303 @@
...
+  `m/1852'/1815'/account'/role/index` over an audited STARK backend (e.g. RISC Zero).
+- Conformance test vectors for the Cardano path covering derivation, signing, and
+  verification, including the anchor-at-`account'` constraint.
+- Public, independent security audit of the proof system and its Cardano integration.
+- At least one wallet adopting Phase 1 (off-chain witness).
+- A documented proposal for the post-Q-day rule as a protocol parameter.
+
+### Implementation Plan
+
+- Rust implementation of `R_pruned-deriv` / `R_pruned-sign` over the RISC Zero zkVM,
```
perturbing: Going the Risc0 route means that the circuit will be bigger than a handwritten one.

The CIPs current rationale is less audit surface, but I think that is not entirely true.

We have in Midnight-zk all the tools (audited) to create such a circuit (I even have a heavily optimized one ready, 114,662 rows @ k=17). I believe that for FRI, you add two orders to this for the blowup.

The only downside is that it is using pairing crypto... That said, the midnight-zk current KZG PCS can easily be swapped out for FRI PCS.

How big is your Risc0 impl?

perturbing (commented):
Hi Pawel very nice CIP; this was also on my mind!

I really like the dual route of signature and proof; it's smart to optionally bind still to the old path (in case the zk has a bug or QC takes a while still).

A few initial review + questions below

CIP-POST-QUANTUM-WALLETS/README.md
```diff
@@ -0,0 +1,303 @@
...
+  A purely additive, signature-layer upgrade avoids that.
+- [ZKPoSP] already specialises to BIP32-Ed25519 and to exactly the "three hardened +
+  two non-hardened" path shape Cardano uses (`m/1852'/1815'/account'/role/index`),
+  making Cardano the natural first deployment.
+
+## Specification
+
+### Background
+
+**How a Cardano wallet makes keys today.** A wallet starts from a single seed (backed
```
AndrewWestberg, line 45: Most wallets don't actually keep the seed phrase around. They derive a root key from it. The root key is what is actually stored encrypted on disk to represent the wallet. The same goes for hardware wallets. They don't store seed phrases. Root key derivation from seed is a one-time operation during wallet creation/import so that it has minimal exposure.

CIP-POST-QUANTUM-WALLETS/README.md
```diff
@@ -0,0 +1,303 @@
+---
+CIP: "?"
+Title: Post-Quantum Zero-Knowledge Signatures for Cardano HD Wallets
```
rphair, line 3:
```suggestion
Title: Post-Quantum ZK Signatures for HD Wallets
```
@paweljakubas @perturbing how about this as a merge of the current PR title & the CIP title?  This is about as much space as we have for a CIP title (abbreviated as necessary, but still recognisable to readers) and includes all the technical concepts without the redundant word `Cardano`.

I've put it into the PR title to see how it looks to everyone but please feel free to update further (we can confirm a title at the upcoming CIP meeting, if not already).

CIP-POST-QUANTUM-WALLETS/README.md
```diff
@@ -0,0 +1,303 @@
...
+    alt Phase 1 — off-chain verification
+        W-->>V: (M, tau, sigma) out-of-band
+        V->>V: verify pi_deriv & pi_sign, cross-check hkR, address = blake2b-224(A)
+        W->>N: submit tx (witness: classical Ed25519)
+        N->>N: validate as today (unchanged)
+    else Phase 2 — native verifier (future)
+        W->>N: submit tx (witness: tau, sigma) [needs proof-size reduction]
+        N->>N: native STARK verification (not Plutus)
+    end
+    N->>N: include in block
```
rphair, line 231:
```suggestion
sequenceDiagram
    autonumber
    actor U as User
    participant W as "Wallet (cardano-addresses)"
    participant P as "ZKPoSP Prover"
    participant V as "Off-chain Verifier"
    participant N as "Node / Consensus"

    U->>W: seed / mnemonic (CIP-9)
    W->>W: derive m/1852'/1815'/account'/role/index<br>A = kL·B<br>address = blake2b-224(A)
    W->>P: request derivation proof (anchor = account')
    P-->>W: tau = (pi_deriv, hkR)   [one-time]
    U->>W: initiate payment<br>W builds tx body M
    W->>P: request signing proof for M
    P-->>W: sigma = (pi_sign, hkR)
    alt Phase 1 — off-chain verification
        W-->>V: (M, tau, sigma) out-of-band
        V->>V: verify pi_deriv & pi_sign, cross-check hkR, address = blake2b-224(A)
        W->>N: submit tx (witness: classical Ed25519)
        N->>N: validate as today (unchanged)
    else Phase 2 — native verifier (future)
        W->>N: submit tx (witness: tau, sigma) [needs proof-size reduction]
        N->>N: native STARK verification (not Plutus)
    end
    N->>N: include in block
```
Semicolons in Mermaid are new lines _of code_: not of labels.  The current syntax error is the same as the one happening here: https://github.com/mermaid-js/mermaid/issues/1445#issuecomment-639553386 — fixed by using `<br>` for line breaks.

CIP-POST-QUANTUM-WALLETS/README.md
```diff
@@ -0,0 +1,303 @@
...
+
+## References
+
+- [ZKPoSP] Botta, Pospieszalski, Ragnoli, Ranvier, "ZKPoSP: Post-Quantum
+  Zero-Knowledge Proofs for Hierarchical Deterministic Wallets",
+  ePrint 2026/1508: https://eprint.iacr.org/2026/1508
+- [CIP-1852] HD (Hierarchy for Deterministic) Wallets for Cardano.
+- [CIP-1854] Multi-signatures HD Wallets.
+- [CIP-0003] Withdrawal scripts and address formats (HD Random vs HD Sequential).
+- [KL17] Khovratovich and Law, BIP32-Ed25519 (EuroS&PW 2017).
```
rphair: @paweljakubas this list of common references is very helpful but it would work better to have these filled out into links.  Please check the end of the raw Markdown of [CIP-0001](https://github.com/cardano-foundation/CIPs/blob/master/CIP-0001/README.md) whose external links at the end of the document can be easily referenced by short terms like the ones you've got between `[]`.

It works OK as you have it, but will be easier for readers & reviewers if these expand into proper links (see [shortcut reference links](https://github.github.com/gfm/#shortcut-reference-link)).

rphair (commented):
thanks @paweljakubas and I feel sure we will be ready to confirm this as a candidate in `Triage` at the next CIP meeting: https://hackmd.io/@cip-editors/141

Now that it's being reviewed, please make sure any further forced pushes also contain any commits to this PR branch from other reviewers / GitHub UI 🙏 (to demonstrate constructive peer review).

cc reviewers on PQ CIPs so far (besides @perturbing): @kozross @hjeljeli32 @mgajda  + ecosystem impact: @ptrdsh @KtorZ

CIP-POST-QUANTUM-WALLETS/README.md
```diff
@@ -0,0 +1,303 @@
+---
+CIP: "?"
+Title: Post-Quantum Zero-Knowledge Signatures for Cardano HD Wallets
+Category: Wallets
+Status: Proposed
+Authors:
+  - Pawel Jakubas <jakubas.pawel@gmail.com>
+Implementors: N/A
+Discussions:
+  - Original PR: https://github.com/cardano-foundation/CIPs/pull/1213
```
rphair:
```suggestion
  - Original PR: https://github.com/cardano-foundation/CIPs/pull/1242
```
apologies @perturbing there was a problem in GitHub UI that caused it to delete _your_ comment when it would not withdraw my own comment upon it from my own review. 😵‍💫

@paweljakubas the URL to use here is `this` URL; labelled "Original" since in general other PR discussions accumulate here over time.

CIP-POST-QUANTUM-WALLETS/README.md
AndrewWestberg: I think a ZK proof showing that leaf key can be derived from the root key is a better system. Requiring starting from the seed phrase continually exposes the seed phrase when generating the proof. Not only that, but it exposes every blockchain wallet that the user might derive from that seed phrase. I see no reason to expose my seed phrase which could be used for both my Bitcoin wallet and my cardano wallet when signing a cardano transaction.

perturbing: Though the CIP does not mention it explicitly right now, I think this is what is done (it does not use a seed phrase). The CIP mentions an optimized reduced path over only the hardened account path onwards in the tree. Note that this CIP does not have CIP-0003 in scope (seed -> root), also because this is not that standardized across implementations (ledger/trezor/icarus) so hard to do in ZK.

This CIP proves knowledge via the root; I think to be exact it tries to proof
> "I know 96 root bytes (kL ‖ kR ‖ chain-code) that derives — entirely in-circuit — to this Ed25519 public key A along m/1852'/1815'/account'/role/index."

And via the optimization, this becomes
> "I know 96 coin-level bytes (kL ‖ kR ‖ chain-code) — the extended key at m/1852'/1815' — that derive, entirely in-circuit, to this Ed25519 public key A along account'/role/index."

CIP-POST-QUANTUM-WALLETS/README.md
paweljakubas: Hi, and thank you for taking the time to look at the cost numbers so carefully — these are genuinely useful points, and they hit exactly where our data is thinnest. Let me respond to each one.

A small caveat first, to be fully transparent: this CIP is written from the ZKPoSP paper (ePrint 2026/1508), and we don't have our own prototype yet. So all numbers below are the paper's, and we've committed to re-measuring them on our own reference implementation before relying on any of them.

1. Machine and context of the timings — You're absolutely right, "16-core box" was too vague. The paper (§9) actually specifies it precisely: a dedicated OVH EU-central cloud server with an `AMD Ryzen 9 9950X3D (16 cores / 32 threads @ 4.5 GHz, 64 GB RAM, no GPU), Ubuntu, Rust over RISC Zero v5.0.0-rc.1`. I've added that exact spec to the CIP's cost table and noted that these are reference figures for that machine, not for end-user devices.

2. Peak memory consumption — Great question, and your intuition is correct: STARK proving does have one dominant memory peak at the end, in the FRI commitment phase. The honest answer is that the paper does not benchmark prover memory — the ~8.4 MB figure it reports is verifier-side. So I don't have a number to give you, and I'd rather not guess. I've added peak-memory measurement (including its profile across the proving phases) to the acceptance criteria. And if it turns out to be prohibitive, the paper's own design already has the escape hatch: the one-time derivation proof can be delegated to a proving server, leaving only the cheap per-tx signing proof to run on-device.

3. WASM in the browser — Agreed, that's the worst-case environment and very much where Cardano wallets live. The paper doesn't cover it, and without a prototype we can't run it yet, so here's our reasoning instead:
- What a browser wallet needs per transaction is verification — ~9–10 ms, constant, sub-10 MB. That's trivially fine in WASM.
- The heavy part is proving (~12.5 s signing, ~156 s one-time derivation on that desktop-class machine); browser WASM will be slower still. The paper's own plan (§11) delegates the one-time derivation proof to a proving server — exactly what Phase 1 assumes.
- I've added a browser/WASM benchmark (prove + verify, timing and peak memory) to the implementation plan, and will publish results as soon as I have them (of course if we agree we are pursuing this path first;-)).

Until then, please read all timings as the paper's reference figures on the machine above — not as a promise about end-user devices.

CIP-POST-QUANTUM-WALLETS/README.md
paweljakubas: good remark. added!

CIP-POST-QUANTUM-WALLETS/README.md
paweljakubas: Thanks — very helpful comment. A few points in response:

1. The Implementation Plan is tentative. I am not committing to one backend; it's a menu of options to pursue, and the CIP now enumerates them explicitly (RISC Zero zkVM as the paper's baseline, Halo2/midnight-zk with the KZG PCS swapped for FRI, Plonky3-style native FRI-STARKs, and Nova as a comparison-only bound). The one hard constraint is a post-quantum prover.

2. You're right about the audit claim. The zkVM buys lower engineering risk, not a smaller audit surface — its own arithmetization and prover still need auditing. I have reworded the rationale accordingly.

3. On quantum-proofness. This is the point we can't bend, and it excludes the options as they currently stand: the midnight-zk circuit (KZG is pairing-based, broken by Shor's), out-of-the-box Halo2 (same pairing issue), and standard Nova (discrete-log folding + trusted setup). **That's why the CIP defaults to FRI-STARKs**. I'd genuinely welcome the Halo2/midnight-zk route with FRI — your 114,662-row @ k=17 circuit is a great starting point — but the PCS swap is real work and re-opens the security argument, and we'd then benchmark it head-to-head against RISC Zero. I'll also try my optimized Nova (I have the groth16 and nova prover) purely as a performance comparison.

4. How big is our RISC Zero impl? Honest answer: I have no prototype yet — the CIP is based on the paper, which reports the signing circuit as constant (two SHA-256 calls via the accelerator, ~12.5 s) and the derivation proof at ~156 s, but doesn't publish per-scheme cycle/segment counts. Also, a RISC-V zkVM trace and a PLONKish circuit aren't directly comparable row-for-row; what I'll benchmark is proving time, peak memory, proof size, and audit cost. I'll publish those numbers as soon as the reference implementation exists. Of course, if there is decision to pursuit this path first ;-)

CIP-POST-QUANTUM-WALLETS/README.md
paweljakubas: Good points, and you've essentially described the design correctly — let me confirm and add one thing we think is important.

We anchor at the extended parent node, not the seed phrase. You're right on both counts: wallets (software and hardware) store extended keys, and root-key derivation from the seed is a one-time operation at wallet creation/import. The CIP's pruned relation is exactly your second formulation: "I know 96 coin-level bytes (kL ‖ kR ‖ chain-code) — the extended key at m/1852'/1815' — that derive, entirely in-circuit, to this Ed25519 public key A along account'/role/index." The seed phrase is never needed at proving time, and no other coin derived from that seed is exposed. CIP-0003 (seed → root) is deliberately out of scope, precisely because that step is not standardized across implementations (Ledger/Trezor/Icarus differ), which would make it fragile to encode in ZK.

Why anchoring at the extended parent node matters here. This is the part we'd like to stress: in Cardano, the account-level extended key account' is the common ancestor of all credentials — payment (role 0), staking (role 2), DRep, and constitutional-committee keys are all derived from that same extended parent node via the non-hardened role/index steps, purely by public-key derivation. So a single derivation proof anchored at account' (generated once) certifies provenance for the entire credential family derived from that account, not just one leaf. That is why the anchor sits at the extended parent node: it is simultaneously the right security boundary (hardened, so the witness binds to a legitimate key) and the right coverage boundary (one proof serves payment + staking + DRep + CC).

So — agreed with your framing, and we'll make sure the CIP says explicitly that the seed phrase is never exposed and that the proof is anchored at the extended key, covering the account-to-payment/DRep/staking derivation structure.

paweljakubas:
Hi Thomas @perturbing , thank you very much for taking a look and giving helpful answers wrt cryptography.

I would like to add one more data point worth sharing: Zcash's ecosystem is converging on the same conclusion. The Tachyon zkVM (`kroma-network/tachyon`, ex-PSE) already implements Halo2 with a FRI polynomial commitment scheme — so "Halo2 with FRI" isn't hypothetical, it's shipped tooling we could reuse. And Zcash's own quantum-readiness roadmap explicitly plans hash-based or STARK-style hardening of its Halo2 proofs, staged over time rather than rushed — largely for the reasons you'd expect: post-quantum proofs are bigger, and migrating prematurely risks breaking hardware-wallet integrations and locking in immature standards. 
That staged posture is essentially the `Phase 1 → Phase 2` reasoning, and it gives us a second real-world vote for the additive, swap-the-commitment-layer approach over re-keying. 

I've added these as references in the CIP's backend-options section.

CIP-POST-QUANTUM-WALLETS/README.md
paweljakubas: I added all references!

rphair: @paweljakubas to elaborate what I was suggesting before: your document still has unlinked text e.g. `[ZKPoSP]` which the reader must either "Find" in the browser or scroll down to your _References_ section to find visually. In either case, they have to remember where they were reading in a very detailed document to continue after following the reference.

GitHub and CIPs so far have 2 ways of resolving this:

### Shortcut reference links (see [GFM spec](https://github.github.com/gfm/#shortcut-reference-link))

These interpolate the external link directly into your tag whenever it's used.  Once defined like this, the links will work exactly as in the document now... but they won't appear as unlinked text anymore:

```
[ZKPoSP]: https://eprint.iacr.org/2026/1508
```

You can see how this works in CIP-0001 as suggested above: _the **end**_ of the Markdown source (the "raw" `README.md`) has a list of links in that format: they're never displayed but they're substituted wherever they're used in the document.  This doesn't allow you to specify your own titles for the references: often not needed since they can get that from the target document.

### Footnotes (see [GFM docs](https://docs.github.com/en/get-started/writing-on-github/getting-started-with-writing-and-formatting-on-github/basic-writing-and-formatting-syntax#footnotes))

If you add a numerical footnote (e.g. `[^1]`) after your terms, papers, CIP sections, etc. it will go to another place in the document like your current _References_ section: with a return link to wherever it was followed from the document.  Following the GitHub markdown docs, your _References_ list items would look like this:

```
## References

[^1]: Botta, Pospieszalski, Ragnoli, Ranvier, "ZKPoSP: Post-Quantum Zero-Knowledge Proofs for Hierarchical Deterministic Wallets", ePrint 2026/1508: https://eprint.iacr.org/2026/1508
[^2]: CIP-1852 | HD (Hierarchy for Deterministic) Wallets for Cardano: https://cips.cardano.org/cip/CIP-1852
...
```

This allows your _References_ to exist as displayed/printed markup and to set your own titles & other related information.  You can see how it looks & how to use it with [CIP-0140](https://github.com/cardano-foundation/CIPs/blob/master/CIP-0140/README.md) (which has the most footnotes) and confirm it also works [on `cips.cardano.org`](https://cips.cardano.org/cip/CIP-0140).

CIP-POST-QUANTUM-WALLETS/README.md
```diff
@@ -0,0 +1,365 @@
...
+
+## References
+
+- [ZKPoSP] Botta, Pospieszalski, Ragnoli, Ranvier, "ZKPoSP: Post-Quantum
+  Zero-Knowledge Proofs for Hierarchical Deterministic Wallets",
+  ePrint 2026/1508: https://eprint.iacr.org/2026/1508
+- [CIP-9] Wallet Mnemonic Sequence. https://cips.cardano.org/cip/CIP-0009
+- [CIP-19] Cardano Addresses. https://cips.cardano.org/cip/CIP-0019
+- [CIP-1852] HD (Hierarchy for Deterministic) Wallets for Cardano.
+- [CIP-1854] Multi-signatures HD Wallets.
```
rphair: Regardless of the format you choose as per https://github.com/cardano-foundation/CIPs/pull/1242#discussion_r3751943372, these items still need links:
```suggestion
- [CIP-1852] HD (Hierarchy for Deterministic) Wallets for Cardano: https://cips.cardano.org/cip/CIP-0019
- [CIP-1854] Multi-signatures HD Wallets: https://cips.cardano.org/cip/CIP-0019
```

CIP-POST-QUANTUM-WALLETS/README.md
```diff
@@ -0,0 +1,365 @@
...
+- [CIP-0003] Withdrawal scripts and address formats (HD Random vs HD Sequential).
```
rphair: This item also needs a link but I'm not sure the CIP name & content are consistent:
* "HD Random vs HD Sequential" is covered in the already referenced CIP-1852 (so this item maybe not be needed).
* "Address formats" are covered in CIP-0019: https://cips.cardano.org/cip/CIP-0019

CIP-POST-QUANTUM-WALLETS/README.md
```diff
@@ -0,0 +1,365 @@
...
+- [KL17] Khovratovich and Law, "BIP32-Ed25519: Hierarchical Deterministic Keys over a
+  Non-linear Keyspace" (EuroS&PW 2017).
```
rphair:
```suggestion
- [KL17] Khovratovich and Law, "BIP32-Ed25519: Hierarchical Deterministic Keys over a
  Non-linear Keyspace" (EuroS&PW 2017): https://input-output-hk.github.io/adrestia/static/Ed25519_BIP.pdf
```
IO has already collected the paper here, if you feel comfortable linking to this repository.

rphair (commented):
@paweljakubas, per your review request: I've been following the technical review with interest, though not qualified to participate beyond presentation issues & accumulating points of others' expert review to assess when this might be complete + soliciting other reviews as may be needed.

Currently we just have to keep working on https://github.com/cardano-foundation/CIPs/pull/1242#discussion_r3751943372 re: the formatting question of whether your references would look better as a footnoted list of references or direct external links from the document.

CIP-POST-QUANTUM-WALLETS/README.md
paweljakubas: fixed

CIP-POST-QUANTUM-WALLETS/README.md
perturbing: Will you, or is it your goal, to add a benchmark on a POC implementation to the CIP?

I think this design makes or breaks with the possibility of running it.

Also because having some exact size metric is important (after all blocks have only 90112 bytes of space).

CIP-POST-QUANTUM-WALLETS/README.md
paweljakubas: done

CIP-POST-QUANTUM-WALLETS/README.md
paweljakubas: done

paweljakubas:
@AndrewWestberg @rphair @perturbing  I have fixed and double checked all references. Also did another round of my research and added prospective proof system as a candidate for Phase 1 - `Lova  — lattice-based IVC`. FYI. I am about to start building it on top of my Nova impl and will soon experiment with it (purely prototype phase but still)

CIP-POST-QUANTUM-WALLETS/README.md
paweljakubas: At some time (within several months) I will have several POCs. Right now I was more into if we are pursuing this path or not. I enumerated several candidates (five atm) that might be interesting to check at Phase 1. My hope/ask would be also that on your side you could also give it a check with your prover:-) Detailed benchmarks of all plus comprehension of trade-offs plus proper analysis of them could point to Phase 2 IMHO. So I would say right now, it would be good to make this strategic decision. Not the last one if we want premium quantum proof blockchain.

CIP-POST-QUANTUM-WALLETS/README.md
rphair: Looks resolved.  @paweljakubas please try to mark these "resolved" on GitHub to help reviewers focus on the issues that still remain (unless you feel they should be highlighted for further commentary).

CIP-POST-QUANTUM-WALLETS/README.md
rphair: Resolving since it looks like the new title is getting along well.

CIP-POST-QUANTUM-WALLETS/README.md
rphair: Resolving: fixed with latest improvements to References list.

CIP-POST-QUANTUM-WALLETS/README.md
rphair: Resolving: apparently fixed with latest improvements to References list (@paweljakubas I still don't think CIP-0003 illustrates the quoted concept but at least the point has been noted).

CIP-POST-QUANTUM-WALLETS/README.md
rphair: Resolving: fixed with latest improvements to References list.

CIP-POST-QUANTUM-WALLETS/README.md
perturbing: Of course; I think time and space-wise verification may fit in phase 1. I just fear that any proof size is just to large to put in blocks (without having 1 tx per block)

CIP-POST-QUANTUM-WALLETS/README.md
```diff
@@ -0,0 +1,379 @@
...
+- Conformance test vectors for the Cardano path covering derivation, signing, and
+  verification, including the anchor-at-`account'` constraint.
+- Public, independent security audit of the proof system and its Cardano integration.
+- Benchmark report for the reference implementation on the [ZKPoSP](#ZKPoSP) paper's machine (AMD Ryzen 9
+  9950X3D) and on a browser/WASM target, covering proving and verification time, proof
+  size, and peak prover memory (including the FRI-commitment-phase peak).
+- At least one wallet adopting Phase 1 (off-chain witness).
+- A documented proposal for the post-Q-day rule as a protocol parameter.
+
+### Proving Backend Options
```
rphair, line 294: These possible implementation options, even though considered during the time period of the surrounding activities, are generally not included in this section: which is always about the _criteria_ for progress.  The best CIPs can do when they rely on preliminary investigation to choose a path is put _the choice itself_ into a timeline with other development activities.

Alternative implementation options are generally fully described in the _Rationale_ section: or maybe _Specification_ if they will be real things rather than just concepts or ideas.  Once defined there, they can be added as items to the _Acceptance Criteria_ or _Implementation Plan_ as appropriate.  Once this subsection is moved, this section will be in the same format as all other CIPs.

rphair (commented):
@paweljakubas thanks for your continued work, in appreciation for the technical discussion & with my apologies that the formatting issues have been a distraction.

Your latest changes to fix the problem I pointed out in https://github.com/cardano-foundation/CIPs/pull/1242#discussion_r3737610178 are very different than what's used in other CIPs: _either_ shortcut (embedded) reference links _or_ Footnotes which would work best for your material.  Instead you have created HTML links, e.g.:
- from: `[ZKPoSP](#ZKPoSP)`
- to: `<a id="ZKPoSP"></a>[ZKPoSP] Botta, Pospieszalski, Ragnoli, Ranvier, "ZKPoSP: ...`

... which:
* only go in 1 direction (unlike Footnotes which link back to their place in the original document);
* aren't highlighted when the reader arrives there (they can only see all the References at the end of the document);
* aren't portable Markdown: so we have no guarantee they will work when published on `cips.cardano.org`... which I'm guessing is important to you since all your other CIP references link there.

I don't want to take time away from your technical writing & review response, and so _if you need me to edit the document to install the footnotes properly then please say the word_ and I'll do it.

Alternatively: if you have a particular reason not to use either of the GitHub Markdown standard linking styles I described in https://github.com/cardano-foundation/CIPs/pull/1242#discussion_r3751943372, then please also confirm this: so the editors can acknowledge your author's prerogative in initial review and be prepared that the reference links won't work on `cips.cardano.org` (since it generally doesn't copy raw HTML).

CIP-POST-QUANTUM-WALLETS/README.md
perturbing: @paweljakubas, is it the goal of this CIP to never put the proofs in blocks?

That is, given that along with normal transactions we also gossip the proofs across the network, a node keeps a cache of the proofs and only adopts blocks of signatures for which it has seen a zk proof? This way consensus forces only zk witnesses + normal ed25519 signatures in blocks.

I think the CIP does not say it now, but I think that we should consider this. This way we have a "cheap" way of making Cardano quantum secure without too many code changes (some nuance here is in place ofc). The only gap that I see that remains is the VRF (not derived this way). Also, people slowly need to switch to derived keys instead of generated ones.

paweljakubas: hi @perturbing That is what I was considering as back-up plan to be frank. In case we are not able to put proofs in blocks. Of course, given new developments coming we could have the following plan.
1. Try to come up with post-quantum proof system that is able to produce proofs small enough to fit in the tx
2. (if 1 not feasible) Try to produce the proofs that fit in several txs (as we can reference them mutually)
3. (if 2 not feasible/practical) A node keeps a cache of the proofs and only adopts blocks of signatures for which it has seen a zk proof (here we need to handle VRF thing)

Sounds like a plan worth adding to CIP?

perturbing: I think that a CIP should gather evidence on what is the best path and then describe that.

So I believe that this CIP is "not final" until more research is done on the topic of block impact and mitigation.

Of course, it is okay to pivot a CIP slightly when implementing it (see, for example, some PRs on the Leios CIP), but those are not big shifts.

paweljakubas: Fair point. I will gather more evidence with the proof systems I am developing/deploying and return with data :+1:

paweljakubas:
hi @rphair I tried to apply all your suggestions in the newest version - I hope I succeeded!
hi @perturbing I added TWO PATHS in https://github.com/paweljakubas/CIPs/blob/paweljakubas/post-quantum-hierarchical-deterministic-wallets/CIP-POST-QUANTUM-WALLETS/README.md#motivation-why-is-this-cip-necessary
for people to understand what is this "game" all about. That this CIP is an "intermediary" solution, a bridge. And at some point there will need to be a "proper" PQ solution. I added that as I was pinged by several people asking me why not falcon/...etc.

CIP-POST-QUANTUM-WALLETS/README.md
rphair: thanks for your persistence @paweljakubas — your approach is something like a "best of both worlds" with some redundancy added to the usual [shortcut reference links](https://github.github.com/gfm/#shortcut-reference-link) so that it produces printable markup.  In my opinion this is now a portable & maintainable document & that's what really matters. ⭐

CIP-POST-QUANTUM-WALLETS/README.md
rphair: Applied in https://github.com/cardano-foundation/CIPs/pull/1242/changes/52958914f7850ca03c8c589e0247ce3e541f3c3b.

rphair (commented):
@paweljakubas it looks like all the issues above are resolved except the technical subjects under consideration — as it should be :) — except the one (https://github.com/cardano-foundation/CIPs/pull/1242#discussion_r3761293613) of the _Proving Backend Options_ still planted in the middle of your _Path to Active_.

We will have a great `Triage` of this at the next CIP meeting (https://hackmd.io/@cip-editors/141) regardless: and if the document ordering issue is not fixed by then, we can confirm my recommendation or get alternate suggestions from other editors.

In any case I think the community at large will be very happy to see this under way.  As soon as we assign a CIP number on Tuesday (which I will recommend) we can push this out a little further across Cardano work areas to get a broad-spectrum review.

CIP-POST-QUANTUM-WALLETS/README.md
```diff
@@ -0,0 +1,436 @@
+---
+CIP: "?"
```
rphair:
```suggestion
CIP: 197
```

rphair (commented):
@paweljakubas this was well received at the CIP meeting: confirmed a candidate as expected, with great appreciation for the proposal even with the disclaimer that this is an "intermediate" rather than general solution of the post-quantum problem for asset security.

We expect that technical review will go on for a while and so please comment when you feel it has stabilised (before a editorial review to merge) or whenever you might need a higher bandwidth discussion for unresolved technical issues — e.g. via a Discord channel or a spot on the CIP meeting agenda.

As soon as possible in the review process please follow the advisement in https://github.com/cardano-foundation/CIPs/pull/1242#discussion_r3761293613 to disperse the _Proving Backend Options_ from where it hangs in the middle of your _Path to Active_.
* Specialist reviewers will be interested in all these topics and should review the document as a whole _after_ this material is properly factored into your _Rationale_ and/or _Specification_.

In the meantime please rename the containing directory to `CIP-0197` and update the "latest rendered version" link in your OP 🎉

paweljakubas:
Hi @perturbing @rphair !

Pls checkout the recent edit that tightens the CIP-197 README to reflect the actual state of the companion implementation in nova-slim (https://github.com/paweljakubas/nova-slim), specifically the end-to-end PoC at cardano/cip197/ (https://github.com/paweljakubas/nova-slim/tree/main/cardano/cip197) (documented in README.md, E2E.md, and tested via scripts/e2e_equivalence.sh). 

**NovaSlim** is a Nova-family folding scheme that produces sub-kilobyte transparent proofs for on-chain verification, with a runtime-selectable commitment layer — Pedersen for speed today, and SIS or Hash commitments that can be swapped in as quantum-hard alternatives mature without changing the proof format or the Aiken verifier.

The goal here is simply to keep the specification accurate and honest about what exists today versus what is still in progress. Four small adjustments:
- Maturity framing — "Production-ready" was too strong for an unaudited experimental project; the table now reads "Implemented (experimental, unaudited)".
- Performance claim — Removed the "~5× faster" SIS-vs-Pedersen figure, which isn't supported by the benchmark data, and replaced it with a qualitative note about matrix–vector products replacing MSMs.
- Circuit status — Clarified that the runnable validation pipeline today uses the VRF stand-in circuit; the full BIP32-Ed25519 circuit is still being built, as documented in the cardano/cip197 README.
- Deployment language — Softened "deployable today" to "technically feasible today" and "prototyping", which better matches the experimental nature of the code.

The cardano/cip197 PoC remains the concrete, tentative path for Phase 2 infrastructure: it proves the fold → compress → slim-proof → verify pipeline works end-to-end on Cardano's BLS12-381 scalar field with Plutus V3. I believe this edit makes the CIP more credible and easier for reviewers to evaluate. Happy to adjust wording if anyone prefers a different tone.

FYI. I have also changed description of the PR

PS. I was working on NovaSlim since Dec 2025 and only during weekend decided to show it to the world;-)

rphair (commented):
@paweljakubas it's great news for Cardano that you think NovaSlim will work here & that you've documented this so comprehensively. 🎉

Now more than ever, especially since more items are going into the ***Proving Backend Options*** which we've asked you to move out of its inappropriate place in the middle of _Path to Active_, before this material gets widespread technical review **we need you to move it** (as I've asked a couple times before above, with reasons) to:
* _Specification_ if you think each of these options will serve different parts of the spec
* _Rationale_ if this comparison will ultimately serve a smaller number of design choices... and the fact that you're leaning on NovaSlim now makes this the logical choice.
* (perhaps both: if you can break it up accordingly)

Although you have graciously given the CIP editors write permission to your branch, I'm hoping not to do this myself: just moving the section bodily would be clumsy compared to the care you've taken with the flow of the rest of the document.  But the editors cannot merge it the way it is now, and we need reviewers to review these options _in place_ rather than changing the structure & maybe the content at the last minute before merge.

paweljakubas:
@rphair Thanks!

I have made proper adjustments. Also added link to nova-slim whitepaper which is my humble attempt to substantiate my claims in more formal fashion. I am both enthusiastic and cautious here. Pls treat it as tentative solution. Although I tried to nail it to the best of my knowledge and abilities, and spent a lot of time on this (also by adding numerous tests and covering use cases)!

I repeat paper/specialization link https://github.com/paweljakubas/nova-slim/blob/master/whitepaper.pdf

I do think we have two backend options here I underlined.

Pls also take notice that I demonstrated e2e of what is the subject of this proposal here : https://github.com/paweljakubas/nova-slim/tree/master/cardano/cip197
Of course, this is experimental and fresh system (IMHO **conjectured PQ** but I need to state that tentatively) but is transparent (no trust setup needed), performant, produces small proofs (sub-kB or so) and has committment modularity (so except classical pedersen, we can use PQ SIS and Hash).

rphair (commented):
@paweljakubas the latest change looks great: the _Path to Active_ is now easy to follow (though as you suggest, we will be cautious about announcing this as a "solution") and the moved material looks natural where you have moved it to.

Since it's been nearly a month since we tagged our limited set of prior PQ-informed reviewers (https://github.com/cardano-foundation/CIPs/pull/1242#pullrequestreview-4885046506) the editors may rely on you to tag others working in this space & to announce when you believe all interested parties have had an opportunity to comment on this.
* Once you do that (or if we notice that "enough" time has gone by) we would mark this for final review & possible merge at an upcoming CIP meeting.

paweljakubas:
on my side, from this position, I might only kindly ask @perturbing or @iquerejeta and his team or IO Research team (@dkaidalov, @kitounliu, ...) to take a look. What is important here is that components of the system are well-established (no new cryptography here), the composition of them is my humble contribution. Of course, I am constantly polishing the system and might try to send the paper to peer-reviewed journal at some point. Also PQ is **conjuctured not proven** (yet) which I honestly underline.

kitounliu:
@paweljakubas Thanks for your work. While [NovaSlim](https://github.com/paweljakubas/nova-slim/tree/master) looks promising, it is still under development and there seems to be several major gaps: 

1. **The headline proof size is not comparable to sound systems.** The 0.4–2.5 KiB slim proof omits exactly the components (PCS openings, fold  verification, norm proofs) that cost bytes. Closing these gaps is estimated to have (with transparent setup): ~5-12 KiB classical proof, ~30-100+ KiB PQ proof.
2. **The verifier does not yet establish soundness.** `verify_slim` checks sumcheck round consistency, Fiat–Shamir re-derivation, and a zero final claim, but never evaluates the circuit's A/B/C MLEs at the random point or opens the committed Z/E there, and never re-verifies the fold (the bundle carries only the final instance). Since relaxed R1CS with a free E is trivially satisfiable, acceptance currently attests honest pipeline execution, not knowledge of a valid witness. An all-zero-polynomial transcript passes against any bundle.
3. **"Hash" and "SIS" are the same assumption.** The Hash commitment is `c = A·v` with Blake2b as a PRF for the matrix entries: Ajtai/SIS in disguise, not hash-based binding. The README's "post-quantum" label for
Hash inherits the same missing-norm-enforcement caveat currently flagged only for SIS.

Overall, in our PQ migration plan I still lean toward positioning the ZK-proof path as a fund-recovery mechanism (preferably verified by a Plutus contract) for use once classical signatures are disabled on Cardano, rather than as a day-to-day signing scheme.

paweljakubas:
hi @kitounliu 

Thank you for the precise and kind critique. I have revised the paper (and the implementation roadmap) to address all three points head-on.

1. Headline proof size is not comparable to sound systems
You are right. The 0.4–2.5 KiB figure is a batch-proving size that omits the components any sound system must carry (PCS openings, fold verification, norm proofs).
**What changed now**: The abstract now leads with a caveat, and a new §5 ("Honest sizing and remaining gaps") gives closed-gap estimates: ~5–12 KiB classical, ~30–100+ KiB post-quantum. These are competitive with Spartan and LatticeFold+, but they are no longer buried in fine print.

2. The verifier does not yet establish soundness
You are right. `verify_slim` checks sumcheck consistency, Fiat–Shamir re-derivation, and a zero final claim, but it never evaluates the circuit MLEs at the random point, never opens the committed witness/error, and never re-verifies the fold. Because relaxed R1CS has a free error vector $E$, an all-zero transcript passes against any bundle.
**What changed now**: §5.2 documents this gap explicitly. The paper now states that the guarantee is batch-proving knowledge soundness only: the prover is trusted during folding, and acceptance attests honest pipeline execution, not knowledge of a valid witness under an untrusted prover.

How I am going to fix it: The roadmap targets a two-step hardening:
- 0.3.0 — Add `az_r`, `bz_r`, `cz_r`, `er_r` to the Rust slim proof and verify algebraic consistency (`az_r·bz_r − u·cz_r − er_r == final_claim`). This prevents the trivial all-zero attack and already exists in the Aiken verifier.
- 0.3.0 — Three extra sumchecks proving `az_r`, `bz_r`, `cz_r` are the correct matrix-vector product evaluations at the random point. This restores knowledge soundness under an untrusted prover. Estimated honest size: ~3–5 KiB for Ed25519 (still under Cardano's 16 KiB limit).

3. "Hash" and "SIS" are the same assumption
You are right. The Hash commitment computes $c = A \cdot v$ where the matrix entries are derived via Blake2b. This is exactly Ajtai/SIS; the "Hash" label refers to the PRF derivation method, not to a separate hash-based binding mechanism.
**What changed now**: The paper now states explicitly that Hash is "operationally SIS/Ajtai in disguise." The missing-norm-enforcement caveat that applies to SIS applies equally to Hash: neither gives a post-quantum guarantee under adversarial witnesses without additional range proofs.

On the QROM work
Today's proofs  are deliberately layered on top of a classically sound protocol. They harden Fiat–Shamir and extraction against quantum adversaries, but they do not close the classical verifier gap. The sequence is: first close the gap (0.3.0), then the QROM theorems lift the completed protocol to the post-quantum setting.

I appreciate the rigor of this feedback — it made the paper materially more honest. The revised version will be in the repo soon.

paweljakubas:
@rphair I will need several weeks to revise the code, paper, etc. I will return here when I am ready. So please be patient, we might have something useful in the end.

@kitounliu thank you once again for taking a look and your meritocratic remarks :gem:  
I plan to adopt all suggestions and, when ready, ask kindly to have a look again! :bow:

rphair:
thanks @paweljakubas & I'm only applying the `Waiting for Author` tag to reinforce that review would be best to continue _after_ you've made your planned revisions... it won't create any prejudice to close this PR especially with your explicit comment above & please take all the time you need. 🙏

dkaidalov:
Thanks @paweljakubas for your proposal, the design looks really interesting. It's enticing because it keeps existing accounts unaffected, which solves a major social challenge in PQ migration (asking all users to transfer their tokens to new accounts).

That said, IMO the size of the witness (the zk proof here) plays a huge role. If the proof size can be brought to levels comparable to native PQ signatures, then this is a really strong proposal. Otherwise it seems impractical for day-to-day tx signing. And changing consensus to accommodate large witnesses (pruning, out-of-band caches, etc.), as was outlined before, would be a major overhaul of the consensus security argument and guarantees — my intuition is that it could be very challenging and may not be worth it.

Another concern is that it introduces complex zkVM dependencies into Cardano core. NovaSlim seems to be a new protocol not deeply studied yet. But even RISC0 or Halo2/FRI as a backend would be risky for core Cardano, beyond the size issues. We are somewhat concerned even about standardized PQ signatures, since they are relatively new crypto primitives — let alone complex zk systems that may still have soundness bugs. Maintaining them in core could be a challenge.

That said, I think this design is at least useful as a backup mechanism for classical accounts that don't voluntarily migrate to PQ-secure accounts in time. Whether it suits day-to-day transactions will depend on benchmarks.

Agreed that we need more evidence to finalize the CIP and understand how best to position it.

CIP-0197/README.md
```diff
@@ -0,0 +1,543 @@
...
+| Plutus `maxTxExecutionUnits` | 14M memory / 10G steps per tx | No STARK verifier exists in Plutus; verification is orders of magnitude beyond this budget |
+| Native scripts (V1–V3) | threshold multisig only | Cannot express arbitrary NIZK verification |
+
+**However, NovaSlim [NOVASLIM] already bypasses the size constraints:** its slim
+proofs are **~0.4–2.5 KiB** (well under `maxTxSize`) and its Aiken verifier runs in
+**Plutus V3** on BLS12-381 scalar arithmetic. This makes NovaSlim the only backend
+with a working Phase 2 *infrastructure* path today — the verifier checks
+sumcheck consistency and algebraic correctness, but does not yet evaluate the
+circuit MLEs or open commitments, so it is suitable for **fund-recovery**
+(trusted prover) rather than day-to-day untrusted signing; see [Proving backend
```
dkaidalov, line 292: even in the fund-recovery setting the verifier must be sound against an untrusted prover, since after classical signatures are disabled the claimant has no other credential — so recovery doesn't lower the bar for the proof system

rrtoledo:
Thanks @paweljakubas for your proposal. The hybrid design really is interesting and could be a way to bridge Cardano to PQ safety.

I am writing this comment now, hopping on @dkaidalov, as I want to stress on some security privacy concerns. This proposal needs to make sure that the security and privacy of all mentioned solutions and primitives are fairly assessed, and ensure it is a distinct step in any deployment plan.

For security, we need to make sure that we reach at least 128 bit of security. This is not the case now for most zkVMs that are based on STARK: they usually have a base security of 80 bits and grind to achieve 100 bits. We should hence update the STARK times and proof size for 128 bit secure proofs.

As for privacy, we have to be very careful that whatever solution we choose, we do not disclose any information on the seed and secret key. Hence either the computation must be done either locally, on TEEs or via MPC. We also need to make sure that the zk property is supported, properly implemented and activated - I would . Many papers do not look at adding zk as compression is their main application. Zk in lattices and folding is often hit and miss: either they do not support it or zk relies on the decider itself or uses random instances. The decider is often not defined and random instances need to gather enough entropy, which is not well studied.

paweljakubas:
thank you VERY much for taking a look and great comments. Highly appreciated @dkaidalov @rrtoledo !

I am in the middle of making progress with nova-slim and significant uplift.

Reply to @dkaidalov 
Thanks for the feedback — fully agree on the positioning. Some clarifying points on where NovaSlim actually sits on each concern:

**Proof size.** Native PQ signatures run ~1–8 KB; NovaSlim's slim proof is ~0.4 KiB and the honest-sound variant (with PCS openings, fold verification, norm certs) is 5–12 KiB classical / 30–100+ KiB PQ. So it is already in the range of native PQ signatures, not orders of magnitude larger.

**Consensus impact.** No native-verifier or consensus change is required. The Aiken verifier runs in Plutus V3 on BLS12-381 scalar arithmetic, and the slim proof (~0.4 KiB) sits comfortably within maxTxSize. No pruning, no out-of-band caches.

**zkVM complexity.** NovaSlim is not a zkVM — it's a Nova-family folding scheme (NIFS + transparent sumcheck + HashPC commitments). No RISC Zero, no Halo2/FRI, no STARK stack; just scalar field arithmetic and BLAKE2b. So it does not drag a zkVM dependency into core.

**Maturity.** Agreed — NovaSlim is younger (it has just been born to be honest :sweat_smile:  ) than STARKs, the concrete PQ label is conjectured (upgradable to "standard-assumption (tight)" via Module-SIS), and it is experimental/unaudited. That's exactly why the CIP positions it as a fund-recovery mechanism for classical accounts that don't migrate in time, not as day-to-day signing. I'll re-open these numbers as benchmarks on the real BIP32-Ed25519 circuit land.

Reply to @rrtoledo 
Thanks — these are the right questions. Point by point:

**128-bit security.** NovaSlim is not a STARK, so the 80/100-bit concern doesn't carry over directly. For the PQ path we validate the Module-SIS parameter sets with core-SVP estimates (tools/sis_core_svp_output.json): Conservative-I ~2^1196 classical / 2^1085 quantum, Balanced-I 2^897 classical, both exceeding 128-bit by a wide margin. The classical (Pedersen) path is 254-bit scalar-field DLP.

**Privacy / no seed disclosure.** NovaSlim does not claim zero-knowledge — I state this explicitly in the paper (the slim proof reveals the sumcheck transcript, round polynomials, challenges, and the claimed product evaluation, but not the witness truth tables). For this application the computation runs locally in the wallet (or on a trusted prover the user controls); the public input (transaction hash) is already public on-chain. No TEE or MPC is claimed or needed.

**ZK correctness.** This is precisely why we deliberately do not attempt ZK. The concerns you raise — ZK "relying on the decider itself," random-instance entropy, unsupported ZK in lattice/folding schemes — are exactly the failure modes we avoid by treating NovaSlim as knowledge-sound-but-not-zero-knowledge. The commitment binding (SIS/Hash/Module-SIS) is independent of any ZK machinery, and norm enforcement targets pre-fold witness shortness.

**Bottom line.** I do agree ZK in folding/lattice schemes is often hit or miss; my proposal avoids the question entirely by not making a ZK claim, and I document what the proof does and does not reveal. If a privacy-preserving variant is ever required, it would be a separate construct on top — not this one.

rrtoledo:
Hi again @paweljakubas , thanks for the thorough answer. :) 

*128-bit security.* I was not referring here to Novaslim but the STARKs you mention elsewhere in the paper, e.g RISC0 proof size.

*Privacy.* That's my more contentious point, Perhaps I am misunderstanding what you are doing though: I though you were generating a proof of knowledge on the secret key seed and a proof of knowledge of preimage of `hKr`? If so, without enabling zk, you are leaking information about the seed and preimage. More particularly, after seeing enough transactions (and so enough signing proofs on`hKr`), an adversary would be able to recompute `hKr` and so compute new signing proofs on it and using the derivation certificate without even knowing the seed. As such you do need privacy, and any proving system without zero-knowledge cannot be used. 
While you are talking about using hash-based STARK in the beginning of the Rationale, you still push in the backend selection and open questions for NovaSlim which does not support zk.

*Zk Correctness* Understood but zk might be needed as I explained above. You also do not describe your decider for NovaSlim.

*General question.* Why do you not consider lattice based solutions such as LatticeFold, Neo, Labrador which present smaller proofs than STARKs?

PR #1243, opened 2026-08-11
CIP-0159 | Add notes & fields for starting account balance

aniketd:
* Correct CDDL for both balance interval fields to have domain of credentials instead of reward accounts.
* Add detail about unregistered accounts also checked along with intervals.

### DIFF
--- a/CIP-0159/README.md
+++ b/CIP-0159/README.md
@@ -326,12 +326,49 @@ be mentioned:
 
 1. Using the account balance interval does *not* require a witness from the associated credential.
 2. To declare that a certain asset in the `AccountValue` has a specific balance of `n`, the asset's
-   balance interval must be set to `[n, n+1)`.
+   balance interval must be set to `[n, n+1)`. For ADA, a bare `coin` is accepted as a shorthand for
+   the same assertion.
 3. If the account balance interval is used for an asset not supported in the current whitelist, the
    transaction will fail *Phase 1 Validation*.
+4. The referenced account must be registered; an interval on an unregistered account fails
+   *Phase 1 Validation*.
+5. The network id of each reward account must match the network of the ledger the transaction is
+   submitted to; an interval on an account address of the wrong network fails *Phase 1 Validation*.
+6. Interval checks are part of *Phase 1 Validation*, but they are only evaluated when the top-level
+   transaction's `isValid` flag is true. A transaction marked `isValid = false` is processed for
+   collateral only and applies no withdrawals or deposits, so its interval assertions — which guard
+   those effects — are not reached. This holds at every level of a nested transaction: each
+   `account_balance_intervals` map is checked independently at the top level and within each
+   sub-transaction, against the account balances in the ledger state as of that level, before that
+   level's own withdrawals and direct deposits are applied.
 
 Plutus scripts will be able to see the set account balance intervals as part of their
-`ScriptContext`. See the [New Plutus Script Context section](#new-plutus-script-context).
+`ScriptContext`, keyed by credential: the network id of the reward account is not carried into the
+script context, in the same way that it is not for withdrawals. See the
+[New Plutus Script Context section](#new-plutus-script-context).
+
+### Starting Account Balance Intervals
+
+`account_balance_intervals` are evaluated against account balances for each transaction level as that level is
+processed. In a nested transaction, a sub-transaction's intervals therefore see the account balances
+*threaded through* any earlier sub-transactions, and so cannot express an assertion about the balances
+at the very start of the whole transaction. A separate top-level field,
+`starting_account_balance_intervals`, fills this gap: it asserts intervals against the account balances
+at the start of the whole transaction, before any sub-transaction or top-level withdrawal or direct
+deposit is applied. It uses the same representation as `account_balance_intervals`, requires the
+referenced accounts to be registered and to carry the correct network id, requires no witness, and is
+only present in the top-level transaction body.
+
+This field is deliberately absent from sub-transaction bodies. Every sub-transaction observes the same
+whole-transaction starting balances, so a per-sub-transaction copy could assert nothing that the
+top-level field cannot already express. In keeping with the CIP-0118 design — where top-level guards
+are responsible for the holistic view of a transaction, while other script purposes focus on the
+contents of an individual transaction — a sub-transaction builder that needs assurance about the
+starting balances can require it through a script in `required_top_level_guards`.
+
+```cddl
+starting_account_balance_intervals = {+ reward_account => account_balance_interval}
+```
 
 ### New Ledger State
 
@@ -433,6 +470,7 @@ are all forced to charge *~1 ADA* due to the `minUTxOValue` requirement.
 2. Partial withdrawals from account addresses in sub-transactions, or in a top-level transaction
    but only when plutus v1-v3 scripts are not used in it.
 3. Account balance intervals validated as part of *Phase 1 Validation*.
+4. Starting (whole-transaction) account balance intervals validated as part of *Phase 1 Validation*.
 
 **CDDL Changes**
 ```cddl
@@ -441,18 +479,23 @@ are all forced to charge *~1 ADA* due to the `minUTxOValue` requirement.
 ; Same definition as current withdrawals.
 direct_deposits = {+ reward_account => coin}
 
-account_balance_intervals = 
-  { + reward_account => 
-        [ inclusive_lower_bound: coin, exclusive_upper_bound: coin / nil ]  /
-        [ inclusive_lower_bound: coin / nil, exclusive_upper_bound: coin ] 
-  }
+account_balance_intervals          = {+ reward_account => account_balance_interval}
+starting_account_balance_intervals = {+ reward_account => account_balance_interval}
+
+; A `coin` means exact (balance == coin).
+account_balance_interval =
+    [inclusive_lower_bound : coin, exclusive_upper_bound : coin/ nil]
+  / [inclusive_lower_bound : coin/ nil, exclusive_upper_bound : coin]
+  / coin
 
 transaction_body = 
   {   0  : set<transaction_input>         
   ,   1  : [* transaction_output]      
   ...
-  , ? 23 : direct_deposits ; new field
-  , ? 24 : account_balance_intervals ; new field
+  ; fields 23 (sub_transactions) and 24 (required_top_level_guards) come from CIP-0118.
+  , ? 25 : direct_deposits ; new field
+  , ? 26 : account_balance_intervals ; new field
+  , ? 27 : starting_account_balance_intervals ; new field (top-level body only)
   }
 ```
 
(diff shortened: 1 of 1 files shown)


rphair (commented):
Thanks @aniketd for the upkeep on this CIP; I've set this for `Triage` at our CIP meeting next week (https://hackmd.io/@cip-editors/141): we may move it up to `Last Check` if we can authenticate this in the meantime, especially if you or other reviewers can link it to work in progress on `cardano-ledger`.

Other CIP-0159 contributors please feel free to review: @fallen-icarus @GeorgeFlerovsky @lehins @polinavino @colll78 @Quantumplation @michele-nuzzi @mpizenberg @AlexDochioiu

CIP-0159/README.md
```diff
@@ -329,10 +329,31 @@ be mentioned:
    balance interval must be set to `[n, n+1)`.
 3. If the account balance interval is used for an asset not supported in the current whitelist, the
    transaction will fail *Phase 1 Validation*.
+4. The referenced account must be registered; an interval on an unregistered credential fails
+   *Phase 1 Validation*.
+5. In a nested transaction, account balance intervals are validated independently at the top level
+   and within each sub-transaction, against that level's balances before its own withdrawals and
+   direct deposits are applied, and only when the enclosing transaction is phase-2 valid.
```
carlostome: I don't think the following is correct: "and only when the enclosing transaction is phase-2 valid."
Account balance interval validation is part of phase-1 validation for both sub-transactions and the top-level transaction. For sub-transactions it does not depend on whether the whole tx is phase-2 valid.

CIP-0159/README.md
```diff
@@ -329,10 +329,31 @@ be mentioned:
...
 
 Plutus scripts will be able to see the set account balance intervals as part of their
 `ScriptContext`. See the [New Plutus Script Context section](#new-plutus-script-context).
 
+### Starting Account Balance Intervals
+
+`account_balance_intervals` are evaluated against each transaction level's balances as that level is
+processed. In a nested transaction, a sub-transaction's intervals therefore see the balances
```
carlostome: I think the phrasing is a bit unclear. Balance intervals are evaluated against accounts (in the ledger state) not against transaction balances.

CIP-0159/README.md
```diff
@@ -329,10 +329,31 @@ be mentioned:
...
+*threaded through* any earlier sub-transactions, and so cannot express an assertion about the balances
+at the very start of the whole transaction. A separate top-level field,
+`starting_account_balance_intervals`, fills this gap: it asserts intervals against the account balances
+at the start of the whole transaction, before any sub-transaction or top-level withdrawal or direct
+deposit is applied. It uses the same representation as `account_balance_intervals`, requires the
+referenced accounts to be registered, requires no witness, and is only present in the top-level
```
carlostome: For consistency, I think that it would make sense to add this field to sub-transaction's body as well. I don't foresee any technical difficulty in doing so.

If this is motivated by a concrete use-case which is exclusive to having this field in the top-level transaction body, I'd suggest to include it as an example.

CIP-0159/README.md
aniketd: Yes absolutely, I will make this worded better.

CIP-0159/README.md
aniketd: Yes, I also feel now that the wording is incorrect. In the spec, we have the `account balance intervals` checks happening in `SUBENTITIES` and `ENTITIES`, which is part of the `valid` branch (`SUBLEDGER-V` and `LEDGER-V` and missing in `SUBLEDGER-I` and `LEDGER-I`). So, how do you like the rewrite to be the following:

> 5. The check runs only when the top-level transaction's isValid flag
> is true (the script-valid branch). A transaction marked isValid =
> false is processed for collateral only and applies no withdrawals or
> deposits, so its interval assertions — which guard those effects —
> are not evaluated. This holds at every level of a nested transaction:
> each interval is checked independently at the top level and within
> each sub-transaction, against the account balances in the ledger state
> as of that level, before that level's own withdrawals and direct
> deposits are applied.

CIP-0159/README.md
aniketd: > If this is motivated by a concrete use-case which is exclusive to having this field in the top-level transaction body, I'd suggest to include it as an example.

I can't think of anything that enforces the exclusivity. I will wait for someone more knowledgeable about the intended plan to chime in here.

kingslavcho:
Two things worth flagging here: 

removing the min-UTxO floor for account deposits also removes the cost floor on unsolicited deposits.
Today, dusting a wallet, sending trivial amounts across many addresses for tracking/clustering purposes is bounded by minUTxOValue: each target costs the attacker close to 1 ADA in committed capital. Since account deposits are witness-free, floor-free, and direct_deposits = {+ reward_account => value} lets a single transaction hit many reward accounts at once, this CIP would let someone dust a very large number of accounts for close to the marginal tx fee alone. Reward addresses also tend to be long-lived and public (delegation/staking), which makes them a more attractive dusting target than fresh payment addresses.
Is this trade-off intentional, or should there be some floor/rate-limit specifically for unsolicited deposits (as opposed to deposits the recipient's own protocol expects, e.g. a wallet's own fee flow)? Would be good to see this addressed.

In the new AccountState, balance appears to hold staking rewards and arbitrary direct deposits (fees, micropayments, etc.) as one combined figure, with no on-chain separation between the two.
This means there's no protocol-level way to distinguish "reward income" from "someone paid me," which complicates reward-yield accounting/reporting for anyone building on this (exchanges, tax tooling, staking dashboards) and since one reward address can now naturally serve both roles, it probably increases address reuse relative to current practice.

Was this considered and accepted as a trade-off, or is separate tracking worth adding?

rphair:
@kingslavcho the first item was addressed in the original proposal (and so likely doesn't have to appear in this update): [CIP-0159 > Whitelisted Native Assets Only](https://github.com/cardano-foundation/CIPs/tree/master/CIP-0159#whitelisted-native-assets-only).  I don't understand your 2nd item completely so maybe @fallen-icarus could answer that.

fallen-icarus:
@kingslavcho The CIP still enforces a deposit but unlike minUTxO, it is the *receiver* who pays the deposit when they whitelist the asset. The whitelisting mechanism exists to prevent dusting attacks on account addresses.

As for reward income history, rewards are paid out at epoch boundaries without being part of any transaction. Meanwhile deposits only happen _inside_ transactions. So there actually is an on-chain "footprint" which can be used to distinguish between the two types of events.

CIP-0159/README.md
fallen-icarus: AFAIU it was so that the top-level smart contracts can know the full account balance diff without having to waste execution units inspecting each individual sub-tx. cc @colll78 

Since we can't change the script context of the sub-tx based on the other sub-txs in the full tx, I don't think you can set the `starting_account_balance_intervals` to be anything but the specific sub-tx's starting balance intervals which seems redundant and wasteful.

lehins: This discussion surfaced during the Ledger meeting today and I've mentioned that it will be better to keep this field to the top level transaction only.
If this information is necessary by a sub-transaction builder it can always be enforced by inclusion of a specific script added to `requireTopLevelGuards`.

The whole script purpose design that we narrow down on with respect to nested transactions is that it is top level guards that will be responsible for the holistic view of a transaction, while all other script purposes will be focused on the contents of individual [sub | top-level]-transaction.

rphair:
@aniketd please don't force-push unless that push somehow contains all the commits that you & others have already made to this branch.  Since other people will review this work multiple times after progressive edits (generally true for everything on this repo) we need to see a change history and correlate committed changes to threaded discussions.  (This is requested [in CIP-0001](https://github.com/cardano-foundation/CIPs/blob/master/CIP-0001/README.md#1a-authors-open-a-pull-request).)

rphair (approved):
@aniketd although we didn't get much feedback from potential implementors — though I'm satisfied enough with @fallen-icarus's own review — we have had a fair enough opportunity to dispute this change from everyone tagged in https://github.com/cardano-foundation/CIPs/pull/1243#pullrequestreview-4912767781.

@carlostome it also looks like @aniketd has resolved all of your own feedback with more recent commits (here & upstream) but please post here if not, and your own agreement on consistency with the Ledger spec would be appreciated (and assumed by default; cc @lehins).

So without any further dispute we should be ready to merge this from `Last Check` at the next CIP meeting: https://hackmd.io/@cip-editors/143

rphair:
**p.s.** if you are showing 2 or 3 separate lines above, adding some reviewers while removing others, it isn't me (**edit**: problem fixed, bad mouse)

perturbing (approved):

aniketd:
@perturbing @rphair there is an update I would like to make to this work today and will post here for a review again please, if it is not a problem.

rphair:
That's fine @aniketd — I can see the GitHub UI inconsistency yesterday also rejected my application of the `Last Check` tag: hopefully that's all resolved now.  But in any case we'll wait for your update before merge.  (**p.s.** it was a faulty mouse making multiple clicks when single-clicking items from the PR lists: GitHub UI is OK)

aniketd:
I have pushed the update now. cc. @carlostome @lehins :eyes: :pray:

CIP-0159/README.md
```diff
@@ -441,33 +504,43 @@ are all forced to charge *~1 ADA* due to the `minUTxOValue` requirement.
...
 
-The plutus script context will be pre-emptively upgraded to support the fields needed in the second
-delivery phase. This way the features can be turned on with a hardfork and plutus scripts written
-for the first delivery phase can immediately take advantage of the new features without having to be
-recompiled.
+The plutus script context is upgraded to expose `account_balance_intervals`. Note that the exposed
+`AccountBalanceInterval` is ADA-only, matching this delivery phase, so the multi-asset phase will
+require a further script context change rather than being a hardfork switch for already-compiled
+scripts. `starting_account_balance_intervals` is not exposed — see
+[Starting Account Balance Intervals](#starting-account-balance-intervals).
```
fallen-icarus, line 533: The original goal was to not require scripts to be recompiled when the second delivery phase is released. Migrating liquidity to new script versions is always a pain. Is there no way to abstract away the interval differences for plutus scripts between the ADA-only phase and the multi-asset phase?

Like what if the ledger just translates the ADA only interval to the same account balance intervals as in the multi-asset phase?

lehins: FTR. This comment is not really related to this PR, but rather to a comment that I made on the original CIP that we could potentially make PlutusV4 forward compatible with the second phase of the CIP.

@fallen-icarus Unfortunately, we are too pressed on time to be able to achieve this forward compatibility. I know it would have been nice to have it. But, even if we were to try and rush such change for the upcoming hard fork, we would not only risk delaying the hard fork, but we would also risk of not getting such change right for PlutusV4, which would in turn make this forward compatibility totally unusable and we would be in a net downside for PlutusV4.

So, we'll have to postpone all of the multi-asset functionality to the second phase, as it is currently described in the CIP.

rphair:
@aniketd @lehins @carlostome @fallen-icarus this has had all the constructive attention from CIP editors that it's likely to get, even if we wait for the next CIP meeting in 6 days' time... would you prefer to merge this now or wait these several days for further comment?

lehins:
I am fine with merging it now or waiting another week. Either is fine from Ledger perspective.

PR #1244, opened 2026-08-12
CPS-0034? | Extending Plutus Core conformance testing

kwxm:
The Plutus team has received treasury funding to extend the Plutus Core conformance tests to include property tests, which will help people implementing their own Plutus Core evaluators (in particular, implementers of alternative Cardano nodes) to increase their confidence that their implementation behaves identically to the reference implementation.   This CPS describes our proposed plan and the motivation behind it and is intended to initiate a discussion about how best to design the testing framework  to make it useful to external users.

---

([rendered proposal](https://github.com/kwxm/CIPs/blob/kwxm/cps/conformance-property-tests/CPS-0034/README.md))


zliu41:
cc @KtorZ @colll78

CPS-conformance-property-tests/README.md
```diff
@@ -0,0 +1,197 @@
+---
+CPS: "?"
+Title: Portable property-based conformance testing for Plutus Core evaluators
```
rphair:
```suggestion
Title: Standardising Plutus property tests
```
The current PR title `Plutus conformance property tests` is about the right length & specificity for a CIP or CPS, but generally we try to name problem statements according to a phrasing of the problem — or a thing we are trying to achieve — without identifying & naming a solution yet.

Since the objective is to produce standardisation in this testing (?), the current phrase would be a better title for a CIP that would contain the result if you choose to publish it that way (and you would be welcome to).  So a CPS title according to the current standardisation _enquiry_ would be more appropriate at this stage: this, or whatever wording you prefer.

rphair (commented):
thanks @kwxm — looks like a promising CPS and I look forward to setting this up in `Triage` — but our validator script ([currently failing](https://github.com/cardano-foundation/CIPs/actions/runs/31568128892/job/94024179509?pr=1244)) and maybe some human reviewers would expect a CIP here because of the directory name `CIP-conformance-property-tests`.

Renaming a directory is hard to do transparently from the GitHub UI so can you please **ASAP rename the containing directory** to something like `CPS-conformance-property-tests` and so set reviewers' and the validator's expectations appropriately?

Once that's done we can schedule this for confirmation at the next CIP meeting and tag some of the node team representatives for a first look.

kwxm:
> our validator script ([currently failing](https://github.com/cardano-foundation/CIPs/actions/runs/31568128892/job/94024179509?pr=1244)) and maybe some human reviewers would expect a CIP here because of the directory name `CIP-conformance-property-tests`.

Sorry, that was a typo; I wondered what was causing all the problems.  I've changed it now.

CPS-conformance-property-tests/README.md
kwxm: > we try to name problem statements according to a phrasing of the problem

I've changed it to "Extending Plutus Core conformance testing", although the problem that we're trying to solve is really that the current tests are quite limited.  I can change it again if you want.

rphair: Resolved with https://github.com/cardano-foundation/CIPs/pull/1244/changes/9f819afc0dc3ae7c228fa1b25a1bb64b946b754e.

rphair (commented):
Ready for `Triage` at next CIP meeting (anyone interested is invited as always); just a check to confirm this & not a detailed discussion: https://hackmd.io/@cip-editors/141

Tagging some more members of other node teams to start the discussion from that side: @Quantumplation @wolf31o2 @scarmuega @michaeljfazio @michele-nuzzi @mercurial @sierkov

CPS-conformance-property-tests/README.md
rphair: p.s. @kwxm I think "Extending" is a great word to characterise the problem of limitation.

Quantumplation:
@yHSJ can you comment on Sundae's behalf on how this pertains to our existing plutus testing strategy / desires for Amaru?

yHSJ:
> @yHSJ can you comment on Sundae's behalf on how this pertains to our existing plutus testing strategy / desires for Amaru?

Yes, of course.

The existing plutus testing strategy in Amaru:
We pull in the semantic and, now, flat suites, evaluate them, and compare the final state we get (term and costs) to the expected states.

My desires:
I have had some conversation with @kwxm in the Intersect Discord about their conformance suite which has led to some improvements on the Plutus side, which is great. I think all of the concerns listed in the CPS (testing many version combinations, property tests, etc.) are concerns I share as well. Another desire I would have for a conformance suite, which was implied but not explicitly stated, is that it covers many realistic scenarios that we would see on chain.

Regarding the open questions:

> How should generated test cases and expected results be transmitted to a
  client?
  
I think that option 1 (or option 2) would be ideal. It makes it much simpler to incorporate this hypothetical suite of tests into pre-existing harnesses without forcing as many opinions onto clients.
 
 > If a description format is needed for programs, inputs, and expected outcomes
  to support useful client-side diagnostics, what should it look like?

This obviously depends on a decision on the first question. I am partial to JSON files with a well defined schema (which can include an optional description field), for the test fixture format, but there are many reasonable options.

> How should test generation and/or selection be organised across the version
  combinations? Should distinct test collections be produced per combination,
  should tests be tagged with the combinations they apply to, or should
  generation itself be parameterised by the combination under test?

I would suggest tagging tests with the combinations they apply to over either of the other options. That again feels like the least opinionated solution, allowing clients to write their harnesses in whatever way fits them most. It also ties into the next open question...

> What hierarchical or filtering structure, if any, should be offered so that
  clients can list or select subsets of tests (e.g. by category, or by version
  combination) rather than always running the full suite?
  
If tests are tagged with the version combination, that is the most obvious natural filtering method for these tests.

> What are the acceptable bounds on execution time and resource usage for a
  client running the full generated suite, and how do the candidate transmission
  mechanisms compare against those bounds in practice?
  
The short answer is "within reason". One bound could be that the entire test suite can run on the free tier of GH Actions. That obviously depends on the harness implementation as well, but it seems to be a reasonable guideline.

> Can we provide a simple way for third parties to contribute their own test
  cases to the main Plutus repository?
  
This is actually what I care the *most* about, with such a conformance test suite. It is most important to me that this suite not only benefits node diversity, but also benefits *from* node diversity. Outside contribution should be not only simple, but also encouraged. That is why I think writing the generator in Haskell is the wrong decision. While I understand why it would be used in this case, that choice inherently creates barriers to contribution and even consumption. I would personally recommend Rust, it is one of the most widely used languages in the Cardano ecosystem, but there are plenty of reasonable options.


These thoughts are my immediate reactions, and they will obviously evolve with more discussion and more consideration. Ultimately, I think the most important aspect of these tests is that they are *truly* agnostic to the client. While every decision is opinionated, I think the goal of this suite should be to leave as much choice up to the consumer as possible. All decisions should stem from goal.

CPS-conformance-property-tests/README.md
```diff
@@ -0,0 +1,199 @@
+---
+CPS: "?"
```
rphair:
```suggestion
CPS: 34
```

rphair (commented):
@kwxm the editors & observers at the CIP meeting today were happy to confirm this: especially appreciating the participation of @yHSJ @Quantumplation from the first significant node project and noting that representatives from all other apparently viable node projects have also been [invited](https://github.com/cardano-foundation/CIPs/pull/1244#pullrequestreview-4923286280) for review.

As soon as that feedback settles and is fully accommodated into the CPS, please post as such: since the editors generally won't know when the lack of additional review signifies our best consensus.  In the meantime please rename the containing directory to `CPS-0034` and update the "rendered proposal" link in your OP 🎉

CPS-conformance-property-tests/README.md
```diff
@@ -0,0 +1,199 @@
+---
+CPS: 34
+Title: Extending Plutus Core conformance testing 
+Category: Plutus
+Status: Open
+Authors:
+    - Kenneth MacKenzie <kenneth.mackenzie@iohk.io>
+Proposed Solutions: []
+Discussions:
+    - Original PR: https://github.com/cardano-foundation/CIPs/pull/?
```
rphair:
```suggestion
    - Original PR: https://github.com/cardano-foundation/CIPs/pull/1244
```
@Ryun1 do you know why this incomplete URL [didn't cause an error from the validator](https://github.com/cardano-foundation/CIPs/actions/runs/32175031634/job/95834852961?pr=1244)?  Was it because the base URL is actually valid: since `?` is often discarded as a query string?

kwxm:
> In the meantime please rename the containing directory to `CPS-0034` and update the "rendered proposal" link in your OP

@rphair OK, I've done that.  We're encouraging people to provide feedback on this and will let you know when the discussion has settled down.

sierkov:
@kwxm, thank you for looking into this. Improving implementation-independent Plutus conformance testing is a major help for alternative implementations.

Below I’ll answer the specific questions in the CPS, but first I’d like to add some context from the perspective of the [C++ implementation](https://github.com/r2rationality/turbocardano).

Along with Amaru, we are currently working on improving conformance testing for other consensus-relevant components, such as CBOR deserialization and ledger state-transition rules. Some of the points below are therefore a generalization of those discussions, but I believe they apply equally well to Plutus.

## What should a conformance suite demonstrate?

The key requirement is that the suite allows to show that an alternative implementation:
1. is fully conformant with the spec; and
2. interprets the specification in exactly the same way as other implementations.

The second point is particularly important for consensus-critical code: disagreement between two otherwise well-tested implementations is itself something we need to detect and understand.

For serialization and ledger state-transition rules, we are currently leaning toward the following process:

1. Run the same conformance dataset against all participating implementations and collect branch-level code coverage.
2. Map implementation code and branches back to the corresponding specification rules or types.
3. Investigate cases where any implementation does not exercise an expected branch, and classify the cause: implementation bug, dataset gap, specification gap or ambiguity, additional/undocumented functionality, etc.
4. Add or adjust tests where this objectively improves the conformance suite.

I think a similar approach could be useful for Plutus.

This also raises one question that may be worth answering before deciding on the interchange format:

**What constitutes a sufficiently comprehensive Plutus conformance test suite?**

Since the current property tests are generator-based rather than a fixed corpus, I assume that conformance is likely statistical: more generated cases increase confidence. It would therefore be useful to define a standard release-testing profile—for example, fixed seeds, sample counts, coverage expectations, or some combination of these—so that different implementations can make comparable conformance claims.

I would also distinguish between:
* a **small, fast suite** suitable for development and CI; and
* a **full conformance suite** used for releases and implementation audits.

Another note I'd like to add is that, beyond testing functionality, we could also benefit from testing non-functional requirements:
- performance
- memory usage

With that context, my answers to the specific questions are below.

## How should generated test cases and expected results be transmitted to a client?

My preference is an adjusted option ii - the generator to emit **both the flat-encode input program and the expected result and costs** into a client-specified directory as files, leaving comparison to the client. Also, the tooling would ideally support:
* pre-generating a corpus of a given size (number of samples);
* providing a deterministic seed;
* generating independent subsets/features in parallel.

Going with the generator approach but generating the whole data at once and not case by case avoids making the generator or reference implementation part of the critical execution path for every test and makes parallel execution much easier. Our C++ implementation is fast, and we routinely run this kind of testing across 24+ cores, so it would be unfortunate if test generation or IPC became the bottleneck.

As an example, here is our plutus bechmarking dataset alogn with its generator that we use to evaluate the C++ Plutus Machine: [Plutus Benchmarking Dataset](https://github.com/r2rationality/turbocardano/tree/main/experiment/plutus-benchmark).

If the generation performance is slow, I would also provide minimal pre-generated corpus to simplify the onboarding of new implementations.

## What should the description format look like?

I would prefer **as little additional metadata as possible**. For example, by making the directory structure self-explanatory, encoding the version information into the file or directory names:
- ```<version>/<feature>/<test-id>.flat```
- ```<version>/<feature>/<test-id>-exp.json```

## How should tests be organized across version combinations?

I think a **separate dataset or generator configuration per version** would be the easiest to use.
becuase that makes it easy to reproduce a particular historical configuration.

## What filtering or hierarchy should be provided?

A hierarchy such as:

> **Version combination → data type / feature → test id**

would work well for us.

This gives clients a straightforward way to run anything from a single feature to a complete version-combination suite.

## What are acceptable execution-time and resource bounds?

I would make **comprehensiveness the primary constraint**, rather than setting an arbitrary runtime limit first.

The full release or audit suite should contain whatever is needed to achieve the agreed level of conformance confidence. If that takes significant time, that is acceptable for release testing.

A smaller subset can then be maintained for development and CI.

At the same time, the infrastructure should make the full suite as efficient as possible by supporting parallel evaluation and ensuring generation does not become the bottleneck—either through sufficiently fast generation or through pre-generated test corpora.

## How should third-party test contributions work?

I would definitely support them.

However, I think we should treat corpus size as something to optimize rather than simply accumulate tests indefinitely. The ideal corpus is the smallest one that gives us the desired coverage across all participating implementations.

A new test would therefore ideally have an objective reason for inclusion, for example:
* it exercises a previously uncovered implementation or specification branch;
* it exposes a disagreement between implementations;
* it covers a specification boundary or previously missing semantic case; or
* it exercises an important exceptional value or condition, such as division by zero.

Tracking coverage across multiple implementations could therefore also give us a useful way to evaluate proposed additions to the shared conformance corpus.

Overall, I think CPS-0034 is a very useful direction. My main suggestion would be to define the **conformance and coverage target first**, and then choose the corpus size, generation strategy, and transmission mechanism needed to achieve that target efficiently.

PR #1245, opened 2026-08-18
CIP-0156 | Specify index count limit for `multiIndexArray`

Unisay:
Records the 1024-index limit the implementation imposes, as a failure condition alongside the existing out-of-bounds case, and corrects the costing claim: CPU is quadratic in the length of the index list, not linear in it. Without this the CIP and the implementation disagree on which programs are valid, since a call with 2000 indices is valid per the CIP and fails on chain. The implementation is already merged into plutus (https://github.com/IntersectMBO/plutus/pull/7869) and the boundary this amendment makes normative is covered by tests in https://github.com/IntersectMBO/plutus/pull/7908, so `Implementors:` is now filled in and the first acceptance criterion is ticked.

Proposal in this branch: [CIP-0156/README.md](https://github.com/Unisay/CIPs/blob/yura/cip-0156-index-count-limit/CIP-0156/README.md)

### DIFF
--- a/CIP-0156/README.md
+++ b/CIP-0156/README.md
@@ -81,7 +81,7 @@ Failing on first error mirrors `indexArray` and keeps the API simple.
 
 1. Add `multiIndexArray` to Plutus Core spec and runtime.
 2. Define preliminary cost model (linear in index list length for both CPU usage and memory usage).
-3. Write conformance tests covering valid and out-of-bounds cases.
+3. Write conformance tests covering valid, out-of-bounds, and over-the-index-count-limit cases, including the 1024-index boundary.
 4. Extend an E2E test suite to include `multiIndexArray` scenarios.
 5. Benchmark against manual `indexArray` loops to refine costing.
 6. Update formal documentation (`plutus-metatheory`, spec PDF).
--- a/CIP-0156/README.md
+++ b/CIP-0156/README.md
@@ -6,9 +6,11 @@ Status: Proposed
 Authors:
     - Yura Lazaryev <yuriy.lazaryev@iohk.io>
     - Philip DiSarro <info@anastasialabs.com>
-Implementors: []
+Implementors:
+    - IOG Plutus Team <https://iohk.io>
 Discussions:
     - Original PR: https://github.com/cardano-foundation/CIPs/pull/1050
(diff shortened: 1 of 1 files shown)


CIP-0156/README.md
```diff
@@ -6,9 +6,12 @@ Status: Proposed
 Authors:
     - Yura Lazaryev <yuriy.lazaryev@iohk.io>
     - Philip DiSarro <info@anastasialabs.com>
-Implementors: []
+Implementors:
+    - IOG Plutus Team <https://iohk.io>
 Discussions:
     - Original PR: https://github.com/cardano-foundation/CIPs/pull/1050
+    - CIP-0156 | Array-first argument order for multiIndexArray: https://github.com/cardano-foundation/CIPs/pull/1234
+    - CIP-0156 | Specify the index count limit for multiIndexArray: https://github.com/cardano-foundation/CIPs/pull/1245
```
rphair:
```suggestion
    - Array-first argument order: https://github.com/cardano-foundation/CIPs/pull/1234
    - Specify index count limit: https://github.com/cardano-foundation/CIPs/pull/1245
```
It's a general recommendation to keep these titles short as possible since they always display horizontally in narrow areas (though by default have often been assigned by full CIP name).

rphair (commented):
thanks @Unisay & also noting that it adds IOG Plutus Team as implementor.  Tagging our usual list of CIP-0156 reviewers — @zliu41 @colll78 @ana-pantilie @fallen-icarus @SeungheonOh — and marking this for `Triage` to confirm the update while we wait for Plutus experts to settle review on the change.

rphair (commented):
@Unisay somehow this got confirmed as an update without the usual editorial comment on "what to do next" — customarily "my job" after the meetings (though any editor or interested observer at the meeting can do it).

In any case this was confirmed without much ado at the CIP meeting before last.  I'm noticing only now this is highly relevant to costing but neglected to tag @kwxm in my original https://github.com/cardano-foundation/CIPs/pull/1245#pullrequestreview-4962325207: Kenneth, if you find the rationale here satisfactory then I would approve this myself & start moving it along towards merge.

@zliu41 @SeungheonOh (without tagging everyone else again) if we could also have your confirmation that this matches the overall Plutus strategy then we could do this more confidently.

Issue #977, opened 2025-02-01
CIP-0119: Extend dRep metadata for better discovery / adoption


MadOrkestra:
I will open this issue as a starting point for a discussion about an extension of the dRep metadata fields defined in CIP-119 for better discovery on dRep explorers. From an UX perspective, right now we have very little filter options to provide users on frontends to actually find a dRep that fits their needs, because freeform text fields don't do the trick. dRep discovery right now depends purely on recommendation or social media visibility and filter options on dRep explorers are mostly based on one thing: delegation. 

This has already led to a dRep distribution that is far from good, because instead of browsing through 500+ dReps people simply choose dReps with high delegation.

**Information that could be included for better discovery:**

- Location/Region
- Languages spoken (different from language used in the metadata file)
- Association with a stake pool
- Category/Areas of Interest (Yuta has assembled a list of dev dReps, Jenny is collecting female dReps, etc.)
- Ecosystem Roles (Stake Pool Operator, Intersect Member, etc.)
 
The challenge with most of these fields/arrays will be to agree on standards how to create and store this data, so all platforms can implement this for the creation of metadata files and use it as filter-options for displaying search results. A location/region field becomes utterly useless if you can throw "Europe", "Europa", "EU" and "European Union" in there for example.

This needs some work from all of us, at least if we want better dRep discovery and a more decentralized distribution among dReps - Otherwise we will continue the current popularity contest where dReps with existing platforms or already high delegation numbers continue to grow their delegation, leading to a governance system that - in reality - will be based on very few opinions instead of many.

We are already seeing different groups of people working around the proposed CIP119 (the "DRep Collective" is working on an un-official CIP that wants to move dRep data into NFTs, tempo.vote has gone through the lengths of scraping the web to come up with dRep locations, etc.), so I'd urge all of us to find an ecosystem wide solution for this problem, before this ends up in a scenario where platforms and projects start to throw random stuff into metadata files to fulfill their users needs without any other platform to be able to provide the same discovery. 

****References:****
- [CIP-0119](https://github.com/cardano-foundation/CIPs/tree/master/CIP-0119)
- [dRep delegation distribution on cgov.app](https://cgov.app/governance#voting-power-distribution)
- [dev dRep list on X created by Yuta](https://x.com/yuta_cryptox/status/1884434339343348056)
- [Unofficial "CIP148" by the Drep Collective](https://drep-eco.vercel.app/cip148)
[...]

PR #1251, opened 2026-08-23
CIP-0119 | Add discussion of possible metadata extension

rphair:
This discreetly — without modifying the CIP language, which could be arguable over the detail — adds a `Discussions:` link to highlight @Quantumplation's extraordinary exposition of how interested parties could go about extending the metadata vocabulary of CIP-0119 (& in fact any other metadata component) without the usually paralysing concerns over compatibility & consensus:
* https://github.com/cardano-foundation/CIPs/issues/977

I believe this will be especially important as DReps will need to be better differentiated by the community in the course of finding real solutions (in CIPs and/or community dialogue) to:
* [CPS-0033 | DRep Voting Power Concentration](https://github.com/cardano-foundation/CIPs/blob/master/CPS-0033/README.md)

(cc @maureenblack @daniellestanko-gov + @thenic95 — see @MadOrkestra's OP https://github.com/cardano-foundation/CIPs/issues/977#issue-2825153599 for a quick view of how a better metadata vocabulary would help with DRep influence distribution)

I've closed the issue above since the only way to apply its principles would be to produce a CIP with such an extended vocabulary: so this link will ensure this information isn't lost by fading into the issue queue (any more than it has already).

@Quantumplation @Ryun1 @Crypto2099 or anyone else: if it might not be considered enough just to have the `Discussions:` link there, we might also add some text to the CIP to develop the concept of extension & refer to the issue content more explicitly: if anyone offers this I'd be happy to add it to this PR.

(BTW I've also removed the tracking component from the Vimeo links: cc @Thomas-Upfield)

### DIFF
--- a/CIP-0119/README.md
+++ b/CIP-0119/README.md
@@ -8,8 +8,9 @@ Authors:
 Implementors: []
 Discussions:
     - Original PR: https://github.com/cardano-foundation/CIPs/pull/788
-    - Video: https://vimeo.com/912374177/0e9299fb5d?share=copy
-    - Video: https://vimeo.com/915297122/c39f4a739b?share=copy
+    - Video: https://vimeo.com/912374177/0e9299fb5d
+    - Video: https://vimeo.com/915297122/c39f4a739b
+    - Metadata extensions: https://github.com/cardano-foundation/CIPs/issues/977
 Created: 2024-02-07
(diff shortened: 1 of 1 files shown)


Ryun1 (approved):
Good add, thanks Robert!

perturbing (approved):

PR #1252, opened 2026-08-24
CIP-0171 | Add env field to the Aiken schema

nemo83:
## What

Adds an `env` field to CIP-0171's Aiken record layout (constructor 0), bumping that constructor's schema version from 1 to 2. Constructors 1–5 are untouched.

Also corrects the `parameter_list` production to `[ * bytes ]` as an erratum, and documents the residual reproducibility limits and the compatibility impact of the revision.

## Why

Aiken's `aiken build --env <module>` selects an environment module that is compiled into the script. Two builds of the same repository at the same commit with the same compiler version produce **different script hashes** if they used different `--env` values — so a record that omits it is not sufficient to reproduce the build it describes.

Without this field, a verifier that rebuilds and gets a mismatch cannot distinguish "the record is wrong" from "the record is incomplete". The revision makes constructor 0 a fixed six-element list, with empty bytes meaning the build used no `--env` flag — explicitly **not** the same as `--env default`, which must be recorded as `default`.

## On revising constructor 0 in place

The CIP states that a field-layout change takes a new constructor ID, and that IDs must not be reused. Both rules remain in force. This is a one-time exception, narrowed to constructor 0 and to this revision, on three grounds:

1. **Status** — this CIP is `Proposed`, not `Active`.
2. **Single implementation** — constructor 0 has one known implementation, the reference implementation named in the Implementation Plan.
3. **The prior records were replaced, not stranded** — every record that existed on mainnet under the previous layout has been re-published in the revised layout, with one exception: a duplicate submission for a repository and commit already covered by a re-published record, which carried no parameters and never completed verification.

The re-publication is complete and independently checkable on mainnet.

The document also now states that records which do not decode to their constructor's field layout MUST be **ignored** rather than treated as an error — label `1984` is shared with unrelated payloads, so non-conforming data under it is expected rather than exceptional.

Any future layout change, constructor 0 included, must use the standard mechanism and take a new constructor ID. This exception is not a precedent.

## Notes

- The `Amendment PR` link in the header is filled in once this PR has a number.
- Happy to split the `parameter_list` erratum into its own PR if reviewers would prefer it separate from the schema change.

---

[current CIP-0171](https://github.com/cardano-foundation/CIPs/blob/master/CIP-0171/README.md) — [rendered CIP-0171 as modified here](https://github.com/easy1staking-com/CIPs/blob/cip-0171-env-amendment/CIP-0171/README.md)

### DIFF
--- a/CIP-0171/README.md
+++ b/CIP-0171/README.md
@@ -115,8 +115,10 @@ to constructor 0 and to this revision only, granted on three grounds:
    it is the reference implementation named in the Implementation Plan.
 3. **The prior records were replaced, not stranded.** Every record that existed
    on mainnet under the previous Aiken layout has been re-published by its
-   author in the revised layout. No consumer is left holding a record that this
-   document no longer describes.
+   author in the revised layout, with one exception: a duplicate submission for
+   a repository and commit already covered by a re-published record, which
+   carried no parameters and never completed verification. No consumer is left
+   holding a record that this document no longer describes.
 
 Consumers MUST parse constructor 0 as a fixed six-element list per this
 document. Records under any constructor that do not decode to that
--- a/CIP-0171/README.md
+++ b/CIP-0171/README.md
@@ -109,11 +109,9 @@ departs from the two rules stated immediately above:
 Both rules remain in force. This revision is a **one-time exception**, narrowed
 to constructor 0 and to this revision only, granted on three grounds:
 
-1. **Status.** This CIP is `Proposed`, not `Active`. Proposed standards are
-   explicitly revisable.
-2. **Single implementation.** Constructor 0 has one known implementation, and
+1. **Single implementation.** Constructor 0 has one known implementation, and
    it is the reference implementation named in the Implementation Plan.
-3. **The prior records were replaced, not stranded.** Every record that existed
+2. **The prior records were replaced, not stranded.** Every record that existed
    on mainnet under the previous Aiken layout has been re-published by its
    author in the revised layout, with one exception: a duplicate submission for
    a repository and commit already covered by a re-published record, which
--- a/CIP-0171/README.md
+++ b/CIP-0171/README.md
@@ -9,13 +9,14 @@ Implementors: []
 Discussions:
     - Forum Discussion: https://forum.cardano.org/t/cip-idea-smart-contract-source-code-verification-metadata/152403
     - Original PR: https://github.com/cardano-foundation/CIPs/pull/1136
+    - Amendment PR: <url-pending>
 Created: 2026-01-19
 License: CC-BY-4.0
 ---
 
 ## Abstract
 
-This CIP defines a decentralized, on-chain metadata standard for linking Cardano script hashes to their published source code. Using transaction metadata label `1984`, anyone can publish records containing the source repository URL, git commit hash, compiler type and version, and script parameters. Verifiers can then independently compile the source and confirm that the resulting script hash matches an on-chain script, establishing a verified link between the deployed script and its source code origin.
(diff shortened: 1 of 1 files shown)


CIP-0171/README.md
```diff
@@ -9,13 +9,14 @@ Implementors: []
 Discussions:
     - Forum Discussion: https://forum.cardano.org/t/cip-idea-smart-contract-source-code-verification-metadata/152403
     - Original PR: https://github.com/cardano-foundation/CIPs/pull/1136
+    - Amendment PR: https://github.com/cardano-foundation/CIPs/pull/1252
```
rphair:
```suggestion
    - Add env field: https://github.com/cardano-foundation/CIPs/pull/1252
```
In the short space permitted we always try to make these links as specific as possible (to form something like a changelog).

CIP-0171/README.md
```diff
@@ -93,8 +94,64 @@ This design allows each compiler's metadata schema to evolve independently. If a
...
+1. **Status.** This CIP is `Proposed`, not `Active`. Proposed standards are
+   explicitly revisable.
+2. **Single implementation.** Constructor 0 has one known implementation, and
+   it is the reference implementation named in the Implementation Plan.
+3. **The prior records were replaced, not stranded.** Every record that existed
+   on mainnet under the previous Aiken layout has been re-published by its
+   author in the revised layout, with one exception: a duplicate submission for
+   a repository and commit already covered by a re-published record, which
+   carried no parameters and never completed verification. No consumer is left
+   holding a record that this document no longer describes.
```
rphair:
```suggestion
1. **Single implementation.** Constructor 0 has one known implementation, and
   it is the reference implementation named in the Implementation Plan.
2. **The prior records were replaced, not stranded.** Every record that existed
   on mainnet under the previous Aiken layout has been re-published by its
   author in the revised layout, with one exception: a duplicate submission for
   a repository and commit already covered by a re-published record, which
   carried no parameters and never completed verification. No consumer is left
   holding a record that this document no longer describes.
```
`Proposed` standards are only considered revisable with the consent (or due diligence to obtain it) — requested in review — of any concurrent implementations.  Therefore, your original point # 2 (renumbered here, "single implementation") is the <ins>real</ins> reason why this CIP can be implicitly be changed.

rphair (commented):
@nemo83 thanks for the upkeep on this specification: marking for `Triage` at next week's CIP meeting: https://hackmd.io/@cip-editors/142

I agree with your comment below that this would be free to change liberally because yours is the only current implementation.  However I'm not not putting this at `Last Check` — as we normally might for such an update — because we would also categorically check some other features:
* @colll78 and maybe @Ryun1 will likely want to check on the approach to versioning since it's introduced to the CIP with this update.
* @KtorZ might be interested in the Aiken-specific architecture.
* However unlikely this may be: if _anyone_ is working on some unknown implementation of this CIP, please post here and we would be happy for your review as well.

With supportive review all around (or any concerns fixed) we might approve & move this up to `Last Check` at the next meeting.

nemo83:
Thanks @rphair happy to discuss this change at the next CIP meeting. I'm fairly sure nobody is implementing their custom version to submit cip171 txs onchain but they are using the website instead.

Happy to take the required time to update this cip.

CIP-0171/README.md
Ryun1, line 8:
```suggestion
Implementors:
  - [uplc-link](https://github.com/easy1staking-com/uplc-link)
```
maybe worth adding?

rphair (approved):
@nemo83 this was confirmed as a valid update, as expected, at the CIP meeting today with an expected fast track towards `Last Check` at the following meeting.

I had mistaken your Discord posting "back in 2 weeks" for an indication you would be away in the meantime 😅 but now I realise you were simply talking about the next meeting date: so have marked it `Last Check` in the assumption that you will resolve any review issues before then: https://hackmd.io/@cip-editors/143

CIP-0171/README.md
rphair: @nemo83 I believe it's _essential_ to add... though I won't commit this, leaving the label & format to you (and to our CIP validator running below).

perturbing (approved):
Given the above confirmation from @rphair 👍

nemo83:
thanks everyone

PR #1253, opened 2026-08-25
CIP-0170 | Add digest computation specification

### LINKED ISSUE #1249: CIP-0170 | `d` has no canonical serialisation, so indexer key ordering breaks verification
## Summary

CIP-0170 defines `d` on an `ATTEST` transaction as *"The digest of the data being signed in the CESR qb64 variant"* ([spec, `ATTEST` field definitions](https://github.com/cardano-foundation/CIPs/tree/master/CIP-0170)), and its worked example computes that digest over the application metadata at the sibling label — *"The CESR digest of the data at label 1447 is …"*. The spec never fixes a **canonical serialisation** for that data; the words "canonical", "serialisation" and "key order" do not appear in it.

Because JSON object key order is not preserved by the tooling verifiers actually use, two honest parties reading the same transaction can compute two different digests. Validity ends up a property of *how the verifier fetched the transaction* rather than of the attestation itself.

We hit this building an `ATTEST` implementation, on preprod, with a transaction that was never tampered with.

## What we observed

Submitted an `ATTEST` whose sibling payload was constructed in this key order:

```
d, credentialType, schemaVersion, issuedAt, holderRef, evidenceDigest
```

**The on-chain CBOR preserves that order exactly.** Verified byte for byte against the raw transaction via Koios `/tx_cbor`. The transaction builder does not reorder either — confirmed separately by encoding a map with deliberately non-canonical keys and inspecting the resulting CBOR. **The chain and the encoder are both faithful.**

**A JSON indexer API returns a different order.** The same transaction through Koios `/tx_metadata`:

```
d, issuedAt, holderRef, schemaVersion, credentialType, evidenceDigest
```

That is keys ordered by length, then bytewise. The mechanism is documented on both sides:

- `cardano-db-sync` stores transaction metadata in [`tx_metadata.json` as **`jsonb`**](https://github.com/IntersectMBO/cardano-db-sync/blob/master/doc/schema.md).
- PostgreSQL: *"`jsonb` does not preserve white space, **does not preserve the order of object keys**, and does not keep duplicate object keys"*, and *"since shorter keys are stored before longer keys"* ([docs](https://www.postgresql.org/docs/current/datatype-json.html)).

The original order is therefore destroyed **at write time**, so every consumer of a db-sync-backed database necessarily sees the normalised order, not what was submitted. The values are untouched; only the order differs.

Consequence: a digest computed at issuance over the construction order cannot be re-derived by a verifier reading the JSON API, and a completely valid attestation reports INVALID. Ours did. It is on preprod as [`f690640a…`](https://preprod.cardanoscan.io/transaction/f690640a6a77a17926a0f0ca272870f486c90cbabe7f07035c42e3de574d4de1), left there deliberately as a reproducible example.

The corrected one, digesting a canonical form, verifies: [`9f84c95d…`](https://preprod.cardanoscan.io/transaction/9f84c95d81d3ce338c3af0aecf2c0c87dc10ad3b75c1d77772cec7b22cd66444).

## Why this is a spec problem rather than an implementation problem

An implementer following CIP-0170 literally will digest the data as they constructed it. Nothing in the spec suggests that is unsafe, and it round-trips correctly through raw CBOR. It only fails once a verifier reads through an indexer — which is what essentially every verifier will do. The failure is silent, produces a valid-looking transaction that can never verify, and is permanent once on chain.

Scope, stated precisely: this affects `ATTEST`, which is the only transaction type carrying a digest over metadata. `AUTH_BEGIN` / `AUTH_END` are not affected in the same way — their `m` block is reordered identically, but nothing digests it, and `c` is a byte-stream rather than a map. *(Corrected shortly after filing; an earlier revision of this issue overstated the scope.)*

## Suggested resolution

Specify the canonical form `d` is computed over. Any of these would close it:

1. **Mandate a canonical serialisation for the digested data**, stated in the field definition for `d`. Ordering keys by length then bytewise has the practical advantage of matching what db-sync-backed APIs already return, so the round trip becomes a no-op for the common tooling.
2. **Require the digested payload to be self-addressing** (an ACDC/KERI SAID over a defined canonical form), making the digest reproducible from the payload alone.
3. **Define `d` over the raw CBOR bytes** at the sibling label. This is more practical than it may sound: db-sync already stores the untouched bytes alongside the JSON in `tx_metadata.bytes`, so indexers are capable of exposing them — though verifiers would need an API that does.

Whichever is chosen, an explicit warning that indexer JSON does not preserve map key order would save implementers a great deal of debugging.

## Reference implementation

Working `ATTEST` implementation with the canonicalisation, a verifier, and the full write-up: https://github.com/Wistkey/acespeak-cip0170-spike (`READINESS.md` §3.2). Happy to open a PR against the CIP text if a preferred resolution is indicated.

Kammerlo:
Fixes #1249
Specifies how the `d` digest of an `ATTEST` transaction is computed, which the CIP previously left open (no canonical serialisation, see #1249):
- If the transaction carries application metadata under another label, `d` MUST be the CESR Blake3-256 digest of the CBOR bytes of the metadatum value at that label, byte-identical to the transaction's auxiliary data - not a JSON or re-serialised form.
- If no application metadata is present, the data referred to by `d` is defined by the use case, which must specify how a verifier obtains it and what encoding is digested.
- Verifiers MUST NOT recompute `d` from indexer JSON (db-sync `jsonb` reorders keys at write time); the spec lists sources for the raw bytes (tx CBOR, db-sync tx_metadata.bytes, Blockfrost /metadata/cbor).
- Adds a self-contained test vector (positive and a reordered negative) and a rationale paragraph.

This is a clarification only: no change to the metadata format or CDDL, v.v stays 1.0. It matches the existing reference implementations (e.g. [Reeve's KeriService](https://github.com/cardano-foundation/cf-reeve-platform/blob/b46af3799fc87011f5191d0f8f2c30138cc663f7/blockchain_publisher/src/main/java/org/cardanofoundation/lob/app/blockchain_publisher/service/KeriService.java#L40): `blake3_256(CBOR(map))`).

---

Rendered newly added section (most of the update): [Digest computation](https://github.com/Kammerlo/CIPs/blob/docs/update-cip-170/CIP-0170/README.md#digest-computation)

### DIFF
--- a/CIP-0170/README.md
+++ b/CIP-0170/README.md
@@ -148,6 +148,8 @@ What `d` refers to depends on whether the transaction carries application metada
 
 The digest is computed with the algorithm identified by the CESR derivation code (Blake3-256, code `E`, RECOMMENDED), encoded as a CESR primitive in qb64 form, placed in `d` and anchored as the seal of the signer's KEL event. Because `d` covers only the attested data and not label `170` itself, there is no circularity: the payload is serialised and digested first, then the `170` entry is added.
 
+Because the KEL anchor is created before the transaction exists and cannot be retracted once written, implementations SHOULD extract the encoded metadatum bytes from the built transaction and confirm they digest to `d` before submitting. If they differ, the attestation is not verifiable and can only be corrected by anchoring a new event and publishing a new transaction.
+
 > [!WARNING]
 > Verifiers MUST NOT recompute `d` from a JSON representation of the metadata. JSON is not a faithful representation of the on-chain bytes: common indexers normalise map key order at write time (for example `cardano-db-sync` stores `tx_metadata.json` as PostgreSQL `jsonb`, which does not preserve key order), so the information needed to recompute the digest is destroyed before any JSON API returns it. Verifiers MUST obtain the raw metadatum bytes, for example from the transaction CBOR itself (a local node, Koios `/tx_cbor`), from `cardano-db-sync` `tx_metadata.bytes`, or from Blockfrost `/txs/{hash}/metadata/cbor`.
 
--- a/CIP-0170/README.md
+++ b/CIP-0170/README.md
@@ -117,7 +117,7 @@ To create a persistent signature over data with KERI, signers can anchor a diges
  The relevant data recorded for each event includes:
 - **t** — A transaction type of `ATTEST` is to create a verifiable record.
 - **i** — The identifier of the signer in the CESR qb64 variant.
-- **d** — The digest of the data being signed in the CESR qb64 variant.
+- **d** — The digest of the attested data in the CESR qb64 variant. If the transaction carries application metadata under another label, the digest MUST be computed over the CBOR encoding of the metadatum value at that label, byte-identical to the bytes in the transaction's auxiliary data — see [Digest computation](#digest-computation). The digest algorithm is identified by the CESR derivation code; implementations MUST support Blake3-256 (code `E`) and SHOULD use it when creating attestations. The value of `d` MUST be identical to the seal anchored in the KEL event at sequence number `s`.
 - **s** — The sequence number of the KERI event, encoded as a hex string.
 - **v** - Version of the CIP. KERI and ACDC version isn't needed here.
 If the KEL of identifier `i` contains an event at sequence number `s` which has a seal value of `{ d: "{{digest}}" }`, it serves as cryptographically verifiable proof that the data was effectively signed by the controller.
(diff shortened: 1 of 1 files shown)


styamanda (commented):
Thanks @Kammerlo — this is more thorough than I expected, and it resolves both things I asked about: it pins down *which* bytes (the metadatum value only, not the label/value pair or the whole map), and it states plainly that JSON representations cannot be used.

**I verified the test vector independently** and every value in it checks out:

| | |
|---|---|
| Vector 1 CBOR | decodes to exactly the JSON shown, in construction order ✅ |
| Vector 2 CBOR | same content, keys length-then-bytewise ✅ |
| Blake3-256 of vector 1 | `ea4c2099…d75f391` — matches ✅ |
| CESR qb64 | `EOpMIJmAaiP4cZgmDkg8rVtl8YU4dDYf_gxrK2sNdfOR` — matches ✅ |
| Vector 2 qb64 | `EIswjSN1u14y36RGf6TKm-z65R0rMMLjQHnjXofImTmZ` — matches ✅ |

Computed with `signify-ts`' `Diger`, independently of whatever produced them. Having a vector where the *wrong* answer is also given is a nice touch — an implementer who lands on `EIswjSN1…` immediately knows what they did.

Two suggestions, one substantive.

**1. Warn about the ordering between anchoring and building.** The text says to digest "the bytes that will be (or were) submitted", which is right, but understates a trap. `d` has to be anchored in the KEL *before* the transaction is built, and the KEL is append-only — so if the builder's encoding differs at all from what was digested, the attestation is permanently unverifiable and a sequence number has been burned on a bad anchor. Most builders take a JSON-ish object and encode it themselves, so the issuer does not directly control the emitted bytes; `schemaVersion: 1` as `01` versus `1901` is enough to break it.

Something like:

> Because the KEL anchor is created before the transaction exists and cannot be retracted, implementations SHOULD extract the encoded metadatum bytes from the built transaction and confirm they digest to `d` before submitting.

That is a cheap check and it converts a permanent failure into a caught one.

**2. Minor: the worked example is now less self-consistent than the test vector.** The `ATTEST` example at label `1447` still carries `"{{someApplicationMetadata}}"` alongside a concrete digest, so a reader cannot check it — while the new §Digest computation gives fully reproducible values. Using the test vector's payload in the worked example would make the whole document verifiable end to end.

Neither blocks merging from my side. We will implement against this text — our current implementation digests a canonicalised JSON form, which was a workaround for exactly this gap — and I will report back on whether the spec was sufficient to implement from without rediscovering anything.

styamanda:
Implemented against this text and it holds up — here is the report I promised.

**Result:** a conformant attestation on preprod, verifying against the new rule: [`046c0ce9…`](https://preprod.cardanoscan.io/transaction/046c0ce93e03ed6d6f709d8930f2611943b5d93c202aca6263bd38b9a5ca39dc). `d` is the Blake3-256 digest of the CBOR encoding of the metadatum value at the application label, read back from the transaction's raw CBOR. Implementation: https://github.com/Wistkey/acespeak-cip0170-spike (`src/cardano/cbor.ts`).

The spec text was sufficient to implement from. I did not have to guess at anything, and the test vector caught my encoder before the chain did.

**One thing I would add, because it nearly bit me.** The text says to digest the bytes that will be submitted, which is correct — but "encode the value to CBOR" is not a single well-defined operation in practice, and the choice is easy to get wrong invisibly. Concretely, in `cardano-multiplatform-lib`:

- `TransactionMetadatum.from_json(...)` with the detailed tagged schema **preserves** key order
- `encode_json_str_to_metadatum(..., BasicConversions)` **sorts keys lexicographically**

Both are reasonable-looking ways to turn a JSON object into a metadatum, they produce different bytes, and only one matches what the transaction builder emits. Picking the wrong one gives you a digest that no transaction will ever match — and because `d` is anchored in the KEL *before* the transaction is built, and the KEL is append-only, the mistake is already permanent by the time you find out.

That is what the SHOULD I suggested earlier is really protecting against, and I would now put it more strongly than "compare before submitting":

> Implementations SHOULD NOT independently re-serialise the payload to compute `d`. Where the digest must be computed before the transaction exists, implementations SHOULD extract the encoded metadatum bytes from the built transaction and confirm they digest to `d` before submitting, since the KEL anchor cannot be retracted if they differ.

Ours now builds, extracts the metadatum back out of the built transaction, and refuses to submit unless it digests to the anchored value.

**On the two suggestions from my earlier review:** the anchor-ordering one is above. The other — aligning the worked example at label `1447` with the test vector so the whole document is reproducible — I still think is worth doing, but it is cosmetic and I would not hold the PR for it.

For what it's worth as regression material, there are now two superseded attestations of ours on preprod that look well-formed and can never verify: [`f690640a…`](https://preprod.cardanoscan.io/transaction/f690640a6a77a17926a0f0ca272870f486c90cbabe7f07035c42e3de574d4de1) (digested a JSON serialisation) and [`9f84c95d…`](https://preprod.cardanoscan.io/transaction/9f84c95d81d3ce338c3af0aecf2c0c87dc10ad3b75c1d77772cec7b22cd66444) (digested a canonicalised JSON form). Both are left on chain deliberately and are useful for testing that a verifier rejects what it should.

Thanks for turning this round so quickly — happy to see it merge.

rphair (commented):
@Kammerlo @styamanda this is scheduled to confirm as a CIP update (`Triage`) at the next CIP meeting — https://hackmd.io/@cip-editors/142 — although if both of you agree that the changes are finalised I'll approve it myself & promote this to `Last Check` at that meeting.

I'll also tag the other editors for advance review as soon as that happens.  Of course any pending or even hypothetical implementors of CIP-0170 are also encouraged to comment here in the meantime.

Kammerlo:
@styamanda Thanks that's a good point! I added a sentence to clarify that. 
If you are good with it now, then @rphair can promote it.

rphair (approved):
yes @Kammerlo since that appears to resolve the last reservation: pending any further testing as this develops & @styamanda submitting a further update if & when needed.  This will be ready for `Last Check` at the next CIP meeting.

@styamanda if there is anything further you think could/should change to save you & perhaps other implementors the trouble later, please feel free to suggest further changes in the meantime.

Kammerlo:
Thanks a lot @rphair From my point of view this update was important and a really good extension from @styamanda 
I wrote the update according to how we implemented it anyway so turning the unkown into something explicit. 
From my point of view it is good and matches what I had in mind anyway.

perturbing (approved):
Nice addition that pins it 👍

PR #1254, opened 2026-08-29
CIP-???? | Hydra Head UTxO Recognition

skoniog:
**Proposal:** [CIP-hydra-head-utxo-recognition/README.md](https://github.com/skoniog/CIPs/blob/hydra-head-utxo-recognition/CIP-hydra-head-utxo-recognition/README.md)

## Summary

A **Wallets**-category proposal defining how a CIP-30 wallet can recognize, verify, and display funds committed to a Hydra head — closing the "ghost UTxO" problem where wallets hide or flag head-committed funds as phishing because they cannot confirm ownership against L1. This is widely cited as the single largest adoption blocker for consumer-facing Hydra applications.

Requirements were gathered in [cardano-scaling/hydra#2818](https://github.com/cardano-scaling/hydra/issues/2818) (opened by me); this proposal is written against its acceptance criteria. Scope is **recognition and display only** — in-head signing, transaction construction, and head lifecycle are explicitly deferred to a separate future wallet-participation proposal.

## Key design points

- **No new attestation scheme and no new trust assumption.** The proposal builds on an artifact Hydra already produces: every confirmed snapshot is multi-signed by *all* head participants, whose Ed25519 verification keys are published on-chain in the open head's datum, and the same signature is what the L1 head validator itself verifies. A wallet checking it runs the same verification the chain runs. The operator serving the proof is untrusted — it can cause "unverifiable", never forged balances.
- **Precise specification against Hydra protocol version 2** (current deployed scripts, hydra 2.3.0): CDDL for the on-chain datums, the exact signable message encoding, the proof bundle (hydra-node's existing `GET /snapshot`), and the BLS12-381 accumulator binding of individual UTxOs to the signed snapshot (full recomputation or KZG membership proof).
- **Trustless pending-deposit state**: deposit datums embed the deposited outputs in full, so the "my funds vanished" moment is closed from L1 data alone, even with no operator reachable.
- **Mandatory display states** implementing the issue's core requirements: verified funds shown as legitimate balances; unverifiable funds *flagged*, never silently hidden.

## Open design areas flagged for review

1. The signable-encoding freeze is framed as a compatibility commitment *requested of* the Hydra team (coordination via hydra#2818).
2. Discovery (head ID → proof endpoint) proposes a transaction-metadata hint on deposits; this is the area most open to editor input.
3. A membership-proof HTTP endpoint on hydra-node is identified as the one missing implementation piece for the lighter verification path.

The CIP number is left as `?` for editor assignment, per CIP-0001.

## Disclosure

Drafted with AI assistance (Claude); all protocol facts were verified against the Hydra source repository at version 2.3.0, and the disclosure is also carried in the document's Copyright section.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


rphair (commented):
thanks @skoniog & team: this is a well-presented draft & looks like a useful proposal.  Please also remove the initially self-assigned number `198` from the internal review introduction here (instead of "tentatively CIP-0198" you could say "CIP number pending review"):
* https://github.com/cardano-scaling/hydra/issues/2818#issuecomment-5461241435

When done with that internal review please take this out of `Draft` mode and/or post here in the CIP PR thread and we'll begin our initial CIP review with `Triage`, and feel free to invite anyone working in this field for initial response to this content in the meantime (cc @Crypto2099).

PR #1257, opened 2026-08-31
CIP-0198? | Nested Transactions - Service Layer

polinavino:
CIP-118 lets a user author a sub-transaction that does not balance on its own, e.g. having more of a user-defined token in the inputs than the outputs, and vice-versa for ADA (such type of sub-transaction is called a *babel fee offer*).  Another party can include this sub-transaction in a fully valid  top-level transaction, which can then be posted on-chain.  The Nested Transactions CIP specifies the required ledger changes to support this, but the ledger rules alone are not enough to use it: a sub-transaction that does not balance cannot be posted as-is, so the feature does nothing until some other party finds it, judges it worth completing, and builds the top-level transaction that carries it. That party, and the means of reaching it, are entirely off-chain, and no such mechanism exists. This CIP specifies such a mechanism, including all the relevant on-chain infrastructure: the offer- and interest-communication protocol, service discovery and customization, and batch construction. For each of these functionalities, we indicate what functions each of the three roles involved in this infrastructure performs. The three roles (which a single party may combine) are: anonymous publishers of sub-transactions (wallets), services (babel aggregators of sub-txs into top-level txs), and relay nodes. 

[Rendered](https://github.com/polinavino/CIPs/blob/refs/heads/polina/babel/CIP-0198/README.md)


CIP-XXXX/README.md
```diff
@@ -0,0 +1,2071 @@
+---
+CIP: "?"
+Title: Babel Fee Offers — Off-Chain Transmission and Service Layer for Nested Transactions
```
```suggestion
Title: Nested Transactions - Service Layer
```
Even if not anticipating a "family" of CIPs as described above, this would solve the usual brevity problem (since this would be our longest CIP title, and they are often rendered in line horizontally) while pairing naturally with the title of CIP-0118 _and_ leaving such a "family" to be continued by other authors & projects.
* This "family" idea is suggested in CIP-0001 where there are common areas of work emergent or anticipated: [Naming CIPs with similar subjects](https://github.com/cardano-foundation/CIPs/blob/master/CIP-0001/README.md#naming-cips-with-similar-subjects)
* The delimiter character in this case is a keyboard minus sign: never a higher-order character like an em dash as originally posted.

CIP-XXXX/README.md
```diff
@@ -0,0 +1,2071 @@
...
+Category: Network
+Status: Proposed
+Authors:
+    - Polina Vinogradova <polina.vinogradova@iohk.io>
+    - William Wolff <william.wolff@iohk.io>
+    - Dana Alibrandi <dana.alibrandi@iohk.io>
+    - Nicolas Henin <nicolas.henin@iohk.io>
+Implementors: []
+Discussions:
+    - CPS-0015 Intents for Cardano: https://github.com/cardano-foundation/CIPs/pull/779
```
rphair: Would you consider this CIP a `Solution to:` CPS-0015 (and therefore include in the metadata as such)?

CIP-XXXX/README.md
```diff
@@ -0,0 +1,2071 @@
...
+    - CIP-0131 Transaction swaps: https://github.com/cardano-foundation/CIPs/pull/880
+    - CIP-0089 beacon tokens: https://github.com/cardano-foundation/CIPs/pull/466
+    # TODO before submission: this proposal's own PR, and the CIP-118 thread it
+    # follows from.
```
```suggestion
Discussions:
    - Original PR: https://github.com/cardano-foundation/CIPs/pull/1257  
    - CPS-0015 | Intents for Cardano (closed): https://github.com/cardano-foundation/CIPs/pull/779
    - CIP-0089 | Beacon Tokens: https://github.com/cardano-foundation/CIPs/pull/466
    - CIP-0118 | Nested Transactions: https://github.com/cardano-foundation/CIPs/pull/862
    - CIP-0131 | Transaction swaps (closed): https://github.com/cardano-foundation/CIPs/pull/880
```
Corrects typography expected by our validator and capitalisation, and adds the missing threads.  We also have to emphasise that CIP-0131 & CPS-0015 are only valid as "discussions" and not as document references.

CIP-XXXX/README.md
```diff
@@ -0,0 +1,2071 @@
...
+Created: 2026-08-14
+License: CC-BY-4.0
+---
+
+## Abstract
+
+[CIP-118][] lets a user author a sub-transaction
```
rphair: Please search & replace `CIP-118` with `CIP-0118` especially since all other CIPs quoted have the canonical 4-digit numbers.

CIP-0198/README.md
```diff
@@ -0,0 +1,2071 @@
...
+for ADA (such type of sub-transaction is called a *babel fee offer*). 
+Another party can include this sub-transaction in a fully valid 
+top-level transaction, which can then be posted on-chain. 
+The Nested Transactions CIP specifies the required ledger changes to support
+this, but the ledger rules alone are not enough to use it: a sub-transaction
+that does not balance cannot be posted as-is, so the feature does nothing
+until some other party finds it, judges it worth completing, and builds the
+top-level transaction that carries it. That party, and the means of reaching it, are
+entirely off-chain, and no such mechanism exists.
+This CIP specifies such a mechanism, including all the relevant on-chain
```
rphair, line 39:
```suggestion
entirely off-chain, and no such mechanism exists.

This CIP specifies such a mechanism, including all the relevant on-chain 
```
I thinking breaking the very long paragraph into 2 parts (roughly: problem + solution) will be more accommodating for readers.

CIP-0198/README.md
```diff
@@ -0,0 +1,2071 @@
...
+
+### Architecture
+
+Boxes are machines. What shares a box is expected to share a host. A service
+must run beside a full
+node and talks to it locally. Nothing else needs a node: a relay is required to
+hold no chain state, and a wallet publisher may be light.
+
+```text
+  ┌─ USER'S MACHINE ─────────┐        ┌─ RELAY HOST ──────────────┐
```
rphair, line 105: We are using Mermaid more & more often now that it's been fully supported in GitHub and `cips.cardano.org` for a while.  On other sites that display CIP content downstream I've seen `text` presentation get garbled (sometimes displaying proportionally spaced = unreadable) but even on these sites unrendered Mermaid code could be grabbed & parsed in a Mermaid previewer.

(Note editors will happily accept the text-only graphics as well.)

CIP-XXXX/README.md
```diff
@@ -0,0 +1,2071 @@
...
+- **batch construction** — [batch construction](#batch-construction), the
+  [constraint language](#constraint-language) a service plans against, and the
+  [liquidity strategy](#liquidity-strategy) that funds it;
+- **service customization** — the [service profile](#service-profile) and
+  [price hints](#price-hints) an operator publishes, within the
+  [protocol constants](#protocol-constants) everyone shares.
+
+### Architecture
+
+Boxes are machines. What shares a box is expected to share a host. A service
```
```suggestion
Outer boxes are machines. What shares a box is expected to share a host. A service
```
a nitpick: currently the inner `cardano-node` element is also in a box 😅

rphair (commented):
Thanks @polinavino @will-break-it @dalibran @nhenin for such a ground-breaking submission: it's on the CIP meeting agenda tomorrow for `Triage` and please feel welcome to attend (which may be helpful if we end up talking about the CIP scope): https://hackmd.io/@cip-editors/142

I believe it would be a good start to get review from (any time, before or after the meeting):
* @fallen-icarus: deep CIP-0118 review, Beacon Tokens, applications of P2P swaps
* @lehins @carlostome @williamdemeo: Ledger types & encoding herein
* @jpraynaud: the DMQ integration
* @colll78: general architecture, security, efficiency

My main question (proceeding also to a `Title:` discussion below, I hope): it seems this architecture would be applicable to other types of swap offers in addition to babel fee offers; so:
* Is this CIP about the whole `Service Layer for Nested Transactions` and therefore should have a more general title; rather than leading with "Babel Fee Offers" which might only be one of its use cases?
* If so, then would it be more clear to have "Babel Fee Offers" listed as its own category of the specification: perhaps with other sections for other kinds of swap offers?

I am certainly not demanding this, but this seems like such a promising architecture that the use cases of this service — and the service itself — might each be more clear if described _in its own CIP_: <ins>if</ins> you also think that would be easier for a general audience to understand as you may be gathering popular support for work on implementations; e.g. ultimately CIPs titled:
* `Nested Transactions` (CIP-0118 itself)
* `Nested Transactions - Service Layer` (the commonly usable part of this proposal)
* `Nested Transactions - Babel Fee Offers` (the part of this draft for that use case)
* `Nested Transactions - Swap Market` (future / hypothetical)

... following the convention of other related CIPs such as [Web-Wallet Bridge](https://cips.cardano.org/search?q=web-wallet%20bridge).

CIP-XXXX/README.md
rphair: p.s. tentatively applied to this PR title but certainly not "set in stone".

CIP-XXXX/README.md
polinavino: thank you for the feedback! ive got a few things i need to finish this week and then i'll be off for a week after, so it will be addressed after that!

CIP-0198/README.md
```diff
@@ -0,0 +1,2072 @@
...
+at the ledger level.
+
+**Sybil attacks on the registry.**
+Peers are drawn uniformly from the registry, so an attacker's chance of
+surrounding a given service is its share of registry entries (i.e. on-chain 
+registration tokens). 
+Too high a deposit excludes the small operators
+that make the market competitive. Too low allows the attack to happen.
+The required deposit and the number of peers drawn are dependent variables in this
+calculation, and neither means anything chosen alone.
```
fallen-icarus: Can you use a reputation system for relays? Like if a relay passes along a sub-tx that you end up dropping, why would you not trust this relay less? AFAIU it is similar to hot/cold peers in the mainnet gossip's protocol. We can use the deposit to limit the number of relays, and then we can track their reputations for the rest. Low reputation relays are ignored by peers. This gives adaptive sybil resistance which is arguably more realistic than perfect sybil resistance.

Lmk if I misunderstood something.

CIP-XXXX/README.md
```diff
@@ -0,0 +1,2072 @@
+---
+CIP: "?"
```
```suggestion
CIP: 198
```

rphair (commented):
@polinavino we confirmed this as a candidate at the CIP meeting today based on understanding that — regardless of any further partitioning of this material as suggested above (e.g. with "babel fees" factored out into a different CIP) — that `this` PR will be for a CIP describing the essential Service Layer: acknowledged all around to be important to the Cardano economy & abundantly described here already.

We would look forward to hearing your and your team's response to that refactoring suggestion when you get back... in the meantime please rename the containing directory to `CIP-0198` and update the "Rendered" link in your OP 🎉

CIP-0198/README.md
polinavino: theres a proposed additions bit (admittedly misplaced) that discusses that relays and services can in the future have private blacklists from which they drop connections. do you mean that there should also be a discussion of a public shared reputation system?

polinavino:
I agree that refactoring is a good idea, but perhaps for now i can take the discussion of usecases or business cases out of this PR and work on those separately (perhaps even better as CPSs?). 

However, the CIP lists some routing/filter keys tailored for certain usecases (babel fees, DApp subsidization, priority brokerage, etc.) and removing those would significantly reduce the usefulness of the service in its first version.

CIP-0198/README.md
fallen-icarus: Private should be enough, but shouldn't the behavior be standardized? Why is this marked as a "future" feature? Why wouldn't this be part of the MVP since it pertains to security? Would simple rules not work?

polinavino: yeah, it can be included. ill try that now

CIP-XXXX/README.md
rphair: resolved with https://github.com/cardano-foundation/CIPs/pull/1257/changes/1ca521b393a36d492c1f18ae8ad903618fd39112 & https://github.com/cardano-foundation/CIPs/pull/1257/changes/eec8af271f6c90857e013653de447ddf3239fc39

CIP-XXXX/README.md
rphair: resolved in a recent commit 🙏

CIP-0198/README.md
polinavino: i added blacklisting and details about relay peer selection

colll78:
I spent some time thinking about how to approach this. The design I ended up with, which I found to have the best tradeoffs is:

**Build the service layer as a dedicated, Nostr-compatible relay network for Babel offers.**

Essentially we reuse Nostr’s small, already-implemented relay protocol to create a purpose-built Cardano offer network:

- A wallet publishes the same signed offer event to several independent Babel relays.
- Each relay stores it only for the offer’s bounded lifetime.
- A fulfiller bot queries the currently stored offers and then remains subscribed to the live stream.
- Bots merge results from several relays, deduplicate them, validate them against Cardano, apply their own profitability criteria, and submit the completed top-level transaction.
- Anyone can publish, subscribe, run a bot, or operate another compatible relay. There is no canonical server or permissioned registry.

Nostr’s base protocol already supplies event publication, filtered historical queries, a transition from historical results to a live subscription, and publication acknowledgements. Its optional extensions cover relay capability discovery, expiration, proof-of-work, query-completeness hints, and client-to-relay or relay-to-relay set reconciliation.

For pure Babel Fees, the common interoperability layer only needs to move one immutable object:

“Here is a signed, incomplete Cardano sub-transaction that remains fillable until slot/time X.”

The broadcast layer does not need to determine:
- which bot should receive it;
- what a token is worth;
- whether the offer is profitable;
- which offers should be batched together;
- whether the offer was ultimately confirmed;
- or the canonical order in which offers arrived.qq  
Those are decisions for relayers / executor bots, who will due to the financial nature of this, want to be able to fully customize the criteria / logic around this. 

That means the off-chain data structure is essentially a TTL-bounded, replicated grow-only set of signed offers

While an offer is live:
- duplication is harmless;
- reordering is harmless;
- a relay inventing data is harmless after validation;
- relay omission affects liveness and market access, but not ledger safety;
- Cardano resolves conflicting offers and final settlement.

image


### Offer event format
The envelope format as a regular stored Nostr event:
```JSON
{
  "id": "<nostr-event-id>",
  "pubkey": "<fresh one-time transport pubkey>",
  "created_at": 1788530000,
  "kind": "<BABEL_OFFER_KIND>",
  "tags": [
    ["n", "cardano-mainnet"],
    ["v", "1"],
    ["x", "<subtransaction-txid>"],
    ["h", "<blake2b-256-of-exact-envelope-bytes>"],
    ["expiration", "<transport-expiration-unix-time>"],
    ["nonce", "<pow-nonce>", "<pow-target>"]
  ],
  "content": "<base64url-without-padding-of-envelope-cbor>",
  "sig": "<nostr-event-signature>"
}
```

Meanings:
`x` represents the Sub-transaction TxId, it is used for grouping all representations of the same semantic offer
`h` represents envelope_id, it is the Blake2b-256 hash of the envelope bytes, used for deduplicating identical subtx-and-witness payloads. 

This is required because witnesses and auxiliary data can be malleated without changing the sub-transaction body’s TxId. A relay should therefore not discard every later event merely because it shares an offer_id; that could allow a malicious first arrival with useless witnesses to suppress a later usable representation. It may safely deduplicate identical envelope_id values. Bots should group variants by offer_id, perform full chain-aware validation, and retain the smallest valid representation.


**Publisher algorithm**

A wallet:

- Constructs and signs a CIP-118 sub-transaction.
- Locally resolves its inputs and validates that the intended imbalance is correct.
- Encodes the canonical envelope.
- Derives offer_id and envelope_id.
- Generates a one-time Nostr key.
- Performs the required admission proof-of-work.
- Signs one outer event.
- Sends that identical event to several independent relays.
- Records each relay’s OK response.
- Keeps the event locally and retries alternative relays until the acknowledgement policy is met or the offer expires.

**Bot snapshot and live subscription**
A bot opens a Nostr subscription of the form:
```JSON
[
  "REQ",
  "active-mainnet",
  {
    "kinds": ["<BABEL_OFFER_KIND>"],
    "#n": ["cardano-mainnet"],
    "#v": ["1"],
    "since": "<now - MAX_OFFER_TTL - REORG_GRACE>"
  }
]
```

The same subscription gives both required behaviors:
- the relay first sends all matching stored events;
- it sends EOSE;
- the subscription remains open and new matching events arrive live.

**Bot-side matching and fulfillment**
Each bot maintains its own verified offer database:
```
offers
  offer_id
  envelope_id
  source_relays
  validity_upper_bound
  spend_inputs
  offered_value
  needed_value
  scripts
  declared_ex_units
  serialized_size
  first_seen
  last_validated_tip
  status
```

After receiving an event, the bot:
- verifies the outer event and envelope;
- resolves all Cardano inputs;
- confirms the inputs are currently unspent;
- checks witness completeness and authorization;
- evaluates the sub-transaction’s own scripts;
- derives consumed − produced for every asset;
- checks its private price/risk policy;
- places conflicting offers into the same input-conflict set;
- selects one or more compatible offers;
- builds a complete top-level transaction;
- revalidates immediately against the latest tip;
- submits it;
- follows Cardano for inclusion and rollbacks.

For example, a bot locally derives:
```
offer imbalance:
    +5 Foo
```
It then decides whether receiving five Foo is worth paying the Cardano transaction fee and providing the collateral input. 

Another option would be to use Celestia as the offer broadcasting layer, and in many ways it's an improvement over Nostr, but the critical tradeoff is that offer publication would require the publisher to pay fees in TIA, an asset that they are unlikely to have. Under the assumption that users don't have TIA, there would have to be a batcher / relayer who pays the Celestia publication fees on behalf of users and submits their offer messages to Celestia, at which point the whole system has centralized anyway or if we have multiple Celestia offer publication batchers then we just have just kicked the can up the road (we would need to define some message protocol / coordination between offer publication batchers etc).

PR #1262, opened 2026-09-07
CIP-0167 | Align mempool transaction format with trailing in-block `is_valid`

Lucsanszky:
## What

Amends CIP-0167 to reflect the in-block transaction format proposed by the amended CIP-0176 (#1261): instead of a block-level `invalid_transactions` index list, each transaction in a block carries its own `is_valid` flag as the trailing element, after the auxiliary data.

- The Abstract no longer states that blocks are unaffected because the flag is "stored separately from the transaction"; within blocks, the flag is the trailing element of each transaction, set by the block producer.
- The proposed CDDL rule is renamed to `mempool_transaction` and gains a second alternative that accepts the legacy flag position (only with the value `true`) for backwards compatibility during the transition to the era that adopts this CIP; encoders must always produce the flag-less form, and the legacy alternative will be removed in the following era.
- The relationship to the in-block `block_transaction` format (CIP-0176) is spelled out.

## Why

Motivated by IntersectMBO/cardano-ledger#5912.

Submitted in tandem with the CIP-0176 amendment (#1261); the two CIPs describe the two sides of the same transaction wire-format split.

### DIFF
--- a/CIP-0167/README.md
+++ b/CIP-0167/README.md
@@ -16,7 +16,7 @@ License: CC-BY-4.0
 ## Abstract
 
 We propose removing the `isValid` boolean from the CBOR encoding of standalone transactions (e.g. for mempool).
-This would not affect the serialization of the transactions within blocks, since isValid flag is already stored separately from the transaction
+Within blocks, transactions continue to carry an `is_valid` flag, but as the trailing element of each transaction, set by the block producer (see CIP-0176); it is not part of the serialization used by the transaction author.
 
 ## Motivation: Why is this CIP necessary?
 
@@ -40,14 +40,31 @@ transaction = [transaction_body, transaction_witness_set, bool, auxiliary_data/ 
 
 The proposal is to change it to:
 ```cddl
-transaction = [transaction_body, transaction_witness_set, auxiliary_data/ nil]
+mempool_transaction =
+  [transaction_body, transaction_witness_set, auxiliary_data/ nil]
+  / [transaction_body, transaction_witness_set, true, auxiliary_data/ nil]
 ```
+
+The second alternative exists only for backwards compatibility during the transition to the era that adopts this CIP (transactions in a mempool at the hard fork boundary may still carry the flag): the legacy `is_valid` position is accepted on decoding, but only with the value `true`.
+Encoders must always produce the first alternative.
+In the following era the legacy alternative will be removed, so it is strongly recommended to encode transactions without the flag.
+
+Inside a block, a transaction is serialized with a trailing `is_valid` flag appended by the block producer, after all the fields supplied by the transaction author:
+
+```cddl
+block_transaction =
+  [transaction_body, transaction_witness_set, auxiliary_data/ nil, bool]
+```
+
+The in-block format is specified by CIP-0176; it is shown here only to clarify the relationship between the two formats.
 
 ## Rationale: How does this CIP achieve its goals?
 
 Removing the `isValid` flag from standalone transaction serialization simplifies the wire format without changing consensus or ledger semantics.
 
 The trusted local client submission use case might be better expressed in a different way (for example, as a Node-to-Client submit parameter), rather than embedded in the transaction bytes.
+
+Keeping the block-producer-supplied `is_valid` flag as the trailing element of the in-block transaction (rather than as a block-level index list) keeps the author-supplied fields as a contiguous prefix and provides a natural position for future block-producer-supplied fields, such as the proposed `feeChangeAmount`.
 
 ## Path to Active
 


rphair (commented):
thanks @Lucsanszky; this & the concurrent update both look realistic and the maintenance on these CIPs is much appreciated.

@lehins (cc @carlostome @williamdemeo) the CIP editors can approve this & the concurrent update once confirmed consistent with the ledger spec, but also because of implications between Tx and block body I think we should tag reviewers & advocates of both [CIP-0167](https://github.com/cardano-foundation/CIPs/tree/master/CIP-0167) and [CIP-0176](https://github.com/cardano-foundation/CIPs/tree/master/CIP-0176) to be aware of possible impact: @ch1bo @colll78  @Quantumplation @Ryun1 @teodanciu

If neither disputed nor explicitly confirmed this would naturally get tagged `Confirmed` from `Triage` at the next CIP meeting (https://hackmd.io/@cip-editors/143) with `Last Check` for merge at the meeting after that.

rphair:
@lehins given that these 2 update propositions are so closely related, does your approving review
* https://github.com/cardano-foundation/CIPs/pull/1261#pullrequestreview-5146305642

also mean _this_ change looks good to you?

lehins (approved):
> also mean this change looks good to you?

It most certainly does look good! 👍

rphair (approved):
Given that both this & its concurrent PR are confirmed all around from the Ledger side, we only have the remaining question of possible ecosystem impact, as tagged already to a handful of experts above (sorry @fallen-icarus I didn't think of tagging you before; only now does it seem like this might affect transaction dependencies).

And since there's been no objection yet upon presentation this week, we would normally take the next CIP meeting (https://hackmd.io/@cip-editors/143) as an opportunity to find who else in Consensus (cc @perturbing), applications (cc @colll78 @Crypto2099), etc. might object to either of these CIPs.

Therefore, since that's what we would be doing at `Last Check` anyway, I'm approving this (based on apparent sensibility + consistency with that Ledger already plans to do) & marking it as such.  That will give us another week for any objections & other editors (cc @Ryun1 @ch1bo) to point out any problems or other stakeholder that we should wait until getting a review from.

CIP-0167/README.md
```diff
@@ -40,15 +40,32 @@ transaction = [transaction_body, transaction_witness_set, bool, auxiliary_data/
...
+The second alternative exists only for backwards compatibility during the transition to the era that adopts this CIP (transactions in a mempool at the hard fork boundary may still carry the flag): the legacy `is_valid` position is accepted on decoding, but only with the value `true`.
+Encoders must always produce the first alternative.
+In the following era the legacy alternative will be removed, so it is strongly recommended to encode transactions without the flag.
+
+Inside a block, a transaction is serialized with a trailing `is_valid` flag appended by the block producer, after all the fields supplied by the transaction author:
+
+```cddl
+block_transaction =
+  [transaction_body, transaction_witness_set, auxiliary_data/ nil, bool]
+```
```
ch1bo, line 57: This should work fine for Leios. We just need to remember (and ideally specify in [CIP-164](https://github.com/cardano-foundation/CIPs/blob/master/CIP-0164/README.md), that the `block_transaction` is hashed and thus endorsed by Leios - *not* the `mempool_transaction`. In fact, it makes our life easier as we don't need to transmit validity via endorser blocks.

The same applies for [CIP-178](https://github.com/cardano-foundation/CIPs/pull/964): the validity must be part of the preimage.

ch1bo (approved):

CIP-0167/README.md
```diff
@@ -40,15 +40,32 @@ transaction = [transaction_body, transaction_witness_set, bool, auxiliary_data/
 
 The proposal is to change it to:
 ```cddl
-transaction = [transaction_body, transaction_witness_set, auxiliary_data/ nil]
+mempool_transaction =
+  [transaction_body, transaction_witness_set, auxiliary_data/ nil]
+  / [transaction_body, transaction_witness_set, true, auxiliary_data/ nil]
 ```
 
+The second alternative exists only for backwards compatibility during the transition to the era that adopts this CIP (transactions in a mempool at the hard fork boundary may still carry the flag): the legacy `is_valid` position is accepted on decoding, but only with the value `true`.
```
ch1bo, line 48: Should: Signal the alternative as "deprecated" somehow and add it to the path to active

Maybe put a deprecation notice as a comment on the Dijkstra CDDL?

I also noticed that no transaction creation library (e.g. [`cardano-api`](https://github.com/IntersectMBO/cardano-api) or [`cardano-serialization-lib`](https://github.com/emurgo/cardano-serialization-lib)) is on the path to active.

Maybe we should add: open issues with transaction creation libraries to drop `isValid` flag once deprecation is active (after PV12)

rphair:
thanks @ch1bo ... good to know that we can merge this promptly but we can do it @ the meeting so we can also get a proper discussion on the comments you made (cc @Lucsanszky to possibly update the text before merge).

perturbing (approved):

rphair:
With no further reservations expressed, merging with counterpart update https://github.com/cardano-foundation/CIPs/pull/1261.

PR #1265, opened 2026-09-10
CIP-0200? | Transaction Urgency Signalling On Linear Leios

willjgould:
This PR adds CIP-urgency-signalling, which specifies a way for users to flag their transaction as urgent and pay an increased fee for statistically lower latency to inclusion, along with dynamic pricing, on linear-Leios.

The CIP addresses [CPS-prioritising-urgent-transactions](https://github.com/willjgould/CIPs/blob/urgency-signalling/CPS-0031/README.md) (**ed:** at submission time, `Last Check` before merge: https://github.com/cardano-foundation/CIPs/pull/1194).

[Rendered](https://github.com/willjgould/CIPs/blob/cip-urgency-signalling/CIP-0200/README.md)


CIP-urgency-signalling/README.md
```diff
@@ -0,0 +1,1601 @@
...
+  - CPS discussion: https://github.com/cardano-foundation/CIPs/pull/1194
+  - Original PR: https://github.com/cardano-foundation/CIPs/pull/1265
+Solution To:
+  - CPS-0031?: https://github.com/cardano-foundation/CIPs/pull/1194
+Created: 2026-06-24
+License: CC-BY-4.0
+---
+
+<details>
+<summary><strong>Table of contents</strong></summary>
```
rphair, line 22: As @nhenin would well know, both TOC's and section numbers are considered unmaintainable & a general disaster by the CIP editors (most vocally, me).  All Markdown outlines are inherently visible in a sidebar — as easily as clicking an outline icon on GitHub — on both our major CIP platforms = GitHub + `cips.cardano.org`.

I have a feeling AI assistance is being used here (with no prejudice if so, as also discussed earlier) to collate this material into sections...  but this does not imply that such a workflow would be used (or would even be welcome) as this document is further reviewed, after it is merged, or when it might need to be modified by others.  At best it is redundant to the outline structure itself and therefore a duplication of effort that will eventually (and repeatedly) go astray from the primary source of its own Markdown headings.

**TL;DR** With appreciation for your conscientious work I will recommend, until/unless voted down by a majority of CIP editors + reviewers, that the TOC for this & any similar document be deleted.

CIP-urgency-signalling/README.md
```diff
@@ -0,0 +1,1601 @@
...
+
+The ledger enforces the urgency signalling rule: every transaction in a valid Ranking Block must carry a fee that covers the urgent quote for that block. In simulation under severe congestion, the mechanism preserves more urgent-class transaction value than linear-Leios with today's flat fee. Retained value means the modelled gross transaction value that remains at inclusion, before fees. Urgent-class retained value improved across most simulated loads. At light load, the mechanism slightly reduces overall retained value, because transactions on the standard path wait longer while Endorser Blocks fill. The Rationale gives exact figures.
```
rphair, line 90:
```suggestion
The ledger enforces the urgency signalling rule: every transaction in a valid Ranking Block must carry a fee that covers the urgent quote for that block. In simulation under severe congestion, the mechanism preserves more urgent-class transaction value than linear Leios with today's flat fee. Retained value means the modelled gross transaction value that remains at inclusion, before fees. Urgent-class retained value improved across most simulated loads. At light load, the mechanism slightly reduces overall retained value, because transactions on the standard path wait longer while Endorser Blocks fill. The Rationale gives exact figures.
```
At least in the CIP process (both documents & social discussion e.g. on Discord), this is the first time we've ever seen the term `linear Leios` hyphenated: and so I assume it should be changed canonically to remove the hyphen.
* In fact it also appears in _this_ document without the hyphen & so at least needs to be canonicalised toward one or the other... even if now asserted that the hyphenation ***is*** canonical.

I don't think it's that important whether "linear" is capitalised or not, since Leios is already a recognised proper name.  But as you know, capitalising it as `Linear Leios` would make it seem like _that_ is the objective: rather than `Leios` being the objective & "linear" simply being a means to an end (as I understand it be).

However decided, this term spelling & capitalisation needs to be standardised all the way through the document.

CIP-urgency-signalling/README.md
```diff
@@ -0,0 +1,1601 @@
+---
+CIP: "?"
+Title: Transaction Urgency Signalling On Linear-Leios
```
rphair, line 3:
```suggestion
Title: Urgency Signalling in Linear Leios
```
(note Linear here would be capitalised for the same reason as Signalling — but without the hyphen as previously suggested)

Generally we are going for as short a title as possible without readers assuming the wrong scope.  I'll retitle the PR the same as this recommended change but the editors would follow your lead about what will serve the community best... without elements that might be considered redundant:
* "Transaction" in Urgency Signalling appears to be implied from candidate CPS-0031 discussion (https://github.com/cardano-foundation/CIPs/pull/1194): cc @fallen-icarus for possibility another point of view on this.

The rest of the title I think well defines the scope of the CIP with respect to its planned implementation & therefore will remain valid regardless of how the draft may change in response to Leios team review... I'd just be reluctant to include `Transaction` because it's not used consistently in the corresponding CPS.

CIP-urgency-signalling/README.md
```diff
@@ -0,0 +1,1601 @@
...
+
+Full details, including method, configs, per-load tables, paired seed deltas, and figures: [preliminary experiment report](https://github.com/input-output-hk/tiered-pricing/blob/bd3de91fc0afe6cea83ff7842c6ba1e6c2aa892a/docs/phase-2/preliminary-experiment-report.md).
+
+### Prototype
+
+https://github.com/user-attachments/assets/6a4ef69a-516f-4517-bfbd-d7b8a97b09cf
```
rphair, line 1470: I think I can speak for @aldo555 (maintainer of `cips.cardano.org`, as far as I currently know) to say this might not properly embed anywhere but GitHub.  In fact, the context for rendering it properly may disappear as soon as this PR is merged... and so it's not likely to work sustainably even on our own repository.

However the video itself is useful to a broad audience and so of course should be included as a Reference... but using a link in a standard, long-term location.  If there is an "assets" section on the Tiered Pricing or Leios repository then that would be ideal... @ch1bo any ideas for a proper place to put this?

rphair (commented):
Thanks @willjgould @polinavino @nhenin @pagio for another impressive round of progress in this field & for documenting it so comprehensively.

I expect this would be confirmed & assigned a number at the next CIP meeting from `Triage` (https://hackmd.io/@cip-editors/143) because it seems obvious that some form of this work will be published & the work itself will be inevitable.

The editors would expect a long review process for this and I hope @ch1bo @berewt would be able to get us started & tag others who are interested: from `Consensus` as well as other disciplines.  (Editors may not be able to contribute much technical review on our own expect for @perturbing who I hope would contribute regularly.)

There are some points for editorial review (not only some format issues but also re: significance to concurrent work) & I'll try to break them up below so they can be reviewed & discussed individually.

Besides those, we would look to @lehins & team for a dedicated review of the [Ledger characteristics](https://github.com/willjgould/CIPs/blob/cip-urgency-signalling/CIP-urgency-signalling/README.md#ledger) including [rule changes](https://github.com/willjgould/CIPs/blob/cip-urgency-signalling/CIP-urgency-signalling/README.md#ledger-rule-changes). 🙏

CIP-urgency-signalling/README.md
rphair: Title declared OK all around at the CIP meeting today.

CIP-urgency-signalling/README.md
```diff
@@ -0,0 +1,1601 @@
+---
+CIP: "?"
```
rphair:
```suggestion
CIP: 200
```

rphair (commented):
@willjgould @polinavino @nhenin @pagio this was confirmed a candidate at the CIP meeting today... please rename the containing directory to `CIP-0200` and update the "Rendered" link accordingly 🎉

Most of the discussion was to confirm the scope & title are a good match — as we understood it, about an implementation _in_ Linear Leios: with protocol v13 as a prerequisite, confirmed by @ch1bo — rather than this CIP defining a characteristic _of_ Leios.

CIP-urgency-signalling/README.md
willjgould: I have no opinion on this, it was added by request and AI use made it effortless, so I'll happily remove it.

rphair: Resolved with https://github.com/cardano-foundation/CIPs/pull/1265/changes/846667e34acf8d6c9709be06ffed61659896b439.

CIP-urgency-signalling/README.md
rphair: Resolved with https://github.com/cardano-foundation/CIPs/pull/1265/changes/1f1a393eb75acd0aff54ead40a9a46a574b78d5e.

CIP-urgency-signalling/README.md
rphair: Since the term `Transaction` has been kept in the latest edited title I'll assume that's what we'll use going forward (cc @willjgould @polinavino @fallen-icarus) and I do think that keeps the proper scope in mind: rather than some readers wondering what kid of "urgency" the CIP might be talking about... so I'm updating the PR title to be the same as the current `Transaction Urgency Signalling On Linear Leios`.

rphair:
@willjgould thanks for all your updates supporting CIP validation & CIP-0001 compliance; there's just one [current validator issue](https://github.com/cardano-foundation/CIPs/actions/runs/35196551307/job/105121036590?pr=1265) that can be fixed by renaming this directory from `CIP-urgency-signalling` to `CIP-0200` as requested in https://github.com/cardano-foundation/CIPs/pull/1265#pullrequestreview-5215421061. 🙏

fallen-icarus:
This CIP is very long and dense so perhaps I missed it: why is the fee change *required*? Is there an incentive reason, or is it strictly maximizing retained value? IIUC it only retains ~2-3% more. If that understanding is correct, I'm not sure the gains are worth the sacrifice.

Does the fee change create a double-satisfaction problem for plutus scripts? What if a tx is withdrawing funds from the same account which receives the fee change? What does the script context tell the plutus script in that case?

---
(editorial question)

Is there a way to make this CIP less dense for the lay reader without sacrificing substance? This CIP is exhaustive which is great, but it is very difficult to find what is important. For instance, do experiments belong in the spec or should they be in extra sections at the end? Should the spec discuss alternatives? This was a problem with the Linear Leios CIP where it wasn't obvious that only ~50% of RBs can actually have EBs which created confusion.

rphair:
@fallen-icarus re: the editorial question (cc @willjgould): co-author @nhenin has been helpful in this regard for other standards documents, e.g. like you can see here with supplementary, much more detailed material moved into a subdirectory (in this case, `CPD`) with its own `README` file and supporting assets:
* https://github.com/cardano-foundation/CIPs/tree/master/CPS-0021

willjgould:
@fallen-icarus regarding the value point, the retained value gains depend on the volume, so at low load the improvement is indeed modest (even very slightly negative under certain circumstances), but in some scenarios, when load is very high, it's more like 15-20% relative improvement.

> why is the fee change required?

Imagine there's high enough volume for dynamic pricing to kick in and you're submitting a transaction, if there's a queue in the mempool your transaction might not get into the next block. If that's the case, the fee might rise before your tx gets included. As such, a "buffer" amount needs to be supplied (so you can cover the fee increase) unless you _do_ want the transaction to be kicked out when the price rises. Most of the time, the total fee supplied will exceed the fee required at inclusion. That remainder is what we want to be refunding.

> Is there a way to make this CIP less dense for the lay reader without sacrificing substance? This CIP is exhaustive which is great, but it is very difficult to find what is important.

This is something we've discussed...it's challenging. You're right, and I agree with you in substance, I'd much prefer to have a shorter document ("if I had more time, I would have written a shorter letter" and all that). I did write a couple of sections at the start ("Common Misconceptions" and "A note on linear Leios") to try to get across some key points up-front. I could write a "TL;DR"-type-section after the abstract if people think it's worthwhile.

I should also mention: the start of "Specification" and the table in "The recommended construction" were intended to cover most of the key parts.

> For instance, do experiments belong in the spec or should they be in extra sections at the end?

Good point. Most of the experiments were actually put towards the end, so I'll moved the rest and refer to them instead.

> Should the spec discuss alternatives?

Good question. I think a discussion of alternatives should enhance the reader's understanding of the proposed solution and how the authors got there, rather than just existing for the sake of it. I'm happy to take your advice onboard here: if you don't think the discussions of alternatives succeed at that we can remove them (or move them to the end).

fallen-icarus:
The question isn't flat vs dynamic. I accept the dynamic fee. The question is: why does the dynamic fee require a fee refund? We can have a dynamic minimum fee where the over-payment is not refunded.

If the typical latency for my tx is <2 minutes in the worst case, it is unlikely that the quoted fee will move much between when I submit my tx and when it finally gets included. I can very easily look up the current quote and either set it exactly (expecting it not to move) or give a slight buffer of 1-2 updates. I think the CIP says the price can only move at most 6.5% per quote update. So I can deliberately choose to pay `x`, or `x * 1.065`, or `x * 1.13`. As a tx builder, we are talking about a difference of 0.03-0.1 ADA. That's not much of a burden.

If there is a sudden demand shock, the delayed quote updates could be a systemic risk. Allowing people to outbid each other directly would be more anti-fragile. Would it not? Wouldn't you get more retained value too?

---

I like the alternatives, but perhaps it should be in a later section. The way I view research papers, the *Results* section explains the experiment in detail but it is very dry. The *Conclusion* section then discusses the higher level interpretation. When I read papers, I generally skip to the conclusion and only read the results when necessary. In essence, I'm suggesting front-loading the *conclusion* parts and leaving the *results* to an appendix section. So it's effectively:
- Abstract
- Motivation (aka Why?)
- Spec - just the implementation instructions. It can link to appendix sections for further reading
- Rationale (aka Conclusion)
- Extra sections such as Experiments, Alternatives, etc.

I'm open to other ideas here though. I don't want anything removed, just reorganized.

polinavino:
> Does the fee change create a double-satisfaction problem for plutus scripts? What if a tx is withdrawing funds from the same account which receives the fee change? What does the script context tell the plutus script in that case?

You're right. fixing this now.

polinavino:
@fallen-icarus so i looked over the design, and sharpened the changes to the `produced = consumed` calculation specification, see the latest [commit](https://github.com/cardano-foundation/CIPs/pull/1265/commits/0e12e14c0b18a206a443424f5c77569f7eb46cc6)  

Specifying the portions of the fee (going to treasury, going to fee pot, going to change address) as explicit summands in the calculation, separate from regular payments to/from accounts, should suffice to avoid accounting issues.

As for double satisfaction for scripts, this should also not be a problem - scripts dont even see the amount of change going to a change account address, and they also do not see the amounts in the accounts. So, they have no way to use change data as "constraint satisfaction", and it wont exacerbate the double satisfaction problem.

fallen-icarus:
@polinavino What about CIP-159 account balance intervals?

> The question is: why does the dynamic fee require a fee refund? We can have a dynamic minimum fee where the over-payment is not refunded.

Re-iterating my question in case it was missed. These plutus fixes wouldn't be required if this CIP didn't have the fee refund.

polinavino:
> Re-iterating my question in case it was missed. These plutus fixes wouldn't be required if this CIP didn't have the fee refund.

I get how this looks convoluted, and it's true that it prevents plutus scripts from having the total Delta for an individual account (the actual _amount_ in accounts cannot be shown to scripts anyways, even without tiered pricing). However, the tiered pricing change to Plutus script context is minimal, and the only thing this design does not allow scripts to do is base validation decisions on the amount of change given (which cannot be predicted at construction time). 

Even when it comes to basing validation decision on the amount given by the transaction to the treasury, it is reasonable that the unpredictable "fee change" portion of it should not be considered.

fallen-icarus:
@polinavino I appreciate the response but that didn't answer my question on why the fee refund mechanism is required. In fact, as I pointed out above, the fee market would possibly be more anti-fragile without it.

It doesn't look to me like you guys simulated a dynamic fee market with no refund mechanism. Perhaps I'm not understanding your experiments, but it looks like you only tested static fee vs dynamic fee with refund. What about dynamic fee without a refund?

The dynamic fee sets the floor but there is no ceiling. SPOs would be incentivized to include the higher paying ones first, even if the profits are indirect through staking rewards.

If this CIP is actually better without the fee refund, then the complexity introduced by the refund mechanism is unjustified, even if that complexity is small from the ledger's perspective.

colll78:
>If this CIP is actually better without the fee refund, then the complexity introduced by the refund mechanism is unjustified, even if that complexity is small from the ledger's perspective.

Agreed. The complexity here isn't just on the ledger, it's on all downstream tooling. Collateral return outputs are already an unjustified complexity nightmare on the ecosystem. We should do anything and everything possible to avoid repeating that disaster by introducing anything similar unless the concrete benefits justify it (a very high bar to meet). 

Generally, we should be trying to consolidate / shrink the tx representation, instead of growing it. We should aim to express more with less.

pagio:
>  I think the CIP says the price can only move at most 6.5% per quote update. So I can deliberately choose to pay x, or x * 1.065, or x * 1.13. As a tx builder, we are talking about a difference of 0.03-0.1 ADA. That's not much of a burden.

@fallen-icarus  : Isn't the difference price depended ? Meaning, if prices are high the difference will also be high in terms of ADA (and thus the money possibly lost).

polinavino:
> The dynamic fee sets the floor but there is no ceiling. SPOs would be incentivized to include the higher paying ones first, even if the profits are indirect through staking rewards.
If this CIP is actually better without the fee refund, then the complexity introduced by the refund mechanism is unjustified, even if that complexity is small from the ledger's perspective.

The dynamic fee sets the floor and the ceiling for the amount that goes into the fee pot. Any accidental overpayment goes either to the fee change address or the treasury, never to the reward mechanism. So, the change for overpayment scheme is explicitly in place to ensure that overpayment does nothing for the user because the incentives don't exist to put higher-fee-paying txs above others. The reason for this is that allowing users to set the fee going into the fee pot effectively defines not a 2-tier fee system, but an arbitrary-tier fee system. The mempool would not be capable of supporting this at high transaction volumes. The fully dynamic fee approach is better suited to support competition for UTxOs, as well as for allowing priority transactions to evict lower-fee transactions from the mempool by invalidating them (an orthogonal design). The tiered fee mechanism we propose here is not built with these goals in mind.

However - the tiered pricing mechanism would still work same way if we were to remove the change mechanism, with the exception of the user not getting change. That is to say, it appears that no incentives (of parties _other than that transaction author_, who is slightly disincentivized by the potential for overpayment) we rely on here would likely be altered by removing the change mechanism.

fallen-icarus:
@pagio Yes, it is ADA price dependent. But we have adjustable parameters to keep the fees low in USD terms. And the dynamic fee mechanism can help here by lowering the minFee when demand is low. The fee refund mechanism isn't required if the dynamic fees update properly.

AFAIU Midnight has a "congestion component" to its tx fee that works like a dynamic fee, but it does not have a fee refund.

---

> The reason for this is that allowing users to set the fee going into the fee pot effectively defines not a 2-tier fee system, but an arbitrary-tier fee system. The mempool would not be capable of supporting this at high transaction volumes.

@polinavino Okay, so it is because the mempool can't handle constantly evicting/replacing txs with higher paying ones. By having the refund, you can still have the mempool use FIFO for the urgent txs. That answers my question. Thanks!

What counts as "high transaction volumes"? IIRC Linear Leios is targeting ~100-200 TPS. Would that already be a problem for the mempool if it had to constantly reorganize around higher paying txs?

willjgould:
> The complexity here isn't just on the ledger, it's on all downstream tooling.

@colll78 (or @fallen-icarus) would it be possible for you to give some examples showing where and how this would be problematic downstream?

fallen-icarus:
My issue isn't with the tooling. It's with whether the complexity is justified as a stop-gap. AFAIU the motivation was based on this [survey](https://product.cardano.intersectmbo.org/business-interviews/builders/#33-latency-and-wallet-ux-for-trading-use-cases) which says:

>  5-10 second ... confirmation is described as "simply too slow"

yet my stance is that this was selection bias of the typical crypto echo chamber. Less than 3% of BTC is participating in DeFi despite robust institutional and retail adoption. I'm building stuff to target (some of) the 97% sitting on the sidelines (eg, Pogun and the DeFi Kernel) and they do not care about the 5-10 seconds. They are used to >24 hours in TradFi! They care about UTxO contention. This CIP doesn't help with that at all.

It seems you wanted to keep the mempool changes minimal for now which is why you have this fee refund mechanism; but if the mempool needs to be changed in the future anyway to attract the sidelined participants, what point is there in this CIP? If we end up dropping the fee refund mechanism later once the mempool is upgraded, all this complexity would be a complete waste. We don't need this change ASAP so I would personally rather effort be spent on solving the mempool problem than implementing this CIP to appease a handful of users who can already be appeased with L2s like Gummiworm and Midgard instead.

TLDR I think this CIP was motivated by a flawed study and doesn't actually solve a pressing need which makes the temporary complexity unjustified, especially since blockchains (and their off-chain tooling) must be backwards compatible forever.

PR #1267, opened 2026-09-10
CIP-0118 | Remove restriction on provenance of scripts

carlostome:
This PR removes CIP-0118's restriction on the provenance of scripts for the top-level transaction in legacy mode.

This change reflects the current implementation and formal specification of Cardano ledger for Dijkstra era.


rphair (approved):
thanks @carlostome; seems a routine update as per concurrent updates & we can merge this more readily once it's seconded from the Ledger side (cc @williamdemeo @aniketd).

@lehins this change doesn't seem to have a behavioural consequence — at least not in any current implementations — but please advise the editors if so & we'll change the review posture accordingly.

BTW I'm making the title a bit shorter since in general these titles don't have to be so specific when referring to such an atomic change... and for practical purposes (like meetings / agendas) shorter names are better even for short-term (update) entities.

`Last Check` for next meeting: https://hackmd.io/@cip-editors/143

perturbing (approved):
Nice!

@colll78 lets see what you can cook with this similar to your other weird plutus tricks ;)

PR #1264, opened 2026-09-10
CIP-0159 | Simplify details about account balance intervals

aniketd:
Simplify text, remove unclear statements and unnecessary mentions.

---

(updated numbered list content here: [Account Balance Intervals](https://github.com/aniketd/cardano-cips/tree/aniketd/cip-159/CIP-0159#account-balance-intervals))

PR #1268, opened 2026-09-11
CPS-0037? | Application and transaction friction in minUTxO implementation

nhenin:
This CPS documents application and transaction-builder friction in Cardano's current minUTxO implementation, where the ada required to protect persistent UTxO state is carried as ordinary output value.

It describes seven use cases and their current workarounds, and sets goals for reducing the remaining friction while preserving protection against accumulation of live UTxO state. Four open questions address capacity-price changes, compatibility with deployed contracts, incentives to remove unused outputs, and the organization of explicit deposits if a solution uses them.

[Read the rendered CPS](https://github.com/nhenin/CIPs/blob/codex/cps-minutxo-submission/CPS-%3F%3F%3F%3F/README.md).

### DIFF
--- a/CPS-????/README.md
+++ b/CPS-????/README.md
@@ -8,7 +8,8 @@ Authors:
   - Will Gould <will.gould@iohk.io>
   - Polina Vinogradova <polina.vinogradova@iohk.io>
 Proposed Solutions: []
-Discussions: []
+Discussions:
+  - Original PR: https://github.com/cardano-foundation/CIPs/pull/1268
 Created: 2026-07-28
 License: CC-BY-4.0
 ---
--- a/CPS-????/README.md
+++ b/CPS-????/README.md
@@ -1,0 +1,335 @@
+---
+CPS: "?"
+Title: Application and transaction-builder friction in the current minUTxO implementation
+Category: Ledger
+Status: Open
+Authors:
+  - Nicolas Henin <nicolas.henin@iohk.io>
+  - Will Gould <will.gould@iohk.io>
+  - Polina Vinogradova <polina.vinogradova@iohk.io>
+Proposed Solutions: []
+Discussions: []
+Created: 2026-07-28
+License: CC-BY-4.0
+---
+
+## Abstract
+
+Cardano's current minUTxO rule requires every new transaction output to contain a
+minimum amount of ada based on a fixed per-output overhead and the output's serialised
+size. This size-sensitive requirement provides **economic coordination** for
+persistent UTxO state. However, the required ada is represented as ordinary output
+value, which creates additional friction.
+
+The minimum is determined from the complete serialised `TxOut`, including any datum
+or reference script. The ada used to satisfy it remains ordinary coin in `TxOut.Value`,
+alongside the ada and native assets the application intends the output to carry.
+The same ada can serve both purposes. The ledger records no separate deposit or
+backing balance, and control of all ada follows the output's spending condition.
+
+This coupling creates **accidental implementation friction** beyond the **necessary
+resource friction** of protecting the live UTxO set. It can prevent a standalone
+payment below minUTxO, make a token sender supply ada later controlled by the recipient,
+and require ada in application state that does not participate in staking. High
+fan-out and successor-output top-ups amplify the funding and transaction-building
+burden.
+
+This CPS documents these frictions and their current workarounds, and sets goals for
+reducing them while preserving protection of the live UTxO set.
+
+## Problem
+
+### The current minUTxO rule
+
+Under the current Babbage and Conway rule, every newly created output `o` must
+satisfy [[1]](#ref-1):
+
+```math
+\mathrm{coin}(o)
+\geq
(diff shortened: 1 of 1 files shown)


CPS-????/README.md
```diff
@@ -0,0 +1,336 @@
+---
+CPS: "?"
+Title: Application and transaction-builder friction in the current minUTxO implementation
```
rphair, line 3:
```suggestion
Title: Transaction friction in minUTxO implementation
```
I would set a goal for the upcoming CIP meeting (at the latest) to confirm a more concise title & will make an attempt here.  The current phrase has an excellent narrative of the present problem but we need more of a "headline" to attract broad review: it's not (and usually isn't) just about the number of characters.
* The term `current` is never applicable to CIPs and CPSs, which last forever.
* I don't think there's any point avoiding the word `friction` since it's used practically everywhere in the document.  (I would prefer a more literal term but all synonyms used in software development are generally other metaphors.)
* Please correct me if I'm wrong but the `application` friction is from transaction handling and therefore this term is redundant to `transaction`.
* My understanding is that minUTxO is not a "parameter" that would be quoted like `minUTxO` and so we are talking about a concept rather than a literal value: and so it's proper to leave it unquoted as you already have.

According to usual review workflow I'm trying out this revised title in the PR itself so we can get a better view of how it might work (or not) and of course we'll keep trying updates @nhenin @polinavino @willjgould according to your own view.

rphair (commented):
Thanks @nhenin @willjgould @polinavino for another contribution from your rapidly & significantly evolving Ledger stack.  I would look forward to confirming this from `Triage` at our next CIP meeting (https://hackmd.io/@cip-editors/143) so contributors & editors alike can invite others to review it in parallel with its contemporary & related proposals.

@lehins of course the high-level Ledger view would be welcome at this early stage and would help to confirm it next Tuesday (or perhaps wait if something significant needs to be changed first): with input welcome from @aniketd re: #1264 + @carlostome as its reviewer.

Since the nature of this problem resembles the need for the Account Address Enhancement ([CIP-0159](https://github.com/cardano-foundation/CIPs/tree/master/CIP-0159)) we should seek out review from its own authors, contributors & ongoing reviewers: @fallen-icarus (especially for these business / economic cases) @GeorgeFlerovsky @colll78 @Quantumplation @gitmachtl @mpizenberg + @AlexDochioiu for the wallet perspective.

fallen-icarus:
IIUC a viable solution to this was designed at the most recent builderfest, though no CIP has been published on it yet. cc @matteocoppola @lehins @crypto2099

CPS-????/README.md
nhenin: Thanks @rphair. I agree with shortening the title and removing “current”. How about:

**Application and transaction friction in minUTxO implementation**

I would retain “application” because some consequences affect the application’s business flow: for example, a token sender must supply ada that the recipient subsequently controls, or switch to a recipient-funded claim flow. Keeping that term makes this dimension visible alongside transaction construction.

rphair: Thanks for that & I don't think the usual goal of short titles would justify the coherence lost in abbreviating `Transaction` as `Tx`.  I'll update the PR title as such & if we get more feedback @ the meeting tomorrow I'll update it... in the meantime feel free to commit this:
```suggestion
Title: Application and transaction friction in minUTxO implementation
```
In case this also comes up, I should also say I first thought `friction from minUTxO` might be sufficient & then noted (again, correct me if I'm wrong) that the problem isn't with a minimum UTxO value itself but in the way this minimum is implemented.

nhenin:
> IIUC a viable solution to this was designed at the most recent builderfest, though no CIP has been published on it yet. cc @matteocoppola @lehins @Crypto2099

Thanks @fallen-icarus. I've already discussed this with the people you mention, and the ongoing work follows up on the BuilderFest discussions.

We are currently exploring the implications in the Ledger codebase, particularly the output value split and transaction-level deposit netting....

CPS-????/README.md
```diff
@@ -0,0 +1,336 @@
+---
+CPS: "?"
```
```suggestion
CPS: 37
```

rphair (commented):
@nhenin this was confirmed as a candidate at the CIP meeting today in which the prior community discussion that @fallen-icarus mentions in https://github.com/cardano-foundation/CIPs/pull/1268#issuecomment-5666582930 was also highlighted again:
* @Crypto2099 suggested (and the editors would welcome) going into full detail about the Use Cases that you two (plus any others) have discussed already.
* Additional discussion from Buidler Fest or afterward about _any_ of the context of this problem would be welcome to add... the feeling from the meeting today is that filling this out with more community-sourced detail would be better than going forward with a bare CPS of mainly the outline points.

@ch1bo also highlighted the importance of what we would generally ask for — ensuring that the CPS doesn't argue too much in favour of a single solution — in this case, to help prepare for solutions that aren't simply removing or reducing the minUTxO requirement.

In the meantime please rename the containing directory to `CPS-0037` and update the "rendered CPS" link in your OP 🎉

PR #1270, opened 2026-09-16
CPS-???? | Trustworthy Off-chain Message Dissemination

will-break-it:
This CPS describes a gap in how Cardano participants exchange time-sensitive messages off chain: stake pool incident alerts, pool announcements, governance updates and dApp notifications. Existing channels carry them, but no common standard combines publisher authentication against Cardano identities, integrity, and quantified resistance to targeted suppression.

It sets out four use cases and six goals, including resistance to targeted peer selection and practical per-node cost. Its open questions cover adversarial participation, delivery targets, topic populations, small topics, participation costs, and behaviour during a chain halt or fork. A proposed solution follows as a separate CIP pull request.

The statement comes out of a research workstream on trustworthy message dissemination for Cardano, including an analysis of SecureCyclon under a silent adversary and a use-case survey with population estimates to validate with operators. Both are linked from the References section.

[Rendered from this branch](https://github.com/cardano-scaling/CIPs/tree/cps-offchain-message-dissemination/CPS-%3F%3F%3F%3F) ([upstream version](https://github.com/input-output-hk/pubsub/blob/main/docs/cps/README.md))


CPS-????/README.md
```diff
@@ -0,0 +1,144 @@
...
+Status: Open
+Authors:
+    - Will Wolff <william.wolff@iohk.io>
+    - Ezequiel Postan <ezequiel.postan@iohk.io>
+    - Denis Firsov <denis.firsov@gmail.com>
+    - Jesus Diaz Vico <jesus.diaz.vico@gmail.com>
+    - Dana Alibrandi <dalibrandi@gmail.com>
+    - Mauro Jaskelioff <mauro.jaskelioff@iohk.io>
+Proposed Solutions:
+    - Decentralised Pub/Sub Message Dissemination: https://github.com/cardano-foundation/CIPs/pull/1271
```
rphair, line 14: We should make a mental note to add the CIP tag into this label (when it's confirmed, probably about 2 weeks from today) and maybe shorten the label if possible since it appears horizontally — with any other labels — in most contexts.

CPS-????/README.md
```diff
@@ -0,0 +1,144 @@
...
+  operators, authenticating participants by their operational certificates so that Sybil
+  resistance follows from active stake. It states no delivery guarantee and no resistance to
+  targeted censorship, which is what this statement asks a solution to supply.
+- The broader survey the four scenarios were drawn from: <https://github.com/input-output-hk/pubsub/blob/5d6391813904159a04908dbdd20b03d1a56f85d1/docs/actor-use-case-analysis.md>
+- *PubSub Technical Report 1: Three-Layer Stack Findings and a Path Forward* — the evaluation
+  that led to this statement:
+  <https://github.com/input-output-hk/pubsub/blob/5d6391813904159a04908dbdd20b03d1a56f85d1/docs/technical-report-1.md>
+
+### Method notes
+
```
```suggestion
```
Most simply: this should be removed because footnotes generate no markup in source order... so this appears just as a blank section.

IF items used as footnotes also need to be visible with arbitrary markup as a reference list, that's possible as discussed in another document here (https://github.com/cardano-foundation/CIPs/pull/1242#pullrequestreview-4910250239) but I think it looks much better the way you have... and most readers will correctly assume that these references are notes on considerable methods.

rphair (commented):
Thanks @will-break-it & @EzequielPostan @dfirsov @jesusdiazvico @dalibran @mjaskelioff — I believe the CIP review community will be very interested in both the problem definition & its currently offered solution.

I expect most reviewers of either the CIP or the CPS will jump over to the other one, so I'll just tag reviewers for related `Network` projects in the CIP thread... I don't want to bias the CPS review toward the DMQ-based CIP in particular.

In any case we'll bring them up together for confirmation in `Triage` at the next CIP meeting: https://hackmd.io/@cip-editors/144 where you, co-authors & advocates would we welcome to introduce them.

FYI @will-break-it for both these PRs I've had to re-anchor the "rendered" link in the OP to the latest reviewable version in the PR branch itself... this is vital for reviewers to observe current changes as you improve the documents & as they begin to incorporate review.
* As changes might migrate back & forth between here & the upstream versions, please remember not to "force push" here unless it includes the whole commit history that's come from review in this branch. 🙏

CPS-????/README.md
will-break-it: Applied in dd636f4 — heading removed, the footnotes render at the bottom as you describe. Thanks for the pointer to #1242 as well; the plain footnote form reads better here, so I kept it.

fallen-icarus:
An open question I have is "where does PoW fit in?" There is precedence of messaging protocols using PoW for DDoS protection and Sybil resistance. For instance, imagine a reputation system combined with PoW: the higher your reputation, the less PoW you need to compute to pass along your message to the next node. The lower your reputation, the more PoW compute you need to pass along your message. 

PoW could compliment a deposit system. The reputation system could be entirely local: I score my own neighbors in the network and occasionally search for longer range ones. It is like the gossip protocol for mainnet with cold/hot peers.

You can even use PoW for urgency signaling: the more work you do, the more urgently the network can propagate your message. IIRC the Tor Browser does something like this.

